PR #655, opened 2013-06-17
support fish shell in client configure command

94c3:
these changes add support for the fish shell in the client configure
command. the main additions are a printer for exporting enviroment
variables (opamConfigcommand:print_fish_env) and a modification to the
dot file finder heuristic (opamMisc:guess_dot_profile).


dongcarl:
The configuration doesn't work as of now: #2255

PR #1934, opened 2015-01-08
Filter commits so the curves are increasing

Chris00:


Chris00:
The resulting images:
contributors
packages
unique-packages

Chris00:
BTW, this command takes way too long (30 mins) and uses way too much memory (1.7 Gb).  Is there a reason why the data is not collected *per commit* and the accumulation done by the program (which should allow sharing memory between the buckets).

samoht:
>  Is there a reason why [...]

lazyness :p you are very welcome to improve my crappy scripts!

avsm:
Looks significantly better!  Would it be possible to start from the first data point to avoid the big import at 10/2012?

Chris00:
New versions:
contributors
packages
unique-packages

Chris00:
@samoht OK.  I'll see if I have some time...

avsm:
Hurrah, that looks great!

PR #1967, opened 2015-01-29
switch with OPAMSWITCH set: fix generated variables script

AltGr:


AltGr:
It fixes one of yesterday's patches :)

But in 1.2, `opam config env` ignores any OPAMSWITCH set. This can get quite confusing:
```
say OPAMSWITCH=3.12.1
$ opam switch system # switches globally, so OPAMSWITCH has little importance
... ok, now run `eval $(opam config env)`
$ eval $(opam config env)
... now the env is correct for "system"... except that OPAMSWITCH is still set
$ opam switch
... reports 3.12.1
```

now (since yesterday), the interaction would be:
```
say OPAMSWITCH=3.12.1
$ opam switch system # still switches globally
... ok, now run `eval $(opam config env --switch system)`
$ eval $(opam config env --switch system)
... now the env is correct for "system"... OPAMSWITCH is set to "system"
$ opam switch
... reports system, which is set globally, set locally and in sync with the env.
```

also `opam switch` is now much more helpful when you have such variables set or your env is not in sync (that's from a few weeks ago).
what I had forgotten is the variables.sh update which wouldn't use the proper switch in the interaction above, fixed by this patch.

PR #1983, opened 2015-02-07
OpamParallel: allow to specify mutually exclusive jobs

AltGr:
Not used at the moment but will come in very handy for better handling
of the 'install' field.

### DIFF
--- a/src/core/opamParallel.ml
+++ b/src/core/opamParallel.ml
@@ -44,12 +44,14 @@ module type SIG = sig
   val iter:
     jobs:int ->
     command:(pred:(G.V.t * 'a) list -> G.V.t -> 'a OpamProcess.job) ->
+    ?mutually_exclusive:(G.V.t list list) ->
     G.t ->
     unit
 
   val map:
     jobs:int ->
     command:(pred:(G.V.t * 'a) list -> G.V.t -> 'a OpamProcess.job) ->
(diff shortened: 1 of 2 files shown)


src/core/opamParallel.ml
```diff
@@ -44,12 +44,14 @@ module type SIG = sig
   val iter:
     jobs:int ->
     command:(pred:(G.V.t * 'a) list -> G.V.t -> 'a OpamProcess.job) ->
+    ?mutually_exclusive:(G.V.t list list) ->
```
samoht: can this be a function instead of a list of lists? eg:
```
?mutually_exclusie:(G.V.t -> G.V.t -> bool)
``
would be easier to use I think.

AltGr: Well we are working on finite sets, supplied as graphs, so that felt consistent. I would have preferred to add specific undirected edges to the graph actually, but that would be too heavy just for this use.

Also we would need some kind of guarantee that the supplied function is symmetric, and computing the exclusion sets could get more expensive.

As for the expected usage: it's just to make all 'install' actions mutually exclusive at the moment (so that for example (i) we can do some diff on the prefix to register what was installed (ii) the install step of a package may be interactive). So the argument would be just `[[all;install;actions]]` which is quite simple.

PR #1985, opened 2015-02-09
Beta version of the 'features' field

AltGr:
The intent is to have this included in 1.2.1 as a "beta" feature, i.e. not
allowed yet in the repository, but there for people to test (and maybe for
forwards-compat, in case it stays as is, but I wouldn't count on it).

`opam lint` will warn about this.


AltGr:
Closes #1325

Drup:
weee \o/

AltGr:
At the moment features can take any value (bool or string) but they should probably be restricted to bools. Also, they're not automatically advertised yet (unless you manually add a `messages` field) so their docstring is mostly useless.

PR #1997, opened 2015-02-13
Allow <name>.opam in the source of packages

AltGr:
Convenient when there are several packages compiled from the same repo

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -34,6 +34,7 @@
   correct OPAM path
 * ~/.opam/config doesn't refer to OPAM's patch-version anymore, to allow
   downgrading
+* Recognise <name>.opam files and directories when pinning a package to source
 * Lots of bug-fixes
 
 1.2.0
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -1542,7 +1542,18 @@ let pin ?(unpin_only=false) () =
         | Some name -> name
         | None -> raise Not_found
       with e -> OpamMisc.fatal e; raise Not_found
-    else raise Not_found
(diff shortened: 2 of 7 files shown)


avsm:
Can this be a provisional feature for 1.2.1 :)

dbuenzli:
Could you please document the exact behaviour in `opam pin --help`

AltGr:
@avsm: offers more options when pinning, but doesn't affect the repo and base usage, so that should be OK.
@dbuenzli: sure, will do. The idea is that when pinning `pkg` to `dir`, instead of just looking for an opam file at `dir/opam` (or `dir/opam/opam`), it will prefer a `dir/pkg.opam` (or `dir/pkg.opam/opam`).

AltGr:
So unless you have `*.opam` files lying around, previous behaviour should be unchanged.

dbuenzli:
Yes, I read the code... I'm just being annoying so that the docs remain up-to-date (also often trying to write documentation of features first/simultaneously is a good usability test: if you can't explain it well you certainly have something wrong). Make sure the priority list for selection is explicitly documented. 

That being said in the subopams case we now have a strong link between the name we pin to and the opam files that need to be present. This is not the case if you don't have subopams, you can do for example `pin add fancy xmlm-repo` (vs. the expected `pin add xmlm xmlm-repo`) which may have its uses (does it ?). So if you face a repo with subopams you cannot do that. This asks whether maybe an `--opam-file=PATH` where `PATH` is a file path relative to the pinned repo should be added to be able to explicitly tell which opam file you want to select.

PR #1999, opened 2015-02-13
fix failure when the system compiler has a '+' in its version string

### LINKED ISSUE #1946: compiler version printing
When I tried to installed codoc, I got the message: 
`
opam-doc-base is not available because it requires OCaml = 4.02.0.
`

When I did `opam show --raw opam-doc-base` I got:
```
opam-version: "1"
name: "opam-doc-base"
version: "0.1.3"
maintainer: "lpw25@cl.cam.ac.uk"
authors: "Leo White <lpw25@cl.cam.ac.uk>"
license: "ISC"
build: [
  ["assemblage" "setup" "--dumpast=false"]
  [make]
  [make "install"]
]
remove: ["ocamlfind" "remove" "opam-doc-base"]
depends: [
  "cmdliner"
  "xmlm"
  "opam-lib"
  "opam-units"
  "ocamlgraph"
  "ocamlfind"
  "assemblage"
]
ocaml-version: [= "4.02.0"]
```

However, as you can see [here](https://github.com/ocaml/platform-dev/blob/master/packages/opam-doc-base/opam-doc-base.0.1.3/opam). The compiler version is not right. It seems that `+doc` is not printed.

planar:
This bug is a consequence of a2f6ca121b942447086549a1b45cd9332684cd43, which fixes #1946.

All the Travis builds of OCaml are currently failing because opam (wrongly) refuses compiler version strings that contain '+'.

### DIFF
--- a/src/core/opamCompiler.ml
+++ b/src/core/opamCompiler.ml
@@ -25,20 +25,26 @@ module Version = struct
 
   let of_string str =
     if OpamMisc.contains str '+' then
-      raise (Invalid_argument "'+' is not allowed in compiler versions");
+      raise (Invalid_argument
+               "'+' is not allowed in compiler version constraints");
     of_string str
 
   type constr = (OpamFormula.relop * t) OpamFormula.formula
 
+  let cut_version_number str =
+    match OpamMisc.cut_at str '+' with
+    | None -> str
+    | Some (s, _) -> s
+
   let current () =
     match Lazy.force OpamSystem.ocaml_version with
     | None   -> None
-    | Some o -> Some (of_string o)
+    | Some o -> Some (of_string (cut_version_number o))
 
   let system () =
     match Lazy.force OpamSystem.system_ocamlc_version with
     | None   -> None
-    | Some v -> Some (of_string v)
+    | Some v -> Some (of_string (cut_version_number v))
 
   let compare v1 v2 = Debian.Version.compare (to_string v1) (to_string v2)
 


AltGr:
Hm, if this is a problem, we need to sort it out (before 1.2.1). I think having comparison on compiler versions behave differently from comparison on package versions is a pretty bad idea.
* I have no objection on allowing `+` back in compiler versions, but then we need to be clear on the distinction between `name` and `version`: the first is the label of this specific compiler, the second the general OCaml version that it provides. Follows that a package specifying `ocaml-version = "4.02.1"` won't be installable on a compiler with `version: "4.02.1+dev72"` ¹. 
* @bactrian (https://github.com/bactrian/opam-sync-github-prs/blob/master/generate.ml) correctly generates the `version:` field to match the OCaml version without suffix ; I am not clear how this specific issue surfaced from OCaml's tests, wasn't it a bug in opam-repo ?
* Previous OPAM versions simply truncated the `.comp`'s `version:` field on file load, so not using `+` in there gives 100% compatibility with those.

¹ Well, unless we change the global meaning of the version comparison `=` to mean "the shortest of the two is a prefix of the longer one". Which would be in this case equivalent to `ocaml-version >= "4.02.1" & ocaml-version < "4.02.2"` (not considering `~` versions). While this is a bit verbose, this change is a bad idea IMHO because the semantics are too complicated. Adding some fancy syntax like `ocaml-version = "4.02.1*"` or `ocaml-version =~ "4.02.1"` for such cases could be an idea, but doesn't seem worth the added complexity either.

dbuenzli:
Could you provide a reference on the version comparison opam implements for packages ? I'm sure you provided doc about this at a certain point but I can't find it anymore.

AltGr:
It's in the [FAQ](opam.ocaml.org/doc/FAQ.html#HowdoesOPAMtellwhichversionofapackageisnewer)

dbuenzli:
> I think having comparison on compiler versions behave differently from comparison on package versions is a pretty bad idea.

Especially if compiler as packages become a reality.

AltGr:
Ah! I finally located the real source of this error (thanks to the "allowed" failure at https://travis-ci.org/ocaml/opam/builds/50899181): OPAM _itself_ generates the `version:` field for the system compiler from running `ocamlc -version`, leading to this error when you have a development OCaml preinstalled. In that case, the name in `compiler` is just `system` and gives no information about the precise version.

Thanks a lot @planar for noticing this and pushing me in the right direction.

This is quite messy, I'll try to find a sane way to solve this:
* `version:` should, consistently, be the patch-version of OCaml (without suffix)
* We have the `preinstalled` boolean variable so having `system` as the compiler name is not needed, besides not really making sense. There may be checks for "system" lurking around though, we'll need to be careful.
* Just truncating `version:` would be consistent with the old behaviour (if rather itchy, and losing the detailed version information)

AltGr:
Just filed a PR to fix this, at least on the surface since we're already in beta stage: the version reported by `ocamlc -version` will simply be truncated in the generated system compiler definition. I'm not completely happy, since the complete compiler version isn't stored there.

Note that this means that a change in suffix of the system OCaml version won't trigger a reinstall. I believe that was already the case (well, we can suppose people installing dev versions of OCaml know that it happened and can do a `switch reinstall` manually I guess)

I would like, as mentionned above and for consistency, to have the system compiler be defined with its complete version in the name rather than just "system" ; but that would be quite a change, and raises some concerns (the name is used as a key throughout, and you'd need a way to specify wether you want the system switch or the source switch with the same version when creating a new switch). This is not a problem after installation since compilers are referred to by their aliases.

There is also some special magic that allows you to write tests like `ocaml-version = "system"` (rather than `compiler = "system"`, or just `preinstalled`), that I would be very happy removing.

planar:
> Follows that a package specifying ocaml-version = "4.02.1" won't be installable on a compiler with version: "4.02.1+dev72"

Note I'll be very sad if you do that, because I won't be able to use an unmodified OPAM to test development versions of the compiler. Unless you provide a switch to truncate the compiler version.

AltGr:
@planar: yes, that's precisely why I wanted to avoid it. You should be fine now, same behaviour as before except that when you write a .comp by hand, you should explicitely not put `+xxx` in the version, rather than having silently ignored (almost the same as your proposed patch, actually, but with fixed system-compiler version change detection).

You won't yet be able to easily access the complete `4.02.1+dev72` string from OPAM though.

PR #2000, opened 2015-02-14
Fix installation of base packages on "opam init"

AltGr:


AltGr:
And here is PR #2000 :)

PR #2017, opened 2015-02-19
ad hoc fix for csh (more general problem remains)

esumii:
"string_of_env_update" is using OCaml's %S for escaping strings in shell scripts!

### DIFF
--- a/src/client/opamState.ml
+++ b/src/client/opamState.ml
@@ -1970,7 +1970,7 @@ let update_ocamlinit () =
 let string_of_env_update t shell updates =
   let fenv = resolve_variable t OpamVariable.Map.empty in
   let sh   (k,v) = Printf.sprintf "%s=%S; export %s;\n" k v k in
-  let csh  (k,v) = Printf.sprintf "setenv %s %S;\n" k v in
+  let csh  (k,v) = Printf.sprintf "if ( ! ${?%s} ) setenv %s \"\"\nsetenv %s %S\n" k k k v in
   let fish (k,v) = match k with
     | "PATH" | "MANPATH" ->
       let to_space_sep = String.concat " " (OpamMisc.split v ':') in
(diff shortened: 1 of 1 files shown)


AltGr:
Indeed, that was kind of a lazy assumption. Can you give some more background on the issue that you had and this fixes ? Thanks!

src/client/opamState.ml
```diff
@@ -1970,7 +1970,7 @@ let update_ocamlinit () =
 let string_of_env_update t shell updates =
   let fenv = resolve_variable t OpamVariable.Map.empty in
   let sh   (k,v) = Printf.sprintf "%s=%S; export %s;\n" k v k in
-  let csh  (k,v) = Printf.sprintf "setenv %s %S;\n" k v in
+  let csh  (k,v) = Printf.sprintf "if ( ! ${?%s} ) setenv %s \"\"\nsetenv %s %S\n" k k k v in
```
AltGr: Why is this needed ? I am curious how it differs from the simple `setenv %s %S`...

esumii: There are actually two issues (sorry!); as for this line, one of my environments (admittedly old!) uses (t)csh, which raises an _error_ for $PERL5LIB when PERL5LIB is not set yet.

src/client/opamState.ml
```diff
@@ -1989,7 +1989,7 @@ let string_of_env_update t shell updates =
       | "+="
       | ":=" -> (ident, Printf.sprintf "%s:$%s" string ident)
       | "=:"
-      | "=+" -> (ident, Printf.sprintf "$%s:%s" ident string)
+      | "=+" -> (ident, Printf.sprintf (match shell with `csh -> "${%s}:%s" | _ -> "$%s:%s") ident string)
```
esumii: The reason for this change is that (t)csh raises an _error_ for "$MANPATH:..." since ":" is interpreted as a modifier for the variable reference "$MANPATH".

However, it fails to type-check on OCaml 3.12.1 (sorry again!).  A possible alternative would be:

(match shell with `csh -> Printf.sprintf "${%s}:%s" ident string | _ -> Printf.sprintf "$%s:%s" ident string)

src/client/opamState.ml
```diff
@@ -1989,7 +1989,7 @@ let string_of_env_update t shell updates =
       | "+="
       | ":=" -> (ident, Printf.sprintf "%s:$%s" string ident)
       | "=:"
-      | "=+" -> (ident, Printf.sprintf "$%s:%s" ident string)
+      | "=+" -> (ident, (match shell with `csh -> Printf.sprintf "${%s}:%s" ident string | _ -> Printf.sprintf "$%s:%s" ident string))
```
esumii: This line is changed since "$MANPATH:..." raises an error in (t)csh, where ":" means a modifier for the variable reference "$MANPATH".  (This is a second try, as the previous change didn't type-check in OCaml 3.x.)

AltGr:
Ok, that all makes more sense, thanks !

esumii:
I don't think this Travis failure is my fault...:-)

AltGr:
Nope, it's not. Thanks!

PR #2018, opened 2015-02-19
Reset console paramaters after a CTRL-C

samoht:
Fix #2013

PR #2024, opened 2015-02-21
Fixed confirm function for read() interruption

AltGr:


avsm:
Was this causing corrupted terminal state when ctrl-c'ing a y/n prompt?  If so, hurrah on the fix :)

AltGr:
I think Thomas had fixed those already; I had this break out of the confirmation dialog when switching shell tabs, probably a SIGWINCH.

avsm:
Aha yes in #2018

PR #2037, opened 2015-02-27
Back out the optimisation to recompile only when archive checksum changes

avsm:
The observable bug is that when I update the `url` file in a package
to change the `archive` source but not the checksum, this never
propagates after an `opam update`.  The reason is that the update
is skipped due to the optimisation that prevents recompilation of a
package when the location changes, but not the checksum.

Unfortunately, if using OPAM with a `local` remote, this makes it
impossible to fix a broken location without manually munging around
in `~/.opam`.

Debugged with @samoht


AltGr:
Indeed, that shouldn't be done at this level ; and the useful part is already done higher up in [OpamState](https://github.com/ocaml/opam/blob/ab33f70642aef373d102ddf7af52c1c051c31756/src/client/opamState.ml#L186) so I don't think there will be any loss in optimisation.

AltGr:
Superseded by #2037

PR #2038, opened 2015-02-28
Back out the optimisation to recompile only when archive checksum changs

AltGr:
The observable bug is that when I update the `url` file in a package to change
the `archive` source but not the checksum, this never propagates after an
`opam update`.  The reason is that the update is skipped due to the
optimisation that prevents recompilation of a package when the location
changes, but not the checksum.

Unfortunately, if using OPAM with a `local` remote, this makes it impossible
to fix a broken location without manually munging around in `~/.opam`.

Debugged with @samoht


AltGr:
Argh: this actually causes OPAM to detect ALL packages as changed ALL the time. Guess that avoids the bug, but...

PR #2040, opened 2015-02-28
Allow custom commands for solver and downloads through $opam/config or the environment

AltGr:


AltGr:
Travis failure fixed in opam-rt

PR #2041, opened 2015-03-02
Finer verbosity control

### LINKED ISSUE #2034: opam -v is too verbose
I think `opam install -v` is too verbose when downloading distfiles in 1.2.1dev.  It would be useful to have a `-v` which only shows package installation output, and perhaps have a `-vv` or other more verbose step which shows all the various `mv` and `cp` commands being executed.

AltGr:
Closes #2034


avsm:
lgtm!

PR #2075, opened 2015-03-23
Don't error out with --no-switch flag if compiler is installed

### LINKED ISSUE #1813: `opam switch doc -A system --no-switch` gives an error if the compiler exists
`opam switch doc -A system --no-switch` gives an error if the compiler exists, which the documentation for `--no-switch` explicitly says it will not.

We use this in the old opam-doc prototype, which no longer works with opam 1.2. (It will also soon be deprecated so this is not particularly important).

anemator:
Fixes #1813.

### DIFF
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -1466,6 +1466,7 @@ let switch =
     | Some `current, [] ->
       Client.SWITCH.show ();
       `Ok ()
+    | Some `set, [switch]
     | Some `default switch, [] ->
       (match alias_of with
        | None ->
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -1468,19 +1468,12 @@ let switch =
       `Ok ()
     | Some `set, [switch]
     | Some `default switch, [] ->
-      (match alias_of with
-       | None ->
-         Client.SWITCH.switch ~quiet:global_options.quiet ~warning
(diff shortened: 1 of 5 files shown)


AltGr:
Thanks!

The check on the dirname existence is not just here to check if the switch exists though, but also to prevent accidental overwrite: swiches directories are at `~/.opam/` and share the namespace with directories like `repo` and `packages` -- we really want to fail in case someone tries `opam switch install repo http://xxx`.

According to the doc, `opam switch install x` _should_ fail when the switch `x` is already installed, while `opam switch x` (which is an alias for `opam switch set x`) should continue. I think what is really missing here is that the `--no-switch` option isn't honored by `switch set`.

AltGr:
There could be a test on the switch existence before the check for the directory existence, still, though.

anemator:
Thanks for the info!

```--no-switch``` doesn't seem applicable to ```switch set``` according to the documentation: "Set the currently active switch, installing it if needed." Ideally, I'd like to ignore ```--no-switch``` on ```switch set```, but I'm concerned that will break user scripts.

AltGr:
> Ideally, I'd like to ignore --no-switch on switch set, but I'm concerned that will break user scripts.

I think that's how it behaves at the moment.

AltGr:
Thanks a bunch, all good, sorry for the lag!

PR #2084, opened 2015-03-24
add new export format for opam compatible with dose-distcheck

abate:
This set of patches add an alternative format dump for opam (PEF : package export format). It is meant to be the new interface between the dose tools and opam to perform repository analysis. It is also needed for the new version of ows.irill.org, the upcoming new version of the opam weather service. This format is different from cudf as versions are strings and not integers. In the future I plan to add more fields to the export format.

Ex:

   package: cohttp
   version: 0.9.14
   maintainer: anil@recoil.org
   depends: fieldslib (>= 109.20.00) , cstruct (>= 1.0.1) , ounit , uri (< 1.5.0) , uri (>= 1.3.8) , re , ocamlfind
   reccomends: lwt | async
   conflicts: mirage-tcpip-xen (< 0.9.5) , mirage-tcpip-unix (< 0.9.5) , lwt (>= 2.4.7) , lwt (< 2.4.3) , async (> 109.53.02) , async (< 109.15.00)

### DIFF
--- a/src/client/opamState.ml
+++ b/src/client/opamState.ml
@@ -1185,7 +1185,8 @@ let dump_state t oc =
 
     (try
       let c = OpamPackage.Map.find package conflicts in
-      match OpamFormula.to_conjunction c with
+      let n = OpamPackage.name package in
+      match (n,None)::(OpamFormula.to_conjunction c) with
       |[] -> ()
       |cc -> Printf.fprintf oc "conflicts: %s\n"
               (string_of_conjunction OpamFormula.string_of_atom cc);
--- a/src/client/opamState.ml
+++ b/src/client/opamState.ml
@@ -1149,41 +1149,47 @@ let dump_state t oc =
   let conflicts = OpamPackage.Map.map OpamFile.OPAM.conflicts opams in
   let maintainers = OpamPackage.Map.map OpamFile.OPAM.maintainer opams in
   let base = base_packages t in
+  let filter _ = true in
 
   let aux package =
     Printf.fprintf oc "package: %s\n" (OpamPackage.name_to_string package);
     Printf.fprintf oc "version: %s\n" (OpamPackage.version_to_string package);
-    begin try
+    (try
       let m = OpamPackage.Map.find package maintainers in
-      Printf.fprintf oc "maintainer: %s\n" (string_of_conjunction (fun a -> a) m);
-    with Not_found -> () end;
+      Printf.fprintf oc "maintainer: %s\n"
+        (string_of_conjunction (fun a -> a) m);
+    with Not_found -> () );
+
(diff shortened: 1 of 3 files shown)


src/client/opamState.ml
```diff
@@ -1125,6 +1125,69 @@ let universe t action =
...
+      |dd -> Printf.fprintf oc "depends: %s\n" (string_of_cnf OpamFormula.string_of_atom dd)
+    with Not_found -> () end;
+
+    begin try
+      let d = OpamPackage.Map.find package depopts in
+      let formula = (OpamFormula.formula_of_extended ~filter:(fun _ -> true) d) in
+      match OpamFormula.to_cnf formula with
+      |[] -> ()
+      |[[]] -> ()
+      |dd -> Printf.fprintf oc "reccomends: %s\n" (string_of_cnf OpamFormula.string_of_atom dd)
```
AltGr: typo ?

AltGr:
Nice!

I am not sure OpamState is the right place to put it though (it's already bloated) ; since we already have a conversion from OpamState.state to OpamTypes.universe (done by the function `OpamState.universe`), then to the Cudf universe (function `OpamSolver.load_cudf_universe`), it would feel more in place to have the PEF generated at the same place as the Cudf, from the `universe` type: in OpamSolver or in a new module beside it (as the formats are close, it may even be possible to reuse most of the same code without duplicating) . Since it is the place where the solver is called, it will also be smoother if we want to allow choosing the format between Cudf and PEF for solving sometime in the future.

There will be information missing in the `universe` type, but it may be extended (with lazy fields in case the information is costly and unused in the general case).

There are no strict code style guideline in OPAM, but I generally try to keep to 80 columns, put a space after the `|` in pattern-matchings, and don't use `begin..end` much.

Sorry to be so annoying! I'll be very glad when we can handle the version ordering at a lower level, the conversion to integers has proven error-prone and makes reporting a bit less accurate.
Thanks!

abate:
hi, my first version of the was indeed written alongside the cudf converter. I then realized that there were many fields missing from the OpamState.universe type. The idea would be to eventually add all fields available in the opam state to the pef export. While pef could be used instead of cudf (a bit in the same way dose interact with apt-get), it was not my goal to replace the cudf interface. Cudf is tailored to interact with the solver, while PEF as a format to interact with all dose tools, and perform repository analysis not only related to dependency solving. I don't think you want to duplicate all opam state fields in the universe DS. I'll fix the typo and reformat the code to fit 80 columns.

AltGr:
Nice, thanks!
That would probably deserve its own module actually (unless that would force to export too many values from OpamState) -- I really would like to split up OpamState at some point, actually, but never really get the time.

AltGr:
Thanks!

PR #2089, opened 2015-03-26
Fix incorrect contractions in messages.

rneswold:
These patches don't affect the logic of opam or its build environment. It simply fixes some spelling mistakes in messages.


AltGr:
Thanks! :)

PR #2105, opened 2015-04-07
Fix to build OPAM on NetBSD without any installed libraries.

rneswold:
Under NetBSD, un-tar'ing a project's archive may result in an extra file getting created. NetBSD's `tar` recognizes GNU `tar` archives but places extra information in a file outside the directory tree that gets extracted. The extra file breaks the `Makefile` command that tries to move the archive's tree out of `tmp`.

This patch replaces the `mv` command with a slightly more complicated command that only moves directories out of the `tmp` directory. Because we may be leaving a file in `tmp`, the `rmdir` command is replaced with a `rm -rf` command.

This commit allows me to build OPAM on NetBSD with no libraries installed in the base system.


AltGr:
Thanks !

PR #2107, opened 2015-04-11
Add ./admin-script/extract-mini-repository

hnrgrgr:
This script is able to generate a minimalistic and self-contained 
opam-repository for a given set of packages and compiler versions.
The resulting repository will be in './opam-mini-repository.tar.gz'.

```
## Including only the latest compiler version:
./extract-mini-repository ocp-indent lwt

## Including a specific version of OCaml or multiple versions.
./extract-mini-repository ocaml.4.02.1 ocaml.4.01.0 ocp-indent lwt
```

Then, the resulting repository can be used to replace the default opam
repository on computer without internet access and with only few spare
disk space.

```
tar xf opam-mini-repository.tar.gz
opam init --root /tmp/opam default ./opam-mini-repository --comp 4.02.1
```

### DIFF
--- a/admin-scripts/extract_mini_repository.sh
+++ b/admin-scripts/extract_mini_repository.sh
@@ -1,0 +1,143 @@
+#! /bin/sh
+
+set -e
+
+if [ $# = 0 ]; then
+   cat <<EOF
+
+This script is able to generate a minimalistic and self-contained
+opam-repository for a given set of packages and compiler versions.
+The resulting repository will be in './opam-mini-repository.tar.gz'
+
+Example:
+
+ ## Including only the latest compiler version:
+ $0 ocp-indent lwt
+
+ ## Including a specific version of OCaml or multiple versions.
+ $0 ocaml.4.02.1 ocaml.4.01.0 ocp-indent lwt
+
+Then, the resulting repository can be used to replace the default opam
+repository on computer without internet access and with only few spare
+disk space.
+
+  tar xf opam-mini-repository.tar.gz
+  opam init --root /tmp/opam default ./opam-mini-repository --comp 4.02.1
+
+EOF
+   exit 1
+fi
+
+DEFAULT_COMPILER=4.02.1
+OPAM_REPO=https://github.com/ocaml/opam-repository/archive/master.tar.gz
+REPO_DIR_NAME=opam-mini-repository
+
+EXE_PATH=$(readlink -f "$0")
+SOURCE_DIR=$(dirname "${EXE_PATH}")
+TARGET_DIR=$(pwd)
+WORK_DIR=$(mktemp -d)
+
+REQUEST=$*
+COMPILERS=
+
+for package in ${REQUEST}; do
+    case ${package} in
+        ocaml.*)
+            COMPILERS="${COMPILERS} ${package#ocaml.}"
+            ;;
+        *)
+            PACKAGES="${PACKAGES} ${package}"
+            ;;
+    esac
+done
+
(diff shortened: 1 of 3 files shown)


admin-scripts/extract_mini_repository.sh
```diff
@@ -0,0 +1,118 @@
...
+        *)
+            for version in $ARCHIVES; do
+                if [ ${version%+opam.tar.gz} = $1 ]; then return 1; fi
+            done
+    esac
+    return 0
+}
+
+for dir in $(ls -d packages/*/*); do
+    if unrequired_version ${dir##packages/*/}; then
```
avsm: is this substitution a bash-ism? You've used `/bin/sh` at the top of the script.

admin-scripts/extract_mini_repository.sh
```diff
@@ -0,0 +1,118 @@
...
+  return 0
+}
+
+for dir in $(ls -d compilers/*/*); do
+    if unrequired_compiler ${dir##compilers/*/}; then
+        rm -r $dir
+    fi
+done
+
+rmdir --ignore-fail-on-non-empty compilers/*
```
avsm: not portable beyond a GNU userland I think

admin-scripts/extract_mini_repository.sh
```diff
@@ -0,0 +1,118 @@
...
+
+EOF
+   exit 1
+fi
+
+DEFAULT_COMPILER=4.02.1
+OPAM_REPO=https://github.com/ocaml/opam-repository/archive/master.tar.gz
+
+TARGET=$(pwd)
+DIR=$(mktemp -d)
```
avsm: need a `set -e` somewhere, or if this fails then the `rm -rf` below may well delete `/`

avsm: Ignore previous, `set -e` is set at the top

hannesm:
(regarding shells, I really like http://www.shellcheck.net/ to evaluate how standard-conform a script :)

admin-scripts/extract_mini_repository.sh
hnrgrgr: I just checked and it is fully POSIX.

admin-scripts/extract_mini_repository.sh
hnrgrgr: Indeed, fixed in https://github.com/hnrgrgr/opam/commit/27bd4eaa882fefdb76cd983647d5744629f857db
Thanks.

hnrgrgr:
@hannesm Indeed there is lot of clever analysis in spellscheck ! And, it is also available in emacs and vim :-) Thanks for the link.

admin-scripts/extract_mini_repository.sh
michipili: It is definitely not a bashism! :)

AltGr:
Great, thanks !

michipili:
Hmmm I cannot find my comments any more, so I just respond here in the wild!

@hnrgrgr If you do not want learn **awk** in the night, you can learn it before going on lunch break! :D

The manual page is 2 pages long and is more than enough to be started.

   https://www.freebsd.org/cgi/man.cgi?query=awk&sektion=1&n=1

After this, one can start reading others scripts!

hnrgrgr:
Your comments are here : https://github.com/hnrgrgr/opam/commit/164b598ef1f3463073ee33fffbd898cb3bee748c

michipili:
Ahah, this was the *other* repository! Thank you! :)

PR #2113, opened 2015-04-16
Auto install (and run) `foo` on `opam foo` if it has `flags: plugin`

AltGr:


avsm:
See ocaml/opam-repository#3944 for an issue with this.

Issue #2119, opened 2015-04-18
lib-ext target does not compile without ocamlopt


eras:
This is due to lib-ext/Makefile compiling OCamlMakeFile target 'ncl' unconditionally even when OCAMLOPT=no in ../Makefile.config.

PR #2120, opened 2015-04-18
Don't compile lib-ext targets that cannot be compiled due to missing compiler #2119

eras:
So don't compile native libraries without ocamlopt, and don't compile bytecode
libraries without ocamlc.

This addresses the issue #2119.

I first bumped into the issue when compiling without ocamlopt, but decided to add ocamlc support for sake of completeness and to make testing easier. The symptoms when compiling lib-ext without ocamlopt are:

    no -c -I extlib/ extlib/enum.ml
    /bin/sh: 5: no: not found

### DIFF
--- a/src_ext/Makefile
+++ b/src_ext/Makefile
@@ -36,8 +36,8 @@ lib-ext: clone build copy
 	@
 
 build: clone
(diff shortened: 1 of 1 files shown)


AltGr:
Thanks!

PR #2125, opened 2015-04-20
OpamFile: make 'permissive mode' the default without strict or debug

AltGr:
Ref #3944

This makes most file format errors pass silently -- in particular unknown
fields or idents. Some "typing" errors (eg string where an ident is expected)
will lead to the whole field being ignored ; lint will report an error 2 "bad
format" in that case.

Also normalises some messages


AltGr:
ref: ocaml/opam-reoository#3944

PR #2126, opened 2015-04-20
Allow _tag_ `"flags:plugin"` instead of _flag_ `plugin`

AltGr:
ref: ocaml/opam-repository/issues/3944


avsm:
+1, also needs the developer manual to be updated

PR #2128, opened 2015-04-20
Change wording to reflect the command used

### LINKED ISSUE #2127: Change the first line of output from 'opam config report'
The first line of the output of `$ opam config report` should reflect the command.  Right now it reads `# OPAM status report`.  This confuses me every time I try to remember the incantation i the future, I end up trying variations of `$ opam status report` before realising what it should be.

IMHO it should be changed to `# OPAM config report`.

amirmc:
This should help to reduce any confusion in remebering what
the command is that generates this output.

Fixes #2127


amirmc:
I grepped the source for the line and only found this one location.  I wasn't very thorough so please do check before merging.

AltGr:
Thanks! That should do it.

And Travis complaining again on a string-only patch :'(

PR #2131, opened 2015-04-23
Pass the expected checksum to the OPAMFETCH program

planar:
I use a script to do caching on OPAM's download requests (to avoid pounding the servers, to work around overzealous firewalls, and to isolate myself from server failures and removed files). Unfortunately, the caching gets in the way when the contents of a file changes at the same URL.

In most cases, `opam` already knows about it because the checksum is updated in the package repository. This PR makes `opam` pass the checksum to the OPAMFETCH command (as `%{checksum}%`) which allows me to index my cache by checksums and avoid all problems with different files versions at the same URL.


AltGr:
Good idea, thanks!

PR #2134, opened 2015-04-26
This PR will solve a fd leak

ivg:
When `OpamProcess.command` is called with `allow_stdin:false` it will
create new channel for `stdin`, that will be duplicated in the child
process, but never closed in the parent.

The only place that currently disables stdin is in the `OpamCudf`
module.

Looks like, that this was the particular issue, that break BAP
installation ocaml/opam-repository#3994.

This issue starts to manifest on all our yosemite machines, when we're
trying to install BAP, after the aspcud was added to macports.


avsm:
Nice catch! I see this too on Yosemite with any large update.  Should definitely be in 1.2.2

AltGr:
Ow, thanks a bunch for spotting this !

PR #2136, opened 2015-04-27
Make opam-list exit non-zero only when patterns don't match

### LINKED ISSUE #2135: opam list --depends-on ${pkg} -s exits 1 when empty
Instead of just printing a newline and exiting 0 when there are no dependents, `opam list --depends-on ${pkg} -s` exits with code 1.

AltGr:
Fixes #2135


avsm:
merge into 1.2.2? helps with scripting.

AltGr:
Yup

PR #2161, opened 2015-05-22
Separate build and install steps

AltGr:
This uses a new separate action `Build in the concrete action graph,
making use of the build/install separation in opam files.

As a consequence,
- install step is unparallelised, which will allow for easily tracking changes
- the verbose package flag only prints install instructions, not the build anymore (needs doc!)
- the install step could be made to have access to stdin and query the user

### DIFF
--- a/src/solver/opamActionGraph.ml
+++ b/src/solver/opamActionGraph.ml
@@ -20,30 +20,36 @@ module type ACTION = sig
   type package
   module Pkg : GenericPackage with type t = package
   include OpamParallel.VERTEX with type t = package action
-  val to_aligned_strings: t list -> string list
+  val to_string: [< t ] -> string
+  val to_aligned_strings: [< t ] list -> string list
 end
 
-let action_strings ?utf8 x =
-  if utf8 = None && (OpamConsole.utf8 ()) || utf8 = Some true then
-    List.assoc x
-      [`inst,   "\xe2\x88\x97 ";
-       `up,     "\xe2\x86\x97 ";
-       `down,   "\xe2\x86\x98 ";
-       `reinst, "\xe2\x86\xbb ";
-       `rm,     "\xe2\x8a\x98 "]
-  else
-    List.assoc x
-      [`inst,   "install";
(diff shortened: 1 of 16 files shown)


src/solver/opamActionGraph.ml
```diff
@@ -20,30 +20,36 @@ module type ACTION = sig
...
+let name_of_action = function
+  | `Remove _ -> "remove"
+  | `Install _ -> "install"
+  | `Change (`Up,_,_) -> "upgrade"
+  | `Change (`Down,_,_) -> "downgrade"
+  | `Reinstall _ -> "recompile"
+  | `Build _ -> "build"
+
+let symbol_of_action = function
+  | `Remove _ -> "\xe2\x8a\x98 "
```
dbuenzli: Just a comment. I think it's nice when you write verbatim UTF-8 to indicate in a comment to which code point this corresponds to using the `U+XXXX` notation.

samoht:
> the verbose package flag only prints install instructions, not the build anymore (needs doc!)

why? verbose builds are very useful too!

AltGr:
`--verbose` is for everything
`flags: verbose` is intended to let the package display some messages directed to the user: it's nice to have a way to display those at the end, without making the build verbose (_e.g._ for user-setup) .

That's my use of the `verbose` package flag, but if you had other ones, we can add more options!

src/solver/opamActionGraph.ml
AltGr: indeed, good suggestion, thanks

dbuenzli:
> flags: verbose is intended to let the package display some messages directed to the user: it's nice to have a way to display those at the end, without making the build verbose (e.g. for user-setup) .

I'm not sure I get this. Isn't this handled by the `message:` field ?

michipili:
This is *great* to see this functionality arrive in **opam**!

AltGr:
@dbuenzli: yes, except you sometimes need dynamic messages (or even interactivity).

PR #2167, opened 2015-05-29
Fix setting OPAMFETCH=wget

### LINKED ISSUE #2164: Setting OPAMFETCH leads to wrong arguments being passed to wget
Here's an example (using master), where setting `OPAMFETCH=wget` leads to the `curl` arguments `--insecure`, `--retry-delay`, etc. being passed to `wget:

```
$ OPAMFETCH=wget ./src/opam install -vv ocamlfind 
  ∗  install ocamlfind 1.5.5

=-=- Gathering sources =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
[ocamlfind: from default] Command started
+ wget "--write-out" "%{http_code}\\n" "--insecure" "--retry" "3" "--retry-delay" "2" "-L" "-o" "/home/jeremy/.opam/repo/default/archives/ocamlfind.1.5.5+opam.tar.gz.part" "https://opam.ocaml.org/1.3/archives/ocamlfind.1.5.5+opam.tar.gz"
- wget: unrecognized option '--write-out'
- wget: unrecognized option '--insecure'
- wget: unrecognized option '--retry-delay'
- Usage: wget [OPTION]... [URL]...
- 
- Try `wget --help' for more options.
[ERROR] The sources of the following couldn't be obtained, aborting:
          - ocamlfind.1.5.5
        (This might be due to outdated metadata, in this case run 'opam update')
'./src/opam install -vv ocamlfind' failed.
```

AltGr:
Closes #2164. Thanks for reporting!

### DIFF
--- a/src/repository/opamDownload.ml
+++ b/src/repository/opamDownload.ml
@@ -44,7 +44,7 @@ let download_args ~url ~out ~retry ?checksum ~compress =
   let cmd, _ = Lazy.force OpamRepositoryConfig.(!r.download_tool) in
   let cmd =
     match cmd with
-    | [CIdent "wget", _] -> cmd @ wget_args
+    | [(CIdent "wget"|CString "wget"), _] -> cmd @ wget_args
     | [_] -> cmd @ curl_args (* Assume curl if the command is a single arg *)
     | _ -> cmd
   in


avsm:
does this require a 1.2.3 point release?

AltGr:
Don't think so, it can easily be worked around by:
- running on Darwin
- having no curl installed
- using the full command (which is preferred anyway):

        OPAMFETCH="wget --content-disposition --no-check-certificate -t %{retry}% -O %{out}% %{url}%"`

    or the simpler alternative `OPAMFETCH="wget -O %{out}% %{url}%"`

- setting `download-command: wget` in ~/.opam/config

Then the documentation says this should work, so if many users risk to be bitten by this, maybe we should re-release.

Issue #2165, opened 2015-05-27
opam 1.2.2 destroys the $PATH variable when used with the fish shell


hugomg:
After I installed opam 1.2.2, sourcing the .opam/opam/init.fish script ruins the PATH environment variable and makes my shell unusable.

The problem is that Opam is putting quotes around the $PATH:

    set -gx PATH "/home/hugo/.opam/4.02.1/bin" "$PATH";

This sets the PATH to be a 2-element array, with the second element being a space-separated list of the old paths. What we actually want is an array of length N+1 and to get that we should leave out the quotes:

    set -gx PATH "/home/hugo/.opam/4.02.1/bin" $PATH;

There is a similar issue with the MANPATH variable but its a bit more complicated because the semantics of MANPATH are different from the semantics for PATH.  In order for `man` to find the default system man pages, either MANPATH needs to be unset or MANPATH needs to be a colon-separated list with an empty element in it (that it, it must start with `:`, end with `:` or have a `::` somewhere in the middle). If we use the same technique we use for PATH it won't work in the case when MANPATH is previously unset. This is an [unfortunate quirk in fish](https://github.com/fish-shell/fish-shell/issues/2090). 

Anyway, changing the MANPATH assignment to this version seems to work for me:

    if [ 0 -eq (count $MANPATH) ];  set -gx MANPATH ""; end
    set -gx MANPATH $MANPATH "/home/hugo/.opam/4.02.1/man";

The first line is to make sure system manpages still work when MANPATH is initially unset and in the second line we got rid of the quotes just like we did for PATH.

--------

A second possibility would be to set MANPATH to an one-element array containing a colon-separated string in any format we want. This would be more future-proof in case a future version of fish decides to stop special-casing MANPATH but is a bit more tricky to implement (in a way that doesn't break if there are spaces or newlines in the file names).

AltGr:
Related to issues #1797, #1877, #1932, #2063, #2068, #2070.
Thanks for reporting. Why did fish decide to make things so complicated ? :p

The code is at https://github.com/ocaml/opam/blob/master/src/state/opamState.ml#L2037 if you want to have a look. The same code is called with the expanded PATH (for `opam config env`) and "$PATH" (for the `variables.fish` file) though.

hugomg:
Yeah, sometimes fish can be pretty annoying... Too bad its so pretty and tempting :)

The quickest fix I can think of is to avoid quotes if the string is a variable expansion for the array we are working with.

[...]

PR #2169, opened 2015-05-29
Fix Fish shell assignments on array env variables

hugomg:
* Correctly handle the case of the variable being unset
  - This fixes the "missing manpages" bug.

* Do not put quotes around array expansions:
    set -gx PATH  $PATH  "/blah"  # Good
    set -gx PATH "$PATH" "/blah"  # Bad

* Add logic for CDPATH, for future-proofing
  - It is the only other special array variable we have
    to potentially worry about.

### DIFF
--- a/src/client/opamConfigCommand.ml
+++ b/src/client/opamConfigCommand.ml
@@ -138,7 +138,9 @@ let print_sexp_env env =
 let print_fish_env env =
   List.iter (fun (k,v) ->
       match k with
-      | "PATH" | "MANPATH" ->
+      | "PATH" | "MANPATH" | "CDPATH" ->
+        (* This function assumes that `v` does not include any variable expansions
+         * and that the directory names are written in full. See the opamState.ml for details *)
         let v = OpamStd.String.split_delim v ':' in
         OpamConsole.msg "set -gx %s %s;\n" k
           (OpamStd.List.concat_map " " (Printf.sprintf "%S") v)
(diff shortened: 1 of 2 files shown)


hugomg:
Hmm, just noticed that there are also assignments on "/src/client/opamConfigCommand.ml". Dunno what would be the best way to proceed without duplicating a lot of code.

hugomg:
I couldn't fully understand why there is separate logic for generating the shell scripts that go on `variables.fish` and on `opam config env`. Right now, the code that generates variables.fish can handle assignments with variable expansions on the RHS (things like `PATH="$PATH:/blah"`)  and the code for `opam config env` does NOT and assumes that the directory names in the path lists are specified in full.

If this difference between the two functions is not a problem then I think this set of patches should fix #2165

AltGr:
Great, thanks!
`opam config env` is different because it can afford to be a bit more clever: it expands `$PATH`, detects previous instances of opam paths tied to the same opam root and _replaces_ them. And re-sets the variable in full.
`variables.fish` is much simpler, and is normally loaded only once at shell init, so expanding directly over `$PATH` using shell expansion is fine. The fish issue had already been (mostly?) fixed for `opam config env`, but not the `variables.fish` variant, it seems.

avsm:
This also seems like a 1.2.3 point fix candidate?

johshoff:
I'd like to support @avsm's suggestion to make a point release. This breaks the configuration of fish shells when installing the latest release, 1.2.2 (released a year ago, so I'm assuming this is what most package managers have).

cartazio:
when is this being merged? @avsm  ? 
seems like its still biting folks

avsm:
It's included in the forthcoming opam 2.0 release, which has just been tagged as rc1

PR #2180, opened 2015-06-05
Detect number of CPUs on init to enable parallelism

AltGr:
also fixes a duplicate --jobs option on 'opam init' Detection thanks to
@dbuenzli at http://stackoverflow.com/questions/16269393/how-to-get-the-number-of-cores-on-a-machine-with-ocaml

### DIFF
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -241,7 +241,7 @@ let apply_build_options b =
     ();
   OpamStateConfig.update
     (* ?root: -- handled globally *)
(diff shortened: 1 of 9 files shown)


hannesm:
nice feature. unfortunately on FreeBSD, `getconf _NPROCESSORS_ONLN` doesn't work... but `sysctl hw.ncpu` returns `hw.ncpu: 4` ; `sysctl -n hw.ncpu` returns just the value `4` (and a newline)...

PR #2181, opened 2015-06-07
Fix documentation of substs field from "subst"

dsheets:
The documentation of the `substs` field was subtly wrong as it was calling the field `subst`. Please let me know if you can think of other places to make this change or if you'd like me to generate something (`dev-manual.pdf`?).


AltGr:
travis failures failed due to bad md5 for cudf ; probably a transient error on the INRIA gforge...

AltGr:
Thanks! I'll handle the rest (online versions should be fine, which is what matters).

PR #2183, opened 2015-06-08
escape path properly for cygwin after "make cold"

struktured:
Cygwin paths have spaces in them typically, so path expressions need to be surrounded by quotes.


AltGr:
Thanks!

PR #2185, opened 2015-06-09
Make switch paths configurable (honoring switch/config/global-config.config)

AltGr:
**rebase before merge, this duplicates commits**

There is no configuration options for the user yet, but OPAM will now honor the paths defined in the switch's global config (making it less useless!)

### DIFF
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -70,7 +70,7 @@ if [ "$OPAM_TEST" = "1" ]; then
     sudo make install
     make libinstall prefix=$(opam config var prefix)
     # Compile and run opam-rt
-    wget https://github.com/ocaml/opam-rt/archive/1.3.tar.gz -O opam-rt.tar.gz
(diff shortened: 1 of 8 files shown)


src/format/opamVariable.mli
```diff
@@ -36,18 +36,28 @@ module Full: sig
 
   include OpamStd.ABSTRACT
 
-  (** Create a variable local for a given library/syntax extension *)
-  val create: OpamPackage.Name.t -> variable -> t
+  type scope =
+    | Global (** Note: this is attributed to unqualified variables, and may
+                 also design self-referring ones *)
+    | Self (** Variable in a package-specific file referring to that
+               package [-:varname] *)
```
samoht: Shouldn't be `_:varname`?

AltGr: Indeed (attempted `-:` but that didn't go well with the parser. And we don't want to introduce files that don't parse with older versions...)

PR #2186, opened 2015-06-09
use sysctl on FreeBSD to find the amount of online processors

hannesm:
as mentioned in #2180 `getconf _NPROCSSOR_ONLN` does not work on FreeBSD, but `sysctl -n hw.ncpu` does.

### DIFF
--- a/src/core/opamSystem.ml
+++ b/src/core/opamSystem.ml
@@ -457,11 +457,13 @@ let install ?exec src dst =
 let cpu_count () =
   try
     let ans =
-      if OpamStd.Sys.os () = OpamStd.Sys.Win32 then
-        [OpamStd.Env.get "NUMBER_OF_PROCESSORS"]
-      else
-        read_command_output ~verbose:(verbose_for_base_commands ())
-          ["getconf"; "_NPROCESSORS_ONLN"]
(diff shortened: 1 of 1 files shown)


hannesm:
tested by moving `.opam` away, and after a `opam init`, `jobs: 4` occured in `.opam/config`.

AltGr:
Thanks! Probably worth posting back to [stackoverflow]( http://stackoverflow.com/questions/16269393/how-to-get-the-number-of-cores-on-a-machine-with-ocaml) ?

dbuenzli:
I made an update to stackoverflow answer.

avsm:
could we make the overall cpu count usage to be `ncpus - 1`?  It's irritating when all interactivity disappears because all CPUs are used by OPAM.

AltGr:
wow, does that really happen ? What arch ? Maybe with an ssd ? I generally use [ncpus*3/2] without trouble, assuming there is a lot of i/o !

We should be careful not to use that formula on single-core machines :D

AltGr:
(or less: https://lkml.org/lkml/2012/3/31/131)

hannesm:
@dbuenzli thanks!

PR #2216, opened 2015-06-23
Add more fields to the pef export format

abate:
Add few more field to the pef export: author, homepage, licence, tags, bug_reports. The idea is also to add the build commands, but at the moment the format is a bit cumbersome in my opinion, so I'll leave this to further work,


abate:
uhmmm. the travis build failure does not seem related to this patch ...

src/state/opamState.ml
```diff
@@ -1292,6 +1297,24 @@ let dump_state t oc =
         (string_of_conjunction (fun a -> a) m);
     with Not_found -> () );
 
+    (match repository_and_prefix_of_package t package with
```
AltGr: Why do you read the opam files again rather than accessing `t.opams` ? You should use function `OpamState.opam`, unless you specifically want to skip pinned package's overlays, and `OpamPackage.Map.find pkg t.opams` otherwise.

src/state/opamState.ml
```diff
@@ -1265,6 +1265,11 @@ let string_of_cnf string_of_atom cnf =
 let string_of_conjunction string_of_atom c =
   Printf.sprintf "%s" (OpamStd.List.concat_map " , " string_of_atom (List.rev c))
 
+let string_of_stringlist l =
```
AltGr: it's not clear what this function does, if it's only useful locally, put it within the function that uses it.

src/state/opamState.ml
```diff
@@ -1292,6 +1297,24 @@ let dump_state t oc =
         (string_of_conjunction (fun a -> a) m);
     with Not_found -> () );
 
+    (match repository_and_prefix_of_package t package with
+    |None -> ()
+    |Some(repo,prefix) -> (
+      let opam_file = OpamRepositoryPath.opam repo prefix package in
+      let t = OpamFile.OPAM.read opam_file in
+
+      if (OpamFile.OPAM.author t) <> [] then
```
AltGr: match ?

AltGr:
yes, the Travis script on mac has been failing recently, need to find the time to look into it

src/state/opamState.ml
abate: yup ! I didn't understand the interface : too many functions with a similar name... much easier this way.

src/state/opamState.ml
abate: I've moved the 3 printing helpers inside the function. the function ```string_of_stringlist``` cleans up the input (it removes all \n) and simply concatenates the list of strings into one string.

abate:
I hope that my liberal use of ```git push -ff``` didn't messed up with github pull request machinery ...

AltGr:
Nope, I actually like clean rebased branches with push -f better :)

PR #2217, opened 2015-06-23
Refine maintainer warning in Opamfile.validate

abate:
the maintainer field must me not empty and it must contain a valid email
 address

### DIFF
--- a/src/format/opamFile.ml
+++ b/src/format/opamFile.ml
@@ -1614,6 +1614,7 @@ module X = struct
           (names_of_formula true t.depends)
           (names_of_formula true t.depopts)
       in
+      let emailrex = Re_pcre.regexp "<?[a-z0-9!#$%&'*+/=?^_`{|}~-]+(?:\\.[a-z0-9!#$%&'*+/=?^_`{|}~-]+)*@(?:[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\\.)+[a-z0-9](?:[a-z0-9-]*[a-z0-9])?>?" in
       let warnings = [
         cond 20 `Warning
           "Field 'opam-version' refers to the patch version of opam, it \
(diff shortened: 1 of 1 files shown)


dsheets:
Ignoring the unspeakable regex horror, the format `Name <email>` should probably be accepted.

dsheets:
Oops. Perhaps that's what `Re_pcre.pmatch` does... I can't tell because [it's not documented](https://github.com/ocaml/ocaml-re/blob/master/lib/re_pcre.mli).

abate:
Re_pcre uses the standard perl regexps. I've used a regexp that should catch most of the valid email around (there is a super complex regexp to catch them all, but for too comple for my tastes). And yes "Name <email>" is valid. I've tested this regexp with a recent opam repo and it does not trigger any false positives.

Drup:
Parsing emails with regexps is wrong, it's going to break at some point and the patch will not be pretty. It doesn't help catching anything anyway. I'm not even sure *why* you want to validate emails. There is no point.

If you really want only valid emails in the repository, just send them an email.

abate:
if the maintainer field does not have a valid email, how can you contact the maintainer of the package ? The nature of the opam-repository (a github repository) does not allow to keep track of the owner of a package. There is no other way to establish who is responsible for what otherwise. If regexp are the wrong tool, I'm all open to suggestion to do it in a better way ... but I'm really not sure how.

Drup:
The regexp will only tell you that it looks like an email address, not that the email address exist.

Really, the only way to do that is to send an email.

AltGr:
When signing the repository we will necessarily need to include validated, unique ids, and associate public keys to them ; emails or github ids would do, either way we'll need to include out-of-band verification of their authors.

Guaranteeing we have valid email addresses as maintainers would be a good thing, but indeed, there is a long story of failed email checkers, based on regexes or not (I'm still pissed everytime a site refuses my `a+b@s.com` address). The only reliable check, as @Drup points out, is to send a mail and wait for a response.

so this is indeed of concern, but I would postpone it until we get to the signed repo and do it properly (probably using git/GitHub information)

dbuenzli:
FWIW since this information gets published on various page on the www I deliberately obfsucate the adresses I put theses fields and I'd really like `opam lint` not to complain about this.

PR #2221, opened 2015-06-24
Update dose3 (4.0-rc1)

### LINKED ISSUE #2219: update dose dependency to latest release 4.0-rc1
The first candidate release of the upcoming 4.0 : https://gforge.inria.fr/frs/?group_id=4395&release_id=9269
This is a good time to close all the remaining opam-related bugs and remove all the remaining patches in the opam source. In the following months and for the 4.1 release of dose, my idea is to refactor all the code in opam that is general enough, and push it down to dose.

abate:
- Add new dependency to cppo (used by dose3 for conditional compilation)
- remove all remaining dose related patches
- resolve #2219


src_ext/Makefile
```diff
@@ -84,6 +87,16 @@ distclean:
 
 LIB_EXTS = extlib re cmdliner graph cudf dose_common dose_debian dose_algo uutf jsonm
 
+# add the preprocessor directive for OCamlMakefile
+CPPO = cppo/cppo
+addcppo = for ml in $1/*.ml $1/*.mli; do echo '(*pp $(CPPO) *)' | cat - $$ml > temp && mv temp $$ml; done
+
+addcppo:
+	$(MAKE) -C cppo all
+	$(call addcppo,dose/common)
```
AltGr: We can get rid of all shell indirections with a short sed command, and I think this will reduce incompatibilities

```
sed -i "1i(*pp $(CPPO) *)" dose/common/*.ml dose/common/*.mli dose/deb/*.ml dose/deb/*.mli dose/algo/*.ml dose/algo/*.mli
```

Also, the target should probably use a stamp file, not to re-add the header every time. I would have it a prerequisite of the `build` target rather than only in the `lib-ext` target, too.

AltGr:
Sounds good! Can you link to the dose changelog ?
I'd rather wait for the final release than use the rc, if it's coming shortly ?
Also, the dose release needs to be in opam-repository before we can make opam depend on it (`src_ext` is for bootstrap, but e.g. installing opam-lib requires the dependencies to be properly installed)

I'm slightly wary of adding a new dependency (cppo), which has its own environment detection and Makefile (more potential for build issues accross platforms), but cppo is quite small and getting rid of the live patches feels good.

src_ext/Makefile
abate: Now we use sed instead of cat. I don't need we need a timestamp as the target clone is always run before addcppo.

abate:
this is the changelog : https://gforge.inria.fr/scm/browser.php?group_id=4395 : there are a few api changes, but apparently nothing used by opam. 4.0 should come shortly, I'm cleaning up a couple of things before getting it out of the way. I'll change the download link with the final release as soon as it is ready. I'll also prepare a pull request to update dose in the opam-repository. thanks !

abate:
I removed the module debian of dose is it not necessary anymore as the debian comparison function of dose is now in the opam code base. The rc2 release of dose adds a new module Versioning that now contains the debian comparison function and the semantic versioning comparison function. This module depends only on the common module of dose.

abate:
I close this PR and open a new one with a more updated versions

PR #2222, opened 2015-06-25
new functionality to opam lint to detect redundant dependencies / conflicts

abate:
This patch  (to be merged after #2221 ) adds a new functionality to opam lint to detect redundant dependencies or conflicts. For each installable package (this test does not make any sense for a broken package), we check all direct dependencies, one by one, to ensure if, in the context of the current switch, the dependency does not refer to a package that is either broken or missing. If this is the case, the dependency is redundant. Similarly for conflicts. For example : 
```
$./opam lint ../../opam-repository/packages/cairo2/cairo2.0.4.6/opam 
Errors found in ~/Projects/repos/opam-repository/packages/cairo2/cairo2.0.4.6/opam:
  warning  6: In the current switch, some conflicts of this package are redundant:
              The conflict cairo (= 0.4.1) refers to a missing package
...
```
the conflict to cairo 0.4.1 present in the opam file is useless as this package is not present in the archive.

This is just a warning to help maintainers to tight and clean their dependencies and it is an approximation as it is run only in the current switch (all switches can be checked manually by running ```opam switch <opamversion> ; opam list <opamfile>``` multiple times). It requires dose3 >= 4.0-rc1 .


dbuenzli:
A command line switch should be added. You are not necessarily `lint`ing opam files that are related to your `opam`'s universe (or the packages you mention are not yet in the universe).

abate:
You mean you want to lint an opam file that refers to packages that belong to some external repository that is not accessible to the current opam instance. In this case I think this warning can be just ignored, as it is not possible to establish which dependencies are really missing by mistake, or they are missing because not known to opam. 

Then again, in the context of QA, I think this is quite of a corner case, as if a dependency is specified in an opam file, it is supposed to be satisfied somehow... 

Notably, for a package in the official opam-repository, it does not make sense to me to add a dependency to a package that can be satisfied only by a private repository. Similarly to add a dependency to a package that is not yet in a repository.

Did I understand you comment correctly ?

On a related note, I've just noticed that opam lint does not have an option to selectively disable warnings. Maybe something we could add ...

dbuenzli:
> Then again, in the context of QA, I think this is quite of a corner case, as if a dependency is specified in an opam file, it is supposed to be satisfied somehow...

It's not necessarily a corner case. Suppose you make the first release of two dependent packages,  `opam-submit` should not fail on that if `lint` fails, otherwise you will never be able to opam-submit your packages (or have to wait that the dependency is really in before being able to submit the second one).

abate:
but it's a warning, not an error ... does opam-submit fails on opam-lint warning ? This is just to say : "Hey maintainer, maybe there is something wired in your dependencies. Do you want to double check ?" . Yes, not quite a corner case, I agree.

dbuenzli:
> but it's a warning, not an error ... does opam-submit fails on opam-lint warning ?

I don't remember. Another issue to consider is that the check will be done in your universe which may be quite different from the raw  `opam-repository` universe and thus questions the credit to actually give to the warning (or lack of warning).

abate:
opam lint could be run as a commit-hook on the official ```opam-repository``` after a pull request to add/update a package on a selected number of switches : I think this is the direction we are moving to. Clearly the maintainer can run ```opam lint``` on his own machine as well, but at the point he/she is responsible of the universe that he is considering.

AltGr:
This is useful, but should be kept separate from `OpamFile.OPAM.validate`. My main point is that this patch adds a dependency from lib `opam-format` towards lib `opam-solver`, which is highly undesired (if there had to be a dependency, I would actually feel it more natural to be the other way around). Some lightweight tools rely on opam-format for parsing/printing but don't import dose, extlib, etc.

This converges with @dbuenzli's remarks that `lint` has to remain independent from the current switch or configuration. [opam-ci](AltGr/opam-ci), in order to check installability, loads a global universe that is an approximation, as it ignores all availability, OS and switch constraints before calling dose: maybe such an approach would be more suited in this case too.

Also, a note of warning: your PR includes patches that are unrelated to what is announced, and that have been separately pull-requested. In this case, you should either:
 1. rebase on master with just the packages needed (`git rebase -i`)
 2. if there is a dependency that makes it impossible, wait for the previous pull-request to be merged, then rebase before pull-requesting
 3. if there is a dependency and you want to start discussion already, you can file the PR with the patches, but in this case announce in the title that this shouldn't be merged before the parent is merged and it has been rebased.

The reasons for this is that it could get, at best, duplicate patches in the git history, at worst, sneaking up patches that get merged without proper review.

abate:
I see. So, are you suggesting to move everything in opam-solver and then add two steps to the lint function. First call validate_file and then OpamSolver.validate_deps ? 

Reg the patch dependencies, I see your concern. I'll try to state more clearly that his patch should be merged only after #2221

AltGr:
No, that's not what I am suggesting. `opam-publish`, for example, depends on `opam-format` and makes use of `OpamFile.OPAM.validate`, but not on `opam-solver`. It should be separate, and then we could call the two from OpamArg or OpamClient.

abate:
Ok, I need a bit of guidance. If this function can't be in ```opamFile.ml" because it will add a dependency for ```opam-publish``` and you are not suggesting to put it in ```opam-solver```, where should this function live ? Directly in ```opamClient.ml``` or in a separate file in the client submodule ? I'm a bit lost. I still don't have a clear picture of all the opam dependencies.

src/Makefile
```diff
@@ -43,7 +43,7 @@ endif
 
 ifneq ($(HAS_LIBEXT),)
   EXT_INCDIRS = ../src_ext/lib
-  LIBS = unix extlib re cmdliner graph cudf dose_common dose_debian dose_algo uutf jsonm
+  LIBS = unix extlib re cmdliner graph cudf dose_common dose_algo uutf jsonm
```
AltGr: Good one, thanks for spotting this!

src/format/opamFile.ml
```diff
@@ -17,6 +17,7 @@
 open OpamTypes
 open OpamTypesBase
 open OpamStd.Op
+open ExtLib
```
AltGr: Please don't

src_ext/Makefile
```diff
@@ -178,18 +191,15 @@ define PROJ_dose_common
 endef
 export PROJ_dose_common
 
-SRC_dose_debian = architecture.ml format822.ml format822_parser.mly		\
-packages_parser.mly version.ml packages_lexer.mll format822_lexer.mll		\
-packages.ml release.ml sources.ml printer.ml evolution.ml debutil.ml debcudf.ml \
-apt.ml edsp.ml
+SRC_dose_versioning = versioning.ml debian_version.ml semantic_version.ml
```
AltGr: we actually don't need this at all anymore do we ?

AltGr:
Actually, I read to quickly:
- there currently _is_ a dependency of opam-solver towards opam-format (although only for base types and OpamPackage; opam-format could be further split or those lifted to the core lib...). So no reverse dependency is possible at the moment.
- you're calling cudf and dose functions from opam-file, which explains that it did compile, and is out of the question: it only worked because the default built system and META overevaluate dependencies (good hint to fix this). It breaks the `make fast` target, which has finer per-lib dependencies, too (.ocp files).
- the function handling cudf data should be in OpamCudf (and use the printing and manipulation functions that are already there). It may return the same type as OpamFile.validate (possibly through a higher-level function in OpamSolver).
- there may be an extra argument to `lint` in OpamArg to add this check, call this function and merge the results with regular `OpamFile.validate`.
- please use ocp-indent and adjust to the surrounding style.

abate:
I've moved all the code to opamCudf and opamArgs this should fix the all dependency issues. I've the impression there is an easier way to get name and version of a package ...

PR #2233, opened 2015-07-01
Dose3.4.0 rc

abate:
updated version of #2221 ( opam package : ocaml/opam-repository#4337 )


avsm:
@AltGr what's the status of merging this?  Would be convenient to have the latest Dose or merge #2285

AltGr:
This makes the embedded-deps mode (a.k.a. `make lib-ext`) rely on latest dose, but we need to check that it also solves the issues with opam-packaged dose 4.0.

AltGr:
#2291 resolves the issues there were at link-time with newest dose and Debian module.

AltGr:
Also, about e6760d7, we won't need the versioning module right now, although we might find a way to rely on it later.

PR #2237, opened 2015-07-03
Broken URL to OPAM repository

ptroja:
The URL was displayed as https://github.com/ocaml/opam-repository but pointed to http://opam.ocaml.org/doc/Packaging.html

### DIFF
--- a/doc/pages/Packaging.md
+++ b/doc/pages/Packaging.md
@@ -22,7 +22,7 @@ Publish to the OPAM repository:
 * Use the [opam-publish](https://github.com/OCamlPro/opam-publish#opam-publish)
   tool (`opam install opam-publish`)
 * Or by hand:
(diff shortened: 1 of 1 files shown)


dsheets:
I don't think the processor for http://opam.ocaml.org/ will automatically linkify URLs in text (but I could be mistaken). The safest thing to do is to specify the URL as the `href` of the link with the text of the URL.

ptroja:
It will, or at least it does in the Preview of the Github on-line editor.

dsheets:
I believe <https://github.com/ocaml/omd> is the processor for http://opam.ocaml.org/ which does not do this as it is not "specified" by markdown.

E.g.:
```
$ omd
http://opam.ocaml.org/
<p>http://opam.ocaml.org/</p>
```

AltGr:
No link is still better than a wrong link... esp. since the url is to be used for cloning.

PR #2248, opened 2015-07-07
Check checksum format during parsing

yallop:
We recently had a url file with an invalid checksum (with an extra character) sneak through:

http://opam.ocaml.org/builds/20ff97ad14c79484d876c1200ba5e83dd7ca46ae/logs/local-ubuntu-14.04-ocaml-4.02.1/raw/sawja.html

This PR adds checksum validity checking during parsing so that errors can be caught earlier and reported more clearly.

### DIFF
--- a/src/core/opamFilename.ml
+++ b/src/core/opamFilename.ml
@@ -152,6 +152,9 @@ let to_string t =
 let digest t =
   Digest.to_hex (Digest.file (to_string t))
 
+let valid_digest t =
+  try Digest.from_hex t; true with Invalid_argument _ -> false
+
 let touch t =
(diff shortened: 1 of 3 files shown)


AltGr:
Thanks! Two minor issues with the tests:
* opam-rt uses fake checksums -- which breaks... I'll fix this in opam-rt.
* `Digest.from_hex` didn't exist yet in OCaml 3.12, which we are still supporting

yallop:
bba0dcc removes `Digest.from_hex` for 3.12.1 compatibility.

AltGr:
All good, thanks a bunch!

Issue #2098, opened 2015-04-03
mechanism to vendor opam dependencies locally


perrelle:
If I am not mistaken, the procedure to build a local repository when you don't already have opam initialized is to:
1/ Init opam
2/ Install the packages you want to put in you local repo
3/ Move into the .opam directory and remove the log subdirectory
4/ use opam-admin make --resolve with the wanted packages
The installation of the package is unnecessary as you only want to download them. It should probably be possible to build a local repository without an opam root.

samoht:
You can simply run `opam-admin make` at the root of a clone of `opam-repository`. That will download *all* the archives. You can run `opam-admin make --help` to have more options (such as `--resolve` to download a minimum set of packages).

You are not supposed to manually dig into `.opam` ever.

AltGr:
Ouch, that sounds like a bad idea, and won't work! Where did you read about this ?

The proper way is simply to clone https://github.com/ocaml/opam-repository , and run `opam-admin` from there.
Hope that helps !

perrelle:
I didn't read that anywhere. I just tried to fulfill opam-admin make requirements. I wanted to not download too much. Is cloning the whole opam-repository the only solution ?

avsm:
@perelle it would help if you could describe your intended usecase more fully.

perrelle:
You are right. I'm trying to build an "offline distribution" of one of the opam packages. I want to build an archive containing opam (and ocaml and findlib) sources, a local opam-repository and an installation script which bootstraps opam and install the package from the local repository. I realize this is not the original intended use of opam-admin.

avsm:
Thanks -- that's a usecase we really need to support.  I prototyped an `opam-boot` script that started this off (for the case when there's no local OCaml installation at all) https://github.com/avsm/opam-boot

It needs a few more things:

- a local `archive/` directory with the vendor dependencies.  This can be done via `opam admin make --resolve <pkg>`
- a mechanism to get the subset of opam-repository with just the packages needed.  Can't do this via `git`, so it may be ok to just clone the whole thing as a first cut.
- adapt `opam boot` to look for local copies of the files rather than do network traffic, so that they can all be bundled in a distribution.

perrelle:
Thank you. I will try your solution.

AltGr:
[...]

PR #2257, opened 2015-07-15
Feature/mini repo

bobot:
Fix the poblem with virtual package in #2098. The side effect is that `archive/*.virtual` files are now created by `opam mk-repo` but it seems not bad.


AltGr:
A simpler/cleaner way could be to add an option to `opam-admin` to restrict the packages directly, rather than do it through a shell-script (and require a separate channel for packages without archives): something like `opam-admin make --resolve-and-filter <packages>` that turns a full repository into a limited one only for the listed packages.

samoht:
Any update on this? I could use this :-)

AltGr:
We now have a clean way to do this, using `opam admin filter` :)

PR #2260, opened 2015-07-18
Lint warning for ambiguous dev-repo addresses

### LINKED ISSUE #2240: opam lint should check for VCS URLs in dev-repo fields
Whatever is necessary for opam to select the VCS backend should be ensured present by `opam lint`. E.g. `http+hg` or `https://......git` etc.

AltGr:
Closes #2240

### DIFF
--- a/src/format/opamFile.ml
+++ b/src/format/opamFile.ml
@@ -1505,7 +1505,7 @@ module X = struct
              OpamFormat.bad_format ~pos:(OpamFormat.value_pos v) "%s" msg) |>
         (function
           | Git _ | Darcs _ | Hg _ as pin -> pin
-          | Http u -> Git u
+          | Http u -> if conservative then Http u else Git u
           | _ ->
             OpamFormat.bad_format ~pos:(OpamFormat.value_pos v)
               "Unrecognised version-control address")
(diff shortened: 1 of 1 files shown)


dbuenzli:
Are urls of the form `http://example.org/bla.git` deemed ambiguous ?

dbuenzli:
Btw. are these ambiguous addresses treated as errors by opam are the moment ? See https://sympa.inria.fr/sympa/arc/caml-list/2015-07/msg00081.html

AltGr:
Ending in `.git` makes the address unambiguous ; I prefer the prefix url syntax (`git+https://`), but the former is better w.r.t. backwards compatibility.
These are errors, but latest opam shouldn't complain anymore at update -- just silently ignore the field (rather than the whole package). The rationale is that this is an issue for the packager, and no concern to the end-user. Same as these deprecation messages spamming my console every time I run a GTK (or QT, for that matter) program from there: just annoying noise.

samoht:
Some repo still uses CVS. We should take care of them (I don't know exactly how, but we should not fail).

AltGr:
you mean implementing additional support for cvs/svn repos ?

AltGr:
ah, or have the field has simple documentation in that case ? hmm...

PR #2265, opened 2015-07-22
This allows fields with prefix 'x-' in opam files, unused by opam but accessible

AltGr:


src/client/opamClient.ml
```diff
@@ -35,6 +35,7 @@ type package_details = {
   tags: string list;
   syntax: string list Lazy.t;
   libraries: string list Lazy.t;
+  extension: (string * string) list Lazy.t;
```
dbuenzli: Plural maybe ?

src/format/opamFile.ml
```diff
@@ -1510,24 +1535,32 @@ module X = struct
...
         build_test; build_doc; depexts; messages; post_messages;
-        bug_reports; flags; dev_repo
+        bug_reports; flags; dev_repo; extension;
       }
 
     let of_channel filename ic =
       let nv = OpamPackage.of_filename filename in
       let f = Syntax.of_channel filename ic in
-      let permissive = Syntax.check f valid_fields in
+      let permissive = Syntax.check ~allow_extended:true f valid_fields in
```
dbuenzli: If the above is changed `allow_extensions` would be better.

Issue #2247, opened 2015-07-07
Switch and package configuration variables


dbuenzli:
# Switch and package configuration variables

Switch and package configuration variables are respectively per switch and per package key-values bindings that can be accessed in opam files to configure packages.

Configuration variables come in two flavors, input and output variables (better names ? user and state variables ?). The difference being that input variables can be set by the opam end-user while output variables are usually defined by the result of install processes.

More precisely for package configuration variables:

* *input* variables can be set before package installation to configure the package installation process.

* *output* variables can be declared by the package install to provide information about its installation for other packages to use.

For switch configuration variables:

* *input* variables can be set to globally configure packages that may use the variable.

* *output* variables are variables determined by the switch creation procedure

Note that both input and output variables share the same namespace. It is not possible to declare both an input and output variable with the same name.

Also note that the current value of an *input* variable says nothing about the current install state of a switch or a package as they may be set by the end-user at any time without further side-effects. (N.B. if we had precise dependency tracking of variables in the opam files we could actually reinstall any package that depends on a variable whenever it is changed, but that may not be desirable either).

## Input variable declaration

Switch input variables are defined in the switch's `.descr` file and package input variables are defined in the package's `opam` file via the `config-input-vars` field as follows:

```
config-input-vars: [ ([IDENT (BOOL | STRING | [ STRING+ ]) STRING])+ ]
```
That is each input variable needs to define a name, a default value (which can be a variable interpolated string, e.g. in a package it can refer to a switch input variable) and a documentation
string. Example:
```
config-input-vars: [
  [ "modernity" false "Use CPU extensions (SSE2, AES-NI)." ]
]
```

## Output variables declaration

Output variables are defined in the switch's `.descr` file and the package's `opam` files via the `config-output-vars` variables as follows:
```
config-output-vars: [ ([IDENT STRING])+ ]
```
That is each output variable needs to define its name and a documentation string.

For packages, the package must then define the value of each of the declared variables by installing a `opam.config` file with the package in its `lib` directory, i.e. `$opam/$SWITCH/lib/opam.config`.

It is an *install* error if the package declares an output variables and fails to define it in its `opam.config` file.

FIXME I know less about the switch install procedure but I guess something similar could be done.

## Setting input variables

Input variables can be set by using the following invocations:
```
opam config var-set $PKG:$VAR $VALUE
opam config var-set $VAR $VALUE
```
[...]

PR #2268, opened 2015-07-24
Extend 'opam config' interface

AltGr:
- new commands 'set' and 'unset' for custom switch variables
- 'exec' now expands variables in command and arguments
- new command 'expand' to simply opam-evaluate a string

### DIFF
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -1012,7 +1012,8 @@ let config =
     ["exec"]    , `exec    , ["[--] COMMAND"; "[ARG]..."],
     "Execute $(i,COMMAND) with the correct environment variables. \
      This command can be used to cross-compile between switches using \
-     $(b,opam config exec --switch=SWITCH -- COMMAND ARG1 ... ARGn)";
+     $(b,opam config exec --switch=SWITCH -- COMMAND ARG1 ... ARGn). \
+     Opam expansion takes place in command and args.";
     ["var"]     , `var     , ["VAR"],
     "Return the value associated with variable $(i,VAR). Package variables can \
      be accessed with the syntax $(i,pkg:var).";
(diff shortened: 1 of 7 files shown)


src/client/opamArg.ml
```diff
@@ -1020,6 +1021,15 @@ let config =
...
      $(i,PACKAGE), lists all the variables available for these packages. Use \
      $(i,-) to include global configuration variables for this switch.";
+    ["set"]     , `set     , ["VAR";"VALUE"],
+    "Set the given global opam variable for the current switch. Warning: \
+     changing a configured path will not move any files! This command does \
+     not perform any variable expansion.";
+    ["unset"]   , `unset     , ["VAR"],
+    "Unset the given global opam variable for the current switch. Warning: \
+     unsetting built-in configuration variables can cause problems!";
+    ["expand"]  , `expand  , ["STRING"],
```
dbuenzli: eval ?

src/client/opamArg.ml
```diff
@@ -1020,6 +1021,15 @@ let config =
     "Without argument, prints a documented list of all available variables. With \
      $(i,PACKAGE), lists all the variables available for these packages. Use \
      $(i,-) to include global configuration variables for this switch.";
+    ["set"]     , `set     , ["VAR";"VALUE"],
```
dbuenzli: This would later be expanded for package variables right ? (If #2247 gets implemented).

AltGr: Yes indeed; at the moment it just sets in `global-config.config`, so variable names with colons need to be refused. We don't yet keep package's config files so setting their variables would be more misleading than anything at the moment.

src/client/opamArg.ml
AltGr: hmm, `eval` would be better, but in my brain, that connects with something that is going to be executed... so I was afraid of a confusion with `exec`. We also have `eval $(opam config env)` all around the place so I'm trying to avoid it.

Issue #2261, opened 2015-07-18
document the existence of external solvers better


gasche:
Today I tried to point an OPAM user on the caml-list to a documentation of what external solvers are and how to install them. I failed to find such a page. The best I could find is the [Specifying Solver Preferences](https://opam.ocaml.org/doc/Specifying_Solver_Preferences.html) page that seems to have evolved to have a bit of this information (but is too advanced for beginners), and the [online solver farm](http://cudf-solvers.irill.org/index.html) page that takes a different approach.

This issue is related to #1267 , but I don't think it is a duplicate. #1267 is about making external solvers easy to install, but right now there is not even an easy-to-find description of what they *are* and are useful for, which is orthogonal.

AltGr:
Well spotted, thanks. This should probably be close to the install / setup guide pages.

samoht:
Could `opam` itself warn about missing external solvers?

dbuenzli:
> Could opam itself warn about missing external solvers?

When ?

samoht:
When the internal solver goes crazy? For instance, after the user press `[n]` when opam asks him to downgrade all his packages. Or it would be possible to add a hook in the internal solver to print a message when the returned solution is not optimal.

AltGr:
It does already, but at `init`.
The idea of adding an heuristics on top of the heuristics to detect if they went wrong is a bit spooky to me, but if it's done e.g. when hitting the timeout, that sounds like a good idea.

AltGr:
Added http://opam.ocaml.org/doc/Usage.html#Beforeyougetstartedquotexternalsolversquot

gasche:
Thanks! I would like to send a pull-request for minor clarifications, but I cannot find the git repository for the OPAM documentation (why is it not linked in a clear way on the documentation website? the links I can find point to `opam2web` which is not really helpful).

AltGr:
it's simply in the main repo, in [doc/pages](https://github.com/ocaml/opam/tree/master/doc/pages)
(used to be in a github wiki, but that was too constrained)
Indeed, there should be a link to the source on the website.

AltGr:
Links to the source should now appear in the footer (once the patches propagage) :)

gasche:
Over at ocsigen/lwt#177 @hcarty had a recommendation for the external solver documentation:

[...]

PR #2271, opened 2015-07-30
move "External solvers" documentation from Usage to Install (#2261)

gasche:
We also detail how to explicitly specify a specific solver, as
recommended by Hezekiah M. Carty.

I guessed the `#Externalsolvers` id; can someone confirm that this will be correct demangling at release time?

Also, it would be nice if there was a way to ask OPAM which solver it will use, for users to check that the solver-detection heuristics worked (and that the solver is correctly installed, etc.). Is such an option hidden somewhere?


gasche:
I added your explanation of why `aspcud` is recommended to the PR.

AltGr:
Thanks a lot!

PR #2280, opened 2015-08-08
OpamState: pass variables to add_to_env

ygrek:
this is needed for opam-bundle


AltGr:
local_variables used to be only used for `success` and `failure` when printing post-messages, i.e. field-specific, context variables. I guess you're using it for something else ?

ygrek:
this probably qualifies as a hack but see variables defined at https://github.com/ygrek/opam-bundle/blob/master/src/main.ml#L122 and used later

AltGr:
Ah, alright then ; I was mostly curious :)

PR #2285, opened 2015-08-13
Sync local `opam` file with opam-repository

avsm:
This fixes compilation of a direct `opam pin add opam-lib --dev`
command, which currently picks up dose4


samoht:
Merging as it breaks some automatic deployments.

avsm:
Thanks!

PR #2288, opened 2015-08-14
Fix typo

Chris00:


AltGr:
Thanks! There were indeed still some occurences of this _faux ami_

PR #2289, opened 2015-08-17
Support for packages from Atlassian's Stash

antoine1fr:
**Enhance detection of Zip archives**

Stash doesn't put any extension to its release archives. Currently, opam detects Zip archives by checking the filename's extension. This patch uses system's file command instead.

**Add support for packages with everything in the root**

Let's consider the following package tree:

```
/tmp/opam-xxx-yyy/
                 |- Makefile
                 |- src
                   |- toto.ml
```

Currently, opam only moves `src` to `.opam/<switch>/build/<package.a.b.c>`, instead of moving the whole tree.

This patch handles this case by directly moving `opam-xxx-yyy`, if regular files are found in it.

**Notes**

I didn't know whether the Makefile was generated or handwritten, so I added a dependency to the `str` package without really knowing what I was doing. :s

### DIFF
--- a/src/core/opamSystem.ml
+++ b/src/core/opamSystem.ml
@@ -546,13 +546,19 @@ let extract file dst =
       f tmp_dir;
       if Sys.file_exists dst then
         internal_error "Extracting the archive will overwrite %s." dst;
-      match directories_strict tmp_dir with
-      | [x] ->
-        mkdir (Filename.dirname dst);
-        command [ "mv"; x; dst]
+      match files_all_not_dir tmp_dir with
+      | [] ->
+        begin match directories_strict tmp_dir with
+        | [x] ->
+          mkdir (Filename.dirname dst);
+          command [ "mv"; x; dst]
+        | _ ->
+          internal_error "The archive %S contains multiple root directories."
+            file
+        end
       | _   ->
-        internal_error "The archive %S contains multiple root directories."
-          file
+        mkdir (Filename.dirname dst);
+        command [ "mv"; tmp_dir; dst]
   )
 
 let extract_in file dst =
--- a/src/core/opamSystem.ml
+++ b/src/core/opamSystem.ml
@@ -524,7 +524,16 @@ module Tar = struct
 end
 
 module Zip = struct
-  let is_archive f = Filename.check_suffix f "zip"
+  let is_archive f =
+    let ic = open_in f in
+    let c1 = input_char ic in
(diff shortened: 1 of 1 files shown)


antoine1fr:
Ok, I created my branch on top of an old version of the code so it won't build. I'm rebasing my work.

AltGr:
Thanks!
I have however a few reservations on your contribution as it is:
- we don't want to depend on `str`. We already use `re` which subsumes it. There is a compatible `Re_str` module, although you should preferably use `Re` directly (I have pending patches that attempt to unify the style and avoid having `Re_pcre` at one point, `Re_str` at another and then `Re_perl` somewhere)
- I am not sure calling the `file` program is the most portable option -- and we target OSX, and *BSD beside Linux (and we're working on Windows). If it's straight-forward enough, reading the magic number would be a better solution (a function to check magic number could be added to OpamSystem).

All this said, it would indeed be very nice to have better handling of external archive types, and your initiative and contribution are very welcome!

AltGr:
oops, sorry, it seems that was a different issue that made the build fail. My bad.

AltGr:
(we already have code that reads the first bytes of a file in `OpamState.Cache`. That could be factored out to a function in `OpamSystem` that checks if the file starts with a given string)

antoine1fr:
Thank you for your feedback. I will address these issues as soon as possible. :)

samoht:
I guess you want to use something like https://github.com/ocaml/opam/blob/master/src/core/opamSystem.ml#L480  Not sure why it is not used anymore.

AltGr:
Ah, indeed! It's used for tar files, but not for zip.

antoine1fr:
Now that I've rebased my branch on top of *master*, I notice that the *upstream-kind* is http instead of https, even if https was mentioned in the *url* file. Is it normal or should I create a ticket?

antoine1fr:
```
antoine@localhost:~$ opam show cow
             package: cow
             version: 1.3.0
          repository: default
        upstream-url: http://github.com/mirage/ocaml-cow/archive/v1.3.0.tar.gz
       upstream-kind: http
```

AltGr:
Woops, that's a bug indeed. I already have a pending rewrite for this function (`OpamTypesBase.parse_url`) though, so no need to report right now; I'll remember to include this in my tests.

AltGr:
Thanks a bunch!

PR #2290, opened 2015-08-17
Minor tweaks to improve the external API

samoht:


samoht:
Any reason why `OpamArg` was not exported in `opam-lib.client`?

AltGr:
Yes: it used to perform initialisation unsuitable for tools other than opam (like the self-bootstrap thingie). I pulled it out because really bad stuff could happen.

Before OpamArg is shared again, we should split the generic arg parsers and initialisation functions from the opam-exec specific stuff, and move the latter to e.g. OpamMain.

Another reason would be that opam-lib.client otherwise doesn't require cmdliner.

src/solver/opamCudf.mli
```diff
@@ -146,14 +146,25 @@ val remove_all_uninstalled_versions_but: Cudf.universe ->
...
+
+(** true if this package belongs to the roots ("installed manually")
+    packages *)
+val s_installed_root: string
+
+(** true if the package is pinned to this version *)
+val s_pinned: string
+
+(** the number of versions of the package since this one*)
+val s_version_lag: string
```
AltGr: Thanks!

AltGr:
All the rest is cool, thanks !

samoht:
> Before OpamArg is shared again, we should split the generic arg parsers and initialisation functions from the opam-exec specific stuff

I am not sure which opam-exec you are speaking about: is it the [command types and values](https://github.com/ocaml/opam/blob/master/src/client/opamArg.mli#L25-79) and the [global](https://github.com/ocaml/opam/blob/master/src/client/opamArg.mli#L119) and [build](https://github.com/ocaml/opam/blob/master/src/client/opamArg.mli#L130) apply functions?

AltGr:
I meant the main `opam` program.

At the moment, a quick inspection of the code shows that:
- the `apply_build_options`, `apply_global_options` are now safe. The latter would indeed be particularly useful, esp. for programs that should consistently use the "current" switch (whichever way it is defined). The underlying interfaces are now clean but doing the initialisation right can still be a bit tedious (you need to get the opam root using `OpamStateConfig.opamroot`, then init opam-format, then call the function that loads the values from opamroot/config, then call each sub-library's configuration function with the parameters you may want to override from your own command-line options).
- `create_global_options` is very dangerous (because of the self-upgrade stuff)
- otherwise, I think all option parsers should be shared, and
- the commands themselves should be kept out of the lib

Actually, that should be fairly straight-forward: cleanup the create_global_options hack, and otherwise cut the file in the middle. I'll try something right away.

samoht:
For the init bits I am currently using [this](https://github.com/samoht/ciso/blob/3d79b9aebfbeb9fa8d0edbd0f1acbae6e890bed2/lib/ci_opam.ml#L20-L32) I think that's a bad user interface (although much better than the previous one full of magic) but I am not sure at all what this does exactly and if I am doing the right thing :p An helper in opam-lib.client would be quite handy.

AltGr:
#2291 cleans up OpamArg and puts it back into the library. This PR rebases cleanly apart from the atom converter patch, which is superseded.

> For the init bits I am currently using  [this](https://github.com/samoht/ciso/blob/3d79b9aebfbeb9fa8d0edbd0f1acbae6e890bed2/lib/ci_opam.ml#L20-L32)

Yup, that's pretty much it. You can probably skip the `logs_dir` part, which would have your program write to a subdir of `/tmp` by default, but that's it.
It makes perfect sense, since we split into several sub-libraries, that they can be initialised separately. However, indeed, it's very tedious when you just use all of them, and opam-client should provide helpers to initialise all the chain in one go.

Basically, what the `OpamXxxConfig.init` do is initialise their config from the environment while allowing overrides through optional arguments.

AltGr:
@samoht: #2292 should make you happy!
[Interface](https://github.com/OCamlPro/opam/blob/8770656af4ade54640a53d0681276f61f52c1882/src/client/opamClientConfig.mli#L43-L87)
[Implementation](https://github.com/OCamlPro/opam/blob/8770656af4ade54640a53d0681276f61f52c1882/src/client/opamClientConfig.ml#L81-L102)  (just stacking the optional arguments, was lots of fun!)

samoht:
Great, thanks!! do you want me to rebase that branch?

samoht:
I've rebased the patches.

samoht:
And I have simplified and documented the sub-command scheme a bit better.

AltGr:
Great, thanks a bunch!

PR #2292, opened 2015-08-19
Provide a function for easy general initialisation in OpamClientConfig

AltGr:


src/client/opamClientConfig.mli
```diff
@@ -27,21 +27,61 @@ type 'a options_fun =
...
+  ?skip_version_checks:bool ->
+  ?all_parens:bool ->
+  ?log_dir:OpamTypes.dirname ->
+  ?print_stats:bool ->
+  ?sync_archives:bool ->
+  ?pin_kind_auto:bool ->
+  ?autoremove:bool ->
+  ?editor:string ->
+  ?current_switch:OpamSwitch.t ->
+  ?switch_from:[ `Command_line | `Default | `Env ] ->
```
samoht: what does this argument mean?

AltGr: The origin we got the "current switch setting" from: it's used in opam to display the proper `eval $(opam config env --switch)` messages for example.
Probably doesn't deserve to be here, at least not in this form -- passing `~switch` should probably set it to `Command_line` automatically.

samoht:
This looks like a very good addition :-)

PR #2294, opened 2015-08-25
Print exception caught instead of just "Could not fetch repo"

mjambon:
This how we found out that `rsync` was missing.


AltGr:
Thanks!

PR #2302, opened 2015-09-14
MD5 sum for dose 3.3 changed

chambart:
The file http://opam.ocaml.org/archives/dose.3.3+opam.tar.gz changed, an update of the md5 is needed. This is currently preventing the travis script on the compiler repository to run correctly.


AltGr:
Thanks!

PR #2303, opened 2015-09-15
Small changes & fixes that I had stashed locally

AltGr:


AltGr:
gah, newer re needs Bytes. Guess we have to stick with the older one...

Drup:
Huh ? Can't you use ocamlfind's overlay ?

AltGr:
And add ocamlfind as a new dependency to opam, which depends on the ocaml version ? No thanks. Adding opam's overlay to be somehow accessible from re would be a better solution, but I don't see the need just to use one function that is new in re.

Drup:
> And add ocamlfind as a new dependency to opam, which depends on the ocaml version ?

Ah, yeah, good point. It's so easy once inside opam that you forgot it's not that easy outside. :D

PR #2310, opened 2015-09-22
Add implicit dependencies of every package towards the base packages

AltGr:


samoht:
Some packages are using base package as configuration switches: for instance mirage package used to depend or not on base-unix (this is not strictly needed anymore but still good to have I think). Other (lwt?) depends on the presence of base-thread

AltGr:
Since this only concerns the base packages _of the current switch_, I am not sure it will have any effect on the usage you mention, unless I am missing something ?
The variant of the package depending on base-unix is still unavailable if you don't have it, and the other one will be unchanged in this case. If you do have it, both packages will now depend on it, but this shouldn't have any effect ?

Just to clarify my goal: with compilers as packages, you normally can't change base packages, but it's of course convenient in some cases (like my [rolling benches](http://flambda.ocamlpro.com/bench/operf/latest) of the flambda branch) ; so you can pin OCaml, and then upgrade it. This patches ensures everything¹ stays consistent when you do ; so you can do e.g. `opam upgrade ocaml`.


¹ well, not everything yet, the configuration variables aren't updated, so it may get weird e.g. if you really change the compiler version number, or what variants (`native-dynlink`, ...) are available. But this is planned.

AltGr:
I'm merging for now, but don't mean to end the discussion if there are issues with this ; it could be reverted later.

PR #2328, opened 2015-10-21
This allows storing url and descr data in the opam file

AltGr:
Thus having packages needing only a single file on the repo.
It uses the opam-syntax "sections" that always existed but were never used.

Separate files are still allowed and __override__ what is in the opam file.

This should be backwards compatible with the repo, but opam now stores single files internally.
This also means less files to scan, and that they are now in the cache: opam search is much faster.

Needs testing...


AltGr:
A next step could be to rework the repo hierarchy, to allow single-file packages without directories, e.g.:
`~repo/pkg/pkg.version.opam` instead of a directory `~repo/pkg/pkg.version`.

Or we could do without the `.opam` suffix and just turn the directory to a file!

hannesm:
I wonder how many of `descr` are actually per-version (I'd expect a description of a package to not change during the lifetime of a package name) -- if they are all the same, a single `descr` and multiple `pkg.version` opam files would be great to have (I still like separate `descr` files, to quickly `cat` package descriptions)...

AltGr:
Hm, good remark. This would impose some rules on the hierarchy though -- at the moment, the only rule is that you need to have a directory named `<pkg>.<version>`, containing at least an `opam` file, and this at any point in the hierarchy below `<repo>/packages`. The behaviour is unspecified if there are multiple definitions for the same package.

This is a binding choice, so we need to be very careful. A possibility would be to:
* allow opam files (rather than directories) named `<pkg>.<version>` anywhere in the hierarchy.
* have the other files (`url`, `descr`, `files/` subdirectory) affect (and override) _any_ `<pkg>.<version>` files present in the same directory. This would allow the sharing proposed by @hannesm.

This would in practice give two main options for a package in the repository:
* simple, with one file per version: `<repo>/packages/<pkg>/<pkg>.<versions>`, `<repo>/packages/<pkg>/descr`
* like before, with one directory per version, e.g. when `files/` is used: `<repo>/packages/<pkg>.<version>/opam`...

Note that in the presence of sharing, the priority may be more useful the other way around, with the contents of the `opam` file taking precedence over other files.

Another possibility would be to enforce some more structure and have e.g. `<pkg>/<version>.opam`

Yet another scheme would remove all notion of directory hierarchy (except for `files/`), and have `<pkg>.<version>.opam`, `<pkg>[.<version>].descr`, `<pkg>.<version>.url`, `<pkg>.<version>.files/`. This could be nice for management but is fragile wrt user errors and visibility (typos in pkg name, hidden package file in the wrong directory) ; our tools could check that and enforce some policy though.

In any case, opam scans all the files below `<repo>/packages`.

dbuenzli:
I'd say forget about having a single `descr` file for a package. Package descriptions do get tweaked in practice --- since their functionality do evolve over time, new optional sub-libraries and new dependencies are added etc. And the kind of proposed hierarchical overriding always makes it harder than needed to understand what happens in case of problems. 

I'd actually be very happy every package version has everything bundled in a single `opam` file; navigating nested hierarchies is painful.

hannesm:
I agree hierarchical overwriting is painful.  I collected some statistics of the current opam-repository: 1036 packages, 422 have a single release (and thus are not relevant), 42 have a separate descr for each release, 177 have a different descr for some releases, 394 have the same descr for all releases.

Rethinking, having a single descr file feels very cumbersome and premature optimisation. Having each package version contained in a single file sounds like a very good idea.

avsm:
> I wonder how many of descr are actually per-version

I've actually been thinking of storing changelogs in the descr file, until we have first-class support for `opam changelog`.

samoht:
The rational with separating these files at the first place was to avoid unnecessary rebuilds when metadata are modified. Is this still the case?

avsm:
That feature always struck me as a little odd -- wouldn't various fields in the `opam` file also not have "do not rebuild" status, like `license`?

AltGr:
Good point: it shouldn't be a problem, but I'll make sure this is handled 
correctly.

samoht:
@avsm modifying the opam will not trigger a rebuild, that's the whole point. Only changes in the checksum should do (and also in the `url` but that can be discussed). The checksum field was added only for this reason (and not for security, as it uses md5).

avsm:
> modifying the opam file and the descr files will not trigger a rebuild, that's the whole point.

So changing `build:` doesn't currently trigger a rebuild at the moment?  That seems odd...

Drup:
`opam search`'s cold start is still quite slow.. :(

AltGr:
@Drup: according to `strace`, the dirs are still being walked to look for the `findlib` files. The idea is to get rid of those in favor of a `x-findlib:` field in `opam` files, now that extended fields are supported. But do we want to include those in `opam search` blindly ?

Also, pinned packages are not cached (the cache being unique for all switches), but you can't possibly have enough of those to make the delay noticeable ? ;)

Note that on the current version, `opam list` is extremely slow (much slower actually than `search`!) but I'm investigating.

samoht:
> So changing build: doesn't currently trigger a rebuild at the moment? That seems odd...

Indeed. The rational is that if the package is already installed, that means that the build instructions are somehow right :p

AltGr:
The cost of `opam list <arg>` is mostly due to the computation of the installable packages, i.e. Dose's `Algo.Depsolver.trim` called from `OpamSolver.installable`. It seems this jumped up with the addition of dependencies towards the compiler packages in all packages (which actually has no effect on the solution, but is necessary to compute rebuilds and for action ordering in opam).

CCing @abate, who may have an idea on the cost of `Algo.Depsolver.trim` and maybe hints on wether it could be optimised -- from what I gathered with a quick glance, it seems to call the SAT for every package, but there are also caches around the place.

AltGr:
(actually, not adding the compiler dependencies doesn't seem to change much. The cost of the call to `Algo.Depsolver.trim` is around 2 seconds)

Here is [my cudf universe](https://github.com/ocaml/opam/files/30676/univ.txt) -- 3743 packages, close to 30_000 lines -- on which this occurs.

abate:
Algo.Depsolver.trim removes all the packages that are not installable by calling repeatedly calling the solver. That means solving a NP-complete problem 3743 times. There might be room for improvements, but there is a hard computational cost to pay...

AltGr:
Hm, I see -- thanks for the insight. I guess for `opam list` we don't care that much about whether the package are _really_ installable, so we could do a quicker approximation.

abate:
Each time we do an opam update, we can mark all packages that are not installable for each active switch. All subsequent calls will consider only installable packages. I guess the final user will be happier to wait 2 more seconds on update then each time he/she calls opam list ... just an idea.

Drup:
> I guess the final user will be happier to wait 2 more seconds on update then each time he/she calls opam list ... just an idea.

Consider my "list/update" ratio, Yes.

PR #2329, opened 2015-10-21
doc/Manual.md: attempt to clarify version ordering

gasche:
I don't understand much of the current documentation of version order in the manual, which says in particular:

> - version strings are sliced into alternate, possibly empty non-digit / digit
>   sequences, always starting with a non-digit sequence. `1.0~beta2` is thus
>   `["";1;".";0;"~beta";2]`.
>
> - those are sorted lexicographically, using resp. ASCII (with symbol > letter)
>   and number order. For example `a` gives `["a"]`, and `1` gives `["";1]`, so
>  `a` is latest (`"" < "a"`).

What does "resp. ASCII (with symbol > letter)" mean? I understand "symbol" to mean "something other than letter or number" (is this correct?), but in my `utop` I have `Char.code '#' = 35` and `Char.code 'a' = 97`, so do we have `"#" > "a"` as "(symbol > letter)" suggests of `"#" < "a"` as ASCII suggests? I removed these inconsistent mentions of ASCII and symbol/letter ordering (if you have the right answer, tell me and I'll update the patch), and tried to give more explanation about what this alternating sequence stuff means.

### DIFF
--- a/doc/pages/Manual.md
+++ b/doc/pages/Manual.md
@@ -69,16 +69,27 @@ is defined by Version Ordering.
 > <a id="version-ordering">**Version Ordering**</a> follows the basics of the
 > [Debian definition](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Version).
 >
-> It is basically "lexicographical order, with numbers handled properly". In
-> more detail:
+> It is basically "lexicographical order, with numbers handled
+> properly". In more detail (see the [Debian
+> page](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Version)
+> for full details):
 >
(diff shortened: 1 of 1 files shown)


AltGr:
It's _similar_ to the ASCII order, **except** that letters have been put first.

There are more details in the linked [Debian spec](https://www.debian.org/doc/debian-policy/ch-controlfields.html#s-f-Version):

>  The lexical comparison is a comparison of ASCII values modified so that all the letters sort earlier than all the non-letters and so that a tilde sorts before anything, even the end of a part.

Thanks for the improvement!

gasche:
Thanks for the details. I emphasized the link to the Debian spec which is relatively clear, and re-added a remark on non-digit comparisons.

AltGr:
Thanks!

PR #2339, opened 2015-10-30
Allow packages to add env variables

AltGr:
Compilers are allowed to do this, and I am progressively closing
the gap between the two.

### DIFF
--- a/src/state/opamState.ml
+++ b/src/state/opamState.ml
@@ -1999,19 +1999,19 @@ let add_to_env t ?opam (env: env) ?(variables=OpamVariable.Map.empty) (updates: 
   env @ expand_env t ?opam updates variables
 
 let env_updates ~opamswitch ?(force_path=false) t =
-  let comp = compiler_comp t t.compiler in
-
   let add_to_path = OpamPath.Switch.bin t.root t.switch t.switch_config in
   let new_path =
     "PATH",
     (if force_path then "+=" else "=+="),
     OpamFilename.Dir.to_string add_to_path in
+(* Todo: put these back into their packages !
   let perl5 = OpamPackage.Name.of_string "perl5" in
   let add_to_perl5lib =  OpamPath.Switch.lib t.root t.switch t.switch_config perl5 in
   let new_perl5lib = "PERL5LIB", "+=", OpamFilename.Dir.to_string add_to_perl5lib in
   let toplevel_dir =
     "OCAML_TOPLEVEL_PATH", "=",
     OpamFilename.Dir.to_string (OpamPath.Switch.toplevel t.root t.switch t.switch_config) in
+*)
   let man_path =
     let open OpamStd.Sys in
     match os () with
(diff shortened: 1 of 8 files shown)


avsm:
Nice! Would it be possible to have a bit of debugging output (or even comments in the `opam config env` shell output) that ascribe the environment variable to the package? It would help track down the source of these as the repository grows...

src/state/opamState.ml
```diff
@@ -1999,19 +1999,19 @@ let add_to_env t ?opam (env: env) ?(variables=OpamVariable.Map.empty) (updates:
...
 let env_updates ~opamswitch ?(force_path=false) t =
-  let comp = compiler_comp t t.compiler in
-
   let add_to_path = OpamPath.Switch.bin t.root t.switch t.switch_config in
   let new_path =
     "PATH",
     (if force_path then "+=" else "=+="),
     OpamFilename.Dir.to_string add_to_path in
+(* Todo: put these back into their packages !
   let perl5 = OpamPackage.Name.of_string "perl5" in
```
dbuenzli: Btw. never understood what this was doing in my environment.

dbuenzli:
How are environment variable *values* defined ? I doubt `opam` files will know the value in advance. Seems like something that could be done through the `.config` mechanism of #2247

AltGr:
@dbuenzli for now, they are defined with a value from the opam file, that can contain opam variables and allow the [environment update operators](http://opam.ocaml.org/doc/Manual.html#Environmentupdates) -- just like the field in the `.comp` files did. Do you have cases in mind that require more computation to define the environment variable value ?

dbuenzli:
> Do you have cases in mind that require more computation to define the environment variable value ?

Not really but it felt more natural, in the declaration of a state variable of #2247 we could allow to specify an env variable that will hold its value aswell. This would provide a single mecanism for packages to provide information to the switch.

I actually thought that `env:` fields already existed and that they would define an environment for the `build:` and `install:` instructions not sure where I got this idea from. I think a better name than `env:` should be found. Fields currently always apply to the package itself and this has an effect on the switch. Maybe `set-env:` would be better.

dbuenzli:
Btw. environment update operators are dodgy. It means that the value of your environment variable depends on the order in which you installed packages which really doesn't feel like a good idea.

src/state/opamState.ml
AltGr: See cd9c8ea2 for the explanation; now we can get rid of the hack in opam and have this put back in the specific package that needs it!
PATH and MANPATH should probably be the only variables defined from opam, besides of OPAM* variables (well, actually, even those could be defined by the compilers...)

AltGr:
There is indeed a `build-env:` field that has existed for ages, doing what you said. But only a [single package](https://github.com/ocaml/opam-repository/blob/master/packages/sibylfs-lem/sibylfs-lem.0.4.0/opam) on the whole repository uses it.

On update operators: not fond of them, but I don't really see a much better solution. It's not the installation order that will count actually, with this patches the updates will be done in _alphabetical_ order of packages (see [this comment](https://github.com/ocaml/opam/pull/2339/files#diff-477f4eaf05f929c5cac6a24ad69fa183R2023)...)
Anyway, this should be used with care and multiple variables changing the values of the same variables sounds like a bad idea...

PR #2346, opened 2015-11-04
Cache environment updates in a separate file.

OCamlPro-Henry:
The environment updates are cached in <root>/<switch>/environment.
It is updated whenever the <root>/<switch>/state is updated.

This might be used by external tools (e.g. opam-manager) to read the
environment updates without having to depend on `OpamState`. This might
also permit to add a "costly topological sort" when computing the
environment updates.

### DIFF
--- a/src/format/opamFile.ml
+++ b/src/format/opamFile.ml
@@ -437,6 +437,34 @@ module Pinned_legacy = LineFile(struct
 
   end)
 
+
+(** Cached environment updates (<switch>/environment) *)
+
+module Environment = LineFile(struct
+
+    let internal = "environment"
+
+    type t = env_update list
+
+    let empty = []
+
+    let pp =
+      Pp.lines_set ~empty:[] ~add:OpamStd.List.cons ~fold:List.fold_right @@
+      Pp.identity ^+
+      Pp.of_pair "env_update_op"
+        (env_update_op_of_string, string_of_env_update_op) ^+
+      Pp.identity ^+
+      Pp.opt Pp.singleton
+
+    let pp =
+       pp -|
+       Pp.map_list
+         (Pp.pp
+           (fun ~pos:_ (a, (b, (c, d))) -> (a, b, c, d))
+           (fun (a, b, c, d) -> (a, (b, (c, d)))))
+
+  end)
+
 (** (2) Part of the public repository format *)
 
 (** repository index files ("urls.txt"): table
--- a/src/state/opamState.ml
+++ b/src/state/opamState.ml
@@ -1399,11 +1399,6 @@ let switch_state t =
     compiler = t.compiler_packages;
     pinned = t.pinned; }
 
-let write_switch_state t =
-  if not OpamStateConfig.(!r.dryrun) then
-    let f = OpamPath.Switch.state t.root t.switch in
-    OpamFile.State.write f (switch_state t)
-
 let universe t action =
   let opams = (* Read overlays of pinned packages *)
     OpamPackage.Name.Map.fold (fun name pin map ->
@@ -2038,15 +2033,8 @@ let add_to_env t ?opam (env: env) ?(variables=OpamVariable.Map.empty)
   in
   env @ expand_env t ?opam updates variables
 
-let env_updates ~opamswitch ?(force_path=false) t =
-  let add_to_path = OpamPath.Switch.bin t.root t.switch t.switch_config in
-  let new_path =
-    "PATH",
-    (if force_path then PlusEq else EqPlusEq),
-    OpamFilename.Dir.to_string add_to_path,
-    Some "Current opam switch binary dir"
-  in
-(* Todo: put these back into their packages !
+let compute_env_updates t =
+  (* Todo: put these back into their packages !
   let perl5 = OpamPackage.Name.of_string "perl5" in
   let add_to_perl5lib =  OpamPath.Switch.lib t.root t.switch t.switch_config perl5 in
   let new_perl5lib = "PERL5LIB", "+=", OpamFilename.Dir.to_string add_to_perl5lib in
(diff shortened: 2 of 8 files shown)


src/format/opamFile.ml
```diff
@@ -437,6 +437,34 @@ module Pinned_legacy = LineFile(struct
...
+
+    type t = env_update list
+
+    let empty = []
+
+    let pp =
+      Pp.lines_set ~empty:[] ~add:OpamStd.List.cons ~fold:List.fold_right @@
+      Pp.map_right
+        (fun (a, (b, (c, d))) -> (a, b, c, d))
+        (fun (a, b, c, d) -> (a, (b, (c, d)))) @@
```
AltGr: Isn't this equivalent to:
```
Pp.pp
  (fun (a, (b, (c, d))) -> (a, b, c, d))
  (fun (a, b, c, d) -> (a, (b, (c, d)))) -|
```
?

src/format/opamFormat.ml
```diff
@@ -245,6 +245,20 @@ module Pp = struct
...
+    name_constr = t.name_constr;
+  }
+
+  let map_left f g t = {
+    parse = (fun ~pos x -> t.parse ~pos (g x));
+    print = (fun x -> f (t.print x));
+    name = t.name;
+    name_constr = t.name_constr;
+  }
+
```
AltGr: See above, I think you can define a pp with the two functions and use the normal pipe operator `-|` for the same result.

src/state/opamState.ml
```diff
@@ -2083,7 +2067,30 @@ let env_updates ~opamswitch ?(force_path=false) t =
...
     else []
   in
-  new_path :: (man_path @ switch @ root @ comp_env @ pkg_env)
+  man_path @ root @ comp_env @ pkg_env
+
+let env_updates ~opamswitch ?(force_path=false) t =
+  let update =
+    try
+      OpamFile.Environment.read (OpamPath.Switch.environment t.root t.switch)
+    with _ ->
```
AltGr: Forbidden! See https://github.com/ocaml/opam/blob/ba3db33874ff1c09b1384c277d8b903748543087/src/core/opamStd.mli#L256-L261

AltGr:
Wow! What a surprise to see patches on code I wrote so recently (and that may still have some rough edges) :)
This is a very good idea. It'll also be easier to debug and the `opam-init/variables.sh` files can be generated in a simpler way. Does this work well with the display of the variables' origin ? (3a231aa2d6df7822e50f77dc9bfabd8d79b564f8, which yields e.g.:
```
[lg@agaric ~/ocamlpro/opam]% opam config env -v
: Updated by compiler 4.02.3;
CAML_LD_LIBRARY_PATH="/home/lg/.opam/4.02.3/lib/stublibs"; export CAML_LD_LIBRARY_PATH;
: Current opam switch man dir;
MANPATH=":/home/lg/.opam/4.02.3/man"; export MANPATH;
: Current opam switch binary dir;
PATH="/home/lg/.opam/4.02.3/bin:/home/lg/bin:/home/lg/local/bin:/usr/local/bin:/usr/bin:/bin:/usr/bin/X11:/usr/games:/usr/sbin:/sbin"; export PATH;
```
)

AltGr:
I was actually just writing tests on the `setenv:` field, since I wrote it but didn't test at all yet.

AltGr:
Now that I fixed the `setenv` field I need this so that the cost of `opam config env` doesn't increase too much :)

src/format/opamFormat.ml
OCamlPro-Henry: Indeed, fixed.

src/state/opamState.ml
OCamlPro-Henry: Ok! I added a proper test.

src/format/opamFile.ml
```diff
@@ -437,6 +437,38 @@ module Pinned_legacy = LineFile(struct
...
+
+    let pp =
+      Pp.lines_set ~empty:[] ~add:OpamStd.List.cons ~fold:List.fold_right @@
+      Pp.identity ^+
+      Pp.of_module "env_update_op"
+        (module struct
+          type t = env_update_op
+          let to_string = string_of_env_update_op
+          let of_string = env_update_op_of_string
+        end : Pp.STR with type t = env_update_op) ^+
```
AltGr:
```
Pp.pp ~name:"env_update_op" string_of_env_update_op env_update_op_of_string
```
would be shorter :)

src/client/opamClient.ml
```diff
@@ -155,6 +155,9 @@ let with_switch_backup command f =
   let t = OpamState.load_state command OpamStateConfig.(!r.current_switch) in
   let file = OpamPath.Switch.backup t.root t.switch in
   OpamFilename.mkdir (OpamPath.Switch.backup_dir t.root t.switch);
+    OpamFile.Environment.write
+      (OpamPath.Switch.environment t.root t.switch)
+      (OpamState.compute_env_updates t);
```
AltGr: Why is this needed here ? It's not backup-related and the file should already be there, no ?

OCamlPro-Henry: Indeed... removed.

src/format/opamFile.ml
OCamlPro-Henry: Included. Thanks.

src/format/opamFile.ml
```diff
@@ -437,6 +437,34 @@ module Pinned_legacy = LineFile(struct
...
+
+    type t = env_update list
+
+    let empty = []
+
+    let pp =
+      Pp.lines_set ~empty:[] ~add:OpamStd.List.cons ~fold:List.fold_right @@
+      Pp.identity ^+
+      Pp.pp ~name:"env_update_op"
+        env_update_op_of_string string_of_env_update_op ^+
```
AltGr: My  bad, the `of_string` takes an additional `~pos` argument. You can either wrap in `(fun ~pos:_ -> xxx)`, or use `Pp.of_pair "env_update_op" (env_update_op_of_string, string_of_env_update_op)` instead.

Other than that, all good it seems. Thanks!

OCamlPro-Henry: Pff... I did not even try to compile... :-( Fixed.

PR #2348, opened 2015-11-04
OpamFile: allow 'quoted string' in 'LineFile'

OCamlPro-Henry:


AltGr:
What does this do exactly ? Did you find a bug or annoying limitation ? This part of the escaping is a bit fragile, so I am wary of modifying it.

Also, Scanf.unescaped is absent in 3.12 (which we're still supporting for building opam, although we'll ditch it at some point) ; also it's a bit [dangerous](http://caml.inria.fr/mantis/view.php?id=6958)

Since Linefile is not "supposed" to be written by hand, I thought escaping of spacing characters should be enough ; it also gives us a simpler definition of the format than OCaml escapes.

See also the _other_ version of escaping, concerning inline strings, in OpamFormat
(of which I recently added a second variant for cleaner multiline strings between `"""`)

PR #2359, opened 2015-11-17
new pef interface

abate:
This pull request is the result of my work toward the integration of the dose module opam into the Opam code base. It basically leverage the pef to cudf translation routines in dose following the opam pef interface described here : https://github.com/abate/opam/blob/master/doc/design/multi-switch.txt .

These patches try to touch as little as possible of the opam data structures, but it makes also part of the code redundant. A new round of patches will focus on removing old code and refactoring and later.

I don't think this patch should be integrated as is in master, but I'd love a bit of feedback to avoid integration problems along the track. I've merged the lastes opam/master HEAD today. I'll rework my patches in the next few days and update this branch accordingly.

The core of this patch is the OpamState.universe record with two new fields I use to hold the list of pef packages.


abate:
These last two patches of this series is to decouple the OpamState.Types.t data structure in two. One part holding the repository wise data, and a second for switch specific data structure. The idea is to move the opam code to handle multi switch requests and therefore we should be able to operate on multiple switches at the same time. The next round of patches will focus to add an upgrade-all functionality.

src/state/opamState.ml
```diff
@@ -1558,32 +1766,64 @@ let installed_versions t name =
...
+    let sw = OpamSwitch.Set.inter all_switches action_switches in
+    OpamSwitch.Set.fold (fun s acc -> s::acc) sw []
+  in
+  Printf.fprintf stderr "aaaaaaaaaa\n";
+  List.iter (fun par ->
+    List.iter (fun (k,(_,v)) -> Printf.fprintf stderr "%s: %s\n" k v) (List.rev par);
+    Printf.fprintf stderr "\n"
+  ) (pef_package switches t)
+  ;
+  Printf.fprintf stderr "aaaaaaaaaa\n";
```
AltGr: forgot to remove these to debugs ?

src/state/opamState.ml
```diff
@@ -1478,67 +1511,242 @@ let string_of_cnf string_of_atom cnf =
...
+    | _               -> None
+  in
+  OpamFilter.eval_to_bool ~default:true env (OpamFile.OPAM.available opam)
+  && consistent_ocaml_version ct opam
+
+let pef_package ?(orphans=OpamPackage.Set.empty) switches t =
+  let depends   = OpamPackage.Map.map OpamFile.OPAM.depends t.opams in
+  let depopts   = OpamPackage.Map.map OpamFile.OPAM.depopts t.opams in
+  let conflicts = OpamPackage.Map.map OpamFile.OPAM.conflicts t.opams in
+  let maintainers = OpamPackage.Map.map OpamFile.OPAM.maintainer t.opams in
```
AltGr: This seems unneeded and causes each package to be looked up 4 times instead of one

src/format/opamTypes.mli
```diff
@@ -262,9 +273,9 @@ type 'a request = {
 
 (** user request action *)
 type user_action =
-  | Install of name_set (** The 'root' packages to be installed *)
-  | Upgrade of package_set (** The subset of packages to upgrade *)
-  | Reinstall of package_set
+  | Install of (name_set * switch_set) (** The 'root' packages to be installed *)
+  | Upgrade of (package_set * switch_set) (** The subset of packages to upgrade on each switch *)
+  | Reinstall of (package_set * switch_set)
```
AltGr: These feel like they should be `(name * switch)` sets. What are the reasons to have this limitation ?
(I understand it may be more awkward and require more changes to not have a package/name set directly available?)

src/format/opamTypes.mli
```diff
@@ -273,21 +284,14 @@ type user_action =
 
 (** Solver universe *)
 type universe = {
+  u_pefuniv  : ((string * string), Opam.Packages.package) Hashtbl.t;
+  u_options  : (string * string list * string list);
+  u_tables   : Pef.Pefcudf.tables;
```
AltGr: What do these contain exactly ?
Are they really mutable ? (the structure used to be immutable)

src/client/opamClient.ml
```diff
@@ -153,8 +154,9 @@ let details_of_package_regexps t packages ~exact_name ~case_sensitive regexps =
 
 let with_switch_backup command f =
```
AltGr: Does it make sense to have a backup only for the current switch when several can be modified (e.g. through `upgrade_all`) ?

src/format/opamTypes.mli
abate: We can substitute these with sets/maps . They are not supposed to be mutable. On the other side I do many Hashtbl.find, so I think that a O(1) data structure makes sense here. (in particular for the tables/version_map ...

src/client/opamClient.ml
abate: no, you are correct. There are still many things to fix ...

src/format/opamTypes.mli
abate: I imagines the user would ask "install package A on switch sw1 and sw2" more than "install package A on switch sw1 and package B on switch s2" . But yes, you are correct (name * switch) sets are more general, and it should not be to much of a problem to adapt the code.

src/format/opamTypes.mli
AltGr: No problem then, currently the conversion makes it quite costly to convert from opam-state to cudf universe, which is annoying e.g. when you want to just compute a set of transitive dependencies or a topological sort. We already use Graph.imperative -- with a functional interface on top.

It should just be documented, and maybe the types named: what are all the strings ? I am curious about the `u_options` too.

So using hashtables is not a problem, but it's good to know whether they can be muted or not: it has implications on the validity of a computed universe.

src/client/opamClient.ml
AltGr: Now the switch state is directly stored in the export format, so exporting the switch amounts to just backing up the `~/.opam/<switch>/state` file: much cheaper than it used to be!

src/format/opamTypes.mli
abate: u_options is a triple (current switch, available switches, and build options - doc, build, tests) . These are used to then convert pef packages to cudf packages.

src/state/opamState.ml
abate: I don't understand this comment here. pef_package is called on t . For each package I call a function that uses depends,depopts,conflicts and maintainers . For I create these variables once and I use them for every package (let d = OpamPackage.Map.find package depends in ... ) . Or the comment was related to the function related to filters ?

src/format/opamTypes.mli
AltGr: This is great, as it means we won't have to reload different kinds of universes, doing lots of duplicate conversions in the way, just to compute the different dependency graphs. Also, even at the upper level, we could load the universe just once and pass it along, specifying the action, doc, build and test flags just when we need a specific dependency graph (in the current setting, it could be part of the switch state, but for multi-switch, we could have it along the switch-name=>switch-state map in a new `multiswitch-state` structure).

I am not fond of using strings for flags, but not sure what the best interface would be. Maybe keep the higher-level types at this point, and do the conversion in the interface with Dose ? Or just make this more opaque to the higher-level, since we should only need to load once ? Well, that's mostly aesthetic concerns anyway.

Something nice we could do is actually get rid of the `doc` and `tests` flags at the lower level (they'll never change by then, they're defined globally, and even if we wanted to define them per-package we would do this at the upper level). The two "dependency toggles" that we really need are actually `build` and `depopt`, which play a similar role despite being defined completely differently. So if we could have a single universe in Dose, with these two toggles on dependencies, that would be awesome and result in big clarity and performance gains, by making the factorisation of most of the translations possible. What do you think ?

src/state/opamState.ml
AltGr: My thought was that you could loop through packages and, for each package, lookup once in t.opams directly and get all of depends, depopts, conflicts and maintainers ; rather than using specific maps and looking up each package in each of them. But maybe did I miss something ?

PR #2386, opened 2015-12-18
Rename a variable

yallop:
If the keyword `implicit` were ever added to the language, this change would allow OPAM to continue to build.


AltGr:
Thanks!

PR #2389, opened 2015-12-22
make cold: Fix order of PATH variable 

avsm:
so that local OCaml overrides any previously installed one (such as 3.11.x that is too old in CentOS 6)

Same fix needs to go into master, but this 1.2 one unblocks container builds for Travis CI


AltGr:
Thanks!

PR #2397, opened 2016-01-06
Minor simplification in OpamState.

OCamlPro-Henry:
Useful for `opam-manager`. This allows to reuse `update_env` without
building a fake `OpamState.t`.


AltGr:
I think this is now obsolete, please reopen if I'm wrong!

PR #2413, opened 2016-01-22
Unit tests

abate:
Add basic infrastructure to add a functional test suite for opamlib based on oUnit.

At the moment large part of the code is boilerplate to test the solver and state sub-libraries. 
Contributions for real tests are more then welcome.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -314,6 +314,7 @@ opam-tests: opam ALWAYS
 define PROJ_opam-tests
   SOURCES = \
 						tests/testsCommon.ml \
+						tests/testsScenarios.ml \
 						tests/solverTests.ml \
 						tests/stateTests.ml
   RESULT = opam-tests
--- a/src/Makefile
+++ b/src/Makefile
@@ -307,6 +307,20 @@ define PROJ_opam-admin-top
 endef
 export PROJ_opam-admin-top
 
+opam-tests: opam ALWAYS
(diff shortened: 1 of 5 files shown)


samoht:
Note that there might be an overlap with opam-rt: https://github.com/ocaml/opam-rt

but obviously more tests are always better :-)

abate:
if I understand correctly, opam-rt aims to test the application opam. Here I'm trying to add a test suite for the library (function by function), and completely ignoring the user interface.

samoht:
yes I see that and that's a very good thing! Just mentioning opam-rt as there are also some tests with package universe than can maybe be sharing (using the lib as you did, not the cli as opam-rt did).

AltGr:
Awesome, thanks a lot for going through the burden of writing these!

AltGr:
I think we should add those to the travis script already

abate:
At the moment I'm building the infrastructure (slowly adding oUnit - aware testing modules for all main data types) . Once I have a few meaningful tests I'll let you know. 

But I really really hope not to be the only one writing tests, otherwise I'll go crazy :)

Moreover, while writing tests, we can easily realize the level of abstraction of a given function, and the per-requisites to run it. I've already noticed that many functions have side effects, and relies on global variables. To make everything work properly I've to write a few set-up/tear-down brackets functions and maybe I'll re-use the code in opam-rt.

AltGr:
Sadly, without more love, this got outdated.

PR #2422, opened 2016-01-28
Split state root

abate:
Add a global OpamPath.root reference to the current opam root directory and remove the "root' arguments from all functions in OpamPath. The first patch removes global_state.root from the OpamSwitchTypes. The second patch removes all occurences of OpamStateConfig.(!r.root_dir) .


samoht:
I am not sure this go in the right direction. Would be great if opam-lib would rely less on side-effects...

abate:
this patch is mainly to get rid of global_state.root and pave the way the multi-switch patch in the works

AltGr:
hm, Github automatically closed this because I renamed the split-state branch to `next`. That was not my intention.

abate:
not sure what to do with these patches. Do you want me to refresh and make you a new pull request on **next** (including the root patch) ? Otherwise, if you take care of the rot patch, I can resume working on refreshing all other pending patches... The important thing is to get there :)

AltGr:
I started merging it but got caught by stg else again. I'll have it in asap!

@samoht, this doesn't represent any regression in terms of side-effect, merely makes things more consistent (and makes it clear what's going on).
Having something obtained by side-effects sometimes but sometimes through an argument is the worst if it has to always be the same value anyways. And lots of stuff, including global options, are dependent on the opam root, so it's better not to _pretend_ you can pass it through `global_state` if there is no chance it will work.

Of course, it would be cleaner to define it all in a record, pass it everywhere and depend on that only. But our compromise is OK -- global refs only in OpamXxxConfig modules, and intialised at startup -- as long as we make it clear. In other words, replacing:
```
let f root x = process root x
let g () = f OpamStateConfig.(!r.root) x
```
by
```
let f x = process OpamStateConfig.(!r.root) x
let g () = f x
```
is I think a good thing, esp. if we don't trust that `process` doesn't internally depend on `OpamStateConfig.(!r.root)`.

abate:
:+1: let me know when it's ready and I'll proceed to refresh all other patches.

samoht:
@AltGr I completely agree in being more explicit about side-effects. But I also think we should really go towards a side-effect free API if possible... that'd be a much nicer API for consumers (currently, the API is still quite hard to use).

AltGr:
@samoht: I've been spending a huge amount of time improving that and the general code quality since 1.2.2, and it's mostly invisble to the users who are have been waiting for many new features for a long time -- although in doing so I was paving the way to including them easily. The only globals are now very well delimited in the OpamXxxConfig modules now, and represent a good compromise between code clarity and purity imho.

@abate: I'm sorry, but I have already spent too long trying to merge this, and there are just too many conflicts for it to be reasonable. At this stage rewriting it would, I think, be quicker (and certainly less error-prone) :(.
But by looking at opam code, the instances of OpamStateConfig.(!r.root_dir) are actually very few in comparison, so it may be worth it to go the other way around, and make sure to pass the root everywhere -- including from OpamMain, where it is originally computed. I intended anyway to load `global_state` at this point only, as the next stage of the split-state rewrite. As for the quite verbose `st.switch_global.root`, we could simply add a shortcut into the switch state, or maybe even pass the global state to the `OpamPath` functions.

abate:
ok either ways. In my opinion passing 'root' to every single function in the code is a bit useless (considering that is always the same root). If you want, I can give it another shot, and see if I can find another side-effect free solution (that probably is going to be a bit verbose). Just ping me when you think the code is in a stable state, so I can refresh all the multi-arch patches and try to converge.

dbuenzli:
> (considering that is always the same root).

If opam moves to per project package management at a certain point that may no longer be the case.

Issue #2255, opened 2015-07-13
`opam init` breaks fish shell with configuration


dongcarl:
Tested on Ubuntu, Mac OS X, and FreeBSD.

When I invoke opam init, this is inserted in my config.fish

```
. /home/dongcarl/.opam/opam-init/init.fish > /dev/null 2> /dev/null or true
```

I'm not sure exactly how but this line renders the fish shell completely unusable. I have resorted to replacing the aforementioned line with `eval (opam config env)`

AltGr:
Can you give the precise opam version ? (`opam config report`)

The output of manually running `. /home/dongcarl/.opam/opam-init/init.fish` could help.

Thanks for reporting!

dongcarl:
Here's the config report
```
# OPAM config report
# opam-version    1.2.2 
# self-upgrade    no
# os              darwin
# external-solver aspcud $in $out $criteria
# criteria        -count(removed),-notuptodate(request),-sum(request,version-lag),-count(down),-notuptodate(changed),-count(changed),-notuptodate(solution),-sum(solution,version-lag)
# jobs            4
# repositories    1* (http), 1 (version-controlled)
# pinned          0
# current-switch  system*
# last-update     2015-07-11 14:29
```

I've talked to people on the mailing list about this and come up with a fix thanks to Zanchey on fish shell IRC. 

The proper configuration would have the following files:

config.fish
```
# OPAM configuration
. /Users/dongcarl/.opam/opam-init/init.fish > /dev/null 2> /dev/null; or true
```

init.fish
```
# Load the environment variables
. /Users/dongcarl/.opam/opam-init/variables.fish > /dev/null 2> /dev/null; or true
```

The above two files have a semicolon before the "or true"

variables.fish
```
set -gx PATH "/Users/dongcarl/.opam/system/bin" $PATH;
set -gx OCAML_TOPLEVEL_PATH "/Users/dongcarl/.opam/system/lib/toplevel";
set -gx PERL5LIB "/Users/dongcarl/.opam/system/lib/perl5:$PERL5LIB";
set -gx MANPATH $MANPATH "/Users/dongcarl/.opam/system/man";
set -gx OPAMUTF8MSGS "1";
set -gx CAML_LD_LIBRARY_PATH "/Users/dongcarl/.opam/system/lib/stublibs:/usr/local/lib/ocaml/stublibs";
```

The above file doesn't have double quotes around the $PATH and the $MANPATH

[...]

PR #2424, opened 2016-01-29
Fix Fish shell error in opamState.ml

roryokane:
In Fish, `or` is a command of its own, not a syntax element. So you need a semicolon before it to start a new command. Reference: [`fish` documentation: commands – `or`](http://fishshell.com/docs/current/commands.html#or)

This should be the last fix needed to solve https://github.com/ocaml/opam/issues/2255 – the other issues were fixed by 758c3912cd1661d116a3bca47e52c2881a95afd5 (use `source`) and c87a2802fe73e94599c1064afac72c45b7d534f6 (don’t quote array variables).


AltGr:
Thanks a lot!

PR #2444, opened 2016-02-23
opam config exec now does not set OPAMSWITCH unless --switch or OPAMSWITCH

dsheets:
Previously, `opam config exec` would always set the `OPAMSWITCH` environment variable. This causes issues when running processes like shells with `opam config exec -- bash` or similar because the shell's environment is not precisely the same as the parent process's environment. In partiular, the absence of `OPAMSWITCH` was not preserved and therefore the use of interactive opam commands was more unwieldy.

Using `opam config exec` to spawn a shell is a common technique when scripting with opam in order to avoid needing to run a separate command to set up the environment in a subshell. See <http://lists.ocaml.org/pipermail/opam-devel/2016-February/001393.html>.

### DIFF
--- a/src/client/opamConfigCommand.ml
+++ b/src/client/opamConfigCommand.ml
@@ -220,8 +220,9 @@ let exec ~inplace_path command =
     with
     | []        -> OpamSystem.internal_error "Empty command"
     | h::_ as l -> h, Array.of_list l in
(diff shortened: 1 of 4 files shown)


AltGr:
Great, thanks!

PR #2445, opened 2016-02-23
Backport #2444 to 1.2.x

dsheets:
Previously, `opam config exec` would always set the `OPAMSWITCH` environment variable. This causes issues when running processes like shells with `opam config exec -- bash` or similar because the shell's environment is not precisely the same as the parent process's environment. In partiular, the absence of `OPAMSWITCH` was not preserved and therefore the use of interactive opam commands was more unwieldy.

(cherry picked from commit a19c1384177c6080f21516e674929cabbcf34565 submitted as #2444)


AltGr:
Thanks!

Issue #2438, opened 2016-02-18
opam fails to install 3.11.2 (curl: code 404 while downloading http://www.ocamlpro.com/patches/3.11.2_binutils.patch)


JasonGross:
```bash
$ opam switch 3.11.2

=-=- Installing compiler 3.11.2 -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
[default.comp] http://caml.inria.fr/pub/distrib/ocaml-3.11/ocaml-3.11.2.tar.gz downloaded
[ERROR] curl: code 404 while downloading http://www.ocamlpro.com/patches/3.11.2_binutils.patch
Fatal error:
# opam-version    1.2.2
# os              linux
curl: code 404 while downloading http://www.ocamlpro.com/patches/3.11.2_binutils.patch
```

AltGr:
Ah, OCamlPro has migrated its main web server, and this may have been forgotten in the process. I forwarded the issue, thanks for reporting!

AltGr:
This has been fixed. Thanks!

PR #2449, opened 2016-02-28
Add automatic fallbacks using backup-urls

lefessan:
This PR provides a solution to the problem of archives becoming unavailable (because the site is down, or because the compression algorithm changed). It is based on looking up archives, on failure, from a configuration parameter called `backup-urls`, that is initially initialized to OCamlPro's backup of archives (http://opam.ocamlpro.com/backup).

Initial configuration:
backup-urls: [ "http://opam.ocamlpro.com/backup" ]

Disabling the feature:
backup-urls: []

### DIFF
--- a/src/format/opamFile.ml
+++ b/src/format/opamFile.ml
@@ -797,6 +797,7 @@ module ConfigSyntax = struct
     dl_jobs : int;
     solver_criteria : (solver_criteria * string) list;
     solver : arg list option;
+    backup_urls : string list;
   }
 
   let opam_version t = t.opam_version
@@ -804,6 +805,7 @@ module ConfigSyntax = struct
   let switch t = t.switch
   let jobs t = t.jobs
   let dl_tool t = t.dl_tool
+  let backup_urls t = t.backup_urls
   let dl_jobs t = t.dl_jobs
   let criteria t = t.solver_criteria
   let criterion kind t =
@@ -815,6 +817,7 @@ module ConfigSyntax = struct
   let with_repositories t repositories = { t with repositories }
   let with_switch t switch = { t with switch }
   let with_jobs t jobs = { t with jobs }
+  let with_backup_urls t backup_urls = { t with backup_urls }
   let with_dl_tool t dl_tool = { t with dl_tool = Some dl_tool }
   let with_dl_jobs t dl_jobs = { t with dl_jobs }
   let with_criteria t solver_criteria = { t with solver_criteria }
@@ -823,10 +826,13 @@ module ConfigSyntax = struct
                (kind,criterion)::List.remove_assoc kind t.solver_criteria }
   let with_solver t solver = { t with solver = Some solver }
 
+  let backup_url_ocamlpro = "http://opam.ocamlpro.com/backup"
+
   let create switch repositories ?(criteria=[]) ?solver jobs ?download_tool dl_jobs =
     { opam_version = OpamVersion.current;
       repositories ; switch ; jobs ; dl_tool = download_tool; dl_jobs ;
-      solver_criteria = criteria; solver }
+      solver_criteria = criteria; solver;
+      backup_urls = [backup_url_ocamlpro] }
 
   let empty = {
     opam_version = OpamVersion.current;
@@ -837,6 +843,7 @@ module ConfigSyntax = struct
     dl_jobs = 1;
     solver_criteria = [];
     solver = None;
+    backup_urls = [backup_url_ocamlpro];
   }
 
   let fields =
@@ -874,6 +881,9 @@ module ConfigSyntax = struct
       "solver-fixup-criteria", Pp.ppacc_opt
         (with_criterion `Fixup) (criterion `Fixup)
         Pp.V.string;
+      "backup-urls", Pp.ppacc
+        with_backup_urls backup_urls
+        (Pp.V.map_list ~depth:1 Pp.V.string);
       "solver", Pp.ppacc_opt
         with_solver solver
         (Pp.V.map_list ~depth:1 Pp.V.arg);
--- a/src/format/opamFile.ml
+++ b/src/format/opamFile.ml
@@ -826,13 +826,11 @@ module ConfigSyntax = struct
                (kind,criterion)::List.remove_assoc kind t.solver_criteria }
   let with_solver t solver = { t with solver = Some solver }
 
-  let backup_url_ocamlpro = "http://opam.ocamlpro.com/backup"
-
   let create switch repositories ?(criteria=[]) ?solver jobs ?download_tool dl_jobs =
     { opam_version = OpamVersion.current;
       repositories ; switch ; jobs ; dl_tool = download_tool; dl_jobs ;
       solver_criteria = criteria; solver;
-      backup_urls = [backup_url_ocamlpro] }
+      backup_urls = [] }
 
   let empty = {
     opam_version = OpamVersion.current;
@@ -843,7 +841,7 @@ module ConfigSyntax = struct
     dl_jobs = 1;
     solver_criteria = [];
     solver = None;
-    backup_urls = [backup_url_ocamlpro];
+    backup_urls = [];
   }
 
   let fields =
--- a/src/format/opamFile.mli
+++ b/src/format/opamFile.mli
@@ -92,6 +92,10 @@ module Config: sig
 
   val dl_tool: t -> arg list option
 
+ (* Field "backup-urls": a list of urls from which it is possible
+    to download by checksum, i.e. $URL/9/a/b/9ab48749e8201... *)
+  val backup_urls: t -> string list
+
   (** Return the number of download jobs *)
   val dl_jobs: t -> int
 
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,8 @@
+1.3.0
+* New field 'backup-urls' can be set to a list of URLs providing archive
+  mirroring by hash, in case a link to an archive is dead. Currently, such
+  a service is provided by "http://opam.ocamlpro.com/backup"
+
 1.2.2
 * Fixed wrong locks being taken during `switch reinstall` (#2051)
 * Fixed `config report` that wasn't displaying the external solver (#2059)
--- a/src/repository/opamRepository.ml
+++ b/src/repository/opamRepository.ml
@@ -61,10 +61,12 @@ let pull_url package local_dirname checksum remote_url =
   let rec attempt = function
     | [] -> assert false
     | [url] -> pull url
-    | url::mirrors ->
+    | url:: (mirror :: _ as mirrors) ->
       pull url @@+ function
       | Not_available s ->
-        OpamConsole.warning "download of %s failed, trying mirror" s;
+        OpamConsole.warning "download of %s at %s failed, trying mirror %s"
+          (OpamPackage.to_string package) s
+          (OpamUrl.to_string mirror);
         attempt mirrors
       | r -> Done r
   in
--- a/src/state/opamState.ml
+++ b/src/state/opamState.ml
@@ -632,6 +632,7 @@ let descr_opt t nv =
 let descr t nv =
   OpamStd.Option.Op.(descr_opt t nv +! OpamFile.Descr.empty)
 
+(* parse and return the 'url' file for this package *)
 let url t nv =
   OpamStd.Option.Op.(opam_opt t nv >>= OpamFile.OPAM.url)
 
@@ -3100,8 +3101,23 @@ let download_upstream t nv dirname =
   | None   -> Done None
   | Some u ->
     let remote_url = OpamFile.URL.url u in
-    let mirrors = remote_url :: OpamFile.URL.mirrors u in
     let checksum = OpamFile.URL.checksum u in
+    let mirrors = [] in
+    let mirrors = match checksum with
+      | None -> mirrors
+         (* if we have a checksum, we can query the backup_urls *)
+      | Some checksum -> (* checksum is already in hexa *)
+        let backup_urls = OpamFile.Config.backup_urls t.config in
+        List.map (fun backup_url ->
+          let link = Printf.sprintf "%s/%c/%c/%c/%s"
+            backup_url
+            checksum.[0] checksum.[1] checksum.[2]
+            checksum in
+          OpamUrl.of_string link
+        ) backup_urls @ mirrors
+    in
+    let mirrors = OpamFile.URL.mirrors u @ mirrors in
+    let mirrors = remote_url :: mirrors in
     let text =
       OpamProcess.make_command_text (OpamPackage.name_to_string nv)
         (OpamUrl.string_of_backend remote_url.OpamUrl.backend)


dsheets:
This seems like a property of a repository that belongs in the `repo` metadata file of the repository.

src/format/opamFile.mli
```diff
@@ -92,6 +92,10 @@ module Config: sig
 
   val dl_tool: t -> arg list option
 
+ (* Field "backup-urls": a list of urls from which it is possible
+    to download by checksum, i.e. $URL/9/a/b/9ab48749e8201... *)
```
samoht: indexing the files by md5 checksums is a bad idea.

samoht:
There is already a `mirror` field in `url` files, so it makes sense to lift it to repo-metadata.

src/format/opamFile.mli
lefessan: The service is not tied to md5 checksums, it can store any checksum (md5, sha1, etc.). It is restricted to md5 here just because they are the only checksums used by OPAM.

lefessan:
Although it could be seen as a repository meta-data, I don't think it should be, for several reasons:
* the backup service is a general service, that can backup archives from many origins. It could be extended and thus useful for more than just the official repository. Restricting it to be per-repository would prevent a repository from benefiting from what has been archived in another repository;
* many people have local copies of the official repository, but without the associated meta-data. Such local repositories would not benefit from the backup, although they are actually providing the same packages;
* if somebody wants to disable backup-urls, it's probably on a global basis (i.e. for security reasons), and not per-repository.

samoht:
> the backup service is a general service

I agree. So design a general solution which 1/ is not tied up to the opam source-code and 2/ which does force the various packages repo (which are not all public, nor use OCaml) to use OCamlPro's infrastructure.

lefessan:
> I agree. So design a general solution which 1/ does not hard-code hostnames in the opam source-code and 2/ does not force the various package repositories (which are not all public, nor use OCaml) to use OCamlPro's infrastructure.

1/ OPAM is already hardwired to use to http://opam.ocaml.org/, even if you don't use it for OCaml.
2/ What is the most important here: shall we improve the UX of all OCaml users, by providing an automatic alternative to dead urls, or block that feature, because a few users are worried for their own different use of OPAM, of a feature that is not even going to change anything for them.
3/ backup-urls are a fallback mechanism. If all your links are correct in the repository, you will never use the feature. Morover, you can easily disable it automatically:

```
opam init my-repo http://opam.foobar.com/
echo 'backup-urls: []' >> .opam/config
```
Finally, the worst thing that could happen would be your client querying the service for the checksum of an archive that is not there. 

Do you really know anybody in the community that is rightly concerned by a md5 checksum fleeing over the Internet ?

samoht:
Sigh...

1/ yes we are using `http://opam.ocaml.org/` and that's the only hard-coded name that we need to bootstrap the process. There are other bits which are OCaml-specific but they are being moved away.
2/ adding a field in `https://github.com/ocaml/opam-repository/blob/master/repo` will have the exact same UX result, but with a more flexible approach as it let repo maintainers tweak the configuration.

Also I am very worried by relying on md5 for backup purposes... Currently we are using it in opam detect file changes so a collision will just cause a recompilation to not happen (which is kind of a good side-effect, if the upstream archive has been compromised anyway). In your proposal, a collision is quite bad...

lefessan:
> 1/ yes we are using http://opam.ocaml.org/ and that's the only hard-coded name that we need to bootstrap the process. There are other bits which are OCaml-specific but they are being moved away.
> 2/ adding a field in https://github.com/ocaml/opam-repository/blob/master/repo will have the exact same UX result, but with a more flexible approach as it let repo maintainers tweak the configuration.

I have done this patch that would improve the UX of all OPAM OCaml users (no more dead links) in 36 lines of code. I propose that it is merged, so that everybody can start benefiting from it, and then, somebody that is concerned by its apparent misconception could move the default OCamlPro's url to a per-repository metadata.

> Also I am very worried by relying on md5 for backup purposes... Currently we are using it in opam detect file changes so a collision will just cause a recompilation to not happen (which is kind of a good side-effect, if the upstream archive has been compromised anyway). In your proposal, a collision is quite bad...

The likelihood of a md5 collision becomes non-negligible above 2^64 files. And that's for text files, not well-formed OCaml archives. Currently, we have 2^13 files in the backup. Moreover, again, it is a fallback mechanism, when all the official links to the archive are dead.

Moreover, with the new secured repositories, OPAM is probably going to move to better checksums, and the code is trivial to adapt (if needed...).

hannesm:
A backup mechanism - wouldn't you simply want to use sth like `http_proxy` (and `https_proxy`) environment variable, which is respected by `curl`, `wget`, or whatever tool you use to transfer data via http (if not, please fix your download tool of choice)?  I don't think adding this to opam is of much use.  The proxy can use whatever storage/link it wants.

At this point in time, introducing more dependencies onto MD5 (checksums of tarballs) is bad imho, just use the name (full path).

In the future, at the horizon, the repository server, and the tarball providers, does not need to be trusted anymore anyways... which opens all the possibilities of caching and using insecure plain text protocols for transfer (http/ftp/you name it).

src/format/opamFile.ml
```diff
@@ -823,10 +826,13 @@ module ConfigSyntax = struct
                (kind,criterion)::List.remove_assoc kind t.solver_criteria }
   let with_solver t solver = { t with solver = Some solver }
 
+  let backup_url_ocamlpro = "http://opam.ocamlpro.com/backup"
```
AltGr: I would rather use this value when initialising the original config file (at `opam init` time, in `OpamClient`) than set it in the "empty" data structure. This way it can be seen and removed by hand later on. An initial default rather than a permanent, hidden default.
Also, the printer for lists assumes that an empty field of type list is the empty list, so this may cause issues -- e.g. you can't actually unset the field, or it reappears if opam reprints the file.

lefessan: Yes, the printer for list is buggy, it also prints a single-element list as a string, which is misleading for the user (i.e. `opam init` currently prints `backup_urls` as a string...).
I don't really know how to do what you say: it would involve reprinting the `config` file after adding a new field `backup_urls`, but then, to do that, I need to change the current global version of OPAM, since the config file does not have its own version, just the global version of OPAM...

dbuenzli: > I would rather use this value when initialising the original config file

I would rather not. Why can't this be on `opam.ocaml.org` which is community managed ?

OCamlPro-Henry: Le 01/03/2016 8:32, Daniel Bünzli a écrit :
>> @@ -823,10 +826,13 @@ module ConfigSyntax = struct
>>                 (kind,criterion)::List.remove_assoc kind 
>> t.solver_criteria }
>>    let with_solver t solver = { t with solver = Some solver }
>> 
>> +  let backup_url_ocamlpro = "http://opam.ocamlpro.com/backup"
> 
>> I would rather use this value when initialising the original config 
>> file
> 
> I would rather not. Why can't this be on `opam.ocaml.org` which is
> community managed ?

I have no strong opinion on where this backup should be hosted, this 
could probably be hosted on `opam.ocaml.org`, but what's wrong in 
OCamlPro offering this service and some sysadmin time ? Anyway, the 
default URL might be changed by the end-user.

AltGr: It's not a bug, it's syntactic sugar, which is documented at http://opam.ocaml.org/doc/Manual.html#Generalsyntax.
Empty and absent list fields can be distinguished when needed by using an option rather than a list as the high-level type of the field.

I am not sure this was already the case in `1.2.2`, but recent versions of opam ignore new fields (unless `--strict` was passed) to allow the evolution of the formats. Thus, just adding a field that is advisory doesn't require a new version. But it's true that per-file versions would be more convenient in an agile setting.

dbuenzli: > but what's wrong in OCamlPro offering this service and some sysadmin time ?

OCamlPro's website has been 404ing more than once with stuff hosted there (the last occurence being that [one](https://github.com/ocaml/opam/issues/2438)). We don't want to wait for an OCamlPro employee to show up to be able to fix these problems. I'm sure your sysadmin time will be gladly appreciated on opam.ocaml.org.

OCamlPro-Henry: On Tue, Mar 01, 2016 at 12:38:27AM -0800, Daniel Bünzli wrote:
> > @@ -823,10 +826,13 @@ module ConfigSyntax = struct
> >                 (kind,criterion)::List.remove_assoc kind t.solver_criteria }
> >    let with_solver t solver = { t with solver = Some solver }
> >  
> > +  let backup_url_ocamlpro = "http://opam.ocamlpro.com/backup"
> 
> > but what's wrong in OCamlPro offering this service and some sysadmin time ?
> 
> OCamlPro's website has been 404ing more than once with stuff hosted there (the last occurence being that [one](https://github.com/ocaml/opam/issues/2438)). We don't want to wait for an OCamlPro employee to show up to be able to fix these problems.

Sure, mistakes were made during a server migration and things were quickly fixed.

OCamlPro-Henry: That's said, I really don't care were this service is hosted.

lefessan: > OCamlPro's website has been 404ing more than once with stuff hosted there

Github has been down more frequently in the last years, and opam-repository is still hosted on it.

> (the last occurence being that #2438)

Indeed, a good example, it was fixed immediately.

> We don't want to wait for an OCamlPro employee to show up to be able to fix these problems.

Funny to read when the one person who fixes bugs in OPAM (Louis Gesbert) is actually an OCamlPro employee.

> I'm sure your sysadmin time will be gladly appreciated on opam.ocaml.org.
￼
It takes me more time to discuss on this forum everything I would like to do than implementing them and starting them on OCamlPro's servers. If your goal is to demotivate outside contributions to OPAM and its repository, you are quite efficient !

dbuenzli: > Funny to read when the one person who fixes bugs in OPAM (Louis Gesbert) is actually an OCamlPro employee.

You are mixing two things that are completely unrelated. Louis is certainly not the only person fixing bugs with the opam repository, I'm not even sure he is an opam repository maintainer. 

> It takes me more time to discuss on this forum everything I would like to do than implementing them and starting them on OCamlPro's servers.

This is not my problem. You may want to reflect on the way you interact with other people and the community or on the solutions you propose that always try to tie in your company infrastructure in some way.

> If your goal is to demotivate outside contributions to OPAM and its repository, you are quite efficient !

My goal is to have a sustainable environment. I don't see an environment partly in control of OCamlPro or any other company as being a sustainable environment. The environment should be in control of the community which is what opam.ocaml.org is for, as such I don't feel we should have URIs that point to OCamlPro's infrastructure in the opam source itself.

lefessan: > This is not my problem. You may want to reflect on the way you interact with other people and the community 

I have done more for this community than most of the people with who I am interacting here will do in their lifetime. You may want to reflect on that too.

> or on the solutions you propose that always try to tie in your company infrastructure in some way.

OPAM was developed on http://github.com/OCamlPro/opam/, the repository hosted on http://github.com/OCamlPro/opam-repository/, and the website hosted on http://opam.ocamlpro.com/. Everything was moved to http://ocaml.org/ That's what you call "always trying to tie in my company" ? The only thing I have fought for was to recover our access rights, since the "community" took all these repos without keeping our access rights to them...

> I don't see an environment partly in control of OCamlPro or any other company as being a sustainable environment

Without Janestreet and OCamlPro, there would be no ocaml.org and no OPAM. Where do you see a sustainable environment without them ? By the way, OCamlPro was created to help the OCaml community, not the other way around.

> The environment should be in control of the community which is what opam.ocaml.org is for, as such I don't feel we should have URIs that point to OCamlPro's infrastructure in the opam source itself.

So, OCamlPro should keep on creating new software and services, then move them to ocaml.org, so that people like you can believe that OCamlPro did not create them, and that they can disrespect us. Nice.

avsm: @dbuenzli @lefessan, please keep the conversation technical and focussed to the PR at hand please.

My comments:

- It is very useful to have a backup URL feature in OPAM, but I agree with @AltGr that this needs to either be per-remote and configurable, just as other remote-specific URLs are.  A global setting doesn't work well with the design of remotes, which can be in local infrastructure.
- Regarding opam.ocaml.org vs opam.ocamlpro.com, we must at a minimum remain consistent.  The current behaviour of `opam.ocaml.org` is that the archive service can be set to a local service very easily at init time, and nothing is baked into the tool beyond the initial default.  Baking in another domain that is non-ocaml.org just makes this much more difficult.  If possible, it would be very helpful to have instructions on how to set up the backup service, and then anyone can run this service (including --and we would be most grateful to -- OCamlPro).

In general, it is helpful to think of ocaml.org as just one of many "OPAM distributions" that combines OPAM and the ocaml/opam-repository and associated tools like opam2web into one service.  There are other emerging consumers in the open source world (such as the Coq community), industrial users whose codebases arent public, and I would hope commercially supported distributions who can all build services using the same tools. The design of PRs coming into OPAM should reflect that configurability and customisability and not bake in centralised choices where possible.

dbuenzli: > I have done more for this community than most of the people with who I am interacting here will do in their lifetime. You may want to reflect on that too.

Reflecting on this I don't see how this gives you more authority to discuss technical issues about opam.

> Without Janestreet and OCamlPro, there would be no ocaml.org and no OPAM. Where do you see a sustainable environment without them ?

This is precisely the problem at the moment. We should strive for an environment that is able to survive without them and bring more companies in to invest in a company neutral ecosystem. OCamlPro would be the first to benefit from it. 

> By the way, OCamlPro was created to help the OCaml community, not the other way around.

I don't care about intentions, I care about actions. 

> So, OCamlPro should keep on creating new software and services, then move them to ocaml.org, so that people like you can believe that OCamlPro did not create them, and that they can disrespect us. Nice.

If you believe OCamlPro is the only entity creating value in this system and take objecting to hardcode OCamlPro URIs in opam as disrespect I'm afraid there's not much left to be discussed at that point.

lefessan:
Some working code is here, the service is running. If somebody wants to improve on them, he is welcome. I am tired of arguing.. . or is it too much skiing ? ;-)

lefessan:
I removed the default URL linking to OCamlPro in the code, and moved it to CHANGES as documentation. Thus, the current PR does not change anything for users who do not explicitely set the backup-urls field in their configuration file, but still allows advanced users to set it.

avsm:
Is there an associated documentation patch with this feature? Reading the patch, I'm not sure if it's initialised by the user, or by the repo metadata.  Placing it in the repo metadata (i.e. opam.ocaml.org could decide where its backup mirrors are) seems useful.

AltGr:
In this state this seems good to merge, if you could rebase ?
Thanks !

AltGr:
@lefessan: the newer version of opam 2.0 [allows](https://github.com/ocaml/opam/pull/2685) sha256 and sha512 hashes; a function is provided that turns a hash into a relative path (e.g. `sha256/e3/e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855`), so that may be a good way to define a "standard" path to access an archive with a given hash ?

Of course, you could end up with duplicate archives through different hashing algorithms, but since what was specified in the original `url` file is going to be used, that shouldn't be too much of a problem.

hannesm:
@AltGr the currently supported hash algorithms (MD5, SHA256, SHA512) all differ in the output length, thus you won't be able to collide them :)

AltGr:
This is now superseded by the built-in `archive-mirrors:` field (available both per-repo, and globally in a given opam root). See [here](http://opam.ocaml.org/doc/2.0/Manual.html#configfield-archive-mirrors) for the file layout and how to generate a cache.

Thanks!

PR #2453, opened 2016-03-02
Set OPAM_COLUMNS to force the formatting line length

lefessan:
Useful to be able to parse the output of `opam lint`: setting a high-value will print each warning/error on a different line, by preventing the formatter from adding line breaks.

### DIFF
--- a/src/core/opamStd.ml
+++ b/src/core/opamStd.ml
@@ -504,7 +504,9 @@ module OpamSys = struct
   let tty_out = Unix.isatty Unix.stdout
 
   let default_columns =
-    try int_of_string (Sys.getenv "COLUMNS") with _ -> 80
+    try int_of_string (Env.get "COLUMNS") with
+    | Not_found
+    | Failure _ -> 80
 
   let get_terminal_columns () =
     try (* terminfo *)
--- a/src/core/opamStd.ml
+++ b/src/core/opamStd.ml
@@ -503,7 +503,8 @@ module OpamSys = struct
 
   let tty_out = Unix.isatty Unix.stdout
 
-  let default_columns = 80
+  let default_columns =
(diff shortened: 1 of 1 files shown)


dbuenzli:
> Useful to be able to parse the output of opam lint

Shouldn't rather `opam lint` have an option for machine readable output ?

lefessan:
@OCamlPro-Henry: From the code, I assumed `COLUMNS` was only used when a terminal output is detected, whereas `default_columns` is used when printing to a file.

@dbuenzli: yes, but in the meantime, this patch does the trick.

lefessan:
@OCamlPro-Henry Instead of creating a new env variable, I changed the code to use `COLUMNS`, even when printing to files.

src/core/opamStd.ml
```diff
@@ -503,7 +503,8 @@ module OpamSys = struct
 
   let tty_out = Unix.isatty Unix.stdout
 
-  let default_columns = 80
+  let default_columns =
+    try int_of_string (Sys.getenv "COLUMNS") with _ -> 80
```
AltGr: Please use `Env.get` and explicitely catch the exceptions, as in the code below (opam guidelines forbid `try with _`)

AltGr:
Thanks!

AltGr:
(note: Travis failure due do cudf md5 (#2455))

lefessan:
PR updated to follow opam guidelines

AltGr:
Thanks!

avsm:
Shouldn't this be mentioned in the main opam man page?

lefessan:
@avsm: not really, I removed the OPAM_COLUMNS variable, to re-use the already used COLUMNS variable, so the only change is that COLUMNS also works when printing to files, instead of just for ttys (I don't think that behavior was documented).

PR #2456, opened 2016-03-03
add support for displaying active switch in prompt

seliopou:
This pull request modifies the opam installation process put in place a shell function&mdash;namely `__opam_ps1`&mdash;that a user can use to display the currently active opam switch in his shell prompt.

Note that as is, the code does not actually modify the user's shell prompt. The user must explicitly modify a shell configuration file to utilize the function. To place the current opam switch at the beginning of the prompt, the user would defined `PS1` as follows:

```
PS1="$(__opam_ps1 "%s")\u@\h:\w\$"
```
This would result in the following prompt:

```
(system)user@host:~$
```

... where `system` indicates the default system switch is active. If another switch is selected, that part of the prompt would change appropriately.

### DIFF
--- a/shell/opam_prompt.sh
+++ b/shell/opam_prompt.sh
@@ -19,7 +19,7 @@ __opam_ps1()
         ;;
     esac
 
-    local switch_name="$(opam switch show 2>/dev/null)"
+    local switch_name="$(opam switch show --safe 2>/dev/null)"
     if [ -z "$switch_name" ]; then
         return $exit
     fi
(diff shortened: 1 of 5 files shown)


shell/opam_prompt.sh
```diff
@@ -0,0 +1,28 @@
...
+    local printf_format='(%s)'
+
+    case "$#" in
+        0|1)    printf_format="${1:-$printf_format}"
+        ;;
+        *)  return $exit
+        ;;
+    esac
+
+    local switch_name="$(opam switch show 2>/dev/null)"
```
AltGr: You should add `--safe`: this guarantees the opam process won't take any locks or change anything.

AltGr:
Great, thanks!

seliopou:
`--safe` flag added, as suggested.

AltGr:
Thanks!

Issue #2006, opened 2015-02-16
is `curl --insecure` still needed ?


AltGr:
This was added because of compatibility issues on many common installations, in which curl didn't have access to a reasonable list of root certificates (in particular for github). I am wondering if this is still an issue now ?

We still double-check md5s of course.

avsm:
The other point of note is that Cohttp is now in good enough shape to download the vast majority of URLs via the safe TLS stack.  I'm testing this out specifically in the https://github.com/avsm/opam-mirror utility

AltGr:
Not relying on Curl anymore would indeed be quite cool.

avsm:
Frustratingly, Curl now fails on MacOS X on some sites:

```
curl "--write-out" "%{http_code}\\n" "--insecure" "--retry" "3" "--retry-delay" "2" "-OL" "https://ocaml.janestreet.com/ocaml-core/112.17/files/sexplib-112.17.00.tar.gz"
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0000
curl: (56) SSLRead() return error -9841
```

This appears to be due to a `curl` configuration option: http://stackoverflow.com/questions/26461966/osx-10-10-curl-post-to-https-url-gives-sslread-error

@yminsky, did the JS site recently disable SSLv3 to trigger this?  I can workaround it by removing the `https` from the distfiles for now, and hope the md5 checksum is strong enough.

dsheets:
`--insecure` already meant that we had no authentication against CA bundle...

avsm:
The error above was due to protocol negotiation and not CA checking, I think

AltGr:
Travis reports such failures to sourceforge, too (on OSX)
see for example 
https://travis-ci.org/ocaml/opam/jobs/51460132

camlspotter:
Is there any workaround for this issue?

AltGr:
the latter ? Hm, if it doesn't find `curl` in your `PATH`, OPAM will use `wget` instead.
If you have a working `curl` command somewhere, set `OPAMCURL` to point at it (there is no similar way to specify a `wget` command at the moment though, curl-specific command-line args will be used after `$OPAMCURL`).

avsm:
[...]

Issue #55, opened 2012-07-16
https curl errors on debian/wheezy


avsm:
Apologies if this isn't an OPAM-specific, but I'm finding that the wget/curl invocations are breaking on Core due to SSL validation errors.  Is anyone else seeing this on Debian? It works fine for me on MacOS X.

```
https://ocaml.janestreet.com/ocaml-core/108.00.02/individual/type_conv-108.00.02.tar.gz
Resolving ocaml.janestreet.com (ocaml.janestreet.com)... 208.70.147.209
Connecting to ocaml.janestreet.com (ocaml.janestreet.com)|208.70.147.209|:443... connected.
ERROR: The certificate of `ocaml.janestreet.com' is not trusted.
ERROR: The certificate of `ocaml.janestreet.com' hasn't got a known issuer.
[3114] ERROR: Command "wget https://ocaml.janestreet.com/ocaml-core/108.00.02/individual/type_conv-108.00.02.tar.gz" failed
 (see /home/avsm/.opam/log/command1869.{info,err,out})
```

PR #2460, opened 2016-03-04
remove insecure / no-check-certificate flags (see mail on opam-devel, #55 #2006)

hannesm:


hannesm:
(earlier discussion in #55 #2006) -- also, I do use `download-command: [ "fetch" url "-o" out ]` in my `.opam/config` (which authenticates server certificates), so far without problems)

hannesm:
To add some more data, figuring out which https hosts are used in the main opam-repository:
```
2401 github.com
 924 ocaml.janestreet.com
 209 forge.ocamlcore.org
 159 bitbucket.org
  31 gforge.inria.fr
  10 raw.github.com
   6 codeload.github.com
   5 www.lri.fr
   4 people.cs.umass.edu
   4 downloads.sf.net
   3 coq.inria.fr
   1 zoggy.github.io
   1 pkgs.fedoraproject.org
   1 gmplib.org
```

and it turns out, pkgs.fedoraproject.org does not send its full certificate chain (used in zenon, others (omake & findlib) use http on pkgs.fedoraproject.org).  thus, the numbers arein favour of the change (and zenon and/or pkgs.fedoraproject.org should be fixed!)

avsm:
the fedoraproject url is a fallback for missing upstream distfiles; we could host those elsewhere on ocaml.org with a proper cert

hannesm:
sounds good to me!

AltGr:
Thanks!

Issue #2458, opened 2016-03-04
opam lint should be more careful about self-references


dbuenzli:
Since I didn't get [any answer](http://lists.ocaml.org/pipermail/opam-devel/2016-February/001342.html) I'll repeat the question here.

I'm trying to refer to `%{mypkg:etc}%` in `mypkg`'s opam `build:` section. It works perfectly but `opam lint` complains:  

```
warning 41: Some packages are mentionned in package scripts of features, but there is no dependency or depopt toward them
```

So either there's another way or `opam lint` should not complain about this.

AltGr:
You can write `%{_:etc}%`, but I don't think that was in 1.2. Anyway, your point remains and this is a bug in `opam lint`. Thanks for reporting!

PR #2464, opened 2016-03-08
Fix warning 41

lefessan:
Witthout the fix, the comparison prints the warning only for packages which are using a reference to themselves (and obviously, their own name is not in the depends). Currently, it is only "bat"...


dbuenzli:
Related to https://github.com/ocaml/opam/issues/2458

AltGr:
Thanks! The condition was indeed broken
Duplicate of #2458 
This was fixed on branch `next` by c1fb3a3 but not backported yet.

PR #2479, opened 2016-03-14
Document test dependencies

Leonidas-from-XIV:
I was looking at the "packaging" page on OPAM and wasn't sure there was a way to specify dependencies that are only valid for testing. `oasis2opam` did generate a `test` key, so it is possible. Hence a small documentation patch to help other poor souls.

### DIFF
--- a/doc/pages/Packaging.md
+++ b/doc/pages/Packaging.md
@@ -98,6 +98,10 @@ build: [
   ["./configure" "--prefix=%{prefix}%"]
   [make]
 ]
+build-test: [
+  ["./configure" "--enable-tests"]
+  [make "test"]
+]
 install: [make "install"]
 remove: ["ocamlfind" "remove" "project"]
 depends: "ocamlfind"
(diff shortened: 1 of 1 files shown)


AltGr:
The "manual" page contains a more detailed reference on the available fields: http://opam.ocaml.org/doc/Manual.html#opamfield-depends

doc/pages/Packaging.md
```diff
@@ -98,6 +98,10 @@ build: [
   ["./configure" "--prefix=%{prefix}%"]
   [make]
 ]
+build-test: [
+  ["./configure" "--enable-tests"]
+  [make "test"]
+]
```
AltGr: It's worth documenting the `build-test:` field below, but I would rather keep the example as simple as possible, and more importantly close to the template provided by opam, so I'd rather keep this out.

AltGr:
Thanks a lot for helping in improving the doc!

AltGr:
The feature has been extended and documented since this PR was made. Thanks!

PR #2480, opened 2016-03-15
Add a lint check on unclosed variables

AltGr:


AltGr:
Merged as part of the `next` branch

PR #2482, opened 2016-03-16
src_ext: update checksums for cudf 0.7 and dose 3.3

sebastienaubry:


AltGr:
Ah! Just a few minutes too late; but thanks a lot! :)

PR #2496, opened 2016-04-01
Fix off-by-one in state backup timestamp

simonjbeaumont:
The `Unix.tm_mon` is a zero-based month and so should be incremented when using as a human-readable month.

```
$ opam install ...
...
The former state can be restored with opam switch import "/home/simonbe/.opam/4.02.2/backup/state-20160301102642.export"
$ date
Fri Apr  1 10:49:39 UTC 2016
```

The month should be 04, not 03. No, this is not an April fools commit :)

Signed-off-by: Si Beaumont <simon.beaumont@citrix.com>


AltGr:
Indeed, using a human-friendly backup name should make it easier to sort them out. Thanks!

PR #2498, opened 2016-04-02
Update basic guide to recommend ocaml 4.02.3 instead of 4.02.1

Khady:
The current basic usage guide uses ocaml 4.02.1. It is confusing for newcomers as you can see [here](http://irclog.whitequark.org/ocaml/2016-04-02#1459592018-1459592392;).


samoht:
Thanks

PR #2508, opened 2016-04-10
Fix zsh completion

yasu100:


yasu100:
Travis-ci : "OCaml version doesn't match: travis script needs fixing"

PR #2515, opened 2016-04-14
Next

AltGr:
This merges the `next` dev branch, with internal version `2.0~alpha`, into the main `master` dev branch, as the former is reaching maturity and an automatically rewritten repository is in place (`compilers_to_packages.ml` applied at https://opam.ocaml.org/2.0~dev).

At this point `next` is probably more stable than `master` already.

**Compatibility note**: `next` doesn't compile on OCaml earlier than 4.1 anymore, as it fails to handle some subtyping on polymorphic variants (used to hierarchise the file locks). Can we discard¹ ? Travis shows an issue with `4.1` too, but this is only related to the lib-ext scripts and should be easy to fix.

¹ These absolutely doesn't mean opam wouldn't support earlier OCaml versions anymore, only that you won't be able to _compile_ opam without at least 4.1.


dbuenzli:
> doesn't compile on OCaml earlier than 4.1 anymore

That sounds reasonable. FWIW topkg will be 4.01 only aswell and others have been moving there (e.g. `ctypes`). It will be nice to be able to drop 3.12 in cmdliner aswell. Could you please point to a proper commit you feel confident can be used for testing, I'd rather not be on a too moving target.

samoht:
Yes, supporting only 4.01 seems fine to me.

Drup:
@AltGr I suppose it's normal that next opam doesn't convert my installed switches to the brand new scheme ?

AltGr:
@Drup, no, it's not. What version were you previously on, and can you check the opam-version in ~/.opam/config ?

Drup:
@AltGr I was on master, which means 1.3, and here is `.opam/config`:
```
% cat .opam/config
opam-version: "1.3"
repositories: ["codoc" "default" "mirage-dev" "termite"]
installed-switches: [
  "4.03.0"
  "last"
  "4.03.0+local-git-eliom"
  "4.03.0+trunk"
  "4.03.0+beta2"
  "eliom"
  "latest"
]
switch: "4.03.0"
jobs: 4
download-jobs: 4
```

AltGr:
I see, `opam-version: "1.3"` is wrong and was put there by a (bug in an) earlier dev version: current dev versions use `1.3~devX`, but this is _before_ `1.3` according to version ordering, so no upgrade is triggered.
You can probably change `opam-version` manually to `1.3~dev` to have opam trigger the upgrade.

Drup:
I tried the upgrade. It was a bit acrobatic since half of my compilers were old and half of them were new, but it worked after manually adding the new compilers to the translated `.opam/config` file.

PR #2541, opened 2016-05-03
add back pef/opam export functions

abate:
This functionality is needed by ows.irill.org


AltGr:
Thanks! Sorry it got away during the rewrite and thanks for going through the hassle of adding it again.

The [travis script](https://github.com/ocaml/opam/blob/0e119a6291eeb74c2fd46ffcd50f9b8a25c58622/.travis-ci.sh#L67) and [src_ext Makefile](https://github.com/ocaml/opam/blob/0e119a6291eeb74c2fd46ffcd50f9b8a25c58622/src_ext/Makefile#L20-L21) would probably need an update of the dose dependency. I noticed a regression on `3.4.2`, though, related to a patch I submitted upstream (but I could not, at the time, point to the failure case). It fails on [opam-rt](https://github.com/ocaml/opam-rt):
>     - Fatal error in module algo/diagnostic.ml: 
>     -  the package encoding global constraints can't be missing

To reproduce: `make libinstall` from opam, clone `opam-rt` on branch `next`, and run:
```
% make
% ./opam-rt test /tmp/opam-rt-test-reinst reinstall
```

abate:
```
- Fatal error in module algo/diagnostic.ml: 
-  the package encoding global constraints can't be missing
```
I think somewhere in your code you are initializing the constraint problem with global_constraints=false and then calling the solver with global_constraints=true . There is a patch to make a fatal error into a warning, but this is actually a fatal error. This should never happen. I'll dig a bit more this morning to check if this is the case.

I'll also update the bootstrap dependency of opam to the latest dose3 release and send you a pull request.

AltGr:
Opam doesn't seem to reference a `global_constraints` parameter, and I don't remember seeing it; where in dose's interface does it appear ?

abate:
I'm looking into this problem. You are right. global_constraints is not part of the API making it difficult to set it to false. I'm going to fix this in the next release. I've also slightly changed the behavior of Depsolver.check_request_using to return a universe without the dummy package. This should resolve this issue and making the call to the solver more transparent.

abate:
@AltGr I've fixed both problems. I think this is good to go. maybe you can update also the ocp-build scripts to reflect these changes.

AltGr:
It seems you have duplicate commits on your branch, you should run `git rebase origin/master`. Thanks!

hannesm:
@abate @AltGr this PR is against next, and afaict github won't let you retarget a PR to another branch... thus you'll need to close, rebase and open a new PR against master...

AltGr:
Ah indeed, thans @hannesm for pointing that out.

Issue #1993, opened 2015-02-12
Allow variables in version constraints


whitequark:
I want this for cross-compiling, e.g. bin_prot-android would have `depends: [bin_prot {= version}]`, as it pulls the syntax parts from the host package.

AltGr:
That would be convenient in that case, indeed, because you only use a variable that's defined statically from the given opam file ; but it's a whole different story if you use, in general, variables depending on the global state: these would need to be resolved _before_ the solver is called, which could quickly end up with a huge noodle mess in the repo.

At the moment, solver scope ("formulas") and command scope ("filters") are well separated and I am quite sure mixing them would be a bad idea.

There is already, though, the notable exception of the `available:` field, which is evaluated first of all -- but this one doesn't allow package variables (like `version`), only global variables.

AltGr:
As a related note, however, being able to select dependencies based on, say, the OS, would be quite useful ; that would be a filter with only global variables, like `available:`.

It's a different problem, but also related to the dependency formula selectors. Doing it consistently would mean two-stage resolution: 1- pre-process evaluating OPAM variables -2- the result is a package formula that can be fed to the solver. That'd need a very careful syntax update if we want to keep an understandable package format.

AltGr:
OS-dependent dependencies are recently coming up again, and this may be useful.

Currently, package definitions resolve variables (in so-called "filters") in two different scopes:
- the concrete scope, available on package actions (build, installation, removal), with access to all variables on the set of installed packages.
- the global scope, which is limited to non-package variables, and can be used _before_ any scenario is computed. Well-defined self-references like `%{_:name}%` and `%{_:version}%` could be added without problem.

The latter is currently only used to compute the `available:` field, which filters the packages available before calling the solver. There would be no technical problems in allowing filters, with the global scope, on dependencies.

Extending the syntax in a reasonable way seems to be the biggest challenge here. We already have:
```
depends: [ "foo" { build & test & >= "1.2" & < "2.0" } ]
```
The `build` and `test` here are already an extension, and they need to be leading. We could actually generalise them (defining them as variables in this scope) and put a real "filter" at this point. But mixing in the same sequence "filter" terms and "version-formula" terms could be a bit misleading, if at all possible syntactically.

[...]

PR #2547, opened 2016-05-10
Allow version references in the 'depends' and 'depopts' fields

AltGr:
Define 'name' and 'version' in switch scope in opam files, and allow
references.

This makes it possible e.g. for a package to depend on another one at the
same version.

    depends: "bar" { = version }


whitequark:
Wonderful, thank you so much!

AltGr:
Closes #1993

lefessan:
Thanks !

PR #2550, opened 2016-05-10
next: add local subcommand

seliopou:
This pull request adds the `local` subcommand to OPAM for managing directory-local switches, often referred to as a "sandbox". When in such a directory, OPAM will use the switch associated with the directory when executing all commands. For example if the user's current active switch is `switch-1`, running `opam install <pgk>` would normally install `<pkg>` into `switch-1`. However, if the user is in a sandbox directory associated with `switch-2`, then running `opam install <pkg>` will install `<pkg>` in `switch-2`. Similarly, running `opam config env` within the sandbox will display the environment for `switch-2`, rather than `switch-1`.

The only `local` subcommand included in this command is `create`. This allows users to setup a directory as a sandbox, associating with it a switch. The `create` command takes the same arguments as `opam switch create` for configuring the initial state of the switch.

In the future, additional commands may be added below `opam local` in order to assist in the management of sandboxes.

This pull request targets the `next` branch of OPAM development.

As mentioned to @AltGr via email, this pull request intentionally does not include command-line documentation.

### DIFF
--- a/src/client/opamLocalConfig.ml
+++ b/src/client/opamLocalConfig.ml
@@ -62,7 +62,7 @@ let update ?noop:_ = setk (fun cfg () -> r := cfg) !r
 
 let initk k =
   let pred dir = OpamFilename.(exists Op.(dir // ".opamlocal")) in
-  let sandbox_dir = OpamFile.locate_ancestor pred (OpamFilename.cwd ()) in
+  let sandbox_dir = OpamFilename.find_ancestor_dir pred (OpamFilename.cwd ()) in
   let local_file =
     match sandbox_dir with
     | None -> None
--- a/src/client/opamLocalConfig.ml
+++ b/src/client/opamLocalConfig.ml
@@ -1,0 +1,77 @@
+(*----------------------------------------------------------------------------
+    Copyright (c) 2016 Inhabited Type LLC <spiros@inhabitedtype.com
+
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions
+    are met:
+
+    1. Redistributions of source code must retain the above copyright
+       notice, this list of conditions and the following disclaimer.
+
+    2. Redistributions in binary form must reproduce the above copyright
+       notice, this list of conditions and the following disclaimer in the
+       documentation and/or other materials provided with the distribution.
+
+    3. Neither the name of the author nor the names of his contributors
+       may be used to endorse or promote products derived from this software
+       without specific prior written permission.
+
+    THIS SOFTWARE IS PROVIDED BY THE CONTRIBUTORS ``AS IS'' AND ANY EXPRESS
+    OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+    WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+    DISCLAIMED.  IN NO EVENT SHALL THE AUTHORS OR CONTRIBUTORS BE LIABLE FOR
+    ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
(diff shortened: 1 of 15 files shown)


AltGr:
`opam-rt` fails to build with:
```
+ ocamlfind ocamlopt -linkpkg -package cohttp.lwt -package opam-lib.client src/opamRTcommon.cmx src/opamRTinit.cmx src/opamRT.cmx src/opamRTmain.cmx -o src/opamRTmain.native
File "_none_", line 1:
Error: No implementations provided for the following modules:
         Option referenced from /home/travis/.opam/system/lib/opam-lib/opam-format.cmxa(OpamFile)
```

Did you use extlib modules by mistake ? You should use `OpamStd.Option` rather than just `Option` (this _should_ have failed to compile from the start, but, well, build systems...).

seliopou:
Pushed fixes for both pull requests.

src/client/opamLocalConfig.ml
```diff
@@ -0,0 +1,78 @@
...
+let initk k =
+  let pred dir = OpamFilename.(exists Op.(dir // ".opamlocal")) in
+  let sandbox_dir = OpamFile.locate_ancestor pred (OpamFilename.cwd ()) in
+  let local_file =
+    match sandbox_dir with
+    | None -> None
+    | Some dir ->
+      let local_file = OpamFilename.Op.(dir // ".opamlocal") in
+      try Some OpamFile.(Local.read (make local_file))
+      with _ -> None
```
AltGr: I know I'm being picky, but please respect https://github.com/ocaml/opam/blob/master/src/core/opamStd.mli#L280 :
`with e -> OpamStd.Exn.fatal e; None`

src/format/opamFile.ml
```diff
@@ -2559,3 +2609,16 @@ module Comp = struct
   include CompSyntax
   include SyntaxFile(CompSyntax)
 end
+
+let locate_ancestor pred dir =
```
AltGr: This could probably be in `OpamFilename`.

AltGr:
Thanks a lot, this is great work ! Also, sorry for taking so long to give a review. Here are just some minor remarks:

- `OpamFile.Local`: you are relying on `OpamFile.Dot_config`, which was a raw stringmap, and has now evolved to include some fields, plus that stringmap inside a `variables:` section. I don't think this is what you want, and you should rather use an adhoc definition based on the `OpamFile.SyntaxFile` functor. You can see the new `OpamFile.Repos_config` [code](https://github.com/ocaml/opam/blob/580e4880a4742a6a3b606e4c7ceb0992b185028b/src/format/opamFile.ml#L982-L1012) for a clean example of an extensible, one-field file. In your case, the fields definition would be something like:

    ```ocaml
    let fields = [
        "switch", Pp.ppacc with_switch switch
          (Pp.V.string -| Pp.of_module "switch-name" (module OpamSwitch))
    ]
    ```
  Also, I know the current naming of modules within `OpamFile` (which really should be split at some point) isn't quite good, but maybe something like `Dot_local` would be slightly more explicit.
- You're submitting your contribution as BSD3, therefore allowing us to relicense in particular as LGPL to fit the general project license (as long as we keep the software notice). Is that correct ? We're still waiting for the resolution of the "license bug", so although that probably won't be needed, I can't yet exclude the possibility that more paperwork is needed.
- See the small inline comments.

Overall, I am quite impressed by how well you grasped the code and inner conventions, with little or no documentation ☺. This is very helpful, thanks for contributing!

rgrinberg:
What's the status of this? Wondering if a workflow like this will be available in opam 2.0

AltGr:
Local switches are built-in in opam 2.0 since alpha2. Simply use a directory name as switch name (e.g. `opam switch create .`), and the switch folder will be put as `_opam/` below this folder. Then, if no other selection makes precedence and you are running from the same folder or a descendant, that "local switch" will get chosen automatically.

I guess this can be closed. Although this wasn't merged as is, it was useful in designing the new feature, thanks @seliopou.

Issue #2506, opened 2016-04-07
Installing package with the -t flag installs test dependencies of all dependent packages


seliopou:
To reproduce, run the following command:

``` bash
opam install webmachine --deps-only -t
```

This will produce the following output (given that all of webmachine's non-test dependencies are installed):

```
The following actions will be performed:
  - install   ocaml-data-notation 0.0.11   
  - install   ounit               2.0.0    
  - install   astring             0.8.1    
  - install   ocamlify            0.0.1    
  - install   qcheck              0.4      
  - install   ocamlmod            0.0.8    
  - recompile re                  1.5.0    
  - recompile ppx_deriving        3.3      
  - recompile dispatch            0.3.0    
  - recompile cstruct             1.9.0    
  - recompile stringext           1.4.1    
  - install   oasis               0.4.5    
  - recompile ppx_type_conv       113.33.00
  - install   alcotest            0.4.8    
  - recompile ppx_sexp_conv       113.33.00
  - recompile uri                 1.9.2    
  - recompile ipaddr              2.7.0    
  - recompile conduit             0.11.0   
  - recompile cohttp              0.20.0   
===== 8 to install | 11 to recompile =====
Do you want to continue ? [Y/n] 
```

The only test dependency that webmachine has is oUnit. However it has dependencies that use alcotest and qtest as well.

I expect this command to only install oUnit, but it is installing the test dependencies of webmachine's dependencies as well.

seliopou:
Possibly related to #2494.

samoht:
See #2190

AltGr:
Indeed, that's what `-t` does. More flexibility for tests would be nice...
- separating build and run of tests in opam files (#2494 : the above will not only fetch all test-deps, it will also run all tests of dependent packages)
- asking for specific test-depends
- listing test-dependencies (if there were an appropriate flag, you could do `opam install $(opam list -s --test --required-by webmachine)` as a workaround)

I am not sure about the distinctions between building, installing (?) an running the tests, and how we should modify the corresponding `build-test:` field. There is also, at the moment, a parity with the `build-doc:` field. But there is no blocker for separating at least dependency handling of test-dependencies and test-building.

PR #2551, opened 2016-05-10
next: add package subcommand

seliopou:
This pull request adds the `package` subcommand to OPAM, which exposes a command-line interface to  lifecycle events already described by OPAM files. Specifically, this pull requests includes three subcommands:
- `opam package build` runs the `build` script of the OPAM file in the current project;
- `opam package build-test` runs the `build-test` script of the OPAM file in the current project; and
- `opam package build-doc` runs the `build-doc` script of the OPAM file in the current project.

These commands will search an OPAM file in the typical locations, relative the current directory, e.g., `opam`, `opam/opam`, etc. In addition, the commands will also search up the directory tree for an `opam` file and if found, use that to determine which commands to run.

The `opam package` subcommands can be used in any context, but were developed alongside directory sandbox support included in #2550. In conjuction with #2250, users can develop OPAM packages in local sandbox, using the switch associated with that sandbox while leaving the globally-configured switch unchanged, and out of the picture.

This pull request targets the `next` branch of OPAM development.

As mentioned to @AltGr via email, this pull request intentionally does not include command-line documentation.

### DIFF
--- a/src/client/opamPackageCommand.ml
+++ b/src/client/opamPackageCommand.ml
@@ -40,7 +40,7 @@ let build t ?build_test ?build_doc dir =
   let name = OpamPackage.name default_nv in
   let pred dir = None <> OpamPinned.find_opam_file_in_source name dir in
   let dir = 
-    match OpamFile.locate_ancestor pred dir with
+    match OpamFilename.find_ancestor_dir pred dir with
     | Some dir -> dir
     | None     ->
       OpamConsole.error "Not in a local sandbox";
--- a/src/client/opamPackageCommand.ml
+++ b/src/client/opamPackageCommand.ml
@@ -1,0 +1,57 @@
+(*----------------------------------------------------------------------------
+    Copyright (c) 2016 Inhabited Type LLC <spiros@inhabitedtype.com>
+
+    All rights reserved.
+
+    Redistribution and use in source and binary forms, with or without
+    modification, are permitted provided that the following conditions
+    are met:
+
+    1. Redistributions of source code must retain the above copyright
+       notice, this list of conditions and the following disclaimer.
+
+    2. Redistributions in binary form must reproduce the above copyright
+       notice, this list of conditions and the following disclaimer in the
+       documentation and/or other materials provided with the distribution.
+
+    3. Neither the name of the author nor the names of his contributors
+       may be used to endorse or promote products derived from this software
+       without specific prior written permission.
+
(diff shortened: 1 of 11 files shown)


src/client/opamPackageCommand.ml
```diff
@@ -0,0 +1,57 @@
...
+  let dir = 
+    match OpamFilename.find_ancestor_dir pred dir with
+    | Some dir -> dir
+    | None     ->
+      OpamConsole.error "Not in a local sandbox";
+      OpamStd.Sys.exit 1
+  in
+  (* XXX(seliopou): This whole dance is necessary because
+   * OpamAction.build_package reads global state, rather than explicitly taking
+   * the build_test and build_doc arguments *)
```
AltGr: Indeed. It would be best to add them as arguments troughout. Meanwhile, it may be more consistent to remove the argument here and relying on the options being set in `OpamMain`.

Note that since we would like to be able to set these options for specific packages (#2506), this will need to be changed anyway at that point.

src/state/opamAction.ml
```diff
@@ -512,9 +512,19 @@ let remove_package t ?keep_build ?silent nv =
 (* Compiles a package.
    Assumes the package has already been downloaded to [source].
 *)
```
AltGr: I would have written one core function and two exported functions for the two modes, rather than a mode argument; but it's just a a matter of taste, this is fine.

AltGr:
Thanks a lot! This is great and many are looking forward to this feature.

I think you could go further though: I would expect `opam package build foo` to resolve the dependencies of the package before building it, resulting to approximately the same results as:
```
$ opam pin foo . --no-act
$ opam install --deps-only foo
$ opam config exec -- <build commands for foo>
$ opam unpin foo --no-act
```
For that you would need to rely on `OpamClient.install`.

Some more design questions:
- You restrict the use of `opam package install` to local-switches. While the two definitely fit together, it may be useful even for centralised switches, do you think we could relax that constraint ?
- Have you considered running the `install` commands too ? There may be some difficulties to solve (upgrading/recompiling ?), but it would be nice to be able to test them too.

AltGr:
Ah, also, this still seems to target the `next` branch.

samoht:
Tentative initial doc. Feedback welcome: https://gist.github.com/samoht/2188fbd01b1d40a3e2063f3a270bf8fd

AltGr:
Superseded by `opam build` and the newer `opam install <dir>` interface. Thanks!

PR #2553, opened 2016-05-11
Update src_ext dependencies

abate:
Add cppo (to compile extlib and dose3)

extlib: 1.5.3 -> 1.7.0
dose3: 3.3 -> 4.3
ocamlgraph: 1.8.5 -> 1.8.6


abate:
@AltGr, this is the new PR as you asked. With dose3.4.3 I see a couple of fails associated to opam-rt that I leave to you , and the problem with BYTES that I think is just a minor problem with extlib, and I'm going to fix it today hopefully. Note that the hack in the travis script. I should be able to just install ocaml-findlib, but for some reason the package is not installable. Using ocamlfind from opam is a workaround.

AltGr:
Thanks!

abate:
@AltGr  I've fixed both problems. I think this is good to go. maybe you can update also the ocp-build scripts to reflect these changes.

AltGr:
Ah, sorry, my changes to the travis scripts (upgrade of lwt for compat with 4.03) conflict with this.
The remaining issue seems otherwise to just be https://github.com/ocaml/opam/pull/2541#issuecomment-218420067

abate:
I've fixed the problem with global_constraints mentioned in #2541. I'll add a small temporary patch to be removed with the next dose release.

AltGr:
Perfect, thanks!

PR #2554, opened 2016-05-13
Fix bug in ~/.opam format upgrade

AltGr:
that caused installed package definitions (including compiler defs
generated in a previous step) to be lost.

@Drup: sorry!


Drup:
Don't apologize, that's precisely why I do early testing! :D

PR #2555, opened 2016-05-13
Define OPAM_SWITCH_PREFIX when running package commands

AltGr:


AltGr:
Do we actually want this in the global context as well, as part of `$(opam config env)` ? It would then provide a replacement to calling `$(opam config var prefix --safe)`

lefessan:
Yes, I think it would be really nice. Currently, a ./configure has no way to know whether the current ocamlc compiler is managed by OPAM or not (and the environment). Checking that variable would give it a pretty good clue. So, definitively, it should be in 'opam config env' too.

By the way, what is the '--safe' for ?

AltGr:
`--safe` guarantees that opam won't take any lock, ask any question or block in any way. It may be useful e.g. if you upgraded your opam binary but not the format of ~/.opam yet, for example.

AltGr:
Let's already merge this part

PR #2560, opened 2016-05-14
Update the sub-lib packages

AltGr:
relying on more advanced ocp-build features, and the new opam `{=
version}` capability.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -73,3 +73,4 @@ doc/man-html
 doc/tutorials/opam.wiki
 doc/dev-manual/*.out
 doc/man
+src/x_build_libs.ocp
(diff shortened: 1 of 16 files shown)


AltGr:
@lefessan, feedback on my use of ocp-build is welcome.
In particular, I found no way to use a possibly undefined variable or set a default value; the solution I found was to define the default value in `src/opam.ocp`, and to override it through a `src/x_build_libs.ocp` file when necessary. Note the prefix `x_`, which makes sure the file is loaded after `src/opam.ocp`, as ocp-build thankfully appears to respect alphabetical order.

A better/more robust solution would be welcome :)

Issue #2317, opened 2015-09-28
behavior is different when invoked as "opam" vs "/usr/bin/opam"


whitequark:
To reproduce:

```
$ opam --version
1.2.2
$ strace -eexecve opam query
execve("/usr/bin/opam", ["opam", "query"], [/* 70 vars */]) = 0
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=10650, si_uid=1000, si_status=0, si_utime=0, si_stime=0} ---
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=10652, si_uid=1000, si_status=0, si_utime=0, si_stime=0} ---
execve("/home/whitequark/.opam/4.02.3/bin/opam-query", ["opam-query"], [/* 71 vars */]) = 0
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=10653, si_uid=1000, si_status=0, si_utime=0, si_stime=0} ---
+++ exited with 0 +++
$ strace -eexecve /usr/bin/opam query
execve("/usr/bin/opam", ["/usr/bin/opam", "query"], [/* 70 vars */]) = 0
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=10664, si_uid=1000, si_status=0, si_utime=0, si_stime=0} ---
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=10666, si_uid=1000, si_status=127, si_utime=0, si_stime=0} ---
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=10667, si_uid=1000, si_status=0, si_utime=0, si_stime=0} ---
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=10668, si_uid=1000, si_status=0, si_utime=0, si_stime=0} ---
--- SIGCHLD {si_signo=SIGCHLD, si_code=CLD_EXITED, si_pid=10669, si_uid=1000, si_status=0, si_utime=0, si_stime=0} ---
opam: unknown command `query'.
Usage: opam COMMAND ...
Try `opam --help' for more information.
+++ exited with 1 +++
```

I do not believe there is any valid reason for this behavior.

whitequark:
Any progress on getting this into a release?

whitequark:
[...]

PR #2565, opened 2016-05-24
For external plugin search, always do a full PATH search

avsm:
This prevents different behaviour if the command is executed with
`opam` vs `/usr/bin/opam`.

This is a backport to the 1.2 branch, as the bug is fixed in
OPAM trunk already.  See #2317.

### DIFF
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -1983,7 +1983,7 @@ let check_and_run_external_commands () =
       && List.for_all (fun (_,info) -> Term.name info <> name) commands
     then
     (* No such command, check if there is a matching plugin *)
-    let command = opam ^ "-" ^ name in
+    let command = (Filename.basename opam) ^ "-" ^ name in
     let t = OpamState.load_env_state "plugins" in
     let env = OpamState.get_full_env ~force_path:false t in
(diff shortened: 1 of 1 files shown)


src/client/opamArg.ml
```diff
@@ -1983,7 +1983,7 @@ let check_and_run_external_commands () =
       && List.for_all (fun (_,info) -> Term.name info <> name) commands
     then
     (* No such command, check if there is a matching plugin *)
-    let command = opam ^ "-" ^ name in
+    let command = (Filename.basename opam) ^ "-" ^ name in
```
dbuenzli: Shouldn't this simply be `"opam-" ^ name` according to docs the plugins tool have to be called `opam-bla`

avsm: That's effectively what it is, unless the `opam` command itself is named something else. I'm happy to change this to `"opam-" ^ name` to make it more specific. @AltGr?

dbuenzli: > That's effectively what it is, unless the opam command itself is named something else.

Precisely it seems to me you don't want the plugins to fail if `opam` happens to be named differently.

AltGr: Yes, this has been changed to `"opam"^` on 2.0

Issue #1215, opened 2014-03-04
`opam who-owns-this-path` command


maverickwoo:
Feature request:

Given a path under ~/.opam, I want to figure out which (package, version) provides that path by `opam provides that-path`. This is essentially the equivalent of the `provides` command in MacPorts.

This is particularly useful for detecting orphaned files after a package upgrade.

AltGr:
Sub-feature of #502

PR #2568, opened 2016-06-01
Track files installed by packages, allow query and use for removal

### LINKED ISSUE #502: Opam should compute the exact list of installed files
It would be nice if opam could uninstall things by logging what got installed in the first place. This is annoying because if you have packages that don't uninstall correctly then reinstalling them doesn't work. This is particularly bad when you do a large 'upgrade' when opam will uninstall a load of packages, reinstall half, break (package maintainer's fault), and then it's up to the user to remember which packages you had installed beforehand!

AltGr:
Closes #502, #1215

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,4 +1,12 @@
 2.0~alpha
+* Track files installed by packages for cleaner removal and listing, add `opam
+  search --owns-file` (#502, #1215) *this requires packages to properly separate
+  build and install*, or unrelated files could get removed on package uninstall.
+* Allow command wrappers around package build/install/remove commands to be
+  defined in `~/.opam/config`; include sample Linux wrappers to restrict process
+  permissions (e.g. ensure `build:` doesn't install)
+* Reworked `opam show`, with more fields and the ability to select raw fields
+  from the opam file (#2122)
 * Dropped ability to compile opam with OCaml earlier than 4.01.0
 * Version constraints in `depopts:` allowed again, but with consistent semantics
 * Allow mixing filters within dependency constraints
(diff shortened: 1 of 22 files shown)


Drup:
So, this works for both `.install` files and the `install` field ? This is great \o/

AltGr:
Yes, it includes all, so this is a bit redundant since it's an added mechanism, but that's not really a problem. Next, I intend to:
- make build dir unneeded by default for `remove:` (removing the need to download archives)
- add an `uninstall-requires-build-dir` flag
- write a repository-rewriting script to add the above flag where necessary or in doubt (using the infamous findlib-hack from opam, and removing it from opam proper)
- update the recommendations for the `remove:` field in the doc: just skip it if you just have to remove the files you added. Retaining the field for documentation is debatable, but since it's often incorrect, if we go that way CI should check that it indeed uninstalled properly (which is now possible).

Packages doing e.g. `make uninstall` will still need their sources though; maybe we could just suppress their `remove:` field in this case. Note that when there is no `remove:` field, the source is never needed, so removing most `remove:` fields rather than changing the download behaviour might be an alternative to the above.

A nice addition would be to not fail, but just skip the `remove:` step if the package archive can't be downloaded but we have a `.changes` file. Probably once I integrate the patch to do the removals as late as possible, and then include downloads in the actions processing graph rather than a first step.

AltGr:
Where's Travis ?

AltGr:
Filing #2571 seems to have wakened up Travis here too (well, it's the same commit). But in both, @github still displays the commits in the wrong order:
I locally have f10af22 *before* 7cf7b8d

PR #2571, opened 2016-06-02
Track files installed by packages, allow query and use for removal

AltGr:
(re-submit of #2568, that Travis didn't detect and where GH displays
the commits in a wrong order (!?))


samoht:
That's great!

PR #2574, opened 2016-06-03
Add '--upgrade-format' to opam-admin

AltGr:
for now, doing only the compilers to packages transition, but more may
be added later.


avsm:
awesome!

PR #2588, opened 2016-06-18
lint CLI improvements

AltGr:


dbuenzli:
A `-warn-error` option would be nice aswell. In my usage in `topkg` I'd like to selectively silence warnings (number 5) and test for the presence of warning by checking for a non-zero exit code.

dbuenzli:
Btw. I would argue that warning 5 is dubious for OPAM files. You are mixing two things in `opam lint`

1. Linting an OPAM *file*, regardless of context. 
2. Linting an OPAM *package*, i.e. is a directory its name and its files. 

Warning 5. is relevant for 2. only.

AltGr:
1. Yes, this is a start and I intended to extend the `--warnings` argument later on. OCaml has `+N` and `-N` to enable and disable warning N, but also `@N` to mark warning as error. We could use the same, but may also need a way to mark an error as warning: `+N` could be "enable and set as warning", or we could keep it as "enable in its default category" and add a new char for this meaning.

2. Hm, can you remind me what warning 5 is ? I can't find it in my code so it may have been removed already.

dbuenzli:
```
 Warnings found in ~/sync/repos/fmt/_build/fmt-0.8.0-2-g8e65774/opam:
 warning  5: Unexpected version 0.8.0-2-g8e65774, the directory name or pinning implied 8.0-2-g8e65774
```

This is because the OPAM file is in a directory `$PKG-$VERSION` (an extracted distribution tarball) which means we are in 1. and `opam lint` assumes we are in 2. and expects the parent to be `$PKG.$VERSION`.

PR #2589, opened 2016-06-21
Interface improvements, in particular to `opam list`

AltGr:
The long-awaited `opam list` rehaul!

### DIFF
--- a/src/client/opamSwitchCommand.ml
+++ b/src/client/opamSwitchCommand.ml
@@ -495,6 +495,18 @@ let import st filename =
   in
   import_t importfile st
 
+let set_compiler st names =
+  let packages =
+    try List.map (OpamSwitchState.find_installed_package_by_name st) names
+    with Not_found ->
+      OpamConsole.error_and_exit "These packages are not installed: %s"
+        (OpamStd.List.concat_map ", " OpamPackage.Name.to_string
+           (List.filter (not @* OpamSwitchState.is_name_installed st) names))
+  in
+  let st = { st with compiler_packages = OpamPackage.Set.of_list packages } in
+  OpamSwitchAction.write_selections st;
+  st
+
 let guess_compiler_package rt name =
   let package_index =
     OpamRepositoryState.build_index rt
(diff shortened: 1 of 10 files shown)


AltGr:
Note that it could be sound to merge `opam list` and `opam search` now...

To list packages with versions: `opam list --columns=package`

src/client/opamSwitchCommand.ml
```diff
@@ -495,6 +495,18 @@ let import st filename =
...
 
+let set_compiler st names =
+  let packages =
+    try List.map (OpamSwitchState.find_installed_package_by_name st) names
+    with Not_found ->
+      OpamConsole.error_and_exit "These packages are not installed: %s"
+        (OpamStd.List.concat_map ", " OpamPackage.Name.to_string
+           (List.filter (not @* OpamSwitchState.is_name_installed st) names))
+  in
+  let st = { st with compiler_packages = OpamPackage.Set.of_list packages } in
```
dra27: I'm working on a related matter - shouldn't this step also compute the transitive closure of the dependencies of the listed packages? Otherwise the cudf universe has packages which are really base packages depending on other base packages. I'm not sure that that creates a problem at the moment, though it certainly creates a problem in my fork where you can explicitly upgrade base packages.

I'm working on a related issue at the moment: upgrading a compiler package (in this case tcl/tk) which itself draws in a new dependent package (upgrading Tcl/Tk from 8.5 to 8.6 adds a requirement for ZLib)

AltGr: Indeed, this is not perfectly consistent, as currently, when you setup a new switch, the transitive closure of the compiler packages is made into compiler packages, ensuring the core of packages is self-contained.

It was pointed out to me, however, that this constraint might be relaxed without compromising much: you only need to accept that, if dependencies of your compiler packages change, they may get recompiled. For example, if you set `coq` as your compiler, and there is a new version of OCaml, you are guaranteed not to change the coq version, but may need to recompile it to upgrade OCaml.

I am not sure which is best -- there seems to be use-cases for the latter --, but when making the `set-compiler` command I made it completely free of constraints, having this in mind and with the idea that it may allow to experiment with it.

Another approach could be to save only the configured compilers, not the closure, but dynamically compute the transitive closure when loading the universe.

PR #2590, opened 2016-06-21
[experimental] try a more consistent setup without global 'ocaml-version'

AltGr:
This has many consequences, notably regarding system compilers.


AltGr:
This is getting quite usable. The migration (from an earlier .opam) still needs some work, as at the moment package "ocaml" is installed, which is now a virtual package, and on upgrade or fixup opam will pick up one of its possible dependencies randomly (opam-base-compiler, or maybe opam-system, or even some opam-variant).

Also, there is no way yet to get an automatic choice of system compiler. A possible, generic way to fix this could be to:
- add a way to define global variables in `~/.opam/config` using the output of commands
- allow the use of a configuration file on opam init (e.g. `/etc/opamrc` by default). This could be put there by the distribution's opam packages
- define an `ocaml-system-version` variable this way
- add `available: ocaml-system-version = xxx` to the definition of package `ocaml-system.xxx`

This is completely OCaml independent, and may prove useful for many other cases.

AltGr:
All the pieces should now be in place; I'll shortly merge this, and switch the format of the 2.0 repo at opam.ocaml.org/2.0~dev

AltGr:
Tests failing in CI are due to the repository not being rewritten yet (there is no compiler named `ocaml-system`, which is the default, in the older repository format).

PR #2593, opened 2016-06-28
Delay removal of packages as much as possible.

OCamlPro-Henry:
Now that OPAM is able to distinguish between the 'build' and 'install' rules, we want, when reinstalling a package, to build the new version before to remove the old one: whenever the new version fails to build, the old one is still installed and the opam repository is left untouched.

This is what this patch implements.

Unfortunatly, we might not apply this strategy with all packages. When a package A depends on a package B, for preserving the consistency of then .opam directory we still need to remove A before to remove B. Then, we can't wait for B to be built before to remove the old version. Still, we wait for A to be successfully built before to remove B.

### DIFF
--- a/src/solver/opamActionGraph.ml
+++ b/src/solver/opamActionGraph.ml
@@ -230,6 +230,9 @@ module Make (A: ACTION) : SIG with type package = A.package = struct
   let explicit ?(noop_remove = (fun _ -> false)) g0 =
     let g = copy g0 in
     let same_name p1 p2 = A.Pkg.(name_to_string p1 = name_to_string p2) in
+    (* We insert a "build" action before any "install" action.
+       Except, between the removal and installation of the same package
+       (the removal might be postponed after a succesfull build. *)
     iter_vertex (fun a ->
         match a with
         | `Install p | `Reinstall p | `Change (_,_,p) ->
@@ -242,6 +245,12 @@ module Make (A: ACTION) : SIG with type package = A.package = struct
         | `Remove _ -> ()
         | `Build _ -> assert false)
       g0;
+    (* For delaying removal a little bit, for each action "remove A" we add
+       a constraint "build B -> remove A" for transitive predecessors
+       of "A" that do not have dependencies.
+
+       For adding a little bit more delay, we ignore dependencies that do not
+       modify the prefix (see [OpamAction.noop_remove]) *)
     let closed_predecessors = compute_closed_predecessors noop_remove g in
     iter_vertex (function
         | `Remove p as a ->
--- a/src/solver/opamActionGraph.ml
+++ b/src/solver/opamActionGraph.ml
@@ -130,6 +130,7 @@ module Make (A: ACTION) : SIG with type package = A.package = struct
   include OpamParallel.MakeGraph(A)
 
   module Map = OpamStd.Map.Make (A.Pkg)
+  module Set = OpamStd.Set.Make (A.Pkg)
 
   (* Turn concrete actions (only install, remove and build) to higher-level
      actions (install, remove, up/downgrade, recompile). Builds are removed when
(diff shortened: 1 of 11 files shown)


src/solver/opamActionGraph.ml
```diff
@@ -132,7 +132,7 @@ module Make (A: ACTION) : SIG with type package = A.package = struct
   module Map = OpamStd.Map.Make (A.Pkg)
   module Set = OpamStd.Set.Make (A.Pkg)
 
-  (* Turn concrete actions (only install, remove and build) to higher-level
+  (* Turn noop_remove actions (only install, remove and build) to higher-level
```
AltGr: ???

src/solver/opamActionGraph.mli
```diff
@@ -36,16 +36,18 @@ module type SIG = sig
...
   val reduce: t -> t
 
-  (** Expand install actions, adding a build action preceding them. *)
-  val explicit: t -> t
-end
+  (** Expand install actions, adding a build action preceding them.
+      The argument [concrete] is a function that should return `true`
+      on package where the `remove` action is a `noop`
+      (such as `conf-*` package). *)
+  val explicit: ?noop_remove:(package -> bool) -> t -> t end
```
AltGr: I am not completely sure this is where this should be done: even if they don't have an effect on the prefix, the installations and removals of `conf` packages get registered in the set of installed packages, and might e.g. have an effect on the environment. As I understand it, a way to express our "optimisation" could be "does not need to be removed before reinstalling", which could be done by simply turning reinstalls to installs beforehand (and making sure that, in case of `conf` packages, installation doesn't fail if the package is already registered as installed).

OCamlPro-Henry: My comments is indeed misleading. the `noop-remove` really mean "the removal of this package won't touch the prefix". It might have effect on the environment, the previous patch https://github.com/ocaml/opam/pull/2593/commits/0d1275843f841ff77c3efcb55a01e2c648aa944f already handles that. What I am trying to detect here is : whether we might build a package before to remove/install its predecessors from the prefix (in all circumstance, it is build in an environment where the variable installed by its predecessors are removed). Turning reinstall into install for conf-* package does not seem sufficient to detect that AFAIK.

OCamlPro-Henry:
The last patch delays removal one (little) step further: it delays the removal after building packages that only depend on packages that do not modify the prefix (the only side effect we can't trivially simulate while building a package).

The CI requires : https://github.com/ocaml/opam-rt/pull/50

samoht:
> Now that OPAM is able to distinguish between the 'build' and 'install' rules, we want, when reinstalling a package, to build the new version before to remove the old one: whenever the new version fails to build, the old one is still installed and the opam repository is left untouched.

Does it mean that it will break for packages which does not respect that constraint? Is the error easy to detect/signal to the user?

AltGr:
@samoht: yes: the package can now be built, then removed, then installed. So if the installation was in the `build` step, you won't actually get your files (unless they were completely different between the two versions)

To avoid that, we have several guards:
- the rewrite I commited to opam-repository
- an `opam lint` heuristic that warns when there are `remove` but no `install` instructions
- the namespace command wrappers that forbid `build` instructions from writing to the prefix. I have them locally, and they are expected to get into the travis checks (before or when 2.0 is released)

so "don't write to prefix during `build:`" becomes a hard constraint, but it will be checked.

samoht:
Cool, thanks!

> the namespace command wrappers that forbid build instructions from writing to the prefix. I have them locally, and they are expected to get into the travis checks (before or when 2.0 is released)

I would be definitely interested to test them with Docker for Mac to see what cross-compilation magic we can do with them.

AltGr:
The scripts are [here](https://github.com/ocaml/opam/blob/master/shell/wrap-build.sh) and [here](https://github.com/ocaml/opam/blob/master/shell/wrap-install.sh) ; they require `sysctl kernel.unprivileged_userns_clone=1` on Debian, and can be set through `wrap-build-commands: "<path>"` and similar in `~/.opam/config`.

AltGr:
(a safe workaround for repos not separating build and install would be to put everything in `install:` instead of `build:`; you lose all parallelism though, and any benefit from doing removals last, obviously)

Issue #2598, opened 2016-07-01
Set user-agent when downloading from the repo


AltGr:
this would give interesting stats on the official repository, and may help track what versions of opam are used.

PR #2603, opened 2016-07-04
Set user-agent on opam downloads

rgrinberg:
The user agent will appear as something like "opam/1.2.2"

My modest attempt at #2598


AltGr:
Awesome, thanks!

Issue #2520, opened 2016-04-16
Next: synopsis field


dbuenzli:
As reported on list `next` allows to specify the `descr` file in the OPAM file itself. It would be nicer to split the field in two with a `synopsis` field that holds the first lines of `descr` files and have `descr` (which could even be called `description`) hold the rest.

PR #2615, opened 2016-07-22
Cleanup OCaml references, verbose uninstall, add 'synopsis:'

AltGr:


dbuenzli:
The last commit would close #2520, I'd suggest however not to have the behaviour you mention in the commit. It complicates the file format and thus its understanding by both machine and users.

src/format/opamFile.ml
```diff
@@ -2078,6 +2075,9 @@ module OPAMSyntax = struct
          Pp.V.constraints Pp.V.compiler_version);
       "os", no_cleanup Pp.ppacc_opt with_os OpamStd.Option.none
         Pp.V.os_constraint;
+      "descr", no_cleanup Pp.ppacc_opt with_descr OpamStd.Option.none
```
dbuenzli: Kill that otherwise you have more than one way of specifying the same thing...

PR #2627, opened 2016-07-27
Improvements to `opam list` and `opam show`

AltGr:
Thanks @Drup for the feedback!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ bootstrap/
 tests/tmp/
 .*.swp
 src_ext/cudf/
+src_ext/cppo/
 src_ext/dose/
 src_ext/cmdliner/
 src_ext/extlib/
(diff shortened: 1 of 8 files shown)


Drup:
I guess it's not possible to give a list of the possible fields in the man page ?

AltGr:
@Drup from `opam show --help`:
```
       -f FIELDS, --field=FIELDS
           Only display the values of these fields. Fields can be selected
           among name, version, package, synopsis, synopsis-or-target,
           description, <field>:, installed-version, pin, opam-file,
           all-installed-versions, available-versions, all-versions,
           repository, installed-files. Multiple fields can be separated with
           commas, in which case field titles will be printed; the raw value
           of any opam-file field can be queried by suffixing a colon
           character (:), e.g. --field=depopts:.
```
Or did you mean the list of all "opam-file field[s]" ? This could be extracted too, but we need to remove the ones that exist for backwards compatibility only.

Drup:
Ah, I meant "flag", sorry!

AltGr:
They are listed in `opam list`:
```
       --has-flag=FLAG
           Only include packages which have the given flag set. Package flags
           are one of: light-uninstall verbose plugin compiler conf
```
Tags are not, obviously, since they are free-form strings.

Drup:
I missed that one! Great.

PR #2632, opened 2016-08-04
add link to the "distribution" wiki page

c-cube:


AltGr:
Thanks!

This will need a backport to the 1.2 branch to be visible on the main site though (2.0 doc is available already, but in a special entry in the menu, and under subdir `2.0/`)

PR #2634, opened 2016-08-05
Allow per-switch repository selection

AltGr:
Implementation is here but the UI is in progress


AltGr:
... and quite a few other improvements

PR #2651, opened 2016-09-01
Local switches

AltGr:
This PR provides the possibility to use directories as switch handles,
which holds the switch contents below `_opam` at that path. The switch
is then automatically selected if opam is run from the given directory
or a descendent.

External switches are still registered globally in ~/.opam/config,
which allows listings, format upgrade, global repository settings
changes, etc. ; but the case where they aren't registered should be
handled nicely.

### DIFF
--- a/opam-devel.opam
+++ b/opam-devel.opam
@@ -13,7 +13,7 @@ authors: [
 ]
 homepage: "https://opam.ocaml.org"
 bug-reports: "https://github.com/ocaml/opam/issues"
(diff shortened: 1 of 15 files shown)


AltGr:
Note: the switch doesn't register the OPAMROOT it belongs to ; so stuff might get weird if you have several roots and commonly set OPAMROOT, and switches are automatically selected.
The only sane behaviour would be to skip selecting a switch that belongs to an opamroot different from what is selected.

PR #2663, opened 2016-09-09
Big manual rehaul, including documentation of all file formats

AltGr:


Drup:
Nice!

How do we access the new repo fields (browse, upstream) from the command line ?

dbuenzli:
> fields (browse,

I'm not sure this is a very good name and feels a little bit too specific, a repository may actually have a homepage with other instructions, rather than just an html ui to browse the packages. Did you consider reusing the other fields that already exist in opam files (`homepage` or `doc`) ?

doc/pages/Manual.md
```diff
@@ -7,33 +7,228 @@ primarily of use for packagers, package maintainers and repository maintainers.
...
+
+Except in explicit cases, opam only alters files within its opam root. It is
+organised as follows:
+- [`~/.opam/config`](#config): the global opam configuration file
+- `~/.opam/repo/`: contains the mirrors of the configured package repositories
+- [`~/.opam/repo/repos-config`](#reposconfig): lists the configured package repositories and their URLs
+- `~/.opam/repo/<repo>`: mirror of the given repository
+- `~/.opam/opam-init/`: contains opam configuration scripts for the outside world, e.g. shell environment initialisation
+- `~/.opam/cache/`: caches of downloaded files
+- `~/.opam/plugins/<plugin>`: reserved for plugins
```
dbuenzli: Aren't plugins per switch ?

doc/pages/Manual.md
```diff
@@ -7,33 +7,228 @@ primarily of use for packagers, package maintainers and repository maintainers.
...
+- `~/.opam/<switch>`: prefixes of named [switches](#switches)
+
+### Repositories
+
+Repositories are collection of opam package definitions. They respect the
+following hierarchy:
+- [`/repo`](#repo): repository configuration file
+- [`/packages/<pkgname>/<pkgname>.<version>/opam`](#opam): holds the metadata
+  for the given package. `url` and `descr` may also be present, in which case
+  they override what may already be present in the `opam` file
```
dbuenzli: Is there are rationale to the override direction ? `url` and `descr` seems rather a backcompat thing than a feature. If this is the case I would prefer the converse: that way if I'm reading an opam file with these fields I know these are the actual values that will be used, I don't have to wonder if they may be overriden.

doc/pages/Manual.md
```diff
@@ -7,33 +7,228 @@ primarily of use for packagers, package maintainers and repository maintainers.
...
+- [`/packages/<pkgname>/<pkgname>.<version>/files/`](#files): contains files to copy over
+  the source of the given package before it gets used
```
dbuenzli: This should be made more precise. Where exactly are they copied (at the root of the source tree) what happens to subdirs in `files` are they copied at the root or in corresponding subdir starting at the root.

doc/pages/Manual.md
```diff
@@ -7,33 +7,228 @@ primarily of use for packagers, package maintainers and repository maintainers.
...
+- `/archives/<pkgname>.<version>+opam.tar.gz`: cached package archives
+  (as generated by running `opam-admin make`)
+- `/urls.txt`, `/index.tar.gz`: index files useful when serving over HTTP, they
+  can be generated using `opam-admin make --index`
+
+Opam can be configured to use repositories at local or remote (ssh) paths, HTTP
```
dbuenzli: I would rather say that opam repositories can be accessed using one of these mecanisms.

doc/pages/Manual.md
```diff
@@ -7,33 +7,228 @@ primarily of use for packagers, package maintainers and repository maintainers.
...
+URLs, or on version control systems. A repository is set up using
 
-This version of the manual documents the version `1.2` of the file formats, and
-is progressively being updated with the new features of opam `2.0`.
+```
+opam repository add <name> <URL>
+```
+
+and can subsequently be then selected for use in specific switches using `opam
+repository select <name>`. Use `opam repository list --all` for an overview of
```
dbuenzli: I don't like this two step procedure which forces you to introduce the `select` terminology (how do you handle remove `unselect` ?). The whole thing feels bureaucratic from an UI point of view. I think this could be simplified by having something like:

```
opam repo add [--switch SWITCH|--to-list] NAME [URL]
```
Basic semantics is:
```
opam repo add name            # add name to current switch (must exist in the global list)
opam repo add --switch s name # same as above but do this in switch s 
opam repo add name url        # add name/url to global list and to current switch (error if name exists)
opam repo add --to-list name url  # add name/url to global list only
etc.
```

doc/pages/Manual.md
```diff
@@ -7,33 +7,228 @@ primarily of use for packagers, package maintainers and repository maintainers.
...
+Command `opam config env` returns the environment updates corresponding to the
+current switch, in a format readable by your shell, and when needed opam will
+prompt you to run:
+
+```
+eval $(opam config env)
+```
 
-## General file format
+A switch is created using `opam switch create <switch> (<compiler>|--empty)`.
```
dbuenzli: I won't comment further on this, I already made my points more than once, but I think that the story you have about switch creation is utter crap.

doc/pages/Manual.md
```diff
@@ -7,33 +7,228 @@ primarily of use for packagers, package maintainers and repository maintainers.
...
+- `<switch>` can be either a plain name, or a directory name (if containing `/`
+  or starting with `.`). In the latter case the switch is _local_ and instead of
+  being held at `~/.opam/<switch>`, it will be created in the given directory,
+  as a `_opam` subdirectory. Local switches are automatically selected depending
+  on the current directory, see above.
+- If a `<compiler>` is selected, opam will install the corresponding packages
+  and their dependencies in the new switch. These packages will be marked as
+  _base_, protected against removal and unaffected by upgrade commands.
+  `<compiler>` can be selected among packages which have the `compiler` flag
```
dbuenzli: Why not call that flag `base` ? Since that's the terminology you use. This all seems very ad-hoc, you could simply have a notion of base package that once installed is not upgradeable. The thing is that you may want to have more than one of these packages (which immediately breaks your switch init story).

doc/pages/Manual.md
```diff
@@ -7,33 +7,228 @@ primarily of use for packagers, package maintainers and repository maintainers.
...
+  of installed, base, pinned packages
+- `<switch-prefix>/.opam-switch/environment`: contains the environment variable
+  settings for this switch
+- `<switch-prefix>/.opam-switch/reinstall`: list of packages marked for
+  reinstallation (development packages where changes have been detected)
+- [`<switch-prefix>/.opam-switch/config/<pkgname>.config`](#ltpkgnamegtconfig):
+  installed package's, opam specific configuration
+- [`<switch-prefix>/.opam-switch/install/<pkgname>.install`](#ltpkgnamegtinstall):
+  `.install` files used to install the given package
+- `<switch-prefix>/.opam-switch/install/<pkgname>.changes`: file changes done by
```
dbuenzli: What is this ? Do you mean file system changes ?

doc/pages/Manual.md
```diff
@@ -7,33 +7,228 @@ primarily of use for packagers, package maintainers and repository maintainers.
...
+  the source.
+- `<switch-prefix>/.opam-switch/backup`: snapshots of previous states of the
+  switch, and other backup files.
+
+#### Pinning
+
+Pinning is an operation by which a package definition can be created or altered
+locally in a switch.
+
+In its most simple form, `opam pin <package>`, `<package>` is bound to its
```
dbuenzli: Did you reintroduce the `add`less form or is the slip we all do ?

doc/pages/Manual.md
```diff
@@ -7,33 +7,228 @@ primarily of use for packagers, package maintainers and repository maintainers.
...
+
+`opam pin` can further be used to divert the source of a package, or even create
+a new package, using `opam pin <package> <URL>` ; this is also very useful to
+point to a local path containing a development or patched version of the package
+source. If the source you are pointing to contains a matching `<pkgname>.opam/`
+or `opam/` directory, a matching `<pkgname>.opam` or `opam` file, that will be
+used as the initial package metadata.
+
+Whenever an install, reinstall or upgrade command-line refers to a pinned
+package, opam first fetches its latest source. `opam update` otherwise updates
```
dbuenzli: On thing that needs to be clarified about opam update (in the description of repos) is whether opam update updates all the repos in the global list or only the repos used in the current switch (would personally prefer the latter)

doc/pages/Manual.md
```diff
@@ -7,33 +7,228 @@ primarily of use for packagers, package maintainers and repository maintainers.
...
+<bool>          ::= true | false
+<int>           ::= [ "-" ] { <digit> }+
+<string>        ::= ( " { <char> }* " ) | ( """ { <char> }* """ )
+<operator>      ::= { "!" | "=" | "<" | ">" | "|" | "&" }+ | [ ":" ] <operator> [ ":" ]
+<list>          ::= "[" { <value> }* "]"
+<option>        ::= <value> "{" { <value> }* "}"
+<comment>       ::= ( "(*" { <char> }* "*)" ) | ( "#" { <char\newline> }* <newline> )
+```
+
+The OPAM file formats share a common base syntax. The files are utf8 encoded and
```
dbuenzli: UTF-8

doc/pages/Manual.md
```diff
@@ -133,24 +347,25 @@ form: `name1+name2+name3:var` is the conjunction of `var` for each of `name1`,
...
-   package's own OPAM file (when not shadowed by another variable), in which
-   case the variables will relate to the version of the package being defined.
-   Otherwise, the installed version is always used, and most variables will be
-   undefined when the package isn't installed.
+   specific to a given package, for example `foo:installed` is a boolean
+   variable that can be used to check if package `foo` is installed, and
+   `foo:lib` is its library installation directory. The prefix `package-name:`
+   can be replaced by `_:` within the package's own OPAM file, or simply omitted
+   when not shadowed by another variable. In this case, the variable always
+   relates to the version of the package being defined. Otherwise, `foo:bar`
```
dbuenzli: This behaviour seems a bit dodgy. IIUC this means that if I am in the context of `p`, `_:var` refers to the version of the package being defined but `p:var` refers to its previous version ? FWIW I had to use e.g.`p:etc` in the build steps of my own packages as I don't think `_:etc` is available in current OPAM.

doc/pages/Manual.md
```diff
@@ -306,68 +517,18 @@ fields:
...
+first line of the file defines the package's synopsis, while the rest defines
+its long description.
 
-> NOTE: handling directory or files of the form `package-name.opam` is a new
-> feature in 1.2.1. It should only be used for source repositories that may hold
-> more than one OPAM package.
+This information can be embedded in `opam` package definition files using the
+[`synopsis:`](#opamfield-synopsis) and [`description:`](#opamfield-description)
+fields since opam version 2.0, but if a `descr` file is present, it takes
+precedence.
```
dbuenzli: As I already said above I would prefer the converse overriding behaviour.

doc/pages/Manual.md
```diff
@@ -306,68 +517,18 @@ fields:
...
-- from the repository if no metadata subdirectory was found above, excluding the
-  opam file is one was found above
-- from a template, opening an editor, if nothing was found in either the package
-  source or the repositories.
+#### descr
 
-Furthermore, the metadata of pinned package is then stored within the switch (at
-`$(opam show --where <package>)`), and the `opam` file can be edited through
-`opam pin edit`
+Descr is a plain utf8 text file without specific syntactic constraints. The
```
dbuenzli: UTF-8

doc/pages/Manual.md
```diff
@@ -391,28 +552,38 @@ It has the following fields:
 - <a id="urlfield-mirrors">`mirrors: [ <string> ... ]`</a>:
   an optional list of mirrors. They must use the same protocol as the main URL.
 
-### files/
+These contents can be embedded within the [`url {}`](#opamsection-url) section
+of an `opam` file; however, if an `url` file is present besides the `opam` file,
+its contents take precedence.
```
dbuenzli: Same as last comment.

doc/pages/Manual.md
```diff
@@ -391,28 +552,38 @@ It has the following fields:
...
-will be copied over the root of the package source. `opam` file fields like
-`patches:` refer to files at that same root, so patches are typically included
-here.
+will be copied over the root of the package source. [`opam`](#opam) file fields
+like [`patches:`](#opamfield-patches) refer to files at that same root, so
+patches specific to opam are typically included in this subdirectory.
 
 Note that repository archives generated by `opam-admin make` in `archives/`
-_already_ include these overlay files.
+already include these overlay files.
```
dbuenzli: Don't understand what it means exactly.

doc/pages/Manual.md
```diff
@@ -391,28 +552,38 @@ It has the following fields:
...
 
-#### package-name.install
+#### <pkgname>.install
+<a id="packagenameinstall"></a>
 
-This file is convenient to automate package installation without resorting to
-external commands like `make install`. It can be included in `files/` but also
-at the root of the package source, or generated by the build-system.
+This file format describes the installation from a source directory to an
+installation prefix. It will be used by opam if present in the package's source
```
dbuenzli: At the root of the package's source directory.

doc/pages/Manual.md
```diff
@@ -457,20 +631,21 @@ The following are treated slightly differently:
...
+  the file format version, should be `2.0` as of writing.
 
-- <a id="opamfield-name">`name: <string>`</a>,
-  <a id="opamfield-version">`version: <string>`</a>:
+- <a id="opamfield-name">`name: <pkgname>`</a>,
+  <a id="opamfield-version">`version: <version>`</a>:
   the name and version of the package. Both fields are optional when they can be
-  inferred from the directory name (e.g. when the file sits in the repository).
+  inferred from the directory name (e.g. when the file sits in the repository),
+  but should always be set for package definitions within source trees.
```
dbuenzli: What do you mean by that ? People should not commit version numbers to their source repositories.

doc/pages/Manual.md
```diff
@@ -698,3 +866,248 @@ recommended to check the validity and quality of your `opam` files.
...
+  defines a long description (one or more paragraphs) for the package. This can
+  also be defined as the body of an external [`descr`](#descr) file.
+
+- <a id="opamsection-url">`url "{" <url-file> "}"`</a>:
+  defines the URL where the package source can be obtained. This section has
+  contents in the same format as the [`url`](#url) file, and has the same effect
+  as using a separate `url` file.
+
+- <a id="opamfield-setenv">`setenv: [ <environment-update> ... ]`</a>:
+  defines environment variables updates that will be applied upon installing the
```
dbuenzli: What is the environment variable definition policy whenever multiple package define or extend the same variable name ? Something deterministic that doesn't depend on the install sequence has to be defined otherwise this could become a source of non-reproducible builds (and other user support nightmares).

doc/pages/Manual.md
```diff
@@ -698,3 +866,248 @@ recommended to check the validity and quality of your `opam` files.
...
+#### repos-config
+
+This file is at `~/.opam/repo/repos-config` and lists all package repositories
+configured on the system, and their source URLs.
+
+- <a id="reposconfigfield-repositories">`repositories: [ <string> { <URL> } ... ]`</a>:
+  lists the configured repository idents and their URLs.
+
+Note that repositories without URLs can be used. In that case, the data in the
+corresponding directory will be used but never updated by opam ; also, caching
```
dbuenzli: What would opam update there ? opam only *reads* from repos no ?

doc/pages/Manual.md
```diff
@@ -256,46 +497,16 @@ reference name (branch, commit, HEAD...): `git://foo.com/git/bar#master`,
...
+#### repo
+<a id="Repospecification"></a>
 
 The `repo` file, placed at the root of a repository, has the following optional
 fields:
 
 * <a id="repofield-opam-version">`opam-version: <string>`</a>:
-  File format and repository format version, should be `1.2` as of writing.
+  File format and repository format version, should be `2.0` as of writing.
 * <a id="repofield-browse">`browse: <string>`</a>:
```
dbuenzli: As I said in a comment I don't feel like there's the need to introduce a new name here, `homepage` or `doc` would do.

dbuenzli:
@AltGr Thanks for writing this. It seems you already merged this but you may want to check out the comments I made, some of which are related to the way the tool works.

doc/pages/Manual.md
dsheets: `browse` already exists and implies a path structure at the endpoint (`[pkg]/[pkg.version]` relative resolution) that can be used to synthesize package description URLs. Neither `homepage` nor `doc` has this semantic.

dbuenzli: Well then this should be properly documented (and I would still dispute the `browse` name).

dsheets: What would you dispute about `browse`? Do you have a suggestion for an alternative? I believe `browse` was introduced around 1.2.0.

doc/pages/Manual.md
AltGr: Hm, good point on `opam repo add <name>`, that would feel more natural;
then `opam repo remove` whould remove just from the current switch by default ? (And remove the repo if no switch is using it anymore ?)
The manpage can be seen at http://opam.ocaml.org/doc/2.0/man/opam-repository.html

One point that bothered me when designing this UI is that we also need to maintain a "default" repository configuration, to be used for newly created switches.

doc/pages/Manual.md
AltGr: One thing at a time.

doc/pages/Manual.md
AltGr: indeed, `base` and `compiler` are almost synonyms and get a bit mixed up, the story could be clearer.
Dependencies of your `compiler` package get installed at switch init, and are marked as `base` too.

doc/pages/Manual.md
AltGr: yes, this basically records an abridged diff corresponding to the package installation

doc/pages/Manual.md
AltGr: *whistles innocently*
(really, after 1 year people still stumbled on this change; and I think I got error messages mostly alright)

doc/pages/Manual.md
AltGr: The latter, unless you specified `--all`; I'll make it explicit.

doc/pages/Manual.md
AltGr: Hm, no, that was unclear: rest assured, if you are in the opam file defining some version of `p`, `p:var` will always refer to the variable at that version. `_:` doesn't do any magic besides selecting the package name; and indeed, it's new in 2.0.

doc/pages/Manual.md
AltGr: Hmm, maybe that's not good advice. It allows `opam pin add project-dir/` to pick the version up though...

doc/pages/Manual.md
AltGr: - package scripts only "see" updates done by packages they depend on
- otherwise, at the moment it's unspecified, as my idea was that packages really shouldn't depend on env variable update order.
- for consistency, though, we should at least use a topological order on dependencies. The [implementation](https://github.com/ocaml/opam/blob/84bd2759fb343bf1b2ffecbd796d25e5cae7545b/src/state/opamEnv.ml#L201-L201)  currently just follows package order (i.e. alphabetical)

doc/pages/Manual.md
AltGr: It can be used for directly testing and hot-patching local repositories, or have package overlays specific to your opam root. Yes, this is kind of tautological in this case, opam doesn't update since it doesn't have a source to update from, but the fact that the internal directory is only used read-only by opam is worth mentioning.
This is mainly directed towards experiments, at the moment, there is no UI to handle that.

doc/pages/Manual.md
AltGr: Yes, that wasn't documented before, but it is not new in 2.0, in fact, I don't think I've touched it.

doc/pages/Manual.md
dbuenzli: Not that I disagree with the change.

AltGr:
@dbuenzli: thanks a lot for your detailed review. It's merged so that it can more easily be browsed from http://opam.ocaml.org/doc/2.0/Manual.html, but this isn't intended as a freeze of any kind.

Design questions are also still open (I'll keep the manual updated as I go) and will still be during the upcoming Beta phase (I am actually hoping to get more feedback on some of the aspects).

doc/pages/Manual.md
dbuenzli: Just make sure there's a way to pin a package called of one of the `COMMANDS`.

doc/pages/Manual.md
dbuenzli: > then opam repo remove whould remove just from the current switch by default ?

Yes. Basically you want to maintain the invariant that OpAm's command set always works on the current switch by default.

> (And remove the repo if no switch is using it anymore ?)

I'd say no, keep it in the global list (unless you want to track if repos where added with `--to-list` or not). You need a `--to-list` for `remove` as well, that would be used to remove from the list if not used by any switch (and maybe a --force to remove it regardless). Something better than `--to-list` needs to be found so that we can have the same option in both cases (`--globally` ? bheu, we need a good name for the global list and then the option will follow).

AltGr:
About the `browse:`, etc. fields for the `repo:` file, we should be very careful not breaking compat, because this is the one file opam needs to read from different version repositories to process redirections.

doc/pages/Manual.md
AltGr: they are installed per switch, but may need to store information globally. Currently, `opam-publish` uses this to store its auth tokens.

doc/pages/Manual.md
AltGr: Indeed, that is for compatibility; the choice was made originally so that pinning could still work by adding an `url` file to override the package source, but it no longer works that way (pinning is now just an indirection of the package's definition, including url).
Either way could be really confusing though ; having both could also be an error, at least on the developer side. I can change the way it works.

doc/pages/Manual.md
AltGr: I should make the links more obvious, but there is a `files/` subsection that describes the behaviour in more detail.

dbuenzli: Yes I saw that afterwards. Having `more details` link to the details is always nice.

PR #2668, opened 2016-09-13
Correct the extlib-1.5.3 source URL, now that Google Code is retired

avsm:


AltGr:
Thanks! I'll look into the Travis issue, seems unrelated

PR #2674, opened 2016-09-15
Let's wait for beta

AltGr:


CHANGES
```diff
@@ -1,4 +1,4 @@
-2.0~beta
+2.0~alpha4
```
Drup: Oh dear, my opam install is not going to like that, doesn't it ?

AltGr: oh, sorry; you managed to get to beta already ? (that was merged earlier today)
Just edit the field in `~/.opam/config` to put `alpha4` instead and you should be fine.

Thanks again for trying the bleeding edge!

Drup: Yeah, I saw the `beta` update and though "Oh, I haven't updated since a few weeks, that's a good occasion!"

AltGr: On second thought that should cause you no trouble if you were already on alpha3 (the format didn't change in between, so no upgrade should have been performed, and the `opam-version` field is left unchanged in that case)

PR #2681, opened 2016-09-26
Strip packages conflicting with the compiler

AltGr:
in fact stripping all conflicting compiler versions. Solver times back to normal

(we were already hinting to the solver that we wanted to keep the specific package versions,
so it could have done this optimisation itself, but it seems it didn't)

### DIFF
--- a/src/core/opamConsole.ml
+++ b/src/core/opamConsole.ml
@@ -55,8 +55,8 @@ let utf8, utf8_extended =
 
 let timer () =
   if debug () then
-    let t = Sys.time () in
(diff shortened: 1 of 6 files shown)


Drup:
Does it also works when I have no set compiler ?

Drup:
(With a compiler set, it is indeed much much better!)

AltGr:
no, I'm afraid that when no compiler is set, we have no choice but to let the solver manage all the versions of the `ocaml` dependency that are around (and take its time doing so).

PR #2685, opened 2016-09-27
Add sha1, sha256 support

AltGr:
This adds the syntax `algo=hexvalue` to discriminate between hash
algorithms, e.g. `md5=d41d8cd98f00b204e9800998ecf8427e`; when the right
length and no prefix is there, assume md5 for compatibility.

Adds support for sha1 and sha256 hashes, through command-line calls to
`openssl`.


samoht:
I think it is more standard to use `:` instead of `=` but I might be wrong. /cc @dsheets

hannesm:
this is great, but please let us skip sha1... we support a weak algorithm (md5) already, future choices should be sha256 and sha512 (which according to some paper is nowadays even faster on 64bit platforms due to larger block size), but sha1 should be deprecated...

src/core/opamHash.ml
```diff
@@ -0,0 +1,97 @@
...
+let to_json s = `String (to_string s)
+
+let to_path (kind,s) =
+  [string_of_kind kind; String.sub s 0 2; s]
+
+let compute ?(kind=default_kind) file = match kind with
+  | `MD5 -> md5 (Digest.to_hex (Digest.file file))
+  | `SHA1 | `SHA256 ->
+    match
+      OpamSystem.read_command_output ["openssl"; string_of_kind kind; file]
```
samoht: on OSX (at least) you can use `shasum -a [1|256|512|...]` instead of requiring openssl.

hannesm: I'd expect sth similar to `download-command`, and use openssl by default (since it is widely available on lots of platforms), and sha256/sha256sum are rather platform-dependent afaik

AltGr: Yes, I was under the impression that `openssl` would expose a more standard interface. Note that the output differs:
```
[lg@maitake{4.02.3} ~]% openssl md5 /dev/null 
MD5(/dev/null)= d41d8cd98f00b204e9800998ecf8427e
[lg@maitake{4.02.3} ~]% md5sum /dev/null     
d41d8cd98f00b204e9800998ecf8427e  /dev/null
```

AltGr:
@samoht: the reason to choose `=` over `:` is that it can fit in filenames; the `to_path` function prefers something like `md5/ab/abcdef12345...` though, so it's not a concrete aissue.

@hannesm: handling is generic, so adding or removing functions is straightforward (as long as openssl supports them). The only reason to keep sha1 would be to interact with `git`, but I don't see how that would be needed in this module.

src/core/opamHash.ml
hannesm: the advantage of openssl is that its output is consistent on all platforms (linux, BSD, MacOSX); while the output md5sum vs md5 (as it is called on FreeBSD at least; or sha256sum vs sha256) might differ.  certainly openssl is a heavier dependency than md5/sha256, thus depending on md5/sha256 is nicer - but please keep in mind that you'll need platform-specific code (as you already have with wget/curl I believe)...

hannesm: but as mentioned earlier, this can be solved by a `download-command` like configuration option which by default uses `openssl md5 %file% | cut -d '=' -f 2` ;)

PR #2691, opened 2016-09-30
Restrict package names to a sane character set

### LINKED ISSUE #2683: Constrain package names to a reasonable set
I think package names are not constrained enough at the moment. For various reasons (one being that these names do show up in filesystems) I think it would be wiser to constrain them to lowercase US-ASCII identifiers. This means something like `[a-z][a-z_\-]*`

AltGr:
Closes #2683

Anyone who had pinned packages with characters not in the restricted set: sorry about this.


dbuenzli:
The PR is missing an update here https://github.com/ocaml/opam/blob/master/doc/pages/Manual.md#package-formulas

AltGr:
Indeed, thanks!

PR #2702, opened 2016-10-03
Experiment: Unquote package names in opam files

AltGr:
(when they start with a letter; quoted ones are still accepted)

this would make the syntax a little more lightweight. E.g.:

```
depends: [
  "ocaml" {>= "4.01"}
  "ocamlfind"
  "lwt" {>= "2.4.0"}
  "zed" {>= "1.2"}
  "react" {>= "1.0.0"}
  "ocamlbuild" {build}
]
```

becomes

```
depends: [
  ocaml {>= "4.01"}
  ocamlfind
  lwt {>= "2.4.0"}
  zed {>= "1.2"}
  react {>= "1.0.0"}
  ocamlbuild {build}
]
```

You should still not forget to quote versions. ~~And `"0install"`.~~

### DIFF
--- a/src/format/opamFormat.ml
+++ b/src/format/opamFormat.ml
@@ -96,6 +96,24 @@ module V = struct
       (fun ~pos:_ -> function String (_,s) -> s | _ -> unexpected ())
       (fun str -> String (pos_null,str))
 
+  (** Parses either string or ident, prints to ident if valid, to string
+      otherwise *)
+  let string_or_ident =
+    pp ~name:"string-or-ident"
+      (fun ~pos:_ -> function
+         | String (_,s) | Ident (_,s) -> s
+         | _ -> unexpected ())
+      (fun str ->
+         if Some true =
+            OpamStd.String.fold_left (fun acc c ->
+                if acc = Some false then acc else match c with
+                  | 'a'..'z' | 'A'..'Z' -> Some true
+                  | '0'..'9' | '-' | '_' | '+' -> acc
+                  | _ -> Some false)
+              None str
+         then Ident (pos_null, str)
+         else String (pos_null, str))
+
   let string_tr = string -| pp (fun ~pos:_ -> OpamStd.String.strip) (fun x -> x)
 
   let simple_arg =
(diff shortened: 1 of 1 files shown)


hannesm:
this looks nice!  why is there the need to have the version in quotes?

hannesm:
I'd disallow package names with leading numeric symbol (in the main opam repository there is currently only `0install`)

AltGr:
> I'd disallow package names with leading numeric symbol (in the main opam repository there is currently only 0install)

I have an improvement that has them parsing fine too, without having to add an "except if" quoting rule, so they shouldn't be a problem ?

> why is there the need to have the version in quotes?

Thing is that, contrary to package names, they (now) appear in a filter context, where variables are allowed too:
`foo { >= "bar" }` is very different from `foo { >= bar }` which -- in opam 2.0 -- will try and expand the `bar` variable.

That may be error-prone, which might be a reason why we wouldn't want to merge this (although version generally aren't valid idents, which helps finding such errors). There is the possibility to shift the rules to have `bar` for the constant and e.g. `%{bar}%` (or `$bar` ?) for the variable, but this doesn't really remove any pitfalls unless we propagate all the way to filters and commands. And that would be a very big change, which by itself may be confusing:
`build: [ make "install" ] { os = "linux" }` becomes `build: [ $make install ] { $os = linux }`

Might be easier to read, though I think in this context quotes aren't a bad idea.

My opinion is that the possible gain here would be far from warranting a change that big, given how confusing it would be for ppl already familiar with the current format.

Also note that:
- up to now, at the parsing level, opam 1.2.2 and 2.0 should still be compatible
- with this PR, it needs to be investigated, but the 1.2.2 parser might have trouble reading 2.0 files. Opam 1.2 files are still compatible with the new format though.
- with the above change added, the formats become really incompatible and 2.0 would need to discriminate between the file versions to understand the different possible meanings of `build: make`

AltGr:
Seeing the overwhelming feedback, it doesn't seem worth it to break the syntax :p

Closing the experiment.

PR #2721, opened 2016-10-17
Disallow parallel builts, and travis tweaks

AltGr:


AltGr:
Note: this ended up as a big rewrite of the Travis scripts, but I am quite happy about it: we now use the container infrastructure which is much faster, only depend on the opam binary release (1.2) for bootstrapping (no apt package for either opam or ocaml, no ppa, no XQuartz or homebrew), don't need root, and all the bootstrapping process is cached so that builds are much, much faster.

Additionally, the script is now the same (!) for OSX and Linux, and any new OCaml version that is packaged can be added with a line in `.travis.yml`.

PR #2735, opened 2016-10-27
Make 'make libinstall' follow the new multi-lib scheme

AltGr:
i.e. opam-core, opam-format etc. libs, as advertised by the new
corresponding opam packages.


AltGr:
Merged into #2734  (the two together are needed for travis+opam-rt to work correctly)

PR #2820, opened 2017-01-13
Expose `with_tags` in the OpamFile.OPAM signature

avsm:
What are the chances that the one field that I want to set using OpamFile.OPAM isn't exposed? Non-zero probability, it turns out...


AltGr:
Thanks!

Issue #2762, opened 2016-11-26
[Feature Request] Specify exact git commit of a dependency


RalfJung:
We are developing some rather fast-moving (as in, constantly breaking backwards compatibility) research projects and would like to sue opam to express and manage dependencies. The trouble is that generally, project B will need a *specific git commit* of project A. I hoped I could express that by writing in B's `opam` file:
```
depends: [
  "https://.../projectA.git#TheRightCommit"
]
```
However, that doesn't seem to work, the dependency is just ignored. Of course, for a local package I can simulate that by first pinning manually, but I'd like to have something that can automatically install the right versions of everything, for example for a CI build. This also has to work transitively; project A may have a similarly volatile dependency on project X, so in A's `opam` file I would like to give the right commit of `X`.

It would be great if opam could support pin-like notation in the dependency field, so that one can express a dependency on a particular git commit. (As another benefit, this would make it easy to depend on projects that are not available in any repository; one could just give their git URL. There is precedence for doing this e.g. in Rust's cargo.)

`opam config report`:
```
# OPAM config report
# opam-version    1.2.2 
# self-upgrade    no
# os              linux
# external-solver aspcud $in $out $criteria
# criteria        -count(removed),-notuptodate(request),-sum(request,version-lag),-count(down),-notuptodate(changed),-count(changed),-notuptodate(solution),-sum(solution,version-lag)
# jobs            4
# repositories    4* (http), 1 (version-controlled)
# pinned          1 (version), 2 (version control)
# current-switch  coq-8.5*
# last-update     2016-11-25 15:57
```

AltGr:
Thanks for the suggestion; indeed, managing development dependencies, and in general providing a way to share and reproduce consistent dev environments, is the last major concern that we are triving to solve before releasing opam 2.0. There are still open questions on how exactly to manage it, though, and input is welcome.

One issue with the suggested behaviour is that it supposes an open universe of packages, while opam relies on solving installation problems within a closed universe. What I mean by this is that opam supposes prior knowledge of the metadata of all packages (or at least their relationships, ie conflicts and dependencies) before attempting to solve the installation problem, and returning the best solution. The installability problem is NP-complete, and, AFAIK, package managers with a different policy don't handle conflicts or expressive dependencies.

[...]

PR #2822, opened 2017-01-13
Add prototype 'opam build' command

AltGr:
Provides tools for using a project directory from CWD.

Will auto-init and create a local switch unless specified otherwise.

Scans for local opam files and builds based on them.

Can install by copying the artefacts from the local switch to a given prefix.

Handling a local conf file that can give the specifics of how to create the
switch would be a nice addition.

### DIFF
--- a/shell/opam_completion.sh
+++ b/shell/opam_completion.sh
@@ -199,7 +199,7 @@ _opam()
       exec|e)
           if [ $COMP_CWORD -eq 2 ]; then compgen_opt+=(-c)
           else _opam_add_f _opam_flags "$cmd"; fi;;
-      lint)
+      lint|build)
           if [ $COMP_CWORD -eq 2 ]; then
               compgen_opt+=(-f -X '!*opam' -o plusdirs)
           else _opam_add_f _opam_flags "$cmd"; fi;;
--- a/src/client/opamMain.ml
+++ b/src/client/opamMain.ml
@@ -150,7 +150,7 @@ let init =
         $(b,global-variables:)."
   ] in
   let compiler =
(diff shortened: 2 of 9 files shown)


avsm:
This is exciting! Once merged, I can try include it in the CI and try it on the larger set of MirageOS opam packages.  Does it require that the `opam-version` be OPAM2, or can it transparently pin to the existing OPAM1-format files that all the repos have?

avsm:
It would be useful to be able to setup custom remotes and pins for this, as our master branches quite habitually require some customisation of the OPAM universe before things compile.  How about an `.opamrc` that executes the local build commands after the `init`? That could also be interpreted by CI systems for their setup as well.

AltGr:
Indeed. We already have [`.opamrc` files](https://opam.ocaml.org/doc/2.0/Manual.html#opamrc), but these are only useful for the original `opam init` command (they can define global variables, e.g. `sys-ocaml-version`, remotes, a formula to choose the default compiler, etc.) We also have `export` files that can be used to define a given package selection, pins,etc.

None of those will be automatically used by `opam build` at the moment, though.
- an export file with a certain pre-defined name could be used automatically to initialise the switch (possibly depending on e.g. OS). Think "lock files".
- a switch-init file format could be defined, similar to `opamrc`, but tailored for specific switch initialisation: allowing to define switch-local variables, pre-select repositories, choose compiler, etc.  There may be some overlap with the above `export` files, which could be included or selected using a formula.
- for depending on specific pins, see the proposal at #2762

Once properly defined, either wouldn't pose specific difficulties to implement, for interpretation when `opam build` is initially called and creates the local switch. However, what exactly should be done on further calls, if the files changed, isn't obvious and needs careful thought, and, maybe, careful handling.

PR #2827, opened 2017-01-17
PROPOSAL: add 'conflict-class:'

AltGr:
This might be the missing piece we needed most from `provides:`, but without the problems that come with the other ones

### DIFF
--- a/src/client/opamMain.ml
+++ b/src/client/opamMain.ml
@@ -322,6 +322,16 @@ let list =
     Arg.(value & opt (list atom) [] &
          info ~doc ~docs:selection_docs ~docv:"PACKAGES" ["required-by"])
   in
+  let conflicts_with =
+    let doc =
+      "List packages that have declared conflicts with at least one of the \
+       given list. This includes conflicts defined from the packages in the \
+       list, from the other package, or by a common $(b,conflict-class:) \
+       field."
+    in
+    Arg.(value & opt (list package_with_version) [] &
+         info ~doc ~docs:selection_docs ~docv:"PACKAGES" ["conflicts-with"])
+  in
   let coinstallable_with =
     let doc = "Only list packages that are compatible with all of $(docv)." in
     Arg.(value & opt (list package_with_version) [] &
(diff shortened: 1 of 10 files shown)


Drup:
Does this means we can get rid of the difference between `ocaml-base-compiler` and `ocaml-variants` ?

AltGr:
What difference do you mean ?
This would change the way conflicts are defined, i.e. all of `ocaml-system`, `ocaml-core-compiler` and `ocaml-variants` will need to define `conflict-class: "ocaml-core-compiler"`, instead of having the first two define conflicts with the two others, and the last not needing to define any conflict at all.

The big difference is that this solution scales if we increase the number of implementing packages, so this could open the way to having more specific packages instead of `ocaml-variants`, while at the moment all variants are in the same package, with the variant name encoded as the suffix of the version (which doesn't make choosing or conflicting with a given variant easy).

Of course, the virtual `ocaml` package would still need to list the implementors in its `depends:` field (but only once (per version)). Which, although a little cumbersome, may be best for security concerns anyway.

src/client/opamMain.ml
```diff
@@ -2688,6 +2688,287 @@ let clean =
         $switch $all_switches),
   term_info "clean" ~doc ~man
 
+(* BUILD *)
```
avsm: This PR includes `opam-build` -- is that intended?

AltGr: Indeed, it'll be rebased anyway, now that `opam build` is merged

Drup:
@AltGr So concretely, it means we can add a `metaocaml` separate package, and avoid burying it inside `ocaml-variants.4.04+metaocaml` by just adding the conflict-class (and the dependency in `ocaml`) ? That sounds great.

AltGr:
Exactly. OTOH, when adding a new compiler variant, don't forget to put it in the right `conflict-class:`.
A specific `ocaml-dev` compiler containing all the `+prXXX` versions could be nice as well, as these are not really "variants". It would have started being too cumbersome to get above three or four implementing packages without this.

PR #2846, opened 2017-02-08
Avoid name clash with next version of cmdliner.

dbuenzli:
where `Arg.conv` is introduced. It would be nice if this could make it in the subsequent `rc`s before 2.0.0.


AltGr:
Sure! Thanks.

PR #2876, opened 2017-03-08
Add `--ignore-constraints-on` to more easily test dependencies

### LINKED ISSUE #2873: Provide a way to bypass dependency constraints
Following a discussion with @mshinwell

Sometimes you want to temporarily bypass version constraints to test compatibility of packages, e.g. with different compiler version. A new option `--ignore-constraints-on <package>` would be useful for that purpose.

For example, `opam install foo --ignore-constraints-on ocaml` would try to install `foo` normally, but on a package universe where all dependencies towards `ocaml` have had their version constraints stripped, practically meaning that any currently installed version of `ocaml` will be accepted.

AltGr:
Closes #2873

### DIFF
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -112,6 +112,7 @@ let apply_global_options o =
     (* ?dryrun:bool *)
     (* ?fake:bool *)
     (* ?makecmd:string Lazy.t *)
+    (* ?ignore_constraints_on:name_set *)
     (* ?skip_dev_update:bool *)
     ?json_out:OpamStd.Option.Op.(o.json >>| function "" -> None | s -> Some s)
     (* ?root_is_ok:bool *)
@@ -146,15 +147,16 @@ type build_options = {
   skip_update   : bool;
   external_tags : string list;
   jobs          : int option;
+  ignore_constraints_on: name list option;
 }
 
 let create_build_options
     keep_build_dir reuse_build_dir inplace_build working_dir make no_checksums
     req_checksums build_test build_doc show dryrun skip_update external_tags
-    fake jobs = {
+    fake jobs ignore_constraints_on = {
   keep_build_dir; reuse_build_dir; inplace_build; working_dir; make;
   no_checksums; req_checksums; build_test; build_doc; show; dryrun;
-  skip_update; external_tags; fake; jobs;
+  skip_update; external_tags; fake; jobs; ignore_constraints_on;
 }
 
 let apply_build_options b =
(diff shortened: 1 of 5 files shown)


src/client/opamArg.ml
```diff
@@ -915,10 +920,17 @@ let build_options =
        actually performing them. \
        WARNING: This option is dangerous and likely to break your opam \
        environment. You probably want `--dry-run'. You've been warned." in
+  let ignore_constraints_on =
+    mk_opt ~section ["ignore-constraints-on"] "PACKAGES"
+      "Forces opam to ignore version constraints on all dependencies to the \
+       listed packages. This can be used to actually test compatibility, but \
```
dbuenzli: I would remove the `actually`.

src/client/opamArg.ml
```diff
@@ -915,10 +920,17 @@ let build_options =
...
+       expect builds to break when using this. Note that version constraints \
+       on optional dependencies and conflicts are unaffected."
```
dbuenzli: Is this a limitation of the current implementation or a more fundamental one ?

AltGr: Limitation of the current implementation ; but removing version constraints on conflicts would actually mean _less_ installability, which is not the purpose in the expected use-cases. I pondered removing the conflicts to the named package altogether (maybe only when they do have a version constraint ?), but it didn't seem worth the convoluted semantics.

As for removing version constraints on depexts, I am not sure this would be expected, or what purpose it would serve..

dbuenzli: > As for removing version constraints on depexts, I am not sure this would be expected, or what purpose it would serve..

If I have `A` that has a constrainted depext on `B` and want to install `A` without the constraint on `B` no ?

dbuenzli: That is something like `opam install A B --ignore-constraints-on B`

AltGr: err, I meant depopts, not depexts
`depopts: "foo" {>= "3"}` no longer means that the package conflicts with `foo<3` (as it kind of did in opam 1.1, with broken semantics). It merely means that the depopt is ignored if the version doesn't match. So ignoring the constraint would make the depext always honored. Which would only lead to more rebuilds...

dbuenzli: > So ignoring the constraint would make the depext always honored.

But that's precisely the scenario I highlighted above no ?

AltGr: Yes. The only difference would be, with `A` having a depopt on `B >= 2`, that `opam reinstall B --ignore-constraints-on B` would trigger a recompile of `A` even if `B` was `<2`.
I am not sure when or if this may be useful, but probably not in the same cases as ignoring constraints on real depends, which is why I decided not to include depopts.

Of course, if you have scenarios in mind that prove otherwise, I'd be glad to add it.

dbuenzli:
> The only difference would be, with A having a depopt on B >= 2, that opam reinstall B --ignore-constraints-on B would trigger a recompile of A even if B was <2.

That's what I would expect. Can't the proposed scheme lead to situations where a package was installed with a depopts, its depopt is reinstalled but the package is not recompiled and thus we get "makes inconsistant assumptions" errors when it's used ?

PR #2877, opened 2017-03-09
make cold: fix the bootstrap compiler to install the faster compiler

avsm:
Without this, `ocamlc.byte` is installed and the build fails. We might
as well use ocamlopt.opt as the build time isn't that much slower
and we get a more complete OCaml environment.


AltGr:
Thanks!

PR #2886, opened 2017-03-20
Don't let opam-lib leave empty dirs in /tmp

AltGr:


AltGr:
Note: this was a regression since beta2, which actually wiped logs a bit too eagerly on errors (could point you to a log file that no longer existed)

PR #2893, opened 2017-03-25
`opam install <dir>`

AltGr:
This allows to specify opam files or project directories to `opam install`,
superseding the `opam build` prototype.

There are still quite a few rough edges:
- `opam install` only: `upgrade`, `reinstall` and `remove` should be supported
  too (maybe `remove` should unpin as well ?)
- ~~Further calls, with a directory as argument, might be expected to sync, i.e.
  to UNpin packages that were pinned there but don't match anymore (renamed
  packages ?) ??~~ (done)
- ~~opam file format upgrade missing on pinning, the install re-does the update
  that was just done~~ (fixed)


hcarty:
👍 f038997 is my favorite!

PR #2906, opened 2017-04-14
Support for `opam upgrade|reinstall|remove <dir>`

AltGr:
Only affects packages that were pinned to the given target.
Note that specifying an opam file is not supported at the moment, while `opam install <opam-file>` is accepted.


AltGr:
NOTE: need to rebase upon #2904 once merged

PR #2912, opened 2017-04-23
Fix two glitches when removing failed installations

dra27:
I hit this testing ocamlfind on the Windows port. It should be possible to trigger the problem elsewhere by putting an incorrect filename in `ocamlfind.install` (for my particular test I expected the problem - it's to do with opam not knowing to install ocamlfind.exe instead of ocamlfind).

There seems to be a typo in `OpamDirTrack.revert` - it doesn't qualify the filename used when reading the checksum (so I get a stat error).

The other more serious problem is to do with the default directory used for `OpamAction.remove_package` - it assumes that the sources will have been put in the "remove" directory but if something fails during installation then the sources will be in the "build" directory. I have fixed this by adding an optional parameter but this should maybe be thought through so more - initially the uninstall worked because I happened to be in my root **OPAM** source directory so opam merrily ran its own configure script because it was in the current directory! Perhaps if there is no build directory, OPAM should change to a temporary empty directory - or raise an error if it's asked to run a relative-path command? For Windows, this behaviour is very important because `PATH` implicitly begins with `.:` (i.e. an executable in the current directory always takes precedence over `PATH` itself, even if you don't prefix `./`).

### DIFF
--- a/src/client/opamAction.ml
+++ b/src/client/opamAction.ml
@@ -409,7 +409,7 @@ let noop_remove_package t nv =
 
 (* Remove a given package *)
 let remove_package_aux
-    t ?(silent=false) ?changes ?force nv =
+    t ?(silent=false) ?changes ?force ?build_dir nv =
   log "Removing %a" (slog OpamPackage.to_string) nv;
   let name = nv.name in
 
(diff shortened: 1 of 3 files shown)


AltGr:
Well caught, for both issues! For the second one, indeed, seems I forgot the failed-install case when adding "remove directories". Creating an empty temp directory for running build/remove instructions even when there is no source could be a good option, yes.

AltGr:
Thanks

dra27:
I've put a note on my (looong) opam todo list - though I wonder if the potential security implications are enough to warrant tracking that in an issue (there may be other places worth checking?)

AltGr:
Well, on Linux, we can use namespaces (see `shell/wrap-*.sh`) to protect against writes outside of the build/remove dir during build/remove. That's good for CI, but doesn't help with Windows or OSX of course.

avsm:
I wonder if we can use Windows Integrity Levels to give us more safety there -- https://msdn.microsoft.com/en-us/library/bb625957.aspx

avsm:
The Chromium sandbox docs are an interesting read for a cross-platform approach (but not directly usable since they dont do sandboxing directly) https://chromium.googlesource.com/chromium/src/+/master/docs/design/sandbox.md

PR #2915, opened 2017-04-26
Logging improvements

dra27:
Apropos of yesterday's Slack discussion, a few changes to the handling of patch files during opam update.

### DIFF
--- a/src/repository/opamRepository.ml
+++ b/src/repository/opamRepository.ml
@@ -396,10 +396,12 @@ let apply_repo_update repo = function
     Done ()
   | Update_err _ -> assert false
 
-let cleanup_repo_update = function
-  | Update_full d -> OpamFilename.rmdir d
-  | Update_patch f -> OpamFilename.remove f
-  | _ -> ()
+let cleanup_repo_update upd =
+  if not (OpamConsole.debug ()) then
+    match upd with
+    | Update_full d -> OpamFilename.rmdir d
+    | Update_patch f -> OpamFilename.remove f
+    | _ -> ()
 
 let update repo =
   log "update %a" (slog OpamRepositoryBackend.to_string) repo;
--- a/src/core/opamProcess.ml
+++ b/src/core/opamProcess.ml
@@ -484,7 +484,11 @@ let is_failure r = r.r_code <> 0 || r.r_signal <> None
 let is_success r = not (is_failure r)
 
 let safe_unlink f =
-  try Unix.unlink f with Unix.Unix_error _ -> ()
+  try
+    log ~level:2 "safe_unlink: %s" f;
+    Unix.unlink f
+  with Unix.Unix_error _ ->
+    log ~level:2 "safe_unlink: %s (FAILED)" f
 
 let cleanup ?(force=false) r =
   if force || (not (OpamConsole.debug ()) && is_success r) then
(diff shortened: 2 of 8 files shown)


src/repository/opamRepository.ml
```diff
@@ -396,10 +396,12 @@ let apply_repo_update repo = function
...
-let cleanup_repo_update = function
-  | Update_full d -> OpamFilename.rmdir d
-  | Update_patch f -> OpamFilename.remove f
-  | _ -> ()
+let cleanup_repo_update upd =
+  if not (OpamConsole.debug ()) then
+    match upd with
+    | Update_full d -> OpamFilename.rmdir d
+    | Update_patch f -> OpamFilename.remove f
+    | _ -> ()
```
AltGr: Did you make sure the stale file/directory can't cause problems on further calls ?

AltGr:
Looks good, I just inlined a minor question. Thanks!

src/repository/opamRepository.ml
dra27: I wouldn't have thought the file case can cause problems, because it's generated using a temporary filename - I've not tested the directory case (I guess a file/rsync repo uses that?)

AltGr: indeed; I'm mostly concerned about git, but it should be OK.

dra27: It was a git repo I was testing this against - I'll do some further testing with several updates in a row.

One related problem I did hit with the patch failing (because of the CRLF problems) was that it leaves .orig and .rej files in the git repo clone - I don't know whether it would be worth the backend always doing a `git clean -dfx`?

AltGr: Since that shouldn't happen anyway (the patch file is generated and used from within the opam engine), I don't think it's worth worrying about them.

dra27: It really does happen! It's a problem I haven't yet properly decided how to solve. At the moment, I've done it by making `OpamSystem.patch` prefer to use `git apply` rather than `patch`

AltGr: using `git apply` could actually be a good solution, if it's more reliable. At the moment we require `patch`, but not `git`, though.
Part of the plan behind the new update feature was to use a patch-file generator written in OCaml, common to Conex and opam. Since patches can be supplied in packages, though, we need to support the number of different patch formats out there, and it's wiser to rely on an external `patch` command for that.

AltGr: I trust it does happen at the moment, but that should be fixed before release shouldn't it ? :)

dra27: Indeed - but lots of things shouldn't generally happen :smile: I'm just thinking of terms of things which leave ~/.opam in a hard-to-recover state!

My solution with git apply should solve this problem very well - it falls back to patch if git isn't found (so we don't depend on git) and the most likely situation where this could happen is for patches generated when updating a git repo (because git will generate an LF-patch even for a CRLF repo) - but in that case the user must have git, or we couldn't have the git remote!

dra27:
There was an obvious thing we missed with talking about testing this - if the patch applies correctly, opam will still remove the patch file, even with `--debug` - it only keeps them if debug mode is enabled *and* something went wrong.

I don't think there's anything else which I need to do to this GPR, therefore?

AltGr:
Yup, thanks!

PR #2921, opened 2017-04-28
Add ? operator to filter syntax

dra27:
See also https://github.com/ocaml/opam-file-format/pull/4 - this PR should have a commit updating `src_ext` before it is merged.

The `?` operator evaluates to false if the expression is undefined (i.e. contains undefined variables) and true in all other instances.

This operator removes the need for evil uses of `"%{variable}%"` (which expands to an empty string if variable is undefined). In particular, for the `ocaml-system` package, it allows a slightly less unclear `available` field:

```
available:
  sys-ocaml-version = "4.02.2" &
  !(?switch-arch & switch-arch != sys-ocaml-arch) &
  !(?switch-cc & switch-cc != sys-ocaml-cc) &
  !(?switch-libc & switch-libc != sys-ocaml-libc)
```

### DIFF
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -12,6 +12,7 @@ install-bootstrap () {
     opam init --root=$OPAMBSROOT --yes --no-setup --compiler=$OCAML_VERSION
     eval $(opam config env --root=$OPAMBSROOT)
     if [ "$OPAM_TEST" = "1" ]; then
+        opam pin add opam-file-format "git://github.com/ocaml/opam-file-format.git" --no-action --yes
         opam install ocamlfind lwt.2.5.2 cohttp.0.20.2 ssl cmdliner dose3 jsonm opam-file-format --yes
         # Allow use of ocamlfind packages in ~/local/lib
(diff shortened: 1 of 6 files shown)


dra27:
The two temporary commits are obviously just that, but note that https://github.com/ocaml/opam/pull/2921/commits/6b2a19fdf808fbf3d10cfbe9e6f4f8a15940a8dc includes changes which should be made, since lib-ext is presently using an older release of opam-file-format

AltGr:
Thanks! We'll also need to update the doc (https://github.com/ocaml/opam/blob/master/doc/pages/Manual.md#filters and general syntax above)

dra27:
Docs updated, and also the temporary lib-ext pointer changed.

Is there anything else I need to do to this one?

AltGr:
Nope, seems all OK to me. The temporary commits actually don't need to be temporary anymore (it's fine to require the master version of opam-file-format for CI !)
Just waiting for Travis to complete

AltGr:
Thanks!

PR #2922, opened 2017-04-28
Some more customisation for `opam init`; some var resolving fixes in `depends`

AltGr:
Note: stacking the config-files is done per field: the latest
definition for any given field overrides the previous ones.

Some might want to be able to merge e.g. repository definitions from
several config files, though ? That would be more complex... but in
practice, if you redefine e.g. `eval-variables:`, you need to include
the def for `sys-ocaml-version` or the default repository would be
broken.


AltGr:
In that case, could we simply let the user supply the expected hash, interactively or through the command-line, without printing it out ?

PR #2923, opened 2017-04-29
Fix renaming of test and doc

dra27:
beta3 renames the `test` and `doc` predicates to `with-test` and `with-doc`, but opam admin upgrade doesn't convert the predicates from old repositories.

@altgr - I patched this just because the warnings about wrong predicates were clogging my debug logs if the state was recomputed. If it saves you 10 seconds' work, then here you go!

### DIFF
--- a/src/state/opamFormatUpgrade.ml
+++ b/src/state/opamFormatUpgrade.ml
@@ -186,6 +186,8 @@ let opam_file_from_1_2_to_2_0 ?filename opam =
        | "ocaml-native" -> mkvar "native"
        | "ocaml-native-tools" -> mkvar "native-tools"
        | "ocaml-native-dynlink" -> mkvar "native-dynlink"
+       | "test" -> OpamVariable.Full.global (OpamVariable.of_string "with-test")
+       | "doc" -> OpamVariable.Full.global (OpamVariable.of_string "with-doc")
        | _ -> v)
     | _ -> v
   in


AltGr:
I believe `doc` shouldn't be rewritten, as the variable was used to point to the `<prefix>/doc` directory, not as a binary flag ; as for `test`, it didn't exist in 1.2. The `build`, `test`, `doc`, `dev` flags that appeared in dependencies only were handled specifically in 1.2, not as variables, so they should be rewritten correctly.

There is a part that I overlooked, though, and that must be causing your errors, and that is migration from an earlier 2.0 format (up to `2.0.0~beta2`, which is still the format `opam.ocaml.org/2.0` currently serves), which did use the `test` and `doc` variables. If needed, the solution would be to do the rewrite you propose, but only in the `depends` and `depopts` filters.

dra27:
That's possibly what happened - I *think* that I fixed this before starting to use opam admin upgrade on my own repository (so I would have been using the one services opam.ocaml.org). This would seem to be low priority, therefore - once opam.ocaml.org is using a newer converter, everything will be fine without this, I think?

AltGr:
yes, that was my impression; in any case when we release beta3 and switch opam-repository/2.0 to the new format the users of beta2 will get the warnings. I generally try to keep incremental upgrades working, but sometimes have to make a compromise with the complexity involved and number of users inpacted. I apologise to them in advance :)

dra27:
In which case, I wonder if the code for doing the format upgrade is already complicated enough without having a case which won't be needed for very long?

AltGr:
indeed... so OK to close this ?

dra27:
Yes!

PR #2926, opened 2017-05-03
Fix lib-ext build of extlib

dra27:
d7bf17fffe354ea0ca72b385ac500a016d78e4b9 reverted a change in c87c77cfec457e3c1410bce5848326910c9228c4 breaking opam-admin.top when compiled
using lib-ext.


AltGr:
Thanks!

PR #2927, opened 2017-05-03
%<...>% path expansion notation

dra27:
There are many instances in packaging of the form `%{some-directory-variable}%/more/path/components` which result in an amateur/confused-looking mixed back/forward-slash path on Windows.

I propose adding the notation `%<subst>%` such that *subst* is expanded with `/` converted to the path separator for the OS. `%{...}%` may be nested inside `%<...>%` notation, allowing variables to be used within a path expansion. Thus `%<%{lib}%/stublibs>%` on Windows is expanded to the value of
`lib` with `\\stublibs` appended.

### DIFF
--- a/src/format/opamFilter.ml
+++ b/src/format/opamFilter.ml
@@ -69,19 +69,22 @@ let rec map_up f = function
   | FUndef x -> f (FUndef (map_up f x))
   | (FBool _ | FString _ | FIdent _) as flt -> f flt
 
-(* ["%%"], ["%{xxx}%"], or ["%{xxx"] if unclosed *)
-let string_interp_regex =
+(* ["%%"], ["%{xxx}%"], or ["%{xxx"] if unclosed.
+ Path expansion versions use ['<'] and ['>'] instead of ['{'] and ['}']. *)
+let (string_interp_regex, path_interp_regex) =
   let open Re in
-  let notclose =
+  let notclose c =
     rep (alt [
-        diff notnl (set "}");
-        seq [char '}'; alt [diff notnl (set "%"); stop] ]
+        diff notnl (set (String.make 1 c));
+        seq [char c; alt [diff notnl (set "%"); stop] ]
       ])
   in
-  compile (alt [
-      str "%%";
-      seq [str "%{"; group (greedy notclose); opt (group (str "}%"))];
-    ])
+  let string_interp =
+    [str "%%";
+    seq [str "%{"; group (greedy (notclose '}')); opt (group (str "}%"))]]
+  in
+  compile (alt string_interp),
+  compile (alt ((seq [str "%<"; group (greedy (notclose '>')); opt (group (str ">%"))])::string_interp))
 
 let escape_expansions =
   Re.replace_string Re.(compile @@ char '%') ~by:"%%"
@@ -204,9 +207,9 @@ let resolve_ident ?(no_undef_expand=false) env fident =
      | Some (S s) -> FString s
      | None -> FUndef (FIdent fident))
 
-(* Resolves ["%{x}%"] string interpolations *)
-let expand_string ?(partial=false) ?default env text =
-  let default fident = match default, partial with
+(* Resolves ["%{x}%"] and ["%<x>%"] string interpolations *)
+let rec expand_string_aux ?(regex=path_interp_regex) ?(partial=false) ?default env text =
+  let default_f fident = match default, partial with
     | None, false -> None
     | Some df, false -> Some (df fident)
     | None, true -> Some (Printf.sprintf "%%{%s}%%" fident)
@@ -222,15 +225,21 @@ let expand_string ?(partial=false) ?default env text =
   let f g =
     let str = Re.Group.get g 0 in
     if str = "%%" then (if partial then "%%" else "%")
-    else if not (OpamStd.String.ends_with ~suffix:"}%" str) then
-      (log "ERR: Unclosed variable replacement in %S\n" str;
+    else if not (OpamStd.String.ends_with ~suffix:(if str.[1] = '{' then "}%" else ">%") str) then
+      (log "ERR: Unclosed %s in %S\n" (if str.[1] = '{' then "variable replacement" else "path expansion") str;
        str)
     else
-    let fident = String.sub str 2 (String.length str - 4) in
-    resolve_ident ~no_undef_expand:partial env (filter_ident_of_string fident)
-    |> value_string ?default:(default fident)
+      if str.[1] = '<' then
+        expand_string_aux ~regex:string_interp_regex ~partial ?default env (String.sub str 2 (String.length str - 4))
+        |> String.map (function '/' -> Filename.dir_sep.[0] | c -> c)
+      else
+        let fident = String.sub str 2 (String.length str - 4) in
+        resolve_ident ~no_undef_expand:partial env (filter_ident_of_string fident)
+        |> value_string ?default:(default_f fident)
   in
-  Re.replace string_interp_regex ~f text
+  Re.replace regex ~f text
+
+let expand_string = expand_string_aux ~regex:path_interp_regex
 
 let unclosed_expansions text =
   let re =


dra27:
This will obviously also require a docs update, but I'll do that after the notation is agreed!

src/format/opamFilter.ml
```diff
@@ -204,9 +207,9 @@ let resolve_ident ?(no_undef_expand=false) env fident =
      | Some (S s) -> FString s
      | None -> FUndef (FIdent fident))
 
-(* Resolves ["%{x}%"] string interpolations *)
-let expand_string ?(partial=false) ?default env text =
-  let default fident = match default, partial with
+(* Resolves ["%{x}%"] and ["%<x>%"] string interpolations *)
+let rec expand_string_aux ?(regex=path_interp_regex) ?(partial=false) ?default env text =
```
dra27: Note that the use of `~regex` means that this only introduces one recursive call - expansion of `%{...}%` within `%<...>%`. It **doesn't** add recursive expansion of variables by accident!

dbuenzli:
This is going to make things unreadable and people will always forget to do it. Why doesn't opam simply converts path variables to forward slashes on all platforms ? Consumers of these paths can then do the right thing according to the platform (e.g. by using a suitable path library).

AltGr:
I share some of @dbuenzli's concern, in particular about forgetting to add the delimiters. I am not sure it's wise to limit the expansion to forward slashes when it will be consumed by Windows commands, though, but I don't really have a good grasp of how this ends up being handled on Windows.

Since these strings are used as arguments to arbitrary commands, we have no way of knowing wether they are paths or not, so the only two solutions I can imagine (if dropping backward slashes altogether, like @dbuenzli suggests, is not a viable solution) are:
- special syntax, as proposed. Something more light-weight could be possible, e.g. a whole-string qualifier rather than an expansion (I don't remember the details well but we've discussed it already). It could make things more readable, but remains the "forgetting" issue
- make the `/` character special, meaning "the path separator", and rewrite it in _all_ strings, unless it's escaped as `\/`. This may not be as crazy as it sounds, as I don't believe we have many uses of that char outside of paths in practice. Other uses would be in informative strings were it wouldn't break. Of course, this opens a different treatment on Unix and Windows that _could_ lead to stuff breaking...

dra27:
I am virtually certain both from a presentational and technical perspective that forcing forward-slashes on Windows is a mistake (even if a mistake made by a lot of other "cross-platform" projects) and we should be sticking to the actual convention. As a tiny bit of background, while there are various layers of compatibility, the separator at the kernel level is backslash *only* (it is API functions which do the conversion, but that's at the subsystem level). More relevantly, there are quite a few user-level places where forward-slashes do not work (and these are not particularly well-defined places - i.e. there are shell commands where some forward-slashes work and others do not). Most importantly, there are *no* places where back-slashes will not work and you have to use forward-slashes...

Some further thoughts on both comments:
 - I hope that `%<...>%` is preferred to my original prototype, which was `%{<$lib$/stublibs}%` :smile:
 - I also tried @altgr's suggestion (ages ago) of a special literal syntax for paths, though I gave up because the impact on opam's own type system was vast and it was also hard to ensure that a `path` type didn't leak back to a `string` (`OpamFilename.t` is now even more pervasive than it was then, so this is probably less of a problem). However, while this made things neater for some cases, there still remained what happened if you wanted, say, to embed a path in a call to a Microsoft C compiler invocation, say, `/FeC:\Users\DRA\Documents\.opam\default\.opam-switch\build\package\foo.exe` where the `/Fe` must be part of the switch (i.e. you have to embed the path in a string, it can't exist on its own). That example is of course not commonplace, but that's not the point.
 - The suggestion of making `/` special could be very interesting and now would be the time to do it - the `/` prefix is for many Windows commands the equivalent of `-` or `--` (Microsoft's compiler tools almost universally accept either `/` or `-`, but other Windows commands are not as liberal) - but if this were changed now, then it would be Windows which would be affected by the behaviour and it would be Windows where breakages would be likely to take place (e.g. `cmd /c rd /s/q C:\some-temp-directory` fails noisily if you forget to escape the forward-slashes). Also, we should already be suspicious of `/` in a prefix position, as absolute paths should not be being embedded in opam files.
 - As regards forgetting to use the `%<...>%` notation - a corollary of Louis's suggestion is that uses of `/` outside `%<...>%` could reasonably be reported by `opam lint` as probably wrong...

dbuenzli:
> I am virtually certain both from a presentational and technical perspective that forcing forward-slashes on Windows is a mistake (even if a mistake made by a lot of other "cross-platform" projects) and we should be sticking to the actual convention. As a tiny bit of background, while there are various layers of compatibility, the separator at the kernel level is backslash only (it is API functions which do the conversion, but that's at the subsystem level).

This misses my point. I think that your proposal introduces a complexity that we could eschew. More specifically here's what I think should be done.

1. `opam` peforms path variable substitution in `opam` files using forward slashes, the build 
systems should behave correctly and `opam` file writers should use forward slashes in their constants. It should be the duty of their build system to render and use them according to the platform conventions.
2.  `opam` when otherwise provided with path information should always accept either the platform convention or forward slashes.
3. `opam` when otherwise queried for path information should always render it according to the platform conventions. 

Now you may see 1. as already being a heresy but I think it's will make our life simpler without adding more idiosyncrasy to the already confusing `opam` language.

dra27:
For me, it escapes heresy by not requiring or displaying the forward-slashes! Actually I think we're in closer agreement than I possibly realised, but I'm not convinced that a key part of (1) is presently doable (@altgr may have a different opinion). The problem is that there is no actual notion of a "path variable" (at present) - that's partly what this syntax introduces, though it's of course not the only way. Storing forward-slashes internally also shifts considerable complexity back into opam itself.

While it would still add complexity to the language, @altgr's suggestion of a whole string qualifier would at least get rid of the commonplace `%<%{` prefix (there would be a small, but probably insignificant, loss of expressivity with that change).

Overall, I'm afraid I see as requirement what you see as idiosyncrasy. It's also potentially a big (though not insurmountable) ask to expect all build systems (or, in general, commands being executed) to convert the forward slashes (shell invocations, where necessary, become harder - there are build systems, such as nmake, which cannot cope with them.

dbuenzli:
> For me, it escapes heresy by not requiring or displaying the forward-slashes!

The log of invocations would still be heretical. 

> The problem is that there is no actual notion of a "path variable" (at present)

Aren't the `%{bin}%`, `%{lib}%`, etc. variables typed internally ? 

> Storing forward-slashes internally also shifts considerable complexity back into opam itself.

You don't store them forwardly. When you perform substitution in the `build:` section you force rendering of the path to string with the forward slash convention. But of course you need the variables to be typed to know on which you have to do this.

AltGr:
> Aren't the `%{bin}%`, `%{lib}%`, etc. variables typed internally ? 

No, the variables are only booleans or strings at the moment. Besides, that could trigger changes within the variable expansion, but not for the rest of the string, e.g. in the right part of `"%{lib}%/foobar/"`.

I still only see "forward slashes everywhere" or "'/' is a special char meaning <os-specific-path-separator> everywhere" as solutions. The major pitfall of the second solution is that it still means '/' on most OSes, so people won't notice any mistake on their part...

dbuenzli:
> I still only see "forward slashes everywhere" or "'/' is a special char meaning everywhere" as solutions.

I dislike both, especially the second. Isn't `/` anyways forbidden as variable names ? We could then simply have the notation `%{bin/foo/bar}%` do the path translations, i.e. the first segment defines a variable the rest after the first `/` is an arbitrary string on which path translations are performed.

If we want that to be more obvious we could use any other character that is not allowed in variable names, e.g. `|`, `%{bin|foo/bar}%` or even a wink to windows user with `\`, `%{bin\foo/bar}%`.

AltGr:
@dbuenzli I thought "forward slashes everywhere" was what you initially suggested ?
`%{bin/foo}%` is indeed lighter-weight than the original suggestion; it's also closer to @dra27's initial prototype, which used `%{<$lib$/stublibs}%` !
It's less expressive though: what about literal paths that _don't_ use variables ? `%{/foo}%` or `%{./foo/bar}%` ? It's weird and confusing...

dbuenzli:
> @dbuenzli I thought "forward slashes everywhere" was what you initially suggested ?

No I suggested 1. + 2. + 3. Now since it seems you don't have the information internally about what is a path and what is not it seems that if we have 1. we cannot have 3. which I see an ux issue.

AltGr:
Note: we are now attempting to lock the opam file-format (https://github.com/ocaml/opam-file-format/) to allow for tighter integration with dune... so we should consider this shortly if something is to be done.
Cc. @dra27

PR #2935, opened 2017-05-13
Various tweaks

dra27:
A handful of hopefully non-controversial changes.

The .gitattributes file inclusion here is a lot easier than the OCaml one was, as there don't appear to have been any erroneous CRLF commits.


AltGr:
Looks all good, thanks a lot !

dra27:
Thanks! I'll rebase the others later on

Issue #2867, opened 2017-02-28
make clean


UnixJunkie:
Makefile:178: recipe for target 'fastclean' failed
make: *** [fastclean] Error 127
pwd
~/src/opam-2.0.0-beta2

PR #2936, opened 2017-05-13
Support make distclean at toplevel

dra27:
Patch Makefile to support total clean of source tree via make distclean and ensure that it can be run twice (so don't attempt to regenerate Makefile.config)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,6 @@
+ifeq ($(findstring clean,$(MAKECMDGOALS)),)
 -include Makefile.config
+endif
 
 all: opam-lib opam opam-installer
 	@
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -34,13 +40,14 @@ download-ext:
 clean-ext:
 	$(MAKE) -C src_ext distclean
 
-clean: fastclean
```
dra27: But for this change, I would have put this in #2935. The dependency on fastclean introduces a dependency on ocp-build... this isn't necessary the best way of handling that optional dependency! Does it *have* to be here and, if so, could ocp-build detection be put in configure instead?

AltGr: It's written so as to not fail if ocp-build is not present… but yes, that's probably not the best approach.
`rm -rf _obuild/` (instead of `ocp-build -clean`) should actually be enough.

dra27: Actually, I think I originally made this change before you committed e141c2e to fix #2867.

It's up to you if you'd like this changed to `rm -rf _obuild/` instead but for me, the underlying is fine (because it works!)

dra27: Oops - you've merged before I had a chance to change this... you want to push `clean: fastclean` back...

PR #2937, opened 2017-05-13
Developer tweaks

dra27:
This PR adds a few (hopefully) useful features for developing opam:

 - `make -C src_ext cache-archives` fetches all the tarballs for lib-ext and stores them in `src_ext/archives/` and these cached copies are used for future builds when running `make lib-ext` (the build system will download any file which is missing from the cache, but it's only updated by running `make -C src_ext cache-archive` again)
 - Debugging output is removed from bootstrap-ocaml.sh, so that actual errors can be seen...
 - A developer mode is added to allow builds of opam with special options. So far, only `DEVELOPER_CACHE` has been added. This option allows a directory to be specified where opam caches everything downloaded through the `OpamDownload` module. This sits above opam's own caching mechanisms meaning that `opam init` itself can use the cache (if downloading from an http repository, this completely breaks opam update, as the web server will never be queried again - this is considered a feature - you obviously develop using a local git clone for opam-repository, right?!). This mode is exceptionally useful for re-running `opam admin upgrade` when you lost your hashes cache...

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,6 +41,13 @@ it as a second step.
 If you are developing OPAM, you may enable developer features by including the
 `--enable-developer-mode` parameter with `./configure`.
 
+Developer features are controlled by setting additional environment variables:
+
+* `DEVELOPER_CACHE` - specifies the location of a directory to use for storing all
+  downloaded files. This directory must exist or OPAM will ignore the setting. See
+  the code for really_download in src/repository/opamDownload.ml for more details on
+  how this feature operates.
+
 ## Compiling without OCaml
 
 `make cold` is provided as a facility to compile OCaml, then bootstrap opam.
--- a/configure
+++ b/configure
@@ -608,6 +608,7 @@ EXE
 WIN32
 CONF_OCAMLFLAGS
 MCCS_DISABLED
+DEVELOPER_CACHE
 DEVELOPER
 OBJEXT
 EXEEXT
@@ -688,7 +689,8 @@ CC
 CFLAGS
 LDFLAGS
 LIBS
-CPPFLAGS'
+CPPFLAGS
+DEVELOPER_CACHE'
 
 
 # Initialize some variables set by options.
@@ -1335,6 +1337,8 @@ Some influential environment variables:
   LIBS        libraries to pass to the linker, e.g. -l<library>
   CPPFLAGS    (Objective) C/C++ preprocessor flags, e.g. -I<include dir> if
               you have headers in a nonstandard directory <include dir>
+  DEVELOPER_CACHE
+              Developer file cache location
 
 Use these variables to override the choices made by `configure' or to help
 it to find libraries and programs with nonstandard names/locations.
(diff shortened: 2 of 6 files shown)


dra27:
Tiny tweak to the "Option to preserve src_ext tarballs" (the archives target needs to be `.PHONY` because it may now exist as a real cache directory). Also rebased.

AltGr:
Thanks!

> make -C src_ext cache-archives fetches all the tarballs for lib-ext and stores them in src_ext/archives/ and these cached copies are used for future builds when running make lib-ext (the build system will download any file which is missing from the cache, but it's only updated by running make -C src_ext cache-archive again)

Why was this needed besides `make -C src_ext archives` ? I just want to make sure we keep things consistent with the way the `opam-full-xxx` archives that we distribute on GH are generated.

> This mode is exceptionally useful for re-running opam admin upgrade when you lost your hashes cache...

I was considering hosting a simple URL→md5 mapping in a file somewhere, that `opam admin upgrade` could hit upon to avoid downloading all the archives. That would avoid unnecessary load on all containerised upgrades too.

dra27:
The `cache-archives` target is optional - it just allows you to download all the src_ext tarballs to a place where they definitely won't be nuked by `make clean` or `make distclean` ... it's a bit frustrating at 30000ft to clear those away for a test and then find you can't build lib-ext again :smile: (admittedly, 2 years later, it is less problematic because there is good WiFi at 30000ft...). `make archives` will then cp those cached files rather than downloading again.

AltGr:
Oh, alright, sounds OK then.

AltGr:
I'm still a bit unsure about the another layer of optional caching through developer options. The cache for `opam admin upgrade` is actually stored, on purpose, outside of the opam root (`~/.cache/opam-compilers-to-packages/url-hashes`), so shouldn't get wiped too easily ; at least, the two caches should keep the same format (i.e. `URL<TAB>HASH` rather than the opposite ;), and files in a `hashalgo/hash2letters/hash` hierarchy) -- which would allow to copy (link ?) to/from `~/.opam/download-cache`.

Another possibility for adding caching to all uses of `OpamDownload` would be to provide a specific [`download-command`](http://opam.ocaml.org/doc/2.0/Manual.html#configfield-download-command) that includes caching.

dra27:
Both good suggestions - if I split this into two, would it be possible for 7c65dbb and 2e9e183 to be merged, and I'll work on the download caching a bit later?

AltGr:
yes!

dra27:
This is parked for now - only e40a24f is part of the PR, it's just been based on #2956.

AltGr:
Some `warn_errors` hitting: https://travis-ci.org/ocaml/opam/jobs/239054235

AltGr:
Unless I'm mistaken, this is outdated and has been partially merged in other PRs already. Thanks!

dra27:
I haven't (yet) addressed the changes required, but what was there before is now rebased so this looks a little less abandoned!

PR #2938, opened 2017-05-13
lib-ext improvements

dra27:
The major change here is to support `make -j lib-ext`. There are two minor changes which affect future Windows work:

 - A fallback to md5sum is added for verifying checksums in src_ext/Makefile (this makes life simpler, e.g. when downloading the cache-archives in #2937)
 - Various fixes to building native lib-ext using the native ports (don't get too excited, though - the commits necessary to configure opam for native Windows are in another PR...)

Note that for parallelisation to work, the lockfile command is required (open to suggestions for other mechanisms - that was the best one I came up with a couple of years ago in Cygwin).


AltGr:
Thanks!

PR #2939, opened 2017-05-13
Fix configuration and build system for native Windows

dra27:
This PR has some duplication with #2938 over the `$(EXE)` and `$(WIN32)` variables.

This PR corrects the build system to operate in a preconfigured Cygwin set-up for a native OCaml compiler (at present, autodetection of the Microsoft C Compiler is not merged - as for OCaml, Cygwin must be started from within a Microsoft Developer Tools prompt or PATH, INCLUDE and LIB must be correctly set). This PR fixes various things and when combined with #2938 should allow opam to be built using either mingw or msvc via the lib-ext route (note: the compiled opam will not work). This PR changes the following:

 - Handling is included in configure to detect the mingw compilers, as configure now tests the C compiler used by the OCaml compiler. Special handling is included to detect the AR command which OCamlMakefile insists on using directly.
 - Various fixes to the venerable, yet hardly used on Windows, OCamlMakefile
 - ocamldep on virtually all versions (I think possibly 4.05 is changing the behaviour) requires the -slash argument on Windows or the generated dependencies contain syntax errors for projects like opam with subdirectories.
 - A workaround somewhat similar to [msvs-tools](https://github.com/metastack/msvs-tools)'s `msvs-promote-path` is used to allow the Microsoft Linker (link.exe) to be called even when Cygwin's `link` appears before it in `PATH`
 - A better default `PREFIX` is selected for Windows-building (I got tired of accidentally finding C:\usr had been created...)


dra27:
d2ff690 is broken - that use of `AC_DEFAULT_PREFIX` is clearly not what I meant!

dra27:
That's hopefully a bit more like it - also has the benefit of displaying the correct default prefix in `./configure --help`

OCamlMakefile
```diff
@@ -112,7 +112,7 @@ endif
 
 ####################  variables depending on your OCaml-installation
 
-SYSTEM := $(shell ocamlc -config 2>/dev/null | grep system | sed 's/system: //')
+SYSTEM := $(shell ocamlc -config 2>/dev/null | grep "^system:" | sed 's/system: //')
```
AltGr: I prefer a single `| sed -n 's/^system: ?//p` pipe (but this is fine)

AltGr:
Thanks!

AltGr:
Ok to merge once the conflicts are fixed!

OCamlMakefile
dra27: My way is sloppy - I've changed it!

dra27:
Fixed that pipe, and also 3 others!

dra27:
Do let this run through CI, therefore, just in case!

AltGr:
gcc seems to be getting an option it doesn't like (`error: unrecognized command line option ‘-NODEFAULTLIB:LIBC’`), both on Linux and Windows (well clang on OSX has the same complaint too).

dra27:
Oops - that's why we have CI! This should fix it - I'd put `/e` instead of `/p` in one of the sed expressions...

dra27:
Can this one go in too, please?

AltGr:
Yes, I had to re-run the AppVeyor builds but they seem OK now.

PR #2940, opened 2017-05-13
Various build process fixes for native Windows

dra27:
Last tranche (for now) from my windows-build branch. These are most of the less-invasive patches for Windows support. Combined with the rest of today's PRs, this gets the build system to the point that it can correctly compile and install opam using opam-installer (however, that opam is not capable of running opam init at this stage).


dra27:
Small correction to use `String.uppercase_ascii` instead of `String.uppercase` and rebased.

AltGr:
Thanks for taking the time to do all this! Only complaint I can find is a couple overly long lines ;)

dra27:
Oops :blush: - I don't think these were as heinous!

Other PRs now all rebased - the inclusion of 352a51a means that the duplication of the `EXE` variable now disappears.

PR #2941, opened 2017-05-14
Update Ubuntu install instructions

dra27:
Update the install documentation for Ubuntu to make it clear that opam is available via apt for "recent" releases.

Also remove the paragraph about broken releases, since the two affect releases are waaaaay out of support.

### DIFF
--- a/doc/pages/Install.md
+++ b/doc/pages/Install.md
@@ -141,12 +141,16 @@ for Opam usage.
 
 #### Ubuntu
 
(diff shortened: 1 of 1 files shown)


AltGr:
Thanks. This is indeed well behind!
Note: should PR to the 1.3 branch to affect the current version of the site
(https://opam.ocaml.org/doc vs https://opam.ocaml.org/doc/2.0)

PR #2944, opened 2017-05-18
Add prototype bundle generator (in bash)

AltGr:


UnixJunkie:
@shindere maybe you can try this with coccinelle. That should be a harsh test given all its dependencies.

shindere:
Cc @thierry-martinez this may interest you, too.

shindere:
And thanks a lot for having let me know about that!

Issue #2158, opened 2015-05-20
Makefile assumes tar has function modifier z


alanfeldstein:
cd tmp && tar xfz ../extlib-1.5.3.tar.gz
tar: z: unknown function modifier
Usage: tar {c|r|t|u|x}[BDeEFhilmnopPqTvw@[0-7]][bfk][X...] [blocksize] [tarfile] [size] [exclude-file...] {file | -I include-file | -C directory file}...

This comes from
opam-full-1.2.2-rc2/src_ext/Makefile

%.stamp: %.download
    mkdir -p tmp
    cd tmp && tar xf$(if $(patsubst %.tar.gz,,$(URL_$_)),j,z) ../$(notdir $(URL_$_))
...

The makefile seems to be assuming GNU tar. There are at least three possible solutions to this bug:

(1) Have configure look for gtar and use it over tar if found.
This solution is not ideal because gtar is not the standard name for GNU tar.

(2) Simply add the GNU tar requirement to README.md.
This is a lazy solution, pushing responsibility onto the user.

(3) Modify the makefile to call gunzip, followed by tar.
This is the most robust solution, supporting the greatest number of platforms. One person makes the change, benefiting multiple users.

Alan Feldstein
Cosmic Horizon
http://www.alanfeldstein.com/

avsm:
Which platform are you build on, out of curiosity?

alanfeldstein:
Solaris 10

AltGr:
Indeed, (3) seems like a good solution. However, the opam binary itself will later want to "tar xfz"/"tar xfj" the downloaded packages, at installation time -- so we would need to workaround that too.

alanfeldstein:
I could create and submit a fix that employs solution (3), but the fact that the OPAM binary will also want to "tar xfz"/"tar xfj" means that mine would be only a partial solution. If OPAM itself is an OCaml program and I want OPAM to install utop as suggested by "Real World OCaml", which I'm reading to learn OCaml, then I'm not the best candidate to fix the OPAM binary. Rather, I'm motivated to await these fixes while diverting my attention to installing utop from sources.

AltGr:
That's fine, your reports are pretty useful already -- thanks!

michipili:
If this is helpful, I have written [autoconf macros to detect compression options supported by tar](https://github.com/michipili/bsdowl/blob/master/aclocal.m4#L167).  Feel free to plunder this treasure! :)

[...]

PR #2945, opened 2017-05-20
I needed these two changes in order to "make cold" on FreeBSD.

zhenya1007:


src_ext/Makefile
```diff
@@ -70,7 +70,7 @@ endef
 
 %.stamp: %.download
 	mkdir -p tmp
-	cd tmp && $(if $(patsubst %.tar.gz,,$(URL_$*)),bunzip2,gunzip) -c ../$(notdir $(URL_$*)) | tar x
+	cd tmp && tar x$(if $(patsubst %.tar.gz,,$(URL_$*)),j,z)f ../$(notdir $(URL_$*))
```
dra27: This is a revert of e740dd6b8880462c5b581a6b404917451611cb47 and can't go through as-is: see https://github.com/ocaml/opam/issues/2158

I'm being lazy and not booting my FreeBSD VM, but is it really possible to have `tar -j` working yet no `bunzip2`? Would it be better to use `-d` with the `bzip2` or `gzip` commands instead?

zhenya1007:
I think the problem is `tar x`.  The man page says,

>      -f file, --file file
>           Read the archive from or write the archive to the specified file.
>            The filename can be - for standard input or standard output.  The
>           default varies by system; on FreeBSD, the default is */dev/sa0*; on
>            Linux, the default is /dev/st0.

I saw `tar` complain that it cannot read from `/dev/sa0`.  

Let me try `tar xf -`: with any luck that works for everyone.

zhenya1007:
Yup, `tar xf -` works on FreeBSD.  The [Solaris man page I found](http://docs.oracle.com/cd/E19109-01/tsolaris8/817-0879/6mgl9vnhn/index.html) suggests that it should work on Solaris, also.  

With any luck, this change can now be merged.  If you need me to squash the commits in my branch, please let me know.

AltGr:
Thanks!

PR #2956, opened 2017-06-03
Developer tweaks, mark #2

dra27:
This is a follow-on from discussions in #2937.

Add the --enable-developer-mode, the src_ext/archives caching and the tweak to bootstrap-ocaml.sh from #2937.

Additionally, this includes a refactoring of the src_ext subproj definitions. Apart from making the definitions much more concise (and IMO therefore clearer), it also allows the lib-ext copy rule to be more precise. In particular, it correctly installs pack .cmx files which eliminates Warning 58 on recent OCamls.

With this change in place (assuming I've got it right), it should allow the CI to warn with `-warn-error A` and `-warn-error A` is also added if you specify `--enable-developer-mode` to configure.


AltGr:
Thanks! All good to me :)

PR #2971, opened 2017-06-16
Improve handling of repository updates across version, add 'announce:'

AltGr:


doc/pages/Manual.md
```diff
@@ -577,6 +577,9 @@ some specifics of the repository. It has the following optional fields:
   similar [config field](#configfield-archive-mirrors), except a path without a
   `protocol://` prefix is accepted and will be considered relative to the
   repository root (e.g. `cache/`).
+* <a id="repofield-announce">`redirect: [ <string> { <filter> } ... ]`</a>:
```
hannesm: `redirect:` -> `announce:`

avsm:
Great feature! Would this be useful to slightly generalise to add an `motd` on `opam update`? That would let us signpost breaking changes to users more easily for more than just first init.

AltGr:
That's what it is, it will be displayed on every update! There are filters too, so we could add variables `{init}`, `{update}`; but is there a need to ?

doc/pages/Manual.md
AltGr: indeed! Thanks for spotting this

PR #3000, opened 2017-07-16
Use const instead of pure

vramana:
Term.pure is deprecated


vramana:
I haven't checked it. Will check it later.

vramana:
@AltGr 0.9.8 version introduced `Term.const`  and deprecated `Term.pure` so I think we are good.

AltGr:
Great, thanks!

PR #3004, opened 2017-07-22
update default ocaml compiler to latest release

yomimono:
If 4.05.0 seems too recent I'm happy to replace with 4.04.2, but 4.02.1 seems definitely wrong.


AltGr:
Thanks!

PR #3009, opened 2017-07-25
Get rid of jsonm dep

dbuenzli:
Remove the `jsonm` and transitive `uutf` dependency. 

This allows to ease debian bootstrapping: latest versions of `jsonm` and `uutf` use `topkg` and hence `opam-installer` which is provided by `opam` itself.

@AltGr note that the output is now minified which used not to be the case. I hope it's not a problem as it allows for a very simple encoder implementation. If people need to read the json they can always pipe the output into `jsontrip -pp`.

/cc @hendriktews

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,8 +11,6 @@ src_ext/cmdliner/
 src_ext/extlib/
 src_ext/re/
 src_ext/graph/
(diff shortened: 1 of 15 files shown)


AltGr:
Thanks a lot!

PR #3011, opened 2017-07-31
Discard "internal heuristics", add builtin solver using `ocaml-mccs`

AltGr:
Note: not for merge right away, this needs the new build-system to be finalised first


AltGr:
Now based on the merged jbuilder patches, and cleaned up (supporting build without mccs): should be ready.

dra27:
Can I have until Friday afternoon to try this on Windows and review further?

One immediate thought: I think the dynamic selection of `static-linking.sexp` is wrong (because it's not future-proof). I would either have it that `configure` copies `static-linking.sexp.true` if static linking is wanted and the jbuild rule is just `(with-stdout-to ${@} (echo "()"))` with no dependencies (so you get dynamic linking if you build without running configure) or, and this is probably better, define an additional package opam-static in src/client/jbuild with a different library and push the static/dynamic option back to ocaml-mccs (so opam-mccs also defines a findlib package ocaml-mccs.static).

AltGr:
Going through the env is clearly a temporary hack, until there is a better way to select the version we want. Two different targets/aliases would nicely do the trick, but I am not sure how they would relate to the opam package (which shouldn't include both... ?)

AltGr:
Mostly OK now. Remaining issues:
-  `OCAML_VERSION=4.03.0 OPAM_TEST=1 EXTERNAL_SOLVER=` glpk wasn't found, and opam was compiled without internal sover. Therefore, forcing the internal solver fails.
-  `OCAML_VERSION=4.03.0 OPAM_TEST=1 EXTERNAL_SOLVER=aspcud` the check for aspcud version seems broken (on older aspcud)
- `OCAML_VERSION=4.05.0 OPAM_TEST=` tests from the opam repo seem to fail, need to investigate, but this test is new
- `osx` `OCAML_VERSION=4.03.0 OPAM_TEST=1 EXTERNAL_SOLVER=` seems that `re` is still missing ?
- `osx` `OCAML_VERSION=4.03.0 OPAM_TEST=1` "no external solver found" ?

AltGr:
Ah, the bootstrap opam (1.2.2) doesn't correctly extract the opam file from the `ocaml-mccs` archive.
This means we're stalled by https://github.com/ocaml/opam-repository/pull/10075

AltGr:
_Finally_ ! All tests pass :)

PR #3015, opened 2017-08-03
Jbuilder support

dra27:
Depends on https://github.com/janestreet/jbuilder/pull/153, which I'm about to address..

@AltGr - I'll rebase your ocaml-mccs changes onto this in dra27/ocaml-mccs and report back later this afternoon; there were some minor missing changes in my branch to do with not calling `cygpath` on Unix...


AltGr:
Great; I had removed the more clever jbuilder part (dropping support for older OCaml) on my `ocaml-mccs` branch waiting for this.

dra27:
I've just pushed changes to the jbuilder PR and I'm about to see @diml :smile:

dra27:
I think it will be better to get it merged with all build systems intact (but not building with jbuilder means you can't have mccs support) and then remove the other two, just so we have a clear working point in the git history

AltGr:
Not sure I really want to see opam with 3 build systems, even if it's transient, to be honest :crab:

dra27:
Only for one commit!

.travis-ci.sh
```diff
@@ -13,7 +13,12 @@ install-bootstrap () {
     eval $(opam config env --root=$OPAMBSROOT)
     if [ "$OPAM_TEST" = "1" ]; then
         opam pin add opam-file-format "git://github.com/ocaml/opam-file-format.git" --no-action --yes
-        opam install ocamlfind lwt.3.0.0 cohttp.0.22.0 ssl cmdliner dose3 opam-file-format --yes
+        if [ "$OCAML_VERSION" != "4.01.0" ] ; then
+          EXTRAS="lwt.3.0.0 cohttp.0.22.0 ssl"
```
avsm: Why an older version of cohttp here?

dra27: No idea! I was just moving this out of the way as it doesn't build on 4.01.0 and neither does lib-ext (but this was a silly way of trying to do it, which I've just changed).

dra27:
Right, I think this is now properly polished and ready for review!

This introduces Jbuilder as a third building option for Opam (but with the intention that it will shortly be the only build option). There are various notes about how this has been done:
 - Building lib-ext with Jbuilder means that for OCaml 4.01 the dependencies cannot be installed from src_ext - Travis has been updated to reflect this.
 - Jbuilder automatically bootstraps itself in src_ext if its not installed; thus `./configure ; make [lib-ext] jbuilder` will build Jbuilder if necessary
 - The opam build of Jbuilder relies on features merged, but not yet released - thus if building from opam, Jbuilder should be pinned to its dev-repo; lib-ext is presently downloading https://github.com/janestreet/jbuilder/tree/46df511dd4fa5101f871ba85daa19a9540a7cd3c
 - There is some refactoring and reorganisation of the Makefiles and the location of some auxiliary files (scripts, etc.)
 - There is a serious problem with Cudf 0.8 and 0.9 - I've bisected to a perfectly sane change in Cudf and included a patch which reverts the alteration, but we should determine exactly what causes the testsuite to fail when the Cudf parsers hashtables are too small, as it's probably masking something more serious?!
 - I'm not terribly proud of `shell/subst_var.ml`, but I'm not sure of a better (platform-independent) way of extracting information from autoconf for jbuilder. While it will likely remain desirable to run `./configure` as a first step of building Opam, `jbuilder build` should work from a clean checkout/untar of the code (which it does).
 - I've upgrade dall the lib-ext dependencies to their latest versions (at least they were latest versions when I last looked...)
 - Note that if you've built with lib-ext, you get the full vendoring power of Jbuilder - if you edit sources in `src_ext/foo/bar.ml` and run `make jbuilder` then the dependent library will be recompiled at the same time as all the required parts of opam.
 - `jbuilder runtest` works correctly, for completeness
 - If you are building with jbuilder, you should ensure that you have a clean sources tree - artefacts left behind from the normal `make all` route have a habit of confusing jbuilder, as it will use them, instead of building new ones.

AltGr:
This still requires an unreleased-yet jbuilder doesn't it ?

dra27:
Indeed - `opam pin add jbuilder.1.0-beta12 --dev-repo` needed for now

PR #3016, opened 2017-08-04
Switch to jbuilder only

dra27:
Let's see what the CI makes of this!


dra27:
Note that in order to test this, you need to pin jbuilder using `opam pin jbuilder.1.0+beta12 --dev-repo`

dra27:
https://github.com/ocaml/opam-rt/pulls can be merged before this, which will fix the CI. At that point, it should be possible to dump the last commit.

Going all-in on Jbuilder necessarily drops support for OCaml 4.01.0...

dra27:
Huzzah - it passes! I've maintained most of the entry points in the Makefile here, which we may want to consider changing soon - `make all` still works, but it's not quite as efficient as running `jbuilder build`. The use of the `-p` option when generating the `.install` files (e.g. `jbuilder build -p opam-client opam-client.install`) is particularly pleasing, because it means that the individual library packages now build much faster when installed via opam - with the old build system, installing `opam-client` would mean that `opam-core` was compiled 6 times, with jbuilder all the subsequent libraries automatically use the installed `opam-core`.

Note that `make -C tests` no longer makes sense - `tests/Makefile` only works from within `jbuilder runtest` (a slightly strange state of affairs, but the testsuite is too much code to bother porting at the moment). I haven't tested it, but I think that documentation generation should still be working properly.

dra27:
Incidentally, for people like me who like these things: given that opam-repository recently got to 10000 issues, I think commit `11fd8f69` in this branch is the first opam commit which needs 8 letters of the sha to be specified...

AltGr:
> https://github.com/ocaml/opam-rt/pulls can be merged before this, which will fix the CI. At that point, it should be possible to dump the last commit.
Note that by using a same-name branch on opam-rt you can have the tests run with it. It's useful when there are incompatible changes, to avoid a specific patch :)

> Going all-in on Jbuilder necessarily drops support for OCaml 4.01.0...

Yup. It shouldn't be an issue at this point, esp. if we can deliver solver-including binaries.

> I've maintained most of the entry points in the Makefile here, which we may want to consider changing soon

I am for making it as simple as possible, we don't need compat stuff for the build system

> Incidentally, for people like me who like these things. opam-repository recently got to 10000 issues, and I think commit 11fd8f69 in this branch is the first opam commit which needs 8 letters of the sha to be specified...

A good pretext to have a beer on the next occasion :beer: !

AltGr:
Just waiting for e423a91 to be removed before merging

AltGr:
> The use of the -p option when generating the .install files (e.g. jbuilder build -p opam-client opam-client.install) is particularly pleasing, because it means that the individual library packages now build much faster when installed via opam - with the old build system, installing opam-client would mean that opam-core was compiled 6 times, with jbuilder all the subsequent libraries automatically use the installed opam-core.

Although that was done manually, that had to be the case already for compiling the individual opam packages (`opam-format`, etc.) ; that was done in `ocp-build` by setting variables and conditionals in the `build.ocp` files.

dra27:
I'll rebase to remove the commit later (out of the office at the moment)

dra27:
Incidentally, would it be best in general to bump the versions of the opam files as soon as an actual (beta) release is cut (it's how OCaml is released) so that a git pin always has a later version?

AltGr:
Yes. But that means we'd need to plan the versions ahead of time ;)

avsm:
The advantage of bumping the format in the dev version is that you are testing upgrades all the time :-)

dra27:
@AltGr - not really, it just has to be the case that the version released is semantically greater than or equal to the Git version!

dra27:
(in this case, I just mean that when beta 3 was released, you change the opam files in git to beta4... when you released beta3.1, you'd have just left the opam files unchanged :smile:)

dra27:
Regarding `Makefile` simplifications, I think we could delete `src/Makefile` completely and for the root `Makefile`:
 - Replace the existing `all` target with the `jbuilder` target (and rename it to `all`)
 - Delete everything after the `src_ext/jbuild.stamp` down to the `lib-ext` target
 - I think the `tests-%` is broken at the moment (that's a weird setup because `tests/Makefile` is now set to act as though it's been copied to `_build`, so it doesn't work when called directly). I'd just delete this target - it seems, looking at opam-rt, etc., that it would be better to move these tests to a "better" framework than spend effort porting them to work more neatly with the jbuilder environment.

dbuenzli:
Are you sure the dependency story is fine with jbuilder for system packagers ? It may be good to keep the Makefiles around to allow them to bootstrap the OCaml support.

dra27:
It's only `src/Makefile` which I was talking about deleting - the lib-ext and cold bootstrap is all still there (and very much wanted by me on Windows!).

It should still be the case that you can get from a system with just a C compiler environment to a fully bootstrapped opam (`make cold` etc)

PR #3021, opened 2017-08-08
Bump version and update the release script

AltGr:
NOTE: will need rebase


dra27:
Note that we'll need to verify BSD building again (/cc @avsm) - it's possible some of the lib-ext overhaul has introduced problems

PR #3027, opened 2017-08-17
Use jbuilder from src_ext in doc Makefile

tchajed:
This allows building the docs even without jbuilder on the PATH.

There's some code duplication which I'm not sure how to fix, since the relative path needs to be different for the two Makefiles.


AltGr:
Thanks!

PR #3030, opened 2017-08-18
Use HTTPS for all GitHub URLs

tchajed:
This is good practice anyway but I'm also seeing the downloads fail with curl using the HTTP URL.

The remaining, non-GitHub HTTP URLs don't appear to support HTTPS due to certificate issues.


AltGr:
Thanks!

PR #3031, opened 2017-08-18
Fix some build details for OpenBSD

AltGr:


src/client/jbuild
```diff
@@ -31,9 +31,8 @@ let (sha, version) =
 let static = try match Sys.getenv "OPAM_BUILD_STATIC" with
   | "1" | "yes" | "true" ->
     (try match List.assoc "system" J.ocamlc_config with
-       | "linux" -> "gcc"
        | "macosx" -> "clang"
-       | _ -> failwith "Sorry, static builds not supported on your platform"
+       | _ -> "gcc"
```
avsm: any reason not to use `cc` here? that is usually a sensible default

AltGr: this is only the extension of the static-linking file, which is actually quite specific (actually depends on how the installed glpk has been linked). Maybe something by OS would be better here. In any case, this is actually related to the version of the linker, not the compiler.

AltGr:
Some Travis tests on OSX fail due to a stack overflow during the preparation phase... in the bootstrap opam 1.2.2 !
Time to release ?

PR #3043, opened 2017-09-01
Restore Travis

dra27:


dra27:
There is still the OSX failure, which I couldn't figure out - something up with opam-rt?

AltGr:
It's a bug with opam 1.2, which we use for bootstrapping :/
Not sure exactly how to work around it yet.

PR #3047, opened 2017-09-04
fixing outdated link to OpenSUSE build server in Install.md

quicquid:
It appears that the link to the rpm in the installation instructions is not working anymore. I believe [this is the correct one](https://software.opensuse.org/package/opam).

### DIFF
--- a/doc/pages/Install.md
+++ b/doc/pages/Install.md
@@ -107,7 +107,7 @@ cave resolve -x repository/ocaml-unofficial
 
 RPMs for Fedora, CentOS and Red Hat Enterprise Linux are available with
 instructions on the
-[OpenSUSE Build Server](http://software.opensuse.org/download.html?project=home%3Aocaml&package=opam).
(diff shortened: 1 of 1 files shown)


quicquid:
Three of the CI failures come from a missing re library, the fourth one seems to have worked fine but was terminated during phase 7/8 of make world.opt because of missing output. Since I only changed a documentation file, I assume the failures are unrelated.

dra27:
The CI is indeed down at the moment - see #3043

AltGr:
Thanks!

quicquid:
Thanks too! I would be lost without opam :)

ANogin:
https://software.opensuse.org/package/opam only has OpenSUSE packages, but not Fedora or CentOS - do those still exist somewhere? Thanks!

quicquid:
I'm running the OpenSUSE package (opam-1.2.2-3.1) under Fedora 26 without problems.

Issue #3039, opened 2017-08-31
Cannot build Opam 2.0.0~beta4 with built-in solver


sebastienaubry:
Hi,
I am trying to build Opam 2.0.0~beta4 with the built-in mccs solver in order to avoid any dependency.
However, configure tells me "Opam will be built WITHOUT a built-in solver".
Can it be related to the "/usr/bin/ld: cannot find -lglpk" error line in the attached config.log ?
Thanks

[config.txt](https://github.com/ocaml/opam/files/1266640/config.txt)

Drup:
Yes, you need to have glpk installed on your system. It is already packaged in most distribs.

sebastienaubry:
Thank you. I am not root on my system, so I cannot install it. I guess I will have to build either glpk or an external solver. This wasn't needed by the previous versions of Opam.

dra27:
It wasn't needed before, indeed - but the internal solver was also not very good! We should investigate the possibility of vendoring glpk as we do for the other dependencies (this will almost certainly compulsory for Windows anyway)

sebastienaubry:
Thank you. Opam is integrated into our build process, just to use it once in order to install a few OCaml packages. We do not need the best solver ever :-)

sebastienaubry:
Hi,
I have compiled GLPK, so it is now installed in a custom directory.
I would like to use it in order to build Opam 2.0.0~beta4.

I had to use the `--includedir` and `--libdir` parameters of configure in order to get `checking for glp_create_prob in -lglpk... yes` and `Opam will be built WITH a built-in solver`.

But `make all` calls `jbuilder build  opam.install` which tries to build mccs. It generates this command line, which does not take into account the previous `--includedir` and `--libdir`:
```
(cd _build/default/src_ext/mccs/src && /usr/bin/gcc -I /path/to/lib/ocaml -O -fno-defer-pop \
-Wall -D_FILE_OFFSET_BITS=64 -D_REENTRANT -O -fPIC -I ../../cudf -I ../../extlib/src -g \
-I . -o glpk_solver.o -c glpk_solver.cpp)
In file included from glpk_solver.cpp:11:0:
./glpk_solver.h:15:18: erreur fatale: glpk.h : Aucun fichier ou dossier de ce type
 #include <glpk.h>
                  ^
```

AltGr:
[...]

PR #3051, opened 2017-09-07
Support specifying additional library dirs to ./configure (solves #3039)

AltGr:
Using a custom glpk can now be achieved with:
`./configure LIB_PREFIX=local_prefix`
or, equivalently,
`./configure CPATH=local_prefix/include LIBRARY_PATH=local_prefix/lib`

then `make lib-ext; make`.


AltGr:
@dra27 any idea what's up with appveyor ?

dra27:
C++ compiler's missing - you need gcc-g++ I think

PR #3053, opened 2017-09-11
Add option '--unlock-base' to change the compiler

AltGr:
This makes single commands such as `opam upgrade ocaml --unlock` possible


dra27:
https://github.com/OCamlPro/opam/pull/2 should fix AppVeyor

Issue #2996, opened 2017-07-08
v2: status of `base` packages


dbuenzli:
I think that the v2 migration could be an opportunity to review these `base` packages. Here are a few things: 

1. The `base-` prefix should be dropped. We should try to converge to a system where library, package and `ocamlfind` names do coincide.
2. Why are `unix`, `bigarray` and `threads` treated specially ? I think these should just be packages to be installed that depend on the `ocaml` package. 

Other packages already work the way I suggest `uchar`, `result`, `base-bytes` (well that `base-` prefix should die).

dra27:
There's a certain amount of history with these - they're partly a workaround for `provides:`. `unix`, `bigarray` and `threads` get special treatment to allow for exotic compiler switches which do not build these libraries - you can't express that in a dependency (at least today).

`uchar` and `result` are slightly different because those packages actually have build systems which take compiler version into account and which build something if the compiler requires it (cf. the ocamlbuild package and ocamlbuild.base for the other way of doing it). `base-bytes` similarly can work as meta-package because bytes is always built - it's just a version constraint on the compiler.

So at the moment the `base-` packages are meta-packages a little like `conf-`

dbuenzli:
> There's a certain amount of history with these - they're partly a workaround for provides:. unix, bigarray and threads get special treatment to allow for exotic compiler switches which do not build these libraries - you can't express that in a dependency (at least today).

Why not ? The `graphics` package does exactly that.

dra27:
Sure, you can feature-probe instead (though I think that `graphics` package is a bad way of doing it - I would have noticed that when I finally get to doing the others...).

The weakness of that is that you get a poorer user experience - opam declares that you can install a whole load of packages and then that only fails when it can't install `unix`, for example. With the base packages, you get a solver error which, IMO, is rather better - the solver tells you that you can't do it, rather than waiting for an artificial package failure.

dbuenzli:
> The weakness of that is that you get a poorer user experience - opam declares that you can install a whole load of packages and then that only fails when it can't install unix, for example. With the base packages, you get a solver error which, IMO, is rather better - the solver tells you that you can't do it, rather than waiting for an artificial package failure.

Well is that really worth the price of having ad-hoc support for this ? From an end-user experience, in both cases you end up with an error that tells you that a package can't be installed; no big-deal. Not to mention that almost all supported platform does support these libraries. So for now I'd rather have `graphics`-like packages with the correct names and `provides:` once this get implemented.

[...]

PR #3054, opened 2017-09-13
Add a 'post' dependency flag.

### LINKED ISSUE #2649: add a `compiler-bundle:` field
This arose from a discussion with @dra27 who implemented something similar for handling the installation of flexlink/flexdll on Windows.

The idea is that when you install a compiler, it may "come with" other packages that get installed along with it: `opam switch create foo` would create the switch, and then attempt to install `foo` and all its `compiler-bundle` packages (including, of course, all dependencies); the bundled packages don't force any direction of dependency, and are afterwards marked as part of the compiler¹.
It may seem a bit ad-hoc, but actually solves a couple issues:
- Legacy compiler definitions came with packages `base-xxx` that mostly had this semantics. I replaced them with `depends:`, which works only because they are actually dummy packages, and is not very pleasant.
- In the current state, the actual compiler is _e.g._ `ocaml-base-compiler`, and `ocaml` a virtual package. What is weird is that it won't get installed together with the compiler (only when other packages "pull" it), and is not actually part of the "base packages"; putting it in `compiler-bundle` might help reduce the confusion.

A more generic field name could be `recommends:`, which kind of fits with these semantics, but there will be no handling of that for normal packages for the moment.

¹ that may differ from @dra27's idea of the field; would it still be OK for your use ?

AltGr:
This resolves #2649, and provides a solution for ocaml/opam-repository#7543, and a better story for `base` packages (#2996). It also allows to have the `ocaml` package be installed with the switch, rather than lazily afterwards, which apparently created some confusion.

Two notes though:
- maybe find a better name ? Dependencies tagged `reverse` may, or may not, actually be reverse dependencies (e.g. `a` with `depends: ["b" {reverse}]` does not imply that `b` depends on `a`, although that may often be the case). Better avoid this kind of confusion
- ~~I added the flag before realising that it was mostly equivalent to just specifying `!build`, which may actually be easier to understand (a dependency that is not active at build time). So we could remove commit ae4f3764, and adapt the repository rewriting in 697fba64, adding some documentation.~~ EDIT: I take this back, it doesn't really match. We could still use `!build` instead or `reverse`, but it's actually a bit convoluted.

Hence this PR is more for review on the concept at the moment

### DIFF
--- a/doc/pages/Manual.md
+++ b/doc/pages/Manual.md
@@ -823,6 +823,11 @@ files.
 
     * `build` dependencies are no longer needed at run-time: they won't trigger
       recompilations of your package.
+    * `post` dependencies will be installed along with the package, but are not
+      required to build it. This can be used to cut build cycles of
+      interdependent packages, while making sure they get installed together.
+      Note that, in case of failed or interrupted builds, opam can not guarantee
+      the invariant that `!build` dependencies are always installed.
     * `with-test` dependencies are only needed when building tests (when the
       package is explicitely installed with `--with-test`)
     * likewise, `with-doc` dependecies are only required when building the
@@ -839,7 +844,7 @@ files.
 
     Variables in the filtered package formula are evaluated as for
     [`depends:`](#opamfield-depends), with the same specific variables
-    available.
+    available (except for `post`, which wouldn't make sense).
 
     Note that `depopts: [ "foo" { = "3" } ]` means that the optional dependency
     only applies for `foo` version `3`, not that your package can't be installed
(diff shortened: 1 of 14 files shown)


AltGr:
@dra27 could you have a look at appveyor ?

AltGr:
An idea for a better name ? `nonbuild` ? `undepends` ? `supply` ? `affix` ? `post` ? `post-dependency` ?

hcarty:
I like `post` - concise and indicates the intent well I think.

AltGr:
Agreed — renamed to `post`

Drup:
Interesting. It sounds a bit boilerplate-y, but I guess it would provide a decent solution for now.

dra27:
https://github.com/OCamlPro/opam/pull/2 should sort out AppVeyor - it's Cygwin-related.

AltGr:
>  It sounds a bit boilerplate-y, but I guess it would provide a decent solution for now.

Well, that was my impression with the `compiler-bundle:` field, but in this way I feel that it fits quite well within the dependencies/solver/actions/solution abstractions of opam. The added code complexity is also quite small.

PR #3056, opened 2017-09-13
Some small fixes, and jbuilder update

AltGr:


dra27:
Huzzah - you just need Travis to catch up and then you have a whole load of AppVeyor jobs to restart on other PRs!

dra27:
The AppVeyor guys will update their images at some point to have a later Cygwin DLL (we had the same problem with the transition from 2.5.2 to 2.6.0), and when they do it will be possible to remove the `-g` from appveyor.yml and speed the CI builds up slightly.

PR #3059, opened 2017-09-15
Add a 'pin-depends:' field for dev packages depending on specific pins

AltGr:


Drup:
Wait, I'm not sure I'm reading the code right. What happens when I `opam install` a package with some `pin-depends`? Does opam ask interactively to pin things ?

Drup:
Regardless of the semantics, We should have a mode in the linter to disallow this, so that it never makes it through CI in opam-repository.

AltGr:
The field has no effect whatsoever on the repository (as documented [here](https://github.com/ocaml/opam/pull/3059/commits/6744851c4ec6324452df271b4d6fc448e5e30c19#diff-5400dd91306c06d0aae0d7cdc17aa9fdR983)): it is only useful whenever the package is manually source-pinned. As such, allowing it on the repository could add a little noise, but also avoid inconsistencies between the in-source and in-repository `opam` files. Note however that, by the time you release, the package obviously shouldn't depend on any pins anymore, so you may be right about the linting!

Drup:
Ah, I see. So, let me see if I get the semantics right: *if* the package is pinned, *then* tries to pin all the URLs specified in `pin-depends`, non-recursively (why?). Is that correct? I think the documentation should be made a bit clearer (by changing the order of the explanation).

AltGr:
>  Is that correct?

Mostly :)

> if the package is pinned,

_source_-pinned, version pinnings are unaffected

> then tries to pin all the URLs specified in pin-depends, non-recursively (why?)

This is the same rationale as behind npm (and other)'s lock files: when you want to specify a precise build environment, you'd rather do it once (if you want recursive resolution, use released packages!) :
- there might be conflicts (same packages pinned differently) in the transitive `pin-depends:`
- you get more control anyway by specifying all you want to pin at once, in the outermost package. Re-specify transitive pins there if you want them.

This is much more predictable and easier to read this way.

Drup:
Ok. The use case I have in mind (mirage unikernels) is slightly different, but the same remarks should apply.

I think you will need to add a dedicated section to the manual too. :)

AltGr:
I really have no idea about, and can't reproduce, that weird Travis failure with 4.05.0...

AltGr:
Finally found out: https://caml.inria.fr/mantis/view.php?id=7640

PR #3061, opened 2017-09-17
Split out opam-installer package

timbertson:
I'd quite like to use `opam-installer` to install ocaml packages without having the rest of `opam` around.

I guess it's a weird request, making an `opam` package so that I can not use `opam`. I specifically want this for [opam2nix](https://github.com/timbertson/opam2nix-packages), which uses opam package data (and code) to create reproducible ocaml environments via [nix](https://nixos.org/nix/). In that case I'm able to build packages from `.opam` files without opam itself. And it's important to be able to run this environment without `opam` being installed, since accidentally relying on `opam` would subvert nix's reproducability guarantees.

**Note**: There's no `opam.opam` file, which I guess makes sense. But since I moved the `opam-installer` executable into its own package, I don't know what else I need to update to make sure `opam-installer` still ends up installed when you build & install opam.

There's also a weird hack in the build steps, see https://github.com/janestreet/jbuilder/issues/257 for more details.


AltGr:
Thanks! A separate `opam-installer` package can indeed be useful for several cases. One possible improvement would be to make it only depend on `opam-format`, rather than the full `opam-state` stack.

The only use of `opam-format` that is made currently by `opam-installer` is the `OpamPath.Default` module, which contains the definition of the static `bin`, `lib`, `share`, etc. paths. And that brings in dependencies to the whole solver and sync libraries. I guess it wouldn't be too bad to just duplicate these definitions (or move them to `OpamFormat`, but I don't see where they would fit).

> Note: There's no opam.opam file

The _real_ `opam` package is actually `opam-devel.opam`, because we don't want to override the `opam` binary (that would be a recipe for disaster), or even take the package name.

timbertson:
I'm not quite sure I follow - could you please clarify a few things?

> The only use of opam-format that is made currently by opam-installer is the OpamPath.Default module

That's currently in `opam-state` isn't it? Not `opam-format`?

> that brings in dependencies to the whole solver and sync libraries

I tried moving OpamPath into opam-format from opam-state (05185ab351f084ac4b7c8ad450093d3de720c2e7). It still seems to compile OK, so I'm not sure which parts bring in the solver/sync libs.

Does moving OpamPath affect the public API? `opam-state` still depends on `opam-format` so I'm thinking maybe not, as long as modules from transitive dependencies are always visible during compilation.

> we don't want to override the opam binary

Ah, that sounds entirely reasonable :)

AltGr:
> > The only use of opam-format that is made currently by opam-installer is the OpamPath.Default module

> That's currently in opam-state isn't it? Not opam-format?

Indeed, I meant `opam-state`, my bad.

> I tried moving OpamPath into opam-format from opam-state (05185ab). It still seems to compile OK, so I'm not sure which parts bring in the solver/sync libs.
>
>Does moving OpamPath affect the public API? opam-state still depends on opam-format so I'm thinking maybe not, as long as modules from transitive dependencies are always visible during compilation.

Ah, I hadn't actually thought about moving the whole module... but indeed, since the actual opam root is a parameter for all functions, that makes sense as well, and I guess other tools based on `opam-format` may have use for it. There shouldn't be compatibility problems, since `opam-state` depends on `opam-format` already, as you point out. Thanks! Please also update `doc/index.html` to reflect the new location of the module.

It seems fixing the CI could be done by just adjusting `make all install` to still compile and install `opam-installer`.

Thanks a lot!

timbertson:
`doc/index.html` updated, thanks for the feedback :+1:

`opam_installer.exe` was already being built, but the build seemed to be failing because it didn't know that its public name was `opam-installer`. I changed it to build `opam-installer.install` and now it seems happy.

AltGr:
Looks good, thanks!

timbertson:
Hmm, I can't figure out why 4/7 travis builds are failing on https://travis-ci.org/ocaml/opam/builds/277278371?utm_source=github_status&utm_medium=notification.

The main thing that sticks out to me is that all the successful builds built jbuilder themselves (`checking for jbuilder... no`), while all the failed builds used a preinstalled jbuilder. Perhaps there's a bug in earlier versions which are installed on the travis images which affects how executables are found in `jbuilder exec`? Is there a way to bump the system version of jbuilder to check whether this is the issue?

The only outlier is build 2862.5, which didn't find jbuilder but then failed to compile ocaml, which I'm hoping wasn't my fault! ;)

AltGr:
> The only outlier is build 2862.5, which didn't find jbuilder but then failed to compile ocaml, which I'm hoping wasn't my fault! ;)

yes, that one popped on other PRs too, it puzzles me and I don't have a fix yet, but it's certainly not your fault 😌

For the others, I guess Makefile rules doing `$(JBUILDER) exec $(JBUILDER_ARGS) -- opam-installer` need to depend on `opam-installer` being compiled ?

timbertson:
Based on the logs though, each of the failed builds has already built opam-installer.install. e.g:

```
ln -s -f _build/default/src/tools/opam_admin_top.bc opam-admin.top
jbuilder build  opam-installer.install
ln -s -f _build/default/src/tools/opam_installer.exe opam-installer
jbuilder build  -p opam-core opam-core.install
    ocamlopt src/core/opam_core.cmxs
jbuilder exec  -- opam-installer --prefix "/home/travis/local" --mandir "/home/travis/local/share/man" --libdir "/home/travis/local/lib" opam-core.install
Error: Program "opam-installer" not found!
```

previously in the same build I saw it building the actual exe, i.e.:

```
jbuilder build  opam-installer.install
    ocamldep src/tools/opam_installer.depends.ocamldep-output
      ocamlc src/tools/opam_installer.{cmi,cmo,cmt}
    ocamlopt src/tools/opam_installer.{cmx,o}
    ocamlopt src/tools/opam_installer.exe
```

So both the exe and the `.install` file appear to have been built already. That should be enough, shouldn't it?

timbertson:
I updated the build to include `opam-installer` files in the generated `opam-actual.install`, because I think that was why the `jbuilder exec` line failed to find `opam-installer` (it is actually using $PATH, not the local built artifacts... I think).

However, now it can't seem to build `opam-installer.install` at all. It's passing `-p opam-installer`, and it can't find the `opam-format` dependency. The library is in src/format, but the I'm assuming the `-p` means it would need to be externally installed. I'm a bit stuck as to what needs changing now - I assume `opam-installer` should be allowed to depend on the repository version of `opam-format`, does the Makefile need adjusting?

dra27:
@timbertson - I haven't checked this in great detail, but two hopefully helpful pointers:

 1. You are correct about jbuilder exec - it manipulates PATH to reference the correct bin directory inside _build, so you can't jbuilder exec a program unless it has been "installed" (this is the difference between expressly building an executable, which means it'll be in _build/*context*/src/tools/*blah*, vs building its install file, which will also copy/link it in _build/install/*context*/bin/*blah*)
 2. You're hitting the Makefile rules for the individual packages (which is why you're getting `-p`). I think you need a special rule for your opam-installer.install file, because opam-installer is not something you'd ever want to install in opam itself (for the same as opam is not installed in opam - see the detail of the opam-devel package).

AltGr:
This is off topic, but...

>  opam-installer is not something you'd ever want to install in opam itself (for the same as opam is not installed in opam - see the detail of the opam-devel package).

I am not sure there would be a problem with that ; opam doesn't actually rely on `opam-installer` so it should be OK. If you got opam by just copying the pre-compiled binary to your PATH, you may not have `opam-installer` and some tools may still need it (although `jbuilder` shouldn't, when run from opam)

timbertson:
Thanks @dra27, the details about _build/install helped. The issue seems to now be that running `jbuilder build -p <pkg> <something.install>` clears out the `_build/install/default/` directory for anything previously-built. You can verify this by:

```
$ jbuilder build opam-installer.install
$ ls _build/install/default/bin/
opam-installer

$ jbuilder build -p opam-core opam-core.install
$ ls _build/install/default/
doc/  lib/
```

If you drop the `-p opam-core` from the second command, `bin/opam-installer` will remain.

This is what happens with e.g. `make libinstall` - it depends on `opam-installer` but then builds `opam-core.install`, which wipes out `_build/install/default/bin`.

This kind of makes sense (you did ask for _just_ that installed package), so I assume it's intentional on jbuilder's part.

I can't figure out why this doesn't break on `master` though - the same should be happening there, since `opam-core.install` is built rght before a successful `jbuilder exec`: https://travis-ci.org/ocaml/opam/jobs/275471400#L1058

Any ideas?

Given that there is already a `./opam-installer` symlink in place, perhaps we could just invoke that instead of `jbuilder exec opam-installer`?

AltGr:
Seems good! Thanks a lot for the work (Travis + OSX failures are transient, I have been told, and were already there)

timbertson:
Great, sounds like it's nearly ready to merge then - anything left? Want me to rebase / squash anything? (I'm also happy for you to rebase my commits however you like...)

opam-devel.opam
```diff
@@ -21,6 +21,7 @@ build: [
 build-test: [make "tests"]
 depends: [
   "opam-client" {= "2.0.0~beta4"}
+  "opam-installer" {= "2.0.0~beta4"}
```
AltGr: I don't think this is really needed ?

AltGr:
I added two very minor remarks, otherwise, all good, thanks a lot!

Makefile
```diff
@@ -111,22 +114,22 @@ installlib-%: $(JBUILDER_DEP) opam-installer opam-%.install
 	$(if $(wildcard src_ext/lib/*),\
 	  $(error Installing the opam libraries is incompatible with embedding \
 	          the dependencies. Run 'make clean-ext' and try again))
-	$(JBUILDER) exec $(JBUILDER_ARGS) -- opam-installer $(OPAMINSTALLER_FLAGS) opam-$*.install
+	$(OPAMINSTALLER) $(OPAMINSTALLER_FLAGS) opam-$*.install
 
 uninstalllib-%: $(JBUILDER_DEP) opam-installer opam-%.install
```
AltGr: `$(JBUILDER_DEP)` could probably be removed from here (and below) now that the command doesn't use `$(JBUILDER)` anymore.

AltGr:
Thanks a lot!

PR #3065, opened 2017-09-28
Update .gitignore

mroch:
Not sure if you're taking PRs for 1.2 (i noticed there's at least one other unreleased 1.2-only change) but if so:

I built from source and these files were left untracked in my checkout.


AltGr:
There probably won't be a new 1.2 release, but PRs are welcome nevertheless. Thanks!

PR #3069, opened 2017-10-03
This fixes some (built-in) solver cases, and uses latest ocaml-mccs

AltGr:
(from which most of the fixes come)
Also adds solver timeouts


AltGr:
Build fails because ocaml-mccs is not new enough (https://github.com/ocaml/opam-repository/pull/10392)

PR #3070, opened 2017-10-09
Better conflict & formula simplification, 'opam admin add-constraint'

AltGr:


dbuenzli:
That seems excellent ! Would it be possible to specify on the cli a list of packages that should be excluded from the update. In practice you don't want to change these `rev-deps` whose compilation did not fail.

AltGr:
Adding a list of packages to exclude wouldn't be too difficult. Could you open an issue for tracking that ?

`add-constraint` has actually been ready for a few months, but I wasn't happy with the formula simplification algorithm yet.

dbuenzli:
Issue https://github.com/ocaml/opam/issues/3077

Issue #3058, opened 2017-09-14
RFC: variables to check for OS, distribution, etc.


AltGr:
Here is a proposal for new pre-defined variables, with examples. It's mostly inspired from the current state of [opam-depext](https://github.com/ocaml/opam-depext).

First some examples:

| host\variables | arch    | os     | distribution | dist-version | distribution-family |
|----------------|---------|--------|--------------|--------------|---------------------|
| Debian stretch | x86\_64 | linux  | debian       | 9            | debian              |
| OSX Yosemite   | x86\_64 | darwin | homebrew     | 10.10.5      | homebrew            |
| OpenBSD 6.1    | x86\_64 | bsd    | openbsd      | 6.1          | openbsd             |
| multiarch/debian-debootstrap:armhf-jessie container | armv7l  | linux  | debian       | 8            | debian              |
| 32bit/ubuntu:16.04 container | x86\_64 (known docker issue) | linux  | ubuntu       | 16.04        | debian              |
| centos:7 container | x86_64  | linux  | centos       | 7            | rhel                |
| Android 5.1.1  | armv7l  | linux  | android      | 5.1.1        | android             |
| resin/rpi-raspbian container | armv6l  | linux  | raspbian     | 8            | debian              |

* `arch` is the output of `uname -m`, with `i?86` rewritten to `x86`, `amd64` rewritten to `x86_64`. We might want some normalisation on `arm` too ? Note: any solution for Windows ? Assume `x86_64` ?
* `os` is either the output of `Sys.os_type`, or the lowercased output of `uname -s` when the former is `Unix`, except for `bsd` in the case of `*BSD` and `DragonFly` (this differs from `depext`). Note: should the win32/cygwin distinction rather be made at the `distribution` level ? Do we want to translate `darwin` to `osx` ?
* `distribution` is (originally following code from [opam-depext](https://github.com/ocaml/opam-depext/blob/e0e6a3c28b28bdacb3bbde0d9745fbcaa7510b52/depext.ml#L92-L130)):
  - on OSX, the first of `homebrew` or `macport` that has a binary installed (resp. `brew` and `port`) , or undefined otherwise
[...]

PR #3074, opened 2017-10-11
New depexts

### LINKED ISSUE #2919: A better depext
Context: see [this](https://github.com/ocaml/opam/blob/master/doc/design/depexts-plugins) original design document, and the [depext](https://github.com/ocaml/opam-depext) plugin.

This supersedes and closes issue #2172.

### Current status

While we have an implementation of the plugin that has been tested widely for a while, in particular by the CI systems around, it's not included into opam (which means beginners might hit missing depexts; opam tries to be helpful and advises to use it on failures, but it's still one more step). Also, the semantics is a bit weird, and it doesn't support versions of OSes or distributions.

1. The current field `depexts:` is documented [in the manual](http://opam.ocaml.org/doc/2.0/Manual.html#opamfield-depexts). It associates sets of tags to lists of package names.

2. The depext plugin analyses the host system to deduce:
- the package manager to use, with pre-defined commands for checking that a set of package is installed, and installing them
- a set of pre-defined tags based on arch, OS, distribution

For each package, the `depexts:` bindings are searched for keys which are a subset of the inferred tags for the host, and the associated system package names are added to the requirements, to be handled by the system-specific package manager.

### Limitations

The syntax is a bit heavy, leading to redundant specification, typically for debian and ubuntu:
```
depexts: [
  [["debian"] ["libssl-dev"]]
  [["ubuntu"] ["libssl-dev"]]
  [["centos"] ["openssl-devel"]]
]
```
Also, this doesn't handle any notion of versioning (#2168), with only static tags, so while adding tags for each distro/OS version could be possible, that doesn't seem realistic.

### A possible new syntax

Filters and variables are already used throughout in opam files, are more expressive and flexible, and have easier to understand semantics. Instead of the pre-defined tags, we could infer a set of host-describing variables (we already have `arch`, `os`) ; either hardcoded (using the code from `opam-depext`), or even using the new `eval-variables:` field from the main config file, that (lazily) defines variables from shell commands. [This](https://github.com/dra27/opam/commit/32a8338ebeaa4bfc37e37eee3aaa10aa7f405938) proposed patch by @dra27 goes in that direction.

Then a possible syntax, closer to what the other fields use, would be:
```
depexts: [
  ["libssl-dev"] { distro = "debian" | distro = "ubuntu" }
  ["openssl-devel"] { distro = "centos" & distro-version >= "7.2" }
]
```

we also benefit from version ordering as defined in the general opam format.

This still doesn't allow specifying versions of system packages, but handling that across packaging systems doesn't seem easily doable. Besides, we are the only ones to provide many versions for a single given package aren't we ? Eg Linux distributions have one version per package per release.

Remains the question of unavailable depexts: in the example above, what would happen if `distro-version` was less than `"7.2"` ? As written above, the depext would simply be dropped (and the package then fail to install, without a clear explanation). We could help, if these new variables are also available in the `available:` field, by adding:
```
available: [ !(distro = "centos" & distro-version < "7.2") ]
```
but that seems a little redundant. Another possibility would be to further extend the format proposed above, with for example triples `<sys-packages> <condition> <validation>`, where if `<condition>` holds `<packages>` are added, but `<validation>` must hold as well or it's a failure. E.g. 
```
["openssl-devel"] { distro = "centos" } { distro-version >= "7.2" }
```

AltGr:
This closes #2919

This replaces the current obscure and verbose depexts format by lists of
system package names and filters based on opam variables, e.g. :

    depexts: [
      ["libssl-dev"] { os-distribution = "debian" | os-distribution = "ubuntu" }
      ["openssl-devel"] { os-distribution = "centos" & os-version >= "7.2" }
    ]

The old format still parses, and is rewritten by the package migration
function (older packages in 2.0 format won't be rewritten and may get
broken depexts, as we don't want to retain compatibility with beta
format versions)

All tags currently present in `opam-repository` are supported, except
for `source`, which is discarded (they were in the same namespace and
now need to be classified between `arch`, `os`, etc.)

Note that the rewrite needs to know about the variables defined and
their values. This is based on #3058 and
https://gist.github.com/AltGr/5bfc8cea6f01e74b95de79ceaba39369, version
4, but may still change.

Opam itself does not define these variables at the moment (except for
`arch` and `os`, but not with the same normalisation yet), and the
command-line format changes too, which means that current `opam-depext`
will be broken by the changes. The new format is

    opam list --external --vars arch=x86_64,os=linux,os-distribution=debian

instead of

    opam list --external=x86_64,linux,debian

which was more ambiguous and less generic, if shorter. The idea is that
the variables will later be built in opam, so the user won't need to
declare them for normal usage.


dbuenzli:
On aspect that doesn't seem to be handled that was present in #2919 is the ability to specify a version for the *package*.

AltGr:
Well, I don't see how we could better handle sys package versions. What is already possible is, if the package manager supports it, to give a version within the sys-package string: for example, you could use

    "conf-mariadb=10.1.26-0+deb9u1" {os-distribution = "debian"}

I don't think it's worth having more built-in, as it would be difficult to map to the package managers, may not even be supported in most of the cases, and most PMs only have one version per package anyway.

dbuenzli:
This occurs quite often and is painful for CI, I still think it would be better to abstract this into `depext` functionality. This has the following advantage:

1. The `depext` specifier does not need to know about the specifics of all the system package managers syntax. The tool responsible for installing the `depext` can perform a best-effort job 
to satisfy the request and/or warn if it can't.
2. The tool responsible for installing `depext` can install the package and may be able to check that a specific version was installed and warn/error otherwise.

avsm:
Why can't this versioning be handled by `os-version` constraints?

e.g.

```
"conf-mariadb" {os-distribution = "debian" & os-version >= "8"}
```

The common case is that an upstream package is present in (e.g.) Ubuntu 17.04 and that needs to be encoded. The precise version would be nice, but not necessary for us to track.

AltGr:
I think using `os-version` is OK for now, and if we want finer support and manage to implement it for the various PMs managed by `opam-depext`, we could keep the same syntax but parse e.g. `"mariadb-client>=10.1"` — so that this could be done without modifying opam itself.

dbuenzli:
> Why can't this versioning be handled by os-version constraints?

I'm not sure about that, e.g. think of `homebrew` which has nothing to do with the actual operating system version.

But indeed the `depext` declaration may not be the best place to specify this (if only because it wouldn't be DRY), this may need solving at the `conf-*` layer where being able to ask for specific version still remains an open question AFAIR.

AltGr:
There is now a matching version of `opam-depext` here: https://github.com/ocaml/opam-depext/pull/81

avsm:
> I'm not sure about that, e.g. think of homebrew which has nothing to do with the actual operating system version.

Agreed, but homebrew is something of a special case since users are expected 'brew update' and keep on the master to solve compatibility problems. The significant version on osx is the base macos revision, which can be tracked by os-version.

I'm happy with the existing scheme in this PR; it's a huge improvement :-)

samoht:
That looks great to me too.

PR #3078, opened 2017-10-17
Remove check for glpk, that is now provided by ocaml-mccs

AltGr:


avsm:
Transient test failure. I'll go ahead and merge this to unblock the container builds. Thanks!

PR #3085, opened 2017-10-27
This adds build ids to packages and experimental binary caching

AltGr:


avsm:
Is there a description of how the binary caching works?  One thing that would be useful is to be able to compute a list of build-ids of the common-prefix of packages.

So I depend on:

A.1
B.0
C.2
D.3

then having the hash of B.0+C.2+D.3 and C.2+D.3 and D.3 so that I could find the nearest binary cache with the required subset of packages.  Then opam-ci could use that information to progressively store layered caches of package builds.

AltGr:
I am not sure what you mean exactly by «a list of build-ids of the common-prefix of packages», but the cache of each package would be stored independently¹. Since the binary cache describes precisely the state of your dependency cone, you are sure to get the same binary if you get the same version of all dependencies (themselves compiled with the same dependencies), including optional dependencies or their absence.

Of course, other approaches would be possible, and experimenting is the point of having this as plugins to start with. My [current prototype](https://github.com/OCamlPro/opam/blob/3564de52becf19d88b736917f7962d9adebc35e5/shell/opam-bin-cache.sh) is fairly trivial, but seems to work so far.

¹ yes, this assumes packages cleanly install new files, rather than patch (or overwrite) already installed ones. If they do, the order in which two packages are installed could matter, even without a dependency relationship between them (imagine e.g. two different packages adding a line to `findlib.conf`).

Note that packages without a stable source (e.g. pinned, or without an archive hash) will be without a build-id, as opam assumes them to be non reproducible.

PR #3089, opened 2017-11-10
Fix performance issue with the builtin solver on some requests

AltGr:
by tweaking the criteria (uses latest version of ocaml-mccs with
finer-grained criteria to do so without regression on the quality of the
results).


avsm:
+1 from me -- speeds up package installation considerably.

PR #3090, opened 2017-11-13
opamEnv: Replace `tty -s` with `test -t 0`

mc10:
`man tty` states that `tty -s` is deprecated.


AltGr:
... and your solution seems to be POSIX, too. Thanks!

PR #3097, opened 2017-11-16
Following #3058, this brings the system detection code in

### LINKED ISSUE #3058: RFC: variables to check for OS, distribution, etc.

AltGr:
This includes definition of the `arch`, `os`, `os-distribution`,
etc. variables, to be used anywhere in packages. The code of
`opam-depext` will also be much simpler, only the interaction with the
host package manager will remain.

Closes #3058

### DIFF
--- a/doc/index.html
+++ b/doc/index.html
@@ -196,6 +196,8 @@ td.sublib {
   <td>Defines the types holding global, repository and switch states</td></tr>
 <tr><th><a href="ocamldoc/OpamFormatUpgrade.html">opamFormatUpgrade.ml</a></th>
   <td>Handles upgrade of an opam root from earlier opam versions</td></tr>
+<tr><th><a href="ocamldoc/OpamSysPoll.html">opamSysPoll.ml</a></th>
+  <td>Detection of host system (arch, os, distribution)</td></tr>
 <tr><th><a href="ocamldoc/OpamGlobalState.html">opamGlobalState.ml</a></th>
   <td>Loading and handling of the global state of an opam root</td></tr>
 <tr><th><a href="ocamldoc/OpamRepositoryState.html">opamRepositoryState.ml</a></th>
--- a/src/client/opamAction.ml
+++ b/src/client/opamAction.ml
@@ -375,9 +375,10 @@ let make_command st opam ?dir ?text_command (cmd, args) =
     OpamProcess.make_command_text name ~args cmd
   in
   let context =
+    let open OpamStd.Option.Op in
     String.concat " | " [
       OpamVersion.(to_string current);
-      OpamStd.Sys.os_string () ^"/"^ OpamStd.Sys.arch ();
+      (OpamSysPoll.os () +! "unknown") ^"/"^ (OpamSysPoll.arch () +! "unknown");
       (OpamStd.List.concat_map " " OpamPackage.to_string
          OpamPackage.Set.(elements @@
                           inter st.compiler_packages st.installed_roots));
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -996,7 +996,9 @@ let build_options =
        (or the deprecated $(b,\\$OPAMBUILDDOC)) to \"true\"." in
   let make =
     mk_opt ~section ["m";"make"] "MAKE"
-      "Use $(docv) as the default 'make' command."
+      "Use $(docv) as the default 'make' command. Deprecated: use $(b,opam \
+       config set[-global] make MAKE) instead. Has no effect if the $(i,make) \
+       variable is defined."
       Arg.(some string) None in
   let show =
     mk_flag ~section ["show-actions"]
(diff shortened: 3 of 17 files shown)


avsm:
awesome!

avsm:
I'm just wondering if we need opam-depext as a plugin at all.  Could we instead do something with pre-install hooks and helper binaries that check for OS dependencies?  That would avoid all the duplication of `opam pin add -n && opam depext && opam install` that we have at the moment.

AltGr:
Yes, that would be nice, but I still prefer to go one step at a time. Being able to update and fix `opam-depext` independently of opam has been quite valuable.

See the plans at https://github.com/ocaml/opam/blob/master/doc/design/depexts-plugins : proper interconnection would require opam to check all available and installed system packages very fast, so that `depexts` can both affect availability and trigger installation of the system packages. The interface, at the moment, is much more limited as depexts don't have an effect on availability, and aren't checked dynamically.

So, definitely, yes, but I prefer to wait for a good design and proper integration before taking the next step. This is already an important step in this direction!

avsm:
Sounds good, and completely agreed! Thanks for integrating this.

PR #3105, opened 2017-11-21
Allow automatic depext handling as a pre-session hook

AltGr:
Also improves variables so that they can expand to argument lists
Also provides additional variables to session hooks


AltGr:
Also includes many fixes that were already part of #3104, ignore them.

PR #3106, opened 2017-11-21
Add variables to session hooks, allow lists in variables

AltGr:


AltGr:
ah, this was meant to be merged before #3105, which would have been rebased on it. Nevermind.

PR #3109, opened 2017-11-22
In commands, resolve undefined to "" rather than no argument

AltGr:
After some discussion with @dra27, it seems in general safer to hold the 
positional argument with an empty string rather than shift the possible 
remaining arguments.

So for example, with

    build: ["some-command" foo "bar"]

Previously, if `foo` wasn't defined, that would become `["some-command"
"bar"]`. To avoid that, you had to use a string interpolation such as:

    build: ["some-command" "%{foo}%" "bar"]

which is a little heavy under the fingers. With this patch, the behaviour
would be the same.

Obtaining the previous behaviour is still possible by using an explicit 
filter:

    build: ["some-command" foo {?foo} "bar"]

which makes it explicit that the positional argument might be present or not

### DIFF
--- a/doc/pages/Manual.md
+++ b/doc/pages/Manual.md
@@ -763,6 +763,9 @@ files.
       detect that the package is not installed from a release tarball, and may
       need additional preprocessing (_e.g._ `automake`).
 
+    If a term is undefined (_e.g._ an undefined variable), the empty string is
+    used as positional argument.
+
 - <a id="opamfield-install">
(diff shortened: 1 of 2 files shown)


AltGr:
@dra27, I'd like your opinion on this :)

Drup:
IIRC, the documentation says that `foo` and `"%{foo}%"`, so I agree with @dra27 : They should actually be equivalent.

PR #3129, opened 2017-12-03
opam-list: explain in man why some listed packages are underlined

maelvalais:
In the 1.2.2 manual, `opam help list` held the following:

    DESCRIPTION
       This command displays the list of installed packages when called
       without argument, or the list of available packages matching the given
       pattern.

       Unless the --short switch is used, the output format displays one
       package per line, and each line contains the name of the package, the
       installed version or -- if the package is not installed, and a short
       description. In color mode, root packages (eg. manually installed) are
       underlined.

In opam 2.0, the underlined mention has disappeared:

    DESCRIPTION
       List selections of opam packages.

       Without argument, the command displays the list of currently installed
       packages. With pattern arguments, lists all available packages matching
       one of the patterns.

       See section PACKAGE SELECTION OPTIONS for all the ways to select the
       packages to be displayed, and section OUTPUT FORMAT OPTIONS to
       customise the output format.

I propose to re-add the explanation for the underlined packages (as you wish, of course!).

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -319,8 +319,8 @@ let list ?(force_search=false) () =
     `P "Unless the $(b,--short) switch is used, the output format displays one \
         package per line, and each line contains the name of the package, the \
         installed version or `--' if the package is not installed, and a short \
-        description. In color mode, root packages (e.g., manually installed) \
-        are underlined.";
+        description. In color mode, manually installed packages (as opposed to \
+        automatically installed ones because of dependencies) are underlined.";
     `P ("See section $(b,"^selection_docs^") for all the ways to select the \
          packages to be displayed, and section $(b,"^display_docs^") to \
          customise the output format.");
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -316,6 +316,11 @@ let list ?(force_search=false) () =
     `P "Without argument, the command displays the list of currently installed \
         packages. With pattern arguments, lists all available packages \
         matching one of the patterns.";
+    `P "Unless the $(b,--short) switch is used, the output format displays one \
+        package per line, and each line contains the name of the package, the \
+        installed version or `--' if the package is not installed, and a short \
+        description. In color mode, root packages (e.g., manually installed) \
+        are underlined.";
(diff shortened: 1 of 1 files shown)


AltGr:
Tests fail because of an unescaped end-of-line. Otherwise I'm good with the change, thanks!

maelvalais:
Sorry for that 😄 Fixed!

maelvalais:
Maybe we could also re-add the explanation on the displayed columns like in 1.2.2, i.e.,

    Unless the --short switch is used, the output format displays one
    package per line, and each line contains the name of the package, the
    installed version or -- if the package is not installed, and a short
    description. In color mode, root packages (eg. manually installed) are
    underlined.

What do you think?

AltGr:
The reasons to remove that last part were that ① there are now column headers and ② they are now documented in the `OUTPUT FORMAT OPTIONS` section, below the `--columns` option. However, I admit the latter is a bit hidden down in the manpage...

maelvalais:
Oh... you are totally right, I didn't look enough to find the `--columns=` default values. Would be really helpful if a short reminder of the default columns were displayed on top 😃 

By the way, what are _root_ packages? I can't find what this means 😕
Is it the "OPAM root" mentionned in the [FAQ](https://opam.ocaml.org/doc/FAQ.html#What-changes-does-OPAM-do-to-my-filesystem)?

AltGr:
Manually installed packages, as opposed to automatically installed ones (i.e. ones installed because of dependencies). Think "GC roots" for `opam remove --auto-remove`. Indeed, the manpage for `opam list` would be a good place to better explain that :).

maelvalais:
Using the tree analogy where the root is one user-installed package, I understand much better the meaning!!! It is kind of confusing though because the rest of the documentation mentions the "root" as to mean `OPAMROOT`

What about this description:

    In color mode, manually installed packages (as opposed to automatically installed ones
    because of dependencies) are underlined.

AltGr:
Sounds good, thanks a lot!

maelvalais:
Maybe we could also add some extra information for `--column`? The `COLUMNS` argument kind of lacks details, for example the colorization of `name` or the fact that `installed-version` is `--` when the package isn't installed

(nvm, the PR is closed now 😄 )

PR #3135, opened 2017-12-12
install.sh: Fix use of "du" to always use -k

mato:
Fixes the case where on some systems "du" prints the summary with the
unit (K) appended, causing an error like the following:

    ./install.sh: 213: [: Illegal number: 6593020K


AltGr:
Thanks a lot!

PR #3146, opened 2018-01-01
Fix 4.06 compatibility

tuncer:
This boils down to three src_ext changes:

1. Switch to 4.06-compatible ocamlgraph-1.8.8 release.
2. Drop Bytes->String patch for re.
3. Add Damien's safe-strings patch for dose3.
   Source:
   https://gforge.inria.fr/tracker/index.php?func=detail&aid=21552&group_id=4395&atid=13810
   https://gforge.inria.fr/tracker/download.php/4395/13810/21552/6082/0001-fixes-for-4.06-safe-string.patch

PR #3150, opened 2018-01-02
OCaml 4.06.0 updates

dra27:
Apparently at least two of us were working on this! This builds on #3146, but also:
 - Adds `Unix` to `OpamCompat` (to make `Unix.map_file` available for pre-4.06.0 builds)
 - Updates regression testing to 4.06.0 (see https://github.com/ocaml/opam-rt/pull/53)
 - Updates `make cold` to 4.06.0
 - Updates the other src_ext packages to their latest versions

Note that the re bytes patch should have been removed when we switched to Jbuilder, as 4.02.3 became the minimum supported version at that point.


AltGr:
Thanks! I #3146 still needed if we merge this ?

AltGr:
According to Travis the build on mac got compiled without the built-in solver ?

dra27:
I think this PR includes everything done in #3146, yes. I'm just having a look at the Travis result.

dra27:
Ah, I can see what's going on. `OPAM_TEST=1` means that opam should be built without lib-ext but mccs is failing to build because of https://github.com/janestreet/jbuilder/pull/393. opam's configure then notices that it has all dependencies except mccs and so continues without the solver.

I've just pushed another commit which tweaks the lib-ext warning.

dra27:
OK, so now the log suggests that opam should be built with an internal solver (and this is forced by the CI) but then just doesn't do it. The next commit I've added temporarily pins jbuilder (I'm using this as a test case for 393 and I'll probably then merge it and update the pin to janestreet/jbuilder#master).

Incidentally, it would be possible to tweak the build system so that the build of opam would actually fail if you specify `--with-mccs` and mccs is not available (i.e. you don't run `make lib-ext`). The same fix would also allow specifying `--without-mccs` when mccs is installed, however it involves adding various opam-solver sub-packages set-up by configure so as it would just be for cosmetic corner-cases, I think it better to wait until we implement https://github.com/janestreet/jbuilder/issues/397 in Dune.

dra27:
I'm expecting Travis to pass - AppVeyor just needs kicking

AltGr:
Got tired of kicking, so assuming it's still OK :)

PR #3151, opened 2018-01-02
Ensure ocamlfind is from a compatible installation

dra27:
A minor niggle, but rather annoying doing the Windows testing when configure picks up another installation's ocamlfind and assumes there are working packages!

The changes are prepared for upstream in https://github.com/dra27/ocaml-autoconf/tree/various-improvements and noted on [the OCaml Forge](https://forge.ocamlcore.org/tracker/index.php?func=detail&aid=1471&group_id=69&atid=363), but ocaml-autoconf isn't very actively maintained.

### DIFF
--- a/configure
+++ b/configure
@@ -604,6 +604,7 @@ LN_S
 JBUILDER
 FETCH
 OCAMLFIND
+OCAMLOBJINFO
 PATH_PREPEND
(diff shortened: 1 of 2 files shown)


AltGr:
Thanks!

PR #3152, opened 2018-01-03
Silence errors in git rev-parse command

marsam:
Hi:
When opam is built in an isolated environment `git rev-parse` which pollutes the build output


AltGr:
Thanks!

PR #3155, opened 2018-01-03
Windows bootstrap

dra27:
This pull request primarily adds support for Windows to the opam bootstrap scripts and uses these to extend AppVeyor testing to all six Windows ports. The Cygwin ports are switched to use the bootstrap compiler rather than the Cygwin-installed ocaml package.

There are a few minor tweaks included on the way:
 - Since beta15, `jbuild-ignore` files can no longer contain subdirectories (see https://github.com/janestreet/jbuilder/pull/389). As it happens, generating `src_ext/jbuild-ignore` is helpful for another pull request due to land...
 - I think that when I originally changed `MCCS_DISABLED` to `MCCS_ENABLED` there was a definite need for it, but either way - "if not disabled then do stuff" conditions do read oddly in the Makefile versus "if enabled then do stuff", so I've left it!
 - There's a small tweak to `shell/check_linker` (for MSVC building) as I discovered during AppVeyor testing that you get some irritating stderr junk from Git's link command.

The major stuff...

## Automatic detection of Microsoft Visual Studio
This utilises my [MSVS Tools](https://github.com/metastack/msvs-tools) package and means that you don't need to start Cygwin from within a Visual Studio Command Prompt to build an MSVC OCaml. The motivation for this becomes apparent in the full Windows port, since opam needs both a 32-bit and a 64-bit compiler and sadly no one ever told Microsoft about compiler triples...

## `make cold` updates
Windows support is added for `make cold` - it's necessary to be able to specify which port you want built, which is either inferred automatically or specified using the `OCAML_PORT` variable passed to `make`. The script becomes quite a bit longer since it also supports automatic detection of the Microsoft C compiler, and the Windows ports of OCaml lack a configure script and need FlexDLL to be extracted.

The configure script is enhanced so that if a bootstrap compiler has been built then configure always uses it. This means you can *develop* opam using a bootstrap compiler, which has been quite handy for doing the Windows ports.

The final tweak to the `make cold` process is that it also supports being run "offline" which meant moving the location of the OCaml and FlexDLL tarballs to `src_ext/Makefile`

## AppVeyor
I've transferred over some of the scripts I did for OCaml's AppVeyor runs which reduces the time spent in Cygwin setup. Various notes on this:
 - All six ports are built, but the limited test for `opam init` and installation of `findlib` is only done on Cygwin, since it won't work on native Windows without more patches (which will follow very, very soon...)
 - `make cold` (or, rather, the new `make compiler`) is used to get OCaml (for uniformity, this is done for the Cygwin ports, rather than using Cygwin-packaged OCaml). The bootstrap folder (and also the src_ext archives) are cached, so each build only takes a couple of minutes, unless the OCaml version has been bumped.

Compilation on native Windows is dependent on three external patches which are at present applied using appveyor.patch:
 - ocaml-mccs 1.1+5 (the merged https://github.com/AltGr/ocaml-mccs/pull/3 is required - earlier versions don't build on Windows)
 - jbuilder/dune beta17 (the merged https://github.com/janestreet/jbuilder/pull/353 is required - necessary to compile ocaml-mccs with the Microsoft C compiler)
 - FlexDLL 0.38 (which will hopefully include https://github.com/alainfrisch/flexdll/pull/48 - necessary to link with ocaml-mccs in the mingw ports)


AltGr:
Very nice, thanks! Do you think we should merge as-is, or wait for the releases of mccs and "dune" (and possibly FlexDLL) ?

dra27:
The 4.06 changes should go in first and I'll then rebase this for 4.06 (I've already tested it with 4.06). I would say this can then be merged - the jbuilder and ocaml-mccs changes are just awaiting releases and the flexdll change only affects one set of Windows ports, not all three sets, so it's all usable work even without the patch.

dra27:
Hopefully correctly rebase for 4.06 - let's see what AppVeyor thinks!

PR #3164, opened 2018-01-10
Disable caching in OpamSystem.resolve_command

dra27:
`OpamSystem.resolve_command` caches results. This cannot work - the ocamlfind package, for example, installs an `ocaml` script which is intended to wrap the system-installed `ocaml` command. This script is not invoked if `ocaml` has been previously executed, because the path to the system-installed `ocaml` command is cached.

There are various shims which could mitigate this for opam-installed binaries/scripts only, but it seems better just not to cache the results at all.

### DIFF
--- a/src/core/opamSystem.ml
+++ b/src/core/opamSystem.ml
@@ -369,21 +369,8 @@ let resolve_command =
       | _ -> None
     else None
   in
-  let cached_results = Hashtbl.create 17 in
   fun ?(env=default_env) ?dir name ->
-    if dir <> None && is_external_cmd name && Filename.is_relative name then
-      resolve env ?dir name (* no caching *)
-    else
-    let path = env_var env "PATH" in
-    try Hashtbl.find (Hashtbl.find cached_results path) name
-    with Not_found ->
-      let r = resolve env ?dir name in
-      (try Hashtbl.add (Hashtbl.find cached_results path) name r
-       with Not_found ->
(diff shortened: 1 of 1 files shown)


AltGr:
Nicely spotted!
The cache was indeed initially planned for commands used internally (e.g. `tar`, `patch`, `git`, etc.) where it really should be harmless (and makes the output of `strace` a bit tidier). But yes, applying it to packages' command scripts is a big mistake indeed.

Could it be worth adding an optional argument to disable the cache only for commads from scripts ?

dra27:
I *think* it's correct to view it as an unnecessary optimisation (i.e. running those commands will take vastly more time than locating them). Personally, I'd drop it completely - it's already been a subtle issue, and it feels like it could be a subtle one in future (imagine if/when ocaml-git starts installing a `git` command and we expect that to be used).

If opam's internal invocations want to cache the location of `tar`, `patch`, `git` and so forth then I think it should do that at `opam init` time and record the location of the tool in a config file/cache (invalidated only if the tool subsequently can't be found). i.e. it should be more like a configurable option (like `MAKE=make`, etc.)

avsm:
If this PR could be merged soon, it'll make a huge usability difference to me. Right now local switches are very unreliable on the Mac with a system switch due to this.

AltGr:
Yup, merging, I'll consider re-adding for non-script commands later. Thanks!

PR #3165, opened 2018-01-10
travis-ci: test with 4.07.0+trunk+flambda

tuncer:


tuncer:
Let's see if 4.07 trunk works. Also enable flambda to test that at least for the OCaml development branch.

tuncer:
Will have to rebase this once #3150 is merged tomorrow, since 4.07 is unlikely to work if 4.06 isn't. So, to clarify, I expect this to fail for the moment.

dra27:
I'm not convinced this is a very good idea, at least at the moment (at Cambridge, we're in the process of throwing vast amounts of compute power at CI, but it's still in development). OCaml trunk is not expected to be tracked by projects at all times, which means it at least needs to be a permitted failure in Travis... at that point, the test is largely a waste of CPU cycles since allowed failures are not reported very well in the GUI.

This might be worth doing after the feature-freeze for 4.07 in March, but, ignoring homebrew's somewhat naively eager package upgrade policy, I'm not sure why opam has such urgent need to be buildable with 4.07 on release day. It will definitely be able to *install* it release day, because the compiler update to opam-repository is part of the OCaml release procedure.

tuncer:
> I'm not sure why opam has such urgent need to be buildable with 4.07 on release day

I don't have a stronger argument to get this merged than that I want to help avoid a repeat of the 4.06 delay. If we will achieve that differently (the new CI) and this addition is a bad idea, let's close it.

> It will definitely be able to install it release day, because the compiler update to opam-repository is part of the OCaml release procedure.

Do you mean using a 4.06 switch to build the opam tree revision of 4.07.0 release day?

dra27:
> I don't have a stronger argument to get this merged than that I want to help avoid a repeat of the 4.06 delay. If we will achieve that differently (the new CI) and this addition is a bad idea, let's close it.

Travis CI here won't really help, other than telling us in a not very helpful/visible way that it's not working. opam has 3 external dependencies of its own (ocaml-mccs, opam-file-format and opam-rt) and 9 external dependencies on other projects.

4.06.0 was an unusual release because of the safe-string switch. I anticipate 4.07.0 will be a little quieter.

Don't get me wrong, though - having as many packages actually ready and working with new releases of the compiler is not only a good goal, but one we're working towards. I just don't think this is the way to achieve it.

> Do you mean using a 4.06 switch to build the opam tree revision of 4.07.0 release day?

No, I mean that opam compiled with any supported version of OCaml will be able to give you a switch with OCaml 4.07.0 in it - I'm not clear what benefit you're after from being able to compile itself using OCaml 4.07.0 on release day.

tuncer:
I don't know the details of the new CI, but from what I understand, we won't have the same delays again, so let's close this.

> I'm not clear what benefit you're after from being able to compile itself using OCaml 4.07.0 on release day

If you try to use OCaml and opam on a platform that isn't supported with binaries, and you want to use the newest opam with the current OCaml release, then you might need to bootstrap a prior OCaml (or get it from package repo), then bootstrap the needed opam version, to finally use that to opam switch the current OCaml version.

If one is lucky, one can find an easier way to achieving this with an existing install of that operating system. OCaml and opam have so far been one of the easier (like Erlang) toolchains to bootstrap on random platforms. I wouldn't want to miss that.

I must admit it's not clear to me why there's no commitment to having opam build out of the box on release day, if the new CI will catch the issues early. It seems I'm missing a detail or getting something wrong.

avsm:
Just before this is closed -- one option is to make it an 'allowable failure' : see https://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail

The good news is that this will test 4.07 as it heads towards feature freeze, but not fail the whole PR if 4.07 quite reasonable has some transient glitches. The bad news is that I have no idea how to coerce the travis.yml into making just one entry in the row an allowed failure :-)

tuncer:
@avsm, are you saying we should reopen and retry this?

Issue #3161, opened 2018-01-08
Built-in solver limitations


AltGr:
I've just managed to hit another explosion of the internal solver. Context to reproduce:
- Fresh `ocaml-system.4.02.3` switch
- opam at 6a5779a45f6f05290207f0efc7938df6e7b775a2 (just after beta6, 2017-12-20) with built-in solver
- repo as of dec. 18th (083c78814442f463b121c9c43a047907543ab69d)
- just run `opam install github`. The more precise `opam install github.2.3.0` works as well.

Some notes:
- aspcud solves it without problem
- removed the depopts, conflicts, same result
- actually the solution is not that far away: if we explicit the direct dependencies in the request:
  ```
  opam install 'github.2.3.0' tls 'uri>=1.9.0' 'cohttp>=0.17.0' 'lwt>=2.4.4' 'atdgen>=1.10.0' 'yojson>=1.2.0' 'stringext' 'lambda-term' 'ocaml>=4.02.3' 'cmdliner>=0.9.8' 'base-unix' 'ocamlfind' 'ocamlbuild'`
  ```
  this is solved easily by mccs as well. The following workaround even works:
  ```
  opam install github $(opam list -s --required-by github.2.3.0)
  ```
- Actually this can in this case be reduced to just `opam install github cohttp`. There are currently 48 releases of cohttp in the repo. maybe that is a little excessive ?

The above may seem weird but is actually easy to understand if you look closely at the criteria
```
-removed,-count[version-lag,request],-count[version-lag,changed],-changed
```
The only effect of adding the direct dependencies of `github` to the request is to the second criterion here: the solver will optimise the versions of the requested packages first. In the first case, `opam install github.2.3.0`, this criterion can be completely ignored, so we basically fall back to the next one. In the expanded case, the versions of the direct dependencies are optimised first, which apparently drives us to the solution relatively easily.

The next criterion, `-count[version-lag,changed]`, in fact applies to all packages since we are on a fresh switch. That makes a lot, and the resulting formula is probably too big to be well handled by glpk. In fact, in our case the criteria would be equivalent to:
```
-count[version-lag,solution],-new
```
since on our fresh switch, nothing is allowed to be removed, the requested package is fixed and any change has to be an addition.

[...]

PR #3170, opened 2018-01-12
Make `version-lag` cubic

AltGr:
This helps the solver a lot by penalising older versions more (and thus
making the decision of which package to roll back easier), and solves,
albeit not by a large margin, #3161.

### DIFF
--- a/src/solver/opamCudf.mli
+++ b/src/solver/opamCudf.mli
@@ -164,7 +164,7 @@ val s_installed_root: string
 (** true if the package is pinned to this version *)
 val s_pinned: string
 
-(** the number of versions of the package since this one*)
+(** the number of versions of the package since this one, cubed *)
 val s_version_lag: string
 
 (** {2 Pretty-printing} *)
--- a/src/solver/opamSolver.ml
+++ b/src/solver/opamSolver.ml
@@ -185,7 +185,8 @@ let opam2cudf universe ?(depopts=false) ~build ~post version_map package =
         (fun v (i,r) -> if v = version then i,i else i+1, r)
         all_versions (0,0)
     in
-    count - i
+    let linear_lag = (count - i) in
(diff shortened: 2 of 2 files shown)


AltGr:
Note that this, obviously, introduces a bias: it'll make opam favor a solution with everything a little behind over a solution with a single package at a very old version, but the rest at their latest. Not sure which is best, or if we can find real-world cases where this matters, though.

I've tried applying an opposite bias (version-lag very penalising at first, then amortised), but it didn't really help the solver as well.

Drup:
That's an interesting solution. It also confirms I have no intuition on how these kind of solvers work at all, even after having used a MAX-SMT solver in my research. :D 

> it'll make opam favor a solution with everything a little behind over a solution with a single package at a very old version, but the rest at their latest.

I do not think that is a problem in practice. If you really want the last version of something, you can ask for it explicitly. And having a solver that returns solutions to begin with is much more important!

Could we have a lag that is proportional to the number of versions ? Having 2 version late in a package that has 20 versions is not as problematic than a package that has three. This would take care of annoying packages like cohttp without being much of a problem for others.

AltGr:
> Could we have a lag that is proportional to the number of versions ? Having 2 version late in a package that has 20 versions is not as problematic than a package that has three. This would take care of annoying packages like cohttp without being much of a problem for others.

I have been thinking about it, but it's actually hard to have something that makes sense, so better keep the metric as simple as possible. A package with 3 versions may have very small patches between them, or they might be years apart. This is something that Roberto pointed me to first when I mentionned the need to reduce the "age" of packages: it's not clear exactly how to measure the age. Time since release ? Number of versions ? Some fancy semver metric ? Basically it's not a problem if we don't really care about the precise outcome of cases where the metric does indeed change the result, but we need to keep it in mind.

PR #3184, opened 2018-01-19
Simplify OpamCompat module

dra27:
Two leftover artefacts from the Jbuilder conversion:
 - 4.01 isn't supported by Jbuilder, so there's no need for the compatibility layer
 - `-safe-string` was introduced in 4.02.0, so there's no need for it to be provided as a context-specific flag

The generation of the OpamCompat module is starting to get tedious (I'm just having to back-port Buffer.add_utf_8_uchar which, even after removing 4.01, has to be duplicated in 3 files). cppo is already used in lib-ext, so opamCompat.ml* refactored to use a cppo and a `per_module` `preprocess` stanza.

GitHub makes this PR look much more impressive than it really is!


avsm:
We're generally seeing more issues due to `opam` itself being compiled with different versions of OCaml (particularly the execve issue).  Is it about time to just pin opam compilation to a narrower set of versions in opam2, and encourage the use of the `make cold` bootstrap for when there is a mismatch?

dra27:
The only issue might be limiting users of the libraries, rather than the binary... but I don't personally see an issue with constraining either just opam-client or even just the binary itself.

AltGr:
The 4.01 compat layer was indeed no longer needed at all, so good thing to simplify. I am not completely fond of using `cppo`, but well, it does indeed make the thing more DRY, easier to figure and less error-prone, so it seems like a good move, esp. since it's limited to just one file.

Tests failures are due to opamRT not compiling with the small change in interface, I'll see to it.

Thanks!

dra27:
I agree - it'd be nice not to have to make it a hard dependency, but it does seem better than so much repetition!

Ta for dealing with opam-rt - I hadn't had a chance.

PR #3185, opened 2018-01-19
Update README with C++ compiler information

lmouelle:
Running `make` after a `make lib-ext` fails with `error gcc: error trying to exec 'cc1plus': execvp: No such file or directory` when compiling 
`changed_criteria.cpp`. Adding documentation stating gcc-c++ must be installed in system (some distros, e.g. Fedora 27, don't have it in a fresh install).

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,7 @@ Either from an existing opam installation, use `opam pin add opam-devel
 
 * Make sure you have the required dependencies installed:
   - GNU make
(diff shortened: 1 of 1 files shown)


dra27:
This isn't quite right - it's not a required dependency, but you need to configure with the internal solver disabled. However, given that autoconf can of course detect a C++ compiler, we should automate this. There's also an old reference to needing glpk itself

PR #3187, opened 2018-01-23
Update to Jbuilder 1.0+beta17

dra27:
I've updated the opam files as well, given that Windows requires beta17.


dra27:
Both CIs obviously need to pass here, and https://github.com/ocaml/opam-repository/pull/11282 should also be merged first.

dra27:
Jbuilder release pulled

PR #3194, opened 2018-01-26
Allows in-source metadata to sit in opam/, and other fixes

AltGr:


doc/pages/Manual.md
```diff
@@ -965,11 +971,15 @@ files.
...
 
-- <a id="opamfield-setenv">`setenv: [ <environment-update> ... ]`</a>:
-  defines environment variables updates that will be applied upon installing the
+- <a id="opamfield-setenv">`setenv: [ <environment-update> ... ]`</a>: defines
+  environment variables updates that will be applied upon installing the
   package. The updates will be visible to any dependent package, as well as
-  exported to the shell environment through `opam env` and
-  `~/.opam/opam-init/` scripts.
+  exported to the shell environment through `opam env` and `~/.opam/opam-init/`
+  scripts. Note that while it is guaranteed that dependents will see the
```
dbuenzli: This is still a bit unclear to me. Specifically is `opam env` (+ user env at the moment)  the environment in which builds happen on an `opam install` ? I.e. do all packages installed *after* the package that `setenv:`s see the environment update ?

AltGr: It's unspecified. Actually, currently, in practice, both can happen, depending on wether the package doing the `setenv:` is modified during the same run or not (if it's untouched, you'll see the `setenv:`, otherwise not). Basically, builds happen with `opam env` _as if_ all the packages that are modified during the run were removed, + the `setenv:`s of their dependencies.

dbuenzli: > It's unspecified.

I don't think this is a good idea. Especially since for now `opam` still mixes the user environment in an `opam install` people might get different builds whether they `eval $(opam env)`'d or not before doing an `opam install`. I think that given how `opam` works now (no user environment isolation) the only reasonable semantics is that the `setenv` of all installed packages is available in *subsequent* installs.

This also enables to have "environment packages" that allow to setup environment without other packages necessarily being aware of them (these things would typically be installed a switch creation time or could be mandated to be so).

AltGr: > Basically, builds happen with `opam env` _as if_ all the packages that are modified during the run were removed, + `the setenv:`s of their dependencies.

This actually _doesn't_ depend on wether `opam env` was evaluated in the meantime, it will remain true even if it hasn't
(modulo the imperfection of reverting env changes if it has, but it's a separate issue)

dbuenzli: I don't understand, `opam install` tries to fiddle with the user env beyond the addition of `opam env` ?

AltGr: More or less: `opam install` will internally revert the changes to the environment, then re-apply them according to the context.
This is very much what you want when running across switches:
```
eval $(opam env --switch foo)
opam install --switch bar somepkg
```
here we don't want the environment for the build of `somepkg` to be tainted by the `setenv:`s of unrelated switch `foo`. But this remains true even when in the same switch.

PR #3206, opened 2018-02-04
Force git diff to not use external diff

artemkin:
`opam update` failed several times due to external git diff driver.

I use:
```
GIT_EXTERNAL_DIFF=/home/user/.opam/4.05.0/bin/patdiff-git-wrapper
```

```
...
00:01.000  PARALLEL                Collected task for job 0 (ret:0)
00:01.000  PARALLEL                Next task in job 0: /usr/bin/git add .
00:01.005  PARALLEL                Collected task for job 0 (ret:0)
00:01.005  PARALLEL                Next task in job 0: /usr/bin/git diff -R -p refs/remotes/opam-ref --
00:01.010  PARALLEL                Collected task for job 0 (ret:128)
00:01.010  SYSTEM                  rm /home/user/.opam/log/git-diff-19973-5f1023
00:01.010  SYSTEM                  error: Git error: refs/remotes/opam-ref not found.
```

```
$ opam --version
2.0.0~beta6
```

### DIFF
--- a/src/repository/opamGit.ml
+++ b/src/repository/opamGit.ml
@@ -143,7 +143,7 @@ module VCS : OpamVCS.VCS = struct
     (* Git diff is to the working dir, but doesn't work properly for
        unregistered directories. *)
     OpamSystem.raise_on_process_error r;
-    git repo_root ~stdout:patch_file [ "diff" ; "-R" ; "-p" ; rref; "--" ]
+    git repo_root ~stdout:patch_file [ "diff" ; "--no-ext-diff" ; "-R" ; "-p" ; rref; "--" ]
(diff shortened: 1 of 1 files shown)


AltGr:
Thanks a lot! Just one thing to check is that the option won't break compatibility with older versions of `git`, could you do that ? Thanks again.

artemkin:
The question is how to check it. I see this option mentioned in 1.6.0.5 release note (it was 8 years ago). https://github.com/git/git/blob/master/Documentation/RelNotes/1.6.0.5.txt

AltGr:
Ok, this is largely good enough. Thanks a lot!

PR #3215, opened 2018-02-07
Add 'opam admin check' to check installability and cycles

AltGr:


AltGr:
Cycle detection needs some more thought: strongly component analysis can still lead to false positives. Indeed, we now remove all dependency edges when there is a direct conflict between the packages, but this is not done for indirect conflicts (i.e. conflicts to a dependency).

For example, if we have a cycle `a → b → c → d → a`, a conflict between `a` and `c` would be enough to break the cycle and avoid the problem, but won't be detected at the moment (and this is even true if `a` and `c` are two different versions of the same package, which is also a conflict).

AltGr:
It's actually very hard to do better: enumerating and filtering cycles would work, except it gets too large to be realistic (we quickly get the cartesian product of all the versions of each package in the dependency chain).
I tried to find other approaches, but SCC doesn't apply anymore since, if we want to exclude cycles with conflicts, the result is not even a partition anymore: the same package can be in two cycles that are not compatible. And we would still get a cartesian product issue since all versions of the same package are mutually conflicting.
The only remaining possibility would be to reduce the graph first, maybe by combining all versions of a package that have the exact same edges. I am not sure if even that would be enough though.

Drup:
@rdicosmo ? :)

AltGr:
> The only remaining possibility would be to reduce the graph first, maybe by combining all versions of a package that have the exact same edges. I am not sure if even that would be enough though.

I tried that :) (I intend to reuse the code to cleanup the repo, so it was worth it). On the case with all test/doc dependencies and without the fixed `jbuilder`, my 16GB are still not enough :grin:

PR #3216, opened 2018-02-08
Ensure a C++ compiler exists if MCCS is enabled

dra27:
MCCS will be automatically disabled if no C++ compiler is found; an error will occur if `--with-mccs` is specified and no C++ compiler is present.


AltGr:
This is great, thanks!

AltGr:
Apparently this broke builds on opam.ocaml.org, I'm investigating.

AltGr:
Ah no, it seems it's due to escaping and an older `jbuilder` version.

dra27:
Ah, it's that - yes, beta17 is now a real minimum required version, not just a meta-data update!

Issue #3197, opened 2018-01-27
List of tools used by opam


jpdeplaix:
Is there a list available somewhere of the tools that could be used by opam ? (e.g. rsync, tar, patch, make, git, …)

dra27:
Sort of: https://github.com/ocaml/opam/blob/master/src/client/opamClient.ml#L550-L552 and https://github.com/ocaml/opam/blob/master/src/client/opamClient.ml#L601-L609

jpdeplaix:
Nice, that what I was looking for ! Thanks a lot ! :)

AltGr:
I was going to point you to the same place :) . That should be in the README or manual though, but it's so hard to keep them up-to-date!

PR #3217, opened 2018-02-09
[WIP]  Move most of the external tools calls to a single module 

jpdeplaix:
This PR aims at improving the way of describing and looking for external programs calls. This should also ease portability checks for Windows (cc @dra27) and minimal distributions such as Alpine Linux for example by checking which packages to install, what options are not available on the distribution/OS you want to port opam on.

Related to  #3197

This is WIP because of some ```assert false``` to fill up in ```src/format/opamFile.ml``` and calls to ```OpamExternalTools.custom``` for Git/Hg/Darcs/…
Also I would know if the change to the API is ok before continuing further.

### DIFF
--- a/src/core/opamExternalTools.ml
+++ b/src/core/opamExternalTools.ml
@@ -8,96 +8,118 @@
 (*                                                                        *)
 (**************************************************************************)
 
+type t = (string * string list)
+
+let unpack x = x
+let custom x = x
+
+let has_space (cmd, _) = String.contains cmd ' '
+let cmd_to_string (cmd, _) = cmd
+
+let to_string (cmd, args) =
+  (* TODO: Add backslashs to quotes for each args *)
+  let quote x = if String.contains x ' ' then "'" ^ x ^ "'" else x in
+  let args = List.map quote args in
+  String.concat " " (cmd::args)
+
 module Unzip = struct
-  let extract cmd ~file ~dir = cmd "unzip" [file; "-d"; dir]
+  let extract ~file ~dir = ("unzip", [file; "-d"; dir])
 end
 
 module Tar = struct
-  let extract_gzip cmd ~file ~dir = cmd "tar" ["xfz"; file; "-C"; dir]
-  let extract_bzip2 cmd ~file ~dir = cmd "tar" ["xfj"; file; "-C"; dir]
-  let extract_xz cmd ~file ~dir = cmd "tar" ["xfJ"; file; "-C"; dir]
-  let extract_lzma cmd ~file ~dir = cmd "tar" ["xfY"; file; "-C"; dir]
+  let extract_gzip ~file ~dir = ("tar", ["xfz"; file; "-C"; dir])
+  let extract_bzip2 ~file ~dir = ("tar", ["xfj"; file; "-C"; dir])
+  let extract_xz ~file ~dir = ("tar", ["xfJ"; file; "-C"; dir])
+  let extract_lzma ~file ~dir = ("tar", ["xfY"; file; "-C"; dir])
 
-  let create_gzip cmd ~output ~inputs = cmd "tar" ("czhf" :: output :: inputs)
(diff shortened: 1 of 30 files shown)


src/core/opamExternalTools.ml
```diff
@@ -0,0 +1,125 @@
+(**************************************************************************)
+(*                                                                        *)
+(*    Copyright 2018 OCamlPro                                             *)
```
avsm: this isn't copyright ocamlpro :)

src/core/opamExternalTools.ml
```diff
@@ -0,0 +1,125 @@
...
+
+let unpack x = x
+let custom x = x
+
+let has_space (cmd, _) = String.contains cmd ' '
+let cmd_to_string (cmd, _) = cmd
+
+let to_string (cmd, args) =
+  (* TODO: Add backslashs to quotes for each args *)
+  let quote x = if String.contains x ' ' then "'" ^ x ^ "'" else x in
```
avsm: This quoting logic seems like it might be replicated elsewhere in opam. `Bos` also has this

jpdeplaix: Thanks, I'll look around. I tried to be as close as the original code as possible (modulo quoting that is added because that seemed broken anyway)

src/core/opamExternalTools.ml
jpdeplaix: I was just copying the header from the other files. Should I put my name instead ?

dra27: Actually, in this instance I think it depends - is the module predominantly just refactoring?

jpdeplaix: yes, I think so.

src/core/opamExternalTools.ml
AltGr: Note that backslash-quoting in single quotes doesn't work for the shell. You have to write e.g. `'foo'\''bar` ; and tabs should also be supported.

AltGr:
Thanks. I would have liked to discuss this beforehand, but I think it's a good idea.

A few comments:
- please update doc/index.html with the new module
- the name `OpamExternalTools` fits the contents of the module well. However, it feels weird that e.g. `OpamSystem.make_command` now relies on `OpamExternalTools.t` (and `OpamExternalTools.custom`) when its principal use is to call the commands from package scripts (e.g. the contents of the `build:` and `install:` fields, see `OpamAction.make_command`). I would also prefer that `OpamProcess` wouldn't depend on the new module at all, since it doesn't rely on external commands (it'd be nice to avoid the dependency for `OpamStd`, too, but I don't see how to achieve that easily).
- maybe there could be a distinction between tools that opam relies on (`cp`, `mv`, `tar`, `patch`, `git`, etc.), and the ones that are just polled at some point (for example, while I think this is generally a great improvement in readability, in the specific case of `OpamSysPoll`, where commands are just checked once, and are not always expected to exist, the benefit is not clear. It's no longer obvious that `getprop` is only ever useful to detect an Android device, for example)

These are details, but gathering the commands like this is a good thing and changing from a list to a `cmd, args` pair is also a good change (that I am sure you noticed had been done for the low-level command engine, but without reaching most calling points).

One last thing, if you find the combination of `OpamSystem` and `OpamFilename` awkward, I totally agree... this might be the last part I didn't get around to rewriting since 1.0, and the `OpamFilename` types are not particularly convenient, while adding a redundant layer of indirection. So I'd be happy to discuss their replacement if you're so inclined :) Ideally, but this would be a much bigger change, we should have one module with basic definitions (e.g. file names) and system operations, then `OpamExternalTools`, then one based exclusively on our file name types, and defining the operations like `cp` and `mv` just once.

avsm:
Also in the medium term, it would be good to start vendoring in things like ocaml-tar or ocaml-git to reduce the dependency on external tools by default.  That might affect the consideration of names for this module...

PR #3222, opened 2018-02-13
make lib-ext: fix curl call and cache URL

### LINKED ISSUE #3221: Inadequate exit code of curl in Makefile
When running `make lib-ext`, it tries to download tarball caches in `opam.ocaml.org` if the tarball is unavailable from its [original source list](https://github.com/ocaml/opam/blob/master/src_ext/Makefile.sources).  However, the problem is that [the `curl` command](https://github.com/ocaml/opam/blob/1c0dabb3fc7380896d1fe6ad152c9219ae56d97e/src_ext/Makefile#L146) downloading tarballs from `opam.ocaml.org` returns zero exit code even when `opam.ocaml.org` returns 404 not found page.

```bash
$ curl  -LSs -o a4c0e652e56e74c7b388a43f9258d119 https://opam.ocaml.org/2.0~dev/cache/md5/a4/a4c0e652e56e74c7b388a43f9258d119
$ echo $?
0
$ cat a4c0e652e56e74c7b388a43f9258d119
<html>
<head><title>404 Not Found</title></head>
<body bgcolor="white">
<center><h1>404 Not Found</h1></center>
<hr><center>nginx/1.2.1</center>
</body>
</html>
```

1. Adding `-f` option to `curl` may solve the issue.
2. The cache links are broken, e.g., the tarball of cudf `https://opam.ocaml.org/2.0~dev/cache/md5/a4/a4c0e652e56e74c7b388a43f9258d119` is not found. Is this because I am using the main branch?

AltGr:
Closes #3221.

### DIFF
--- a/shell/bootstrap-ocaml.sh
+++ b/shell/bootstrap-ocaml.sh
@@ -3,7 +3,7 @@
 GEN_CONFIG_ONLY=${GEN_CONFIG_ONLY:-0}
 
 if command -v curl > /dev/null; then
-  CURL="curl -OLSs"
+  CURL="curl -OLSfs"
 elif command -v wget > /dev/null; then
   CURL=wget
 else
--- a/src_ext/Makefile
+++ b/src_ext/Makefile
@@ -17,7 +17,7 @@ MD5_flexdll = cc456a89382e60d84130cddd53977486
 
 ifndef FETCH
   ifneq ($(shell command -v curl 2>/dev/null),)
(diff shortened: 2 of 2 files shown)


dra27:
Also line 6 of `shell/bootstrap-ocaml.sh`

AltGr:
Indeed, thanks

AltGr:
Hmm, doesn't `opam-depext` have the issue as well ?

PR #3237, opened 2018-02-28
Eliminate the ocaml-mode jbuild file in the client

dra27:
The good: no OCaml syntax `jbuild` file any more

The bad: the trick to achieve this, but still have the current git SHA is almost as evil as OCaml syntax `jbuild` files

The intermediate: I'll discuss with Dune devs whether we could have a `mode` option in the `rule` stanza which can do this less evily.

### DIFF
--- a/src/client/jbuild
+++ b/src/client/jbuild
@@ -1,41 +1,10 @@
-(* -*- tuareg -*- *)
-
-module J = Jbuild_plugin.V1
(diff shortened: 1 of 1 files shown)


AltGr:
Thanks! I still like this version better, it's more consistent. But yes, if it could be made more straightforward... :)

PR #3239, opened 2018-03-02
Update to dune beta 19.1!

dra27:


dra27:
OK - so there are two workarounds needed in order to support dune beta18, which doesn't feel worth it... beta19 is due to be released this week which should address both, so I'll the PR with beta19 when it's out!

dra27:
The Travis failure here is similarly because the opam packages in the `OPAM_TEST` version need updating.

dra27:
Hopefully this version will now get a full pass - patches not in the correct directories!

Dune beta19 has at least one feature which would be nice, but I have too many open PRs at the moment, and it's getting confusing... let's get this one merged and worry about the next beta later!

dra27:
Right, I had enough of the CI being silly about transient, so I've rewritten it!

AltGr:
Cool, finally getting there, hopefully this won't break so often anymore :)

dra27:
Wohoo!

PR #3241, opened 2018-03-04
Update re to 1.7.2

dra27:
Nice to have another build in src_ext which now natively supports jbuilder! Alas, it temporarily requires a patch (see ocaml/ocaml-re#155)


dra27:
@AltGr - you'll need to blow the Travis build cache in order for the `OPAM_TEST` jobs to work, since they need to update the installed re package.

AltGr:
Still fails after a cache wipe... maybe some travis tweaking is required

dra27:
@AltGr - I don't think you wiped all the caches. I'm fairly sure you need to wipe both the cache for this PR and also the master cache. All my tests with this are working (especially if I manually `rm -rf $OPAMBSROOT` in the script). It also gets rid of the really weird and slightly worrying error installing mccs.

rgrinberg:
Btw the patch here is no longer necessary (as of 1.7.3)

dra27:
I know - I was going to close this one!

PR #3242, opened 2018-03-04
C stubs for native Windows support

dra27:
This PR starts the process of merging some of the larger patches needed for proper native Windows support. My main focus here is to try to get things which will hugely affect both the API and build process in before the release of 2.0.0.

Windows support will necessarily require C stubs. The C code is provided in `src/stubs` but the `jbuild` is only symlinked/copied on Windows, so this library is never built on Unix. opam-core is then altered to `select` between either a dummy implementation of the stubs for Unix or the stubs library for Windows. This may be replaced in the future as and when Dune gets better support for conditional compilation for os-specific things like this.

Prior to Windows 10 1511, colour support is only available via specific Win32 API calls. Tempting as it may be to abandon pre-Windows 10 users, it's a bit mean as Windows 7 and 8 are still supported for quite a few more years. However, in order to translate VT100 sequences into the appropriate calls, it means that anything which may include an escape sequence **must** go via an `OpamConsole` function. This necessitates a few refactorings:
 - There's a little hook added to allow `OpamStd.Config.env` to access `OpamConsole.warning` which means that warning messages from that function display in colour everywhere.
 - I moved `OpamStd.Format.print_table` to `OpamConsole.print_table`. Morally, this function doesn't really have a home in either module - it could be implemented similarly to the trick in `OpamStd.Config.env` with the circular dependency closed at runtime?
 - Tricks with flushing handles for status lines have to go, as clearing the status line requires an API call.

A problem with re-routing all the functions is that `%a` has a different signature for `printf` vs `sprintf` which has affected the public API for some of these changes. It is possible to hook this while maintaining `printf` but it's a **lot** hairier.

UTF-8 support on Windows is quietly improving - OCaml 4.06 added a lot and there's an open PR to push this support back to Windows 8 and earlier. The code here predates that work, but lots of it is still applicable, even with better support in OCaml. A key problem is that the default console fonts on Windows don't include the symbols required (I haven't yet found a good alternative font which actually includes them all - it's possible if we ask Microsoft nicely they might update the Consolas with a few of the missing ones in future...). The code is refactored to allow substitution with different symbols and there's quite a lot of C code for determining what the console font is and what glyphs it contains.

In fixing the compatibility layer so that `Uchar` and `Buffer.add_utf_8_uchar` were available for 4.02.3, I noticed that `module type of` wasn't being "correctly", so that's fixed too.

The following more minor things are also implemented in C:
 - `Unix.getpid` does not return a unique value or the normal notion of "process ID" on Windows (this is known: see [Mantis #4034](https://caml.inria.fr/mantis/view.php?id=4034)). This affects opam's attempts to generate unique log and temporary filenames.
 - `Unix.wait` isn't available on Windows, but opam knows the list of child processes its launched, so it's possible to implement `OpamProcess.wait_one` using the Win32 `WaitForMultipleObjects` call.
 - `opam init` has cause to write to the Windows Registry, so there's a stub for doing that.
 - `opam init` also has cause to offer to update the `HOME` environment variable (Windows users won't have this set by default, but it's a good idea - other software packages pick up on it too). Changing this involves several API calls - one to retrieve the correct value to put in `HOME` and another to broadcast to other Windows processes that the persisted environment has changed.

Finally, there is a huge piece of machinery for the Windows equivalent of `eval $(opam env)` which is one area in which Windows is vastly superior to Unix, since it's never necessary for the user to run it. It is possible on Windows to run `putenv` in the context of your parent (i.e. the shell) using process injection (which is an entirely standard thing to do on Windows - I got the idea from another tool I've used for years doing exactly the same thing). The only non-standard bit is that I've done the injection directly rather than the more conventional use of a DLL, but that's because calling `SetEnvironmentVariable` is not a lot of code, so there's no need for the build system complexity of two DLLs. However, it's necessary to worry about whether you're a 32-bit opam running in a 64-bit shell and vice versa. On Windows, this means that an auxiliary C program is compiled called `opam-putenv.exe` which will always be for the opposite architecture from `opam.exe`. If there's a mismatch between opam and the shell, then it uses the auxiliary (which won't have a mismatch) to do the code injection and update the parent.

It's really neat, and it's been happily working since 2015 when I wrote it!

There's then a refactoring commit for a lovely piece of software called Clink which adds GNU readline support to the Windows Command Processor and also allows prompt customisation and tab completion (both of which I've implemented, but which aren't in this PR). The refactoring is necessary because of how Clink works: `cmd.exe` starts which launches Clink and then clink has the equivalent of `.bash_profile` scripts (the Windows Command Processor doesn't really have these itself) so this is the perfect place to ensure that `opam config env` is run every time you open a new command prompt (that's the only time it is necessary to run `opam config env`, as at the moment I've chosen not to persist opam's environment changes... I may change my mind on that though!). At this point, when opam is setting the environment variables, it needs to run `putenv` not in the parent process, but in the parent of its parent!

This PR is probably best reviewed commit-by-commit rather than as one huge diff.


configure
```diff
@@ -3099,12 +3102,26 @@ EOF
 $as_echo "$OCAML_OS_TYPE" >&6; }
 
 
-if test "${OCAML_OS_TYPE}" = "Win32" ; then
+if  test "${OCAML_OS_TYPE}" = "Win32"; then :
```
avsm: Out of curiosity, is this `:` at the end for portability on Windows?

dra27: That's autoconf, not me :smile: That's coming from `AS_IF` (I try to follow m4sh coding standards, although m4sh, coding and standards I find to be co-oxymoronic terms!). I think it's a cheap way of ensuring that you never get an empty `then` branch, as that would be a syntax error. There's similar things in the `Makefile`s of the OCaml testsuite, IIRC.

AltGr:
Great work! I'll review ASAP. Thanks!

dra27:
I've revised the Console stubs in the last few days (it's all pushed, so as here!). It should now be the case that the Unix and Windows 10 behaviour for the console status line is identical - so it does still use unflushed `\r\027K` instructions to ensure that the status is erased on next output. The `clear_status` call is now just a `flush`. However, to maintain correct console operation for before Windows 10 1511, *all* output after a `status_line` must be via an `OpamConsole` function or the status line won't be erased. Obviously, it's just a display fault if that gets forgotten, but my change means it's now only a display fault on old versions of Windows, not everywhere!

dra27:
I had made a fundamental and stupid mistake with the console function refactoring - `%!` of course makes no sense with `ksprintf`! Three commits added: two of which are patches to the previous work converting erroneous `%!` to actual `flush` calls and the third goes through the entire codebase eliminating unnecessary calls to `%!` (before, these calls were unnecessary because they were passed to functions which flushed anyway; now they're incorrect since `%!` is a no-op).

src/solver/opamActionGraph.ml
```diff
@@ -30,19 +30,27 @@ let name_of_action = function
...
   | `Build _ -> "build"
 
-let symbol_of_action = function
+let symbol_of_action =
+  let open OpamConsole in
+  function
   | `Remove _ ->
-      OpamConsole.utf8_symbol OpamConsole.Symbols.circled_division_slash ""
+      utf8_symbol Symbols.circled_division_slash
+                  ~alternates:[Symbols.greek_small_letter_lambda] "X"
```
AltGr: We already have a fallback to more explicit strings (just above) in case UTF8 is not available. Are the single-letter fallbacks really needed ?
I figure the intent is to allow partial fallback if only some of the symbols are not available ?

src/tools/jbuild
```diff
@@ -17,6 +17,8 @@
    (flags (:standard (:include ../ocaml-flags-standard.sexp) (:include ../ocaml-context-flags.sexp)))
    (libraries (opam-client))))
 
+(include opam-putenv.inc)
```
AltGr: Not really an issue, but we should be aware that we lose on Windows the property that opam can be used from a single installed binary (well, given that all its dependencies are there on course)
In any case, for Windows, we'll need some kind of installer, so that property isn't as valuable that it is on Linux. And it's in any case not a requirement.

src/core/opamStd.ml
```diff
@@ -500,7 +500,7 @@ module OpamString = struct
 
 end
 
-
+let is_windows = Sys.os_type = "Win32"
```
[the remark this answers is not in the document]
dra27: That's true - good opportunity to change it!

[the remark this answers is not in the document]
dra27: (I'd already resolved the conflict!)

src/core/opamConsole.ml
```diff
@@ -322,6 +314,108 @@ let read fmt =
...
+      let str =
+        OpamStd.List.concat_map
+          ("\n" ^ String.make base_indent ' ')
+          (String.concat sep)
+          splits
+      in
+      output_string str;
+  in
+  List.iter print_line table
+
```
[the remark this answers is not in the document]
dra27: I will double-check that's the case, but I think so.

src/solver/opamActionGraph.ml
dra27: Indeed - the alternates list is based on looking at the console fonts in previous versions of Windows and then having the single letter for complete failure. FWIW it was only for completeness - Consolas has something useful on all support versions of Windows.

src/tools/jbuild
dra27: It's not completely lost - it's just that the user would otherwise have to ensure that they are using the correct `opam.exe`. You really don't want to see the gymnastics required to eliminate `opam-putenv.exe` completely :wink:

src/core/opamStubs.mli
```diff
@@ -0,0 +1,130 @@
+(**************************************************************************)
```
AltGr: For this (and other added modules) could you also add a line in `/doc/index.html` ? Thanks

dra27:
Right, fingers crossed that should pass on macOS as well. Since Dune beta18, the `UChar` module in extlib (which is compiled when extlib-compat, rather than extlib, is installed) conflicts with the standard library's `Uchar` module on case-preserving, rather then case-sensitive, file systems (so macOS and Windows). This won't be seen on Windows because `lib-pkg` compiles extlib with `minimal=1` (as the opam `extlib` package does) but was seen on Travis macOS here. I've fixed it by changing the CI to specify extlib explicitly and also altering the opam-core package to conflict with extlib-compat.

AltGr:
Great, thanks!

PR #3248, opened 2018-03-08
Pass fd to functions in OpamFilename.with_flock*

dra27:
File locking behaves differently under Windows - if a file is to be written or read using a lock, then the same fd needs to be used. Alter the OpamFilename.with_flock functions to pass the fd of the lock file
instead of unit.


dra27:
The AppVeyor failure is just Cygwin being silly. The Travis fails are because 89a1bae6ee055b306fde0bb5924d630b5827cd00 in #3242 is not yet merged.

AltGr:
Ok to merge as soon as it passes (i.e. the #3242 dependency)

dra27:
Rebased - expect AppVeyor to pass and Travis to fail the jobs < 4.06.1

PR #3263, opened 2018-03-21
Add sandboxing wrapper based un "bubblewrap"

AltGr:
Much nicer interface than using namespaces directly, and delegates the issues
of namespaces not generally being accessible as user.

### DIFF
--- a/shell/bwrap.sh
+++ b/shell/bwrap.sh
@@ -1,0 +1,53 @@
+#!/bin/bash -ue
+
+# Configure with the following lines in ~/.opam/config:
+#
+#   wrap-build-commands:   ["%{root}%/opam-init/hooks/bwrap.sh" "build"]
+#   wrap-install-commands: ["%{root}%/opam-init/hooks/bwrap.sh" "install"]
+#   wrap-remove-commands:  ["%{root}%/opam-init/hooks/bwrap.sh" "remove"]
+
+if ! command -v bwrap >/dev/null; then
+    echo "The 'bwrap' command was not found. Install 'bubblewrap' on your system, or" >&2
+    echo "disable sandboxing in $OPAMROOT/config at your own risk." >&2
+    echo "See https://github.com/projectatomic/bubblewrap for details." >&2
+    exit 10
+fi
+
+ARGS=(--unshare-net --new-session)
+ARGS=("${ARGS[@]}" --proc /proc --dev /dev)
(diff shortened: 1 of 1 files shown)


shell/bwrap.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash -ue
+
+# Configure with the following lines in ~/.opam/config:
+#
+#   wrap-build-commands:   ["%{root}%/opam-init/hooks/bwrap.sh" "build"]
+#   wrap-install-commands: ["%{root}%/opam-init/hooks/bwrap.sh" "install"]
+#   wrap-remove-commands:  ["%{root}%/opam-init/hooks/bwrap.sh" "remove"]
+
+if ! command -v bwrap >/dev/null; then
+    echo "The 'bwrap' command was not found. Install it, or disable sandboxing in" >&2
```
avsm: Since bubblewrap is a new tool, perhaps add a URL to its source here?  The link between `bwrap` and Bubblewrap wouldn't be clear otherwise

shell/bwrap.sh
```diff
@@ -0,0 +1,52 @@
+#!/bin/bash -ue
```
avsm: Do you definitely need bash here?

AltGr: well, pure `sh` doesn't really support arrays, and properly escaping a list of arguments is tedious without them...

shell/bwrap.sh
AltGr: Good point!

AltGr:
Well that was fast! :)

shell/bwrap.sh
avsm: Makes sense - couldn't find an easy way in /bin/sh indeed

shell/bwrap.sh
```diff
@@ -0,0 +1,53 @@
...
+    case "$1" in
+        ro) B="--ro-bind";;
+        rw) B="--bind";;
+    esac
+    for dir in "$@"; do
+        [ -d "$dir" ] && ARGS=("${ARGS[@]}" "$B" "$dir" "$dir")
+    done
+}
+
+add_mounts ro /usr /bin /lib /lib32 /lib64 /etc /opt /opt /nix/store /home
```
avsm: just noticed `/opt` is repeated here

AltGr: oops, thanks

Issue #3240, opened 2018-03-04
Failing to switch to 4.06.1 with opam 2.0.0~rc


ttamttam:
I'm currently playing with nixos ([http://nixos.org](http://nixos.org)). Of course, the first thing to test is to install a OCaml environment!

Basically, I used nix to install `ocaml` (4.03.0) and `opam` (1.2.2).
I then created switch `4.06.1+flambda`, and installed `opam-devel`.

I then started again from scratch with the newly generated `opam` (2.0.0~rc), but it failed to `opam switch install 4.06.1+flambda`.

From there, I'm not sure about what to check and try?

Here is the whole transcript:
```sh
[matt@pcdell:~]$ ocamlc -version
The program ‘ocamlc’ is currently not installed. It is provided by
several packages. You can install it by typing one of the following:
  nix-env -iA nixos.ocaml
  nix-env -iA nixos.ocaml_3_12_1
  nix-env -iA nixos.ocaml_4_03

[matt@pcdell:~]$ opam --version
The program ‘opam’ is currently not installed. You can install it by typing:
  nix-env -iA nixos.opam

[matt@pcdell:~]$ nix-env -i ocaml opam
installing ‘ocaml-4.03.0’
installing ‘opam-1.2.2’
building path(s) ‘/nix/store/4yac5a5aq48xlj2kh4b29hd4vnjz5gla-user-environment’
created 752 symlinks in user environment

[matt@pcdell:~]$ ocamlc -version
4.03.0

[matt@pcdell:~]$ opam --version
1.2.2

[matt@pcdell:~]$ opam init
Checking for available remotes: rsync and local, git, mercurial, darcs. Perfect!

=-=- Fetching repository information =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
[default] synchronized from https://opam.ocaml.org

=-=- Gathering sources =-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=

=-=- Processing actions -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=
∗  installed base-bigarray.base
∗  installed base-threads.base
∗  installed base-unix.base
Done.

In normal operation, OPAM only alters files within ~/.opam.

During this initialisation, you can allow OPAM to add information to two
other files for best results. You can also make these additions manually
if you wish.

[...]

PR #3267, opened 2018-03-22
Fix undefined 'os' and related variables

AltGr:
ref. #3240 (better fallback than before, but this doesn't fix OS detection on
nixOS)

### DIFF
--- a/src/core/opamStd.ml
+++ b/src/core/opamStd.ml
@@ -554,8 +554,34 @@ end
 
 module OpamSys = struct
 
+  let is_windows = Sys.os_type = "Win32"
+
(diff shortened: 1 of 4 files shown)


src/core/opamStd.ml
```diff
@@ -554,8 +554,34 @@ end
 
 module OpamSys = struct
 
+  let is_windows = Sys.os_type = "Win32"
```
ttamttam: is_windows == Sys.win32

AltGr: Yes. This code was there before OCaml 4.01.0, which introduced it, though.

PR #3268, opened 2018-03-22
Use env PATH in OpamSys.with_process_in (for NixOS)

ttamttam:
#3240


AltGr:
Ah, thanks a lot for the effort! This already got fixed as part of #3267, though (using the already present function for extracting `PATH` on the OS, through some code shuffling).
Best

PR #3273, opened 2018-03-30
This adds automatic support for bubblewrap on Linux

AltGr:
by installing a wrapping script to the opamroot, and setting it as
wrapper automatically by default.
Checks that the `bwrap` utility is installed are also added.

Includes a small refactoring of the repository format upgrade, with
an added config upgrade and automatic repo update pass.


AltGr:
CI expected to fail on Linux until we make sure `bwrap` is properly installed there.

AltGr:
Failing as expected, because Bubblewrap is not available. Travis defaults to Ubuntu "Trusty" (LTS), and it appears that the package is only available from "Artful" (https://packages.ubuntu.com/search?keywords=bubblewrap&searchon=names&exact=1&suite=all&section=all) so we'll need to tweak the CI scripts to get it installed.

AltGr:
Installing bwrap after adding the artful remote is OK, except namespaces are disallowed within the Docker container. Disabled wrapping for our CI for now.

dra27:
I guess it sort of makes that one sandbox doesn't necessarily work inside another?

dra27:
It might work if you have `sudo: true` instead which I think is the correct way to switch from container to VM builds. I think there's a performance hit, though (reduced parallelism and possibly fewer cores?) though it's possible you can apply `sudo: true` to just one job?

AltGr:
`sudo: true` is not a good solution, as it falls back to the former VM backend, which is much slower and has no staging at all: there was quite some work involved for migrating to multi-stage, Docker-based CI and not worth going back.

Although my googling returned nothing, @avsm could point me to docker's `--privileged` option that makes running `bwrap` possible. Not sure if there is a way to pass that to Travis, though ?

src/state/opamFormatUpgrade.ml
```diff
@@ -1008,7 +1008,11 @@ let from_2_0_beta_to_2_0_beta5 root conf =
        (OpamFile.Config.eval_variables conf))
     conf
 
-let latest_version = v2_0_beta5
+let v2_0 = OpamVersion.of_string "2.0"
+
+let from_2_0_beta5_to_2_0 _ conf = conf
+
+let latest_version = v2_0
```
dra27: I know we're *really* close to release, but is that wholly safe? It's OK having `opam-version: "2.0"` not have a final meaning until actually released, but this could result in having the opamroot version having to be greater than the release version for 2.0?!

dra27: (I'm thinking of OCaml, where bumping versions and so forth is always the last commit before release it tagged)

Issue #2925, opened 2017-05-02
opam init -a should recommend another shell syntax


zapashcanon:
Hi,

When running `opam init -a`, the output contains:

``` sh
# To setup the new switch in the current shell, you need to run:
eval `opam config env`
```

It should be:

```sh
# To setup the new switch in the current shell, you need to run:
eval $(opam config env)
```

See [this explanation](https://github.com/koalaman/shellcheck/wiki/SC2006).

dra27:
Both syntaxes are valid POSIX, and you're really not going to hit the nesting issues which can plague backtick command quoting. However, there are users whose default shell doesn't support `$()` (I believe Solaris 10 doesn't, from a quick Google).

On the basis that it's less typing and, even in 2017, we'd have to add code to detect whether `$()` is actually valid, I'd say it's best left as it is.

AltGr:
Ah, I had been wanting to do this for a while (I only ever use `$()` myself, and backticks are more annoying to type on some exotic non-english keyboards) -- and actually did in 42913bc30bd7cb356ad17ccfa5ea95fb54366e73 -- but I didn't know about Solaris 10. Should I revert ?

Note that we already detect the shell, and have a different escape for `fish`, so reverting to backticks on plain `sh` only would be just one line (except in the manpages, but who cares).

dbuenzli:
> and backticks are more annoying to type on some exotic non-english keyboards

And write in markdown...

> Should I revert ?

Since it's valid POSIX I would say no.

AltGr:
> And write (or c&p) in markdown...

They were more annoying to write in the Cmdliner man syntax though ;)

dbuenzli:
> They were more annoying to write in the Cmdliner man syntax though ;)

If you talk about the `$` I would even say that the behaviour used to be close to random. `cmdliner` 1.0.0 fixed that see [here](http://erratique.ch/software/cmdliner/doc/Cmdliner.html#doclang).

AltGr:
Yes yes, I have made sure to use the 1.0.0 syntax. I should probably update the `cmdliner` version requirement, though, we're still advising 0.9.8.

zapashcanon:
@AltGr, I looked at 42913bc, it looks like one character hasn't been removed, I hl you on the commit.

PR #3274, opened 2018-04-02
Fix opam env messages

dra27:
This PR fixes the display of the `eval $(opam env)` messages. In particular:

 - #2925 converted the ``eval `opam config env` `` message to `eval $(opam config env)`. At this point, it was necessary to add a special case for `csh` which doesn't support that syntax.
 - `OpamEnv.eval_string` only ever guessed the shell from `$SHELL` even when `--shell` had been passed to opam.
 - `opam init` always displayed `eval $(opam env)` regardless of the shell, as did `--help` screens.

This PR:
 - Adds an extra case for `` `csh`` which uses backticks.
 - Extends `OpamEnv.eval_string` to have a `shell` parameter. This is quite an invasive change - in particular, it means that `switch`, `install`, `pin`, `remove` and `reinstall` commands now have a `--shell` parameter. I was quite haphazard about where `~shell` got put... if you're happy with the idea of this change, this should be rationalised (i.e. should the argument always be named and where should it go in argument listed).
 - `opam init` now displays the correct command for the given shell.
 - The `--help` screens for the `switch`, `config` and `env` commands now display the command based on `$SHELL` (i.e. `OpamStd.Sys.guess_shell_compat`). It's possible to peek the `--shell` argument here, but even I think that having `SHELL=bash opam config --shell=fish --help` work correctly is more effort than is necessary :wink:

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -236,7 +236,7 @@ let compute_upgrade_t
        ~upgrade:upgrade_atoms
        ())
 
-let upgrade_t ?strict_upgrade ?auto_install ?ask ?(check=false) ~all atoms t =
+let upgrade_t ?strict_upgrade ?auto_install ?ask ?(check=false) ~shell ~all atoms t =
   log "UPGRADE %a"
     (slog @@ function [] -> "<all>" | a -> OpamFormula.string_of_atoms a)
     atoms;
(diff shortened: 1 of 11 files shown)


dra27:
Hmm, the CI golf going on here has revealed that `ignore` is perhaps used a little too liberally in the code base...

AltGr:
Thanks! I am not sure about adding a `--shell` parameter to `switch`, `install`, etc. though, esp. since it's only used to print an advisory message. Maybe it would be best to store the shell info (which should be given at init) in `~/.opam/config`, in case shell detection fails ?

AltGr:
Does not merge anymore, what is the status of this ?

dra27:
This has been merged as the superior #4427!

PR #3275, opened 2018-04-02
Better shell detection

dra27:
At present, opam bases shell detection entirely on `$SHELL`. This is fine for the default shell, but doesn't cope with a shell launched from another shell.

This PR adds various mechanisms based on `/proc` and `ps` which I believe should work for both Linux and the BSDs. If all three mechanisms fail, it falls back to $SHELL. Note that in this context, "fail" means "doesn't identify a shell" (so, for example, if for some strange reason you start `fish` using a symlink called `fishy`, then opam will continue to inspect `$SHELL`).

### DIFF
--- a/src/core/opamStd.ml
+++ b/src/core/opamStd.ml
@@ -750,20 +750,25 @@ module OpamSys = struct
         with e ->
           fatal e;
           try
-            let c = open_in_bin ("/proc/" ^ string_of_int ppid ^ "/cmdline") in
-            begin try
-              let s = input_line c in
-              close_in c;
-              Some (String.sub s 0 (String.index s '\000'))
-            with
-            | Not_found ->
-                None
-            | e ->
-                close_in c;
+            with_process_in "ps" (Printf.sprintf "-p %d -o comm=" ppid)
+              (fun ic -> Some (input_line ic))
+          with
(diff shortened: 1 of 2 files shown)


dra27:
This PR also fixes the following fairly easy-to-hit case:

foo.sh:
```
#!/bin/bash

eval $(opam env)

# stuff
```
and then run:
```
$ SHELL=fish fish
> ./foo.sh
```
Before, bash would get fish instructions and explode...

src/core/opamStd.ml
```diff
@@ -711,8 +717,45 @@ module OpamSys = struct
...
-    with Not_found -> `sh
+    let parent_guess =
+      if Sys.win32 then
+        None
+      else
+        let ppid = Unix.getppid () in
+        let dir = Filename.concat "/proc" (string_of_int ppid) in
+        try
+          Some (Unix.readlink (Filename.concat dir "exe"))
+        with _ ->
```
AltGr: `e -> Exn.fatal e;` ?

AltGr:
Seems clear enough, thanks! Please just see the detail about never catching `Sys.Break` unlessyou re-raise.

src/core/opamStd.ml
dra27: Done - in all places!

dra27:
And a CI pass!

PR #3276, opened 2018-04-03
Add message to bug-report links, in case of build failure (close #1929 )

### LINKED ISSUE #1929: Add link to relevant bug-trackers on error
Now that the `bug-reports` field is used a bit more pervasively, it would be nice to use that information and display the relevant bug-tracker uris in case of errors.

rjbou:
From issue https://github.com/ocaml/opam/issues/1929

### DIFF
--- a/src/client/opamSolution.ml
+++ b/src/client/opamSolution.ml
@@ -170,7 +170,7 @@ let sanitize_atom_list ?(permissive=false) t atoms =
   atoms
 
 (* Pretty-print errors *)
-let display_error (n, error) =
+let display_error t (n, error) =
   let f action nv =
     let disp =
       OpamConsole.header_error "while %s %s" action (OpamPackage.to_string nv) in
@@ -189,7 +189,12 @@ let display_error (n, error) =
   | `Install nv        -> f "installing" nv
   | `Reinstall nv      -> f "recompiling" nv
   | `Remove nv         -> f "removing" nv
-  | `Build nv          -> f "compiling" nv
+  | `Build nv          ->
+    (f "compiling" nv;
+    let bugreports = OpamFile.OPAM.bug_reports (OpamSwitchState.opam t nv) in
+    OpamConsole.errmsg "You can check if there is an open issue for %s at:\n%s\n"
+      (OpamConsole.colorise `bold (OpamPackage.name_to_string nv))
+      (OpamStd.Format.itemize (fun x -> x) bugreports))
   | `Fetch nv          -> f "fetching sources for" nv
 
 module Json = struct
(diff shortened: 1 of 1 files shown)


rjbou:
@dbuenzli As it is only if an error is raised when compiling, it is also a software concern. The message do not invite to create an issue, but see if there is an open one.

dbuenzli:
> but see if there is an open one.

Generally this will be on the wrong tracker and failing to see one people will open one (again on the wrong tracker).

dra27:
If a package is pinned to a dev repo then, yes, it may well be a software issue - but a package failing to build is almost always a packaging issue. If it doesn't build for whatever reason, the remote should have disabled the package for that environment and any upstream info on that should be linked from there.

dra27:
(Given the relatively high number of bug reports which come to opam's tracker, I completely agree that it would be a good idea to advertise a tracker)

dra27:
Having access to the code again, a ~couple~ ~quartet~ quintet of thoughts:

 - It's obviously a good moment to duplicate `bug-reports` from `OpamFile.OPAMSyntax` to `OpamFile.RepoSyntax`.
 - opam-repository can obviously be updated at once, but it might be an idea (and easy) to add an heuristic for working out bug trackers based on the `upstream` field - i.e. recognising gitlab, bitbucket, github URLs and guessing the issue tracker.
 - If `OpamSwitchState.is_pinned` is false or `OpamSwitchState.is_version_pinned` is true, then the issue tracker should be the remote. If the package is pinned to a repo or a path, then the package's issue tracker is probably appropriate although perhaps a more verbose message displaying both trackers (and a reminder that the package is effectively pinned for development) would be better.
 - There isn't a direct way in `OpamSwitchState` to get from the package back to its remote, but this could easily be added by altering the `OpamRepositoryState.build_index` to augment the packages with a pointer back to their repository or, slightly more hackily, it could be reconstructed from `metadata_dir`. The index update is slightly better, as a repo or path-pinned package would have a different `metadata_dir` and it'd be easier to display both the package bug-reports URL **and** the remote URL if the package also includes the remote where it *originally* came from.
 - The `bug-reports` field would obviously be committed in the repo, so there's a question of what to display when working from a clone. The message could perhaps usefully include a reminder for file-based remotes that the repo is local, so it could be a local problem! I expect that trying to infer that a repo clone of a remote is not the upstream one (i.e. that https://github.com/dra27/opam-repository.git is not `upstream`) is a step too far...

XVilka:
Isn't this a good addition for the [2.1.0 milestone](https://github.com/ocaml/opam/milestone/23)? Shouldn't introduce new bugs but improves UX a bit.

PR #3279, opened 2018-04-05
Remove French punctuation spacing from messages

dra27:
Also, at the risk of kicking a hornets' nest: a few minor word changes here and there.

cc @yallop


AltGr:
Thanks!
Good to merge, but let's wait for more reviews so all related changes can be gathered into the same PR

PR #3280, opened 2018-04-05
Try to improve opam init message about configuration files

dra27:
The list of configuration files being consulted (the first "NOTE" from `opam init`) displays the files in ascending order of priority which is not wholly intuitive. I've tweaked the wording so that the note now appears as:
```
[NOTE] Will configure from /home/dra/.opamrc, then /etc/opamrc, and finally from built-in defaults.
```
or
```
[NOTE] Will configure from /home/dra/.opamrc and then from built-in defaults.
```

I've also added, but then not used, a `?last_sep` parameter to `OpamStd.List.concat_map` which allows you to have lists written `x and y` or `x, y and z` (or even `x, y, and z` if you are so inclined). There are definitely other places in the client where the messages could be augmented by use of that parameter.

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -226,9 +226,15 @@ let init =
     in
     let init_config =
       try
-        OpamConsole.note "Will configure from built-in defaults%s."
-          (OpamStd.List.concat_map ~nil:"" ~left:", " ~last_sep:" and " ", "
-             OpamFile.to_string config_files);
+        let others =
+          match config_files with
+          | [] -> ""
+          | [file] -> OpamFile.to_string file ^ " and then from "
+          | _ ->
+              (OpamStd.List.concat_map ~nil:"" ~right:", and finally from " ", then "
+               OpamFile.to_string (List.rev config_files))
+        in
+        OpamConsole.note "Will configure from %sbuilt-in defaults." others;
         List.fold_left (fun acc f ->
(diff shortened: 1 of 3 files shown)


AltGr:
Nice, thanks!
About `?last_sep`, I believe there is a function `pretty_list` or similar in OpamStd that does that. Maybe the opportunity could be taken to factor uses within the concat_map function ?

src/client/opamCommands.ml
```diff
@@ -226,9 +226,15 @@ let init =
...
       try
-        OpamConsole.note "Will configure from built-in defaults%s."
-          (OpamStd.List.concat_map ~nil:"" ~left:", " ", "
-             OpamFile.to_string config_files);
+        let others =
+          match config_files with
+          | [] -> ""
+          | [file] -> OpamFile.to_string file ^ " and then from "
+          | _ ->
+              (OpamStd.List.concat_map ~nil:"" ~right:", and finally from " ", then "
```
AltGr: Didn't you mean to use `~last_sep` here ?

AltGr: @dra27 ping ?

AltGr: Ah, no, nevermind ^^

AltGr:
Thanks!

src/client/opamCommands.ml
dra27: Sorry :)

PR #3285, opened 2018-04-10
`opamrc update`: recommended & required tools, init-scripts

rjbou:
This PR allow to set up recommended and required tools, and init scripts in the opamrc.
It add also the option `--show-default-opamrc` to `opam init`.
The user can use this option to see the built-in configuration, and uses it as template for an opamrc file.
Note that opamrc fields override the built-in ones.


src/state/opamEnv.mli
```diff
@@ -87,7 +87,7 @@ val update_user_setup:
 
 (** Write the generic scripts in ~/.opam/opam-init needed to import state for
     various shells *)
```
AltGr: explain the new argument ?

src/client/opamInitDefaults.ml
```diff
@@ -32,20 +32,73 @@ let eval_variables = [
   "OCaml version present on your system independently of opam, if any";
 ]
```
AltGr: There is quite a bit of evaluation below: it'd be best to make `init_config` (and its dependencies) a function of `unit`

src/client/opamClient.ml
```diff
@@ -690,8 +680,35 @@ let init
...
+          if bypass_checks then false else
+          let all_ok = init_checks config in
+          if not all_ok &&
+             not (OpamConsole.confirm "Continue initialisation anyway ?")
+          then OpamStd.Sys.exit_because `Configuration_error
+          else not all_ok
+        in
+        let custom_scripts =
+          let gst_env =
+            OpamPackageVar.resolve_global (OpamGlobalState.load `Lock_none)
```
AltGr: same remark as above (use `OpamSysPoll.variables`)

src/client/opamClient.ml
```diff
@@ -690,8 +684,33 @@ let init
...
+        let dontswitch =
+          if bypass_checks then false else
+          let all_ok = init_checks config in
+          if not all_ok &&
+             not (OpamConsole.confirm "Continue initialisation anyway ?")
+          then OpamStd.Sys.exit_because `Configuration_error
+          else not all_ok
+        in
+        let custom_scripts =
+          let env =
```
AltGr: same remark as above

src/client/opamCommands.ml
```diff
@@ -144,7 +144,7 @@ let init =
     `P "Any field from the built-in initial configuration can be overriden \
         through $(i,~/.opamrc), $(i,/etc/opamrc), or a file supplied with \
         $(i,--config). The default configuration for this version of opam is:";
-    `Pre (OpamFile.InitConfig.write_to_string (OpamInitDefaults.init_config));
+    `Pre (OpamFile.InitConfig.write_to_string (OpamInitDefaults.init_config()));
```
AltGr: Oops, forgot this wqs included verbatim here! Since now it can get quite big, and we have `--show-opamrc`, better remove it! :laughing:

dra27:
Sorry to put a spanner into the works - these fields with lists are a problem for the "cascade" of files. Recommending that a user always uses the built-in opamrc as a template just to disable bwrap will cause them to override a system-installed `/etc/opamrc` completely.

I really don't think we can release this without some way of filtering what's already there. For example, I it should be possible to remove `bwrap` from `required-tools`, not just override the entire `required-tools` list. Two possibilities:

 - some kind of operators for the lists (e.g. `required-tools -= "bwrap"` or, in a more syntax-compatible way, `required-tools-filter-out: [ "bwrap" ]` or `required-tools-filter: [ "-bwrap" "+mysandboxtool" ]`)
 - a set language, a bit like Dune uses (e.g. `required-tools: [ "current \ bwrap"`)

Some of this could be fixed by bumping the `opam-version` of opamrc in future, but it seems better to sort this now, rather than have weird upgrade paths in the future for opamrc files.

AltGr:
I don't really agree this is needed as much for this case, and we need to bother: you _can_ indeed cascade files (default, /etc/opamrc, ~/.opamrc), but in practice I see little use in really stacking definitions, esp. since this is just for `opam init` (so unless you use multiple roots, `~/.opamrc` is just one-shot).

For this case, `/etc/opamrc`, being system-wide should be consistent with the presence or absence of `/usr/bin/bwrap`, so I don't see an issue either ^^.

rjbou:
> Recommending that a user always uses the built-in opamrc as a template just to disable bwrap will cause them to override a system-installed /etc/opamrc completely.

The purpose of `--show-default-opamrc` is to show only the built-in configuration, to build his own. If there is no opamrc file installed, it can be used as an opamrc to edit. But if there is one, the user can update his own, based on the default configuration.

avsm:
@altgr wrote:
> but in practice I see little use in really stacking definitions

Hang on, I see quite a lot of use for distros in stacking definitions, especially to control the sandboxing by default.  I imagine the *recommended* Debian or Fedora packages will involve installing an `/etc/opamrc` with the package with a pointer to the appropriate distro-specific sandbox setup (which might differ from ours)

PR #3288, opened 2018-04-12
travis: restrain cudf to 0.9

rjbou:
In travis build, opam automatically install (dependence) cudf.0.8, which broke the build because of camlp4 system package (success even if campl4 is not installed).


AltGr:
This is another occurence of the "cubic version-lag numerical installability" (#3239) (obfuscated by a bug on the camlp4.xxx-system package) that sometimes leads to unoptimal solutions on opam 2.0.0~rc1 (which we now use for bootstrapping in travis, unfortunately). Here it's worked around by fixing the dep version.

PR #3290, opened 2018-04-12
Big optimisation pass on universe loading

AltGr:
by reviewing the whole loading process, this achieves x2 speedup on a basic `opam install` run.
Involved changes:
- optimise and make tail-rec some functions on the manipulation of formulas (e.g. add cases for CNF conversion when trivial, which is thankfully most of the time)
- review opam to cudf conversion to be based on computing maps from the universe and running `Map.merge` (rather than converting packages one by one)
- simplify handling of optional dependencies when activated, since these are only used to build the dependency graph
- factorise as much as possible the loading between runs with different flags (build, post, opt), including CNF conversion

This also includes a fix in computing the dependency graph, which was
over-approximating dependencies in some cases with conjunctions of
version constraints (e.g. `a>=3 & a<4`.

### DIFF
--- a/src/client/opamAdminCheck.ml
+++ b/src/client/opamAdminCheck.ml
@@ -67,7 +67,7 @@ let installability_check univ =
   let graph =
     OpamCudf.Graph.of_universe @@
     OpamSolver.load_cudf_universe
-      ~depopts:false ~build:true ~post:true univ packages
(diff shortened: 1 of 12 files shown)


avsm:
wow, nice speedup!

Issue #2999, opened 2017-07-15
Warn if `opam update` was not invoked for sometime.


dbuenzli:
I noticed that nowadays when I `brew install` packages automatically update which may be a right thing to do with a less often used system package manager.

The `brew` behaviour may be a bit extreme for a developer oriented package manager but the same problem does happen for more casual opam users. The OCaml opam repository sometimes gets issues about people who simply didn't `opam update` for quite sometime (it also happens to me btw.).

I think `opam` could suggest to `opam update` when an install fails or set of constraints cannot be satisfied and the last `opam update` is greater than a couple of days.

Issue #2951, opened 2017-05-31
Usability issue with `opam update <package>`


gasche:
On the beginners list:

>  I also just tried `opam update utop` and got the following answer:
> > Cannot update the package utop because it is not pinned
>
> What does "pinned" mean here ?

The error message should be changed to take into account the common case where one considers `update` and `upgrade` as synonyms. I would suggest the following:

> To install the latest version of the package, you should use `opam upgrade utop` instead. The `update` command is used to update package *repositories*, or in the special case where a package is *pinned* to a development repository on your machine.

hcarty:
I'm fairly comfortable with opam and I still make this mistake sometimes. I like @gasche's suggestion for improved wording.

dra27:
opam could even offer to do opam upgrade in that context...

gasche:
I think that, besides what we do when users hit the current usability bug, it would be interesting to think about how to *fix* the usability bug -- the current proposals are recovery strategies rather than fixes.

(I started writing the paragraph below with "I know this breaks compatibility so I'm not suggesting this change right now, but it's informative to lay it out still.". I'm not convinced anymore, maybe we can actually do the change right now?)

Would anything bad happen if we made `upgrade` and `update` actual synonyms? If I understand correctly, currently `update` (1) syncs repository metadata (2) syncs pin metadata and (3) runs `upgrade` (but the tool asks for confirmation so this last part can be denied). This means that the extra expressivity coming from having a separate `upgrade` is only there to upgrade stuff without synchronizing metadata first. I don't see right now why that is useful, but I think it could be relegated to an option flag on the common `update/upgrade` command: `--no-metatadata-update`?

dra27:
It's an interesting idea - are there are other package managers which do this (off the top of my head, the `update` then `upgrade` model is very familiar - though I agree that it's trivial easy in all of them to forget which one you mean)?

There is one bonus to having the separate steps - assuming packages are cached, `opam upgrade` doesn't have to involve network access, where `opam update` always does.

I wonder if perhaps keeping the two commands, but deprecating `update` in favour of `sync`?

dbuenzli:
> The error message should be changed to take into account the common case where one considers update and upgrade as synonyms. I would suggest the following:

I think there should be absolutely no error message here. It seems to me that the right semantics would be to simply update the OPAM repo in which the package belongs.

dbuenzli:
(Note if you want to lump update and upgrade there's `opam update -u`)

[...]

PR #3294, opened 2018-04-13
Messages at `update` and `install`

rjbou:
This PR is about warning and error message:

- at `update`, if nondev packages are given as argument, an explicit message is displayed, and returns with the good return code (Sync Error). It is related to #2951.
- at `install`, if an `update` has not been done since a while (3 weeks), a note invites the user to update repositories. It is related to #2999.

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -518,7 +518,7 @@ let update
           (OpamPackage.Set.to_json updates);
       (success, not (OpamPackage.Set.is_empty updates)), st
   in
-  repo_update_failure = [] && dev_update_success,
+  repo_update_failure = [] && dev_update_success && remaining = [],
   repo_changed || dev_changed,
   rt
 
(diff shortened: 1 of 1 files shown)


src/client/opamClient.ml
```diff
@@ -442,9 +454,16 @@ let update
...
         "The following are not development packages (no dynamic or version \
-         controlled upstream) and can't be updated individually. What you \
-         want is probably to update your repositories: %s"
-        (OpamPackage.Set.to_string nondev_packages);
+         controlled upstream) and can't be updated individually: %s\nYou may \
+         want to update your repositories with just %s or to upgrade your \
+         package%s with %s %s"
+        (OpamPackage.Set.to_string nondev_packages)
+        (OpamConsole.colorise `bold "opam update")
+        (if OpamPackage.Set.cardinal nondev_packages = 1 then "" else "s")
```
AltGr: I recently added `is_singleton` to the module, but it doesn't matter here :)

AltGr:
Thanks!

Issue #3292, opened 2018-04-13
opam2-rc1: unable to install switches from non-default repositories?


gasche:
In order to try the latest ocaml-multicore releases, I did

```bash
$ opam repository add multicore git+https://github.com/ocamllabs/multicore-opam.git
```

which seems to make the packages available as expected:

```bash
$ opam show ocaml-variants -f available-versions | grep -o "\\S*+multicore"
4.02.2+multicore
4.04.2+multicore
4.06.1+multicore
```

however, `opam switch create` apparently cannot find these switches:

```bash
$ opam switch list-available | grep multicore
# no output

$ opam switch create multicore ocaml-variants.4.06.1+multicore
[ERROR] No compiler matching 'ocaml-variants.4.06.1+multicore' found,
        use 'opam switch list-available' to see what is available,
        or use '--packages' to select packages explicitely.

$ opam install ocaml-variants.4.06.1+multicore
Your request can't be satisfied:
  - no matching version

No solution found, exiting
```

Is there something transient that is wrong with my opam2-rc1 setup? Can other people reproduce the problem with this switch description?

AltGr:
This confusion, due to the new semantics of `opam repo add`, should hopefully be avoided by this recently added message:
```
% opam repository add multicore git+https://github.com/ocamllabs/multicore-opam.git
[multicore] Initialised
[NOTE] Repository at git+https://github.com/ocamllabs/multicore-opam.git doesn't define its version, assuming it's 1.2.

=-=- Upgrading repositories from older opam format =-=-=-=-=-=-=-=-=-=-=-=-=-=-=
Upgrading repository "multicore"...
[NOTE] Repository multicore has been added to the selections of switch 4.05.0 only.
       Run `opam repository add multicore --all-switches|--set-default' to use it in all existing switches, or in newly created
       switches, respectively
```

Can you confirm ?

AltGr:
Note that what you really want, in this case, is the single command:
```
opam switch create multicore ocaml-variants.4.06.1+multicore --repos multicore=git+https://github.com/ocamllabs/multicore-opam.git,default
```

that will configure the new repository only in the switch where you want it :)

gasche:
Thanks! When doing `opam repo`, the following warning should have told me that something was going on, but I didn't realize:

```
[NOTE] These are the repositories in use by the current switch. Use '--all' to see all configured repositories.
```

[...]

PR #3295, opened 2018-04-13
add an explicit warning about the single-switch behavior of `repo add` (#3292)

gasche:
The output looks roughly like this (plus emphasis).

```
   add NAME [ADDRESS] [QUORUM] [FINGERPRINTS]
       Adds under NAME the repository at address ADDRESS to the list of
       configured repositories, if not already registered, and sets this
       repository for use in the current switch (or the specified scope).
       ADDRESS is required if the repository name is not already
       registered, and is otherwise an error if different from the
       registered address. The quorum is a positive integer that
       determines the validation threshold for signed repositories, with
       fingerprints the trust anchors for said validation.

       Note: By default, the repository is only be added to the current
       switch. To add a switch to other repositories, you need to use the
       --all or --set-default options (see below). If you want to enable a
       repository only to install of of its switches, you may be looking
       for opam switch create --repositories=REPOS.
```

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -1477,6 +1477,14 @@ let repository =
      address. The quorum is a positive integer that determines the validation \
      threshold for signed repositories, with fingerprints the trust anchors \
      for said validation.";
+    " ", `add, [],
+    (* using an unbreakable space here will indent the text paragraph at the level
+       of the previous labelled paragraph, which is what we want for our note. *)
+    "$(b,Note:) By default, the repository is only be added to the current switch. \
+     To add a switch to other repositories, you need to use the $(b,--all) or \
+     $(b,--set-default) options (see below). If you want to enable a repository \
+     only to install of of its switches, you may be looking for \
+     $(b,opam switch create --repositories=REPOS).";
     "remove", `remove, ["NAME..."],
     "Unselects the given repositories so that they will not be used to get \
      package definitions anymore. With $(b,--all), makes opam forget about \


gasche:
cc @dbuenzli I looked for a nice way to have multi-paragraph labelled lists (the code here does not build an explicit manpage document, but basically it is turned into a list of `I (cmd ^ args, text)` by `make_subdoc`, and I couldn't find one.

I resorted to adding a new `I ()` block immediately after (which is what this patch does), but if the label is empty this is equivalent to a non-indented paragraph, so I had to use an invisible-but-non-empty label by using an unbreakable space as the label. It feels kind of dirty. I haven't found a documented way to do this in `cmdliner.mli`. I am not sure what is the use of *not* indenting paragraphs with an empty label (you could just use `P` for that), but applications already rely on that so I guess it is too late to change the behavior of this corner case anyway.

dbuenzli:
> I am not sure what is the use of not indenting paragraphs with an empty label (you could just use P for that), but applications already rely on that so I guess it is too late to change the behavior of this corner case anyway.

I'm not sure either. These answers are certainly to be found in the underlying groff language (use `--help=groff` to see the source). More precisely [here's](https://github.com/dbuenzli/cmdliner/blob/452760b9be8c23e1835e0912efb79592061b305d/src/cmdliner_manpage.ml#L409-L410) the `I case generation.

gasche:
Oh, I thought you were somehow doing the alignment work yourself, sorry. The instances of the [groff manual](https://manpages.debian.org/jessie/groff/groff_man.7.en.html) I could find online do not specify any particular semantics to `.TP` without a label.

For what its worth, I looked at the manpages of my system, `man man` displays a multi-paragraph description of a label, and it uses `.IP` for the second paragraph:

```groff
.TP
.BR \-K ", " \-\-global\-apropos
Search for text in all manual pages.
This is a brute-force search, and is likely to take some time; if you can,
you should specify a section to reduce the number of pages that need to be
searched.
Search terms may be simple strings (the default), or regular expressions if
the
.B \-\-regex
option is used.
.IP
Note that this searches the
.I sources
of the manual pages, not the rendered text, and so may include false
positives due to things like comments in source files.
Searching the rendered text would be much slower.
```

```
       -K, --global-apropos
              Search  for  text in all manual pages.  This is a brute-force search, and is likely to take some time; if you can, you should specify a sec‐
              tion to reduce the number of pages that need to be searched.  Search terms may be simple strings (the default), or  regular  expressions  if
              the --regex option is used.

              Note  that  this searches the sources of the manual pages, not the rendered text, and so may include false positives due to things like com‐
              ments in source files.  Searching the rendered text would be much slower.
```

AltGr:
Thanks a lot! That's indeed better :)

gasche:
@AltGr sorry, I just spotted a minor typo in my patch: "is only be added" should be "is only added". Can I let you fix this?

AltGr:
sure, thanks!

Issue #1447, opened 2014-06-12
default bash environment setup only configures interactive shells


dsheets:
Right now, `opam init -a --bash` (well, not exactly, `--bash` doesn't exist) writes to `~/.bashrc` if `~/.bash_profile` doesn't exist. If the shell is unknown, `~/.profile` is used.

The fundamental difference between `~/.bashrc` and `~/.bash_profile`/`~/.profile` is that the latter are run for all *login* shells whereas the former is run for *interactive* shells. The profile files usually execute `~/.bashrc` but `~/.bashrc` contains

```sh
# If not running interactively, don't do anything
[ -z "$PS1" ] && return
```

usually near the top. Because `opam` appends its session configuration to the end of the interactive shell initialization, only interactive sessions are configured for opam.

This is painful when trying to use opam binaries via ssh like

```sh
ssh host "bash -l -c \"ocamlfind list\""
```

If `~/.bash_profile` does not exist, should opam just create it? If `~/.profile` exists, should opam just append its configuration to that?

Configuring for only interactive shells by default seems a little broken.

dsheets:
Is `ssh host "bash -l -c \"ocamlfind list\""` working now? Or impossible?

AltGr:
It *should* :)

Issue #3143, opened 2017-12-29
local switches: shortcut for eval in older project


avsm:
Right now, I shell alias this for whenever I switch back to building a project with a local switch:

```
eval $(opam env --switch=`pwd`)
```

It would be nice to have a shortcut for this, or some other recommendation for how to use it more seamlessly.  `opam exec bash` works fine as well, but obviously requires a subshell.

AltGr:
Just `eval $(opam env)` is a shortcut for it, since opam will choose the local switch by default. **Unless**, of course, `OPAMSWITCH` is already set, because that overrides the automatic local switch selection.

Also, `--switch .` works as well :). Or even `--sw .` if you want it as short as possible!

avsm:
Oh! I was going by the guidelines that opam gives you when creating a new local switch.

I suspect what's going on is that I have multiple local switches, and I am changing directory between them. At this point, `OPAMSWITCH` is set from the first one, and so creating the second switch gives me a big command line.

Perhaps `$(opam env)` could be a special case to ignore OPAMSWITCH?  It's a case where we want to explicitly ignore the old switch setting in order to change the current local switch to the current directory.

AltGr:
I think the latter could be even more confusing in other cases. But I guess the source of your confusion here boils down to this:

```
[~/ocamlpro/ocaml-top]% opam env
OPAM_SWITCH_PREFIX='/home/lg/ocamlpro/ocaml-top/_opam'; export OPAM_SWITCH_PREFIX;
CAML_LD_LIBRARY_PATH='/home/lg/ocamlpro/ocaml-top/_opam/lib/stublibs:'; export CAML_LD_LIBRARY_PATH;
OCAML_TOPLEVEL_PATH='/home/lg/ocamlpro/ocaml-top/_opam/lib/toplevel'; export OCAML_TOPLEVEL_PATH;
MANPATH=':/home/lg/ocamlpro/ocaml-top/_opam/man'; export MANPATH;
PATH='/home/lg/ocamlpro/ocaml-top/_opam/bin:/home/lg/local/bin:/usr/local/bin:/usr/bin:/bin:/usr/games'; export PATH;
```
Here the current switch is inferred from CWD, hence opam __doesn't__ set the `OPAMSWITCH` variable (it's only set when needed, i.e. the switch needs to be explicit for opam to find it again). But:
```
[~/ocamlpro/ocaml-top]% opam env --sw .
[...]

PR #3304, opened 2018-04-20
Shell config update

rjbou:
This PR contains an update of shell configuration scripts to have always opam variables set (evaluation at each command). It was inspired from [direnv](https://github.com/direnv/direnv) (thanks @samoht). Available for all opam handled shells.
For `bash`, I kept sourcing scripts on profile config files (cf. #1447), assuming that profile files are sourced in `.bashrc`. Added a message for the user in this case.

### DIFF
--- a/src/state/opamEnv.ml
+++ b/src/state/opamEnv.ml
@@ -650,7 +650,11 @@ let display_setup root ~dot_profile shell =
 let check_and_print_env_warning st =
   if (OpamFile.Config.switch st.switch_global.config = Some st.switch ||
       OpamStateConfig.(!r.switch_from <> `Command_line)) &&
-     not (is_up_to_date st) then
+     (not (is_up_to_date st) &&
+      (dot_profile_needs_update OpamStateConfig.(!r.root_dir)
+         (OpamFilename.of_string @@
+          OpamStd.Sys.(guess_dot_profile @@ guess_shell_compat ()))) = `yes)
+  then
     OpamConsole.formatted_msg
       "# Run %s to update the current shell environment\n"
       (OpamConsole.colorise `bold (eval_string st.switch_global
--- a/src/state/opamEnv.ml
+++ b/src/state/opamEnv.ml
@@ -371,6 +371,35 @@ let init_sh        = "init.sh"
 let init_zsh       = "init.zsh"
 let init_csh       = "init.csh"
 let init_fish      = "init.fish"
+let opam_env_sh    =
+  "\
+\  _opam_env_hook() {\n\
+\    local previous_exit_status=$?;\n\
+\    eval \"$(opam env --shell=bash)\";\n\
+\    return $previous_exit_status;\n\
+\  };\n\
+\  if ! [[ \"$PROMPT_COMMAND\" =~ _opam_env_hook ]]; then\n\
+\    PROMPT_COMMAND=\"_opam_env_hook;$PROMPT_COMMAND\";\n\
+\  fi\n\
+"
+let opam_env_zsh   =
+  "\
+\  _opam_env_hook() {\n\
+\    eval \"$(opam env --shell=zsh)\";\n\
+\  }\n\
+\  typeset -ag precmd_functions;\n\
+\  if [[ -z ${precmd_functions[(r)_opam_env_hook]} ]]; then\n\
+\    precmd_functions+=_opam_env_hook;\n\
+\  fi\n\
+"
+let opam_env_fish  =
+  "\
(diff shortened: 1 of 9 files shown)


samoht:
I am not sure to understand when do these scripts are actually called. Is this for internal opam commands? Or is it for every user commands? In the later case maybe we should have a way to disable them?

rjbou:
These scripts are generated at `opam init`. They can be found in folder `~/.opam/opam-init/`. User can add (or allow `opam` to add) sourcing `init.xx` script in its shell init file. 
For example, in a bash environment (`opam init` extract):
```
=-=- Required setup - please read -=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=-=

  In normal operation, opam only alters files within ~/.opam.

  However, to best integrate with your system, some environment variables
  should be set. If you allow it to, this initialisation step will update
  your bash configuration by adding the following line to ~/.profile:

    test -r /tmp/opamroot/opam-init/init.sh && . /tmp/opamroot/opam-init/init.sh > /dev/null 2> /dev/null || true

  Otherwise, every time you want to access your opam installation, you will
  need to run:

    eval $(opam env)

  You can always re-run this setup with 'opam init' later.

Do you want opam to modify ~/.profile ? [N/y/f]
(default is 'no', use 'f' to choose a different file)
```

This script, if it's a terminal, set evaluation of `opam env` for each command (not only `opam `ones). Otherwise, it sources the `variable.xx` file, containing last opam environment variables.

On current setup, only `variable.xx` scripts where sourced, at shell loading. For each change (switch for example), user had to run `eval $(opam env)`. In this new setup, it's done in background, new environment variables are always at well set.

AltGr:
I am wondering wether it would be a good idea to re-enable making `--set-switch` the default when necessary in this case. I don't like such adhoc cases in general, but the following interaction should really be avoided:

```
% cd foo
(opam: selected switch ~/foo, environment updated)
% eval $(opam env --switch 4.06.0)
(opam: selected switch ~/foo, environment updated)
```
See #3143 for context. In this case, there can be no point in the command if not using `--set-switch`.

One small glitch is that, if you issue `opam switch bar` in a shell, other terminals without a local switch or OPAMSWITCH set will only sync up on the next prompt. I guess there is not much we can do about that...

And no, this shouldn't spawn any subshells, it's just an `eval` ?

rjbou:
> Moreover, does it mean that we are creating a new shell on every command?

> And no, this shouldn't spawn any subshells, it's just an `eval` ?

Yes, just a `eval $(opam env)` is done for each prompt (cf. [bash example](https://github.com/ocaml/opam/pull/3304/files#diff-234457b80298a00240e5fcd7591e5432R376)).

rjbou:
It is now disabled by default, at `opam init` a message allows user to enable it. To enable/disable, it is done by re-running `opam init`.

src/state/opamEnv.ml
```diff
@@ -371,37 +373,40 @@ let init_sh        = "init.sh"
 let init_zsh       = "init.zsh"
 let init_csh       = "init.csh"
 let init_fish      = "init.fish"
+let opam_env_sh    =
+  "_opam_env_hook() {\n\
```
AltGr: the `\` should be below the opening `"` for correct alignment. `ocp-indent` should do it that way automatically... ?

src/state/opamEnv.ml
```diff
@@ -435,10 +440,52 @@ let string_of_update st shell updates =
...
+  if eval_env then
+    let thenn =
+      match complete with
+      | None -> opam_env
+      | Some cplt -> Printf.sprintf "%s\n  %s" opam_env cplt
+    in
+    if_st thenn (Some variables)
+  else
+    Printf.sprintf "%s\n%s"
+      (export_in_shell shell ("OPAM_NOEVAL_ENV", string_of_bool @@ not eval_env,
```
AltGr: Should be `OPAMNOEVALENV`. Ugly, I agree, but consistent with all other env variables opam reads.

src/state/opamEnv.ml
```diff
@@ -609,8 +658,14 @@ let display_setup root ~dot_profile shell =
         && not complete_zsh then
           not_set
         else ok in
-      [ ("init-script"     , Printf.sprintf "%s" pretty_init_file);
-        ("auto-completion" , completion);
+      let eval_env =
+        try
+          if bool_of_string @@ Sys.getenv "OPAM_NOEVAL_ENV" then ok else not_set
```
AltGr: `OpamStd.Config.env_bool "NOEVALENV"` (with the variable as `OPAMNOEVALENV` above)

PR #3317, opened 2018-04-27
Cache for pin download

rjbou:
When running `opam pin distant_url`, there was a first download of the archive
or git repo to guess package name, and another ones for every package name
found. This PR adds a cache for these case of pin, in order to download
only once.

### DIFF
--- a/src/repository/opamRepositoryPath.ml
+++ b/src/repository/opamRepositoryPath.ml
@@ -15,6 +15,18 @@ let create root name = root / "repo" / OpamRepositoryName.to_string name
 
 let download_cache root = root / "download-cache"
 
+let pin_cache_dir =
+  let dir =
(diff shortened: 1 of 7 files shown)


src/repository/opamRepositoryPath.ml
```diff
@@ -15,6 +15,19 @@ let create root name = root / "repo" / OpamRepositoryName.to_string name
 
 let download_cache root = root / "download-cache"
 
+let pin_cache_dir =
+  let dir =
+    lazy (OpamFilename.Dir.of_string (OpamStd.Sys.tmp ()) /
+          (Printf.sprintf "opam-pin-cache-%d" @@ OpamStubs.getpid ()))
```
AltGr: `OpamSystem.temp_basename "opam-pin-cache"` ? Or even `OpamSystem.mk_temp_dir ()` ?
Either of which would need to be exported.
The added random bits would mitigate /tmp race attacks.

PR #3324, opened 2018-04-30
A few fixes

rjbou:
Move message, remove comment, clean config

### DIFF
--- a/src/state/opamEnv.ml
+++ b/src/state/opamEnv.ml
@@ -698,13 +698,6 @@ let setup_interactive root ~dot_profile shell =
 
   OpamConsole.header_msg "Required setup - please read";
 
-  let eval_env =
-    OpamConsole.confirm ~default:false
-      "Additionally, a hook can be added to your shell to ensure that \
-       the opam environement remains in sync. Add a hook?"
-  in
-  if eval_env then write_init_shell_scripts root ~completion:true ~eval_env;
-
   OpamConsole.msg
     "\n\
     \  In normal operation, opam only alters files within ~/.opam.\n\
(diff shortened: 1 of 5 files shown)


src/state/opamEnv.ml
```diff
@@ -725,6 +718,12 @@ let setup_interactive root ~dot_profile shell =
     (OpamConsole.colorise `cyan @@ OpamFilename.prettify dot_profile)
     (OpamConsole.colorise `bold @@ source root ~shell (init_file shell))
     (OpamConsole.colorise `bold @@ "eval $(opam env)");
+  let eval_env =
+    OpamConsole.confirm ~default:false
+      "Additionally, a hook can be added to your shell to ensure that \
+       the opam environement remains in sync. Add a hook?"
```
dra27: There's one too many e's in environment

AltGr: (or one too few `n`s ;) )
more seriously, shouldn't that be even later, after the `"Do you want to modify"` question ?
In practice, to the user, this question makes little sense if the answer to the other one is `no`, since they won't source the file anyway (they are more likely to just run `eval $(opam env)`) ; so it should probably only be asked if the other answer was `y` (or `f`).

AltGr:
@dra27: it's in the `InitConfig` (`opamrc`) file, there is no purpose in having it in `~/.opam/config` file as well.

src/state/opamEnv.ml
rjbou: Ok to have it after the "Do you want to modify" question, but if user doesn't want to modify it's `.profile` file, it does not mean that he won't use `init` scripts (it's my case). Agree that it's for non-beginner user...

rjbou:
@dra27 Besides, those scripts are written at init step, there is no need to keep them in config file. If the user want to change them, the action does not concern current config file, but he has to run `opam init` again, with its `opamrc` specifying new init scripts.

src/state/opamEnv.ml
```diff
@@ -725,19 +718,29 @@ let setup_interactive root ~dot_profile shell =
...
+          OpamConsole.msg "Alright, assuming you changed your mind, \
+                           not performing any changes.\n";
+          false
+        | Some f -> update (Some (OpamFilename.of_string f))
+      end
+    | _ -> update None
+  in
+  let eval_env =
+    OpamConsole.confirm ~default:false
+      "Additionally, a hook can be added to init script to ensure that \
```
AltGr: Alright, but then be more precise `"Additionally, a hook can be added to opam's init scripts to ensure that the shell remains in sync with the opam environment when they are loaded"`. Not sure it should be `"Additionally"`, in case you answered `n` above, so maybe `"A hook can also..."`

AltGr:
Actually, isn't the same true for `required-tools`? Shouldn't they remain in `opamrc` only ?

rjbou:
True, `required-tools` and `recommended-tools` are only used at init step...

src/state/opamEnv.ml
```diff
@@ -725,19 +718,30 @@ let setup_interactive root ~dot_profile shell =
...
+                           not performing any changes.\n";
+          false
+        | Some f -> update (Some (OpamFilename.of_string f))
+      end
+    | _ -> update None
+  in
+  let eval_env =
+    OpamConsole.confirm ~default:false
+      "A hook can be added to opam's init script to ensure that the opam \
+       environment remains in sync with the opam environment when they are \
```
AltGr: `the opam environment remains in sync with the opam environment` ? ;)

AltGr: also, either `script*s*`, or `when *it is* loaded`.

PR #3326, opened 2018-05-02
Add sandboxing support for OSX

AltGr:


src/state/sandbox_exec.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/sh -ue
+
+POL='(version 1)(allow default)(deny network*)(deny file-write*)'
```
avsm: is the default behaviour of the Linux one also to deny network connections?  That's a sensible default, but probably shouldn't diverge between operating systems

src/state/sandbox_exec.sh
```diff
@@ -0,0 +1,38 @@
...
+        add_mounts ro "$OPAM_SWITCH_PREFIX/.opam-switch"
+        add_mounts rw "$PWD"
+        ;;
+    remove)
+        add_mounts rw "$OPAM_SWITCH_PREFIX"
+        add_mounts ro "$OPAM_SWITCH_PREFIX/.opam-switch"
+        [ "X${PWD#$OPAM_SWITCH_PREFIX/.opam-switch}" != "X${PWD}" ] && add_mounts rw "$PWD"
+        ;;
+    *)
+        echo "bwrap.sh: unknown command $COMMAND, must be one of 'build', 'install' or 'remove'" >&2
```
avsm: this bwrap.sh could be $0 to be more accurate

src/state/sandbox_exec.sh
AltGr: yes, it's the same on Linux. There are now built-in mechanisms for package needing to download extra files, so better to enforce them.

src/state/sandbox_exec.sh
AltGr: Indeed, thanks!

avsm:
working fine for me with a fresh switch -- might need some testing to find out what happens for rc1->rc2 users

src/state/sandbox_exec.sh
avsm: Thanks, I agree that these are excellent defaults.

PR #3329, opened 2018-05-03
Add an option --disable-sandboxing

AltGr:
for the brave, or working within unprivileged containers

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -434,8 +434,8 @@ let update
     else if all then all_repos
     else OpamSwitchState.repos_list st
   in
-  let packages =
-    if repos_only then OpamPackage.Set.empty else
+  let packages, ignore_packages =
+    if repos_only then OpamPackage.Set.empty, OpamPackage.Set.empty else
     let packages = st.installed ++ st.pinned in
     let packages =
       if names = [] then packages else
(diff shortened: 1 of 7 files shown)


dra27:
Just to clarify the detail of what I mean, I'd propose that by default `--disable-sandboxing` would display a warning (or error?) that opam was built without that support but if you pass, say, `--enable-disable-sandboxing` then the resulting opam would honour the option

AltGr:
I was on the same opinion as @dra27 here initially, which is why we preferred not to provide any facility for disabling sandboxing (see https://github.com/ocaml/opam/blob/34631c5a8e480c35110f05ea443dc22ae92709c2/doc/pages/FAQ.md#--why-does-opam-require-bwrap);

but after discussions with @avsm, who insisted this made some uses (for e.g. containers) unnecessarily difficult, decided to add it still. I am not sure it matters that much actually, if the option is presented with due warnings. Merging this now to avoid conflicts (the PR also contains some other fixes), but the discussion remains open and the option could easily be withdrawn.

avsm:
We definitely want the majority of users to use sandboxing, but that's no reason to punish users that (for whatever reason) do not or cannot get it work easily.  I propose that sandboxing is enabled by default on every platform that supports it, and that we actively attempt to support currently unsupported ones (especially FreeBSD, OpenBSD and Windows).  Once we have coverage there, a future version of opam can make it a hard requirement.  For the moment for the 2.0 release, large warning messages if the user manually disables it seems fine.

PR #3337, opened 2018-05-06
Fix ocaml-system and ocaml scripts for Windows

dra27:
The config-generation scripts in the `ocaml` and `ocaml-system` packages don't take the `.exe` extension on Windows **and Cygwin** into account. Proposing this now to solve Cygwin, as that should be otherwise working in 2.0.0.

The `set_binary_mode_in` is for native Windows - it should have no effect on any other platform, but on Windows it will cause the trailing `\r` in the output of `ocamlc -where` to be stripped (which is the behaviour required).


AltGr:
Thanks!

Issue #3334, opened 2018-05-06
opam upgrade should explain things being kept back better.


pmetzger:
Say you type `opam upgrade` and see this:

```
$ opam upgrade 
Everything as up-to-date as possible (run with --verbose to show unavailable
upgrades).
The following would require downgrades or uninstalls, but you may upgrade them
explicitly:
  - lwt.4.0.1
$
```

So what seems to be going on here is (for example) that I have utop 2.1.0  installed which conflicts with lwt 4.0.1, and I'm currently at lwt 3.3.0.

What would be nice would be if the message explained this. Right now it's confusing. First, the line that says:

```
The following would require downgrades or uninstalls, but you may upgrade them
explicitly:
```

is confusing. (Trust me, it confused me at first.) It might be better if it said something like:

```
The following packages are not being upgraded because the new versions
conflict with other installed packages that would need to be downgraded
or uninstalled. However, you may "opam upgrade" these packages explicitly,
which will ask permission to downgrade or uninstall the conflicting packages.
```

And second, it would be nice if it said something like:

```
- lwt.4.0.1 (3.3.0 installed, 4.0.1 conflicts with utop (all versions).)
```

(Perhaps some would argue for the second part only if you say `--verbose` but I think it should always tell you.)

dra27:
I agree - and especially with Jane Street packages like in your referenced example, it would be handy to have an option for something clearer. It probably shouldn't be done by default, because you need to invoke the solver each time, but I wonder if `--verbose` could try the following algorithm:

 - Take the first package which can't be upgraded and solve `opam upgrade X`.
 - Note package downgrades and removals required
 - Remove any packages which are upgraded as part of that solution to their targets from the list
 - etc.

That should allow you a messages like:

```
Requiring utop to be removed:
 - lwt.4.0.1
```

or, from your previous examples:
```
Requiring async_graphics to be removed:
 - etc.
```

There might possibly be some additional cleverness for the possibility of individual sets conflicting with each other, though that might be taking it too far.

pmetzger:
So changing the error message to be clearer (I wrotea new one above) should be a simple commit. Maybe that could go in first, and then the rest of this could be considered separately.

dra27:
Yes, indeed - do you want to do a PR?

AltGr:
```
The following packages are not being upgraded because the new versions
conflict with other installed packages that would need to be downgraded
or uninstalled. However, you may "opam upgrade" these packages explicitly,
[...]

PR #3339, opened 2018-05-07
Improve an error message in "opam upgrade"

pmetzger:
The error message given when an "opam upgrade" leaves some
packages alone because newer versions would conflict with other
installed packages was confusing to naive users. The new one is a bit
longer and much more explicit.

This is a partial fix for #3334.


AltGr:
Thanks!

pmetzger:
Eeep! I hope the CI system still passes. I didn't do a full build myself.

PR #3343, opened 2018-05-09
Allow file-depends to include deleted files

dra27:
Currently, for `file-depends`, opam can detect when a file changes or when it is deleted. What it can't detect at the moment is when a file **appears**.

This alters the logic for `file-depends` so that if the checksum given is zero then the file is expected **not** to exist.

This allows, for example, the `ocaml-system` package to depend on the *lack* of `graphics.cmi` and so trigger a recompile if the user on Debian, for example, has `ocaml-nox` installed and then adds `ocaml`.


AltGr:
Hmm, interesting and seems a good idea ; thanks! Did you check if the manual needed an update ?

PR #3344, opened 2018-05-09
Refactor 'opam init', remove 'opam config setup'

AltGr:
The new code should be more straightforward, with a static init script, and
the two options (env_hook, completion) controlled by the presence or
absence of the specific scripts.


AltGr:
Another addition could be to use the check for wether the .profile line was added, and be able to remove it as well.

PR #3346, opened 2018-05-10
Various bits 'n bobs

dra27:
This is a first chunk of #3260, separated off. Hopefully these are the non-controversial parts.


AltGr:
Thanks!

PR #3347, opened 2018-05-10
Fix file locking on Windows

dra27:
Another bit from #3260.

These three commits are worth looking at individually.

 - `Close all lock files on error`: this, as you noted in #3260, this shouldn't be necessary, but given that any coding error could result in this, I'd prefer to keep it. It could possibly be extended to display warnings if `OpamSystem.release_all_locks` actually has to release any locks? It would seem to be much better to have warnings displayed, which need fixing, than to get error messages about being unable to remove a directory (in this case `~/.opam`), as failing to delete the directory in this instance will probably result in worse error messages for the user on the next command they run.
 - `Fix file locking on Windows`: this is sound, though could potentially result in the loss of the lock to another opam process. This could be fixed by converting locks to be pairs of read and write locks if you want? A benefit of doing that is the code would cease to be Windows-specific (it would just retain a comment for why it can't be done the more obvious way).
 - `Fix reading/writing of OpamRepositoryState cache`: is a simple refactoring to prevent opening a file twice, since this doesn't work on Windows if it's locked for writing (even by the same process).

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -761,8 +761,10 @@ let init
          else OpamFile.InitConfig.default_compiler init_config), custom_scripts
       with e ->
         OpamStd.Exn.finalise e @@ fun () ->
-        if not (OpamConsole.debug ()) && root_empty then
-          OpamFilename.rmdir root)
+        if not (OpamConsole.debug ()) && root_empty then begin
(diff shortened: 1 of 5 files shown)


AltGr:
Seems fine, thanks!

PR #3348, opened 2018-05-10
Allow native Windows to use Cygwin tools

dra27:
Yet another bit from #3260.

~This also borrows a commit from #3346, so it will need rebasing before merge.~

This set of patches address two fundamental issues:
 - It allows native Windows opam to call Cygwin tools correctly, by converting their arguments to Cygwin-style paths when necessary.
 - It allows native Windows opam to appear to be able to execute shell scripts directly (by interpreting `#!` comments) and also ensures that command line arguments passed to Cygwin executables are correctly encoded/escaped, since they don't follow the same rules as native Windows executables. Note that this has **nothing** to do with shell escaping: this is dealing with the conversion of arguments as a string list to a single string (which is how Windows passes the command line) by the calling process being recovered as the *same* string array by the invoked process.

### DIFF
--- a/src/core/opamProcess.ml
+++ b/src/core/opamProcess.ml
@@ -9,8 +9,187 @@
 (*                                                                        *)
 (**************************************************************************)
 
+open OpamCompat
+
 let log ?level fmt =
   OpamConsole.log "PROC" ?level fmt
+
+let cygwin_create_process_env prog args env fd1 fd2 fd3 =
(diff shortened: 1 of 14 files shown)


AltGr:
Wow, this is another heavy piece of work ! LGTM :)

rjbou:
No, ok for me. Thanks a lot for this work!

PR #3349, opened 2018-05-10
Adapt patch behaviour for LF/CRLF compatibility

dra27:
A rather more controversial part of #3260!

~This relies on ~two~three bits of #3346, but the actual content of this PR still needs work/discussion anyhow.~

Further comment ~to follow~ below.

### DIFF
--- a/src/core/opamSystem.ml
+++ b/src/core/opamSystem.ml
@@ -821,13 +821,320 @@ let lock_none = {
 let lock_isatleast flag lock =
   lock_max flag lock.kind = lock.kind
 
+(* read_lines processes the lines of a text file in a platform-agnostic manner
+   (which is to say that it preserves '\r' on all platforms). It has two modes
+   of operation - in the Lines mode, it returns a string list of the lines, in
+   CrLf mode, it returns true if every line ends "\r\n". *)
+type _ read_return = Lines : string list read_return
+                   | CrLf  : bool read_return
+
+let read_lines : type s . s read_return -> string -> s = fun mode file ->
+  let ch =
+    try open_in_bin file
+    with Sys_error _ -> raise (File_not_found file)
+  in
+  let read_lines f =
+    let rec read_lines cr acc =
+      match input_line ch with
+      | line ->
+          let length = String.length line in
+          let acc = f line acc in
+          read_lines (cr && length > 0 && line.[length - 1] = '\r') acc
+      | exception End_of_file ->
+          close_in ch;
+          let acc =
+            if cr then
+              List.rev_map (fun s -> String.sub s 0 (String.length s - 1)) acc
+            else
+              List.rev acc
+          in
+          (acc, cr)
+    in
+    read_lines true []
+  in
+  match mode with
+  | Lines ->
+      read_lines OpamStd.List.cons |> fst
+  | CrLf ->
+      read_lines (fun _ _ -> []) |> snd
+
+let translate_patch ~dir orig corrected =
+  (* CRLF detection with patching can be more complicated than that used here,
+     especially in the presence of files with mixed LF/CRLF endings. The
+     processing done here aims to allow patching to succeed on files which are
+     wholly encoded CRLF or LF against patches which may have been translated to
+     be the opposite.
+
+     The resulting patch will *always* have LF line endings for the patch
+     metadata (headers, chunk locations, etc.) but uses either CRLF or LF
+     depending on the target file. Endings in the patch are always preserved for
+     new files. The benefit of always using LF endings for the metadata is that
+     patch's "Stripping trailing CRs from patch" behaviour won't be triggered.
+
+     There are various patch formats, though only the Unified and Context
+     formats allow multiple files to be patched. I tired of trying to get
+     sufficient documented detail of Context diffs to be able to parse them
+     without resorting to reverse-engineering code. It is unusual to see them
+     these days, so for now opam just emits a warning if a Context diff file is
+     encountered and does no processing to it.
+
+     There are various semantic aspects of Unified diffs which are not handled
+     (at least at present) by this function which are documented in the code
+     with the marker "Weakness". *)
+  let process_chunk_header result line =
+    match OpamStd.String.split line ' ' with
+    | "@@"::a::b::"@@"::_ ->
+        (* Weakness: for a new file [a] should always be -0,0 (not checked) *)
+        let l_a = String.length a in
+        let l_b = String.length b in
+        if l_a > 1 && l_b > 1 && a.[0] = '-' && b.[0] = '+' then
+          try
+            let f (_, v) = int_of_string v in
+            let neg =
+              OpamStd.String.cut_at (String.sub a 1 (l_a - 1)) ','
+                      |> OpamStd.Option.map_default f 1
+            in
+            let pos =
+              OpamStd.String.cut_at (String.sub b 1 (l_b - 1)) ','
+                      |> OpamStd.Option.map_default f 1
+            in
+            result neg pos
+          with e ->
+            OpamStd.Exn.fatal e;
+            (* TODO Should display some kind of re-sync warning *)
+            `Header
+        else
+          (* TODO Should display some kind of re-sync warning *)
+          `Header
+    | _ ->
+        (* TODO Should display some kind of warning that there were no chunks *)
+        `Header
+  in
+  let process ch (state, file_mode) line =
+    let length = String.length line in
+    let next_state =
+      let state =
+        match state with
+          (* Weakness: no effort is made to identify the diff or Index: lines *)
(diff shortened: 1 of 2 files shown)


src/core/opamHash.ml
```diff
@@ -82,31 +82,93 @@ let to_json s = `String (to_string s)
...
+          let len = len kind in
+          make kind (String.sub l (String.length l - len) len)
+        | _ -> failwith "openssl error"
+      with OpamSystem.Command_not_found _ | Exit ->
+        make kind (OpamSHA.hash kind file)
+  in
+  let probably_binary name =
+    List.mem (Filename.extension name) [".zip"; ".tar"; ".gz"; ".tgz"; ".tbz"; ".txz"; ".tlz"]
+  in
+  if OpamStd.Option.map_default ((=) primary) true target || probably_binary file then
```
AltGr: I do like the `Std.Option` combinators, but `target = None || target = Some primary` would probably be easier to read here ^^.

dra27: Oh all right, then 🙂

dra27:
Finally, some notes on this!

The first issue is the hashing. openssl has nothing to solve - the files are fundamentally different here because (usually) of Git. Internet protocols typically use CRLF (email, http, etc.), so this doesn't come up since the standards already declare the data to be hashed as being supposed to have CRLF line endings.

I have to confess that it just feels tediously necessary, rather than wrong to me. All that is potentially happening is that we identify a simple and, at least I hope, non-exploitable transformation of the file which may have occurred. The alternative would be to force (presumably) Unix line endings, but this seems highly unsatisfactory to me, especially as the one likely to get pinged about the confusing bug reports :wink:

dra27:
I guess we could re-implement patch, yes. There would be another benefit, in that I've certainly hit a problem with the `patch` command on macOS not being able to cope with a patch+rename operation (which Git will happily emit).

My only concern is how easy it is to replicate the same kind of fuzzy patching as standard patch, but in fairness I think that that's just a matter of seeing whether the patch applies near to the line numbers given when it doesn't apply exactly - I think the context still has to match exactly.

dra27:
Impact for each of these:

 - The hashing change I think should definitely be in 2.0.0 as it means that the whole of 2.x will behave the same way - it could be really annoying (i.e. result in more bug reports) if opam 2.0 Unix users have systems which can't cope with the odd CRLF submitted file in opam-repository, etc.
 - The patch change has the same argument - it's just annoying if those patches generated against one kind of repo can't be applied to another.

rjbou:
Thanks for the work! 
I'm wondering if it is the good choice to add it in 2.0.0 as there is still some work to be done: most TODO are for warning, but there is also TODO that are `assert false` that could happen.
As git can handle CRLF (with config option `core.autocrlf`), we can for the moment have a local config using it in 2.0.0, and add a further handling in future version.

dra27:
@rjbou - no, Git can't handle this, that's the point. You still don't know what's actually checked into the repo (see myriad guides on how to convert repos to use core.autocrlf which weren't using it correctly in the first place - the OCaml change was a nightmare)

dra27:
I'm extremely happy to polish this patch off properly for 2.0.0, it just needs the further discussion as to whether it's headed in the correct direction.

dra27:
Incidentally, for context diffs, one thought I had had was to eliminate the processing - i.e. just emit context diffs as they are, possibly with a warning. Surely no one is actually using them these days for real patches?

src/core/opamSystem.ml
```diff
@@ -821,11 +821,236 @@ let lock_none = {
...
+                    let neg =
+                      OpamStd.String.cut_at (String.sub a 1 (l_a - 1)) ','
+                              |> OpamStd.Option.map_default f 1
+                    in
+                    let pos =
+                      OpamStd.String.cut_at (String.sub b 1 (l_b - 1)) ','
+                              |> OpamStd.Option.map_default f 1
+                    in
+                    `Processing (`Unified, orig, target, crlf, `Chunk (neg, pos))
+                  with _ ->
```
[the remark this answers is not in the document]
dra27: <sigh> Yes, indeed - I feel the need for a compiler warning to help lazy programmers like myself with this one!

dra27:
OK, this needs some more testing (principally that I must rebase #3260 to use this version instead of the old one), but this revised version:

 - Removes various TODO items (including deleting the temporary patch file 😊)
 - Adds a missing case of ensuring CR is *stripped* if the target file uses LF-endings
 - Skips Context diffs with a warning - the remaining `assert false`s are genuinely unreachable cases, therefore.

There is a note about dealing with renaming files. This is something Git certainly does but, as noted in https://github.com/ocaml/opam-repository/pull/11207#issuecomment-358594175, this is not portable, so its absence here is not (I think) critical - it should be the case that the patch will pass through unaltered since the code would incorrectly identify it as a new file.

dra27:
Well, the important OS is working :wink:

AltGr:
Thanks!
I guess was @rjbou omitted to say about using git's `core.autocrlf input` is that it would solve client problems, on Windows, _if_ we enforce that the repository itself is purely in Unix format, which is an option and should already be 99% the case. I am not completely sure even that holds, though. It also assumes that patch files are supplied from the repository, while in practice they could come from the package source itself (although I doubt that is used in practice ?)

About the patches, since we already restricted patches to apply with `-p 1` only on opam 2 (without too much trouble, just a couple had to be fixed) — I guess it would be OK to restrict the patch format some more. I'd presonnally be fine as long as we handle the `git format-patch` format, but it'd be easy to check what we actually have in the repository.

And about the hashing function, does that mean that hashing a whole file is now equivalent to hashing the string as obtained through `open_in`, as opposed as `open_in_bin` ?

src/core/opamSystem.ml
```diff
@@ -821,13 +821,344 @@ let lock_none = {
 let lock_isatleast flag lock =
   lock_max flag lock.kind = lock.kind
 
+(* read_lines processes the lines of a text file in a platform-agnostic manner
+   (which is to say that it preserves '\r' on all platforms). It has two modes
+   of operation - in the Lines mode, it returns a string list of the lines, in
+   CrLf mode, it returns true if every line ends "\r\n". *)
+type _ read_return = Lines : string list read_return
+                   | CrLf  : bool read_return
```
AltGr: why not just define two functions ?

src/core/opamSystem.ml
```diff
@@ -821,13 +821,344 @@ let lock_none = {
...
+     depending on the target file. Endings in the patch are always preserved for
+     new files. The benefit of always using LF endings for the metadata is that
+     patch's "Stripping trailing CRs from patch" behaviour won't be triggered.
+
+     There are various patch formats, though only the Unified and Context
+     formats allow multiple files to be patched. I tired of trying to get
+     sufficient documented detail of Context diffs to be able to parse them
+     without resorting to reverse-engineering code. It is unusual to see them
+     these days, so for now opam just emits a warning if a Context diff file is
+     encountered and does no processing to it.
```
AltGr: Agreed, maybe just add a line to the doc where it explains the patch format (and that they must apply with `-p 1`) that they should be unified diffs.

src/core/opamSystem.ml
dra27: Where's the fun in that 🙂 More seriously, they're very similar, but not quite identical functions

AltGr: i'd argue that you could factorise the core in `read_lines_aux`, and define the two functions with just the last lines. But well, doesn't really matter ^^.

src/core/opamSystem.ml
dra27: I updated the sentence in the docs

dra27:
Docs updated and I also factored out the chunk processing code to `process_chunk_header` at @rjbou's suggestion (contrary to my previous doubt, I had clearly just been lazy - the handling *is* the same in both cases!)

dra27:
@AltGr - indeed, the problem is that you don't know what the patch files themselves may be. While we can enforce that opam-repository uses Unix line-endings, you can't in general enforce that, even by `core.autocrlf` - `.gitattributes` can override it, and if the files are already in the repo with `\r\n` endings, then neither `core.autocrlf` or `.gitattributes` will automatically change that (in fact, they cause problems as it means that the files will always be detected as having changed, even on normal checkout)

dra27:
For the hashing, sort of - it means that it will hash the file both as if it had been read with `open_in_bin` (old behaviour) and also as though it had been opened **on Windows** with `open_in` (where `"\r\n"` gets reduced to `"\n"`) - but note that the `"\r\n"` translation is being done on all OSes here - normally, of course, `open_in_bin` and `open_in` are identical on Unix.

rjbou:
Thanks!

PR #3350, opened 2018-05-10
Deal with Windows path conventions (backslashes, .exe, etc.)

dra27:
And the last portion of #3260 split off.

This borrows parts of ~#3346~ and #3348 but also contains at least bit which needs further discussion.

Further comment to follow.

### DIFF
--- a/src/client/opamAction.ml
+++ b/src/client/opamAction.ml
@@ -88,13 +88,14 @@ let process_dot_install st nv build_dir =
                 let base' =
                   {base with c = OpamFilename.Base.add_extension base.c "exe"} in
                 if OpamFilename.exists (OpamFilename.create build_dir base'.c) then begin
-                  OpamConsole.warning "Added .exe to %s; .install file should be fixed" (OpamFilename.Base.to_string base.c);
+                  OpamConsole.warning ".install file is missing .exe extension for %s" (OpamFilename.Base.to_string base.c);
                   (base', true)
                 end else
                   (base, false)
               else
                 (base, false) in
             let src_file = OpamFilename.create build_dir base.c in
+            if append then warning (OpamFilename.to_string src_file) `Add_exe;
             let dst_file = match dst with
               | None   -> OpamFilename.create dst_dir (OpamFilename.basename src_file)
               | Some d ->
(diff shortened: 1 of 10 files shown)


dra27:
The two commits about installation add useful warnings and shims to allow programs to be installed correctly (which should be becoming less common with Dune, etc. anyway).

The problem part is the variable interpolation. This is indeed arising from files mentioned in `substs` fields (that's where the one above came from). The problem here is that `substs` command is blind to what it's actually doing. Thoughts?

src/client/opamAction.ml
```diff
@@ -66,10 +66,25 @@ let process_dot_install st nv build_dir =
           OpamFilename.mkdir dst_dir;
         );
         List.iter (fun (base, dst) ->
+            let (base, append) =
+              if exec && not (OpamFilename.exists (OpamFilename.create build_dir base.c)) then
+                let base' =
+                  {base with c = OpamFilename.Base.add_extension base.c "exe"} in
+                if OpamFilename.exists (OpamFilename.create build_dir base'.c) then begin
+                  OpamConsole.warning "Added .exe to %s; .install file should be fixed" (OpamFilename.Base.to_string base.c);
```
AltGr: It's best in general to avoid displaying messages that only make sense to developpers/packagers to the end users, unless they have a clear way to report; but I am not sure how best to do this.

dra27: Perhaps also noting that the package maintainer should be informed?

dra27:
I had some thoughts on the `substs` problem:

 - Adding a predicate/flag to the substs command to either enable or disable escaping
 - Possibly changing the default to be with escaping of backslashes ... but I haven't checked uses in opam-repository to see whether that would break existing packages (but it's probably too late to do that)
 - Instead of relying on just `opam-version` being the first line, instead see if the file parses as an opam file and enable the heuristic then (that seems a much better idea)

dra27:
Refactored the interface for the warnings and altered so that running with `--strict` turns these warnings into errors - see [this AppVeyor log](https://ci.appveyor.com/project/dra27/opam/builds/25492165/job/sf096inaln05qw31#L457)

PR #3354, opened 2018-05-14
Doc fix, formula printer & upgrade message

### LINKED ISSUE #3318: install opam man pages in .opam/[version]/man directory
Right now, no manual pages are installed for opam, one reads the man pages with `opam --help`, but this precludes the possibility of reading the man pages with alternative manual readers like Emacs's `M-x man` support.

rjbou:
This PR contains:
- Some documentation fixes
- Formula printer: print inner parenthesis, even if it is in a `Block`, to avoid conflicts on regenerated `opam` files 
- Improve upgrade message in case of conflict (related to #3334)
- Fix manpage installation: they need to be generated before `opam-actual.install` generation, in order to be installed by `opam-installer` (close #3318)

### DIFF
--- a/doc/pages/FAQ.md
+++ b/doc/pages/FAQ.md
@@ -87,8 +87,9 @@ You can manually disable package build sandboxing and remove bwrap from
 the required dependencies (at your own risk). You can use the built-in
 configuration as a template to create or update an `opamrc` file: run `opam
 init --show-default-opamrc >~/.opamrc`, then edit that file and remove or
-modify the `init-scripts:` and `wrap-*:` fields as well as the `bwrap` line
-from the `required-tools:` field, and finally retry `opam init`.
+modify the [`init-scripts:`](Manual.html#opamrcfield-init-scripts) and
+`wrap-*:` fields as well as the `bwrap` line from the `required-tools:` field,
+and finally retry `opam init`.
 
 See also the [wrap entry](Manual.html#configfield-wrap-build-commands) section
 in the manual.
--- a/doc/pages/Manual.md
+++ b/doc/pages/Manual.md
@@ -721,7 +721,7 @@ The default, built-in initial config of <span class="opam">opam</span> can be se
   of alternative commands to look for in the `PATH`, optionally a specific error
   message to display if none of them is found, and a filter that can make the
   check conditional.
-- <a id="configfield-init-scripts">`init-scripts: [ [ <string> <string> ] { <filter> } ... ]`</a>:
+- <a id="opamrcfield-init-scripts">`init-scripts: [ [ <string> <string> ] { <filter> } ... ]`</a>:
   These scripts will be written verbatim into the hook directory
   (`~/.opam/opam-init/hooks`) upon initialisation. The first string is the file
   name of the script, the second its raw contents, and the filter allows to
@@ -1331,12 +1331,14 @@ for <span class="opam">opam</span>.
   specify wrappers around every command executed during the corresponding action
   of any package. The command-line elements will be prefixed to the package
   command, so for example command `[ make "install" ]` with wrapper
-  `[ "time" "-o" "/tmp/timings" "-a" ]` will result in the command
-  `[ "time" "-o" "/tmp/timings" "-a" make "install" ]`.
-  As `init-scripts:` are stored in the hook directory, when using a wrapper
-  script defined by `init-scripts:`, use the variable `%{hook}%` as prefix for
-  you script filename.
-  The filters are evaluated in the same scope as the package commands.
+  `[ "time" "-o" "/tmp/timings" "-a" ]` will result in the command `[ "time"
+  "-o" "/tmp/timings" "-a" make "install" ]`. The filters are evaluated in the
+  same scope as the package commands.
+
+    As [`init-scripts:`](#opamrcfield-init-scripts) are stored in the hook
+    directory, when using a wrapper script defined by
+    [`init-scripts:`](#opamrcfield-init-scripts), use the variable `%{hook}%` as
+    prefix for you script filename.
 - <a id="configfield-post-build-commands">`post-build-commands: [ [ <term> { <filter> } ... ] { <filter> } ... ]`</a>,
   <a id="configfield-post-install-commands">`post-install-commands: [ [ <term> { <filter> } ... ] { <filter> } ... ]`</a>,
   <a id="configfield-post-remove-commands">`post-remove-commands: [ [ <term> { <filter> } ... ] { <filter> } ... ]`</a>:
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -315,6 +315,65 @@ let upgrade_t ?strict_upgrade ?auto_install ?ask ?(check=false) ~all atoms t =
              (OpamPackage.names_of_packages t.compiler_packages)
          in
          let unopt = unopt -- base in
+         let conflicts =
+           let get_formula pkg =
+             OpamStd.Option.map
+               (fun opam ->
+                  OpamFilter.filter_formula ~default:false
+                    (OpamPackageVar.resolve_switch ~package:pkg t)
+                    (OpamFile.OPAM.conflicts opam))
+               (OpamSwitchState.opam_opt t pkg)
+           in
+           OpamPackage.Set.fold (fun unopt_pkg map ->
+               let set =
+                 OpamSwitchState.conflicts_with t
+                   (OpamPackage.Set.singleton unopt_pkg) latest in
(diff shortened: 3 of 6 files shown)


src/format/opamFormat.ml
```diff
@@ -462,10 +462,39 @@ module V = struct
   let package_atom constraints =
     map_option pkgname constraints
 
+  (* These two functions are duplicated from [OpamFormula] but we need to have a
+     it here because of a change on [Block] handling: to have a coherent
+     printing, we must not always discard them *)
+  let rec ands_to_list = function
+    | Empty -> []
+    | And (e,f)
+    | Block (And (e,f)) ->
```
AltGr: Is even this line needed ? Not having it would preserve cases such as `[ "foo" ("bar" & "baz") ]`, while with the line it would be re-printed as `[ "foo" "bar" "baz" ]`. Not sure if it is best, but at least it should ensure that `parse (print x) = x`, which is no longer the case if we don't preserve all blocks ?

The other approach would be to give up on the `Block` constructor, and consequently on preserving user parentheses ; but I think it's best if they remain for now (even if just for preserving camelus 1.2→2.0 rewrite consistency).

doc/pages/Manual.md
```diff
@@ -1329,12 +1329,14 @@ for <span class="opam">opam</span>.
   specify wrappers around every command executed during the corresponding action
   of any package. The command-line elements will be prefixed to the package
   command, so for example command `[ make "install" ]` with wrapper
-  `[ "time" "-o" "/tmp/timings" "-a" ]` will result in the command
-  `[ "time" "-o" "/tmp/timings" "-a" make "install" ]`.
-  As `init-scripts:` are stored in the hook directory, when using a wrapper
-  script defined by `init-scripts:`, use the variable `%{hook}%` as prefix for
-  you script filename.
-  The filters are evaluated in the same scope as the package commands.
+	`[ "time" "-o" "/tmp/timings" "-a" ]` will result in the command `[ "time"
```
AltGr: ???

doc/pages/FAQ.md
```diff
@@ -87,8 +87,9 @@ You can manually disable package build sandboxing and remove bwrap from
 the required dependencies (at your own risk). You can use the built-in
 configuration as a template to create or update an `opamrc` file: run `opam
 init --show-default-opamrc >~/.opamrc`, then edit that file and remove or
-modify the `init-scripts:` and `wrap-*:` fields as well as the `bwrap` line
-from the `required-tools:` field, and finally retry `opam init`.
+modify the [`init-scripts:`](Manual.html#opamrcfield-init-scripts) and
+`wrap-*:` fields as well as the `bwrap` line from the `required-tools:` field,
+and finally retry `opam init`.
```
AltGr: Hmm, now there is `--disable-sandboxing`. But not sure if we want to advertise it, or even if it'll remain...

src/client/opamClient.ml
```diff
@@ -315,6 +315,63 @@ let upgrade_t ?strict_upgrade ?auto_install ?ask ?(check=false) ~all atoms t =
              (OpamPackage.names_of_packages t.compiler_packages)
          in
          let unopt = unopt -- base in
+         let get_formula pkg get =
+           OpamStd.Option.map
+             (fun opam ->
+                OpamFilter.filter_formula ~default:false
+                  (OpamFilter.deps_var_env ~build:true ~post:false
+                     ~test:true ~dev:false ~doc:false)
```
AltGr: why this choice ? My guess is that this should be
- `build:true`, `post:true` (these are the "base" dependencies as seen by the solver)
- `dev` should be determined from `OpamSwitchState.is_dev_package` IIRC
- `test` and `doc` should be determined from `!OpamStateConfig.r.build_{test,doc}`, and wether the package is a member of the `requested` set.

Most of this should be handled correctly already by `OpamPackageVar.all_depends` (plus the possible `version` variable)

src/client/opamClient.ml
```diff
@@ -315,6 +315,63 @@ let upgrade_t ?strict_upgrade ?auto_install ?ask ?(check=false) ~all atoms t =
...
+               let set =
+                 OpamSwitchState.conflicts_with t
+                   (OpamPackage.Set.singleton unopt_pkg) latest in
+               OpamPackage.Set.fold (fun installed_pkg map1 ->
+                   match get_formula installed_pkg  OpamFile.OPAM.conflicts with
+                   | None -> map1
+                   | Some conflicts_formula ->
+                     if OpamFormula.verifies conflicts_formula unopt_pkg then
+                       OpamFormula.fold_left
+                         (fun map2 (n,formula) ->
```
AltGr: Why use a different name (`map2`) if it's always the same accumulator being passed among the different depth levels ?

src/client/opamClient.ml
```diff
@@ -333,15 +390,40 @@ let upgrade_t ?strict_upgrade ?auto_install ?ask ?(check=false) ~all atoms t =
            OpamConsole.formatted_msg
              "%s (run with --verbose to show unavailable upgrades).\n" hdmsg;
          if not (OpamPackage.Set.is_empty unopt) then
-           (OpamConsole.formatted_msg
+           (let bullet =
+              OpamConsole.(colorise `red
+                             (utf8_symbol Symbols.circled_division_slash "--"))
+              ^ " "
```
AltGr: We normally use this symbol only to mean "package removal"

src/client/opamClient.ml
```diff
@@ -315,6 +315,63 @@ let upgrade_t ?strict_upgrade ?auto_install ?ask ?(check=false) ~all atoms t =
...
+                     if OpamFormula.verifies conflicts_formula unopt_pkg then
+                       OpamFormula.fold_left
+                         (fun map2 (n,formula) ->
+                            if OpamPackage.name unopt_pkg = n
+                            then
+                              OpamPackage.Map.update unopt_pkg
+                                (OpamStd.List.cons (installed_pkg, formula)) [] map2
+                            else map2
+                         ) map1 conflicts_formula
+                     else map1
```
AltGr: While not strictly correct in the case of disjunctions (you only consider atoms, so if a package has `depends: "foo" { (< "3" | > "4") & !=5 }` and you have `foo=5`, you would retain both `< "3"` and `!=5`), I think it is fine as it is (it's just a little artefact in the worst — very rare — case, and more complicated than it deserves to do better). Maybe a comment could say so, though.

rjbou: Disjunctions are handled: each package formula is checked using `OpamFormula.check_version_formula`, then added to the map for printing.

src/client/opamClient.ml
rjbou: Personal preference :)

doc/pages/FAQ.md
rjbou: imho, it shouldn't be in this FAQ field. Maybe another field at bottom, targeted for CI needs?

AltGr: Yeah... on the other hand, I feel uncomfortable with willingly providing confusing information when there is an easier way... so we'd better remove the explanation of `opamrc` altogether ?

src/client/opamClient.ml
AltGr: I'd argue that you risk mixing the different `acc`s and using the wrong one, since they all have the same type, which would cause hard-to-spot bugs, while you are safer with using the same name along the chain with masking ;)

doc/pages/Manual.md
```diff
@@ -1331,12 +1331,14 @@ for <span class="opam">opam</span>.
...
   of any package. The command-line elements will be prefixed to the package
   command, so for example command `[ make "install" ]` with wrapper
-  `[ "time" "-o" "/tmp/timings" "-a" ]` will result in the command
-  `[ "time" "-o" "/tmp/timings" "-a" make "install" ]`.
-  As `init-scripts:` are stored in the hook directory, when using a wrapper
-  script defined by `init-scripts:`, use the variable `%{hook}%` as prefix for
-  you script filename.
-  The filters are evaluated in the same scope as the package commands.
+  `[ "time" "-o" "/tmp/timings" "-a" ]` will result in the command `[ "time"
+  "-o" "/tmp/timings" "-a" make "install" ]`. The filters are evaluated in the
```
AltGr: I am not sure if it's still the case, but at some point a carriage return within the backquotes would break formatting.

rjbou: It doesn't break locally, but a check on server is needed.

doc/pages/FAQ.md
rjbou: If we remove also this explanation, some people could be stuck: users that are not admin, or just can't install `bwrap` (exotic install).

PR #3368, opened 2018-05-23
opam show: string fields printed with quotes, as lists

### LINKED ISSUE #3365: opam info for tags is hard to parse
```
$ opam show -f tags: root1d
"scientfic" "root finding"
$ opam show -f tags: mirage-xen-ocaml
org:mirage
```

This is quite hard to parse from OCaml -- sometimes the output is string-quoted (and `%S` from `scanf` works) and sometimes the output is bare.  It would be useful to make this a bit more systematic before 2.0 so that it can be depended on for use in shell and script output.

rjbou:
this PR closes #3365

### DIFF
--- a/src/client/opamListCommand.ml
+++ b/src/client/opamListCommand.ml
@@ -470,7 +470,7 @@ let version_color st nv =
 let mini_field_printer ?(prettify=false) ?(normalise=false) =
   if normalise then OpamPrinter.Normalise.value else
   function
-  | String (_, s) -> s
+  | String (_, s) when prettify -> s
   | List (_, l) when prettify &&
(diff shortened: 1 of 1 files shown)


avsm:
Does this escape contents inside the strings? That `%s` might be better off as `%S` instead

rjbou:
As `tags:` (and most of fields) are not triple quote strings, they are already escaped in the `opam` file, which is kept by the format. 
I updated the PR to match list printing, *i.e.* use `OpamPrinter.value` in case of single string field printing.

avsm:
That makes sense, thanks :)

PR #3377, opened 2018-05-28
opam init --bare: Copy bash completion script to opam-init

anemator:
Prior to this patch, a fresh install of opam (master) did not copy complete.sh to the opam-init folder.

### DIFF
--- a/src/state/opamEnv.ml
+++ b/src/state/opamEnv.ml
@@ -369,9 +369,9 @@ let eval_string gt ?(set_opamswitch=false) switch =
 let shells_list = [ SH_sh; SH_zsh; SH_csh; SH_fish ]
 
 let complete_file = function
-  | SH_sh -> Some "complete.sh"
+  | SH_sh | SH_bash -> Some "complete.sh"
   | SH_zsh -> Some "complete.zsh"
-  | _ -> None
+  | SH_csh | SH_fish -> None
 
 let env_hook_file = function
   | SH_sh | SH_bash -> Some "env_hook.sh"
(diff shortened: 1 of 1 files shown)


src/state/opamEnv.ml
```diff
@@ -391,7 +391,7 @@ let init_file = function
   | SH_fish -> "init.fish"
 
 let complete_script = function
-  | SH_bash -> Some OpamScript.complete
+  | SH_sh -> Some OpamScript.complete
```
dra27: Shouldn't this be `| SH_bash | SH_sh -> ... ?

I wonder if something is going wrong with shell detection, rather than that this function needs fixing?

dra27: Yes, having looked further - I think it's more important to investigate why `OpamStd.Sys.guess_shell_compat` is returning `SH_sh` on your system rather than `SH_bash`. Are you able to debug that yourself, or would you like some pointers?

anemator: Shell detection seems to work, I checked on MacOS 10.13.4 and Ubuntu 18.04, so I don't think it's that. From my understanding of the code, we're just generating scripts for every shell on init, not just the one in use. Note:  bash and sh are the same entry in shells_list:
https://github.com/ocaml/opam/blob/4f2753428d11ca021b3cb684559e8fa2f119c686/src/state/opamEnv.ml#L367-L369

The source of the problem is the following code:
https://github.com/ocaml/opam/blob/4f2753428d11ca021b3cb684559e8fa2f119c686/src/state/opamEnv.ml#L517-L523

When we loop through ```shells_list```, the ```filef``` and ```scriptf``` functions have to match on the same shell variant, but ```complete_file``` ignores ```SH_bash``` and ```complete_script``` ignores ```SH_sh```. The ```complete.sh``` script itself is set up to exit early if the shell is not bash:
https://github.com/ocaml/opam/blob/4f2753428d11ca021b3cb684559e8fa2f119c686/src/state/shellscripts/complete.sh#L1

dra27: Oh, indeed, how silly of me! In which case, I wonder if just for completeness (excuse pun) sake it should be `| SH_bash | SH_sh ->` (cf. the others).

AltGr:
Thanks!

Issue #3362, opened 2018-05-22
opam switch create on NetBSD


UnixJunkie:
Hello,

With opam 2.0.0~rc:
```
$ opam switch create 4.06.1
[...]
#=== ERROR while compiling ocaml-base-compiler.4.06.1 =========================#
"./configure": command not found.
```

dra27:
Could post the content of the ellipsis?

UnixJunkie:
https://gist.github.com/UnixJunkie/191649441b3227b501f150393abc5a94

dra27:
This is definitely with the original rc, not rc2?

If you run with `OPAMKEEPBUILDDIR=yes`, the compiler source tree should be left behind on the error (assuming you don’t tell opam to delete the switch, obviously) - is `configure` executable?

UnixJunkie:
yes, not yet using rc2.

dra27:
Are you able to try the second part?

UnixJunkie:
2.0.0-rc2 doesn't compile in there:
```
'~/bin/ocamlc.opt' -g -w -40 -o boot.exe unix.cma boot.ml
File "boot.ml", line 1:
Error: Error on dynamically loaded library: /usr/pkg/lib/ocaml/stublibs/dllunix.so: /usr/pkg/lib/ocaml/stublibs/dllunix.so: Undefined PLT symbol "caml_strdup" (symnum = 133)
Makefile:27: recipe for target 'src_ext/jbuilder/_build/install/default/bin/jbuilder' failed
gmake: *** [src_ext/jbuilder/_build/install/default/bin/jbuilder] Error 2
```

UnixJunkie:
I think
```
OPAMKEEPBUILDDIR=yes
```
is ignored by 2.0.0-rc

rjbou:
If you answer no to the last question 
```Switch initialisation failed, clean up? ('n' will leave the switch partially installed) [Y/n]```
is file `~/.opam/4.06.1/.opam-switch/{build,sources}/ocaml-base-compiler.4.06.1/configure` exists and is an executable?

AltGr:
> 2.0.0-rc2 doesn't compile in there

hmm, this one seems to have to do with the bootstrap of the (bundled) jbuilder... will have to ask the devs.
[...]

PR #3378, opened 2018-05-28
Add tar handling: ignoring pax_global_header file

rjbou:
This PR closes # #3362.

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -229,7 +229,7 @@ let init =
            adding a line to the `~/.profile' file.";
         Some false, info ["n";"no-setup"] ~doc:
           "Do not update the user shell configuration to setup opam. Also \
-           implies $(--disable-shell-hook), unless $(--interactive) or \
+           implies $(b,--disable-shell-hook), unless $(b,--interactive) or \
            specified otherwise";
       ])
   in
--- a/src/core/opamSystem.ml
+++ b/src/core/opamSystem.ml
@@ -647,7 +647,12 @@ let extract_job ~dir file =
     else if try not (Sys.is_directory dir) with Sys_error _ -> false then
       internal_error "Extracting the archive would overwrite %s." dir
     else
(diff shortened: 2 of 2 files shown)


src/core/opamSystem.ml
```diff
@@ -647,7 +647,12 @@ let extract_job ~dir file =
     else if try not (Sys.is_directory dir) with Sys_error _ -> false then
       internal_error "Extracting the archive would overwrite %s." dir
     else
-    match files_all_not_dir tmp_dir with
+    let flist =
+      OpamStd.Op.(
+        files_all_not_dir tmp_dir |>
+        List.filter (not @* OpamStd.String.contains ~sub:"pax_global_header"))
+    in
+    match flist with
```
AltGr: From my understanding, we not only need to remove the `pax_global_header` dir, but also move the contents of the other directory beside it to the root ?

src/client/opamCommands.ml
```diff
@@ -330,23 +330,19 @@ let init =
...
       if reinit then
-        let init_config =
-          get_init_config ~no_sandboxing
-            ~no_default_config_file:no_config_file ~add_config_file:config_file
-        in
         OpamClient.reinit ~init_config ~interactive ~dot_profile
           ?update_config ?env_hook ?completion
           (OpamFile.Config.safe_read config_f) shell
       else
         OpamEnv.setup root
```
AltGr: This wasn't really duplicate: we don't need to call it in this case, where it is not needed (because it has some side-effects like printed messages and possibly errors)

rjbou: ok!

src/core/opamSystem.ml
rjbou: To have the same behavior no need to move the contents. With `tar` <1.14, there is a just an `pax_global_header` extra file at archive extract root, which confuses opam because there is different handling of copy for single-directory and several-file archive content.

PR #3381, opened 2018-05-29
improve the formatting of docstrings with odoc master

avsm:
When generating the docs with `dune build @doc` and odoc master,
the new parser emits some warnings. This fixes some of the markup
to be more compliant.


AltGr:
Thanks!

PR #3392, opened 2018-06-05
Provide 'opam switch link SWITCH', some linked switches fixes, etc.

AltGr:


Drup:
Does this means it should be possible to get a list of all the switches on the system, including the local ones?

AltGr:
That was already the case, except that opam can't assume it knows about them all. The info is useful for knowing what switch a package is installed in with `opam show`, for example.
I've just improved the synchronisation so that local switches don't only get registered at `opam switch create` time.

Drup:
Ok, nice. :)

Issue #3380, opened 2018-05-28
Feature proposal: `opam install --justinstall`


AltGr:
There is currently still something that feels like a hole in our workflows, when working in an opam-aware project: these two are very good
- use `opam install . --deps-only` to have opam handle the local source dependencies
- use `make` to compile by hand

These can be enough, but sometimes, you need to also test the _installation_ of your package by hand, and things get a bit worse:
- either run `opam install .`, but if you do it repeatedly, there will be no incremental compilation, so this can get quite long
- or run `make install` (possibly calling `opam-installer` if you use `.install` files), to use the results of your `make`. But then the resulting installation gets into your opam prefix, without being registered by opam, so this is a bit unclean (I admit I am using this a lot, though).

The option `--reuse-build-dir` and `--inplace-build` have somehow been attempts at helping here, but are not really satisfactory to that purpose: they are a bit heavy to use, bring some quirks and don't always behave as expected (e.g. `--inplace-build` will re-run `./configure` with its own options from the `opam` file in your project dir, which may get annoying).

---

Enough for the summary of the issue, here is an idea to fill that gap:
let the user run the builds with their own tools, but let opam handle just the install part, from the source directory, and assuming the build was done properly. Could be something like:
```
opam install --justinstall foo
```
assuming `foo` is a locally-pinned package, or a directory containing one or more opam packages, run their installation instructions, including handling their `.install` and `.config` files if any, and register their installation into opam (then possibly recompiling dependents if any). Another plus is that uninstallation can then be cleanly handled by opam.

This could typically be used just after `make`, or directly from a Makefile `install` rule. Implementation-wise, care should be taken that the command is not allowed to modify any dependency of `foo`.

I'd like to have the input from the users here!

Another idea could be to just extend `opam-installer` (or a related command built-in opam) so that it registers the installation of the package within opam. It would certainly be simpler, but wouldn't handle installation outside of `.install` files, consistency with the package definition, or recompilation of dependents.

Drup:
That sounds quite useful. I think it could also solve many of @gasche's woes regarding compiler development with opam.

gasche:
The feature you propose sounds very nice. I have no strong opinion over whether it should be in `opam install` or `opam-installer` -- having it in the `opam` binary makes the setup slightly simpler, but well.

[...]

PR #3394, opened 2018-06-06
Add `--assume-built` option

rjbou:
This PR is related to this issue #3380.
If a package `foo` is locally pinned, option `--assume-built` can be set to skip build steps and perform only install commands. 
If a package is not locally pinned and the option is set, it will be skipped.
Packages dependencies are *locked* on their current version: if a dependency `bar` of the package `foo` needs an upgrade, and a `foo` install with `--assume-built` is launched, opam won't propose to upgrade the dependency  `bar` and reinstall `foo`. 

EDIT:
- Add the option to `reinstall`
- Always reinstall `foo` with the option (even if no changes detected)
- if a package `baz` depends on `foo`, on `foo` reinstall `baz` will be recompiled/reinstalled (even if non-locally pinned).

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -1197,15 +1197,20 @@ let remove t ~autoremove ~force names =
   let atoms = OpamSolution.sanitize_atom_list t names in
   remove_t ~autoremove ~force atoms t
 
-let reinstall_t t ?ask ?(force=false) atoms =
+let reinstall_t t ?ask ?(force=false) ~assume_built atoms =
   log "reinstall %a" (slog OpamFormula.string_of_atoms) atoms;
+
+  let atoms =
+    if assume_built then filter_unpinned_locally t atoms (fun x -> x)
+    else atoms
+  in
 
   let reinstall, not_installed =
     get_installed_atoms t atoms in
   let to_install =
     if not_installed <> [] then
       if
-        force ||
+        force || assume_built ||
         OpamConsole.confirm "%s %s not installed. Install %s?"
           (OpamStd.Format.pretty_list
              (List.rev_map OpamFormula.short_string_of_atom not_installed))
@@ -1226,6 +1231,12 @@ let reinstall_t t ?ask ?(force=false) atoms =
   let requested =
     OpamPackage.Name.Set.of_list (List.rev_map fst atoms) in
 
+  let t, atoms =
+    if assume_built then
+      assume_built_restrictions t atoms
+    else t, atoms
+  in
+
   let request = OpamSolver.request ~install:atoms ~criteria:`Fixup () in
 
   let t, solution =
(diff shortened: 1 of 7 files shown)


src/client/opamCommands.ml
```diff
@@ -1180,9 +1180,14 @@ let install =
        definitions, and is typically useful to provide more constrained \
        dependencies that describe a precise development environment."
   in
+  let assume_built =
+    Arg.(value & flag & info ["assume-built"]
+           ~doc:"Just install the package from its pinned directory, but don't \
+                 actually run build.")
+  in
```
AltGr: We should give more explanation and examples of use here: «For use on locally-pinned packages: assume they have already been correctly built, and only run their installation instructions, directly from their source directory. This skips the build instructions and can be useful to install packages that are being worked on. Implies $(i,--inplace-build).»

src/client/opamClient.ml
```diff
@@ -1001,7 +1018,26 @@ let install_t t ?ask atoms add_to_roots ~deps_only =
...
+          t.packages
+      in
+      let universe =
+        OpamSwitchState.universe t
+          ~requested:(OpamPackage.names_of_packages pkg_of_atoms)
+          Install
+      in
+      let fixed =
+        OpamPackage.Set.of_list
+          (OpamSolver.dependencies ~build:true ~post:true
```
AltGr: Shoudln't `post` be `false` here ? We _could_ also not consider `build` deps, since they don't trigger recompilations, but I'd rather have them, seems simpler.

src/client/opamClient.ml
```diff
@@ -1001,7 +1018,26 @@ let install_t t ?ask atoms add_to_roots ~deps_only =
...
+             ~depopts:true ~installed:true universe pkg_of_atoms)
+      in
+      atoms @ (OpamSolution.eq_atoms_of_packages fixed)
```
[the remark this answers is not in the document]
rjbou: There is no further handling of conflict messages, in all cases we have the usual 
```
  - foo → bar >= 3
      no matching version
```

src/client/opamClient.ml
```diff
@@ -1028,19 +1064,21 @@ let install_t t ?ask atoms add_to_roots ~deps_only =
...
+let install t ?autoupdate ?add_to_roots
+    ?(deps_only=false) ?(assume_built=false) names =
   let atoms = OpamSolution.sanitize_atom_list ~permissive:true t names in
   let autoupdate_atoms = match autoupdate with
     | None -> atoms
     | Some a -> OpamSolution.sanitize_atom_list ~permissive:true t a
   in
   let t = update_dev_packages_t autoupdate_atoms t in
-  install_t t atoms add_to_roots ~deps_only
+  install_t t atoms add_to_roots ~deps_only ~assume_built
```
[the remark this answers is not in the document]
rjbou: done in the PR new version

src/client/opamCommands.ml
```diff
@@ -1324,6 +1333,15 @@ let reinstall =
...
   ] in
+  let assume_built =
+    Arg.(value & flag & info ["assume-built"]
+           ~doc:"For use on locally-pinned packages: assume they have already \
+                 been correctly built, and only run their installation \
+                 instructions, directly from their source directory. This \
+                 skips the build instructions and can be useful to install \
+                 packages that are being worked on. Implies $(i,--inplace-build). \
+                 No locally-pinned packages will be skipped.")
+  in
```
rjbou: Not sur about that, duplicated. Maybe a message redirecting to `opam install` manpage and its `--asume-built` option?

rjbou:
I didn't squash some commits, intentionnally. Will do it on last version, before merging.

src/client/opamCommands.ml
AltGr: Can be added as a top-level value in `OpamArgs` and used directly in the `Term.(pure foo $x $y $OpamArgs.assume_built)`

PR #3397, opened 2018-06-07
Download error message

### LINKED ISSUE #3393: Have a more specific error message when checksum of an archive fails
Checksum errors seem to be quite a common problem with, for instance, github or others changing how archives are generated and thus changing their checksums.

Right now there seems to be no specific error message for checksum failures and we only get the same error message as when the archive couldn't be reached (i.e. `[ERROR] The sources of the following couldn't be obtained, aborting:`)

Could we have a specific error message for checksum failures ?

rjbou:
This PR closes #3393.
It adds two levels of download error message: a short one to be displayed on normal mode and a more complete one on verbose mode (when it is needed, when the message itself is short, it is the same on both modes). For the moment, no specific error download message is displayed to user, it is only filled on json file output when option is set.

### DIFF
--- a/src/client/opamAction.mli
+++ b/src/client/opamAction.mli
@@ -15,12 +15,13 @@ open OpamTypes
 open OpamStateTypes
 
 (** [download t pkg] downloads the source of the package [pkg] into its locally
-    cached source dir. Returns [Some errmsg] on error, [None] on success.
+    cached source dir. Returns [Some (short_errmsg option, long_errmsg)] on error,
+    [None] on success. See {!OpamTypes.Not_available}.
 
     This doesn't update dev packages that already have a locally cached
     source. *)
 val download_package:
-  rw switch_state -> package -> string option OpamProcess.job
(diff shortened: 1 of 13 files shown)


src/client/opamAction.mli
```diff
@@ -20,7 +20,7 @@ open OpamStateTypes
     This doesn't update dev packages that already have a locally cached
     source. *)
 val download_package:
```
AltGr: Can the returned type be documented ?

src/format/opamTypes.mli
```diff
@@ -49,7 +49,7 @@ type std_path =
 (** Download result *)
 type 'a download =
   | Up_to_date of 'a
-  | Not_available of string
+  | Not_available of string option * string
```
AltGr: This should be documented too

rjbou: fixed

src/client/opamAction.mli
rjbou: fixed

PR #3399, opened 2018-06-08
if 'ccache' is installed, add its cache directory to the sandbox (rw)

### LINKED ISSUE #3395: sandboxing breaks ccache
I'm now using opam2~rc2. On my Fedora system, running

```
opam switch create . 4.06.1
```

fails with the following error

```
#=== ERROR while compiling ocaml-base-compiler.4.06.1 =========================#
# context     2.0.0~rc | linux/x86_64 |  | https://opam.ocaml.org/2.0#bed714e6
# path        ~/tmp/repro/_opam/.opam-switch/build/ocaml-base-compiler.4.06.1
# command     ~/.opam/opam-init/hooks/sandbox.sh build make world
# exit-code   2
# env-file    ~/.opam/log/ocaml-base-compiler-22947-1872ec.env
# output-file ~/.opam/log/ocaml-base-compiler-22947-1872ec.out
### output ###
# make -C byterun  all
# make[1]: Entering directory '/home/gasche/tmp/repro/_opam/.opam-switch/build/ocaml-base-compiler.4.06.1/byterun'
# echo "/home/gasche/tmp/repro/_opam/lib/ocaml/stublibs" > ld.conf
# echo "/home/gasche/tmp/repro/_opam/lib/ocaml" >> ld.conf
# sed -n -e '/^  /s/ \([A-Z]\)/ \&\&lbl_\1/gp' \
#        -e '/^}/q' caml/instruct.h > caml/jumptbl.h
# gcc -c -O2 -fno-strict-aliasing -fwrapv -Wall -fno-tree-vrp  -g -D_FILE_OFFSET_BITS=64 -D_REENTRANT -DCAML_NAME_SPACE  interp.c
# ccache: error: Failed to create temporary file for /home/gasche/.ccache/tmp/tmp.cpp_stderr: Read-only file system
# make[1]: *** [Makefile:205: interp.o] Error 1
# make[1]: Leaving directory '/home/gasche/tmp/repro/_opam/.opam-switch/build/ocaml-base-compiler.4.06.1/byterun'
# make: *** [Makefile:384: coldstart] Error 2
```

It turns out (I had no idea) that `ccache` is enabled by default on Fedora OSes. Could you add `$HOME/.ccache` to the list of directories to the write whitelist of the sandbox?

For reference (if other users encounter the issue), a local fix is to pass `CCACHE_DISABLE=1` explicitly to the command-line:

```
CCACHE_DISABLE=1 opam switch create . 4.06.1
```

of course, many other OPAM packages (than the compiler) invoke `gcc`, so selectively disabling ccache all the time will get old fast -- then export it in your shell environment, etc.

gasche:
fixes #3395 

cc @AltGr, @avsm 

I have tested the patchset on my machine (Linux, ccache installed). I have not tested it on macos (so the `sandbox_exec.sh` change is untested), or on a machine without ccache.

@AltGr is there a way to factorize common code between the two files?

This implementation ignores the CCACHE_DISABLED setting: it will add the cache directory (rw) even if ccache is available but disabled in the system. I could take this into account (there is a line of the `--print-config` output that tells whether ccache is currently disabled), but I thought it wasn't worth the trouble and extra script complexity.

### DIFF
--- a/src/state/shellscripts/sandbox_exec.sh
+++ b/src/state/shellscripts/sandbox_exec.sh
@@ -1,4 +1,4 @@
-#!/bin/sh -ue
+#!/bin/bash -ue
 
 POL='(version 1)(allow default)(deny network*)(deny file-write*)'
 POL="$POL"'(allow file-write* (literal "/dev/null"))'
--- a/src/state/shellscripts/sandbox_exec.sh
+++ b/src/state/shellscripts/sandbox_exec.sh
@@ -13,12 +13,29 @@ add_mounts() {
 
 add_mounts rw "${TMPDIR:-/tmp}"
 
+# C compilers using `ccache` will write to a shared cache directory
+# that remain writeable. ccache seems widespread in some Fedora systems.
+add_ccache_mount() {
+  if command -v ccache > /dev/null; then
+      CCACHE_DIR=$HOME/.ccache
+      ccache_dir_regex='cache_dir = (.*)$'
+      local IFS=$'\n'
+      for f in $(ccache --print-config); do
+        if [[ $f =~ $ccache_dir_regex ]]; then
+          CCACHE_DIR=${BASH_REMATCH[1]}
+        fi
+      done
+      add_mounts rw $CCACHE_DIR
+  fi
+}
+
 # This case-switch should remain identical between the different sandbox implems
 COMMAND="$1"; shift
 case "$COMMAND" in
     build)
         add_mounts ro "$OPAM_SWITCH_PREFIX"
         add_mounts rw "$PWD"
+        add_ccache_mount
         ;;
     install)
         add_mounts rw "$OPAM_SWITCH_PREFIX"
--- a/src/state/shellscripts/bwrap.sh
+++ b/src/state/shellscripts/bwrap.sh
@@ -25,12 +25,29 @@ add_mounts() {
 
 add_mounts ro /usr /bin /lib /lib32 /lib64 /etc /opt /nix/store /home
 
+# C compilers using `ccache` will write to a shared cache directory
+# that remain writeable. ccache seems widespread in some Fedora systems.
+add_ccache_mount() {
+  if command -v ccache > /dev/null; then
+      CCACHE_DIR=$HOME/.ccache
+      ccache_dir_regex='cache_dir = (.*)$'
+      local IFS=$'\n'
+      for f in $(ccache --print-config); do
+        if [[ $f =~ $ccache_dir_regex ]]; then
+          CCACHE_DIR=${BASH_REMATCH[1]}
+        fi
+      done
+      add_mounts rw $CCACHE_DIR
+  fi
+}
+
 # This case-switch should remain identical between the different sandbox implems
 COMMAND="$1"; shift
 case "$COMMAND" in
     build)
         add_mounts ro "$OPAM_SWITCH_PREFIX"
         add_mounts rw "$PWD"
+        add_ccache_mount
         ;;
     install)
         add_mounts rw "$OPAM_SWITCH_PREFIX"


gasche:
The first PR version would output the result of `ccache --help` to the standard output, which is unpleasant in the logs. I'll fix it now and rebase.

gasche:
(It's just a debugging `echo` that I forgot to remove, sigh...)

src/state/shellscripts/sandbox_exec.sh
```diff
@@ -13,12 +13,29 @@ add_mounts() {
...
+# that remain writeable. ccache seems widespread in some Fedora systems.
+add_ccache_mount() {
+  if command -v ccache > /dev/null; then
+      CCACHE_DIR=$HOME/.ccache
+      ccache_dir_regex='cache_dir = (.*)$'
+      local IFS=$'\n'
+      for f in $(ccache --print-config); do
+        if [[ $f =~ $ccache_dir_regex ]]; then
+          CCACHE_DIR=${BASH_REMATCH[1]}
+        fi
```
AltGr: The `bwrap` script is bash-only, but this one is declared as `sh`; is this still OK ?

AltGr:
Thanks a lot!

src/state/shellscripts/sandbox_exec.sh
gasche: Oh, I hadn't thought of that. If you are not willing to depend on `bash` on OSX, maybe we could just give up on `ccache` support in OSX for the time being? I'm not very excited at trying to figure out the right grep+sed magic -- although I suppose that `ccache --print-config | grep -o 'cache_dir = .*$' | sed 's/cache_dir //g'` should work, I cannot test it.

AltGr: Maybe that's the safest, let's ask an OSX user (@avsm ?) about it ?

avsm: I'm a little confused: bwrap is Linux-only, where you can (more-or-less) depend on bash being available.  The bwrap script should never run on OSX should it?

Either way, bash is the default on osx, but an ancient version (3.2.57), so you need to be careful not to use bash4-isms. The portability danger is on OpenBSD, where `/bin/sh` is `ksh` and not `bash` or `dash`.

AltGr: it's actually the `sandbox_exec` script that we are discussing here (and which was not bash-only until now) ; but I managed to make the `bwrap` and `sandbox_exec` scripts quite similar, at least for their interesting parts.

@gasche: OK to merge then, just make the `#!` bash for `sandbox_exec` and we're good. Thanks!

gasche:
I made `sandbox_exec.sh` use bash in the shebang and rebased.

Note: I also sent a PR to `ccache` to have easier access to the value of configuration variables and avoid the grepping, https://github.com/ccache/ccache/pull/270 . We won't be able to use this in a while, though :-)

src/state/shellscripts/sandbox_exec.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/sh -ue
+#!/bin/bash -ue
```
hannesm: FWIW, I think the posix[tm] way to do this is `#!/usr/bin/env bash` -- this may even lead to success on non-GNU/Linux systems that have bash installed optionally.

gasche: Thanks, I proposed this change in #3405.

AltGr: Discussing this in the other PRs; in this case it's OSX, though, which should always have `bash` in `/bin` ?

PR #3404, opened 2018-06-11
Fixes: Env var & delimiters, unversionned opam file pin, doc

### LINKED ISSUE #3391: opam breaks man

When no MANPATH is set, running the opam init script sets the MANPATH in a way that it is broken. This breaks every man page.

```
$ echo $MANPATH
/home/r/.opam/system/man:/home/r/.opam/system/man
$ man git
No manual entry for git
See 'man 7 undocumented' for help when manual pages are not available.
$ MANPATH= man git
# <man page appears>
```

```
# opam config report
# opam-version      2.0.0~rc # but this is really rc2
# self-upgrade      no
# system            arch=x86_64 os=linux os-distribution=debian os-version=testing
# jobs              8
# repositories      4 (http), 5 (local), 1 (version-controlled) (default repo at 61536b7f)
# pinned            1 (rsync)
# current-switch    system
```

### LINKED ISSUE #3403: Minimal opam file incompatibility on v1.2 and v2 when pinning
I have the following minimal `todolist.opam` file:

```
opam-version: "1.2"
name: "todolist"
version: "1.0.0"
maintainer: "Your Name <email@example.com>"

depends: [
  "jbuilder" {build}
  "caqti"
  "caqti-lwt"
  "caqti-driver-postgresql"
]
```

I'm having trouble pinning this package with opam 2.0.0~rc2.

With opam.1.2.2 (working as expected, installing the dependencies):

```
$ opam pin add todolist .
Package todolist does not exist, create as a NEW package ? [Y/n] y
todolist is now path-pinned to /Users/bobbypriambodo/Projects/todolist

[todolist] /Users/bobbypriambodo/Projects/todolist/ synchronized
[todolist] Installing new package description from /Users/bobbypriambodo/Projects/todolist

todolist needs to be installed.
  ∗  install conf-m4                 1           [required by ocamlfind]
  ∗  install jbuilder                1.0+beta20  [required by todolist]
  ...
```

With opam.2.0.0~rc2 (not working):

```
$ opam pin add todolist .
Package todolist does not exist, create as a NEW package? [Y/n] y
[todolist.~dev] synchronised from git+file:///Users/bobbypriambodo/Projects/todolist#master
[NOTE] No package definition found for todolist.~dev: please complete the template
Press enter to start "vim" (this can be customised by setting EDITOR or OPAMEDITOR)...
```

When I press enter, an editor opened with a newly created opam file template which I don't want. Changing `opam-version` to `"2.0"` gives the same outcome. From the [Packaging](https://opam.ocaml.org/doc/2.0/Packaging.html) doc, the necessary fields should only  be `opam-version` and `maintainer`. Do you know where I did wrong?

```
# opam config report
# opam-version      2.0.0~rc2
# self-upgrade      no
# system            arch=x86_64 os=macos os-distribution=homebrew os-version=10.12.6
# jobs              3
# repositories      1 (http) (default repo at 3faa926b)
# pinned            3 (git)
# current-switch    default
```

rjbou:
This PR closes #3391.
- It fixes `OpamStd.split_delim` implementation which had the same behaviour than `OpamStd.split`. Now they respectively keep & clean empty strings from the returned list.
- An option is added to `OpamStd.Sys.split_path_variable`, permiiting to retrieve a cleaned path or no.
- A fix on environment variables construction, in order to keep the trailing delimiter when it should be.

edit:
- Fix #3403: on local pin, versionned directory, if opam file is not versionned retrieve the one found locally in source directory
- ~Fix # 3398: Escape json extended ascii characters~
- Update pinning doc

### DIFF
--- a/src/client/opamPinCommand.ml
+++ b/src/client/opamPinCommand.ml
@@ -441,6 +441,7 @@ and source_pin
     OpamPath.Switch.Overlay.tmp_opam st.switch_global.root st.switch name
   in
 
+  let opam_local = OpamFile.OPAM.read_opt temp_file in
   OpamFilename.remove (OpamFile.filename temp_file);
 
   let opam_opt =
@@ -472,6 +473,20 @@ and source_pin
       opam_opt >>+ fun () ->
       OpamPackage.Map.find_opt nv st.installed_opams >>+ fun () ->
       OpamSwitchState.opam_opt st nv)
+  in
+
+  let opam_opt =
+    match opam_local, opam_opt with
+    | Some local, None ->
+      OpamConsole.warning
+        "Couldn't retrieve opam file from versioned source, \
+         using the one found locally.";
+      Some local
+    | Some local, Some vers when not (OpamFile.OPAM.effectively_equal local vers) ->
+      OpamConsole.warning
+        "Opam file have significant uncommitted changes, using the versioned one";
+      opam_opt
+    | _ -> opam_opt
   in
 
   if not need_edit && opam_opt = None then
--- a/doc/pages/Manual.md
+++ b/doc/pages/Manual.md
@@ -187,6 +187,12 @@ pinning a package, the source is searched for metadata in an `opam` or
 directory. You can also replace that file by a directory containing an `opam`
 file and optionally other metadata, like a `files/` subdirectory.
 
+As the `package` argument is optional, `opam` guesses package name from the
+`<URL>` or the `opam` file found. Note that for local VCS pinning, when given
+without package name, `opam` retrieves the locally found `opam` file, even if not
+versioned. If this file is versioned, `opam` relies on the versioned
+version.
+
 Whenever an install, reinstall or upgrade command-line refers to a pinned
 package, <span class="opam">opam</span> first fetches its latest source. `opam
 update [--development]` is otherwise the standard way to update the sources of
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -2350,9 +2350,14 @@ let pin ?(unpin_only=false) () =
   let guess_names url k =
     let from_opam_files dir =
       OpamStd.List.filter_map
-        (fun (nameopt, f) -> match nameopt with
-           | None -> OpamFile.OPAM.(name_opt (safe_read f))
-           | some -> some)
+        (fun (nameopt, f) ->
+           let opam_opt = OpamFile.OPAM.read_opt f in
+           let name =
+             match nameopt with
+             | None -> OpamStd.Option.replace OpamFile.OPAM.name_opt opam_opt
+             | some -> some
+           in
+           OpamStd.Option.map (fun n -> n, opam_opt) name)
         (OpamPinned.files_in_source dir)
     in
     let basename =
@@ -2390,7 +2395,7 @@ let pin ?(unpin_only=false) () =
       match found with
       | _::_ -> found
       | [] ->
-        try [OpamPackage.Name.of_string basename] with
+        try [OpamPackage.Name.of_string basename, None] with
         | Failure _ ->
           OpamConsole.error_and_exit `Bad_arguments
             "Could not infer a package name from %s, please specify it on the \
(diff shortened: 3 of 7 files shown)


src/client/opamPinCommand.ml
```diff
@@ -474,6 +475,16 @@ and source_pin
       OpamSwitchState.opam_opt st nv)
   in
 
+  let opam_opt =
+    match opam_local, opam_opt with
+    | Some local, None ->
+      OpamConsole.warning
+        "Couldn't retrieve opam file from versionned source, \
+         using the one found locally.";
+      Some local
```
AltGr: I am wondering if the case when both are defined and different (`Some local, Some vers when local <> vers`; or maybe `when not (OpamFile.OPAM.effectively_equal local vers)`) would not deserve a warning too: "not using uncommitted version of opam file xxx".

But there is already the warning about uncommitted changes, so that might be redundant.

src/core/opamJson.ml
```diff
@@ -27,7 +27,8 @@ let adds_esc b s =
       match String.get s i with
       | '"' -> flush b start i; adds b "\\\""; loop next next
       | '\\' -> flush b start i; adds b "\\\\"; loop next next
-      | '\x00' .. '\x1F' | '\x7F' (* US-ASCII control chars *) as c ->
+      | '\x00' .. '\x1F' | '\x7F' (* US-ASCII control chars *)
+      | '\x80' .. '\xFF' (* Extended ASCII chars *) as c ->
```
dbuenzli: You can't do this here, this will break UTF-8 output. The module assumes strings are UTF-8 encoded. 

Btw. could you please make one PR per issue you fix.

dbuenzli:
Also on #3403 I wonder if there's really anything to fix. For me it works as expected, a `git` pin considers only what is in git.

AltGr:
@dbuenzli there was still a consistency problem, because then we shouldn't use the non-versionned `opam` files for package detection (which would actually be more involved). It was a bit weird to detect the package definition file for pinning, but then say that no definition was found.

dbuenzli:
> @dbuenzli there was still a consistency problem, because then we shouldn't use the non-versionned opam files for package detection (which would actually be more involved).

Then why do you use the non-versionned `opam`  file at all ? Just consider only what is in the VCS.

With this the system becomes idiosyncratic to understand. Now if I have an `opam` VCS file with uncommited changes why won't `opam` take it into account while it did take into account things that were not in the VCS before ? Keep these semantics simple.

src/client/opamPinCommand.ml
rjbou: @AltGr It would be complementary: on update, uncommitted changes warning is shown, in pinning, uncommitted opam changes warning is shown.

dbuenzli:
> Then why do you use the non-versionned opam file at all ? Just say no opam file was found, only consider what is in the VCS.

See also https://github.com/ocaml/opam/issues/3408

I really think the way this is done too complicated and should be carefully reviewed (better, *documented* so that one can see if it can explains the behaviour in simple words and can then inform the design).

I think the confusion by @bobbypriambodo in #3403 is a confusion due to a change in defaults (and it's right to be confused by a change in defaults), I don't think it's a confusion in expectations.

bobbypriambodo:
> I think the confusion by @bobbypriambodo in #3403 is a confusion due to a change in defaults

I confirm this, it did throw me off because I remembered opam v1.2 worked that way. However I do agree with @dbuenzli that the behavior is consistent and reasonable. Perhaps it only warrants a documentation and/or a warning if opam cannot find an `opam` file? Something like `no package definition is found -- perhaps you forgot to commit the opam file?` (definitely could be better).

rjbou:
> I really think the way this is done too complicated and should be carefully reviewed (better, documented so that one can see if one can explains the behaviour in simple words, which will inform the design).

Indeed, it should be added in the new documentation. For the UX part, user is alerted by messages when uncommitted changes are discarded on upgrade, when an unversioned opam file is taken on pinning, when uncommitted opam file changes are discarded, etc.

PR #3405, opened 2018-06-12
Posix shebangs

gasche:
This PR, on top of #3399 (which switched sanbbox_exec.sh to `bash` instead of `sh`), use POSIX-friendly shebangs `/usr/bin/env bash -ue` instead of `/bin/bash -ue` as suggested by @hannesm. I don't know enough about the porability issues underhand to decide whether that's a good move, but the PR is here for consideration.

### DIFF
--- a/src/state/shellscripts/sandbox_exec.sh
+++ b/src/state/shellscripts/sandbox_exec.sh
@@ -1,4 +1,5 @@
-#!/bin/bash -ue
+#!/usr/bin/env bash
+set -ue
 
 POL='(version 1)(allow default)(deny network*)(deny file-write*)'
 POL="$POL"'(allow file-write* (literal "/dev/null"))'


hannesm:
thanks @gasche.  to give some motivation: most BSD systems do not contain a `/bin/bash`, but `bash` is a package installed as `/usr/local/bin/bash`.  to develop portable scripts, you can either stick to `sh` (see http://pubs.opengroup.org/onlinepubs/9699919799/utilities/sh.html), or use a path-independent way of calling out to `bash` -- as done by @gasche in here.  I only now realise that these sandbox scripts are meant to work on Linux only, so it's slightly overengineered...

avsm:
I think the Linux-only scripts should probably stay as `/bin/bash` -- the `/usr/bin/env` just complicates things with multiple arguments unnecessarily.

gasche:
Thanks for the feedback, I rebased the PR to only affect sandbox_exec.sh.

AltGr:
Isn't `#!/usr/bin/env bash -ue` incorrect? I believe arguments don't work there and you should write
```
#!/usr/bin/env bash
set -ue
```
On my (Debian) system it doesn't ignore the option but gives `/usr/bin/env: 'bash -ue': No such file or directory`

AltGr:
Thanks!
`sandbox_exec` is OSX only for the time being, though: are there cases where calling `env` would be needed there ?

gasche:
I fixed the parameters (I had tried on my machine where it worked fine) with a separate call to `set`. I don't know anything about OSX, so if you believe we could just stay with `/bin/bash` we can also close this PR.

AltGr:
Thanks!

PR #3406, opened 2018-06-12
use #!/usr/bin/env bash instead of #!/bin/bash

hannesm:
following along #3405, I `git grep /bin/bash` in this repository and used `/usr/bin/env bash` instead. /cc @dra27 does this work on windows (see `msvs-detect`, `check_linker`) as well?

In `opam_installer.ml` I couldn't find any use of `bash` specifics when outputting the shell script, thus instead I made it to output `#!/bin/sh` instead of `#!/bin/bash`.


shell/bundle.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -ue
+#!/usr/bin/env bash -ue
```
avsm: Be really careful with multiple arguments here -- this may reduce portability on soe systems; https://stackoverflow.com/questions/4303128/how-to-use-multiple-arguments-for-awk-with-a-shebang-i-e/4304187#4304187

AltGr:
Most of these scripts are used for maintainance of the opam source itself, and not within an opam installation, so this is non-critical (although always nice to have better portability without requiring manual fiddling for calling the shell, of course!)

Good point on the `opam-installer` gen-script feature though!
Just be careful with arguments in the shebang, though, as Anil pointed out (you need to use e.g. `set -ue` within the script for these cases)

hannesm:
i added a commit, turns out more shell scripts only used sh (checked with https://www.shellcheck.net)

AltGr:
Thanks!

PR #3407, opened 2018-06-14
Make hash function LF/CRLF-agnostic

dra27:
Splits off the completed commit from #3349

### DIFF
--- a/src/core/opamHash.ml
+++ b/src/core/opamHash.ml
@@ -82,31 +82,93 @@ let to_json s = `String (to_string s)
 let to_path (kind,s) =
   [string_of_kind kind; String.sub s 0 2; s]
 
-let compute ?(kind=default_kind) file = match kind with
-  | `MD5 -> md5 (Digest.to_hex (Digest.file file))
-  | (`SHA256 | `SHA512) as kind ->
-    try
-      if not OpamCoreConfig.(!r.use_openssl) then raise Exit else
-      match
-        OpamSystem.read_command_output ["openssl"; string_of_kind kind; file]
-      with
-      | [l] ->
-        let len = len kind in
-        make kind (String.sub l (String.length l - len) len)
-      | _ -> failwith "openssl error"
-    with OpamSystem.Command_not_found _ | Exit ->
-      make kind (OpamSHA.hash kind file)
+let compute ?target ?(kind=default_kind) file =
+  (* TODO This process is woefully inefficient. *)
+  let primary =
+    match kind with
+    | `MD5 -> md5 (Digest.to_hex (Digest.file file))
+    | (`SHA256 | `SHA512) as kind ->
+      try
+        if not OpamCoreConfig.(!r.use_openssl) then raise Exit else
+        match
+          OpamSystem.read_command_output ["openssl"; string_of_kind kind; file]
+        with
+        | [l] ->
+          let len = len kind in
+          make kind (String.sub l (String.length l - len) len)
+        | _ -> failwith "openssl error"
+      with OpamSystem.Command_not_found _ | Exit ->
+        make kind (OpamSHA.hash kind file)
+  in
+  let probably_binary name =
+    List.mem (Filename.extension name) [".zip"; ".tar"; ".gz"; ".tgz"; ".tbz"; ".txz"; ".tlz"]
+  in
+  if target = None || target = Some primary || probably_binary file then
+    primary
+  else
+    let buffer = Buffer.create (Unix.((stat file).st_size)) in
+    let ch = open_in_bin file in
+    let no_eol_at_eof ch =
+      seek_in ch (in_channel_length ch - 1);
+      input_char ch <> '\n'
+    in
+    let rec add_cr line buffer =
+      Buffer.add_string buffer line;
+      Buffer.add_string buffer "\r\n";
+      process 2
+    and rem_cr line buffer =
+      let l = String.length line in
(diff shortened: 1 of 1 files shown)


cfcs:
Nice, now I can backdoor OPAM resources and have the checksum check out as validated! :tada:

AltGr:
Good for you! Just have to make sure your backdoor is only made of `\r` characters though... You'd be better off forging md5s, though, since that's still what's in most of the repository currently

cfcs:
PS: Why aren't these considered binary?
- `xz`
- `bz2`
- `tar.gz`
- `tar.bz2`
- `tar.xz`

cfcs:
@AltrGr they can contain of `\n` too, since this patch lets me inject those after any `\r` in the file. Plenty of wiggle-room room for a decent payload.

cfcs:
@AltGr I would of course prefer just being able to pre-image MD5s, but unfortunately I don't have a computer capable of running the best known pre-image algorithm which has a time-complexity of 2^123.4...

hannesm:
I think I understand the issue this PR tries to solve: using git and eolstyle=auto leads to different newline sequences for patches (please correct me if I'm wrong).

now, this solution, AFAICT, computes the hash twice, once with `\r\n` once with `\n`, and see whether any of the hashes equals the `target` (the provided hash from the repository).  this may be fine for text files, but arbitrary binary files may have other semantics for 0x0a and 0x0d (where now insertion is allowed).

I'd suggest another approach: 
- specify that hashes of text files are with newline line endings (likely opam-publish and other tools _on windows_ will need to be adjusted).
--> this can be checked by camelus upon PR!
- for checksum validation _on windows_ do the "remove_cr" bits and compare that hash.
- to discover whether it is a text or binary file, instead of making a blacklist of "likely binary files", let's make a whitelist of "known text files" -- this includes (a) files without an ending (such as `opam`, `descr`), (b) `.diff`, (c) `.patch`, (d) `.sh`, (e) anything else?
- this list of file endings should as well be documented, instead of being hidden ad-hoc in code
- this list could as well be part of the repository configuration, so adding new file endings won't require updates of the opam package itself, but is well-contained in the opam repo.

AltGr:
Thanks @hannesm for the explanation. I am naturally always wary of meddling with hash functions, but unfortunately it seems that we do end up wanting to validate different files with a single hash... Do you think it's a mistake to integrate this into the release in its current state ?
@cfcs thanks for raising interest on the matter, if in an uncommon approach ;)

hannesm:
> @AltGr wanting to validate different files with a single hash... Do you think it's a mistake to integrate this into the release in its current state ?

yes.

an alternative to my above proposal would be to store a pair of hashes, "unix" and "windows" (or have the possibility to do so -- of course painful for contributors without access to a windows machine or for users of windows when there's no windows hash provided).

I still struggle with the "replace `\n` with `\r\n`" from a theoretical perspective: taking a random text file, this may patch lines which include (e.g. as string literal) expressions such as:
`+ Printf.printf "foo bar baz\r\nbalabl\r\nlabla" ;`
now, if we replace all `\r\n` with `\n` on windows to get rid of the line endings, this will break (i.e. compute the wrong hash).  I'm not too familiar with the semantics of `input_line ch`, esp. whether this avoids the corner case above (and if there are other corner cases)!?

dra27:
@cfcs - I don't think the patch lets you inject `\n`. the only place `\n` is put is where it would have already been in the file. It's processed using `input_line`, so the only `\r` characters which are ever analysed are those which already had a `\n` following them.

".xz" and ".bz2" could certainly be added to the list (I think I may have come up with that original list based on a grep of opam-repository). The ".tar." entries are unnecessary - `Filename.extension "foo.tar.gz"` is `".gz"` and so already included.

cfcs:
Buffering the whole file in memory is also kind of problematic IMO.

Consider:
```ocaml
    let buffer = Buffer.create (Unix.((stat file).st_size)) in 
```
This is bounded by the implementation to be between `1` and `Sys.max_string_length`.

- 32-bit OCaml: This will fall somewhere between `1 byte` and `16 MB` regardless of the file size, so as long as there are no OPAM packages with resources > 16MB that's fine, otherwise it crashes during the hashing.

- 64-bit OCaml: Here we have `Sys.max_string_length` of ~130 PB, so we'll likely run out of memory during `Buffer.create` way before we crash in the hashing part.

cfcs:
@dra27 You're correct on both counts! But on the 2nd count I think @hannesm has a point that whitelisting the filenames would be cleaner and would need less adjustment. I'm not sure I'd add `.sh` to the list since that may or may not create problems in whatever tool is used to evaluate `.sh` on Windows.

dra27:
@hannesm - this can be for patches or any text file (e.g. OCaml sources - scripts included in the `files` directory). As to whether the file semantically makes sense, in this section I don't care - it's about whether the file retrieved was valid, or not (so, for example, if you check out a shell script which, I should add, will not typically have a `.sh` extension but be an extension-less filename then it may validate but then fail later when `bash`, `sh`, whatever barfs on the `\r`).

I don't believe that any solution which involves Camelus or adding extra checksums works, because people will not do it (and there are remotes other than opam-repository).

Bear in mind that you can have a git repo which will check out a file with `\r\n` endings **on Unix**. What will have happened there is that a Windows developer checked the file into Git. So you can't just restrict stripping the CR to Windows only.

I don't quite follow your printf'd example - the hashing here is not changing the code, it's just changing the code which was considered for hashing?

PR #3412, opened 2018-06-15
bwrap script: allow writes to /tmp

AltGr:
Previously, `/tmp` was made a dedicated tmpfs; but that completely
breaks running from an opam root within /tmp.


rjbou:
Thanks!

avsm:
How does this work if TMPDIR isn't `/tmp`?

Issue #3411, opened 2018-06-15
behaviour when moving local switches around


AltGr:
Issues arise when one has a local switch and e.g. renames its parent directory, e.g. `foo/_opam` becomes `bar/_opam`. It's difficult to handle, and in any case the OCaml ecosystemdoesn't perfectly handle relocation just yet, but we can probably do better (esp. when the shell hook in is place).

- `_opam/.opam-switch/environment` becomes wrong which leads to `opam env` malfunctionning. For this we should add a check that `OPAM_SWITCH_PREFIX` matches. First opam reads `$OPAM_SWITCH_PREFIX/.opam-switch/environment` and reverts it (here it can't be found) ; then it reads `<current-switch-prefix>/.opam-switch/environment` and applies it (here it's found but obsolete). This leads to `opam env` infinitely adding to e.g. `PATH`... Fixing `OPAM_SWITCH_PREFIX` would avoid that, but still not load a correct environment: the file should actually be re-generated, but that's not allowed from the shell hook (read-only mode), so maybe just bail out in that case ?

- the switch has registered in `~/.opam/config` won't be correct anymore, but that should already be handled fine by the next opam calls (removed and added switches (once used) are detected and automatically updated)

Thanks to @samoht for reporting the original issue to me

PR #3413, opened 2018-06-15
Ensure OPAM_SWITCH_PREFIX is correctly set by `opam env`

AltGr:
even if the files have been moved / the switch renamed.

This is not a complete fix to #3411 (doesn't add a warning, and the PATH
will only become correct once a package operation has been done), but
avoids the pathological cases.


rjbou:
Thanks!

PR #3414, opened 2018-06-15
LF/CRLF-agnostic hashing take 2

dra27:
While going through points made subsequent to merging #3407, I noticed that the hashing code I'd written was not doing precisely what I'd originally intended it to do. This PR:

 - Adds a few extra binary file extensions to the weak heuristic which aborts this scan, as suggested by @cfcs
 - Adds a missing check for hashing empty files
 - Most importantly, switches the logic for allowing an alternate hash to be considered. Based on the first line read from the file, the function will remove `\r` from the end of **every** line (or ~abort~ otherwise *try adding `\r` to every line*) or add `\r` to the end of every line ~**only** if none of the lines already had `\r` at the end~. In the previous implementation, the detection of the former was faulty and the detection of the latter was missing.

As regards the memory "problems" (which is why the weak heuristic for "this is probably a binary file" is there in the first place), there are several options:
 - Separate implementation of MD5 which exposes the state (i.e. as @AltGr's SHA implementations already do) - it's only done this way because OCaml's `Digest` module doesn't expose its workings.
 - Size restriction on the availability of the transform. That said, I suppose a 16MB script is not inconceivable.

My RTT will be slower for the rest of the evening, as I'm not supposed to respond to GitHub PRs while driving... I haven't yet tested this patch, but wanted to get the points from it up for further discussion before leaving.

### DIFF
--- a/src/core/opamHash.ml
+++ b/src/core/opamHash.ml
@@ -103,10 +103,11 @@ let compute ?target ?(kind=default_kind) file =
   let probably_binary name =
     List.mem (Filename.extension name) [".zip"; ".tar"; ".xz"; ".lz"; ".gz"; ".tgz"; ".tbz"; ".txz"; ".tlz"]
   in
-  if target = None || target = Some primary || probably_binary file then
+  let size = Unix.((stat file).st_size) in
+  if target = None || target = Some primary || size = 0 || probably_binary file then
     primary
   else
-    let buffer = Buffer.create (Unix.((stat file).st_size)) in
+    let buffer = Buffer.create size in
     let ch = open_in_bin file in
     let no_eol_at_eof ch =
       seek_in ch (in_channel_length ch - 1);
--- a/src/core/opamHash.ml
+++ b/src/core/opamHash.ml
@@ -101,7 +101,7 @@ let compute ?target ?(kind=default_kind) file =
         make kind (OpamSHA.hash kind file)
   in
   let probably_binary name =
-    List.mem (Filename.extension name) [".zip"; ".tar"; ".gz"; ".tgz"; ".tbz"; ".txz"; ".tlz"]
+    List.mem (Filename.extension name) [".zip"; ".tar"; ".xz"; ".lz"; ".gz"; ".tgz"; ".tbz"; ".txz"; ".tlz"]
   in
   if target = None || target = Some primary || probably_binary file then
     primary
--- a/src/core/opamHash.ml
+++ b/src/core/opamHash.ml
@@ -109,53 +109,67 @@ let compute ?target ?(kind=default_kind) file =
   else
     let buffer = Buffer.create size in
     let ch = open_in_bin file in
-    let no_eol_at_eof ch =
+    let has_eol_at_eof = lazy (
       seek_in ch (in_channel_length ch - 1);
-      input_char ch <> '\n'
+      input_char ch <> '\n')
     in
-    let rec add_cr line buffer =
+    let rec add_cr line =
       Buffer.add_string buffer line;
-      Buffer.add_string buffer "\r\n";
-      process 2
-    and rem_cr line buffer =
-      let l = String.length line in
-      if l = 0 || line.[l - 1] <> '\r' then
-        primary
-      else begin
-        Buffer.add_substring buffer line 0 (l - 1);
-        Buffer.add_char buffer '\n';
-        process 1
-      end
-    and process strip =
+      match input_line ch with
+      | line ->
+          Buffer.add_string buffer "\r\n";
+          add_cr line
+      | exception End_of_file ->
+          if Lazy.force has_eol_at_eof then
+            Buffer.add_string buffer "\r\n";
+          complete ()
+    and rem_cr line =
+      Buffer.add_substring buffer line 0 (String.length line - 1);
+      Buffer.add_char buffer '\n';
       match input_line ch with
       | line ->
           let l = String.length line in
-          if l = 0 || line.[l - 1] <> '\r' then
-            add_cr line buffer
-          else
-            rem_cr line buffer
+          if l = 0 || line.[l - 1] <> '\r' then begin
+            match input_line ch with
+            | _ ->
+                (* A line didn't end with '\r' - try again adding '\r' instead *)
+                seek_in ch 0;
+                Buffer.clear buffer;
+                add_cr (input_line ch)
+            | exception End_of_file ->
+                if Lazy.force has_eol_at_eof then begin
+                  (* The file should have ended with '\r' - try again *)
+                  seek_in ch 0;
+                  Buffer.clear buffer;
+                  add_cr (input_line ch)
+                end else
+                  complete ()
+          end else
+            rem_cr line
       | exception End_of_file ->
-          let contents =
-            let offset =
-              if no_eol_at_eof ch then
-                strip
-              else
-                0
-            in
-            let reqd = Buffer.length buffer - offset in
-            let result = Bytes.create reqd in
-            Buffer.blit buffer 0 result 0 reqd;
-            result
-          in
-          match kind with
-          | `MD5 ->
-              md5 (Digest.to_hex (Digest.bytes contents))
-          | (`SHA256 | `SHA512) as kind ->
-              make kind (OpamSHA.hash_bytes kind contents)
+          complete ()
+    and complete () =
+      let content = Buffer.to_bytes buffer in
+      match kind with
+      | `MD5 ->
+          md5 (Digest.to_hex (Digest.bytes content))
+      | (`SHA256 | `SHA512) as kind ->
+          make kind (OpamSHA.hash_bytes kind content)
     in
-    let result = process 0 in
+    (* Already checked above that the file is non-empty *)
+    let line = input_line ch in
+    let l = String.length line in
+    let result =
+      if l > 0 && line.[l - 1] = '\r' then
+        rem_cr line
+      else
+        add_cr line
+    in
     close_in ch;
-    result
+    if Some result = target then
+      result
+    else
+      primary
 
 let compute_from_string ?(kind=default_kind) str = match kind with
   | `MD5 -> md5 (Digest.to_hex (Digest.string str))


dra27:
(I'm certain that the removing `\r` only from all lines is the right thing to do and was what I was trying to do before - patch files are weird, and I need to think some more about rejecting adding `\r` to every line wrong, because there are valid patch files with contain `\r\r\n`)

dra27:
Revised version - Windows-formatted patches was indeed why I hadn't done that before. The function now determines if it will be CR-stripping or CR-adding based on the first line read. CR-adding is done "blindly" (to allow for patches which may legitimately have multiple CRs) but CR-stripping aborts if a line is found with a CR **and tries again doing CR-adding instead**.

hannesm:
@dra27 thanks for working on this.  Please bear with my restricted exposure to the windows ecosystem.  My angle on this is from the viewpoint on how to add signing to opam, and (exaggerating!) if we just accept any checksum for any file in opam, that means I can't rely on opams checksum mechanism.  Now, I want to have a system which is easy to understand, works nicely, and doesn't let any (MitM-)attacker tamper with checksums.  Before #3407 was merged, this was simple: file X is hashed (resulting in H), and the hash is compared to H', the checksum distributed with the repository.  Now with #3407 being merged, there exists a whole family of files X, X', X'', which succeed in checksum comparison with H'.

I tried in the other issue to figure out whether a solution that hashes only once is sufficient.  In my honest opinion, my opinion is still that any opam repository should only contain one sort of line endings (please see below for some empirical evidence).  Opam2 requires _all patch files_ to be applicable with `patch -p1` - this is very strong assumption, but we cope with it - in a similar vain I believe we can cope with the line endings.

Now back to your proposal: I need to convince myself that all X and X' for which H matches H' are "basically the same" - there must not be any malicious X' that can replace X to execute malicious code.  To achieve that, I try to find what is the minimal variation between X and X' we really need to succeed on all platforms?

From reading your comments here, I'd say `valid input` (or a legitimate text file) has the following form:
- a file with only newline (`\n`) line endings (based on analysis of the first line) -> here adding `\r\n` at the end of each line.
- a file with only carriage-return (`\r\n`) line endings (also based on analysis of the first line) -> here removing all the `\r` at the end of each line.

Is this sufficient, or are there more cases which need to be handled? (I especially struggle with your following remarks, I'd need to see examples of legitimate files with this syntax:
- `only if none of the lines already had \r at the end`
- `CR-stripping aborts if a line is found with a CR and tries again doing CR-adding instead`)

Given that binary files may contain arbitrary `\n` and `\r\n` sequences, which are interpreted by the binary parser, it is crucial to not leverage this trial-hashing mechanism to any binary file.  Your proposed solution is a list of file endings which are binary.  This list is an ad-hoc specified list in `opamHash`, not extensible, and not documented anywhere.  For me, this sounds like a potential source of trouble - I only need to find a single binary format which is not covered by this list and where replacing `\n` by `\r\n` leads to a different semantics.  Updating this list requires releasing a new opam version and having all the users update to it.  That sounds very brittle to me.

I proposed instead to use a list of file endings which are text files, and to include this list as user-configurable option in their `.opam/config`.  In addition, I propose that this very same list of file endings is used in checking utilities such as `opam lint` and `Camelus` to warn if you're referencing a text file that does not contain such a file ending.

To get some numbers on the causa, I analysed the 28360 files in the `packages` directory of the opam-repository.  _NONE_ of these has any CRLF line ending (are we overengineering a non-issue here?) in my opam-repository checkout.  Thereof, the following file endings (discovered by ` basename $x |  tr . \\n | tail -n1`) are popular.  I'm all in for unifying the file endings of these files, e.g. require patches to end with `.patch` or `.diff`, and shell scripts ending with `.sh`.  That'd leave us with `opam`, `META`, `descr`, and `url` without any file ending.
```
   1 _oasis
   1 0002_alpine_linux
   1 buffers
   1 cc
   1 dose
   1 dose3
   1 Fluorine-version
   1 freetennis
   1 mli
   1 obuild
   1 ocp
   1 py
   2 conf
   2 desr
   3 cpp
   4 pkgconfigure
   5 pc
   5 travis
   6 ocaml-stub
   7 configure
   7 r
   8 desc
  11 corebuild
  17 c
  18 in
  33 META
  58 ml
  69 sh
  73 diff
 334 patch
 837 install
8828 url
9002 descr
9023 opam
```

If others (esp. opam maintainers) are convinced there's need for such a trial hashing mechanism on all platforms, opening the attack surface drastically, then please provide documentation and an extensive test suite of this trial hashing mechanism where others can understand what is solved within this code, and what are "good inputs" vs "bad inputs".

Sorry for the many words.

dbuenzli:
> on the causa, I analysed the 28360 files in the packages directory of the opam-repository. NONE of these has any CRLF line ending

This result is not really meaningful. `opam` doesn't work on windows so it's unlikely you'll have users who do not use LF line endings... What @dra27 is trying to do is to cope (in a way I do not understand, see below) with the arrival of future users whose default line ending will be CRLF. 

>  (are we overengineering a non-issue here?)

I didn't follow the discussion closely, but that is my impression. I do not see what the problem is if in opam repos we have `opam` files with LF and CRLF ending. My text editor is smart enough to adapt to those transparently. So just hash what's there.

dra27:
@hannesm - thanks for this, and also sorry for a long reply. Bearing with your restricted exposure to Windows is causing me to review all this further and make tweaks, so it's most definitely all worthwhile, whichever final conclusion we end up at!

If I may start with clearing up some misconceptions in the motivation for this. Note that opam-repository only doesn't contain CRLF endings because *I* fixed it (after spending several years tracking other files which came and went with CRLF endings) in https://github.com/ocaml/opam-repository/pull/11545 and I note in that PR that it should precede the inclusion of a `.gitattributes` file. However, analysis of the *present* state of opam-repository doesn't tell us much - opam-repository presently doesn't support Windows; this feature is to do with readiness of the repository to support Windows in the (near) future.

I wholly agree with your principle that the repository itself should only contain LF endings - but possibly not in the way you mean, as I simply mean that autocrlf should be using used, meaning that the files are always stored in Git's file system LF-only but the *checkouts* of the files may be transformed. Now, while opam may choose with a Git remote to force core.autocrlf to false, this does not allow for the quite normal situation of adding one's local clone of opam-repository to opam as a **File** remote (so one where opam cannot tweak any settings). I strongly regard it as an unnecessary bar to usability for opam to arbitrarily require these checkouts - and indeed any other repositories laid out on disk - to be in one or other format.

Now, while I believe trust issues are extremely important, I'm afraid I rank usability - especially consistent usability - as higher on the basis that we may engineer the most secure and brilliant system in the history of systems, but have no users because the complexity of operating it ends up so high. #3260 is a sufficiently functioning version of opam to build a complete test on Windows - compiled without #3407, this gives errors like this [on native Windows](https://ci.appveyor.com/project/dra27/opam/build/1.0.358/job/wms5coakrbiabmmi#L188) but a complete pass on Unix systems as shown in the overall test matrix results [for that build](https://ci.appveyor.com/project/dra27/opam/build/1.0.358). My overarching consideration is that that scenario leads to yet another blog post, online comment, tweet, whatever about how awful and unapproachable OCaml is (especially on Windows), and so I really struggle to prioritise a highly theoretical security concern over a solution prevents that usability from occurring. (fingers crossed I'm never made to regret that prioritisation, but do please read on!)

Another slight misconception in the original discussion is that this is very far from about patch files. **All** files included in the `files` directory have checksums in opam 2. This affects, for example, the `gen_ocaml_config.ml.in` script used in the [ocaml package](https://github.com/ocaml/opam-repository/tree/2.0.0/packages/ocaml) , the `ocamlfind.install` file included in the [ocamlfind package](https://github.com/ocaml/opam-repository/tree/2.0.0/packages/ocamlfind), and so forth. Also, note that where restricting patches to `patch -p1` (and now even recommending unified diffs, not context diffs, since #3349) affects **all** uses of it, the difference here is having files which work in some places and not others (I find the comparison a teeny bit chalk and cheese, as a result).

In terms of my implementation as it stands, note that the "probably_binary" thing was included for vague reasons of performance, not security - i.e. an heuristic to assume that it was not worth considering the other form of the file because it name suggested it was *probably* binary. So I would say that the discovery of a file format for which this is an actual exploitable problem (I would suggest exploitable for system control, rather than simply DoS or otherwise "irritating") then we would be revisiting this entire mechanism, not updating the list of file extensions. Importantly, to approach this from the other direction I would like to see a function "unequivocally_text" - i.e. where my function is allowed to say `false` for a binary function, your function must always say `true` for a text file (at least to meet my usability requirement).

As things stand, a checksum in the repository refers to one, or possibly two, files:
 - X
 - X' which, if every occurrence in X of `\n` is preceded by `\r`, is X with `\r\n` replaced with `\n`. Otherwise, it is X with `\n` replaced with `\r\n`.

This is slightly more general than the version you mention of looking at the first line and only adding `\r` if the first line doesn't have one, because patch files specifically may have mixed line endings (i.e. there are valid text files which may have mixed `\r\n` and `\n` endings and which remain valid, albeit with a complaint from the `patch` utility, with the transformation `\n` -> `\r\n`).

There are some further hardenings which I agree could sensibly be done. In particular, the way I've implemented it here has made **all** instances of hash checking CRLF-agnostic. Hashes are checked in the following places in the opam sources:
 - `OpamAction.prepare_package_source`: the hashes of `files` listed in `extra-files` are hashed here
 - `OpamPinCommand.source_pin` and `OpamPinCommand.edit` both verify `extra-files` in the same way via the internal function `copy_files`
 - `OpamDownload.download` and `OpamDownload.download_as` both verify checksums of any *downloaded* file via the internal function `really_download`
 - `OpamRepository` uses are mixed - some are because of downloading, some are because of rsyncing
 - `OpamRepositoryBackend.check_digest` is unused, so it's not clear what its context is
 - `OpamFileTools.lint*` also checks `extra-files` when linting
 - `OpamSwitchState.load` uses it for checking external file changes (like the system compiler being upgraded)
 - `OpamUpdate.pinned_package` uses it in two places - one to check the `opam` file and the other to check `extra-files`

It could certainly be changed so that downloading files never allows the alternate version of a file to be considered (this would need a little bit of care in `OpamRepository` - I think it's only used for archives, not for general files as well). The computation of the `opam` hash when updating a pinned package and also the checking of external files could similarly only admit the primary hash.

Tricks which try to work in the other direction - i.e. working out what is a text file - I believe will be doomed to irritating corner cases. For example, shell scripts do not have extensions; scanning based on characters can end up excluding valid (at least for now) OCaml files which might have weird embedded strings. Note also that when linting an opam file, you do not know the opam remote, so storing the configuration in the remote would provide for yet another situation of a previously-thought valid opam remote being rejected later in the process.

My implementation also does all this silently, which in retrospect is certainly not good. This could be fixed:
	• opam could emit a warning if the alternate form of the file is valid. In terms of consistency, I think it would be sensible for Camelus to be ensuring that the checksums in opam-repository are all based on LF checkouts and a consequence of this could also be that the warning would not be displayed by default on Windows.
	• An option `--strict-checksums` (a la both `--require-checksums` and `--strict`) could be added which would prevent the use of the alternate form completely (it could even be baked as a `configure`-time option for setting it as a default). Note that using this option would expose the security-conscious Unix user to the possibility of a checksum error as shown above - and obviously this option would be totally impractical on Windows (cue comments about whether the security-conscious should ever be using Windows…).

dra27:
> > (are we overengineering a non-issue here?)
>
> I didn't follow the discussion closely, but that is my impression. I do not see what the problem is if in opam repos we have opam files with LF and CRLF ending. My text editor is smart enough to adapt to those transparently. So just hash what's there.

The problem is not if the repo contains files with a mix of endings - the problem is that when I check out the git remote on my Windows system, I expect (because it's the default for my Git system on my operating system) to get the files with CRLF endings and I'll checksum them that way. When you check out my commits on your system, you'll get the files with LF endings and consequently broken checksums.

dbuenzli:
@dra27 that looks like a git repository configuration issue, there seem to exist quite a few [options](https://git-scm.com/docs/gitattributes) you can add to the `.gitattributes` of your git `opam` repository repository (sic) that can solve the problem (e.g. forcing checkouts to be LF).

I have the impression that you are trying to solve the problem at the wrong level here.

dra27:
No, it is not a configuration problem - opam should absolutely not be dictating how I configure my files just for the purpose of checksumming! The files are valid files in opam with either line ending. Further, this is something which can be easily got wrong - be it reading the `opam` from for a pin (so any Git repo), a private remote or whatever. Having got it "wrong", it's then non-trivial to try to "fix" it, just to keep checksumming happy. It's a usability disaster.

Forcing checkouts to be LF-only on Windows (for anything other than a shell script) means you aren't writing a portable system anymore.

dra27:
What I'm willing to give on here is the precise mechanism by which the file is found to be valid, but that the file may have had its line-endings changed is something I feel opam must be capable of dealing with.

dbuenzli:
> No, it is not a configuration problem - opam should absolutely not be dictating how I configure my files just for the purpose of checksumming! 

It is to me. There's nothing like configuring "files". Files are data just respect the data. Configure the git repos via `.gitattributes` so that it checkouts your file as they are rather than perform this ugly translation. Again editors are able to cope with these issues transparently.

> The files are valid files in opam with either line ending.

But different as far as the common notion of file hashing ("look at these bytes") is concerned. It seems a bit odd that because of the way you "configure" your files you should change the notion of hashing. 

> Further, this is something which can be easily got wrong - be it reading the opam from for a pin (so any Git repo), a private remote or whatever. Having got it "wrong", it's then non-trivial to try to "fix" it, just to keep checksumming happy. It's a usability disaster.

I'm not sure I see the problem with pins. You don't checksum anything there do you ? What goes wrong and what is the non-trivial fix ? The documentation on how to do a private repository can indicate which gitattributes should be set for this not to be a problem.

> Forcing checkouts to be LF-only on Windows (for anything other than a shell script) means you aren't writing a portable system anymore.

Well that seems a bit of an extreme point of view. But in any case I'm sure you can simply disable that translation, letting CRLF and LF files exist as they want, which should make you happy then.

dra27:
> > No, it is not a configuration problem - opam should absolutely not be dictating how I configure my files just for the purpose of checksumming!

> It is to me. There's nothing like configuring "files". Files are data just respect the data. Configure the git repos via .gitattributes so that it checkouts your file as they are rather than perform this ugly translation. Again editors are able to cope with these issues transparently.

"ugly translation" => you have no interest in portability. It's not ugly. It's the OS. Deal with it. I don't care about editors - I care about the definition that a text file on Windows has CRLF endings. FWIW, my editor does not deal with it transparently, at least by my definition of transparently (meaning I can't tell which it was) - it informs me that the file has come from another OS on every read and write.

> > The files are valid files in opam with either line ending.
> 
> But different as far as the common notion of file hashing ("look at these bytes") is concerned. It seems a bit odd that because of the way you "configure" your files you should change the notion of hashing.

Indeed - this is an aspect which I too do not like. I'm not overly keen on some of the alternative solutions, either, but they all at least involve respecting that the file I provided. However, we don't seem to be talking about those (yet).

> > Further, this is something which can be easily got wrong - be it reading the opam from for a pin (so any Git repo), a private remote or whatever. Having got it "wrong", it's then non-trivial to try to "fix" it, just to keep checksumming happy. It's a usability disaster.

> I'm not sure I see the problem with pins. You don't checksum anything there do you ? What goes wrong and what is the non-trivial fix ? The documentation on how to do a private repository can indicate which gitattributes should be set for this not to be a problem.

The opam file of a pin overlay is check-summed. I haven't double-checked, but I'm fairly sure that when git-pinning, you can provide `files` in the repo along with the opam file.

The fix is non-trivial if you only discover that you should have set .gitattributes *after* trying to sync with it. It's usually fastest at that point to blow away the opam remote entirely and try again which, as I've mentioned, is really great for that next "opam is impossible to use on Windows" Discuss post.

> > Forcing checkouts to be LF-only on Windows (for anything other than a shell script) means you aren't writing a portable system anymore.

> Well that seems a bit of an extreme point of view. But in any case I'm sure you can simply disable that translation, letting CRLF and LF files exist as they want, which should make you happy then.

No - this is the point. I find the requirement to have to disable the translation unacceptable - apart from anything else, the translation is Git's **recommended** default on Windows. My extreme points of view have arisen from the tiny amount of time I've spent working on this over the last 3 years...

dbuenzli:
> "ugly translation" => you have no interest in portability.

I guess that ends the discussion for me.

dbuenzli:
> you have no interest in portability.

These are not my words and the accusation you cast entirely false. I have always cared in my software packages about portability and especially Windows. If that can serve you memory I even went to meetings about porting opam to Windows with you at Microsoft Research Cambridge while I have little interest in the Microsoft platform for myself. 

Portable software entails trade-offs that one has to accept and make sure to make the right ones. I think you are not doing the right one in this case by trying to hack the hash function.

> I'm happy to discuss alternate ways of doing this which may not require weird effects with the hashing but which still respect the underlying fact that the file line endings may be different.

My proposal does exactly that. But again I no longer want to be part of this discussion.

dra27:
This is twisting what each of us is saying out of necessary context. Describing the portability issue that this is attempting to fix as "ugly" is hardly constructive. Holding opinions like that I said *implies* you hold a lack of interest in portability. That this is clearly not true, as of course I am aware, was intended to suggest that you should change the original opinion, because I find the description to be in conflict with caring about portability.

Being pointed to essentially the beginner's guide for .gitattributes in the manual on something with which it should hopefully be clear I'm really quite familiar is, to put it mildly, irritating. So far, the solutions you propose involve imposing changes on the way the Git repositories are created to prevent the situation from arising which I do not find pallatable.

Alternate solutions which might involve not hacking the hash function and also not hacking with the legitimate operation of another piece of software:

 - Relaxing, or even removing, the need for checksums on `extra-files` in a VCS repo (while MD5 is still around, the very existence of this surprised me, given that Hg/Git/etc. are already checksummed). To work effectively, this would need to be extended to local remotes (or with something clever to identify that a local remote is a checkout of a VCS remote?)
 - Something along the lines of always recording the checksums as over LF-ending files, regardless of what's on the disk, which I think is part of an idea @hannesm had proposed.
 - Storing multiple checksums - at first glance, this sort of looks like what's permitted now with sha256 and sha512, but of course it would always be the case that one checksum would be able to fail, which isn't great.

The second two suffer from the problem both of needing to determine exactly which files to do this to and also ensuring that the hashes are maintained (a small benefit of the mechanism I currently have here is that putting the md5sum of the file as found on your computer is always valid).

I could imagine that the first solution could work - it effectively means you stop double-checking Git's work (for git remotes) by assuming that the files checked out from a commit are actually the files which were checked in (modulo CRLF translation, which one now no longer care about) or that local files have already been trusted by the user.

I had considered implementing the first alternative - I considered it to be heavierweight and riskier than what I've already done - that opinion I'm certainly not fixed on.

Pax, I hope.

dbuenzli:
> Holding opinions like that I said implies you hold a lack of interest in portability.

In general it's better not try to model humans with implications. And I reserve my right to call things I see as ugly as such.

> Being pointed to essentially the beginner's guide for .gitattributes in the manual on something with which it should hopefully be clear I'm really quite familiar is, to put it mildly, irritating.

You are not the only one following this discussion. Personally I have never been offended by hyperlinks but will try to remember I should not link to docs when I discuss with you.

> So far, the solutions you propose involve imposing changes on the way the Git repositories are created to prevent the situation from arising which I do not find pallatable.

If this is only needed on the creation of `opam` repository git repositories. I think it is a reasonable portability trade-off. It could even be automated via an `opam repo create` command. If that doesn't work with pins though then I don't think it is reasonable.

cfcs:
Re: the argument: While straw men arguments are a fun tool too provoke other people, I think it would be suiting to keep the personal attacks on people's motivations or intelligence to a minimum here.

I'm still missing a clear justification for this change, ie is there a place where I can learn the motivation for this patch?
What is it exactly that doesn't work now?

dra27:
@cfcs - https://github.com/ocaml/opam/pull/3414#issuecomment-397882744 references two builds showing the problem in action. The motivation is included in #3260, which originally contained #3349, which originally contained #3407, which lead us here!

cfcs:
@dra27 Hmm, my main concern here is really that there should only be one canonical version of a file, and hashing twice is both dangerous and slow. I would *really* like to avoid that.
That being said, I also believe that the hashing *needs* to be streaming, since the 16 MB limit IMHO is unacceptable. `digestif` can do this, I don't know if there are other OCaml libraries for hashing that exposes a streaming interface; but I feel that `$whatever` random library having a crap API is not a good reason to introduce such constraints.

1) Adding a file `.git/info/attributes` inside the cloned git repo with a line `* -text` turns of the text attribute for this clone. We could combine that with the `--template` clone or whatever.

2) It seems to be that setting this to `autocrlf = true` is a mistake in any case (compared to either `autocrlf = false` or `autocrlf = native`). When this weird conversion is turned on (with `= input` or `= true` or `=native` in the case of Windows), the files are still stored with LF inside the git objects. It appears to me that you should be able to do a (CRLF-clean) checkout ignoring the `.gitattributes` if you `git clone` with `--template`, but I haven't tested this. Combined with 1) this could maybe be a solution for OPAM?

3) Since the transformation is only applied to files classified as `text` (which in turn is also governed by `.gitattributes`), we could list the `crlf` setting and the `text` status for each file we're hashing and select the hashing algorithm based on that. In that case we can hash with CRLF newlines **iff** we got a CRLF'ed version in the checkout, and normal hashing otherwise.
Example: `find . -type f -print0 | git check-attr --stdin -z crlf text`

PR #3429, opened 2018-06-22
added shell/nfsopam

UnixJunkie:


rjbou:
Thanks for this script!
I am not sure it's best to have it in the opam repository, but if you would put it in a gist we'd be glad to add a link in the FAQ.

UnixJunkie:
why scatter the info?

AltGr:
well, from experience, `shell/` tends to contain old/deprecated/unmaintained stuff (I plead guilty here). I'd prefer to tidy it up, and keep such scripts that are not really related to opam sources outside of the main tree to help with maintenance. This is a valid use-case though, and your script is likely to prove useful: the FAQ seems like a good place to add (a link to) it, wherever you choose to host it ?

UnixJunkie:
OK, I will create a special repo for it, but its kind of overkill.

UnixJunkie:
ok, the repo is here:
https://github.com/UnixJunkie/nfsopam

dra27:
Alternatively, you could keep it in a gist or just a branch of your own opam fork, which would be less overkill

AltGr:
Keeping open until we add to the FAQ, so we don't forget.

PR #3437, opened 2018-06-29
fix some typos in man pages in the client

avsm:


dra27:
LGTM - you should exercise your commit rights :wink:

avsm:
I can never distinguish 'admin ownership rights' from 'commit rights' and so err on the side of caution, but this is so trivial that I might as well... :-)

PR #3428, opened 2018-06-21
Update download error message

rjbou:
This PR update the download error message display: 
- Fix a lost short message propagation
- by default always show the short one, and the long one in verbose mode

PR #3439, opened 2018-07-02
Don't raise error when retrieving mirror cache fails

rjbou:
This PR completes commit 5ec161401d51c2e153e5467bead902fcab21bc4b (https://github.com/ocaml/opam/pull/3428#issuecomment-401756482), that introduced an error raised on the first mirror cache fail, and some typos.


avsm:
Thanks! I’ll merge this more quickly than usual to unbreak the opam-ci, as the images take some time to build

PR #3442, opened 2018-07-10
Redirect shell hook stderr to /dev/null

### LINKED ISSUE #3431: issue with shell wrapper + opam being desinstalled
If for some reasons the opam binary is not available anymore, every shell invocation will display `-bash: opam: command not found` which is a bit annoying.

rjbou:
This PR fixes #3431, stderr is dropped in order to not pollute user shell in case of inconsistent opam state or uninstalled opam.
- Working on bash, sh, zsh, and fish (csh do not permit redirecting only stderr)
- With fish, do not handle inexistant opam binary, the error is raised at fish level

### DIFF
--- a/src/state/shellscripts/env_hook.sh
+++ b/src/state/shellscripts/env_hook.sh
@@ -1,6 +1,6 @@
 _opam_env_hook() {
  local previous_exit_status=$?;
- eval $(opam env --shell=bash --readonly);
+ eval $(opam env --shell=bash --readonly 2> /dev/null);
  return $previous_exit_status;
 };
 if ! [[ "$PROMPT_COMMAND" =~ _opam_env_hook ]]; then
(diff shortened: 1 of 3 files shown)


src/state/shellscripts/env_hook.sh
```diff
@@ -1,6 +1,6 @@
 _opam_env_hook() {
  local previous_exit_status=$?;
- eval $(opam env --shell=bash --readonly);
```
AltGr: errr... shouldn't the redirect be _outside_ of the `$( )` ? :confused:

rjbou: Redirection must be inside `$( )`, but only stderr must be redirected

AltGr: ah yes, indeed the redirect doesn't apply to the shell expansion, which is the part writing to stderr. Good.

Issue #3433, opened 2018-06-24
Could not update local repository: processed-patch failed


madroach:
On OpenBSD I regularly get this error when updating from my local opam repository:
```
<><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
[local] synchronised from file:///home/madroach/src/opam-repository
[ERROR] Could not update repository "local": "/usr/bin/patch -p1 -i /home/madroach/opam/log/processed-patch-63718-42bb52" exited with code 2
```
I can work around this problem by removing and re-adding the repository.
```
# opam config report
# opam-version      2.0.0~rc3 
# self-upgrade      no
# system            arch=x86_64 os=openbsd os-distribution=openbsd os-version=6.3
# solver            builtin-mccs+glpk
# install-criteria  -removed,-count[version-lag,request],-count[version-lag,changed],-changed
# upgrade-criteria  -removed,-count[version-lag,solution],-new
# jobs              1
# repositories      1 (local)
# pinned            0
# current-switch    default
```

AltGr:
Thanks for reporting. If you have tested an earlier beta or rc, it would be useful to know whether the error is new or wals already there? Thanks.

madroach:
I'm not 100% sure, but I think I saw it also with rc2, but did not test any earlier version. 1.2 had no problems updating repositories.

madroach:
OK, I found the problem. Question remains on how to fix it.
Half of the problem is that OpenBSD ``patch`` will always save backups.
This behaviour can be turned of with this patch:

``` diff
Index: src/core/opamSystem.ml
--- src/core/opamSystem.ml.orig
+++ src/core/opamSystem.ml
@@ -1131,7 +1131,9 @@ let patch ~dir p =
      raise Not_found);
   let p' = temp_file ~auto_clean:false "processed-patch" in
   translate_patch ~dir p p';
-  make_command ~name:"patch" ~dir "patch" ["-p1"; "-i"; p'] @@> fun r ->
+  make_command
+    ~env:Array.(append default_env [|"POSIXLY_CORRECT=yes"|])
+    ~name:"patch" ~dir "patch" ["-p1"; "-i"; p'] @@> fun r ->
     if not (OpamConsole.debug ()) then Sys.remove p';
     if OpamProcess.is_success r then Done None
     else Done (Some (Process_error r))
```

The other half of the problem is OpenBSD ``patch`` not supporting binary diffs.
A binary diff is generated when diff is run over ``.git/index``.
[...]

PR #3443, opened 2018-07-10
No CRLF translation for opam generated patches

rjbou:
This PR is a partial fix for #3433. For patches generated by opam (http & local backends), CRLF translation is not performed.


AltGr:
LGTM!

PR #3444, opened 2018-07-10
Use gpatch on openBSD

rjbou:
This PR is a partial fix for #3433, more specifically https://github.com/ocaml/opam/issues/3433#issuecomment-400346546.
It conflicts with #3443, need to be rebased before merge.

### DIFF
--- a/doc/pages/Install.md
+++ b/doc/pages/Install.md
@@ -120,6 +120,15 @@ cd /usr/ports/sysutils/opam
 make install
 ```
 
+#### FreeBSD
+
+Opam is available in the ports and packages tree on FreeBSD 11 or higher.
+
+```
+cd /usr/ports/devel/ocaml-opam
+make install
+```
+
 #### OSX
 
 Opam packages for [homebrew](http://mxcl.github.com/homebrew/) and
--- a/src/client/opamInitDefaults.ml
+++ b/src/client/opamInitDefaults.ml
@@ -37,7 +37,12 @@ let os_filter os =
 
 let linux_filter = os_filter "linux"
 let macos_filter = os_filter "macos"
+let openbsd_filter = os_filter "openbsd"
+let freebsd_filter = os_filter "freebsd"
 let sandbox_filter = FOr (linux_filter, macos_filter)
+
+let gpatch_filter = FOr (openbsd_filter, freebsd_filter)
+let patch_filter = FNot gpatch_filter
 
 let wrappers ~sandboxing () =
   let cmd t = [
@@ -95,7 +100,8 @@ let required_tools ~sandboxing () =
     Some "A download tool is required, check env variables OPAMCURL or OPAMFETCH",
     None;
     ["diff"], None, None;
-    ["patch"], None, None;
+    ["patch"], None, Some patch_filter;
+    ["gpatch"], None, Some gpatch_filter;
     ["tar"], None, None;
     ["unzip"], None, None;
   ] @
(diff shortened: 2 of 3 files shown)


AltGr:
LGTM!

AltGr:
hah, conflicted with your other PR I just merged!

rjbou:
Rebased. Still need to be tested on openBSD.

hannesm:
on my FreeBSD, I got a similar message as noted in https://github.com/ocaml/opam/issues/3433#issuecomment-400345921 (using opam2  rc2), and was wondering whether opam should require and call out to GNU patch on FreeBSD as well (the port is named devel/patch, the binary installed by it gpatch).

rjbou:
ok, I'll add it  for FreeBSD also then.

hannesm:
while I've seen this once so far, I'm not entirely sure when this happens... is there (for the OpenBSD patch) a reproducible test case which fails with their BSD patch, which I could test on FreeBSD?

dra27:
@hannesm - I probably won’t be able to do it today, but I was planning on adding a binary test to #3456 both to ensure that the patch rewriter no longer messes them up and also to act as a test for this issue.

hannesm:
what I forgot to mention, the above works fine when using `gpatch` -- so I guess: yes, the FreeBSD `patch` command is not well-suited for opam, and `gpatch` should be used instead (plus opam requires the port `devel/patch` as dependency).

rjbou:
Thanks for your test! I'll clean & keep it then.

> how does opam generate a binary diff (and why? - i thought it only every cared about text files)?

For two reasons, on some repo update 1.2 -> 2.0, there is a delete of a remaining `index.tar.gz` file. As the update process is done using `patch`, the removal too. And we can't be certain that custom repo doesn't contain binary files.

rjbou:
ping @jorisgio for commit https://github.com/ocaml/opam/pull/3444/commits/c661ad76822a79b8cdc365057ece42cea722196f, I'm not sure about the FreeBSD version.

jorisgio:
@rjbou wow, i'm sorry about that, I was not even aware i was still listed as maintainer of this port. IIRC i tried to notify various person but it's blurry now, it's been years since i last booted a *BSD.
I've just tried to open a PR but apparently i forgot my password and only managed to lock myself out of freebsd bug tracker...
Will retry later, but i'm afraid i cannot help much.

hannesm:
@jorisgio there is a "lost password" link at https://bugs.freebsd.org/bugzilla/ and the email address you've using is mentioned at https://svnweb.freebsd.org/ports/head/devel/ocaml-opam/Makefile?revision=427548&view=markup#l13 -- I don't know how the governance of FreeBSD ports works (and wasn't able to find much information at https://www.freebsd.org/doc/en/books/porters-handbook/makefile-maintainer.html), but feel free to propose me (hannes AT mehnert dot ORG) as maintainer for `devel/ocaml-opam`. thanks!

rjbou:
@jorisgio no worries :) I pinged you because you were still listed as maintainer. Thanks @hannesm to agree to be the next maintainer!

hannesm:
maybe another approach would be (or is this already possible?), similar to `download-command` in `.opam/config`, have a `patch-command` there?

rjbou:
We could consider a future better solution (in-built patch? as we have some constraints using installed one). For the opam2 release, we just include the `gpatch` choice for BSD in order to [handle the problem](https://github.com/ocaml/opam/issues/3433#issuecomment-400359312).

One of the worry of having the `patch-command` in the config file would be that the user given command must have the same constraint as `patch -p1` in order to correspond to internal handling.

Issue #3434, opened 2018-06-25
Sandbox: opam root not visible


copy:
[ocb-stubblr](https://github.com/pqwy/ocb-stubblr) is an ocamlbuild plugin that invokes the opam binary during compilation. The Linux version of opam's sandbox script makes `$OPAM_SWITCH_PREFIX` visible, but not `$OPAMROOT`, so the ocp-stubblr's opam invokation can't detect the opam root (if it doesn't happen to be in `/home`):

```
ocamlfind ocamlopt -package unix -package ocamlbuild -linkpkg -package ocb-stubblr myocamlbuild.ml /media/samsung/opam2/4.06.1/lib/ocamlbuild/ocamlbuild.cmx -o myocamlbuild
[ERROR] /media/samsung/opam2 exists, but does not appear to be a valid opam root. Please remove it and use `opam init', or specify a different `--root' argument
Failure: Ocb_stubblr: error running opam.
```

rjbou:
Related [mirage issue](https://github.com/mirage/mirage-platform/issues/200).
I reproduce my message here:
_During build, `OPAMROOT` is not binded. Usually it is in `HOME`, which is binded `ro`. At build stage, opam should not be invoked, especially with a command that could change opam state (that needs `OPAMROOT` to be `rw` binded). Needed variables can be found in the environment, defined/found in the opam file and transmitted, or set at this stage._

rjbou:
Resolved via #3540. Thanks!

PR #3446, opened 2018-07-10
Add /rw/usrlocal to bwrap.sh

reynir:
On Qubes `/usr/local/` is usually a symlink to `/rw/usrlocal/`.

I had trouble installing `cpuid` because it calls out to `opam`, but since `/usr/local/` is a symlink to the unmounted path `/rw/usrlocal/` the `opam` binary in `/usr/bin/` got precedence.

### DIFF
--- a/src/state/shellscripts/bwrap.sh
+++ b/src/state/shellscripts/bwrap.sh
@@ -23,7 +23,7 @@ add_mounts() {
     done
 }
 
-add_mounts ro /usr /bin /lib /lib32 /lib64 /etc /opt /nix/store /home
+add_mounts ro /usr /bin /lib /lib32 /lib64 /etc /opt /nix/store /rw/usrlocal /home
 
 # C compilers using `ccache` will write to a shared cache directory
 # that remain writeable. ccache seems widespread in some Fedora systems.


copy:
Related issue: https://github.com/ocaml/opam/issues/3434 and a potentially better way to fix the problem: https://github.com/pqwy/ocb-stubblr/pull/15

reynir:
@copy I tend to agree on your PR, but I think this PR will be useful for Qubes in any case since

* `/usr/local/bin/` is usually in `$PATH`
* `/rw/userlocal/` / `/usr/local/ ` is where you install persistent software in AppVMs

So you would likely want `/usr/local/` in your sandbox anyway, I think.

Another question I have is why Qubes decided to make it a symlink and not a bind mountpoint. If it was the latter then I don't think this would have been an issue. I'll have to ask the Qubes people...

dra27:
This seems a bit hacky - would it not be better to examine the mount-points which are going to be added and, if they are symlinks, add the target of the symlink as well?

reynir:
@dra27 I can add something like this instead?
```
if [ -L "/usr/local" ]; then
    ARGS=("${ARGS[@]} --ro-bind "$(realpath /usr/local)" "/usr/local")
fi
```

reynir:
Do you mean adding a check in `add_mounts` if it's a symlink? If so, then `/usr/local` has to be added to the list explicitly, or it should recursively check for symlinks (probably a terrible idea!)

reynir:
@AltGr I went with that initially after seeing `/nix/store`. Should I remove the latest commit with symlink detection then?

AltGr:
Yes, using `readlink` seems too brittle to me: we could do it if really needed, after careful thought, but it doesn't seem worth it at the moment.
Thanks!

reynir:
Ok, I removed the commit. It's available in the `qubes-readlink` branch for reference.

reynir:
FYI my PR to Qubes that makes the VMs use bind mount instead of symlink was merged. Eventually this workaround would not be necessary, but I'm afraid that could potentially take years :-)

AltGr:
Thanks!

PR #3448, opened 2018-07-11
Pin cache fix

rjbou:
This PR contains pin cache related fixes:
- Cache directory created in tmp dir
- Use pin cache only for no vcs backend

/ @samoht

### DIFF
--- a/src/repository/opamRepository.ml
+++ b/src/repository/opamRepository.ml
@@ -172,7 +172,8 @@ let pull_from_upstream
    else
    let pin_cache_dir = OpamRepositoryPath.pin_cache url in
    let url, pull =
-     if OpamFilename.exists_dir pin_cache_dir then
+     if OpamUrl.(match url.backend with | #version_control -> false | _ -> true)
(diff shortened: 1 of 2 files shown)


src/client/opamPinCommand.ml
```diff
@@ -482,9 +485,13 @@ and source_pin
...
          using the one found locally.";
       Some local
-    | Some local, Some vers when not (OpamFile.OPAM.effectively_equal local vers) ->
+    | Some local, Some vers when
+        not OpamFile.(OPAM.effectively_equal
+                        (OPAM.with_url URL.empty local)
+                        (OPAM.with_url URL.empty vers)) ->
       OpamConsole.warning
-        "Opam file have significant uncommitted changes, using the versioned one";
+        "%s opam file have significant uncommitted changes, using the versioned one"
```
AltGr: `%s's opam file has non-trivial…` ?
"significant" could be interpreted as "large" in this context, which is misleading.

PR #3449, opened 2018-07-12
Fix binary patching

dra27:
This is still slightly in progress.

The first commit fixes `OpamSystem.translate_patch` corrupting binary patches by incorrectly applying CRLF transformations. The revised version both detects that a file has mixed line endings (i.e. some are CRLF and some are LF) and does the same thing with *all* the chunks of the patch. If either the file being patched or the patch chunks themselves have mixed endings, then no translations occur. The edge cases for binary files which happen to be CRLF or LF encoded are also sound (the patch will either maintain them and, if it changes the "endings" then the `+` lines and `-` lines would have different endings which would also disable the function).

However, doing that has made the original function even uglier than it was and so the following commit (which is the work-in-progress part - I just haven't tidied it up) rewrites the function. Before, `translate_patch` read the entire patch into memory (mostly reasonable) and folded the list of lines once. The problem is that the chunk line-ending detection means that lines now have to be cached in the state machine, which is a ~bit~lot hideous. The new version accepts that the file is read 3 times - the first to determine if GNU `patch`'s CR-stripping will come into effect, the second time to work out which lines may or may not require CR characters to be added or removed and then the third time to actually write the patch.

It also seems prudent to start testing this, and I've written a rudimentary test harness for that function. There are a couple of commits added which control the debug output a bit further, which allows the log to be used as a reference file.

### DIFF
--- a/src/core/opamSystem.ml
+++ b/src/core/opamSystem.ml
@@ -821,58 +821,69 @@ let lock_none = {
 let lock_isatleast flag lock =
   lock_max flag lock.kind = lock.kind
 
-(* read_lines processes the lines of a text file in a platform-agnostic manner
-   (which is to say that it preserves '\r' on all platforms). It has two modes
-   of operation - in the Lines mode, it returns a string list of the lines, in
-   CrLf mode, it returns true if every line ends "\r\n". *)
-type _ read_return = Lines : string list read_return
-                   | CrLf  : bool option read_return
-
-let read_lines : type s . s read_return -> string -> s = fun mode file ->
+let get_eol_encoding file =
   let ch =
     try open_in_bin file
     with Sys_error _ -> raise (File_not_found file)
   in
-  let read_lines f =
-    let rec read_lines cr acc =
-      match input_line ch with
-      | line ->
-          let length = String.length line in
-          let acc = f line acc in
-          let cr =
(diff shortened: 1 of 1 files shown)


AltGr:
Just tested on an old repo of mine that gave me errors like:
```
+ /usr/bin/patch "-p1" "-i" "/home/lg/.opam/log/processed-patch-2688-58c514" (CWD=/home/lg/.opam/repo/default-git)
- patching file packages/ocamlspot/ocamlspot.4.02.1.2.3.0/opam
- Hunk #1 FAILED at 1.
- Hunk #2 FAILED at 14 (different line endings).
- 2 out of 2 hunks FAILED -- saving rejects to file packages/ocamlspot/ocamlspot.4.02.1.2.3.0/opam.rej
- patching file packages/ocamlspot/ocamlspot.4.02.1.2.3.0/opam.orig
- Reversed (or previously applied) patch detected!  Assume -R? [n] 
- Apply anyway? [n] 
- Skipping patch.
- 1 out of 1 hunk ignored -- saving rejects to file packages/ocamlspot/ocamlspot.4.02.1.2.3.0/opam.orig.rej
- patching file packages/ocamlspot/ocamlspot.4.02.1.2.3.0/opam.orig.rej
```
and I can confirm it seems to work fine with this new version !

PR #3451, opened 2018-07-13
"opam admin add-hashes" helper to add more secure hashes to the repo

AltGr:
This is still WIP: cache should be saved at some interval, not just at
the end, and hashes other than source (e.g. extra-sources) are not
considered.


AltGr:
Should now be complete!

PR #3454, opened 2018-07-13
Make '--locked' a full-fledged build option

AltGr:
there is no reason it should be available in 'opam install' and not
'opam pin'; this also avoids the issue that 'opam update PKG' would not
accept the --locked option and detect a changed opam file.

There is still a few corner-cases, the combination with the
pinned-package-update can sometimes be a bit strange.

### DIFF
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -149,19 +149,21 @@ type build_options = {
   jobs          : int option;
   ignore_constraints_on: name list option;
   unlock_base   : bool;
+  locked: string option;
 }
 
 let create_build_options
     keep_build_dir reuse_build_dir inplace_build make no_checksums
     req_checksums build_test build_doc show dryrun skip_update
-    fake jobs ignore_constraints_on unlock_base = {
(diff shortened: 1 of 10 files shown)


rjbou:
Missing `guess_name` handling? in `OpamPinned.get_name`, name are retrieved only from the `.opam`.

AltGr:
`guess_name` uses `OpamPinned.files_in_source`, which behaviour is changed by the option; so that should be OK ?

rjbou:
Nop, because [here](https://github.com/ocaml/opam/pull/3454/files#diff-8e8884272e348cf9c755f05001bd507cR78), if locked, filename is replaced by this one. Then [here](https://github.com/ocaml/opam/blob/master/src/state/opamPinned.ml#L72), `name_of_opam_filename` is called with the new filename that doesn't recognize locked files, so it takes the directory name as package name.

AltGr:
ah indeed, `name_of_opam_filename` needs to be patched as well. Thanks!

PR #3456, opened 2018-07-16
Patch rewrite tests

dra27:
This splits off the tests added in #3449. Three tasks remain for this PR:
 - ~The debug_sections should be converted to be an int option String.Map.t so that there is internally at least the option to have different levels for different sections.~
 - ~`OPAMDEBUGSECTIONS` needs adding (either just supporting a list of sections, or full-blown support for specifying levels as well)~
 - The test cases and the harness are still the result of my iterative hacking, and need tidying up...

### DIFF
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -236,6 +236,9 @@ let help_sections = [
   `P "$(i,OPAMCURL) can be used to select a given 'curl' program. See \
       $(i,OPAMFETCH) for more options.";
   `P "$(i,OPAMDEBUG) see options `--debug' and `--debug-level'.";
+  `P "$(i,OPAMDEBUGSECTIONS) if set, limits debug messages to the space-separated \
+      list of sections. Sections can optionally have a specific debug level \
+      (for example, $(b,CLIENT:2) or $(b,CLIENT CUDF:2), but otherwise use `--debug-level'.";
   `P "$(i,OPAMDOWNLOADJOBS) sets the maximum number of simultaneous downloads.";
   `P "$(i,OPAMDRYRUN) see option `--dry-run`";
   `P "$(i,OPAMEDITOR) sets the editor to use for opam file editing, overrides \
--- a/src/core/opamConsole.ml
+++ b/src/core/opamConsole.ml
@@ -496,7 +496,17 @@ let timestamp () =
     (int_of_float (1000.0 *. msec))
 
 let log section ?(level=1) fmt =
-  let debug_level = OpamCoreConfig.(!r.debug_level) in
+  let debug_level =
+    let debug_level = OpamCoreConfig.(!r.debug_level) in
+    let sections = OpamCoreConfig.(!r.debug_sections) in
+    if OpamCoreConfig.StringMap.is_empty sections then
+      debug_level
+    else
+      match OpamCoreConfig.StringMap.find section sections with
+      | Some level -> level
+      | None -> debug_level
+      | exception Not_found -> 0
+  in
   if level <= abs debug_level then
     let () = clear_status () in
     let timestamp = if debug_level < 0 then "" else timestamp () ^ "  " in
--- a/src/core/opamConsole.ml
+++ b/src/core/opamConsole.ml
@@ -13,7 +13,7 @@ open OpamCompat
 
 (* Global configuration *)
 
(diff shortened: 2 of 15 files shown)


rjbou:
The conflict comes from doc change for [`temp_file`](https://github.com/ocaml/opam/pull/3453/commits/ea01e46bfd7305c3b5f6120273923b5c1182a390).

src/core/opamConsole.ml
```diff
@@ -496,19 +496,28 @@ let timestamp () =
...
+    let debug_level = OpamCoreConfig.(!r.debug_level) in
+    let sections = OpamCoreConfig.(!r.debug_sections) in
+    if OpamCoreConfig.StringSet.is_empty sections || OpamCoreConfig.StringSet.mem section sections then
+      debug_level
+    else
+      0
+  in
+  if level <= abs debug_level then
     let () = clear_status () in
+    let timestamp = if debug_level < 0 then "" else timestamp () in
```
rjbou:
```suggestion
    let timestamp = if debug_level < 0 then "" else (timestamp ()) ^ "  " in 
```

In order to not have remaining spaces at the beginning of the line. But this implies to update the tests/patcher-test.reference file.

src/core/opamConsole.ml
```diff
@@ -496,19 +496,28 @@ let timestamp () =
...
     if Sys.win32 then begin
       (*
        * In order not to break [slog], split the output into two. A side-effect
        * of this is that logging lines may not use colour.
        *)
       win32_print_message `stderr (Printf.sprintf "%s  %a  "
```
rjbou:
```suggestion
       win32_print_message `stderr (Printf.sprintf "%s%a  "
```

src/core/opamConsole.ml
```diff
@@ -496,19 +496,28 @@ let timestamp () =
...
-        (timestamp ()) (acolor_with_width (Some 30) `yellow) section);
+        timestamp (acolor_with_width (Some 30) `yellow) section);
       Printf.fprintf stderr (fmt ^^ "\n%!") end
     else
       Printf.fprintf stderr ("%s  %a  " ^^ fmt ^^ "\n%!")
```
rjbou:
```suggestion
       Printf.fprintf stderr ("%s%a  " ^^ fmt ^^ "\n%!")
```

tests/patcher.ml
```diff
@@ -0,0 +1,107 @@
+let set_debug_level n sections =
+  let debug_sections =
+    List.fold_left (fun set elt -> OpamCoreConfig.StringSet.add elt set) OpamCoreConfig.StringSet.empty sections
```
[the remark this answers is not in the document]
dra27: Indeed - although now it's a map!

src/core/opamConsole.ml
dra27: Done!

src/core/opamStd.ml
```diff
@@ -1370,7 +1386,7 @@ module Config = struct
     in
     OpamCoreConfig.(setk (setk (fun c -> r := c; k)) !r)
       ?debug_level:(env_level "DEBUG")
-      ?debug_sections:None
+      ?debug_sections:(env_sections "DEBUGSECTIONS")
```
rjbou: One last thing, it need to be added in [doc](https://github.com/ocaml/opam/blob/master/src/client/opamArg.ml#L238).

dra27: Done!

src/client/opamArg.ml
```diff
@@ -236,6 +236,9 @@ let help_sections = [
   `P "$(i,OPAMCURL) can be used to select a given 'curl' program. See \
       $(i,OPAMFETCH) for more options.";
   `P "$(i,OPAMDEBUG) see options `--debug' and `--debug-level'.";
+  `P "$(i,OPAMDEBUGSECTIONS) if set, limits debug messages to the space-separated \
+      list of sections. Sections can optionally have a specific debug level \
+      (for example, $(b,CLIENT:2)), but otherwise use `--debug-level'.";
```
rjbou:
```suggestion
      (for example, $(b,CLIENT:2) or $(b,CLIENT CUDF:2), but otherwise use `--debug-level'.";
```

rjbou:
lgtm,thanks!

PR #3459, opened 2018-07-16
Update src_ext extlib to 1.7.5

dra27:
Necessary for OCaml 4.07.0 `lib-ext` building (see https://github.com/ocaml/homebrew-ocaml/issues/11)


AltGr:
Thanks!

PR #3468, opened 2018-07-18
Fix regression on the concurrent lock message

AltGr:


dra27:
`%!` can't be used any more - can't remember the detail, but this isn't the right fix.

dra27:
(that's a TODO list item for me as in the middle of a rehearsal!)

dra27:
Looking at the definition, shouldn't `formatted_msg` be flushing already (see definition of `OpamConsole.print_message`)?

AltGr:
Probably didn't look closely enough... but it definitely wasn't flushing the unfinished line!

dra27:
It’s going ksprintf, so the `%!` shouldn’t be doing anything - is there another effect?

AltGr:
Indeed, the channel should be flushed... and can't reproduce today, must have seen an artifact. Sorry for the noise, will dig more if it happens again.

Issue #3455, opened 2018-07-14
opam switch should use make -j


pmetzger:
When opam switch builds a new compiler, it should use `make -j N` where `N` is taken from the `jobs:` configuration parameter. This would dramatically speed up building compilers.

pmetzger:
(Note it might be that opam 2.0 already does this; I'm still using opam 1.2. If so, please close this.)

dra27:
This is an opam-repository issue, not an opam issue - opam should definitely not invoke `make -j` by default.

I can short-circuit your raising the issue on opam-repository, though - the compiler's build system is not stable enough to do `make -j` as a matter of course.

pmetzger:
I thought that "opam switch" invocations were managed by opam itself, and not by opam-repository. Presumably that's wrong?

I've been told by a lot of people that invoking `make -j` for the compiler is stable. If that's not the case, then that should be fixed -- there are now even laptops with six cores and twelve threads, and single threaded builds are feeling very slow -- but of course that issue needs to be raised in the ocaml repo.

dra27:
That is indeed wrong - in both 1.2.2 and 2.0, the commands executed for setting up switches are in the repository.

We test the parallel build, but I’m not convinced that that’s a good default to have without some kind of fallback on error. Out of curiosity, who/where is “lots of people”?

pmetzger:
@gasche and others have told me it's find for doing clean builds (which is what opam does.) There was a discussion on discuss.ocaml.org, and some discussion on the discord chat channel as well. It certainly speeds up the build a great deal.

If it's buggy, it should be debugged, but this is probably not the right place to discuss that of course.

gasche:
Parallel clean builds are tested in the CI and seem to have worked robustly in the last year or so on `trunk`, but I also agree with @dra27's incomfort at making that the default for unaware users, without another fallback strategy than "just run the command again". Given that installing a new switch is not a very frequent operation for users, it does seem sensible to wait until we move from "we know it's broken but it seems to work in practice" to "we trust it to work properly" to make this the default.

pmetzger:
What would be needed to gain confidence beyond it working for a year in the CI system?

I rebuild switches often enough that I find it a bit irritating, but I understand the reluctance if it still isn't considered robust. That said, would there be any obvious way to make it optional so someone like me can just turn on parallel builds?

pmetzger:
> Would you care to elaborate on what is your need for rebuilding switches?

[...]

PR #3474, opened 2018-07-21
Updates for OCaml 4.07.1

dra27:
- Three tweaks to the `src_ext` vendoring script.
 - Minor fix to AppVeyor bootstrap compaction
 - Shift the cold compiler to 4.07.~0~1 and include 4.07.~0~1 in the test matrix
 - ~Disable Cygwin64 testing - it's been only been enabled because the cached copy of OCaml 4.06.0 was still in place. flexdll needs patching, I've nagged the guy who wrote the patch to finish it...~ (cherry-picked in cac57424)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ src_ext/camlp4
 src_ext/findlib
 src_ext/ocamlbuild
 src_ext/topkg
+src_ext/seq
 src_ext/*.*stamp
 src_ext/*.tbz
 src_ext/*.tar.gz
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -106,7 +106,7 @@ EOF
         fi
         ./configure --prefix ~/local -no-graph -no-debugger ${CONFIGURE_SWITCHES:-}
         if [[ $OPAM_TEST -eq 1 ]] ; then
-          make -j world.opt
+          make -j 4 world.opt
         else
           make world.opt
         fi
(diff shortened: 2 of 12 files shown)


dra27:
Oh dear - that last commit is a bit worrying!

dra27:
@pmetzger - just for interest (apropos of #3455), that final commit is an example of `make -j` causing problems with OCaml 4.07.0 (still trying to investigate why, sadly...)

pmetzger:
@dra27 🙁

pmetzger:
@dra27 I wonder if there's some way we could build something to deliberately exercise lots of randomized build orderings in order to shake out the make -j issue. But that should be discussed in the OCaml mantis or some such I suspect.

dra27:
Given that it’s repeatable, and apparently only affecting this specific platform, I fear it will turn out to be an annoying `make` thing...

pmetzger:
Just one note: Travis has a fairly short timeout based on wall clock time and not on CPU time. (Dumb I think, but I didn't design Travis.) I don't think that will cause problems for this in particular, but you should be aware of it in case it comes up.

dra27:
The build cache is being a nuisance here - https://travis-ci.org/dra27/opam/builds/407124218 is running with the caches blown away

dra27:
Just a ping on this one - rebased and good-to-go

rjbou:
Great! Thanks!

PR #3476, opened 2018-07-23
make cold: check md5

rjbou:


shell/bootstrap-ocaml.sh
```diff
@@ -13,12 +13,22 @@ fi
...
 URL=`sed -ne 's/URL_ocaml *= *//p' ../src_ext/Makefile | tr -d '\r'`
+MD5=`sed -ne 's/MD5_ocaml *= *//p' ../src_ext/Makefile | tr -d '\r'`
 V=`echo ${URL}| sed -e 's/.*\/\([^\/]\+\)\.tar\.gz/\1/'`
 FV_URL=`sed -ne 's/URL_flexdll *= *//p' ../src_ext/Makefile | tr -d '\r'`
 FLEXDLL=`echo ${FV_URL}| sed -e 's/.*\/\([^\/]*\)/\1/'`
 if [ ! -e ${V}.tar.gz ]; then
   cp ../src_ext/archives/${V}.tar.gz . 2>/dev/null || ${CURL} ${URL}
 fi
+
+ACTUALMD5=`md5sum ${V}.tar.gz | sed -e 's/^[^a-f0-9]*\([a-f0-9]*\).*/\1/'`
```
hannesm: is this supposed to work on all platforms?  if so, `md5sum` is not universal -- on FreeBSD and OpenBSD, it should be ``ACTUALMD5=`md5 -q ${V}.tar.gz` `` (where `-q` is the quiet mode and only prints the checksum, not the filename).

hannesm: yes! (even though OpenBSD ships LibreSSL, this still provides an `openssl` binary which meets the requirements here (accepts `md5` and outputs in the same format).

rjbou: great! thanks :)

PR #3487, opened 2018-07-30
Respect user's TMPDIR when invoking bwrap

3noch:
Before this change, the invocation to `bwrap` assumed `TMPDIR=/tmp` and/or that scripts hardcode `/tmp`. This binds `/tmp` in the bubblewrap to the host `TMPDIR` and then points all tmp-related env vars to `/tmp`. Notably, this makes `opam` work better on NixOS. It may also affect which temp dir gets used on other systems.

### DIFF
--- a/src/state/shellscripts/bwrap.sh
+++ b/src/state/shellscripts/bwrap.sh
@@ -1,4 +1,6 @@
-#!/bin/bash -ue
+#!/usr/bin/env bash
+
+set -ue
 
 if ! command -v bwrap >/dev/null; then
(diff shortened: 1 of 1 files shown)


3noch:
Note that the only build failure happens in the build of ocaml itself.

dra27:
The Cygwin64 build failure is not your fault - see #3474

AltGr:
This will break if `$TMPDIR` is not defined, you should write `${TMPDIR:-/tmp}` instead.

Thanks for the contribution!

rjbou:
LGTM! Thanks!

PR #3490, opened 2018-08-01
do not execute validation hook if update is empty

hannesm:
if the update is empty, there's no reason to execute the validation hook


hannesm:
NB: I'm not sure whether the logic for `Update_empty` should be instead of here in `OpamRepository.update`.

rjbou:
> NB: I'm not sure whether the logic for Update_empty should be instead of here in OpamRepository.update.

Indeed, to me it should be integrated in `OpamRepository.update`: the pattern matching on update is done there for this purpose (what if error, what if empty, what if else), and no need to call the validation to directly out the function. 
Maybe adding a log entry, something like `log "Update empty, no validation permorfed"`.

Thanks!

rjbou:
We need to keep the [msg](https://github.com/ocaml/opam/blob/960dad9df190436b853efe18d90583c829b315c8/src/repository/opamRepository.ml#L429) and the [vcs post-treatment](https://github.com/ocaml/opam/blob/960dad9df190436b853efe18d90583c829b315c8/src/repository/opamRepository.ml#L464), even if the update is empty (ftm, action with vcs/hg backend). I think that "cleaner" way would be to keep only `apply_repo_update` & `repo_update_complete` calls (`cleanup_repo_update` does nothing in this case): 
```ocaml
   | Update_empty ->
    log "update empty, no validation performed";
    apply_repo_update repo upd @@+ fun () ->
    B.repo_update_complete repo.repo_root repo.repo_url
```

Afterwards, is it possible to squash the commits, in order to merge a clean PR (the second commit reverts the first one).

Thanks!

hannesm:
I re-added those function calls in db841b1.

I'm a bit curious: `apply_repo_update` is a local function which only outputs (using both `OpamConsole.msg` and `log`), IMHO this should just `assert false` in the `Update_empty` branch.

For `repo_update_complete`, I went through the source and could only see any effect with `hg` (which I do not have installed so I can't test this).

Feel free to merge+squash this PR (the GitHub UI should allow you do do this).

rjbou:
Indeed, they are the two reasons why I suggested that modification:
- `apply_repo_update`: as all update logging is located there, it is in order to not move that message & keep them "grouped" in that function
- `repo_update_complete`: as there is at least one backend that perform an action even is the update is empty, it is to keep the same behavior. I didn't test the `hg` backend, i'll ping @AltGr  on that.

> Feel free to merge+squash this PR (the GitHub UI should allow you do do this).

Nice, discovering it :)

hannesm:
anything I should do to move this PR forward?

PR #3516, opened 2018-08-24
Don't continue with checksum check if openssl sha512 is unavailable

Bronsa:
On some macOS versions openssl doesn't come with sha256 support by default:
```
[~] openssl sha512 < /dev/null
openssl:Error: 'sha512' is an invalid command.
```

If that's the case the installation script currently fails with a checksum mismatch, which is not desiderable

### DIFF
--- a/shell/install.sh
+++ b/shell/install.sh
@@ -158,7 +158,7 @@ bin_sha512() {
 }
 
 check_sha512() {
-  if command -v openssl > /dev/null && ! openssl sha512 2>&1 < /dev/null | grep Error > /dev/null; then
+  if command -v openssl > /dev/null && [ "$(openssl sha512 2>&1 < /dev/null)" = "cf83e1357eefb8bdf1542850d66d8007d620e4050b5715dc83f4a921d36ce9ce47d0d13c5d85f2b0ff8318d2877eec2f63b931bd47417a81a538327af927da3e" ]; then
     sha512=`openssl sha512 "$TMP/$OPAM_BIN" 2> /dev/null | cut -f 2 -d ' '`
     check=`bin_sha512`
     test "x$sha512" = "x$check"
--- a/shell/install.sh
+++ b/shell/install.sh
@@ -158,7 +158,7 @@ bin_sha512() {
 }
 
 check_sha512() {
-  if command -v openssl > /dev/null; then
+  if command -v openssl > /dev/null && ! openssl sha512 2>&1 < /dev/null | grep Error > /dev/null; then
     sha512=`openssl sha512 "$TMP/$OPAM_BIN" 2> /dev/null | cut -f 2 -d ' '`
     check=`bin_sha512`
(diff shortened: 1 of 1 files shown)


rjbou:
It would be better to rely on return code than grep an error message (they depends of the os, the locale, etc.). For example, here I don't have `Error`, which  makes the condition true even if it fails:
```shell
$ openssl sha522 2>&1 < /dev/null 
Invalid command 'sha522'; type "help" for a list.
$ openssl sha522 > /dev/null 2>&1 < /dev/null 
$ echo $?
1
$ openssl sha512 > /dev/null 2>&1 < /dev/null 
$ echo $?
0
```

Thanks for the contribution!

Bronsa:
Good point! how does it look now?

Bronsa:
Actually, we've just found yet another weirdness with macOS: on some systems a missing command still returns with a code of 0 :/

 ```sh
[~]> openssl sha513 2>&1 < /dev/null
openssl:Error: 'sha513' is an invalid command.

Standard commands
asn1parse         ca                certhash          ciphers
crl               crl2pkcs7         dgst              dh
dhparam           dsa               dsaparam          ec
ecparam           enc               engine            errstr
gendh             gendsa            genpkey           genrsa
nseq              ocsp              passwd            pkcs12
pkcs7             pkcs8             pkey              pkeyparam
pkeyutl           prime             rand              req
rsa               rsautl            s_client          s_server
s_time            sess_id           smime             speed
spkac             ts                verify            version
x509

Message Digest commands (see the `dgst' command for more details)
gost-mac          md4               md5               md_gost94
ripemd160         sha               sha1              sha224
sha256            sha384            sha512            streebog256
streebog512       whirlpool

Cipher commands (see the `enc' command for more details)
aes-128-cbc       aes-128-ecb       aes-192-cbc       aes-192-ecb
aes-256-cbc       aes-256-ecb       base64            bf
bf-cbc            bf-cfb            bf-ecb            bf-ofb
camellia-128-cbc  camellia-128-ecb  camellia-192-cbc  camellia-192-ecb
camellia-256-cbc  camellia-256-ecb  cast              cast-cbc
cast5-cbc         cast5-cfb         cast5-ecb         cast5-ofb
chacha            des               des-cbc           des-cfb
des-ecb           des-ede           des-ede-cbc       des-ede-cfb
des-ede-ofb       des-ede3          des-ede3-cbc      des-ede3-cfb
des-ede3-ofb      des-ofb           des3              desx
rc2               rc2-40-cbc        rc2-64-cbc        rc2-cbc
rc2-cfb           rc2-ecb           rc2-ofb           rc4
rc4-40

[~]> echo $?
0
```

I've made a patch that checks directly the return value, this seems to be the most secure approach

rjbou:
erf, If we can't rely on return code anymore...
It's ok for me like that. I'll squash and merge the PR.
Thanks!!

Bronsa:
Awesome, thanks!

PR #3529, opened 2018-09-04
Patch mccs to build on darwin.

xplat:
Contains https://github.com/AltGr/ocaml-mccs/pull/15 .


rjbou:
I think that It's better to have the fix merged on ocaml-mccs and then update opam dependency, instead of having the same fix in two different places.

rjbou:
`ocaml-mccs` PR was merged and a new version published. The referenced PR updates `mccs` archive, which is less complex than having to update the patch accordingly.
Thanks!

Issue #3503, opened 2018-08-10
opam 2.0 can't be built on Cygwin64


yannl35133:
I have tried different methods of installing opam (the Windows WIP guide, `make cold`, the full archive on [the opam website](https://opam.ocaml.org/2.0-preview/doc/Install.html), none of which worked for me, and each causing different errors.
I am new to opam, and the only thing I want is to get merlin and its code styling, please let me know how to get it.
If you want error reports, ask me what exactly to put in them.

AltGr:
Windows support is still pretty much work in progress, but @dra27 has been doing tremendous work, part of which has been merged into opam 2.0.0, and we are soon getting there!

I'll let him elaborate :)

dra27:
opam 2.0.0 should be working in Cygwin, yes - that's as in compiled with Cygwin's gcc, not any of the native ports of OCaml (mingw-w64 or MSVC).

Are you using 32-bit Cygwin or 64-bit Cygwin? 64-bit Cygwin is completely broken for OCaml at the moment, I'm afraid.

yannl35133:
I am using 64-bit Cygwin, and most of my tries were with mingw, so I guess this answers my problems. I am only using Cygwin for opam and OCaml, so I am going to retry with 32-bit Cygwin right now.
Thank you very much, I'll tell you if it worked with 32-bit Cygwin

yannl35133:
It worked out, I got no problems, perfect.
So maybe you should put it somewhere in 'README.md' that opam (2.0.0) does not work yet with 64-bit Cygwin.
Anyway, thanks a lot.

PS: Should I close the issue now, even though my problems are still there ?

dra27:
Good, glad it's working on Cygwin32!

Up to @rjbou where to record the Cygwin64 issue - opam itself is not broken, the issue is that you can't build OCaml. I realise that that distinction is of limited comfort when your end goal is simply being able to use opam, but my point is that there's nothing that would be fixed in opam, other than bumping the version of OCaml built in `make cold`.

For me, this issue could sensibly be renamed "opam 2.0 can't be built on Cygwin64" and left open to track this?

rjbou:
Updated related readme documentation, and yes, better keep the issue open like that it is can be more easily found.
Thanks for reporting!

mclagett:
I'm experiencing fork issues when trying to install Opam on cygwin running on windows 10.  Here's the error:  

[...]

PR #3532, opened 2018-09-05
doc: update readme, remove cygwin64 install

rjbou:
This PR is related to #3503.
It updates the readme as opam does not compile with cygwin64.
/cc @dra27


dra27:
Ah, I'm afraid this PR is not necessary. Cygwin is used in two contexts: for the 4 native Windows ports, it provides the build system (which is what is described in these parts of the docs) and Cygwin64 is better than Cygwin32. However, this is for when you're building opam to run natively on Windows, not opam to run within Cygwin. If you're building opam to run within Cygwin - e.g. Cygwin's own opam package - then that is broken on Cygwin64 at the moment.

rjbou:
oh, ok, Thanks for the explanation!

PR #3538, opened 2018-09-11
Update OpenBSD support, OpenBSD requires gtar

adamsteen:
OpenBSD tar does not support the ‘J’ flag

### DIFF
--- a/src/client/opamInitDefaults.ml
+++ b/src/client/opamInitDefaults.ml
@@ -44,6 +44,9 @@ let sandbox_filter = FOr (linux_filter, macos_filter)
 let gpatch_filter = FOr (openbsd_filter, freebsd_filter)
 let patch_filter = FNot gpatch_filter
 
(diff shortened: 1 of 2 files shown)


rjbou:
Thanks for the contribution!

Ok for me also, I'll just ping @hannesm about FreeBSD, to check if it is needed to  also add it (cf. #3444) or keep the change OpenBSD specific.

hannesm:
from FreeBSD `tar` man page (bsdtar 3.3.2 - libarchive 3.3.2 zlib/1.2.11 liblzma/5.2.4 bz2lib/1.0.6):
```
     -J, --xz
             (c mode only) Compress the resulting archive with xz(1).  In
             extract or list modes, this option is ignored.  Note that this
             tar implementation recognizes XZ compression automatically when
             reading archives.
```

so a `gtar` should not be needed on FreeBSD

rjbou:
Thanks!

Issue #3534, opened 2018-09-06
opam2 -  bwrap - hook/sandbox.sh does not allow to mount exotic path


ErwanGa:
Hi
I'm using opam in my build system.
I'm currently switching on opam2 and I'm currently facing an issue while creating a new switch.

It stops at ocaml-variants build 
```
# Configuring OCaml version 4.06.1
# Configuring for host x86_64-unknown-linux-gnu ...
# Configuring for target x86_64-unknown-linux-gnu ...
# Using compiler /work/CG-2.1/_build/vm-current/host/bin/x86_64-linux-gcc.
# ./configure: 1: ./configure: /work/CG-2.1/_build/vm-current/host/bin/x86_64-linux-gcc: not found
# Compiler family and version: .
# [ERROR!] Unable to compile the test program.
#  Make sure the C compiler /work/CG-2.1/_build/vm-current/host/bin/x86_64-linux-gcc is properly installed.

[ERROR] The compilation of ocaml-variants failed at "/work/CG-2.1/_build/_opam/opam-init/hooks/sandbox.sh build
        ./configure -prefix /work/CG-2.1/_build/_opam/vm-guichet -cc
        /work/CG-2.1/_build/vm-current/host/bin/x86_64-linux-gcc -as
        /work/CG-2.1/_build/vm-current/host/bin/x86_64-linux-as -aspp
        /work/CG-2.1/_build/vm-current/host/bin/x86_64-linux-gcc -c -partialld
        /work/CG-2.1/_build/vm-current/host/bin/x86_64-linux-ld -r -no-curses -fPIC".
```
The _opam/opam-init/hooks/sandbox.sh file seems not able to mount an exotic path for gcc compilers

``` 
 build)
        add_mounts ro "$OPAM_SWITCH_PREFIX"
        add_mounts rw "$PWD"
        add_ccache_mount
        ;;
``` 
is enough  not as my $PWD is something like /work/CG-2.1/buildsystem/ and does not contains /work/CG-2.1/_build/

I've updated mine as following (and adapt my build system with the required new variable)
``` 
 build)
        add_mounts ro "$OPAM_SWITCH_PREFIX"
        add_mounts rw "$PWD"
        if  [ -n "${OPAM_USER_PATH_RO-}" ]; then
           add_mounts ro "$OPAM_USER_PATH_RO"
        fi
        if  [ -n "${OPAM_USER_PATH_RW-}" ]; then
[...]

PR #3540, opened 2018-09-12
Adding way to mount unusual path in bwrap sandbox 

ErwanGa:
- add usage OPAM_USER_PATH_RO environment variable for setting
    unusual paths in mount bind_ro list

    - remove some unusual path from the script (/nix/store /rw/usrlocal)

    - fix problem of build in the sandbox
        Some modules generates an error if the complete infra of {opam_root}
        is not seen in the sandbox
             -> [ERROR] /work/CG-2.1/_build/_opam exists, but does not appear
                to be a valid opam root.....
         They also need to write in $(dirname $OPAM_SWITCH_PREFIX)/log
        =>
            Need to mount $(dirname $OPAM_SWITCH_PREFIX) in ro
        but need to mount $(dirname $OPAM_SWITCH_PREFIX)/log in rw

### DIFF
--- a/src/state/shellscripts/bwrap.sh
+++ b/src/state/shellscripts/bwrap.sh
@@ -27,7 +27,11 @@ add_mounts() {
     done
 }
 
-add_mounts ro /usr /bin /lib /lib32 /lib64 /etc /opt /nix/store /rw/usrlocal /home
+# remove some unusual pathes (/nix/stored and /rw/usrlocal )
+# use OPAM_USER_PATH_RO variable to add them
+# the OPAM_USER_PATH_RO format is the same as PATH
+# ie: export OPAM_USER_PATH_RO=/nix/store:/rw/usrlocal
+add_mounts ro /usr /bin /lib /lib32 /lib64 /etc /opt /home
 
 # C compilers using `ccache` will write to a shared cache directory
 # that remain writeable. ccache seems widespread in some Fedora systems.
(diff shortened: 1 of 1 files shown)


src/state/shellscripts/bwrap.sh
```diff
@@ -49,16 +54,43 @@ add_ccache_mount() {
...
+        #     Please remove it and use `opam init', or specify a different `--root' argument
+        # They also need to write in $(dirname $OPAM_SWITCH_PREFIX)/log
+        #  => Need to mount $(dirname $OPAM_SWITCH_PREFIX)/log in rw
+        add_mounts rw "$(dirname $OPAM_SWITCH_PREFIX)/log"
+        add_mounts ro "$(dirname $OPAM_SWITCH_PREFIX)/repo"
+        add_mounts ro "$(dirname $OPAM_SWITCH_PREFIX)/config"
+        add_mounts ro "$(dirname $OPAM_SWITCH_PREFIX)/config.lock"
+        add_mounts ro "$(dirname $OPAM_SWITCH_PREFIX)/download-cache"
+        add_mounts ro "$(dirname $OPAM_SWITCH_PREFIX)/lock"
+        add_mounts ro "$(dirname $OPAM_SWITCH_PREFIX)/opam-init"
```
rjbou: I'm afraid there is some confusion between `OPAMROOT` which contains [global configuration files](http://opam.ocaml.org/doc/2.0/Manual.html#opam-root) and `OPAM_SWITCH_PREFIX` which contains [local switch configuration and installation structure](http://opam.ocaml.org/doc/2.0/Manual.html#Structure). It's the opam root that contains those files, not the switch prefix. Plus, at build stage, opam don't need to access to its opam root (if so, it's a [packaging issue](https://github.com/ocaml/opam/issues/3434)).
I've checked and I can confirm that it works as is: having opam root and opam switch in an unlisted directory, launching an install from another one unrelated.
What is you configuration? You can have internal values of root & switch with `opam config list`.

rjbou:
Thanks for the contribution!

I'll ping @reynir to inform about the `/rw/usrlocal` change.

PR related to #3534.

src/state/shellscripts/bwrap.sh
ErwanGa: Ok
Seems to be package issue.
Shall I do another pr without this part of the patch ?

rjbou: > Seems to be package issue.

What is the package you tried to install?

> Shall I do another pr without this part of the patch ?

No need to create a new one, you can update this PR by updating the branch you used. You can squash commits in order to have a clean history.

ErwanGa: I've such issue with package cpuid  @ version 0.1.1.

I'll update the pr asap

rjbou: Usually, cpuid can be installed as `$HOME` is ro mounted and it uses opam to access configuration (via `ocb-stubblr`, cf. above links). The `ocb-stubblr` PR is not yet merged fixing it, so meanwhile it still need to access opam root, but just in ro.
I tried using the new `OPAM_USER_PATH_RO` variable set to an unusual opam root and it installed with no problem.

ErwanGa: Understood...
PR has been updated. It now only contains the OPAM_USER_PATH_RO patch

rjbou:
Thanks!

PR #3541, opened 2018-09-13
Locked state and signals, avoid 

rjbou:
When opam is blocked (waiting for a lock to be released) signals abort this blocking state, like SIGWINCH wich is launched when a window is resized. This PR catch interruption caused by signals in order to ignore them.
Terminal signals still terminates the program.

### DIFF
--- a/src/core/opamSystem.ml
+++ b/src/core/opamSystem.ml
@@ -816,7 +816,10 @@ let rec flock_update
            Unix.(lockf fd F_ULOCK 0);
          Unix.lockf fd (unix_lock_op ~dontblock:true flag) 0
        with Unix.Unix_error (Unix.EAGAIN,_,_) ->
-         if dontblock then raise Locked;
+         if dontblock then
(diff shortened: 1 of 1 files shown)


AltGr:
Note that opam normally handles `SIGWINCH`, and re-reads the window width to adapt line-wrapping. Ignoring it, even temporarily, will break this (non vital) feature if you resize the window while opam is waiting for the lock. Wouldn't it be possible to retry the locking after the signal was handled ? ("It wouldn't be practical" is a valid answer, signals are a mess...)

rjbou:
A solution would be to reraise the `SIGWINCH` once the lock released. But it isn't very precise as catched exception  `Unix.Unix_error (Unix.EINTR,"lockf",_)` don't specify which signal is the source of the interruption, it can be `SIGWINCH` or `SIGPIPE`.

AltGr:
Sorry, I misunderstood and thought you were using `Sig_ignore`. This is not the case, and will cause no problems with `SIGWINCH` :)

PR #3566, opened 2018-09-27
doc/dev-manual: zero date (version is already part of title)

### LINKED ISSUE #3562: the opam debian package is not reproducible
someone brought to my attention that opam2, now in buster, is not entirely reproducible.  the very kind people from the reproducible builds project host a website containing their results: https://tests.reproducible-builds.org/debian/rb-pkg/buster/amd64/diffoscope-results/opam.html

what do we see there? the (optional) opam-docs package, which installs the doc/dev-manual (still at version 1.1.2 AFAICT), does contain - as common in latex - a timestamp of when the tex was translated to a pdf.

I think this can easily be avoided by using some waternarking (`%%VERSION%%`) in the tex - or unset `\date{}`. The timestamp of when latex was run is not very informative, it is much more interesting which opam release this document is targeting, or the git head!?

I haven't looked inside the dev-manual at all, but if it is specific for opam1, maybe it could be deleted or needs to be updated.

hannesm:
fixes #3562 (and solves the issue at https://tests.reproducible-builds.org/debian/rb-pkg/buster/amd64/diffoscope-results/opam.html)


rjbou:
As the dev manual is unmaintained/obsolete, maybe we can delete it completely?

rjbou:
Thanks

PR #3570, opened 2018-09-30
Minor cleanup bwrap.sh

jpathy:
Use bash pattern substitution instead.


src/state/shellscripts/bwrap.sh
```diff
@@ -55,7 +55,7 @@ case "$COMMAND" in
     build)
         # mount unusual path in ro
         if  [ -n "${OPAM_USER_PATH_RO-}" ]; then
-           add_mounts ro $(echo ${OPAM_USER_PATH_RO} | sed 's|:| |g')
+           add_mounts ro ${OPAM_USER_PATH_RO://:/ }
```
dra27: This, and all of the others, have a stray `:` before the `//`

PR #3571, opened 2018-09-30
minor: improve documentation of opam repo add

hannesm:
I'm still not entirely sure what: `If you want to enable a repository only to install its switches, you may be looking for $(b,opam switch create --repositories=REPOS).` is supposed to mean - does a repository contain `switches`? can I `opam switch create --repositories=RRR` where `RRR` is not known to opam before?

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -1549,9 +1549,9 @@ let repository =
     (* using an unbreakable space here will indent the text paragraph at the level
        of the previous labelled paragraph, which is what we want for our note. *)
     "$(b,Note:) By default, the repository is only added to the current \
-     switch. To add a switch to other repositories, you need to use the \
+     switch. To add a repository to other switches, you need to use the \
      $(b,--all) or $(b,--set-default) options (see below). If you want to \
-     enable a repository only to install of of its switches, you may be \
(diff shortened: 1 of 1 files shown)


AltGr:
Yes, as per `opam switch create`:

>  `--repositories=REPOS`
>        When creating a new switch, use the given selection of repositories
>        instead of the default. REPOS should be a comma-separated list of
>        either already registered repository names (configured through e.g.
>        opam repository add --dont-select), **or NAME=URL bindings, in which
>        case NAME should not be registered already to a different URL, and
>        the new repository will be registered.** See opam repository for more
>        details. This option also affects list-available.

tests/jbuild
```diff
@@ -3,4 +3,4 @@
 (alias
   ((name runtest)
    (deps ((files_recursively_in .) ../src/client/opamMain.exe ../src/tools/opam_check.exe))
-   (action (run make all))))
+   (action (run gmake all))))
```
rjbou: Is it a change you wanted to push? It seems a local-BSD modification, not [portable](https://travis-ci.org/ocaml/opam/jobs/435208178#L3219).

rjbou:
Thanks!

PR #3572, opened 2018-09-30
Fix configure tests

### LINKED ISSUE #3551: configure for 2.0.0 does not allow FETCH=no
I'm constructing the `Portfile` for the MacPorts opam package.

As we don't want to actually fetch any dependencies, I'm using the version of the distfile that contains all dependencies in it already. However, if "curl" or "wget" isn't available, configure fails. As I don't actually _want_ to fetch anything, having to depend on one of those two is suboptimal.

I can set `FETCH` in the environment to `no`, which prevents this, but then a test in the configure script complains that either curl or wget must be available.

For now I'm patching this out and the build proceeds just fine, but it would be nice to have an official configure flag that says "don't depend on a fetch program, I'm using the cached version of the dependencies".

dra27:
Addresses two shortcomings of the configure script:

 - not having wget or curl is now only a hard-error if `src_ext/archives` doesn't contain the archives (i.e. if `make -C src_ext cache-archives` has not been run) which means that this should no longer be a requirement for building with the "full" tarball
 - the C++ test for MCCS is now moved to the correct place and the message as to whether the solver will be built should work correctly with `--disable-checks`. Note that not having a C++ compiler is not considered an error if `--disable-checks` is specified because `configure` is permitted to believe that the mccs library exists even if it couldn't detect it.

cc @hannesm, @pmetzger

Fixes #3551

### DIFF
--- a/configure
+++ b/configure
@@ -5112,7 +5112,7 @@ $as_echo "not found" >&6; }
 
 
 if test "x${with_mccs}" = "xno" && test "x$OCAML_PKG_mccs" != "xno"; then :
-  as_fn_error $? "Option --without-mccs is not supported. You need to uninstall the 'mccs' package" "$LINENO" 5
+  as_fn_error $? "Option --without-mccs is not available without uninstalling the 'mccs' package" "$LINENO" 5
 fi
 
 
--- a/configure
+++ b/configure
@@ -6637,11 +6637,14 @@ bindir="`eval echo ${bindir}`"
 mandir="`eval echo ${mandir}`"
 mandir="`eval echo ${mandir}`"
 
-if { test "x$MCCS_ENABLED" = "xfalse" && test "x${hasalldeps}" = "x"; } ||
-       { test "x$OCAML_PKG_mccs" = "xno" && test "x${hasalldeps}" = "xtrue"; }; then :
+if test "x$MCCS_ENABLED" = "xfalse"; then :
+
   echo "Opam will be built WITHOUT a built-in solver"
+
 else
+
   echo "Opam will be built WITH a built-in solver"
+
 fi
 echo
 echo Executables will be installed in ${bindir}
--- a/configure
+++ b/configure
@@ -5116,22 +5116,9 @@ if test "x${with_mccs}" = "xno" && test "x$OCAML_PKG_mccs" != "xno"; then :
 fi
 
 
+if test "x$MCCS_ENABLED" = "xtrue"; then :
 
-echo
-
-
-if test "x${enable_checks}" != "xno" && {
-       test "x$OCAML_PKG_extlib" = "xno" ||
-       test "x$OCAML_PKG_re" = "xno" ||
-       test "x$OCAML_PKG_cmdliner" = "xno" ||
-       test "x$OCAML_PKG_ocamlgraph" = "xno" ||
-       test "x$OCAML_PKG_cudf" = "xno" ||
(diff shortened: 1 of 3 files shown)


pmetzger:
I still don't quite understand why disabling checks should be the way that I specify that I don't want to fetch anything under any circumstances. Checks are fine. Fetching is a problem.

dra27:
Disabling checks doesn’t do that, all it ever did was continue if neither wget nor curl was found.

pmetzger:
So can we change the name to something more intuitive, like `--disable-fetch`?

dra27:
No, both for compatibility and because it doesn’t disable fetching!

dra27:
FWIW, where does it say it’s required?

Your package is of course your own concern, but your patch doesn’t “disable” fetching per se - there are several parts of the build system which will guess at fetching anyway.

The point here is that the full tarball *won’t* do a fetch during the build and this patch prevents an error if your system happens not to include a fetch command (as it’s not needed).

If you truly wish to be *disabling* network access then you either need to actually disable it (or use a sandbox).

pmetzger:
> FWIW, where does it say it’s required?

MacPorts policy.

> your patch doesn’t “disable” fetching per se - there are several parts of the build system which will guess at fetching anyway

I'll have to disable those more thoroughly then.

> If you truly wish to be disabling network access then you either need to actually disable it (or use a sandbox).

Or inform the build system that the fetch tool is `/usr/bin/false`. 😉

dra27:
> > FWIW, where does it say it’s required?

> MacPorts policy.

I was hoping for an actual link, if you have one? I couldn't come across one to satisfy my curiosity...

pmetzger:
Some of this stuff isn't well written down. The MacPorts guide contains some of this stuff, but this particular thing isn't. The deal is, though, that with very few exceptions, the Portfile is intended to describe the entire set of inputs into a constructed package for auditing purposes. (The exceptions are a handful of ports that pull directly from a source control system for unfortunate and unavoidable reasons — they don't fetch other stuff during build either.) There are also some other considerations, like the fact that our fetch phase is again supposed to be distinct and controllable. In theory you can build almost all of MacPorts with a mirror of the sources and no live network access.

pmetzger:
(There may be stronger documentation from Debian. They're much better at policy documents. Finding the right one may be a bit of a challenge though.)

hannesm:
> The point here is that the full tarball won’t do a fetch during the build and this patch prevents an error if your system happens not to include a fetch command (as it’s not needed).

thanks @dra27, I haven't tested this PR yet (due to travelling), but am happy with that.  This is sufficient for my FreeBSD package (which currently uses `--disable-checks`).

dra27:
@hannesm - excellent, thanks. I have asked for this to be considered for 2.0.1 as well, just for the convenience of the display being better.

dra27:
@pmetzger - I think you're taking the no network access thing far too literally. Debian is not a source distro, so it's a weak comparison - even more so as their opam source package [depends on curl](https://packages.debian.org/source/sid/opam) and their [patch series](https://salsa.debian.org/ocaml-team/opam/tree/master/debian/patches) doesn't touch the fetch stuff :wink:

Gentoo's [ebuild guide](https://devmanual.gentoo.org/ebuild-writing/functions/index.html), to give a better example, uses the phrase "Avoid network access"...

rjbou:
Thanks!

PR #3574, opened 2018-10-02
Allow build to specify GNU sed executable

rneswold:
Some uses of `sed` in the `Makefiles` specify the `-i` option which is only found in GNU `sed`. On my BSD system, GNU `sed` gets installed as `gsed`. This pull request still defaults to using `sed`, but allows the user to specify a different executable by defining the `SED` variable.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,6 +17,7 @@ else
 endif
 
 OPAMINSTALLER = ./opam-installer$(EXE)
+SED ?= sed
 
 ALWAYS:
 	@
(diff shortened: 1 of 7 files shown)


rjbou:
lgtm,thanks for the contribution!

dra27:
Oops - the offending one was almost certainly added by me. I didn't realise that `-i` wasn't portable, (though I guess it does something useful, but not compulsory so it's not surprising!). I agree this change is nice, but it would also be sensible to eliminate the use of the non-portable flag (at least in `src_ext/Makefile`) which I'll aim to include in the port of the build system to Dune which I've been working on.

hannesm:
> sensible to eliminate the use of the non-portable flag

I agree with that, would be great to have this in the next release (a non-dependency to GNU sed, but using POSIX sed).

rjbou:
#3603 merged, closing this one.
Thanks!

PR #3588, opened 2018-10-09
Update FAQ URIs for bubblewrap

nekketsuuu:
This PR updates URIs in error messages for bwrap. The new URI is <http://opam.ocaml.org/doc/FAQ.html#Why-does-opam-require-bwrap>.


rjbou:
Thanks!

Issue #3576, opened 2018-10-03
Failure writing temporary files in MacOS X's /var/folders


herbelin:
Hi, after upgrading to opam 2.0.0, I get fatal `couldn't create cache file '/var/folders/...'` and `unable to open output file '/var/folders/...'` errors on MacOS 10.11.1 for many packages (e.g. `ocamlfind`, `ocamlbuild`, `camlp5`, ...). This is apparently due to sandboxing permissions, since, if I add the line `add_mounts rw "/var/folders/my/local_cryptic_name_folder` to `sandbox.sh`, it stops failing. However, I have no idea what is the right thing to do. Do you have an idea whether this is a problem of configuration specifically on my installation or an opam-level problem? For instance, what are the policies wrt temporary files, their location, and there permissions?

`opam config report` gives:
```
# opam config report
# opam-version      2.0.0 
# self-upgrade      no
# system            arch=x86_64 os=macos os-distribution=homebrew os-version=10.11.1
# solver            builtin-mccs+glpk
# install-criteria  -removed,-count[version-lag,request],-count[version-lag,changed],-changed
# upgrade-criteria  -removed,-count[version-lag,solution],-new
# jobs              4
# repositories      5 (http), 1 (local), 2 (version-controlled) (default repo at d25a75a2)
# pinned            0
# current-switch    4.06.1
```

rjbou:
Thanks for reporting.
On MacOS X, the TMPDIR var is `rw` mounted, but when undefined the default value is `/tmp` ([here](https://github.com/ocaml/opam/blob/master/src/state/shellscripts/sandbox_exec.sh#L15)).
It appears that `/var/folders` is used for caching, but it should be defined in the `TMPDIR` (according this [blog post](http://www.magnusviri.com/Mac/what-is-var-folders.html)).  As there is randomisation on this folder, is it not possible just give `rw` permission to the specific session folder.
On solution would be, in case `TMPDIR` is undefined, get the information from `DARWIN_USER_TEMP_DIR` or `DARWIN_USER_CACHE_DIR`. Can you check that they are well defined even if you changed the default setup?
If an MacOS user have a suggestion, or a objection / validation, it's very welcome!

Note that on the linux script, we added the variable `OPAM_USER_PATH_RO` (cf . #3540 & #3534), in order to let user locate its opam root in unusual paths (but not useful for build which need to be `rw`).

herbelin:
[...]

PR #3597, opened 2018-10-10
Sandboxing on MacOS: considering the possibility that TMPDIR in unset

herbelin:
This is a PR to address #3576 (see discussion there).

~~I added a test which presumably makes that the fix works on both linux and MacOS X (on MacOS X, we use `getconf` to get the per-user temporary directory).~~

@rjbou clarified that the file `sandbox_exec.sh` is actually only for MacOS X. I updated the PR to simplify the comment explaining the situation.

Tested on MacOS 10.11.1 both with TMPDIR set and unset.

### DIFF
--- a/src/state/shellscripts/sandbox_exec.sh
+++ b/src/state/shellscripts/sandbox_exec.sh
@@ -12,7 +12,20 @@ add_mounts() {
     esac
 }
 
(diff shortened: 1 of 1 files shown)


rjbou:
Thanks!

PR #3603, opened 2018-10-12
avoid sed -i, a GNU sed extension.  use > YYY.tmp and mv YYY.tmp YYY instead

hannesm:
this is an alternative of #3574 /cc @dra27 @rneswold


dra27:
LGTM,  thanks!

rjbou:
Thanks!

PR #3605, opened 2018-10-13
[Sandbox] make /var readonly instead of empty and rw

### LINKED ISSUE #3604: access to `/var/lib/mlocate/mlocate.db`
Hi,

I'm updating a `conf-ida` package from `opam 1.2` to `opam 2.0`. And I stuck trying to install it on `ubuntu/xenial64`.

I use `locate` in my shell script to find a program path: `locate -w -r idaq64$`. It worked with previous opam, but now I got the error: 
`locate: can not stat () `/var/lib/mlocate/mlocate.db': No such file or directory`.
I assume that's because of the sandboxing in opam 2.0. But I can't use anything else instead of `locate`, since IDA can be installed at any place.

So, is it a way to permit an access to the database?

The whole purpose of conf package is to interact with a system, so are there any specific rules
applicable to conf packages that I can use in order to solve the problem?

bobot:
Should fix #3604 

What is the security logic of this file? Why not all `/` ro , instead of sherry picking? It should be stated in the file at least.


rjbou:
Thanks!

Issue #3600, opened 2018-10-11
ocaml-version >= "4.02.3" not satisfied by ocaml 4.06.1 on old switch after upgrade to opam 2


Blaisorblade:
This is a followup on #3594 (which might be the same or a different issue) /cc @rjbou.

TLDR:
```
$ opam install coq
[ERROR] coq = 8.8.0 unmet availability conditions: ocaml-version >= "4.02.3"
```
Currently running `opam reinstall -y ocaml-base-compiler` in case that helps.

```
pgiarrusso@SilenceOfTheLambdas:~/git/Coq/dot-calculus$ opam list
# Packages matching: installed
# Name              # Installed       # Synopsis
base-bigarray       base
base-threads        base
base-unix           base
camlp5              7.06.10-g84ce6cc4 Preprocessor-pretty-printer of OCaml
conf-m4             1                 Virtual package relying on m4
num                 1.1               The legacy Num library for arbitrary-precision integer and rational arithmetic
ocaml               4.06.1            The OCaml compiler (virtual package)
ocaml-base-compiler 4.06.1            Official 4.06.1 release
ocamlfind           1.8.0             A library manager for OCaml
ott                 0.28              Ott is a tool for writing definitions of programming languages and calculi
pgiarrusso@SilenceOfTheLambdas:~/git/Coq/dot-calculus$ opam install coq-equations
The following dependencies couldn't be met:
  - coq-equations → coq >= 8.8
      unmet availability conditions, e.g. ocaml-version >= "4.02.3"

No solution found, exiting
$ opam install coq
[ERROR] coq = 8.8.0 unmet availability conditions: ocaml-version >= "4.02.3"
```

```
pgiarrusso@SilenceOfTheLambdas:~/git/Coq/minidot/ecoop17$ opam switch
#  switch                   compiler                    description
   4.05.0                   ocaml-base-compiler.4.05.0
→  coq-8.8.0                ocaml-base-compiler.4.06.1  base-bigarray.base base-threads.base base-unix.base ocaml-base-compiler.4.06.1
   coq-8.8.1                ocaml-base-compiler.4.06.1  base-bigarray.base base-threads.base base-unix.base ocaml-base-compiler.4.06.1
   coq-8.8.1-equations-1.0  ocaml-base-compiler.4.06.1
   coq-8.8.2                ocaml-base-compiler.4.07.0  coq-8.8.2
(reverse-i-search)`env': opam ^Cv
[...]

PR #3610, opened 2018-10-16
Upgrade pin source

rjbou:
This PR contains a fix for format upgrade, it add extraction of archived source files of version-pinned packages. opam1.2 stores the archive file while opam2 the extract.
Related to #3600.

### DIFF
--- a/src/core/opamFilename.mli
+++ b/src/core/opamFilename.mli
@@ -194,6 +194,9 @@ val install: ?exec:bool -> src:t -> dst:t -> unit -> unit
     directory if possible. Otherwise, the symlink is absolute. *)
 val link: ?relative:bool -> target:t -> link:t -> unit
 
+(** Returns true if the given file is an archive (zip or tar) *)
+val is_archive: t -> bool
+
 (** Extract an archive in a given directory (it rewrites the root to
     match [Dir.t] dir if needed) *)
 val extract: t -> Dir.t -> unit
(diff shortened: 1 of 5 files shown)


src/core/opamFilename.mli
```diff
@@ -194,6 +194,9 @@ val install: ?exec:bool -> src:t -> dst:t -> unit -> unit
     directory if possible. Otherwise, the symlink is absolute. *)
 val link: ?relative:bool -> target:t -> link:t -> unit
 
+(** Returns trus is the given file is an archive (zip or tar) *)
```
AltGr:
```suggestion
(** Returns true if the given file is an archive (zip or tar) *)
```

src/state/opamFormatUpgrade.ml
```diff
@@ -958,13 +958,20 @@ let from_2_0_alpha3_to_2_0_beta root conf =
...
+                let nv = OpamPackage.package_of_name state.sel_installed name in
+                sources_dir / OpamPackage.to_string nv
+            in
+            OpamFilename.move_dir ~src:d ~dst;
+            (* Extract version-pinned archives to source dirs *)
+            match OpamFilename.files dst with
+            | file::[] when OpamFilename.is_archive file ->
+              OpamFilename.extract_in file dst;
+              OpamFilename.remove file
+            | _ -> ()
```
AltGr: Wouldn't it be better to extract directly without moving the archive first ?
This could cause a bug in the (very unlikely) case where `foo.tar` contains a `foo.tar` file.

rjbou: fixed!

PR #3618, opened 2018-10-19
Port build system to Dune

dra27:


dra27:
This can be merged, although it's not quite complete - I totally forgot that I haven't upgraded the `lib-ext` system from Jbuilder to Dune - it still works of course, while Dune is in transition.

dra27:
I will aim to do that part over the next few days...

PR #3617, opened 2018-10-19
Eliminate renames in patches

dra27:
I forgot to use `git format-patch --no-renames` when I generated the patches for seq.

Recommended for 2.0.1 as mentioned earlier.

PR #3620, opened 2018-10-23
Eliminate renames in patches (#3617)

rjbou:
* Eliminate renames in patches

* ensure that `patch` removes zero-length files

This fixes the build on some platforms like Alpine where
this behaviour is not active by default, and thus results
in empty `.ml` files being generated in the source tree.

* Revert "ensure that `patch` removes zero-length files"

This reverts commit f6d001761256eec3e9db640f82e377823b44be21.

* Workaround strict POSIX patching

POSIX patch will leave src_ext/seq/seq.ml{,i} as zero-length files but
patch -E is non-POSIX.


rjbou:
This is a cherry-pick of #3617

PR #3628, opened 2018-10-24
Fix #3627: use `git -c diff.noprefix=false diff`

### LINKED ISSUE #3627: `opam update` fails because of custom git settings `diff.noprefix`
`opam update` fails on the iris-dev repo if my `~/.gitconfig` contains:
```
[diff]
        noprefix = true
```

That changes the output of `git diff` to default to `-p0` not `-p1` — but `opam` generates a diff with `git "diff" "--no-ext-diff" "-R" "-p" "refs/remotes/opam-ref" "--"` and tries to apply that with `patch -p1` during `opam update`.

The minimal possible fix seems to reset this option via `git -c diff.noprefix=false "diff"` (I might try this), but always skipping the local git config files could be more robust.

`man git` has this suggestion:
```
       GIT_CONFIG_NOSYSTEM
           Whether to skip reading settings from the system-wide $(prefix)/etc/gitconfig file. This environment variable can be used along with $HOME and $XDG_CONFIG_HOME to create a predictable environment
           for a picky script, or you can set it temporarily to avoid using a buggy /etc/gitconfig file while waiting for someone with sufficient permissions to fix it.
```

## Command outputs

Install works:
```
$ opam repo add --all-switches iris-dev https://gitlab.mpi-sws.org/FP/opam-dev.git
[iris-dev] Initialised

<><> Upgrading repositories from older opam format ><><><><><><><><><><><><>  🐫
Upgrading repository "iris-dev"...
```

Update doesn't:
```
$ opam update

<><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><>  🐫
[coq-released] synchronised from http://coq.inria.fr/opam/released
[iris-dev] synchronised from git+https://gitlab.mpi-sws.org/FP/opam-dev.git
[ERROR] Could not update repository "iris-dev": "/usr/bin/patch -p1 -i /Users/pgiarrusso/.opam/log/processed-patch-22224-c743ac" exited with code 1
[default] no changes from https://opam.ocaml.org/

<><> Upgrading repositories from older opam format ><><><><><><><><><><><><>  🐫
Upgrading repository "coq-released"...

<><> Synchronising development packages <><><><><><><><><><><><><><><><><><>  🐫
[coq-iris-builddep.dev] no changes from file:///Users/pgiarrusso/git/Coq/iris-coq/build-dep
[coq-stdpp.dev.2018-10-15.1.b7f1f6b1] no changes from git+https://gitlab.mpi-sws.org/robbertkrebbers/coq-stdpp.git#b7f1f6b178091ca6e3d363208d7875d73321903c
```

After undoing the migration via `cd ~/.opam/repo/iris-dev; git reset HEAD; git checkout .; git clean -fdx`, the update works again, *once*.

The failure appears to be when running `patch -p1` against a `-p0` patch generated by `/usr/local/bin/git "diff" "--no-ext-diff" "-R" "-p" "refs/remotes/opam-ref" "--"`.

Full output of `opam update -vv` (warning: 6M) here:
https://gist.github.com/Blaisorblade/24189e423677a61e524273d705091c6c. The issue is that `git diff` is indeed supposed to generate `-p1` patches.

Blaisorblade:
Fix #3627 by ensuring `git` produces a `-p1` diff. This is a very limited
fix but appears to work in my usecase. I have reviewed the other uses of
`git diff` in the same file, but it appears their output is not passed
to `patch` hence this change is unnecessary.

### DIFF
--- a/src/repository/opamGit.ml
+++ b/src/repository/opamGit.ml
@@ -33,6 +33,8 @@ module VCS : OpamVCS.VCS = struct
       git repo_root [ "init" ];
       (* Enforce this option, it can break our use of git if set *)
       git repo_root [ "config" ; "--local" ; "fetch.prune"; "false"];
+      (* We reset diff.noprefix to ensure we get a `-p1` patch and avoid <https://github.com/ocaml/opam/issues/3627>. *)
+      git repo_root [ "config" ; "--local" ; "diff.noprefix"; "false"];
       (* Document the remote for user-friendliness (we don't use it) *)
       git repo_root [ "remote"; "add"; "origin"; OpamUrl.base_url repo_url ];
     ] @@+ function
@@ -156,8 +158,7 @@ module VCS : OpamVCS.VCS = struct
     (* Git diff is to the working dir, but doesn't work properly for
        unregistered directories. *)
     OpamSystem.raise_on_process_error r;
-    (* We reset diff.noprefix to ensure we get a `-p1` patch and avoid <https://github.com/ocaml/opam/issues/3627>. *)
-    git repo_root ~stdout:patch_file [ "-c" ; "diff.noprefix=false" ; "diff" ; "--no-ext-diff" ; "-R" ; "-p" ; rref; "--" ]
+    git repo_root ~stdout:patch_file [ "diff" ; "--no-ext-diff" ; "-R" ; "-p" ; rref; "--" ]
     @@> fun r ->
     if not (OpamProcess.check_success_and_cleanup r) then
       (finalise ();
--- a/src/repository/opamGit.ml
+++ b/src/repository/opamGit.ml
@@ -158,7 +158,8 @@ module VCS : OpamVCS.VCS = struct
     (* Git diff is to the working dir, but doesn't work properly for
        unregistered directories. *)
     OpamSystem.raise_on_process_error r;
-    git repo_root ~stdout:patch_file [ "diff" ; "--no-ext-diff" ; "-R" ; "-p" ; rref; "--" ]
+    (* We also reset diff.noprefix here to handle already existing repo. *)
+    git repo_root ~stdout:patch_file [ "-c" ; "diff.noprefix=false" ; "diff" ; "--no-ext-diff" ; "-R" ; "-p" ; rref; "--" ]
     @@> fun r ->
     if not (OpamProcess.check_success_and_cleanup r) then
       (finalise ();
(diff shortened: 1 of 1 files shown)


rjbou:
Thanks for the fix!

Instead of putting the option in the diff command, it would be better to add it in the local config of the internal git repo [here](https://github.com/ocaml/opam/blob/master/src/repository/opamGit.ml#L30). 

It's [`OpamVCS.fetchs_repo_upgrade`](https://github.com/ocaml/opam/blob/master/src/repository/opamVCS.ml#L37)  which performs the first `init` and then uses the same directory to update it.

Blaisorblade:
> Instead of putting the option in the diff command, it would be better to add it in the local config of the internal git repo [here](https://github.com/ocaml/opam/blob/master/src/repository/opamGit.ml#L30).

Makes sense, done and testing locally.

Still, I will note if I just change `init`, it seems existing repos won't be fixed; I can remove and readd the repo, but that's slightly not ideal. Not that I'd insist on this.

Overall, `opam update` is slow enough here that I guess resetting the config after init wouldn't hurt? Or one could pass the config to each command, but adding wrappers to avoid code duplication.

rjbou:
> It seems existing repos won't be fixed

Indeed. We can have both: the local config updated, and just adding it to that git command. Other git diffs don't generate patches (`--quiet` option), so there is no effect on them.
Like that new init won't be affected, existing repo are handled, and with the comments, new added diffs will have the good option to set or no (the wrapper can be set  only if need be).

rjbou:
ping @Blaisorblade

Blaisorblade:
@rjbou Sorry for the delay — been busy! I am not sure what you are asking for — in particular I don't get:
> Like that new init won't be affected, existing repo are handled, and with the comments, new added diffs will have the good option to set or no (the wrapper can be set only if need be).

So, should we both set `diff.noprefix=false` both in the repo config and pass it to diff commands? Alternatively, maybe you can sketch which code you have in mind, and then I can revise/test it (or feel free to finish the PR yourself if that's easier).

rjbou:
What I wanted to say is that, with the comment on `git diff`command, if a new git diff command will be added in this file, developer will see the comment and don't forget it (or set a wrapper on that time to avoid duplication, and don't forget about `noprefix` option).
What I proposed is to merge your two addition (at init & at diff). If you are busy ftm, i can update your PR with this modification.

Blaisorblade:
> What I proposed is to merge your two addition (at init & at diff). If you are busy ftm, i can update your PR with this modification.

Please do, thanks!

rjbou:
Updated and rebased on master in the new pr #3788.
Thanks for the contrib!

Blaisorblade:
BTW, now I finally get what you meant — I thought you were talking about comments to the opam users (who are also developers)! Sorry for the confusion and thanks for taking it up :-)

PR #3637, opened 2018-10-25
Port build system to Dune (2.0 version)

dra27:
This is a version of #3618 for the 2.0 branch. It borrows a missing commit from #3474 to include 4.07.1 in the test matrix (was it intention not to include the bumping of the cold compiler from 4.06.0 to 4.07.1 - commit https://github.com/ocaml/opam/pull/3474/commits/3e51306e634294ed28de5f948c92ca4c43864c3e?)

The main thing this PR does is to remove the `configure`-changes, as those are more user-facing than is desirable for a maintenance branch. I've done this with a separate commit as this allows the possibility when cherry-pick things in future to revert the commit and then rebase, if that turns out to be simpler when back-porting other things.

GitHub's interface doesn't do too well at displaying the renames neatly, but if you checkout this branch and use patdiff to compare each jbuild file with the new dune file you'll find that the changes are minimal!

PR #3646, opened 2018-10-30
configure-dune: fix developer mode option

rjbou:
Make fails after a `configure` with `enable-developer-mode`:
```
$ make 
dune build --profile=release  opam-installer.install opam.install
        cppo src/core/opamCoreConfig.pp.ml (exit 2)
(cd _build/default && /.opam/rsw/bin/cppo '-D DEVELOPER' src/core/opamCoreConfig.ml) > _build/default/src/core/opamCoreConfig.pp.ml
/.opam/sw/bin/cppo: unknown option '-D DEVELOPER'.
```
Because of wuotes surrounding the cppo option `-D DEVELOPER`, expanded from dune `readlines` variables.
ping @dra27


dra27:
Aha, thanks - I'd spotted that this wasn't working, but hadn't had a chance to put the fix together. Should only affect master, as this change wasn't in the 2.0 of the pull request IIRC

rjbou:
updated without `-e` option (not portable).

@dra27 according https://github.com/ocaml/opam/pull/3637/commits/c138917cc276199a72a9dc0c151a15d7ad1c6d88#diff-e2d5a00791bce9a01f99bc6fd613a39dR3954 & https://github.com/ocaml/opam/pull/3637/commits/091f26a5ffdfa7d998af8d2d05457082028ba0e9#diff-e2d5a00791bce9a01f99bc6fd613a39dL3954, it is included

dra27:
@rjbou - the commit you point to is the one which _removed_ it from the 2.0 branch! When I did the Dune port for 2.0, I rebased the master patch and then added a commit (that one) which reverted changes which weren't wanted. The 2.0 has https://github.com/ocaml/opam/blob/2.0/src/core/opamCoreConfig.ml.in which is processed by `configure` instead of `cppo`

PR #3657, opened 2018-11-06
pinning: don't vcs-pin unversionned opam files

rjbou:
At the moment, when pinning a local package in a  vcs folder, opam automatically vcs-pin it, and takes into account unversioned opam files.
This PR changes this behavior by changing the pinning url: if the opam file is versioned, it is kept as a vcs-pin, otherwise it is transformed into a path-pin.
To restrict opam to pin only versioned opam files, the option `-k git` (or other vcs) can be used.

The main usefulness of this behavior is with recursive pinning, where it is applied to folders not just opam files.

### DIFF
--- a/src/client/opamAuxCommands.ml
+++ b/src/client/opamAuxCommands.ml
@@ -109,6 +109,40 @@ let opams_of_dir d =
         acc)
     [] files
 
+let opams_of_dir_w_target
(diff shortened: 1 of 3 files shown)


AltGr:
Closing because this will be part of #3499

PR #3661, opened 2018-11-06
Handle symlinks in bwrap

### LINKED ISSUE #3660: Sandboxing breaks gcc on centos
I'm having trouble with sandboxing on CentOS7.

During `opam init` as it tries to compile `ocaml-base-compiler`:

```
ocaml-base-compiler-4.07.0: gcc: error trying to exec 'cc1': execvp: No such file or directory
```

I got into a sandbox via:

```
bwrap --unshare-net --new-session --proc /proc --dev /dev \
  --bind /tmp /tmp --setenv TMPDIR /tmp --setenv TMP /tmp --setenv TEMPDIR /tmp \
  --setenv TEMP /tmp --tmpfs /run --tmpfs /var \
  --ro-bind /usr /usr --ro-bind /bin /bin --ro-bind /lib /lib --ro-bind /lib64 /lib64 \
  --ro-bind /etc /etc --ro-bind /opt /opt --ro-bind /home /home \
  bash
```

`gcc` is having trouble finding its sub-programs:

```sh
# Outside sandbox:
$ gcc -print-prog-name=cc1
/usr/libexec/gcc/x86_64-redhat-linux/4.8.5/cc1
# Inside sandbox:
$ gcc -print-prog-name=cc1
cc1
```

The search paths are weird:
```sh
# Outside sandbox:
$ gcc -print-search-dirs
install: /usr/lib/gcc/x86_64-redhat-linux/4.8.5/
programs: =/usr/libexec/gcc/x86_64-redhat-linux/4.8.5/:...lots more...
# Inside sandbox:
$ gcc -print-search-dirs
install: /bin/../lib/gcc/x86_64-redhat-linux/4.8.5/
programs: =/bin/../libexec/gcc/x86_64-redhat-linux/4.8.5/:/bin/../libexec/gcc/:...lots more...
```

I'm not sure why the search paths are different inside the sandbox, but... outside the sandbox, `/bin` is a symlink to `/usr/bin`, so `realpath /bin/../libexec` resolves to `/usr/libexec`. Inside the sandbox, `/bin` is a mount, so it resolves to `/libexec` which doesn't exist.

Changing `sandbox.sh` to do `--symlink usr/bin /bin` instead of `--ro-bind /bin /bin` fixes it, but only on platforms where `/bin` is a symlink. :/ And I don't know why gcc's search paths are different inside the sandbox; if they weren't, things would work.

GCC: `gcc-4.8.5-28.el7_5.1.x86_64` RPM
OS: `CentOS Linux release 7.5.1804` (per /etc/centos-release)
bubblewrap: 0.3.0 RPM from [here](https://centos.pkgs.org/7/epel-x86_64/bubblewrap-0.3.0-1.el7.x86_64.rpm.html)

```
# opam-version      2.0.1
# self-upgrade      no
# system            arch=x86_64 os=linux os-distribution=centos os-version=7
# jobs              23
# repositories      1 (http) (default repo at e5263e9b)
# pinned            0
# current-switch    default
```

mroch:
On some OSes, like CentOS and Arch Linux (so I've read), `/bin` is a symlink to `/usr/bin`, `/lib` is a symlink to `/usr/lib`, etc. 

Mounting them directly with `--ro-bind` resolves the symlinks. This breaks paths like `/bin/../libexec`: when `/bin` is a symlink to `/usr/bin`, `/bin/../libexec` points to `/usr/libexec`; but if it gets resolved first, it points to `/libexec`.

Fixes #3660

### DIFF
--- a/src/state/shellscripts/bwrap.sh
+++ b/src/state/shellscripts/bwrap.sh
@@ -17,15 +17,37 @@ ARGS=("${ARGS[@]}" --bind "${TMPDIR:-/tmp}" /tmp)
 ARGS=("${ARGS[@]}" --setenv TMPDIR /tmp --setenv TMP /tmp --setenv TEMPDIR /tmp --setenv TEMP /tmp)
 ARGS=("${ARGS[@]}" --tmpfs /run)
 
-add_mounts() {
+add_mount() {
     case "$1" in
         ro) B="--ro-bind";;
         rw) B="--bind";;
+        sym) B="--symlink";;
     esac
+    ARGS=("${ARGS[@]}" "$B" "$2" "$3")
+}
+
+add_mounts() {
+    local flag="$1"; shift
     for dir in "$@"; do
       if [ -d "$dir" ]; then
-        ARGS=("${ARGS[@]}" "$B" "$dir" "$dir")
+        add_mount "$flag" "$dir" "$dir"
       fi
+    done
+}
+
+# Mounts the standard system paths. Maintains symlinks, to handle cases
+# like `/bin` -> `usr/bin`, where `/bin/../foo` resolves to `/usr/foo`,
+# not `/foo`. We handle symlinks here but not in `add_mounts` because
+# system paths are pretty much guaranteed not to accidentally escape into
+# off-limits directories.
+add_sys_mounts() {
+    for dir in "$@"; do
+        if [ -L "$dir" ]; then
+            local src=$(readlink -f "$dir")
+            add_mount sym "$src" "$dir"
+        else
+            add_mounts ro "$dir"
+        fi
     done
 }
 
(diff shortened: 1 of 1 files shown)


rjbou:
Thanks for the contribution!
There is a case that is not handled: if resolved symlink is contained in mounted directories list. But I'm not even sure that it can happen on a usual state, and mounting them automatically could result to a breach. Environment variable `OPAM_USER_RO_PATH` can be used for that special case...

mroch:
yeah, i suppose that could happen but I think this might actually be a desirable behavior change in that case. previously, if `/bin` was a symlink to `/malicious`, `/malicious` would get mounted. now it'd be a broken symlink and you'd have to explicitly add `/malicious`.

but I also scoped it to just the system paths with the assumption that those are not likely to point unusual/dangerous places.

Issue #3662, opened 2018-11-07
Opam remove -a tries to remove ocaml-config and fails


NathanReb:
It seems the `--auto-remove` option of `opam remove` does not cope very well with the new `ocaml-config` package.

This is triggered by the ocaml-ci `.travis-docker.sh` script and was first reported here: https://github.com/ocaml/ocaml-ci-scripts/issues/265
The issue contains all the details as to how to reproduce the bug.

It also looks similar to this issue: https://github.com/ocaml/opam/issues/3461

I may not understand the `--auto-remove` option very well but it seems to me that if a package can't be removed because other packages depend on it, the command shouldn't fail and simply leave the package alone since it's then not meant to be auto-removed isn't it?

rjbou:
This is because of the switch state not updated when a base package have a new dependency. And as the autoremove relies on it to lock them...
Thanks for reporting!

rjbou:
Update of switch-state is one of the problem, even with it, opam should have kept `ocaml-config` in the package to keep. A bug on dependencies construction (fixed in #3670) didn't included it because of cycles on the graph (`ocaml-base-compiler` ⇄ `ocaml`).

PR #3666, opened 2018-11-07
Don't attempt to auto-remove base dependencies

dra27:
It is possible to end up with dependencies of base packages which are not themselves base packages. Don't allow these packages to be removed by `opam remove -a`.

This fix deals with the situation where a switch has been upgraded to include `ocaml-config`. `opam upgrade` should automatically put these packages into the base when they are recompiled (I believe @rjbou is working on that).

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -1158,9 +1158,23 @@ let remove_t ?ask ~autoremove ~force atoms t =
       OpamPackage.Set.of_list
         (OpamSolver.reverse_dependencies ~build:true ~post:true
            ~depopts:false ~installed:true universe packages) in
+    let base =
+      (* It is possible, either through unintended behaviour (fixed in PR#3667),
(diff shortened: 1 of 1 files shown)


rjbou:
Related to #3662 (issue) & #3667 (snd part of the resolution)

src/client/opamClient.ml
```diff
@@ -1158,9 +1158,12 @@ let remove_t ?ask ~autoremove ~force atoms t =
       OpamPackage.Set.of_list
         (OpamSolver.reverse_dependencies ~build:true ~post:true
            ~depopts:false ~installed:true universe packages) in
+    let base =
```
rjbou: Maybe adding a comment to explain why we don't take `universe.u_base` or `t.compilers_package`?

rjbou:
lgtm! thanks!

rjbou:
Inner problem fixed in #3666.

Issue #3659, opened 2018-11-06
Sandboxing on MacOS is too strict


silene:
On MacOS, `sandbox-exec` is called with the policy `(deny network*)`, which breaks programs that rely on unix sockets for inter-process communications. Here is an example of the needed permissions:
```
(allow network-bind (path "/private/var/tmp/rmk-GjXRh"))
(allow network-outbound (path "/private/var/tmp/rmk-GjXRh"))
```

Adding `(allow network* (remote unix))` after `(deny network*)` in `src/state/shellscripts/sandbox_exec.sh` seems to be sufficient and it should still forbid applications from actually accessing the network.

```
# opam config report
# opam-version      2.0.1 
# self-upgrade      no
# system            arch=x86_64 os=macos os-distribution=homebrew os-version=10.12.6
# solver            builtin-mccs+glpk
# install-criteria  -removed,-count[version-lag,request],-count[version-lag,changed],-changed
# upgrade-criteria  -removed,-count[version-lag,solution],-new
# jobs              1
# repositories      2 (http) (default repo at e5263e9b)
# pinned            0
# current-switch    default
```

avsm:
Do you need domain sockets for builds, or just for running tests?

silene:
It is the build system itself that uses unix sockets, so I need it for potentially everything.

gasche:
For context: the build system is [remake](https://github.com/silene/remake), created by @silene, as a mix of `make` and (dbj's/apenwarr's) `redo`, and it uses unix sockets for inter-process IPC.

@silene: my current understanding is that your proposal is fairly reasonable: I see no strong security need to prevent local sockets during build. Would you send an actual Pull Requet with your proposed sandboxing-script changes?

avsm:
This does look like it'll allow access to any domain sockets even outside of the build root.  From a security perspective, it might be nice to refine this to only allow the sockets to be dropped in the opamroot.  remake appears to support setting REMAKE_SOCKET to a suitable value.

My general worry is that we're widening the attack surface on osx a lot with this patch. There are an awful lot of services listening on domain sockets throughout the system...

silene:
> This does look like it'll allow access to any domain sockets even outside of the build root.

How so? The file corresponding to the socket still needs to be accessible.

> remake appears to support setting REMAKE_SOCKET to a suitable value.

[...]

Issue #3654, opened 2018-11-05
Check that `foo.opam.locked` is up-to-date


AltGr:
With the option `--locked`, it would be useful to have a warning if the base `opam` file is newer and/or has dependencies that the `opam.locked` file doesn't have.

It's easy to make changes to the `opam` file and forget to update the `opam.locked` :)

Issue #3652, opened 2018-11-03
opam does not respect `--no-checksums`


co-dan:
OPAM does not seem to respect the `--no-checksums` flag (or the $OPAMNOCHECKSUMS env variable).

I don't have enough knowledge to produce a minimal test case, but at this point in time the following triggers the incorrect behaviour for me:

```
opam install --no-checksums coq-stdpp.1.1.0 -vv
+ /home/dan/.opam/4.07.0/bin/ocamlc "-vnum"
- 4.07.0
The following actions will be performed:
  ∗ install   ocaml-config 1
  ↻ recompile ocaml        4.07.0
  ∗ install   ocamlfind    1.8.0             [required by coq]
  ∗ install   camlp5       7.06.10-g84ce6cc4 [required by coq]
  ∗ install   num          1.1               [required by coq]
  ∗ install   coq          8.8.2             [required by coq-stdpp]
  ∗ install   coq-stdpp    1.1.0
===== ∗ 6   ↻ 1 =====
Do you want to continue? [Y/n] y

<><> Gathering sources ><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
[camlp5.7.06.10-g84ce6cc4] found in cache
[coq.8.8.2] found in cache
+ /bin/tar "xfz" "/home/dan/.opam/download-cache/sha256/3a/3a9b90684a710b23b2f83f020304f0fda48219725d7a9ea881ae49effd9afdeb" "-C" "/tmp/opam-22016-ccabea"
[num.1.1] found in cache
+ /bin/tar "xfz" "/home/dan/.opam/download-cache/md5/71/710cbe18b144955687a03ebab439ff2b" "-C" "/tmp/opam-22016-c61e7b"
[ocamlfind.1.8.0] found in cache
+ /bin/tar "xfz" "/home/dan/.opam/download-cache/md5/5d/5d693cd1953a0dd74920b43d183bc26c" "-C" "/tmp/opam-22016-7afd23"
+ /bin/tar "xfz" "/home/dan/.opam/download-cache/md5/a7/a710c559667672077a93d34eb6a42e5b" "-C" "/tmp/opam-22016-a00279"
[...]

PR #3669, opened 2018-11-08
2.0.2: backported commits

rjbou:
Backported commit for 2.0.2 release.

| commit   | issue | descr |
| --- | --- | --- |
| 4d39333 |  | system: suffix .out for read_command_output stdout files |
| 38dd76d | #3622  | state config: resolve default opam root path |
| 4eb7be9 |  | filename: add concat and resolve operator |
| 8aff7ad |  | Fix some cases of listing coinstallable packages |
| 4952b36 |  | pin: add aux files on pinning 1.2 opam files |
| 60e5cfb |  | option: deprecated no-aspcud |
| 1f19a2b | #3654 | locked: check consistency with opam file when reading lock file to suggest regeneration message |
| 7e44f01 | #3508 | pin: upgrade pin depends on pinning |
| fe1f88f | #3390 | option: reinsert deprecated option alias-of & no-autoinstall to display deprecated message |
| 426e0f2 |  | arg: add a deprecated option message display function |
| 25c6267 |  | doc: update man pages |
| d002196 |  | show: remove pin depends messages |
| 0a6b63e | #3662 | Fix closure computation in the presence of cycles |
| 471d34c | #3660 | Handle symlinks in bwrap |
| 62753f8 | #3659 | Sandbox: allow access to unix sockets|
| 4f3d616 | #3652 | Do not check hash with --no-checksum on pull_upstream |
| f85ba9a |  | man: add OPAMNOCHECKSUMS doc |
| a5cb01c | #3660 | Sandbox: handle symlinks |
| 52bd2b9 | #3652 | Do not check hash with --no-checksum on pull_upstream |
| e9c701c | #3607 | Sandbox: change one-line conditional to if statement which was incompatible with set -e |
| 016252e | | Format Upgra#de: add function to upgrade opam file and its auxiliary files |
| 3997d9f | #3604 | Sandbox: make /var readonly instead of empty and rw |
| e0826ac | #3600 | Format upgrade: extract archived source files of version-pinned packages |
| 7d0a673 | | Core: add is_archive in OpamSystem and OpamFilename |
| bc86afd | #3614 | Init: don't fail if empty compiler given |
| 7521abb | #3625 | Lint: fix light_uninstall flag for error 52 |
| 58918d8 |  | Update cold compiler to 4.07.1 |


avsm:
Lgtm, except a question about upgrades: if the sandbox scripts are updated between opam versions (as they are here), are the hooks in an already-initialised opam installation also updated, or do users need to reinitialise their setup to get the latest versions?

rjbou:
An action is needed to be performed by users, an `opam init --reinit -n` . It is usually mentioned in the blog post announcement.

AltGr:
LGTM !

rjbou:
Last update!

avsm:
> It is usually mentioned in the blog post announcement.

Most users would probably never see this :-( It would be really nice to have something like @altgr's hashed hooked idea to alert users from the command line.  It's been a very common user report since the 2.0 release.

dra27:
> An idea to solve this could be to keep a static definition of all known by-default previous configs (or preferably just their hashes). Then when we detect an opam version bump, we can upgrade automatically states that match the known hashes, and just print a warning when we detect a modified or custom version (a little like Debian does for modified configuration files). This could apply both to the script file and to the hooks configuration, allowing for a `--disable` option in the latter case (assuming we want to provide such an option :/ )

This sounds brittle - especially once you start considering CRLF/LF joy and also third-party opams. For this to work, you need to know what the default configuration would have been when the root was initialised and what the actual configuration used was - why not just record both of those as actual files in the root (with no "actual" configuration file written if the built-in opam default was used). An upgrade of opam can then detect field-by-field whether the defaults have altered and take action accordingly:

- If the configuration was actually the default, then automatic upgrade seems reasonable
- If the specific field was the default value, then automatic upgrade is possible reasonable?
- If the specific field was different, then either a warning or an interactive prompt is wanted?

A similar mechanism could be used to warn about the need to propagate changes if the /etc/opamrc file gets updated.

avsm:
How about just unconditionally showing a `motd` with the changelog if the version of opam has changed?  That would just inform all users of the changes in the release.

dra27:
That would also be a good idea, apart from the unconditionally bit - it should may be display it only once or have the ability to suppress it once you've seen it... if you repeatedly show a message which someone knows will appear, then they stop paying attention completely and don't notice when it *changes*

PR #3670, opened 2018-11-08
Fix closure computation in the presence of cycles

AltGr:


rjbou:
When a cycle is present (typically `ocaml-base-compiler` ⇄ `ocaml`), relying on the successor to determine if a package should be marked gives an incomplete graph (that what caused the _non inclusion_  of `ocaml-config`in the package to not remove in #3662).

PR #3682, opened 2018-11-26
Allow environment variables to be used for system polling

toots:
This PR makes it possible to change the default values polled for the system running `opam`. This is particularly useful for [opam-cross-windows](https://github.com/ocaml-cross/opam-cross-windows) which makes use of depexts tagged for `mxe`. Previously, this was passed through the `--external=mxe` option but the value for that option has been removed with `opam 2.0`.

### DIFF
--- a/src/state/opamSysPoll.ml
+++ b/src/state/opamSysPoll.ml
@@ -35,14 +35,18 @@ let normalise_arch raw =
   | s -> s
 
 let arch_lazy = lazy (
-  let raw = match Sys.os_type with
-    | "Unix" | "Cygwin" -> OpamStd.Sys.uname "-m"
-    | "Win32" ->
-      if Sys.word_size = 32 && not (OpamStubs.isWoW64 ()) then
-        Some "i686"
-      else
-        Some "x86_64"
-    | _ -> None
+  let raw =
+    match OpamStd.Env.getopt "OPAM_ARCH" with
+      | Some arch -> Some arch
+      | None ->
+        (match Sys.os_type with
+        | "Unix" | "Cygwin" -> OpamStd.Sys.uname "-m"
+        | "Win32" ->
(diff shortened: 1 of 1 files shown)


toots:
Hi. Any opinion on this? I'm available to do more if needed, make it command line options etc. This would really be useful for the opam-cross-windows depext, which is currently broken with `opam 2.0`..

rjbou:
So sorry for the delay, thought already answered, but seems that my answer hasn't been posted :/

I was saying that there is no need for code modification, you can find same behaviour with option (they have changed with 2.0.0). Previous `--external` arguments is now `--vars`:
```
       --vars=[VAR=STR,...]
           Define the given variable bindings. Typically useful with
           --external to override the values for arch, os, os-distribution,
           os-version, os-family.
```
It is used as follow
```
$ opam show conf-pkg-config -f depexts:
["pkg-config"] {os-distribution = "debian"}
["pkg-config"] {os-distribution = "ubuntu"}
["pkg-config"] {os-distribution = "archlinux"}
["pkgconfig"] {os-distribution = "fedora"}
["pkgconfig"] {os-distribution = "centos"}
["pkgconfig"] {os-distribution = "mageia"}
["pkgconfig"] {os-distribution = "rhel"}
["pkgconfig"] {os-distribution = "oraclelinux"}
["pkgconfig"] {os-distribution = "alpine"}
["devel/pkgconf"] {os = "freebsd"}
["devel/pkgconf"] {os = "openbsd"}
["pkg-config"] {os = "macos" & os-distribution = "homebrew"}
["pkgconf"] {os = "freebsd"}

$ opam list --external conf-pkg-config --vars=os-distribution=debian
pkg-config

$ opam list --external conf-pkg-config --vars=os-distribution=fedora
pkgconfig

$ opam list conf-pkg-config --external --vars  os=openbsd,os-distribution=nothing
devel/pkgconf
```

toots:
Oh awesome, thanks!

PR #3698, opened 2018-12-13
Fix opam installer

Bronsa:
- the SHAs have been updated for 2.0.2 but version has been kept at 2.0.1, meaning the wrong release is being downloaded and the checksum fails
- on my system `uname -s` comes back as `Darwin`, not `darwin`

### DIFF
--- a/shell/install.sh
+++ b/shell/install.sh
@@ -5,7 +5,7 @@ set -ue
 # (c) Copyright Fabrice Le Fessant INRIA/OCamlPro 2013
 # (c) Copyright Louis Gesbert OCamlPro 2014-2017
 
-VERSION='2.0.1'
+VERSION='2.0.2'
 TAG=$(echo "$VERSION" | tr '~' '-')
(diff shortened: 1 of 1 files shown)


shell/install.sh
```diff
@@ -130,7 +130,7 @@ case "$ARCH" in
     *) ARCH=$(echo "$ARCH" | awk '{print tolower($0)}')
 esac
 
-OS=$(uname -s || echo unknown)
+OS=$(uname -s | awk '{print tolower($0)}' || echo unknown)
 case "$OS" in
     darwin|macos) OS=macos;;
     *) OS=$(echo "$OS" | awk '{print tolower($0)}')
```
dra27: This line is no longer necessary - indeed the whole `case` can become
```sh
if [ "$OS" = "darwin" ] ; then
  OS=macos
fi
```

Bronsa:
Applied the suggested changes

dra27:
LGTM, thanks!

PR #3699, opened 2018-12-13
Fix installer

Bronsa:
Very small followup to my previous PR: it looks like in certain shell implementations `$((` is not a valid token


dra27:
Ugh, yes indeed! This is why backticks are superior to `$()` :wink:

rjbou:
Thanks!

Issue #3700, opened 2018-12-13
opam 2.0.2 throwing weird errors during install phase


pmetzger:
I'm trying to upgrade MacPorts to install 2.0.2. Unfortunately, I run into this during the install phase. It happens over and over:

```
cmdliner error: Undefined variable $($b,\$OPAMBESTEFFORT) in "Don't fail if all requested packages can't be installed: try to install as many as possible. Note that not all external solvers may support this option (recent versions of $(i,aspcud) or $(i,mccs) should). This is equivalent to setting $($b,\\$OPAMBESTEFFORT) environment variable."
```

(BTW, if it's not obvious, I'm the MacPorts maintainer for OCaml, and what I'm trying to do here is upgrade the MacPorts provided package. I can't ask anyone else to help me, I'm supposedly the "expert".)

avsm:
Some more details would be useful about the build/install environment. Is MacPorts doing anything special in the install phase, such as setting some environment variables for fakeroot or similar? How was the binary compiled?

avsm:
It might be simpler to actually do the installation during the build phase, and just copy the files over in the install phase :-)

dra27:
@avsm - it wouldn't fix the problem, unfortunately! The issue is that some targets are unconditionally rebuilt (for development convenience)

avsm:
So what's different from the Homebrew package then, which works?  There's clearly a workaround for this issue or else all the package managers would see it...

```
    ENV.deparallelize

    system "./configure", "--prefix=#{prefix}", "--mandir=#{man}"
    system "make", "lib-ext"
    system "make"
    system "make", "man"
    system "make", "install"
```

dra27:
@avsm - assuming you’ve got 2.0.2 installed with homebrew, run `opam —help`, then quit the pager and see what appeared on stderr (which should be that same error message)

avsm:
Hm, oddly enough I can't repro this with a 2.0.2 binary on my laptop, but an upgrade on my desktop does happen with `opam --help` on Homebrew as well.  I'll chalk the laptop failed repro down to some developer/upgrade hitch...

pmetzger:
@dra27 Can you post the  patch needed to fix this? I can then have MacPorts apply it during the patch phase. Or, if 2.0.3 will be out within a few days, I can just wait for that.

dra27:
[...]

PR #3706, opened 2018-12-14
Fix typo in man pages

dra27:
Same as #3700, but on 2.0 branch and with a CI test for the future.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -33,15 +33,17 @@ all: man html pages
 man:
 	rm -rf man
 	mkdir -p man
-	$(OPAM) $(HELPFMT) > man/opam.1
+	$(OPAM) $(HELPFMT) > man/opam.1 2> man/err
 	for i in $(TOPICS); do\
-	  $(OPAM) $$i $(HELPFMT) > man/opam-$$i.1;\
+	  $(OPAM) $$i $(HELPFMT) > man/opam-$$i.1 2>> man/err ; \
 	done
-	$(OPAM) admin $(HELPFMT) > man/opam-admin.1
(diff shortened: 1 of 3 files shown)


dra27:
(assuming you're happy with it, 7fd531 needs to go on master too)

doc/Makefile
```diff
@@ -33,15 +33,17 @@ all: man html pages
...
 	done
-	$(OPAM) admin $(HELPFMT) > man/opam-admin.1
+	$(OPAM) admin $(HELPFMT) > man/opam-admin.1 2>> man/err
 	for i in $(TOPICS_ADMIN); do\
-	  $(OPAM) admin $$i $(HELPFMT) > man/opam-admin-$$i.1;\
+	  $(OPAM) admin $$i $(HELPFMT) > man/opam-admin-$$i.1 2>> man/err ; \
 	done
-	$(OPAM_INSTALLER) $(HELPFMT) > man/opam-installer.1
+	$(OPAM_INSTALLER) $(HELPFMT) > man/opam-installer.1 2>> man/err
+# Dune doesn't (yet) support --no-print-directory
```
avsm:
```suggestion
# TODO Dune will support --no-print-directory in 1.7.0 and higher
```

dra27: 🙂 At the time I pushed this branch, I had only just opened the PR in Dune and didn't want to presume until it was merged. This comment could be changed on master, but I don't think we want to constrain Dune to >= 1.7.0 on the 2.0 branch

avsm: Makes sense!

rjbou:
Closing the PR, 
- 626baf5 is 2.0.3 release
- 7fd8531 will be in 2.0.4 release (cf. referenced pr)
Thanks!

Issue #3671, opened 2018-11-12
On fish, opam doesn't set MANPATH, then complains it isn't set


talex5:
opam always warns me that my environment is incorrect:

```fish
~ $ opam sw --debug
00:00.002  GSTATE                 LOAD-GLOBAL-STATE @ /home/user/.opam
00:00.002  SWITCH                 list
#  switch   compiler                    description
   4.04.2   ocaml-base-compiler.4.04.2  4.04.2
   4.05.0   ocaml-base-compiler.4.05.0  4.05.0
→  4.07.0   ocaml-base-compiler.4.07.0  4.07.0
   default  ocaml-system.4.02.3         default
00:00.014  ENV                    Not up-to-date env variables: [MANPATH]

[WARNING] The environment is not in sync with the current switch.
          You should run: eval (opam env)
```

The log message indicates that the warning is because MANPATH isn't set correctly. However, the suggested command doesn't set MANPATH:

```fish
~ $ opam env --shell=fish
set -gx OPAM_SWITCH_PREFIX '/home/user/.opam/4.07.0';
set -gx CAML_LD_LIBRARY_PATH '/home/user/.opam/4.07.0/lib/stublibs:/home/user/.opam/4.07.0/lib/ocaml/stublibs:/home/user/.opam/4.07.0/lib/ocaml';
set -gx OCAML_TOPLEVEL_PATH '/home/user/.opam/4.07.0/lib/toplevel';
set -gx PATH '/home/user/.opam/4.07.0/bin' '/usr/local/go/bin' '/home/user/bin' '/usr/local/bin' '/usr/bin' '/bin' '/usr/local/games' '/usr/games';
```

This appears to be intentional, as `variables.fish` contains:

```fish
if [ (count $MANPATH) -gt 0 ]; set -gx MANPATH $MANPATH '/home/user/.opam/4.07.0/man'; end;
```

https://github.com/ocaml/opam/blob/c35eb977d19457d3f13b31c1aa60de4fb210865d/src/state/opamEnv.ml#L425

```fish
~ $ opam config report
# opam config report
# opam-version      2.0.1 
# self-upgrade      no
# system            arch=x86_64 os=linux os-distribution=debian os-version=9.5
# solver            builtin-mccs+glpk
# install-criteria  -removed,-count[version-lag,request],-count[version-lag,changed],-changed
# upgrade-criteria  -removed,-count[version-lag,solution],-new
# jobs              3
# repositories      1 (http) (default repo at efd53853)
[...]

Issue #3005, opened 2017-07-22
Fish shell variables.fish is breaking PATH and MANPATH


abc-mikey:
I've been following the instructions [here](https://github.com/realworldocaml/book/wiki/Installation-Instructions) for installing Ocaml, Opam and Utop.

I found that there is currently a problem with the environment setup script for opam for fish shell. 

_~/.opam/opam-init/variables.fish_
```
set -gx PATH "/home/michael/.opam/4.02.1/bin" "$PATH";
set -gx OCAML_TOPLEVEL_PATH "/home/michael/.opam/4.02.1/lib/toplevel";
set -gx PERL5LIB "/home/michael/.opam/4.02.1/lib/perl5:$PERL5LIB";
set -gx MANPATH "$MANPATH" "/home/michael/.opam/4.02.1/man";
set -gx CAML_LD_LIBRARY_PATH "/home/michael/.opam/4.02.1/lib/stublibs";
```

Here the quotes around `$PATH` on the first line are casing the existing path to be concatenated with spaces and treated as a single path. This actually completely breaks fish shell for all subsequent logins if you have added this script to your _~/.config/fish/config.fish_ profile. 

My guess is that there is the same problem for `$MANPATH` being quoted but I'm not seeing it as I don't have any existing man paths.

ksqsf:
Yes, I just experienced this.

Changing that line to
```fish
set PATH "/home/michael/.opam/4.02.1/bin" $PATH;
```
fixes this problem.

abc-mikey:
Yes that works, actually I just experienced another issue relating to this script getting sourced. 

The opam fish script is appending to / setting `MANPATH`. For me this is unset by default and fish has a check in `"/usr/share/fish/functions/man.fish` so that when MANPATH is unset it looks in the path returned by the command `manpath` which on my system correctly returns the various man paths on my system along with the ocaml path:

```
$manpath
/home/michael/.opam/4.02.1/man:/usr/local/share/man:/usr/share/man
```

However as this is being set by the opam script it is preventing this being run so that *only* the ocaml path is being included, *breaking all my other man pages*.

This can be fixed by commenting out the line or probably better adding a check to see if MANPATH is already set and only then adding the ocaml path to it.

AltGr:
You didn't report your opam version (nor `opam config report`). I'm assuming that it's 1.2.2, in which case this has been reported and fixed already.

[...]

PR #3728, opened 2019-01-23
fish unconditionally display manpath when fish version is 2.7 or later

gregory-nisbet:
Hello, I would like feedback on this change and advice on how to write a test for it. I briefly looked through the test directory and saw some scaffolding for "fake packages", but could not figure out what the right way to set up a "fake switch" is for testing purposes.

My system opam packaged by Manjaro (version `2.0.1`, sha256sum `2b76b39ed8d53923b81985b618d4ef13fca1eabb7110bb477c7f9a3e4ae01767`) does not produce a fish shell fragment for the `MANPATH` environment variable, causing `opam` to erroneously report that the environment is out of sync.

As a workaround, I have been using the following `fish` function. `setenv` is a valid fish command and the quoting strategy used by `OpamConfigEnv.print_csh_env` seems to work for fish if the paths contain sensible characters.

    function opam_env
      eval (opam config env --shell=csh)
    end

I removed a conditional and changed it so that `OpamConfigEnv.print_fish_env` always prints the `MANPATH`, since none of the other `print_*_env` functions appear to have dedicated logic for the `MANPATH` environment variable.

I tested it by `evaling` fragments generated by the development version of `opam` and checking for the presence or absence of warnings.

    ~/.../opam> setenv MANPATH '/'
    Exit 1

    ~/.../opam> opam switch
    #   switch   compiler                    description
        4.05.0   ocaml-base-compiler.4.05.0  4.05.0
    ->  default  ocaml-system.4.07.1         default

    [WARNING] The environment is not in sync with the current switch.
              You should run: eval (opam env)

    ~/.../opam> eval (~/.../opam/opam/_build/install/default/bin/opam config env --shell=fish)

    ~/.../opam  > opam switch
    #   switch   compiler                    description
        4.05.0   ocaml-base-compiler.4.05.0  4.05.0
    ->  default  ocaml-system.4.07.1         default

No tests failed as a result of this change. At least one test probably should have failed, though.

Thanks in advance,
Greg

### DIFF
--- a/src/client/opamConfigCommand.ml
+++ b/src/client/opamConfigCommand.ml
@@ -177,9 +177,9 @@ let rec print_fish_env env =
   (* set manpath if and only if fish version >= 2.6 *)
   let manpath_cmd v =
     OpamConsole.msg "%s" (
-      "awk -v f=$FISH_VERSION " ^
-      "'BEGIN { split(f, fs, \".\"); goodfish = fs[1] > 2 || (fs[1] == 2 && fs[2] >= 6); if (goodfish) { exit 0 } else { exit 1 } }' " ^
-      "; and "
+      (* test for existence of disown builtin, introduced in fish 2.7 *)
+      (* depends on `string' builtin, introduced in fish 2.3 *)
+      "builtin string match -q disown (builtin -n); and "
     ) ;
     set_arr_cmd "MANPATH" v in
   match env with
--- a/src/client/opamConfigCommand.ml
+++ b/src/client/opamConfigCommand.ml
@@ -174,6 +174,14 @@ let rec print_fish_env env =
               (OpamStd.Env.escape_single_quotes ~using_backslashes:true v))
          v)
   in
+  (* set manpath if and only if fish version >= 2.6 *)
+  let manpath_cmd v =
+    OpamConsole.msg "%s" (
+      "awk -v f=$FISH_VERSION " ^
+      "'BEGIN { split(f, fs, \".\"); goodfish = fs[1] > 2 || (fs[1] == 2 && fs[2] >= 6); if (goodfish) { exit 0 } else { exit 1 } }' " ^
+      "; and "
+    ) ;
+    set_arr_cmd "MANPATH" v in
   match env with
   | [] -> ()
   | (k, v, _) :: r ->
@@ -185,8 +193,7 @@ let rec print_fish_env env =
           * opamState.ml for details *)
          set_arr_cmd k v
        | "MANPATH" ->
-         if OpamStd.Env.getopt k <> None then
-           set_arr_cmd k v
+         manpath_cmd v
        | _ ->
          OpamConsole.msg "set -gx %s '%s';\n"
            k (OpamStd.Env.escape_single_quotes ~using_backslashes:true v));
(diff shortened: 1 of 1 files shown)


rjbou:
Thanks for the proposal!

This special case for `MANPATH` has been introducing because otherwise it broke `MANPATH` with old fish version, cf. #3005 and especially this [comment](https://github.com/ocaml/opam/issues/3005#issuecomment-320456270). 
I don't know if fish 2.6 is yet largely  used, from what I see, it is still the default on [debian stable](https://repology.org/metapackage/fish/versions).

On your proposal, changes will bring back to errored MANPATH setting (cf. 74f8bbd). Is there another way in fish to set `MANPATH` to avoid opam not up to date message and that don't break old fish versions?

> No tests failed as a result of this change. At least one test probably should have failed, though.

opam have an included test suite that tests only main operation (install, pin, update, upgrade, etc.). Shell features are not tested automatically, only via manual testing.

rjbou:
Related to #3671

gregory-nisbet:
@rjbou 

I changed it to use an `awk` fragment to check the fish version.
`fish` does not appear to have the ability to manipulate strings very well by itself. Please let me know if another solution (such as using `sh`) would be better than using `awk`.

In the course of trying to rebase this fix onto opam's master, I unintentionally pulled in some other commits. I am trying to fix it.

gregory-nisbet:
Hi everyone,

I looked at the fish changelog here: https://fishshell.com/release_notes.html and decided to test for the presence of the `disown` command (introduced in 2.7) using `string match` introduced in 2.3.

The command in structured in such a way that fish versions prior to 2.3 will attempt to execute a nonexistent `string` command and exit abnormally so the `MANPATH` won't be set.

Unfortunately, I can't redirect the stderr output of failing to execute a nonexistent command to `/dev/null`, for example:

    > builtin qqch 2>/dev/null 
    fish: Unknown builtin “qqch”
    builtin qqch 2>/dev/null
                ^

    > builtin qqch 1>/dev/null 2>/dev/null
    fish: Unknown builtin “qqch”
    builtin qqch 1>/dev/null 2>/dev/null
                ^

gregory-nisbet:
@rjbou @dra27 please let me know if there's anything else I should do.

rjbou:
Thanks for the changes!
From release notes, `disown` builtin command has been added in [2.6.0](https://github.com/fish-shell/fish-shell/blob/master/CHANGELOG.md#fish-26b1-released-may-14-2017), as `MANPATH` issue fixed in [2.7.0](https://github.com/fish-shell/fish-shell/issues/4158), maybe use another new builtin command to check, as [`argparse`](https://github.com/fish-shell/fish-shell/blob/master/CHANGELOG.md#fish-27b1-released-october-31-2017)?
Also, a proposal to not have fish 2.3.0 users have the error `string` message: use `grep` instead of `string match`. On my 2.4.0 fish (untested on other versions)
```fish
~> builtin -n | grep argparse > /dev/null 2>&1
~> echo $status
1
~> builtin -n | grep string > /dev/null 2>&1
~> echo $status
0
```
Having error messages (from shell) is an unwanted behaviour, even if it is sometimes just cosmetic, but in this case, when shell hook is enabled `opam env` would print an error message at each command prompt.
Note that I'm not a fish user :) maybe some other command than `argparse` or `grep` suits better, but it needs to not produce any superfluous messages.

gregory-nisbet:
@rjbou All right. We should be ready to go. I went with `argparse` and `grep`, but wrapped `grep` in a `/bin/sh` so it doesn't produce any stray output if `grep` doesn't exist.

rjbou:
Thanks! It's a go for me!

PR #3729, opened 2019-01-25
opam show: reintroduce file only information display

### LINKED ISSUE #3721: opam show on files is slow
```
> opam --version
2.0.2
> time opam show ./cmdliner.opam --normalise --no-lint --field=authors:,bug-reports:,depends:,dev-repo:,doc:,homepage:,license:,maintainer:,synopsis:,tags:,version: > /dev/null

real	0m0.223s
user	0m0.154s
sys	0m0.060s
```
I guess some stuff must be going on to enable evaluation with respect to the current repo/switch. But I'm really interested in the raw fields like `opam-query` gave us. Would it maybe be possible to add an option that skips unneeded setup ?

rjbou:
Reintroduce a more complete version of deprecated `--file` option with new option `--just-file`: permit to show opam files, their fields, without loading switch state.
Fixes #3721

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -685,10 +685,7 @@ let show =
     let doc =
       Arg.info
         ~docv:"FILE"
-        ~doc:"DEPRECATED: use an explicit path argument as package instead. \
-              Get package information from the given FILE instead of from \
-              known packages. This implies $(b,--raw) unless $(b,--fields) is \
-              used. Only raw opam-file fields can be queried."
+        ~doc:"DEPRECATED: see $(i,--just-file)"
         ["file"] in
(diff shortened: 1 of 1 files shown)


dbuenzli:
Thanks this looks like exactly what I need (that is IIUC I can still query multiple files in one invocation as before). 

Note however that the commit says `add --query option`, then it seems  the doc string of `--file` mentions it's deprecated in favour of `--file-only` but the option defined is called `--just-file` !

rjbou:
Oups, last minute name sed-changes...

I kept the same behaviour than normal `opam show` but for local file, you can query multiple files and directories. It don't error if non opam-file fields are queried, they are just empty.

PR #3742, opened 2019-02-01
sandbox: on osx, always rw mount /tmp

rjbou:
If `TMPDIR` is set to something else than `/tmp`, it is not mounted, which can be blocking for some programs.
Related to ocaml/opam-repository#13339

### DIFF
--- a/src/state/shellscripts/sandbox_exec.sh
+++ b/src/state/shellscripts/sandbox_exec.sh
@@ -13,11 +13,11 @@ add_mounts() {
     esac
 }
 
+# Even if TMPDIR is set, some applications uses /tmp directly
+add_mounts rw /tmp
+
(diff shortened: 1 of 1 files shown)


avsm:
This is ok by me, but I'm suspicious about the root cause of that lascar issue. I think it's worth tracking down to figure out if there's a worse bug lurking somewhere in the compiler...

rjbou:
Yes, of course, that's why I didn't set this PR as _fixing_ the lascar issue, only related

avsm:
Indeed, I see no harm in `/tmp` rw always, although keeping it as-is is useful for finding TMPDIR bugs I guess! We *could* set TMPDIR to something different in opam-repository CI to catch such bugs earlier.

Issue #3730, opened 2019-01-25
diff: opam update is painfully slow (2 minutes)


fpottier:
Hello -- I have been suffering this for a long time, but I just realized that maybe it is abnormal.

`opam update` is painfully slow: about two minutes on my mid-2015 MacBook Pro, an otherwise fast machine. (It is similarly slow on my desktop machine which runs Ubuntu.) Here is a time measurement:
```
real	1m59.466s
user	0m5.039s
sys	0m10.931s
```

My configuration:
```
# opam config report
# opam-version      2.0.2 
# self-upgrade      no
# system            arch=x86_64 os=macos os-distribution=macports os-version=10.11.6
# solver            builtin-mccs+glpk
# install-criteria  -removed,-count[version-lag,request],-count[version-lag,changed],-changed
# upgrade-criteria  -removed,-count[version-lag,solution],-new
# jobs              4
# repositories      2 (http) (default repo at 312209df)
# pinned            0
# current-switch    4.07.1
```
The two repositories are OCaml and Coq.
I have reinstalled aspcud 1.9.1_2 via MacPorts; this makes no difference.

I am attaching a [log.txt](https://github.com/ocaml/opam/files/2798076/log.txt) produced by `opam --debug --verbose`.

rjbou:
If we reassemble logs, we can see that it is diff operation takes a lot of time:
- 23s for coq repo
```
00:00.749  PARALLEL                        Starting job 0 (worker 1/3): 0
00:00.749  REPOSITORY                      update coq-released at /Users/fpottier/.opam/repo/coq-released from http://coq.inria.fr/opam/released
00:00.749  CURL                            pull-repo-update
00:00.749  SYSTEM                          mkdir /Users/fpottier/.opam/repo/coq-released.new
00:00.749  SYSTEM                          mkdir /var/folders/qv/429y8rgj4jd166x167cws9tw0000gp/T/opam-91602-5a61c5
00:00.767  PARALLEL                        Next task in job 0: /opt/local/bin/wget --content-disposition -t 3 -O /private/var/folders/qv/429y8rgj4jd166x167cws9tw0000gp/T/opam-91602-5a61c5/index.tar.gz.part http://coq.inria.fr/opam/released/index.tar.gz -U opam/2.0.2
Processing  1/2: [coq-released: http]
00:01.127  PARALLEL                        Collected task for job 0 (ret:0)
[...]

PR #3745, opened 2019-02-07
WIP Dirty optimisation to repository loading

AltGr:
Makes things *much* faster on older HDDs by keeping local repositories as
.tar.gz instead of thousands of scattered files. The implementation isn't very
nice at the moment though.

This only affects cases where the marshalled cache is not present, and
when the repository files are not yet in the OS disk cache.


rjbou:
Failing tests are from `repo_root` still set to the temporary directory when it tries to read it afterwards.

rjbou:
Related to #3730

Issue #3694, opened 2018-12-10
opam --locked does not correct the version of already-installed packages


jvillard:
Consider for instance [this opam.locked](https://github.com/facebook/infer/blob/ed82a0e572a453870c228e1b27eaa3927f9e38af/opam.locked) file. I would expect opam to downgrade javalib in this scenario (since the file specifies version 2.3.5) but it doesn't:

```console
$ opam install --deps-only --locked infer .
[starting in an empty switch, this installs all the dependencies at the correct version]
$ opam upgrade
...
  ↗ upgrade   javalib     2.3.5 to 3.0
  ↗ upgrade   sawja       1.5.4 to 1.5.6
...
$ opam install --deps-only --locked infer .
Nothing to do.
$ # javalib has not been downgraded to 2.3.5
$ opam show javalib --field=version 
3.0
```

AltGr:
This indeed appears to be a bug. What I think probably happens is that `--locked` is not correctly taken into account if the package is already pinned (which it was after the first `install` command).

We'll look into it, thanks for reporting!

AltGr:
To confirm, does the following workaround work for you ?
- before the second `opam install` command, run `opam unpin -n infer`

jvillard:
It reports `[NOTE] infer is not pinned.`.

rjbou:
Didn't noticed it at first look, but there is just a confusion about the `--locked` option: it can be set with or without argument, in this latter the argument is the suffix for the locked file (default `.locked`).
In your commands, you use the `--locked`, then name the package to install `infer`, but what is interpreted by opam is that you set the new lock file suffix as `infer`.
That's also why there is a different behaviour between cited `opam.locked` and master's one ; at the end, installed packages are only based on the `opam` file and as javalib `<3.0` restriction was added...
Just moving `--locked` to makes it with no argument really install locked packages (e.g. from [this state/commit](https://github.com/facebook/infer/commit/ed82a0e572a453870c228e1b27eaa3927f9e38af)):
```
$ opam install --deps-only infer  .  --locked
The following actions will be performed:
  ↘ downgrade javalib  3.0 to 2.3.5
  ↘ downgrade dune     1.6.2 to 1.5.1
  ↘ downgrade cmdliner 1.0.3 to 1.0.2
  ↘ downgrade sawja    1.5.6 to 1.5.4 [uses javalib]
  ↻ recompile jbuilder transition     [uses dune]
[...]

PR #3746, opened 2019-02-08
Add `lock` option

### LINKED ISSUE #3734: Integrate `opam lock` in opam
As lock mechanism is handled internally in opam for install, it is logical to integrate `opam-lock` in opam, as a new subcommand.
New additional _feature_: possibility to lock package from repository, and no more just local files.

### LINKED ISSUE #3769: opam-lock does not include dependencies marked as {dev}
If a dependency is marked with `{dev}` in `project.opam` then it is not carried over to `project.opam.locked` by `opam lock project.opam`.

opam 2.0.3 and opam-lock 0.2

rjbou:
Integrates lock option in opam. In order to do that, I've made also some needed changes:
- split `--locked` into `--locked` && `--lock-suffix`, no change on `OPAMLOCKED`
- add new `get_remote_url` on `OpamVCS`that retrieve repo remote url , and in case of hash (branch or commit) check that they are present in distant repo (used for `pin-depends`).

`lock` option is mainly the same than `opam-lock` plugin, except it is more integrated in opam uses: not only take files, but paths and packages, check for uninstalled dependencies, nonexistent packages.

Closes #3734, closes #3769 and related to #3694

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -3099,6 +3099,19 @@ let lock =
     `P "If paths (filename or directory) are given, those opam files are locked. \
         If package is given, installed one is locked, otherwise its latest version.";
     `P "Fails if all mandatory dependencies are not installed in the switch.";
+    `S "What is changed in the locked file?";
+    `P "- $(i,depends) are fixed to their specific versions, with all filters \
+        removed (except for the exceptions below";
+    `P "- $(i,depopts) that are installed in the current switch are turned into \
+        depends, with their version set. Others are set in the $(i,conflict) field";
+    `P "- `{dev}`, `{with-test}, and `{with-doc}` filters are kept if all \
+        packages of a specific filters are installed in the switch. Versions are \
+        fixed and the same filter is on all dependencies that are added from \
+        them";
+    `P "- $(i,pin-depends) are kept and new ones are added if in the \
+        dependencies some packages are pinned ";
+    `P "- pins are resolved: if a package is locally pinned, opam tries to get \
+        its remote url and branch, and sets this as the target URL";
     `S "ARGUMENTS";
     `S "OPTIONS";
   ]
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -3108,198 +3108,30 @@ let lock =
            "Only lock direct dependencies, rather than the whole dependency tree.")
   in
   let lock_suffix = OpamArg.lock_suffix "OPTIONS" in
-  let get_git_url url nv dir =
-    let module VCS =
-      (val OpamRepository.find_backend_by_kind url.OpamUrl.backend)
-    in
-    let open OpamProcess.Job.Op in
-    OpamProcess.Job.run @@
-    VCS.get_remote_url ?hash:url.OpamUrl.hash dir @@| function
-    | Some u ->
-      (if u.OpamUrl.hash = None then
-         OpamConsole.warning
-           "Referenced git branch for %s is not available in remote: %s, \
-           use default branch instead."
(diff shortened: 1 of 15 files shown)


rjbou:
Also closes #3769.
Added handling of `dev`, `with-test`, and `with-doc` variables.
For the moment, it's the default behaviour, maybe worth adding `--no-{dev,test,doc}` option to not have opam lock populated automatically.
To determine if they have to be added, we check the presence of _all_ `dev` packages, and if present, these packages and their dependencies (that also have a `dev` filter) are added in lock file. Same for `with-test` and `with-doc`.
In the same time, check for outdated opam lock file is released, only non dev/test/doc packages are checked (cf. AltGr/opam-lock/issues/11).

src/client/opamCommands.ml
```diff
@@ -3061,6 +3061,75 @@ let clean =
...
+        state of their installed dependencies, and outputs modified versions of \
+        the opam file with a $(i,.locked) suffix, where all the (transitive) \
+        dependencies and pinnings have been bound strictly to the currently \
+        installed version.";
+    `P "By using these locked opam files, it is then possible to recover the \
+        precise build environment that was setup when they were generated.";
+    `P "If paths (filename or directory) are given, those opam files are locked. \
+        If package is given, installed one is locked, otherwise its latest version.";
+    `P "Fails if all mandatory dependencies are not installed in the switch.";
+    `S "How is locked the opam file?";
```
AltGr: What is changed in the locked file?

src/client/opamCommands.ml
```diff
@@ -3061,6 +3061,75 @@ let clean =
...
+    `P "- $(i,depends) with no filter other than version constraint";
```
AltGr: "depends are fixed to their specific versions, with all filters removed (except for the exceptions below)"

src/client/opamCommands.ml
```diff
@@ -3061,6 +3061,75 @@ let clean =
...
+    `P "- $(i,depopts) that are installed in current switch, with their version
```
AltGr: "the" current switch +"are turned into depends"

src/client/opamCommands.ml
```diff
@@ -3061,6 +3061,75 @@ let clean =
...
+        set. Others are set in the $(i,conflict) field";
+    `P "- `{dev}`, `{with-test}, and `{with-doc}` filter are kept if all \
+        packages of a specific filters are installed in the switch. Version are \
+        fixed and the same filter is on all dependencies that are added from \
```
AltGr: +s

src/client/opamCommands.ml
```diff
@@ -3061,6 +3061,75 @@ let clean =
...
+        them";
+    `P "- $(i,pin-depends) are kept and new ones are added if in the \
+        dependencies some package are pinned ";
```
AltGr: package +s

src/client/opamCommands.ml
```diff
@@ -3061,6 +3061,75 @@ let clean =
...
+    `P "- pin are resolved: if a package is local pinned, opam tries to get \
```
AltGr: pin +s
s/local/locally
+ and sets this as the target URL.

src/core/opamSystem.ml
```diff
@@ -842,12 +842,18 @@ let rec flock_update
            Unix.(lockf fd F_ULOCK 0);
          Unix.lockf fd (unix_lock_op ~dontblock:true flag) 0
        with Unix.Unix_error (Unix.EAGAIN,_,_) ->
-         if dontblock then raise Locked;
+         if dontblock then
```
AltGr: part of an already merged PR, skip the commit!

PR #3752, opened 2019-02-13
Optimise repository loading

AltGr:
Repositories stored in `~/.opam/repo` contain thousand of tiny files
and directories, and can take a huge amount of time to load on HDDs,
networked or old filesystems. This is not visible in normal use
because there is a marshalled cache, but can cause `opam update` to be
extremely slow (it now needs to compute a diff on the files, which
requires re-loading the repository drom disk), or big lags when
changing your opam version (and the full file tree is not yet in the
OS memory cache).

The solution proposed here is extremely pragmatic, yet quite
efficient: we store the repository contents as `.tar.gz` files in
`~/.opam/repo` instead. Rather than resorting to a complex in-memory
structure, we just untar them to `/tmp` when they need to be read, and
re-tar them after modification (`opam update`, or format upgrade
only). Then we let the OS disk cache do the job: in normal operation,
the tree never needs to be flushed to disk, and loading the `.tar.gz`
is orders of magnitude faster than loading the individual files.

Note that this is done even for `rsync` or `git` repositories, which
is not particularly clean, but works. In the case of `git`, it would
be possible to just store a bare repository, and use `git` to extract
the individual files (this could even be done explicitely, directly to
memory, see how `Camelus` performs). But it does not seem worth the
additional implementation cost at the moment.

### DIFF
--- a/src/state/opamRepositoryState.ml
+++ b/src/state/opamRepositoryState.ml
@@ -127,31 +127,18 @@ module Cache = struct
 
 end
 
-let with_repo_root gt repo f =
+let get_repo_root gt repo repos_tmp =
   let uncompressed_root = OpamRepositoryPath.root gt.root repo.repo_name in
   let tar = OpamRepositoryPath.tar gt.root repo.repo_name in
   if OpamFilename.exists tar &&
      not (OpamFilename.exists_dir uncompressed_root) then
-    OpamFilename.with_tmp_dir @@ fun tmp ->
+    let open OpamFilename.Op in
     OpamFilename.extract_in
       (OpamRepositoryPath.tar gt.root repo.repo_name)
-      tmp;
-    f OpamFilename.Op.(tmp / OpamRepositoryName.to_string repo.repo_name)
+      (Lazy.force repos_tmp);
+    Lazy.force repos_tmp / OpamRepositoryName.to_string repo.repo_name
   else
-    f uncompressed_root
-
-let with_repo_root_job gt repo fjob =
-  let uncompressed_root = OpamRepositoryPath.root gt.root repo.repo_name in
-  let tar = OpamRepositoryPath.tar gt.root repo.repo_name in
-  if OpamFilename.exists tar &&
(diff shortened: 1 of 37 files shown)


AltGr:
Broken for packages which use `extra-files` at the moment

AltGr:
Note that, on existing opam installs, this will work transparently and tar-gzip the repo dirs on the next `opam update`

src/state/opamRepositoryState.ml
```diff
@@ -127,13 +127,13 @@ module Cache = struct
 
 end
 
-let load_repo_opams repo =
-  let t = OpamConsole.timer () in
+let load_opams_from_dir repo_name repo_root =
+  (* FIXME: why is this different from OpamPackage.list ? *)
```
rjbou: Just a bunch of several modification from https://github.com/ocaml/opam/commit/423bb84b76e5fa5e04110b3e1ea7018617f1eb69 to https://github.com/ocaml/opam/commit/eebaa69e4a9f9aae77d07ad5b698c6487e40e619 i'd say...

src/state/opamRepositoryState.mli
```diff
@@ -49,18 +49,45 @@ val build_index:
...
+ *     containing the extracted structure of the given repository, and cleans it up
+ *     afterwards if temporary. The basename of the directory is guaranteed to
+ *     match the repository name (this is important for e.g. [tar]) *\)
+ * val with_repo_root:
+ *   'a global_state -> repository -> (OpamFilename.Dir.t -> 'b) -> 'b
+ * 
+ * (\** As [with_repo_root], but on jobs *\)
+ * val with_repo_root_job:
+ *   'a global_state -> repository ->
+ *   (OpamFilename.Dir.t -> 'b OpamProcess.job) -> 'b OpamProcess.job *)
```
rjbou: remaining comments ?

rjbou:
LGTM! thanks!
Related to #3721

PR #3759, opened 2019-02-22
Update working dir

rjbou:
`working-dir` option didn't copy all local files, just uncomiited changes of version controlled ones, which leads to some rsync errors (if a file is deleted), and not handling new ones.
This PR changes this behavior by copying all current source directory (except `_build` directory). It may result to some build failure is current source directory is not clean.

### DIFF
--- a/src/repository/opamVCS.ml
+++ b/src/repository/opamVCS.ml
@@ -102,21 +102,11 @@ module Make (VCS: VCS) = struct
     match OpamUrl.local_dir repo_url with
     | None -> pull_url repo_root None repo_url
     | Some dir ->
-      VCS.versioned_files dir
-      @@+ fun files ->
       let files =
         List.map OpamFilename.(remove_prefix dir)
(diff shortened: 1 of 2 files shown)


src/repository/opamVCS.ml
```diff
@@ -125,9 +115,18 @@ module Make (VCS: VCS) = struct
...
+      let stdout_file =
+        let f = OpamSystem.temp_file "rsync-files" in
+        let fd = open_out f in
+        List.iter (fun s -> output_string fd s; output_char fd '\n') files;
+        close_out fd;
+        f
+      in
+      let args = [
+        "--files-from"; (Lazy.force convert_path) stdout_file;
+        "--exclude"; "_build"; (* tiny clean: exclude _build directory *)
```
avsm: I believe rsync supports `.gitignore` files, so a future improvement to this hardcoded _build directory could be adding `--filter=":- .gitignore"` to the command line.

PR #3776, opened 2019-03-10
`of_json`

gasche:
This branch adds `of_json` deserializers to a lot of internal OPAM datatypes. The idea is for `opam-lib` users to be able to exchange data safely (@Armael and I are playing with computing list of co-installable packages at the MirageOS retreat).

The second half of the pull-request is a set of `Crowbar` generators for the internal data structures, checking that the serializers/deserializers roundtrip correctly.

One thing not implemented in the PR is an `OpamJson.of_string` function to decode JSON. I guess that, for this, it would be nicer to add a dependency to an existing JSON parser. (Edit: actually we don't need a JSON decoder in OPAM, see below.)


gasche:
@Armael just pointed out that we don't actually know to add any JSON-decoding logic in OPAM for our use-case: users of opam-lib can just link to their favorite JSON library that uses the same polymorphic variant, and everything will work fine. Yay.

gasche:
(Thanks @yomimono for merging https://github.com/stedolan/crowbar/pull/47 which made my life easier.)

dbuenzli:
> One thing not implemented in the PR is an `OpamJson.of_string` function to decode JSON. I guess that, for this, it would be nicer to add a dependency to an existing JSON parser.

That was the case in the past, see https://github.com/ocaml/opam/commit/1663fe033bc99bc3c992fa0058859ac4fc3de0c7. 

It's better if tools like opam minimize their their dependencies otherwise we quickly run into bootstrapping issues; this both for the dev tools the dependencies themselves may want to use and sometimes things we don't see but downstream system packagers actually run into.

gasche:
I made some changes suggested by @rjbou : the `.opam` file for `opam-crowbar` is only in the `src/crowbar` subdirectory to make sure that people using `opam install .` at the root do not draw in the additional dependencies.

AltGr:
Great, thanks !

PR #3779, opened 2019-03-12
Set opam env variables once

gahr:
env_hook.csh uses precmd to set its environment variables, meaning they will be set every time the prompt is about to be displayed. I don't see anything in the variables that require them to be kept in sync with the shell state (e.g., cwd), so I think precmd is overkill. Isn't it enough to just set them at startup?

### DIFF
--- a/src/state/shellscripts/env_hook.csh
+++ b/src/state/shellscripts/env_hook.csh
@@ -1,1 +1,1 @@
-alias precmd 'eval `opam env --shell=csh --readonly`'
+eval `opam env --shell=csh --readonly`


AltGr:
Setting the variables at shell startup is already handled by updating your shell init files (done at `opam init` time, and only if you allow it). Indeed, `precmd` is not required for that workflow.

Opam 2 introduced "local switches", though, that actually get selected automatically depending on your CWD: this is where the shell hooks do come in handy. They're still opt-in, and you can remember to run `eval $(opam env)` when needed if you prefer.

PR #3781, opened 2019-03-12
Refactor the return types of OpamSolution.{apply,resolve_and_apply}

Armael:
The goal of this PR is to make `OpamSolution.apply` return a bit more information about what happened, in particular for the jobs that failed.

This is joint work with @gasche .


AltGr:
Looks good, thanks!

PR #3783, opened 2019-03-14
Introduce `drop` to replace `ignore (unlock ..)` for more lock-type-safety

gasche:
This PR, coming out of joint work with @Armael, can be summarized by the corresponding mli documentation:

```ocaml
(** Releases any locks on the given global state and then ignores it.
    Using [drop gt] is equivalent to [ignore (unlock gt)],
    and safer than other uses of [ignore]
    where it is not enforced by the type-system
    that the value is unlocked before it is lost.
*)
val drop: 'a global_state -> unit
```

The idea is applied to all three kinds of states-with-locks (global, switch, repos), and then `drop` is consistently used within the whole opam codebase, to remove `ignore` on possibly-still-locked locks as a code smell. See the commit message of the first commit for more design details.

Example from the patch:

```diff
- ignore (OpamSwitchState.unlock st);
- ignore (clear_switch gt switch)
+ OpamSwitchState.drop st;
+ OpamGlobalState.drop (clear_switch gt switch)
```

I didn't find any instance where locks were silently ignored incorrectly. Besides the fact that the resulting code is generally nicer, the point of porting the whole opam codebase is to encourage clueless users of opam-lib (us, for example) to follow the same style and avoid mistakes there.

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -373,7 +373,7 @@ let init =
         ?repo ~bypass_checks ~dot_profile
         ?update_config ?env_hook ?completion shell
     in
-    OpamStd.Exn.finally (fun () -> ignore (OpamRepositoryState.unlock rt))
+    OpamStd.Exn.finally (fun () -> OpamRepositoryState.drop rt)
     @@ fun () ->
     if no_compiler then () else
     match compiler with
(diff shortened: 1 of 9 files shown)


AltGr:
Indeed, this looks better, thanks!
The idea was to rely on the `with_` functions as much as possible, but in practice many use-cases ended up much more complex than that :)

PR #3786, opened 2019-03-18
JSON (de)serialization for OpamParallel graphs

gasche:
This PR provides JSON in/out functions for OpamParallel graphs, and in particular for the `OpamSolver.solution` type.

I needed to introduce new functions to the OpamParallel API, and to make the signatures less abstract (to reveal that the graph is unlabelled). See the commit messages for explanations.

(feature request from @Armael in the context of our marracheck work)


AltGr:
Thanks!

Issue #3727, opened 2019-01-23
opam does not update git branches


RalfJung:
On our CI, we (as of now) are using the following commands to update all installed packages and, if needed, install the package defined in `build-deps/` (that directory contains an `opam` file):
```
opam init --no-setup --disable-sandboxing --reinit -y
opam update
opam upgrade -y --all build-dep/
```
This used to work fine (not always with the same command sequence), but stopped reliably updating packages sometime around Dec 19th.  You can see this, for example, in [this CI log](https://gitlab.mpi-sws.org/FP/iris-coq/-/jobs/22121):
```
[prepare-opam] Done! Coq version we are going to use:
The Coq Proof Assistant, version 8.10+alpha (December 2018)
compiled on Dec 18 2018 15:23:34 with OCaml 4.07.0
version     dev
source-hash 4c733a92
```
That source-hash is outdated, there have been new commits in the branch that version "dev" points to since then.

On Dec 19th, we changed two things: we updated to opam 2.0.2, and we changed the command sequence from
```
opam update
opam upgrade -y --all build-dep/
```
to
```
opam init --no-setup --disable-sandboxing --reinit -y
opam upgrade -y --all build-dep/
```
I did the `opam init` change because the opam release notes said to do this after the update, and of course this being CI I cannot just interactively run this once, I have to make sure it runs each time an old cached opam root is used.

I assumed the changed command sequence might be the issue (`opam init` might not have *all* the effects of `opam update`), so I added `opam update --development`, but that doesn't seem to update git repos (which btw I think is a bug, the docs make it sound like `opam update --repositories && opam update --development` is the same as `opam update`).  So now I added a full `opam update`, redundantly updating the repositories, but it *still* does not update coq.

Another example is [this CI log](https://gitlab.mpi-sws.org/FP/LambdaRust-coq/-/jobs/21848) which shows a build failure because coq-stdpp was not properly updated. This was using the `opam init && opam upgrade` command sequence. coq-stdpp is pinned to a git branch, and at that point the repository already contained a new function `tc_to_bool`, but opam failed to update coq-stdpp and hence the build fails because that method is missing.  We had this failure for *all* our projects, and only fixed it (worked around it, really) by changing the URL of the pinned git repository -- *that* seems to trigger the right update.

```
# opam config report
# opam-version      2.0.2 
# self-upgrade      no
# system            arch=x86_64 os=linux os-distribution=debian os-version=9
# solver            builtin-mccs+glpk
[...]

PR #3792, opened 2019-03-21
Add OPAMDROPWORKINGDIR env variable

rjbou:
Mainly for CI.
It permit to not block update of packages previously installed with `--working-dir` (or that happen to have a dirty source tree).

Related to #3727 and more especially https://github.com/ocaml/opam/issues/3727#issuecomment-459016232.

/cc @RalfJung

### DIFF
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -272,6 +272,9 @@ let help_sections = [
   `P "$(i,OPAMNO) answer no to any question asked.";
   `P "$(i,OPAMNOASPCUD) Deprecated.";
   `P "$(i,OPAMNOCHECKSUMS) enables option --no-checksums when available.";
+  `P "$(i,OPAMDROPWORKINGDIR) overrides packages previously updated with \
+      $(b,--working-dir) on update. Without this variable set, opam would keep them \
+      unchanged unless explicitely named on the command-line.";
   `P "$(i,OPAMNOSELFUPGRADE) see option `--no-self-upgrade'.";
   `P "$(i,OPAMPINKINDAUTO) sets whether version control systems should be \
       detected when pinning to a local path. Enabled by default since 1.3.0.";
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -564,7 +564,9 @@ let update
          nondev_packages)
     in
     let dirty_dev_packages, dev_packages =
-      if names <> [] then OpamPackage.Set.empty, dev_packages else
+      if names <> [] || OpamClientConfig.(!r.drop_working_dir) then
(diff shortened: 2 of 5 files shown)


src/client/opamArg.ml
```diff
@@ -272,6 +272,8 @@ let help_sections = [
   `P "$(i,OPAMNO) answer no to any question asked.";
   `P "$(i,OPAMNOASPCUD) Deprecated.";
   `P "$(i,OPAMNOCHECKSUMS) enables option --no-checksums when available.";
+  `P "$(i,OPAMDROPWORKINGDIR) disable working-dir handling on update. \
```
AltGr: This isn't extremely clear. Maybe something like
```suggestion
  `P "$(i,OPAMDROPWORKINGDIR) overrides packages previously updated with $(b,--working-dir) on update. Without this variable set, opam would keep them unchanged unless explicitely named on the command-line."
```

src/client/opamClient.ml
```diff
@@ -564,7 +564,9 @@ let update
          nondev_packages)
     in
     let dirty_dev_packages, dev_packages =
-      if names <> [] then OpamPackage.Set.empty, dev_packages else
+      if names <> [] || OpamStd.Config.env_bool "DROPWORKINGDIR" = Some true then
```
AltGr: Please add an option to `OpamClientConfig.t` (or `OpamStateConfig.t`) ?

src/client/opamArg.ml
```diff
@@ -272,6 +272,8 @@ let help_sections = [
...
+      Don't clean source directory";
```
AltGr:
```suggestion
```

src/repository/opamGit.ml
```diff
@@ -202,7 +202,16 @@ module VCS : OpamVCS.VCS = struct
...
+    | { OpamProcess.r_code = 0; _ } ->
+      (git dir ["ls-files"; "--others"; "--exclude-standard"]
+       @@> function
+       | { OpamProcess.r_code = 0; OpamProcess.r_stdout = []; _ } ->
+         Done false
+       | { OpamProcess.r_code = 0; _ }
+       | { OpamProcess.r_code = 1; _ } as r ->
+         OpamProcess.cleanup ~force:true r; Done true
+       | r -> OpamSystem.process_error r
+      )
```
AltGr: Does this fix a bug ?

rjbou: dirty source check was done only by considering uncommitted changes, that was incomplete. Here, check is done giving untracked files also. It is complementary with #3759.

AltGr: :+1:

PR #3798, opened 2019-03-24
Include config variables of base packages in opam config report

dra27:
This is part of something else I've been working on, but seems to be a good feature in its own right, so I've separated it. This change includes the values of any configuration variables defined by the base/compiler packages of the switch in the output of `opam config report`. So you get output like this instead...

```
# opam config report
# opam-version         2.0.0 (aec142f0f32f027f2244993886f54ee363cdf35c) 
# self-upgrade         no
# system               arch=x86_64 os=linux os-distribution=ubuntu os-version=18.10
# install-criteria     -removed,-count[version-lag,request],-count[version-lag,changed],-changed
# upgrade-criteria     -removed,-count[version-lag,solution],-new
# jobs                 3
# repositories         1 (http), 2 (version-controlled) (default repo at 26bdd6eb)
# pinned               0
# current-switch       default
# ocaml:native         true
# ocaml:native-tools   true
# ocaml:native-dynlink true
# ocaml:stubsdir       /usr/local/lib/ocaml/stublibs:/usr/local/lib/ocaml
# ocaml:preinstalled   true
# ocaml:compiler       system
# ocaml-system:path    /usr/local/bin
```

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -961,7 +961,7 @@ let config =
        | [file] -> let oc = open_out file in dump oc; close_out oc; `Ok ()
        | _ -> bad_subcommand commands ("config", command, params))
     | Some `report, [] -> (
-        let print label fmt = OpamConsole.msg ("# %-17s "^^fmt^^"\n") label in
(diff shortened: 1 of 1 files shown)


AltGr:
Seems useful, thanks!

PR #3799, opened 2019-03-24
Change OpamStateTypes.switch_state.conf_files from package_map to name_map

dra27:
Currently, the API for accessing `.config` files is a little awkward - only one `.config` file may be installed, and they physically never refer to version numbers, yet the API uses `OpamPackage.t` with slightly awkward calls to `OpamPackage.package_of_name` (awkward because it forces you to query the set of installed packages in order to get a fake version of the package to look up a file from a map which was keyed on name only, not name+version!).

This PR changes the type of `conf_files` to be a `name_map`. The semantics of `OpamSwitchState.package_config` are subtly different in that a `.config` file can be loaded for a package which is not installed - that at present would indicate a "corrupt" OPAMROOT but happens to be rather useful for something I'm working on...

### DIFF
--- a/src/client/opamSolution.ml
+++ b/src/client/opamSolution.ml
@@ -403,8 +403,8 @@ let parallel_apply t ~requested ?add_roots ~assume_built action_graph =
     let t =
       { !t_ref with
         installed = visible_installed;
-        conf_files = OpamPackage.Map.filter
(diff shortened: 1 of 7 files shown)


AltGr:
Seems cleaner indeed, thanks!

PR #3800, opened 2019-03-25
Fix pinnings not installing/upgrading already pinned packages

AltGr:


AltGr:
Travis failure unrelated with the patch
(we do `opam install lwt` as a basic test, and `lwt` now depends on `seq`, which is apparently broken on 4.02.3 — @kit-ty-kate  ?)

AltGr:
@rjbou, is 2.0.3 affected bu this regression ? If so, this would probably deserve a 2.0.4 ?

rjbou:
On travis failure, it's from the image, it have an ocaml 4.02.3 installed, but not with `ocamlbuild`, and seq depends on `ocamlbuild.0` that depend on `base-ocamlbuid`.

>  this would probably deserve a 2.0.4 ?

it's contained in 2.0.2, so i'd say yes. We can do a 2.0.4 fixes release.

rjbou:
@altgr fixed, just need to rebase

AltGr:
Rebased (twice), CI build still blocked with the ocamlbuild/seq issue on 4.02.3... :/
@rjbou an idea ?

rjbou:
I suppose that the cache contains an ocaml version without ocamlbuild. I deleted it and relaunched, let's see

rjbou:
I was the cache, 4.02.3 build succeed.

AltGr:
Thanks!

rjbou:
related to #3727

PR #3802, opened 2019-03-26
Use a character that displays better on terminals for download actions

AltGr:


dra27:
This changes from a character which isn't available in `Consolas` (the default Windows 10 terminal font) to one which is 👍

PR #3803, opened 2019-03-26
Proposal: refresh the Unicode symbols we use for actions

AltGr:
the new ones should display better on standard terminals

Not sure it looks better though:

▲ upgrade
▼ downgrade
◆ install
◇ remove
◈ reinstall
⬇ download
λ build

(current set:
↗ upgrade
↘ downgrade
∗ install
⊘ remove
↻ reinstall
▼ download
λ build
)

### DIFF
--- a/src/core/opamConsole.ml
+++ b/src/core/opamConsole.ml
@@ -98,6 +98,11 @@ module Symbols = struct
   let downwards_arrow = Uchar.of_int 0x2193
   let up_down_arrow = Uchar.of_int 0x2195
   let downwards_double_arrow = Uchar.of_int 0x21d3
+  let black_down_pointing_triangle = Uchar.of_int 0x25bc
+  let black_up_pointing_triangle = Uchar.of_int 0x25b2
+  let white_diamond = Uchar.of_int 0x25c7
+  let black_diamond = Uchar.of_int 0x25c6
+  let white_diamond_containing_black_small_diamond = Uchar.of_int 0x25c8
   let downwards_black_arrow = Uchar.of_int 0x2b07
 end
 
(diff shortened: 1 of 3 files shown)


dra27:
If we change these, can we pick ones which are available on the Windows Terminal (I haven’t yet checked...)

dra27:
Actually, I just looked into all this again (the last time I checked all this, Windows 10 was still in beta...). I "discovered" (by which I mean actually installed) https://github.com/dejavu-fonts/dejavu-fonts which has all of these symbols (and all the old ones...)

So, I think recommending that console font is probably the way to go, and just make sure that the default Windows Console font has options listed in the alternates (which it already does).

avsm:
I think the existing symbols for reinstall and remove are clearer than the newer proposed ones.  I do like the install diamond _vs_ the existing star though.  I'm not sure how to evaluate them beyond subjectivity though -- if the existing ones don't render on some terminals then they clearly need to be improved :-)

dbuenzli:
I think I agree with @avsm, I find some of the new symbols a bit cryptic. Though I always found the forbidden something (if I still have my road sign reading abilities) equated with "remove" to be misleading.

Just for the fun and the subjectivity I tried to look for a purely arrow based set drawn from the same block of characters so that a font might succeed in having them designed in a cohesive manner.  I failed on build though which you may find a bit pessimistic. But in any case that would be my proposal 

↗ upgrade
↘ downgrade
→ install
← remove
↻ reinstall
↓ download
💥build

directions on the right always means something you will have, on the left means gone; and hence note the sign for reinstall. 

Now this is all very culturally centered, luckily for right-to-left scripts we can change that to:

↖ upgrade
↙ downgrade
← install
→ remove
↺ reinstall
↓ download
💥build

for top-to-bottom I tried a 90° rotation but my brain can't get used to it, I'll let someone in the know figure out.

AltGr:
Indeed, I like the new up/down grade, but not sure about the other ones.
On the older selection, maybe an "X mark" ❌would be more obvious for removals.

Note that this is intended to improve readability, but is quite secondary:
- colors also give information
- wherever it is important, there is always a printed word to explicit the action
- (build actions only need to be printed on failure, they are normally hidden)

Example:
Spectacle bw1048

PR #3812, opened 2019-04-03
Some more Opam*State.drop use

rjbou:


AltGr:
Good cleanup, thanks!
I don't think that `drop` should be used for the cases that are already enclosed in `with_`, though, it seems redundant.
To be perfectly type consistent, the function passed to `with_` could actually return a state, although it's actually not needed: that would avoid these ignored values. But I remember I tried to have it written like that at some point, and it ended up adding more noise than it removed.

rjbou:
Indeed, it ~is~ was  a `sed`dy pr :)
fixed, with another drop with an unlock

src/client/opamClient.ml
```diff
@@ -637,6 +637,7 @@ let update
           (OpamPackage.Set.to_json updates);
       (success, not (OpamPackage.Set.is_empty updates)), st
   in
+  OpamSwitchState.drop _st;
```
AltGr: Please remove the leading `_` if actually using the variable, `_foo` should be reserved for documentation of unused variables.

src/client/opamCommands.ml
```diff
@@ -2169,6 +2166,7 @@ let switch =
           ~local_compiler
           switch
       in
+      OpamGlobalState.drop _gt;
```
AltGr: Leading `_`, see above

src/client/opamConfigCommand.ml
```diff
@@ -259,10 +259,10 @@ let env gt switch ?(set_opamroot=false) ?(set_opamswitch=false)
        in
        log "Missing environment file, regenerates it";
        if not (OpamCoreConfig.(!r.safe_mode)) then
-         (let _st =
+         (let _, _st =
             OpamSwitchState.with_write_lock st @@ fun _st ->
             (OpamFile.Environment.write env_file upd), _st
-          in ());
+          in OpamSwitchState.drop _st);
```
AltGr: Leading `_`, see above

src/client/opamClient.ml
rjbou: I kept the `_` because they are not really used, just kept to be dropped. Updated!

PR #3813, opened 2019-04-03
init defaults: add `getconf` to required tools 

rjbou:
except for win32 & freebsd (cf. [`cpu_count`](https://github.com/ocaml/opam/blob/master/src/core/opamSystem.ml#L563)), to avoid errors like
```
$ opam --version
2.0.3
$ opam init
[NOTE] Will configure from built-in defaults.
Fatal error:                                           "getconf": command not found.        
```

@dra27, does it need a special handling for cygwin?


dra27:
Sorry for the lag - this should be fine for Cygwin, which has `getconf`

PR #3816, opened 2019-04-04
Fix auto upgrade of `ocaml-system`

### LINKED ISSUE #3708: opam system compiler changing fails to upgrade
With opam 2.0.2, I did a `brew upgrade` which upgraded the system compiler from 4.06.0 to 4.07.1.  The resulting system fails to upgrade however:

```
$ opam upgrade
[WARNING] File /usr/local/bin/ocamlc, which package ocaml-system.4.06.1 depends upon, was changed on your system. ocaml-system has been marked as
          removed, and will be reinstalled if necessary.
[WARNING] Upgrade is not possible because of conflicts or packages that are no longer available:
The following dependencies are the cause:
  - ocaml → ocaml-base-compiler = 4.06.1
      conflict with the base packages of this switch
  - ocaml → ocaml-system = 4.06.1
      unmet availability conditions: sys-ocaml-version = "4.06.1"
  - ocaml → ocaml-variants >= 4.06.1
      conflict with the base packages of this switch

You may run "opam upgrade --fixup" to let opam fix the current state.

The former state can be restored with:
    opam switch import "/Users/avsm/.opam/default/.opam-switch/backup/state-20181216091405.export"
yak:opam-repository avsm$ opam upgrade --fixup
[WARNING] File /usr/local/bin/ocamlc, which package ocaml-system.4.06.1 depends upon, was changed on your system. ocaml-system has been marked as
          removed, and will be reinstalled if necessary.
[ERROR] It appears that the base packages for this switch are no longer available. Either fix their prerequisites or change them through 'opam list
        --base' and 'opam switch set-base'.
  - ocaml → ocaml-base-compiler = 4.06.1
      conflict with the base packages of this switch
  - ocaml → ocaml-system = 4.06.1
      unmet availability conditions: sys-ocaml-version = "4.06.1"
  - ocaml → ocaml-variants >= 4.06.1
      conflict with the base packages of this switch

No solution found, exiting

The former state can be restored with:
    opam switch import "/Users/avsm/.opam/default/.opam-switch/backup/state-20181216091413.export"
yak:opam-repository avsm$ ocaml -version
The OCaml toplevel, version 4.07.1
```

rjbou:
System compiler upgrade (closes #3708) using `opam upgrade --fixup`: base packages are unlocked automatically in order to update compiler packages to current system installed, or an `ocaml-variant` if no `ocaml-system` is found.

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -442,7 +442,31 @@ let upgrade_t
 let upgrade t ?check ~all names =
   let atoms = OpamSolution.sanitize_atom_list t names in
   let t = update_dev_packages_t atoms t in
-  upgrade_t ?check ~strict_upgrade:(not all) ~all atoms t
+  let unlocked_base, t =
+    let base_to_update = t.compiler_packages %% t.remove in
+    if OpamPackage.Set.is_empty base_to_update then
+      None, t
+    else
+      (log "unlocking base packages to fix  %s"
+         (OpamPackage.Set.to_string base_to_update);
+       Some t.compiler_packages,
+       { t with compiler_packages = OpamPackage.Set.empty })
+  in
+  let t = upgrade_t ?check ~strict_upgrade:(not all) ~all atoms t in
+  match unlocked_base with
+  | None -> t
+  | Some unlocked ->
+    (* we put back updated base package *)
+    let compiler_packages =
+      t.compiler_packages ++
+      (OpamPackage.Set.map
+         (fun p ->  OpamSwitchState.get_package t (OpamPackage.name p))
+         unlocked)
+    in
+    let diff = compiler_packages -- unlocked in
+    if not (OpamPackage.Set.is_empty diff) then
+      log "new base packages are %s" (OpamPackage.Set.to_string diff);
+    { t with compiler_packages }
 
 let fixup t =
   log "FIXUP";
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -447,7 +447,7 @@ let upgrade t ?check ~all names =
     if OpamPackage.Set.is_empty base_to_update then
       None, t
     else
-      (log "unlocking base packages to fix  %s"
+      (log "unlocking base packages to fix %s"
          (OpamPackage.Set.to_string base_to_update);
        Some t.compiler_packages,
        { t with compiler_packages = OpamPackage.Set.empty })
--- a/src/state/opamStateTypes.mli
+++ b/src/state/opamStateTypes.mli
@@ -142,7 +142,7 @@ type +'lock switch_state = {
   (** The set of packages which needs to be reinstalled *)
 
   remove: package_set;
-  (** The set of packages which need to be removed *)
+  (** The set of packages which needs to be removed *)
 
   (* Missing: a cache for
      - switch-global and package variables
--- a/src/state/opamStateTypes.mli
+++ b/src/state/opamStateTypes.mli
@@ -141,6 +141,9 @@ type +'lock switch_state = {
   reinstall: package_set;
   (** The set of packages which needs to be reinstalled *)
 
+  remove: package_set;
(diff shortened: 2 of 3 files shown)


dra27:
This doesn't feel like the right user-approach for this. It's odd (in a psychological sense) that having installed a shiny new OCaml system compiler, we must take the view that the opam switch is broken and requires "fixing"! I've in the past discussed an adaptation where the dependencies of base packages which are also roots are not considered fixed - so, for example, if the `ocaml-config` package is updated, it would be nice to have a switch automatically recompile because that package has changed, but not allow the solver to upgrade `ocaml`, `ocaml-system`, `ocaml-base-compiler`, etc. without express permission.

The use of the roots has an interesting consequence, because it depends on whether you did `opam switch create foo ocaml` or `opam switch create foo ocaml-system`. In the latter, I don't think the default should ever select a base compiler - if you actually remove the system compiler, I think it OK for you to expect to do something manual.

rjbou:
Indeed, worth adding it in the simple upgrade process. Updated.

src/client/opamClient.ml
```diff
@@ -442,7 +442,31 @@ let upgrade_t
 let upgrade t ?check ~all names =
   let atoms = OpamSolution.sanitize_atom_list t names in
   let t = update_dev_packages_t atoms t in
-  upgrade_t ?check ~strict_upgrade:(not all) ~all atoms t
+  let unlocked_base, t =
+    let base_to_update = t.compiler_packages %% t.remove in
+    if OpamPackage.Set.is_empty base_to_update then
+      None, t
+    else
+      (log "unlocking base packages to fix  %s"
```
avsm:
```suggestion
      (log "unlocking base packages to fix %s"
```

src/state/opamStateTypes.mli
```diff
@@ -141,6 +141,9 @@ type +'lock switch_state = {
   reinstall: package_set;
   (** The set of packages which needs to be reinstalled *)
 
+  remove: package_set;
+  (** The set of packages which need to be removed *)
```
avsm:
```suggestion
  (** The set of packages which needs to be removed *)
```

(just to match the grammar style of the entry above)

rjbou:
@avsm thanks!

AltGr:
I think this is the right approach, but doesn't go far enough:
- it is only applied on `opam upgrade`, which I think is a mistake: all the commands eventually calling `OpamSolution.apply` should go through this code (e.g. if I `opam install foo` with a removed system compiler, the handling is needed)
- the compiler is removed from the base packages altogether, letting the solver choose the new one. While the compiler should remain a "root" package, which could maybe help, I think we can do better with a specific heuristic:
  * for each base package that has been marked, if there is at least one version available, mark as `wish_install`
  * otherwise, we could either abort altogether with an error
  * or issue a warning, and try to remove the constraints altogether (like proposed atm). The solver by default tries to reduce removals first and foremost, so someone having a given OCaml system version will probably get switched to a similar compiled version if any packages required it.

  No need to ask for confirmation here, since the operations will be validated by the user anyway.

Anyway, we should probably settle on the more general case (#3817) and come back to this afterwards, since the former might help or make the situation clearer.

AltGr:
Note also that `opamSolver.ml` has this piece of code already, which may change the above:
```
  let keep_map =
    OpamPackage.Set.fold (fun nv acc ->
        if OpamPackage.Set.mem nv universe.u_available
        then OpamPackage.Map.add nv `Keep_version acc
        else if OpamPackage.has_name universe.u_available nv.name
        then OpamPackage.Map.add nv `Keep_package acc
        else acc)
      (packages %% universe.u_base) OpamPackage.Map.empty
  in
```

So if a `base` package is not available at this point, but has other versions available, just mark it as forced install, withdrawing its version constraint. If not available at any version, forget the constraint already. This was intended to allow system compiler upgrades, but the process got blocked before this could get called I think.

rjbou:
Superseded by #3894

PR #3814, opened 2019-04-03
Remove & reinstall of system wide changed packages

rjbou:
If a package depends on a system file, its hash is stored in the config file, it permits to detect if the files changed or has been removed. In that case, it was only removed from installed packages in order to be reinstalled.
This PR propose a deeper management of these packages modfication, by 
- adding them automatically to the reinstall packages list (even if nothing is needed to be installed)
- changing their action graph: we need the package to be removed in any case, even if the build fails.
Current action graph precedence adds remove action after the build, it is changed for these packages and their dependences

possible side effect: new `ocaml-system` is not installable, opam deleted all your packages in the current switch.

PR #3817, opened 2019-04-04
Forget about system wide changed packages and its depends

rjbou:
If a package depends on a system file, its hash is stored in the config file, it permits to detect if the files changed or has been removed. In that case, it was only removed from installed packages in order to be reinstalled.
This PR propose another solution than #3814: instead of forcing re move & reinstall, it just forbid you to use these package unless system files are fixed.

### DIFF
--- a/src/client/opamSolution.ml
+++ b/src/client/opamSolution.ml
@@ -943,13 +943,67 @@ let resolve t action ~orphans ?reinstall ~requested request =
       (`A (List.map (fun s -> `String s) (Array.to_list Sys.argv)));
     OpamJson.append "switch" (OpamSwitch.to_json t.switch)
   );
-  Json.output_request request action;
-  let r =
-    OpamSolver.resolve
-      (OpamSwitchState.universe t ~requested ?reinstall action)
-      ~orphans
-      request
+  let universe =
+    OpamSwitchState.universe t ~requested ?reinstall action
   in
+  let request =
+    let remove =
+      OpamPackage.Set.filter (fun pkg ->
+          let n = OpamPackage.name pkg in
+          let f (n',_) = OpamPackage.Name.compare n n' = 0 in
+          not ((List.exists f request.wish_install)
+               || (List.exists f request.wish_upgrade)))
+        t.remove
+    in
+    if OpamPackage.Set.is_empty remove || action = Remove then request
+    else
+    let forbidden =
+      OpamSolver.reverse_dependencies ~depopts:true ~build:true ~post:true
+        ~installed:false universe remove
+      |> OpamPackage.Set.of_list
+      |> (fun f -> OpamPackage.Set.diff f t.remove)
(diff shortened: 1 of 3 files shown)


src/client/opamSolution.ml
```diff
@@ -943,13 +943,67 @@ let resolve t action ~orphans ?reinstall ~requested request =
...
+    else
+      (let lst =
+         OpamStd.List.concat_map ~left:"" ~right:"" ~last_sep:"and" ", "
+           (fun n ->
+              OpamConsole.colorise `underline (OpamPackage.Name.to_string n))
+       in
+       let sgl_rem = OpamPackage.Set.cardinal t.remove = 1 in
+       let sgl_op = List.length (forbid_inst @ forbid_upg) = 1 in
+       OpamConsole.warning
+         "Packages %s depends on a changed or missing system dependency. \
```
AltGr: Since we change what the user actually asked for, we should ask for a confirmation here.
What happens if the user request becomes empty ?

rjbou:
> I am not sure handling this at the request level is sufficient

We can handle this by modifying  universe that is given in argument to the solver, no?

> Also, we would need to also consider the dependency cone of ext_file_changed packages: if one package there is removed, installed or changed, that would trigger a rebuild, that we might not be able to process (in the case where the file was removed).

Dependency cone of `ext_file_changed` is handled in [`OpamSolution`](https://github.com/ocaml/opam/pull/3817/files#diff-ad92d5ee99a3f222e3233d030c90f611R960). It is packages from dependency cone that are removed from request, not just `ext_file_changed` ones.

src/client/opamSolution.ml
rjbou: It will end up on a `Nothing to do`.

rjbou:
Another UI concerns... If package `foo` have its external dependency changed. To reinstall the package should we allow `opam install foo` or keep requiring the reinstall operation `opam reinstall foo`.

src/client/opamSolution.ml
rjbou: Added confirm for continuation

src/client/opamSolution.ml
```diff
@@ -943,13 +943,67 @@ let resolve t action ~orphans ?reinstall ~requested request =
...
+    OpamSwitchState.universe t ~requested ?reinstall action
+  in
+  let request =
+    let remove =
+      OpamPackage.Set.filter (fun pkg ->
+          let n = OpamPackage.name pkg in
+          let f (n',_) = OpamPackage.Name.compare n n' = 0 in
+          not ((List.exists f request.wish_install)
+               || (List.exists f request.wish_upgrade)))
+        t.remove
```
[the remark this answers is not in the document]
rjbou: Updated this whole part, to refine `ext_file_changed` packages detection.

rjbou:
Closing as resolved by #3894

PR #3825, opened 2019-04-12
update deprecated ocaml-re use

rjbou:


AltGr:
No need to bump the minimum `re` version required ?

rjbou:
updated!

AltGr:
Thanks! It seems `src_ext/Makefile.sources` still has 1.8.0 though :)

rjbou:
Updated `src_ext` and rebased

Issue #3828, opened 2019-04-16
`opam clean -c` seems incompatible with `opam pin add -k git (...)`


erikmd:
### Steps to reproduce the issue

```
# This is not specific to dune, it is just an example package with a committed dune.opam
$ opam pin add -n -y -k git dune "https://github.com/ocaml/dune#a84b95949285966952ea6e89568362c6a5afd5e9"
$ opam install dune
$ opam clean -c
$ opam update

<><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
[coq-released] no changes from https://coq.inria.fr/opam/released
[default] synchronised from https://opam.ocaml.org

<><> Synchronising development packages <><><><><><><><><><><><><><><><><><><><>
[ERROR] Could not synchronize /home/coq/.opam/4.05.0/.opam-switch/sources/dune from
        "git+https://github.com/ocaml/dune#a84b95949285966952ea6e89568362c6a5afd5e9":
        "/usr/bin/git fetch -q" exited with code 1 "error: https://github.com/ocaml/dune did not send
        all necessary objects"
Now run 'opam upgrade' to apply any package updates.
```

### System info

```
# opam config report
# opam-version      2.0.4 
# self-upgrade      no
# system            arch=x86_64 os=linux os-distribution=debian os-version=9
# solver            builtin-mccs+glpk
# install-criteria  -removed,-count[version-lag,request],-count[version-lag,changed],-changed
# upgrade-criteria  -removed,-count[version-lag,solution],-new
# jobs              1
# repositories      2 (http) (default repo at db98cef2)
# pinned            1 (git)
# current-switch    4.05.0
```

### Additional info

This issue showed up in this coq-community CI job: <https://travis-ci.com/coq-community/math-classes/jobs/192817437>, that motivated that workaround patch: https://github.com/coq-community/docker-coq/commit/332ff995f9d18e276fa03c48d98e3f3fbc176a93

The issue seems related to #3485, albeit that latter issue didn't mention `opam clean -c`.

Being able to clean the opam download cache in this context would be useful to earn ~110 MB when building Docker images of Coq:

```
$ docker run --rm -it coqorg/coq:dev
  $ du -s .opam
  2670684 .opam
  $ opam clean -c
  Clearing cache of downloaded files
  $ du -s .opam
  2557616 .opam
```

erikmd:
[...]

PR #3831, opened 2019-04-19
Don't remove git cache objects that may be used

AltGr:
This is a quickfix, we can probably do better than that ; see #3828

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -3082,7 +3082,14 @@ let clean =
     if download_cache then
       (OpamConsole.msg "Clearing cache of downloaded files\n";
        rmdir (OpamPath.archives_dir root);
-       cleandir (OpamRepositoryPath.download_cache root));
+       List.iter (fun dir ->
+           match OpamFilename.(Base.to_string (basename_dir dir)) with
+           | "git" ->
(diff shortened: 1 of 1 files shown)


erikmd:
Hi @rjbou @AltGr, I've seen you put the milestone 2.1.0 for this PR #3831 (which will be important to be able to run `opam clean -c` again in `coqorg/coq:dev`-based images and thus save ~110MB…)

Do you confirm that the upcoming release of opam will be 2.1.0, or do you plan to make a point release 2.0.x beforehand? (in which case would this PR be eligible to be shipped along?)

erikmd:
@rjbou @AltGr actually I've just seen that you opened some [milestone for 2.0.6](https://github.com/ocaml/opam/milestones) :-)

Hence my question: would it be feasible for you to integrate the fix from PR #3831 in opam 2.0.6 ?

Kind regards, Erik

rjbou:
There will be 2.0.6 & 2.1.0~beta releases. What we plan is that the 2.0 branch is the LTS one until 2.2.0, and its releases contains only bug fixes. By default, all PR are tagged 2.1.0 until backported to 2.0 branch for a release (cf. backported commits PRs).
We can integrate then this PR into 2.0.6.

erikmd:
OK great! Thanks @rjbou for your explanation :+1:

PR #3841, opened 2019-05-14
Update guix install instructions

roptat:
Guix has changed its terminology, so the full system is now called Guix System instead of GuixSD (see [this note](https://www.gnu.org/software/guix/manual/en/html_node/Introduction.html#DOCF2) in the manual).

Guix also added a simpler alias to `guix package -i` called `guix install` and we try to encourage users to use it most of the time.

Thanks :)


rjbou:
Thanks!

PR #3845, opened 2019-05-22
Add support for Z3 as a solver backend

AltGr:
Support will be compiled and linked in if Z3 is installed on the opam
switch. This requires the latest update of the Z3 package
(https://github.com/ocaml/opam-repository/pull/14173) for proper
static linking.

Z3 has been shown to perform much better on pathological cases,
avoiding many timeouts, while being only a bit slower on most basic
cases. The cost is a doubled binary size, with static linking... Also,
the compilation time of Z3 should be considered.

For now, I made opam-solver depend on `mccs | z3` ; in reality, both
should be depopts, but we prefer people not to end up with opam
without a solver by mistake (it requires an external `aspcud` at
runtime in that case).

Perhaps we should keep `mccs` a dependency, and make `z3` a depopt,
though, to avoid the latter from being randomly chosen with its huge
build time.

If both are linked in, the current version still default to mccs if
the solver to use is unspecified. This is easy to change.

### DIFF
--- a/opam-solver.opam
+++ b/opam-solver.opam
@@ -22,9 +22,11 @@ build: [
 ]
 depends: [
   "opam-format" {= "2.0.0"}
-  "mccs" {>= "1.1+9"} | "z3" {= "4.8.4"}
+  "mccs" {>= "1.1+9"}
   "dose3" {>= "5"}
   "cudf" {>= "0.7"}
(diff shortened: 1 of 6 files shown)


AltGr:
This should at least wait for https://github.com/ocaml/opam-repository/pull/14173 to be merged.

PR #3852, opened 2019-05-29
pin-depends: update confirmation message to add skip option

### LINKED ISSUE #3840: opam pin add should provide interactively "skip this pin-depend"
what I currently get from ``opam pin add yyy `pwd`#HEAD`` when the `yyy.opam` contains `pin-depends` is:

```bash
The following additional pinnings are required by caldav.~dev:
  - webmachine.dev at git+https://github.com/roburio/ocaml-webmachine.git#webdav
Continue? [Y/n] n
You can specify --ignore-pin-depends to bypass
```

what I'd like to get:
```bash
Continue or skip? [Y/s/n] s
...opam continues to install yyy...
```

otherwise I've to run the command twice and put the `--ignore-pin-depends` on commandline.

please let me know what you think about that :)

rjbou:
fixes #3840

### DIFF
--- a/src/client/opamPinCommand.ml
+++ b/src/client/opamPinCommand.ml
@@ -369,8 +369,7 @@ let rec handle_pin_depends st nv opam =
              ~ignore_extra_pins:true)
          st extra_pins
      else
-     if
-       OpamConsole.confirm
+     if OpamConsole.confirm
          "Try to install anyway, assuming `--ignore-depends`?"
      then st else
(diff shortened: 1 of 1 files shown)


rjbou:
Updated to a double confirm instead of a choice: `--yes` compatible & more friendly

hannesm:
just wanted to say thank you for this patch, I used it in recent days several times :)

Issue #3846, opened 2019-05-24
Should opam install --deps-only impose target package constraints?


maroneze:
I am trying the 4.08+rc1, and I'd like to do a `opam install --deps-only frama-c` to quickly obtain all dependencies. However, the `frama-c` package is marked as incompatible with 4.08 (it was released before 4.08 was out), so opam fails with:

```
The following dependencies couldn't be met:
  - frama-c → ocaml < 4.08.0
      base of this switch (use `--unlock-base' to force)
```

I agree this makes sense most of the time, but since the `frama-c` package will not be installed anyway, it could make sense to emit a warning saying "Unmet dependencies for target package, do you want to continue?" and allow the user to try it anyway.

By the way, `--unlock-base` does not help, it simply downgrades OCaml.

rjbou:
The `deps-only` option installs the set of needed package in order to compile (or install later) the package. It is often used for local install & or switch creation `opam {install|create switch} .`. in order to have the current switch consistent with the package, constraints can't be dropped. In your case, if `frama-c` is incompatible with `ocaml.4.08.0`, you won't be able to compile `frama-c` if it is installed.
In some case, user can need to don't be restrained by constraints, you can use the option `--ignore-constraints-on`.

But you issue led to another one: conflicts are take into account when `deps-only` is used. It shouldn't be the case, as we don't want yet to install the package itself, but have an compliant environment.

rjbou:
You can also change `frama-c` constraints with pin editing it:
`opam pin edit frama-c`.
Thanks for the report!

PR #3853, opened 2019-05-29
install: remove conflicting packages if deps-only

rjbou:
In case of dependencies only install, there is no need to keep restriction of conflicting package.
Related to #3846

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -981,7 +981,8 @@ let filter_unpinned_locally t atoms f =
          None))
     atoms
 
-let install_t t ?ask atoms add_to_roots ~deps_only ~assume_built =
+let install_t t ?ask atoms add_to_roots
+    ~deps_only ?(ignore_conflicts=false) ~assume_built =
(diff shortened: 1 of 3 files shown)


AltGr:
Actually... A bit late, sorry, but I'm just having a second thought on this.
I mean, after `opam install foo --deps-only`, you expect to be able to compile `foo` by hand. If conflicting package `bar` is kept installed on the system, wouldn't that compromise your compilation ?

The proper response to #3846 might be to allow a "relaxed mode", akin to `--best-effort`. Unfortunately, `--deps-only --best-effort` won't work as you would hope at the moment, and would be actually pretty difficult to implement.

As a side-note, `--best-effort` could probably be improved with a report of wether it was partial or not, and what was skipped...

rjbou:
It depends if the conflict is related to build or install. Having the conflicting package installed doesn't always interfere with the build..
Maybe adding a message to inform that a conflicting package is kept?
_Note that as it is a quite minimal modification, having minimal consequences,  i don't insist on merging it._

rjbou:
Updated it as an option `--ignore-conflicts`

Issue #3823, opened 2019-04-11
Add —packages option to opam exec


Khady:
Now opam has built in support for lock files. It is great. But there is no convenient way to run a command in a context that for sure respects the constraints of an opam (or opam.locked) file.

For now we have to do

```
opam install ./my_project.opam ./second_project.opam --deps-only --locked
opam exec -- dune build
```

Instead it would be nice to have a single command that at least makes sure that the switch and an opam file are matching. 

```
opam exec --packages=./my_package.opam,./second_project.opam --locked -- dune build
```

rjbou:
This option would allow install the package and execute the command given this install? Or just check that dependencies specified in the opam file are installed?

Note that if you local opam file contains those build commands, you can use  [`--inplace-build`](https://opam.ocaml.org/doc/man/opam-install.html#lbAG) option to build in the local directory and not in the opam internal build directory.

Khady:
I would already be very happy with "just check that dependencies specified in the opam file are installed".

Unfortunately --inplace-build is not powerful enough for our use case as we have different commands we want to run in the context of one (or multiple) opam file, not only a build command.

rjbou:
From my point of view, it's not the purpose of `opam exec` to perform switch & package checking, especially if it have to handle all the logic of the filtering and opam file selection. It is intended to act on the field of opam/shell environment.
For check purpose, it is possible to add an `install` option, but it won't resolve the problem of two commands to launch.
One solution, is to resolve this at a shell level, or with a plugin that merges the two commands.

cc @AltGr

Khady:
> to resolve this at a shell level

Do you mean to create a shell script that launches the two commands?

My problem with that is how slow `opam install` is, even when using lock files. Close to 2 seconds to check that there is nothing to do.

rjbou:
You can launch the install command with `--debug` to find what is the time consuming operation. Maybe the solver?

You can combine `opam show`& `opam list` to see if a depends package is missing (without calling solver), depending on how you combine them, you can gain some little time (for me 0.5s).

rjbou:
There is now the option `opam install --check` (#3854) without calling the solver, that permit to save some time.
Thanks for the report!

PR #3854, opened 2019-05-29
Check package dependencies

rjbou:
Add an option `--check` to install command. It checks that dependencies of given packages are satisfied in the switch. ~If not, report missing ones, and asks to install them (equivalent to install package with `--deps-only` option). Another option `--no-inst` can be used if just the check is required.
Fails with exit code `1` if dependencies are missing.~ edit: cf. further comment

Related to #3823

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -1128,6 +1128,57 @@ let install t ?autoupdate ?add_to_roots
   let t = update_dev_packages_t autoupdate_atoms t in
   install_t t atoms add_to_roots ~deps_only ~assume_built
 
+let check_installed t atoms =
+  let available = (Lazy.force t.available_packages) in
+  let uninstalled = OpamPackage.Set.Op.(available -- t.installed) in
+  let pkgs =
+    OpamPackage.to_map
+      (OpamFormula.packages_of_atoms available atoms)
+  in
+  let test = OpamStateConfig.(!r.build_test) in
+  let doc = OpamStateConfig.(!r.build_doc) in
+  let env p =
+    OpamFilter.deps_var_env ~build:true ~post:true ~test ~doc
+      ~dev:(OpamSwitchState.is_dev_package t p)
+  in
+  OpamPackage.Name.Map.fold (fun name versions map ->
+      let compliant, missing_opt =
+        OpamPackage.Version.Set.fold (fun version (found, missing) ->
+            if found then (found, missing)
+            else
+            let pkg = OpamPackage.create name version in
+            let cnf_formula =
+              OpamSwitchState.opam t pkg
+              |> OpamFile.OPAM.depends
+              |> OpamFilter.filter_formula (env pkg)
+              |> OpamFormula.to_cnf
+            in
+            let missing_conj =
+              List.filter
+                (List.for_all (fun ((n,_vc) as atom) ->
+                     OpamPackage.Set.for_all
+                       (fun p -> not (OpamFormula.check atom p))
+                       (OpamPackage.packages_of_name t.installed n)))
+                cnf_formula
+            in
+            if missing_conj = [] then true, None
+            else false, Some (pkg,missing_conj))
+          versions (false,None)
+      in
+      if compliant then map
+      else
+      match missing_opt with
+      | None -> assert false (* version set can't be empty *)
+      | Some (pkg, missing_conj) ->
+        OpamPackage.Map.add pkg
+          (OpamPackage.names_of_packages
+             (List.fold_left (fun names disj ->
+                  OpamPackage.Set.union names
+                    (OpamFormula.packages_of_atoms uninstalled disj))
+                 OpamPackage.Set.empty missing_conj))
+          map
+    ) pkgs OpamPackage.Map.empty
+
 let remove_t ?ask ~autoremove ~force atoms t =
   log "REMOVE autoremove:%b %a" autoremove
     (slog OpamFormula.string_of_atoms) atoms;
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -1246,9 +1246,15 @@ let install =
        --destdir) to revert."
       Arg.(some dirname) None
   in
+  let check =
+    mk_flag ["check"]
+      "Check if dependencies are installed. If some packages are missing, \
+       it will ask if you want them installed and launch install of \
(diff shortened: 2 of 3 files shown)


src/client/opamClient.ml
```diff
@@ -1128,6 +1128,38 @@ let install t ?autoupdate ?add_to_roots
...
+        |> OpamFormula.ands_to_list
+      in
+      let missing_conj =
+        List.filter (fun f -> not (OpamFormula.verify_set f t.installed)) cnf_formula
+      in
+      if missing_conj = [] then map
+      else
+      let uninstalled = OpamPackage.Set.Op.(available -- t.installed) in
+      OpamPackage.Map.add to_check_pkg
+        (List.flatten (List.map (fun form ->
```
AltGr: `List.fold_left (fun names disj -> OpamPackage.Name.Set.union names @@ OpamPackage.names_of_packages (OpamFormula.packages disj))` ?

src/client/opamClient.ml
```diff
@@ -1128,6 +1128,38 @@ let install t ?autoupdate ?add_to_roots
...
+    OpamFilter.deps_var_env ~build:true ~post:true ~test ~doc
+      ~dev:(OpamSwitchState.is_dev_package t p)
+  in
+  OpamPackage.Set.fold (fun to_check_pkg map ->
+      let cnf_formula =
+        OpamSwitchState.opam t to_check_pkg
+        |> OpamFile.OPAM.depends
+        |> OpamFilter.filter_formula (env to_check_pkg)
+        |> OpamFormula.cnf_of_formula
+        |> OpamFormula.ands_to_list
```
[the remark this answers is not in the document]
rjbou: `OpamFormula.to_cnf` returns an `atom cnf` while we manipulate a `formula` here. It is then simpler as there is no more need to call the `OpamFormula.verify_set` function.

src/format/opamFormula.ml
```diff
@@ -293,6 +293,14 @@ let verifies f nv =
       check_version_formula cstr (OpamPackage.version nv))
     name_formula
 
+let verify_set f set =
+  let dnf = (dnf_of_formula f) in
+  dnf <> Empty &&
+  eval (fun (name, cstr) ->
+      OpamPackage.packages_of_name set name
+      |> OpamPackage.Set.exists (fun nv ->
+          check_version_formula cstr (OpamPackage.version nv))) dnf
```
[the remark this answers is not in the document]
rjbou: no more relevant

src/client/opamClient.ml
rjbou: indeed

rjbou:
There is no difference, it's even less useful to use `opam install foo --check --yes` instead of `--deps-only`. The choice to add interactive mode is just to not let user launch another command in case of missing dependencies. It can just advice to install with `--deps-only`, and keep the check to its first purpose...

AltGr:
I found a small inconsistency in the code: the `check_installed` function takes atoms as arguments, but then aggregates the results for all their matching packages, while we would probably expect them to match for at least one (e.g. if I do `opam install --check-depends foo` and `foo` has 5 versions, I probably don't care if the older ones have missing dependencies, as long as one of them can be installed).

> not let user launch another command in case of missing dependencies.
While there are cases where I agree it's needed for user-friendliness, I don't think this is a problem in general, if it enables us to have a cleaner interface (i.e. where it is clear what one command does and doesn't do). Since this function is by design directed to scripting, I'd rather indeed just terminate with a return value of 1.

src/client/opamClient.ml
```diff
@@ -1130,38 +1130,53 @@ let install t ?autoupdate ?add_to_roots
...
-      let missing_conj =
-        List.filter
-          (List.for_all (fun ((n,_vc) as atom) ->
-               OpamPackage.Set.for_all
-                 (fun p -> not (OpamFormula.check atom p))
-                 (OpamPackage.packages_of_name t.installed n)))
-          cnf_formula
+  OpamPackage.Name.Map.fold (fun name versions map ->
+      let compliant, missing_opt =
+        OpamPackage.Version.Set.fold (fun version (found, missing) ->
```
AltGr: If I am not mistaken, `found = true` ⇔ `missing = None`, so maybe this can be slightly simplified ?

src/client/opamCommands.ml
```diff
@@ -1292,26 +1287,20 @@ let install =
...
-                       OpamPackage.Name.to_string
-                       (OpamPackage.Name.Set.elements names)))
-                 (OpamPackage.Map.bindings missing));
-           not (no_inst) && OpamConsole.confirm ~default:true "Install them?")
-      else deps_only
-    in
-    if check && not deps_only then
-      OpamStd.Sys.exit_because `False;
+    if check then
+      (let missing = OpamClient.check_installed st atoms in
```
AltGr: For consistency, and easier use in scripts, I would output in both cases:
- a readable message to stderr (`OpamConsole.errmsg`)
- the raw list of missing packages to stdout

Otherwise, this is perfect! Thanks :)

src/client/opamClient.ml
rjbou: It would be cleaner on the idea, but more _complex_ code without the `found` argument: `None` can't be the first accumulator value (ends on the first call), in that case, as `missing` contains a package, first value need to be calculated by taking an element of the set & apply part of the fold arg function on it.

src/client/opamCommands.ml
rjbou: With a raw list, we don't have the information of missing package -> atom package. I Suggest
```
p1.v1: m1, m2, m3
p2.v2: m4, m5
```
fi, current dispay is
```
- p1.v1: m1, m2 and/or m3
- p2.v2: m4 and/or m5
```

src/client/opamClient.ml
AltGr: Aah, good point indeed. You could start with the acc `Some []`, though.

rjbou: not really, because `missing: (package * atom disjunction list) option` (and there is no empty package) :)

src/client/opamCommands.ml
rjbou: note: lint err output is quieted to not interfere with std{err,out}

src/client/opamClient.ml
AltGr: Right :)

rjbou:
Updated accordingly

PR #3881, opened 2019-06-23
Read jobs variable from OpamStateConfig

dra27:
The jobs variable was always read from the opam config file, so didn't reflect the `OPAMJOBS` variable or `--jobs parameter`. This hurts https://github.com/ocaml/opam-repository/pull/14257, as it means there's no way to prevent parallelism.

In general, using `jobs` to control build system parallelism is of course imperfect, although for OCaml itself this is more likely to be sound, since the compiler is usually a bottleneck during package installation.

This PR makes `jobs` behave in the same way as `make`, reading the `OpamStateConfig` value, which is influenced by `OPAMJOBS` and also `--jobs`. With this change, `OPAMJOBS=1000 opam config var jobs` displays 1000 instead of the value of `jobs` recorded at `opam init` time.

### DIFF
--- a/src/state/opamPackageVar.ml
+++ b/src/state/opamPackageVar.ml
@@ -64,7 +64,7 @@ let resolve_global gt full_var =
     | _ ->
       match V.to_string var with
       | "opam-version"  -> Some (V.string OpamVersion.(to_string current))
-      | "jobs"          -> Some (V.int (OpamFile.Config.jobs gt.config))
+      | "jobs"          -> Some (V.int (OpamStateConfig.(Lazy.force !r.jobs)))
       | "root"          -> Some (V.string (OpamFilename.Dir.to_string gt.root))
       | "make"          -> Some (V.string OpamStateConfig.(Lazy.force !r.makecmd))
       | _               -> None


dbuenzli:
Actually I always thought the `jobs` variable you use in build descriptions and `OPAMJOBS` as different things. 

The first one is the build system level parallelism of a package build, the second one is the opam package build level parallelism (how many packages opam should build in parallel). 

I'm not sure it's a good idea to confuse the two.

dra27:
It may be what you thought, but it's not the description of the variable!

<pre><b>jobs</b>              <font color="#729FCF"><b>3</b></font>               # The number of parallel jobs set up in opam configuration
</pre>

(if this were to be merged, that description should be updated)

What would you propose instead?

dbuenzli:
Well the manual says for config:
```
jobs: <int>: the number of concurrent jobs to run for build processes. The default value is calculated from the number of cores.
```
And under global variables: 
```
jobs: opam's jobs (-j) parameter, i.e. the number of parallel builds opam is allowed to run
```
So it is all a bit confusing. 

I think best would have been:
```
--jobs OPAMJOBS  # opam level //
--build-jobs OPAMBUILDJOBS  # package level //
```
But the variable being used in `opam` files is called `jobs` and that's it would be defined via `--build-jobs` so I suspect it will too error prone with people specifying `--jobs` on the cli instead of the `-build-jobs` they want.

Not super convinced but maybe this is better:
```
--jobs OPAMJOBS  # package level //
--install-jobs OPAMINSTALLJOBS  # opam level //
```

dbuenzli:
(what I suspect it that it doesn't concerns only `install`s , maybe `--tasks`, `OPAMTASKS` meh)

AltGr:
@dbuenzli indeed this is a confusion that has been running for a long time: this variable indeed controls the number of build jobs  opam is allowed to run in parallel, but ended up being used as the `--jobs` parameter in build scripts, which I didn't foresee but is a legitimate use. I agree these should be different.

(wasn't there an old issue discussing that a long time ago, with ideas on how to compute the relation between the two etc? )

dbuenzli:
> (wasn't there an old issue discussing that a long time ago, with ideas on how to compute the relation between the two etc? )

I remember discussing this but when I wrote the above I couldn't find the discussion.

dra27:
Given ocaml/opam-repository#14257, could this be considered for 2.0.5?

PR #3883, opened 2019-06-24
Correct unnecessary alternation in URL parser

dra27:


dra27:
Ah - @rjbou points out that this commit was intentionally reverted in https://github.com/ocaml/opam/pull/3754

PR #3884, opened 2019-06-25
switch install: fix multiple compiler candidates

### LINKED ISSUE #3874: creating local switch only works if there already is a switch
I wanted to start using local switches.
Locally, with several existing global switches (one being 4.07.1), `opam switch -y create ./ --deps-only` works fine (log says `installed ocaml-system.4.07.1`).
On travis-ci, after `opam init -y -a --bare` (without having any switches installed), it fails with `No available version of ocaml-base-compiler satisfies the constraints` ([see log](https://travis-ci.com/goblint/analyzer/jobs/207815684#L536)).
There it only works if I specify a compiler version, e.g., `opam switch -y create ./ --deps-only ocaml-base-compiler.4.07.1`.
I would expect that local switches also install a compiler if there is none yet.
The version constraint in the `opam` file is `"ocaml" {>= "4.04.1" & < "4.08"}`.

rjbou:
On local switch creation, if no compiler is specified or & single infered, there is the choice between several candidates that comply with the local opam file, defined as a conjunction that leads to an unresolvable solver request.

Fixes #3874

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -2084,23 +2084,23 @@ let switch =
       | packages, _ -> packages
     in
     let compiler_packages rt ?repos switch compiler_opt =
-      let open OpamStd.Option.Op in
-      let single_lst p = (p >>| (fun x -> [x])) +! [] in
       match packages, compiler_opt, OpamSwitch.is_external switch with
       | None, None, false ->
-        OpamSwitchCommand.guess_compiler_package ?repos rt
-          (OpamSwitch.to_string switch)
-        |> single_lst, false
+        OpamStd.Option.to_list
+          (OpamSwitchCommand.guess_compiler_package ?repos rt
+             (OpamSwitch.to_string switch)), false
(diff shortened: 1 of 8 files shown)


rjbou:
ping @AltGr especially for 62fcdb9

src/client/opamCommands.ml
```diff
@@ -2083,19 +2084,24 @@ let switch =
       | packages, _ -> packages
     in
     let compiler_packages rt ?repos switch compiler_opt =
+      let open OpamStd.Option.Op in
+      let single_lst p = (p >>| (fun x -> [x])) +! [] in
```
AltGr: Worth adding `OpamStd.Option.to_list` maybe ? :D
(which made its way into 4.08.0 IIRC)

dra27: `Option.to_list` is indeed in 4.08.0

rjbou: A good occasion actually :)

PR #3887, opened 2019-06-26
Bump src_ext Dune to 1.6.3

dra27:
Allows compilation with OCaml 4.08.0 - this is for 2.0 branch only, `master` should be bumped to latest Dune (though there's some work to deal with the `jbuild` deprecation warning needed).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ all: opam opam-installer
 	@
 
 ifeq ($(DUNE),)
-  DUNE_EXE = src_ext/dune-local/_build/install/default/bin/dune$(EXE)
+  DUNE_EXE = src_ext/dune-local/_build_bootstrap/install/default/bin/dune$(EXE)
   ifeq ($(shell command -v cygpath 2>/dev/null),)
     DUNE := $(DUNE_EXE)
   else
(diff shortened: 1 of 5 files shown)


avsm:
Thanks, I think it would be useful to cut this as a 2.0.5 point release

avsm:
Incidentally, why not just go up to latest Dune 1.10.x?

dra27:
Because the `jbuild` deprecation warning breaks building the man pages, and I'm still working on that fix!

dra27:
@avsm - presumably also OK to cut it as 2.0.4-1 release tarball and submit to homebrew, etc. directly? It feels odd to cut a release with no actual features or bug fixes.

avsm:
Just cherry pick in #3886 and you have a bugfix release :-) I think it's far easier for packaging if we're consistent with our versioning.  For instance the `-1` suffix is also used by debian as an extra version so it starts to clash.

dra27:
The version wouldn't change it - it's just a new release tarball and any version bumping in Debian is theirs to do. But you're right that a quick release can also be cut.

avsm:
Generally speaking, I view a build system upgrade as a reasonably major change -- there may be unexpected regressions or side effects that we don't anticipate. From a packagers perspective, newer versions of Dune may do something surprising that affects their packaging infrastructure and require some tweaks. A "recut" of a point release should be reserved for when we are absolutely sure there are no weird side effects (for example, documentation fixes or README changes or things of that nature).

PR #3888, opened 2019-06-26
Support Dune 1.7.0 and later

### LINKED ISSUE #3870: opam 2.0.4 doesn't compile when using dune 1.10.0
opam 2.0.4 doesn't compile when using dune 1.10.0.

The build fails with:

```
Note: You can use "dune upgrade" to convert your project to dune.
File "src_ext/result/jbuild", line 1, characters 0-0:
Warning: jbuild files are deprecated, please convert this file to a dune file instead.
Note: You can use "dune upgrade" to convert your project to dune.
make[1]: *** [Makefile:41: man] Error 1
make[1]: Leaving directory '/tmp/opam/build/doc'
make: *** [Makefile:171: man] Error 2
Command '['/bin/sh', '-c', u'./configure && make lib-ext && ./configure && make && make install prefix="$DISTDIR" "$@"', '-']': Build failed with exit code 2
Command failed with exit status 1
```

Perhaps it is interpreting these jbuild warnings as build errors?
https://github.com/ocaml/opam/blob/d6871934bcfecdcd737b305cf0fb44e6093038af/doc/Makefile#L46

Building opam without dune installed fixes the problem.

dra27:
Dune 1.7.0 displays a deprecation warning for jbuild files which needs to be suppressed when generating the man pages.

Fixes #3870.

### DIFF
--- a/doc/Makefile
+++ b/doc/Makefile
@@ -43,7 +43,8 @@ man:
 	done
 	$(OPAM_INSTALLER) $(HELPFMT) > man/opam-installer.1 2>> man/err
 # Dune doesn't (yet) support --no-print-directory
-	@if grep -v "^Entering directory '" man/err > man/err2 ; then cat man/err2 ; false ; fi
+	@sed -f man.sed man/err > man/err2
+	@if test -s man/err2 ; then cat man/err2 ; false ; fi
 
 man-html: man
 	rm -rf man-html
--- a/doc/man.sed
+++ b/doc/man.sed
@@ -1,0 +1,7 @@
+/^Entering directory '/d
+/^File .*jbuild", line 1, characters 0-0:/{
+  :a
+  N
+  /Note: You can use.*to dune/!ba
+}
+/Warning: jbuild files are deprecated/d


dra27:
I'd be grateful if someone with access to a Mac which isn't running GNU sed (or a BSD box in the same limited position!) could verify that that `sed` script is entirely POSIX.

doc/Makefile
```diff
@@ -43,7 +43,8 @@ man:
 	done
 	$(OPAM_INSTALLER) $(HELPFMT) > man/opam-installer.1 2>> man/err
 # Dune doesn't (yet) support --no-print-directory
```
avsm: Is there an upstream dune pr for this `--no-print-directory` feature? Worth noting in the comment if so.

avsm:
I do have full dune ports of all the dependencies in src_ext in https://github.com/dune-universe/

Would it be better to just cut src_ext tarballs from those git repos instead?

doc/Makefile
dra27: It's in Dune 1.7.0 - on the `master` branch PR (coming shortly) for Dune 1.10, the comment is gone...

dra27:
I know 🙂 I prefer the `src_ext` to remain upstream tarballs as far possible. The ones causing "pain" here are all packages which upstream were using `jbuilder`, so I expect all we need to do is update to their latest versions.

dra27:
(you're applying "master" reviewing to a maintenance branch!)

dra27:
It will of course be worth back-porting #3902 once it’s finalised

PR #3890, opened 2019-06-26
Typos fixed downstream by Debian

dra27:
Taken from https://sources.debian.org/patches/opam/2.0.3-1/0002-Fix-spelling-error-in-manpage.patch/
with some further amendments.

NB I've based this PR on 2.0 - the commit cleanly rebases onto master as well.

cc @mehdid


rjbou:
master pr #3891

PR #3892, opened 2019-06-27
Clarify message in opam init --yes

dra27:
`opam init -y` asked the user setup configuration question and then answers no: advertise `--shell-setup` instead.

### DIFF
--- a/src/state/opamEnv.ml
+++ b/src/state/opamEnv.ml
@@ -671,24 +671,28 @@ let setup
         (OpamConsole.colorise `cyan @@ OpamFilename.prettify dot_profile)
         (OpamConsole.colorise `bold @@ source root shell (init_file shell))
         (OpamConsole.colorise `bold @@ "eval $(opam env)");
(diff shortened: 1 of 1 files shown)


src/state/opamEnv.ml
```diff
@@ -671,24 +671,28 @@ let setup
...
-          with
-          | None   ->
-            OpamConsole.msg "Alright, assuming you changed your mind, not \
-                             performing any changes.\n";
-            None
-          | Some f -> Some (OpamFilename.of_string f)
-        end
-      | _ -> None
+      if OpamCoreConfig.(!r.answer = Some true) then begin
+        OpamConsole.warn "Shell not updated in non-interactive mode: use --shell-setup";
```
AltGr:
```suggestion
        OpamConsole.warning "Shell not updated in non-interactive mode: use --shell-setup";
```

dra27: 😊

PR #3478, opened 2018-07-24
Bump version to 2.1.0

AltGr:

PR #3894, opened 2019-06-27
Define switch invariants rather than "base packages"

AltGr:
~NOTE: this is WIP and should be rebased once #3478 is merged to avoid
patch duplication~

Currently, we define switches with a set of "base packages", that
normally correspond to the compiler, and that are strongly locked
(unless `--unlock-base` is specified). This is somehow inherited from
the older "compiler definitions" that weren't packages.

This PR proposes to switch to a much more flexible notion of "switch
invariant", which can be an arbitrary dependency formula (same format
as `depends:`, minus the extended formulas).

- The invariant is defined in `SWITCH/.opam-switch/switch-config`
- It can be automatically inferred from the base packages (so the
  upgrade should be smooth. No global upgrade needed)
- This allows easy (and controllable) automatic upgrades of compilers
  (typically 'ocaml-system')
- The solver will always enforce it (using a virtual package
  internally, without consequences on the optim criteria)
- Base packages are still there, but informative only (and they help
  for migration). They will be automatically updated on changes

This works quite well, but some small parts are still missing on top
of it:

- [X] Switch creation needs updating to generate the invariants, either
  inferred from the arguments on the CLI, or the otherwise known base
  packages
- [X] `--unlock-base` won't work anymore, it needs to affect the
  invariant instead of the base packages
- [X] provide CLI options for switch invariant control (show and change)
- [X] check stuff that relies on base packages, like `opam fixup`
- [ ] we may consider extending the `switch-export` format to include
  invariants
- [ ] there is some regression on conflict messages


AltGr:
Auto-merge with the merged #3960 was completely broken (duplicated definitions), hence the failures.
Rebasing...

AltGr:
Rebased. I am still working on improving the conflict messages, but it ended up being much more difficult than expected.

So I say we don't stall this any longer, knowing that there will be a regression on conflict error messages when the compiler versions are in cause; that'll be improved in another PR asap

src/client/opamCommands.ml
```diff
@@ -2391,18 +2421,28 @@ let switch =
...
-         `Ok ()
-       with Failure e -> `Error (false, e))
+         let st = OpamSwitchCommand.set_invariant ~force st invariant in
+         let st =
+           if no_action || OpamFormula.satisfies_depends st.installed invariant
+           then st
+           else OpamClient.install_t
+               st ~ask:true [] None ~deps_only:false ~assume_built:false
+         in
+         OpamSwitchState.drop st;
```
rjbou: No drop with a `with_`?

AltGr: It's not needed when using `with_`, but here it's a different, modified `st` that we would need to ignore anyway ; so calling `drop` is a bit better than `ignore`...

AltGr:
The last test on appveyor apparently failed due to OCaml 4.10, but otherwise all tests seem OK.

PR #3897, opened 2019-06-27
VT100 escape sequences appearing on Windows 10 Console

dra27:
During `opam init`, things proceed nicely until a second or so after the repository tarball is downloaded when the "Processing" message suddenly appears as:

```
[K[[32mdefault[0m] Initialisedlt[0m: http]
[KProcessing  1/1:
[36m<><>[0m [01mRequired setup - please read[0m
[36m<><><><><><><><><><><><><><><><><><><><><><><>[0m   
```
The problem is that processes invoked by `opam` gain `stdin` and therefore the Console and clear the Console mode (in particular, Cygwin processes seem to do this). They're running in the background, so it doesn't matter whether they behave nicely and restore the console mode on termination, because the message may well be displayed while they're running.

This PR offers two "fixes":
- firstly the `OpamConsole` mode is altered to force-set the mode on every print (this is ugly, but not much more ugly than what was there before)
- secondly the default for `allow_stdin` is declared `false` on Windows, and a couple of places where `allow_stdin` should clearly be `false` (`cygpath` invocation and the download command) are set

Either of these on their own is sufficient to fix the Windows problem.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -115,6 +115,7 @@ New option/command/subcommand are prefixed with ◈.
   * Fix `opam exec` on native Windows when calling cygwin executables [#4588 @AltGr]
   * Fix temporary file with a too long name causing errors on Windows [#4590 @AltGr]
   * CLI: Add flag deprecation and replacement helper [#4595 @rjbou]
+  * Win32 Console: fix VT100 support [#3897 @dra27]
 
 ## Test
   * Make the reference tests dune-friendly [#4376 @emillon]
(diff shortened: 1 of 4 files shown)


XVilka:
Will be this a part of 2.0.6? Seems a reasonable user experience bug fix.

dra27:
This isn’t ready to go - the CI failures are relevant

dra27:
I don't intend merging https://github.com/ocaml/opam/pull/3897/commits/890c297d3b6d92775df1a2e8722fead6e5888dcd (the stdin change) in 2.1.

dra27:
Good to go with CI!

AltGr (approved):
LGTM

PR #3899, opened 2019-06-28
Fix environment file handling for empty switches

dra27:
`opam switch create foo --empty` doesn't write an `environment` file. At this point, `opam config revert-env` will not return any updates. `opam env` will regenerate file. This PR fixes both `opam switch create --empty` to write the `environment` file and also `opam config revert-env` to regenerate it if it's missing (so belt-and-braces approach). Finally, the environment is also not checked in `opam switch create --empty` because of the same branch, and this is fixed (it hurts Windows, which actually updates the environment at this point; it doesn't hurt Unix if you have the shell hook installed).

### DIFF
--- a/src/client/opamConfigCommand.ml
+++ b/src/client/opamConfigCommand.ml
@@ -217,6 +217,25 @@ let print_eval_env ~csh ~sexp ~fish env =
   else
     print_env env
 
+let ensure_env_aux ?(set_opamroot=false) ?(set_opamswitch=false) ?(force_path=true) gt switch =
+  let env_file = OpamPath.Switch.environment gt.root switch in
+  if not (OpamFile.exists env_file) then
+    Some (OpamSwitchState.with_ `Lock_none gt @@ fun st ->
+          let upd =
+            OpamEnv.updates ~set_opamroot ~set_opamswitch ~force_path st
+          in
+          log "Missing environment file, regenerate it";
+          if not (OpamCoreConfig.(!r.safe_mode)) then
+            (let _, st =
+               OpamSwitchState.with_write_lock st @@ fun st ->
+               (OpamFile.Environment.write env_file upd), st
+             in OpamSwitchState.drop st);
+          OpamEnv.add [] upd)
+  else
+    None
+
+let ensure_env gt switch = ignore (ensure_env_aux gt switch)
+
 let env gt switch ?(set_opamroot=false) ?(set_opamswitch=false)
     ~csh ~sexp ~fish ~inplace_path =
   log "config-env";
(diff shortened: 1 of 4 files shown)


src/client/opamConfigCommand.ml
```diff
@@ -217,6 +217,25 @@ let print_eval_env ~csh ~sexp ~fish env =
   else
     print_env env
 
+let ensure_env_aux ?(set_opamroot=false) ?(set_opamswitch=false) ?(force_path=true) gt switch =
+  let env_file = OpamPath.Switch.environment gt.root switch in
+  if not (OpamFile.exists env_file) then
+    Some (OpamSwitchState.with_ `Lock_none gt @@ fun st ->
```
AltGr: A good reason to not take the write lock directly in this case ? Lock upgrades should be avoided unless absolutely necessary (and the scenarios for deadlocks have been considered)

AltGr:
LGTM besides the comment above!

src/client/opamConfigCommand.ml
dra27: Happy to change this, but this is a refactoring - it's exactly the code deleted below.

AltGr: Oh right! I should ask @rjbou then :)

src/client/opamSwitchCommand.ml
```diff
@@ -187,7 +187,11 @@ let remove gt ?(confirm = true) switch =
 
 let install_compiler_packages t atoms =
   (* install the compiler packages *)
-  if atoms = [] then t else
+  if atoms = [] then begin
+    OpamFile.Environment.write (OpamPath.Switch.environment t.switch_global.root t.switch) (OpamEnv.compute_updates t);
+    OpamEnv.check_and_print_env_warning t;
+    t
```
AltGr: Won't it be redundant to print the env warning at this point ?

dra27: There's always the update to `PATH` - you should still run `opam env` even for an empty switch?

src/client/opamConfigCommand.ml
rjbou: `env` file is only regenerated & written if a write lock can be taken (not `--safe`, cf. 3691). We don't want to block `opam env` because of the missing environment file.

AltGr: Yes, was I was suggesting is stg along the lines of:
```
  if not safe_mode && not OpamFile.exists env_file then OpamSwitchState.with_ `Lock_write gt @@ ... (* regenerate *)
  else OpamSwitchState.with_ `Lock_none gt @@ ...
```

the point would just be to avoid the use of `OpamSwitchState.with_write_lock` really, since we could detect the need beforehand and load with a write lock directly, which is always better & safer than upgrading existing locks.

PR #3900, opened 2019-06-28
Bits 'n bobs

dra27:
Shedding the weight of some simpler commits:

- `opam-admin.top` - prior to Jbuilder, this was built using `ocamlmktop`. I converted this badly, because the existing code now works as a toplevel, but you can't run a script by having `#!opam-admin.top` or `opam-admin.top foo.ml`. The commit generates the `let _ = Topmain.main ()` in a second file which is used to create `opam-admin.top` and the actual `opam_admin_top.ml` is now used to generate a library. This means that scripts have two choices - you can either `#require "opam-admin.top"` in your script, or you can run an `.ml` file with `opam_admin_top` (although you need `opam_admin_top.cmi` in the same directory as your script).
- `%{exe}%` does exactly what it says on the tin. The main reason for making it a global is to standardise the name used (so there's no temptation for `%{ext_exe}%`, etc. This extension is intentionally blank on Cygwin because Cygwin has its own official handling for automatically adding `.exe`.
- Back when originally working on this, I was converting 1.2 repos to 2.0 regularly - it was a nuisance for flexdll that the `ocaml` dependency kept getting re-added after I'd manually removed it... this legacy commit doesn't do that if the `opam-version` of the file is already `2.0`. The principle may be useful in the future...
- Also in `opam admin upgrade` was an attempt to download URLs using their filename. Fine for a lot of things, but troublesome on Windows when you start having `?` characters, etc., so this patch downloads patches and other referenced files using a safe filename (it's only to compute their checksums).
- I'd like, prior to any kind of global depext variable support, to introduce `sys-ocaml-arch`, `sys-ocaml-libc` and `sys-ocaml-cc` to allow accurate selection of the system compiler on Windows (this may prove unnecessary with the other work we've been looking into on switch parameters, but I don't think these variables would ever do any harm in future).

### DIFF
--- a/src/state/opamFormatUpgrade.ml
+++ b/src/state/opamFormatUpgrade.ml
@@ -162,7 +162,8 @@ let opam_file_from_1_2_to_2_0 ?filename opam =
     aux available
   in
   let pkg_deps =
-    if NMap.mem ocaml_wrapper_pkgname pkg_deps ||
+    if OpamVersion.compare (OpamFile.OPAM.opam_version opam) v2_0 >= 0 ||
+       NMap.mem ocaml_wrapper_pkgname pkg_deps ||
        OpamFile.OPAM.has_flag Pkgflag_Conf opam
     then pkg_deps
     else NMap.add ocaml_wrapper_pkgname Empty pkg_deps
--- a/src/state/opamFormatUpgrade.ml
+++ b/src/state/opamFormatUpgrade.ml
@@ -65,6 +65,8 @@ let upgrade_depexts_to_2_0_beta5 filename depexts =
          None)
     depexts
 
+let v2_0 = OpamVersion.of_string "2.0"
+
 let opam_file_from_1_2_to_2_0 ?filename opam =
   let ocaml_pkgname = OpamPackage.Name.of_string "ocaml" in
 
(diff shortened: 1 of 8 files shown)


src/state/opamFormatUpgrade.ml
```diff
@@ -19,6 +19,8 @@ let slog = OpamConsole.slog
 
 exception Upgrade_done of OpamFile.Config.t
 
+let v2_0 = OpamVersion.of_string "2.0"
```
rjbou: `v2_0` defined twice (cf. [L1042](https://github.com/ocaml/opam/blob/16ed21c0dd262a57afbff827e42738a93355d09b/src/state/opamFormatUpgrade.ml#L1042)). By convention, at the beginning of the files these versions are not defined ([for ex.](https://github.com/ocaml/opam/blob/16ed21c0dd262a57afbff827e42738a93355d09b/src/state/opamFormatUpgrade.ml#L332)) and defined only in front of their translation functions.

dra27: This is fixed - although for some reason GitHub is still display this line, even though 3c6ddbe1 doesn't have it!

rjbou: reintroduced by 5e4798f6 ^^

dra27: 🤦‍♂

Issue #3735, opened 2019-01-31
feature request: instead of hardcoding curl and wget, allow to specify a download-program at configure-time


hannesm:
curl and wget are atm hardcoded into opam. They are widely deployed. But FreeBSD (or OpenBSD, likely others) distribute a fine download-program with their base system (`fetch` on FreeBSD, `ftp` on OpenBSD). If opam would allow to specify the download command during configuration time (as in `.opam/config`), this would allow us on BSD to use the builtin binary, and not require an external program to be installed via ports.

WDYT? I can try to come up with a PR if you agree that this would be a good idea. Another option may be to detect and use `fetch` / `ftp` on the specific systems.

dra27:
I think it's possible to do this already with `OPAMFETCH` but it'd be good to improve the default on BSDs (it could also allow Windows to use powershell's `Invoke-WebRequest`, with the same benefit for all modern Windows systems of not needing a dependency to be installed).

hannesm:
@dra27 `OPAMFETCH` sounds like an environment variable!? I wouldn't know how to pass this during a port build on FreeBSD. I suggest a configure argument. :)

dra27:
@dbuenzli beat me to it! `wget` is difficult to portably have not follow redirects (and the default `curl` invocation includes `-L`) so it would be good if the docs specified that the command should follow redirects, yes.

hannesm:
@dbuenzli I'm aware of this, and already use it on my computer. What I ask in this issue is to have a configure-time way to specify this. I couldn't find a way to modify `.opam/config` system-wide.

dra27:
Ohhhh, sorry @hannesm - I'd completely misread the feature request.

This can be done now by overridding `src/client/opamInitDefaults.ml`, but at the moment that would have to be done by patching it, but local patches are presumably OK in ports? You want the `dl_tool` function to return `Some [(CString "fetch ..."), None]` in the `None` case so that `.opam/config` always has `download-command` on BSD. You can either use `OpamStd.Sys.os ()` to determine if it's OpenBSD or FreeBSD or just have a different patch, as I assume they use different package repos?

For the feature request itself, as I said initially, I think it would be better to alter the defaults rather than require it to be specified at `configure`-time.

rjbou:
It is possible to add in *BSD package install of an `/etc/opamrc` file containing the `download-command` field, it will be taken into account at opam init or reinit, and then automatically added in the opam config file.
[...]

PR #3904, opened 2019-06-30
default to use "fetch" on FreeBSD / "ftp" on OpenBSD

hannesm:
thanks a lot to @rjbou for the commit, I adjusted (as requested in #3735 the command line arguments of `fetch`/`ftp` in the last commit (and rebased the others to master), namely:
- pass user-agent
- pass output filename
- pass URL

I read through the command-line arguments passed to curl, and that looks pretty similar (e.g. fetch follows redirects automatically). I also tested this on FreeBSD, working nicely (don't have a OpenBSD to really test it, maybe someone else (/cc @avsm @adamsteen) would like to do this?

> @rjbou On default init config, I've only put `fetch`/`ftp` as required tools, and in OpamDownload, I've set `fetch`/`ftp` as default command, curl as fallback. In that case, should we add `curl` in recommended_tools:?

in contrast to Linux, both BSDs ship their distribution as kernel plus userland / world -- and the latter not only contains basic utilities (such as a C compiler etc.), but also fetch/ftp - of course you can compile your custom *BSD system which does not include any such tools (and then opam should need curl/a C compiler/make/...) - but that's not anything I recommend to support via opam (plus the download command can already be overridden via OPAMFETCH). So IMHO this is good as is, we don't really need to suggest people to install curl.

Some output from opam:
```
+ /usr/bin/fetch "https://opam.ocaml.org/cache/sha256/e2/e294aa1f920205da2c9f0ab909e2755b070702cb4fdb47a94b9c18e55cdf774d" "-o" "/usr/home/hannes/.opam/download-cache/sha256/e2/e294aa1f920205da2c9f0ab909e2755b070702cb4fdb47a94b9c18e55cdf774d.tmp.part" "--user-agent" "opam/2.1.0~beta"
- /usr/home/hannes/.opam/download-cache/sha256/e2/e294aa1f920205d          47 kB  308 kBps    00s
```

### DIFF
--- a/src/client/opamInitDefaults.ml
+++ b/src/client/opamInitDefaults.ml
@@ -39,6 +39,8 @@ let linux_filter = os_filter "linux"
 let macos_filter = os_filter "macos"
 let openbsd_filter = os_filter "openbsd"
 let freebsd_filter = os_filter "freebsd"
+let not_open_free_bsd_filter =
+  FNot (FOr (openbsd_filter,  freebsd_filter))
 let win32_filter = os_filter "win32"
 let sandbox_filter = FOr (linux_filter, macos_filter)
 
@@ -73,6 +75,16 @@ let bwrap_string () = Printf.sprintf
     bwrap_cmd
 
 let req_dl_tools () =
+  let msg =
+    Some "A download tool is required, check env variables OPAMCURL or OPAMFETCH"
+  in
+  let default =
+    [
+      ["curl"; "wget"], msg, Some not_open_free_bsd_filter;
+      ["fetch"], msg, Some freebsd_filter;
+      ["ftp"], msg, Some openbsd_filter
+    ]
+  in
   let open OpamStd.Option.Op in
   let cmd =
     (OpamStd.Env.getopt "OPAMFETCH"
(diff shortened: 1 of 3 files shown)


adamsteen:
@hannesm just to confirm, the easiest way to test this, checkout `hannesm:fetch` and build, then get opam to update/pull in dependencies? I don't have lots of time or brain power at, so i didn't want to waste it figuring out extra stuff.

hannesm:
@adamsteen yes:
- checkout hannesm:fetch branch of opam
- run ./configure --prefix=/tmp/opam/ ; gmake ; gmake install
- setenv PATH /tmp/opam/bin:$PATH
- opam install -vvv <some-not-downloaded-and-installed-package>

in the output of the last step you should see `ftp` being used (and not curl)

adamsteen:
@hannesm I have been deep in solo5/OpenBSD vmm recently and didn't have an opam initialised but

```
asteen:/home/asteen:2221$ opam --version
2.1.0~beta
asteen:/home/asteen:2222$ opam init -vvv
[NOTE] Will configure from built-in defaults.
+ /usr/bin/getconf "_NPROCESSORS_ONLN"
- getconf: _NPROCESSORS_ONLN: unknown variable


+ /usr/bin/ftp "https://opam.ocaml.org/index.tar.gz" "-o" "/tmp/opam-87462-e7966f/index.tar.gz.part" "-U" "opam/2.1.0~beta"
[ERROR] Could not update repository "default": OpamDownload.Download_fail(_, "Download command succeeded, but resulting file not found")
[ERROR] Initial download of repository failed
'opam init -vvv' failed.
```

the `ftp` worked but i think something else with the beta stopped me, not sure.

cfcs:
@adamsteen @hannesm I think the problem might be that you need to use BSD-style argument ordering (options `-o` and `-U` before the freestanding URL), so `-o` and `-u` currently do not get accepted. Why the BSD tool wouldn't complain about extraneous arguments I have no idea, that seems silly.

So the download works, but it will use the default filename of `index.tar.gz` instead of the provided path, which seems consistent with the error message.

In general it would also be good custom to terminate the series of options with `--` which is supported on both GNU tools and FreeBSD `fetch` (I checked), I think OpenBSD `ftp` should also accept this.
Doing so prevents proceeding arguments starting with a dash from being interpreted as options/flags rather than unnamed arguments.

Something like:
```ocaml
let fetch_args = [
  CString "-o", None; CIdent "out", None;
  CString "--user-agent", None; user_agent, None;
  CString "--", None ; (* End list of options *)
  CIdent "url", None;
]

let ftp_args = [
  CString "-o", None; CIdent "out", None;
  CString "-U", None; user_agent, None;
  CString "--", None ; (* End list of options *)
  CIdent "url", None;
]
```

hannesm:
@adamsteen thanks for testing, I opened #3910 to address your issue.

PR #3907, opened 2019-07-01
Make reinstall handling stricter

AltGr:
i.e. don't allow to change packages transitively depending on
something marked for reinstall without doing the reinstall.

The first two patches just add helpers to the library:

- a function to compute a dependency graph directly on the opam
  packages (going through Dose and using back-and-forward conversion
  though). Unused at the moment but may come in handy for some uses
  (e.g. opam2web)

- a Set.fixpoint function, used e.g. to directly compute transitive
  dependency cones directly on opam packages (without going through
  Dose this time)


rjbou:
LGTM thanks!

rjbou:
Doesn't this close #3817?

PR #3910, opened 2019-07-02
pass url last in download commands, this is necessary for ftp at least

hannesm:
I could reproduce the issue seen by @adamsteen while testing `ftp` on OpenBSD -- once the URL is parsed, all other arguments are ignored. This lead to `ftp` downloading the archive, but instead of saving at the location opam intended, it was saved as `filename`, which opam did not expect.


AltGr:
Thanks!

rjbou:
Thanks @adamsteen for testing

PR #3902, opened 2019-06-28
Build the man-pages using dune

AltGr:
this avoids a big annoyance on every `make install`

PR #3911, opened 2019-07-02
Add missing Cmdliner.Manpage.escape calls

dra27:
#3893 uses Windows separators in various opam messages - the ones which are cmdliner terms need to be escaped, which this PR addresses. This PR is also a prerequisite of #3902.

### DIFF
--- a/src/client/opamAdminCommand.ml
+++ b/src/client/opamAdminCommand.ml
@@ -36,7 +36,7 @@ let admin_command_man = [
         holding package definition within subdirectories. A 'compilers%s' \
         subdirectory (opam repository format version < 2) will also be used by \
         the $(b,upgrade-format) subcommand."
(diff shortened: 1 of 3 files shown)


AltGr:
```
File "src/client/opamArg.ml", line 843, characters 43-66:
Error: Unbound module Cmdliner_manpage
```
Need to upgrade cmdliner requirement ?

dbuenzli:
Maybe not. `Cmdliner_manpage` is not an API. Here's [the value](https://erratique.ch/software/cmdliner/doc/Cmdliner.Manpage.html#VALescape) one should be looking for.

dra27:
Oops - my bad. I think I must have done that soon after tracking down that escaping bug and so had the internal module name in my mental cache!

PR #3915, opened 2019-07-03
Fix cppo detection in configure

dra27:
This is a cosmetic change, since `cppo` is a dependency of several of opam's dependencies, but `opam` checks for the `cppo` ocamlfind package, rather than the `cppo` tool itself. This works fine on an opam installation, owing to https://github.com/ocaml/dune/issues/2353, but fails on Debian, where the `cppo` package (correctly) does not install a `META` file.

With this change, `./configure` now doesn't display the "run make lib-ext" message because of a missing `cppo` package. I have changed the detection to look at `$CPPO`, but this is academic since several of the libraries depend on cppo.

This is for master and 2.0

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -4,7 +4,7 @@
 # Shell scripts, autoconf, etc. must have LF endings, even on Windows
 *.sh text eol=lf
 *.zsh text eol=lf
-configure text eol=lf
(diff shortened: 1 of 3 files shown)


rjbou:
lgtm!

PR #3926, opened 2019-07-11
shell/install.sh: Properly update to 2.0.5, fixing "Checksum mismatch" error

0cjs:
The previous commit updating this to 2.0.5 updated the SHA512
hashes to that version, but forgot to update $VERSION and so was
still downloading the 2.0.4 version of $OPAM_BIN, causing a
checksum mismatch.

### DIFF
--- a/shell/install.sh
+++ b/shell/install.sh
@@ -5,7 +5,7 @@ set -ue
 # (c) Copyright Fabrice Le Fessant INRIA/OCamlPro 2013
 # (c) Copyright Louis Gesbert OCamlPro 2014-2017
 
-VERSION='2.0.4'
+VERSION='2.0.5'
 TAG=$(echo "$VERSION" | tr '~' '-')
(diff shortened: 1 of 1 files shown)


0cjs:
This type of error looks easy enough to commit that this script should probably be tested in CI.

Unfortunately, the script is currently interactive. One way to fix this would be to add an additional command-line option for the install location that would disable the interactive prompt (using the install directory specified on the command line instead), allowing a test to install to a scratch directory and then ensure that the file was correctly installed.

AltGr:
Thanks!
This would be difficult to test with CI since it depends purely on the "outside" environment; that script is kept in opam's master just for convenience actually. But it would indeed be good to have tests nevertheless.

0cjs:
To me it doesn't seem difficult at all to test with CI so long as the CI has the ability to download files from github.com. Is that an issue for the current CI system?

If that's not an issue, I can try to find some time to cons up a test.

Issue #3500, opened 2018-08-08
install.sh is hard to use for automated installs


brendanlong:
I'm in the process of switching our CI box over to opam 2.0, which took about an hour and the final version doesn't use the install script at all (we just download the latest binary and drop it in /usr/local/bin). There are a few things that could make this easier for other people in the future:

 - Accept the install dir as an argument instead of requiring user input. It *does* work to do `echo /usr/local/bin | install.sh`, but it's not ideal.
 - Have an option that doesn't install a switch by default (we can't use 4.07 right now because of Core, so building it just adds several minutes to my test cycle right now)
 - Make it possible to pass `--disable-sandboxing` through to `opam init` (or add a `--no-init` option). We do plan to turn on sandboxing, but right now bubblewrap isn't packaged for any distro we use.

AltGr:
Indeed, the script is intended for end-users, and I don't believe it helps in any aspects for CI, compared to just downloading the archive: you don't need automatic selection of the binary, or backup of an old install, or automatic initialisation.

But we should certainly advertise it as such more clearly!
Note that you also have the option of using pre-built Docker images for various configurations from https://hub.docker.com/r/ocaml/opam2/

brendanlong:
That makes sense. It wasn't very hard once I stopped trying to use the installer.

0cjs:
Recent events seem to have demonstrated that help for CI is needed to test this script itself. Yesterday's version update from 2.0.4 to 2.0.5 had a bug in the commit (fix in PR #3926), causing it to download the wrong version and display "Checksum mismatch."

PR #3929, opened 2019-07-11
Sanity checks on shell/install.sh in Travis

dra27:
Addresses https://github.com/ocaml/opam/issues/3500#issuecomment-510399942 and https://github.com/ocaml/opam/pull/3926#issuecomment-510404204

This borrows some stuff from OCaml's Travis script to determine if a PR touches `shell/install.sh`. I've refactored the script to have the data which changes all in one place (which should reduce the temptation to update the sha512s and not the version) but also changed the names to refer to the version in one place. I refactored the URL slightly so that a little of `fgrep` eval could allow the Travis script to determine the URLs in the script.


dra27:
An example of this failing can be seen in [this Travis log](https://travis-ci.org/ocaml/opam/jobs/557382790#L728-L729)

rjbou:
LGTM, thanks @dra27!
I was thinking about exporting the `install.sh` script to its own repository. It is on master, but not consistent with it (it follows last release which can be `master` or `2.0`), and in its current form, don't permit _simply_ to have other releases with their checksum.

dra27:
It shouldn't be too bad to make it work for multiple 2.x versions - the `case` with the checksums could be untied from `$VER` and `$VER` and `$TAG` could be computed as part of command-line parsing. That might be a good idea when 2.1 hits and you could want to install either 2.0.x or 2.1.x.

I wouldn't put it in its own repository, but it could be worth putting it in a distinct branch (so a new one with an unrelated history). See, for example, https://github.com/dra27/opam/tree/shell which I just extracted from opam's `master` branch.

PR #3931, opened 2019-07-12
show: don't require colon for fields

### LINKED ISSUE #3664: opam show -f and field names consistency
Start from [merlin's opam file](https://github.com/ocaml/opam-repository/blob/master/packages/merlin/merlin.3.2.2/opam).
```
$ opam show merlin -f version
3.2.2
$ opam show merlin -f depends
[ERROR] No printer for "depends". Did you mean the opam field "depends:" (with a colon)?
$ opam show merlin -f depends:
"ocaml" {>= "4.02.1" & < "4.08"}
"dune" {build}
"ocamlfind" {>= "1.5.2"}
"yojson"
"craml" {with-test}
```
Is there any reason for that inconsistency? It looks like a bug to me.

rjbou:
opam show no more requires a colon for non _inferred_ fields `:`. Now, by defaults, `opam show pkg -f <field>` displays computed content if it exist, field content otherwise. If `--raw` is specified, it's the field content that is displayed (old `-f <field>:`). For backward compat, `-f <field>:` is still supported. ~it gives the same result than `-f <field>`~

Fixes #3664

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -668,8 +668,9 @@ let show =
                 OpamListCommand.field_names
               ^". Multiple fields can be separated with commas, in which case \
                 field titles will be printed; the raw value of any opam-file \
-                field can be queried by suffixing a colon character (:), e.g. \
-                $(b,--field=depopts:).")
+                field can be queried by combinig with $(b,--raw). For backward \
+                compatibilty, it is possible to query field ending with a colon \
+                (:), it is the same than without.")
         ["f";"field"] in
     Arg.(value & opt (list string) [] & doc) in
   let show_empty =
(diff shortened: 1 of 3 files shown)


AltGr:
> For backward compat, -f <field>: is still supported, it gives the same result than -f <field>.

Shouldn't it behave as `-f <field> --raw` ?

rjbou:
Indeed, updated.

src/client/opamCommands.ml
```diff
@@ -668,8 +668,9 @@ let show =
                 OpamListCommand.field_names
               ^". Multiple fields can be separated with commas, in which case \
                 field titles will be printed; the raw value of any opam-file \
-                field can be queried by suffixing a colon character (:), e.g. \
-                $(b,--field=depopts:).")
+                field can be queried by combinig with $(b,--raw). For backward \
+                compatibilty, querying fieldi by suffixing a colon character \
```
AltGr:
```suggestion
                compatibility, querying fields by suffixing a colon character \
```
(2 typos)
Not sure it's worth mentioning the `backwards compatibility` bit too.

PR #3934, opened 2019-07-17
Further sanity checks

dra27:
This brings #3929 to the 2.0 branch (checking `install.sh` is not strictly necessary, since it shouldn't be used from here, but it also pulls the infrastructure in). It then extends it with some more things from OCaml's CI to verify that `configure` is correctly committed.


dra27:
Example failure in https://travis-ci.org/ocaml/opam/jobs/559897064#L770

dra27:
Right, that's better - I've switched the Travis CI image from Trusty to Xenial, which means it pulls in the correctly patched version of autoconf. We could use the [same patch as OCaml](https://github.com/ocaml/ocaml/blob/trunk/autogen#L17-L29), but it's ugly and I think that @AltGr, @rjbou and I are all using versions of `autoconf` which include that patch anyway.

rjbou:
LGTM, thanks!

dra27:
Cool - I'll open the parallel branch for `master`

PR #3936, opened 2019-07-18
Build man pages using Dune (2.0 version)

dra27:
Rebase and patching of #3902.


dra27:
https://github.com/ocaml/opam/pull/3936/commits/0a29619f9da6e7ad3c35e8cf67d93f314e0460b0 should also go on `master`. It's necessary to ensure that `sudo make install` doesn't run `dune`.

@AltGr - the alternative to that last commit is simply to remove the dependencies of the `install:` target - in other words, you'll get an error if you run `make install` without first running `make [all]`. However, that means you have to do `make && make install` (or `make all install`, strictly without `-j`) which I don't think is what the GNU standards recommend which is that `make install` should build, but not unnecessarily rebuild.

PR #3937, opened 2019-07-19
Build the man pages with Dune

dra27:
Freshly rebased version of #3902


dra27:
@AltGr - you’ll need to blow away AppVeyor’s cache so it knows to rebuild with the cmdliner patch (it only rebuilds if the version is also bumped, and last I looked Daniel had yet done a release with the patch)

dbuenzli:
@dra27 always feel free to ping me if you need stuff, I usually don't make release asap on non critical patches to accumulate a few but I can if it simplifies your life.

dra27:
@dbuenzli - don't worry, I would! It's not critical for us until building the Windows version of opam becomes commonplace (so opam 2.2) at which point if you've had no other reason to do a release of cmdliner in the meantime, I'd definitely ping 🙂

dra27:
Doing a release of `cmdliner` causes lots of people to produce CO₂ rebuilding their switches :wink:

PR #3938, opened 2019-07-19
bump configure to 2.0.5

rjbou:


dra27:
This is already in #3934

Issue #3932, opened 2019-07-15
`opam switch create . --locked`: No available version of ocaml-base-compiler satisfies the constraints


vogler:
Using this [opam file](https://github.com/goblint/analyzer/blob/master/opam) I can do `opam switch create . --deps-only ocaml-base-compiler.4.07.1` to setup a switch with all dependencies.
After `opam lock` I get this [opam.locked file](https://github.com/goblint/analyzer/blob/master/opam.locked).
I expected to be able to setup a switch with the locked versions as shown [here](http://opam.ocaml.org/blog/opam-2-0-beta5/).
However I get: ([full log](https://travis-ci.com/goblint/analyzer/jobs/216124663))
~~~
$ opam switch -y create ./ --deps-only --locked ocaml-base-compiler.4.07.1
[ERROR] Could not resolve set of base packages:
        Your request can't be satisfied:
          - No available version of ocaml-base-compiler satisfies the constraints
~~~
Maybe it's related to #3874 

If I try it without specifying a compiler, I get this strange behavior (no error, but it only installs the compiler and not the rest of the dependencies):
~~~
$ rm -rf _opam
$ opam switch create . --deps-only --locked
[WARNING] Lock file opam.locked is outdated, you may want to re-run opam lock:
          Dependencies present in opam file not in lock file:
            - ounit

[WARNING] Lock file opam.locked is outdated, you may want to re-run opam lock:
          Dependencies present in opam file not in lock file:
            - ounit

The following additional pinnings are required by goblint.dev:
  - goblint-cil.1.7.4 at git+https://github.com/goblint/cil.git
Continue? [Y/n] y
[goblint-cil.1.7.4] synchronised from git+https://github.com/goblint/cil.git
goblint-cil is now pinned to git+https://github.com/goblint/cil.git (version 1.7.4)
goblint is now pinned to file:///home/travis/a (version dev)

<><> Gathering sources ><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
[ocaml-base-compiler.4.07.1] found in cache

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
∗ installed base-bigarray.base
∗ installed base-threads.base
∗ installed base-unix.base
∗ installed ocaml-base-compiler.4.07.1
∗ installed ocaml-config.1
∗ installed ocaml.4.07.1
Done.
~~~

[...]

PR #3939, opened 2019-07-19
lock: add warning in case of missing locked file

rjbou:
Related to #3694 & #3932

### DIFF
--- a/src/state/opamPinned.ml
+++ b/src/state/opamPinned.ml
@@ -128,7 +128,11 @@ let check_locked default =
                       consistent)
               else "")));
        fl)
-    else default
+    else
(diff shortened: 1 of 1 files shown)


src/state/opamPinned.ml
```diff
@@ -128,7 +128,11 @@ let check_locked default =
                       consistent)
               else "")));
        fl)
-    else default
+    else
+      (OpamConsole.warning "Locked file %s doesn't exist, uses %s instead."
```
AltGr:
```suggestion
      (OpamConsole.warning "Lock-file %s doesn't exist, using %s instead."
```

PR #3944, opened 2019-07-28
Fix misspelled environment variable in man page

tchajed:
The usage of this variable is correctly spelled: https://github.com/ocaml/opam/blob/abf8fd8d9e2e0cd743e88d61a3dc82661b733e22/src/repository/opamRepositoryConfig.ml#L121

only the man page is wrong.


dra27:
The fix is good  - but you've rebuilt `configure` on a system without the `runstate` patch. Can you redo the patch without those alterations to `configure`?

tchajed:
Sure, but I did so because CI failed the Hygeine job asking to run `make configure` (actually, it still says that, presumably because I don't have the runstate patch: https://travis-ci.org/ocaml/opam/jobs/564719091).

AltGr:
Thanks!

PR #3866, opened 2019-06-12
Sorting opam fields

rjbou:
Add sorting on opam fields, available as a function in opam-lib & as an `opam show` option.
This pr contains several things:
~- a compare function for formulas~
~- a sort function for formulas~
~- a sort function for filtered_formulas; this one calls also `OpamFilter.simplify_extended_version_formula` on filters~
→ on a seperate PR #3945 
- a sort opam file function:
  - fields are sorted given alphabetical order: author, tags, depends, depopts, depexts, conflicts, pin_depends, extra_files, extra_sources
  - `filtered_fomulas` of `depends`, `depopts`, and `conflicts` are sorted given alphabetical order and their filter : ~`build < <nothing> < test < doc < post`~ alphabetical order too

Related to ocaml/opam2web#173

~_This PR is still a __WIP__. Some discussion needed on **fields sorting**_~

PR #3945, opened 2019-07-31
Sorting formulas functions

rjbou:
Intially, this is part of #3866 

from https://github.com/ocaml/opam/pull/3866#issue-287591110 : 
- a compare function for formulas
- a sort function for formulas
- a sort function for filtered_formulas; this one calls also `OpamFilter.simplify_extended_version_formula` on filters

### DIFF
--- a/src/format/opamFormula.ml
+++ b/src/format/opamFormula.ml
@@ -172,6 +172,27 @@ type version_formula = version_constraint formula
 
 type t = (OpamPackage.Name.t * version_formula) formula
 
+let rec compare_formula f x y =
+  let rec compare_atom x = function
+    | Empty -> 1
+    | Atom y -> f x y
+    | Block y -> compare_atom x y
+    | And (y,z) | Or (y,z) ->
+      let r = compare_atom x y in
+      if r <> 0 then r else compare_atom x z
+  in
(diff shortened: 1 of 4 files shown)


src/format/opamFormula.ml
```diff
@@ -172,6 +172,27 @@ type version_formula = version_constraint formula
...
+  | Empty, Empty -> 0
+  | Empty, _ -> -1
+  | _ , Empty -> 1
+  | Atom x, Atom y -> f x y
+  | Atom x, y -> compare_atom x y
+  | x , Atom y -> -1 * (compare_atom y x)
+  | Block x, y | x, Block y -> compare_formula f x y
+  | (And (x,y) | Or (x,y)), (And (x',y') | Or (x',y')) ->
+    let r = compare_formula f x x' in
+    if r <> 0 then r else compare_formula f y y'
```
hannesm: hmm, doesn't this make "a || b" and "a && b" equal? is this intended that `And` and `Or` are the same?

rjbou: As compare has been made with the idea of inner element comparison only (for sorting opam fields), it was intended. But incorrect for a compare function. Fixed in #3960!

PR #3947, opened 2019-08-01
update: don't update installed dev package that is not pinned

rjbou:
When pinned package is unpinned without removal, its sources continue to be updated on `opam update`.

### DIFF
--- a/src/state/opamSwitchState.ml
+++ b/src/state/opamSwitchState.ml
@@ -403,7 +403,13 @@ let get_package st name =
     OpamPackage.max_version st.packages name
 
 let is_dev_package st nv =
-  match opam_opt st nv with
+  let opam_opt =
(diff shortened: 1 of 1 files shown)


AltGr:
I fear this is not correct for repository dev packages, which may need updating even though they are not pinned. An additional check should be done on the URL from the in-repository package definition (`st.repos_package_index`) in case the package is not pinned.

rjbou:
In fact, a check if the package is pinned is aleady done in [`OpamPackageVar.is_dev_package`](https://github.com/ocaml/opam/blob/master/src/state/opamPackageVar.ml#L117) but only for `http` backend, Shouldn't be better to put the in-repo package definition there?

rjbou:
Related also to #3991, the package is considered as a dev package, so dirty check is launched on a non git directory.

PR #3956, opened 2019-08-15
Add -j jobs to the dune example command

kit-ty-kate:
As per comment of @pictyeye in https://github.com/ocaml/opam-repository/pull/14678#discussion_r314073818


rjbou:
Thanks!

PR #3957, opened 2019-08-15
Do not fail if $HOME/.ccache is missing

mseri:
See https://discuss.ocaml.org/t/dune-1-11-1-compilation-failed/4248


rjbou:
Thanks for the PR!
It remains a tricky case that is not handled by current proposition : ccache configuration file only system wide, `cache_dir` changed in this conf file, and `$HOME/.ccache` directory missing.
Moving existence directory check as a condition for `add_mount` call would handle all cases.

mseri:
Indeed! This was actually checked in `bwrap.sh` in the `add_mounts` function, I have reverted the change and updated `sandbox_exec.sh` moving the check as you suggested.

rjbou:
Thanks! If no objection, i'll change PR description when merging.

mseri:
Sure, do as you see fit!

PR #3971, opened 2019-08-29
Add script for package number plotting

AltGr:
This used to be in 'opam admin stats', but was removed and didn't scale anymore anyway

### DIFF
--- a/shell/pkg-stats.sh
+++ b/shell/pkg-stats.sh
@@ -1,0 +1,36 @@
+#!/bin/sh -ue
+
+if [ ! -d "packages" ] || [ ! -d ".git" ] ; then
+  echo "Please run from a git clone of an opam repository"
+  exit 2
+fi
+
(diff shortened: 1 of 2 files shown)


avsm:
welcome back my old friend! I enjoyed the surprise graph in the platform talk -- and I'm sure @dra27 did as well :-)

shell/pkg-stats.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash -ue
```
avsm: at first glance, this looks like it could be `/bin/sh` as not using anything bash specific (makes it portable)

dra27: I don't think `|&` is portable, although that could be trivially changed. Why not just use the standard `#!/usr/bin/env bash` and `set -ue` below... I know it'll force you to install `bash` on your OpenBSD boxes :wink:

PR #2224, opened 2015-06-26
Adhering to SPDX standard for licence naming

camillem:
Following SPDX standard would  reduce ambiguity, improve interoperability and eases legal compliance.
Thanks !

PR #3976, opened 2019-09-02
Add a lint check for SPDX licenses

AltGr:
and update the documentation.
Supersedes and closes #2224


AltGr:
I feel sorry for the CI :D 
It has a "replay" button, but no "cancel" :/

dbuenzli:
Is there a syntax that allows to silence the warning ?

AltGr:
Unfortunately no, I am not sure if we should add something for custom licenses ?
Lintian has a generic way of declaring "I know I have this warning and am OK about it", but well...

AltGr:
(merging now, but this can ofc still be amended)

dbuenzli:
> Unfortunately no, I am not sure if we should add something for custom licenses ?

I mean those do exist in practice so it would be nice to be able to specify them without having `opam` constantly warn about them. 

For example in `odig` I have a few fonts license that I'm not sure I'm able to recast as `spdx` licenses (have to check again) so at the moment [this](https://github.com/b0-system/odig/blob/f9599095f8291ba0e01b53c33ca2adc888f1b2ac/opam#L6) is what I have so that people are made aware and need to pay attention to the fact  there's a bit more than just `ISC` in there.

PR #3982, opened 2019-09-11
Working-dir fixes

rjbou:
* Don't fetch dev packages when `working-dir` is set
* Copy uncommited modified versioned file

### DIFF
--- a/src/repository/opamGit.ml
+++ b/src/repository/opamGit.ml
@@ -231,6 +231,18 @@ module VCS : OpamVCS.VCS = struct
       OpamProcess.cleanup ~force:true r; Done true
     | r -> OpamSystem.process_error r
 
+  let modified_files repo_root =
+    git repo_root ~verbose:false [ "status" ; "--short" ] @@> fun r ->
+    OpamSystem.raise_on_process_error r;
+    let files =
+      OpamStd.List.filter_map (fun line ->
+          match OpamStd.String.split line ' ' with
+          | ("A" | "M" | "AM")::file::[]
+          | ("R"|"RM"|"C"|"CM")::_::"->"::file::[] -> Some file
+          | _ -> None) r.OpamProcess.r_stdout
+    in
+    Done files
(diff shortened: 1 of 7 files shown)


src/repository/opamGit.ml
```diff
@@ -231,6 +231,18 @@ module VCS : OpamVCS.VCS = struct
...
     | r -> OpamSystem.process_error r
 
+  let modified_files repo_root =
+    git repo_root ~verbose:false [ "status" ; "--short" ] @@> fun r ->
+    OpamSystem.raise_on_process_error r;
+    let files =
+      OpamStd.List.filter_map (fun line ->
+          match OpamStd.String.split line ' ' with
+          | ("A" | "M")::file::[]
+          | ("R"|"RM"|"C"|"CM")::_::"->"::file::[] -> Some file
```
rjbou: Can a file be `copied` and `modified`?

AltGr: I believe so, yes... After a certain threshold of modifications, it will be detected as removed and created. IIRC, this is computed by git after the fact for display, but it doesn't really make a difference to it.

AltGr:
Can you document what this does exactly ?

IIUC, when using `--working-dir`, it uses the VCS on the origin to list the versionned files that have been modified ; then it does a VC sync, followed by a rsync of these files.
Is this correct ? Could you remind me of how it differs from what we had before ?

rjbou:
There was several evolution of the feature. First based on only rsync, but it can quite long if the source contains submodules or have a huge git history. Then moved to git + rsync : ① do a vcs sync ② rm deleted files (versioned or no files) ③ rsync unversioned files + modified-uncommitted  ones.
This PR is a fix for ③, rsync was only done on unversioned files.

rjbou:
And before that, it was an rm deleted files + rsync of versionned ones.

rjbou:
Updated with doc

PR #3992, opened 2019-09-25
config: add 'config setopt' and 'setopt' (sub)commands to modify global config fields

rjbou:
Modifiable fields are `download-command`, `download-jobs`, `jobs`, `best-effort-prefix-criteria`, `solver`, `global-variables`, `eval-variables`, `repository-validation-command`, `solver-criteria`, `solver-upgrade-criteria`, `solver-fixup-criteria`.
On revert mode, value is reverted to its init config value if available, deleted otherwise.

ecit: cf https://github.com/ocaml/opam/pull/3992#issuecomment-557638448

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -870,9 +870,9 @@ let config ?(setopt=false) () =
     "Set the given variable globally or in the current switch. Warning: \
      changing a configured path will not move any files! This command does not \
      perform any variable expansion.";
-    "set-opt", `set_opt, ["[switch|global]"; "NAME"; "[VALUE]"],
+    "set-opt", `set_opt, ["[switch|global]"; "FIELD[(+=|-=|=)VALUE]"],
     "Set the given opam configuration field in the global configuration file.\
-     If $(b,VALUE) is omitted, $(b,FIELD) is set to its default initial \
+     If $(b,op VALUE) is omitted, $(b,FIELD) is set to its default initial \
      configuration, as after a fresh init (use `opam init \
      show-default-opamrc` to display it)";
     "expand", `expand, ["STRING"],
@@ -959,43 +959,31 @@ let config ?(setopt=false) () =
     | Some `list, params ->
       OpamGlobalState.with_ `Lock_none @@ fun gt ->
       (try `Ok (OpamConfigCommand.list gt
-      (List.map OpamPackage.Name.of_string params))
+                  (List.map OpamPackage.Name.of_string params))
        with Failure msg -> `Error (false, msg))
     | Some `set_var, ["sw"|"switch"; var; value] ->
       (wrap_st @@ fun st ->
-      OpamConfigCommand.set_var_switch st
-        (OpamVariable.Full.of_string var) (Some value));
+      OpamConfigCommand.set_var_switch st var (Some value));
       `Ok ()
     | Some `set_var, ["sw"|"switch"; var] ->
       (wrap_st @@ fun st ->
-      OpamConfigCommand.set_var_switch st
-        (OpamVariable.Full.of_string var) None);
+      OpamConfigCommand.set_var_switch st var None);
       `Ok ()
     | Some `set_var, ["gl"|"global"; var; value] ->
       (wrap_gt @@ fun gt ->
-      OpamConfigCommand.set_var_global gt
-        (OpamVariable.Full.of_string var) (Some value));
+      OpamConfigCommand.set_var_global gt var (Some value));
       `Ok ()
     | Some `set_var, ["gl"|"global"; var] ->
       (wrap_gt @@ fun gt ->
-      OpamConfigCommand.set_var_global gt
-        (OpamVariable.Full.of_string var) None);
+      OpamConfigCommand.set_var_global gt var None);
       `Ok ()
-    | Some `set_opt, ["sw"|"switch"; field; value] ->
+    | Some `set_opt, ("sw"|"switch")::fvs ->
       (wrap_st @@ fun st ->
-      OpamConfigCommand.set_opt_switch st field (Some value));
+       List.fold_left OpamConfigCommand.set_opt_switch st fvs);
       `Ok ()
-    | Some `set_opt, ["sw"|"switch"; field] ->
-      (wrap_st @@ fun st ->
-      OpamConfigCommand.set_opt_switch st field None);
-      `Ok ()
-    | Some `set_opt, ["gl"|"global"; field; value] ->
+    | Some `set_opt, ("gl"|"global")::fvs ->
       (wrap_gt @@ fun gt ->
-      OpamConfigCommand.set_opt_global gt field (Some value));
-      `Ok ()
-    | Some `set_opt, ["gl"|"global"; field] ->
-      (wrap_gt @@ fun gt ->
-      OpamConfigCommand.set_opt_global gt field None);
+       List.fold_left OpamConfigCommand.set_opt_global gt fvs);
       `Ok ()
     | Some `expand, [str] ->
       OpamGlobalState.with_ `Lock_none @@ fun gt ->
@@ -1141,31 +1129,26 @@ let config ?(setopt=false) () =
         "Subcommand set is deprecated. Use set-var switch %s %s instead."
         var value;
       (wrap_st @@ fun st ->
-      OpamConfigCommand.set_var_switch st
-        (OpamVariable.Full.of_string var) (Some value));
+      OpamConfigCommand.set_var_switch st var (Some value));
       `Ok ()
     | Some `unset, [var] ->
       OpamConsole.warning
         "Subcommand set is deprecated. Use set-var switch %s instead." var;
       (wrap_st @@ fun st ->
-      OpamConfigCommand.set_var_switch st
-        (OpamVariable.Full.of_string var) None);
+      OpamConfigCommand.set_var_switch st var None);
       `Ok ()
     | Some `set_global, [var; value] ->
       OpamConsole.warning
         "Subcommand set-global is deprecated. Use set-var global %s %s instead."
         var value;
       (wrap_gt @@ fun gt ->
-      OpamConfigCommand.set_var_global gt
-        (OpamVariable.Full.of_string var) (Some value));
+       OpamConfigCommand.set_var_global gt var (Some value));
       `Ok ()
     | Some `unset_global, [var] ->
       OpamConsole.warning
         "Subcommand set-global is deprecated. Use set-var global %s instead."
         var;
-      (wrap_gt @@ fun gt ->
-      OpamConfigCommand.set_var_global gt
-        (OpamVariable.Full.of_string var) None);
+      (wrap_gt @@ fun gt -> OpamConfigCommand.set_var_global gt var None);
       `Ok ()
     | command, params -> bad_subcommand commands ("config", command, params)
   in
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -915,16 +915,6 @@ let config ?(setopt=false) () =
       mk_subcommands commands
   in
   let open Common_config_flags in
-  let wrap_gt set =
-    OpamGlobalState.with_ `Lock_write @@ fun gt ->
-    OpamGlobalState.drop @@ set gt
-  in
-  let wrap_st set =
-    OpamGlobalState.with_ `Lock_none @@ fun gt ->
-    OpamSwitchState.with_ `Lock_write gt @@ fun st ->
-    OpamSwitchState.drop @@ set st;
-    OpamGlobalState.drop @@ gt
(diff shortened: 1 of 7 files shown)


dbuenzli:
Can we get a better name ? There's already `config set` if I'm reading instructions or a script I'm not sure I'd manage to make a difference in my mind between `config set` and `config setopt`.

dbuenzli:
(Also how does it relate to `set-global` ? )

src/client/opamCommands.ml
```diff
@@ -951,6 +962,12 @@ let config =
     | Some `subst, (_::_ as files) ->
       OpamGlobalState.with_ `Lock_none @@ fun gt ->
       `Ok (OpamConfigCommand.subst gt (List.map OpamFilename.Base.of_string files))
+    | Some `setopt, [var;value] ->
+      OpamGlobalState.with_ `Lock_none @@ fun gt ->
```
AltGr: Not `Lock_none` no ?

rjbou:
We could, I'm not completely satisfied with `setopt`. `set-config-field` is redundant & unclear, `set-global-field` is too close to `set-global`, etc. Call for proposal! :)

`set-global` is kind of a subset of `setopt`: `set-global` permits to add variables into the `global-variables` field in the config file, while `setopt` permit to change several fields. `config set-global var_foo bar` is equivalent to `config setopt global-variables var_foo bar` when `global-variables` is empty in `config`.

src/client/opamConfigCommand.mli
```diff
@@ -52,6 +52,13 @@ val set: full_variable -> string option -> unit
 (** Sets or unsets global config variables *)
 val set_global: full_variable -> string option -> unit
 
+(** [set_config_field gt field value] sets global config [field] to [value] in
+    <opamroot>/config file. Modifiable fields are a subset of all defined file
+    in [OpamFile.Config.t]. If the given [value] is [None], [field] is reverted
+    to its initial value as defined in [OpamInitDefaults.init_config], to
+    default value otherwise ([OpamFile.Config.empty]).*)
+val set_config_field: 'a global_state -> string -> string option -> unit
```
AltGr: `rw global_state`

src/client/opamConfigCommand.ml
```diff
@@ -382,6 +382,88 @@ let variable gt v =
...
+        let gt = {gt with config = new_config} in
+        OpamGlobalState.write gt;
+        (), gt
+      in
+      OpamConsole.msg "Field %s %s in %s%s.\n"
+        (OpamConsole.colorise `underline field)
+        (if value = None then "reverted" else "updated")
+        (OpamFile.to_string (OpamPath.config gt.root))
+        (OpamStd.Option.to_string ~none:"" (Printf.sprintf " with value '%s'") value);
+      OpamGlobalState.drop gt
```
AltGr: If the function takes a `gt`, it's probably not its job to discard it ?

src/client/opamCommands.ml
rjbou: The `write` lock is taken in the end, when it is finally needed to write (checks are ok).

dbuenzli:
> `set-global` is kind of a subset of `setopt`: `set-global` permits to add variables into the `global-variables` field in the config file, while `setopt` permit to change several fields. `config set-global var_foo bar` is equivalent to `config setopt global-variables var_foo bar` when `global-variables` is empty in `config`.

Did you try to write the documentation that explains in plain words how all of this works ? I think if you did you would realize this is becoming unwieldy.

Why not try to solve this with a single command and a clear flag with a reasonable default to indicate the context. There's a good example of this in the tool you used to create your PR namely:

```
git config [--get | --set] [--local | --global | --system | --worktree ] NAME [VALUE]
```

I certainly don't have in my mind all the config nuts and bolt `opam` allows but something along these lines could certainly do:

```
opam config set [--switch | --global | --global-var ] NAME [VALUE]
opam config get [--switch | --global | --global-var ] NAME [VALUE]
opam config delete [--switch | --global | --global-var ] NAME
```
In fact `-global-var` may not even be needed if you have a shorter name for this dictionary in the global config file and you introduce a dot notation for `NAME`. For example setting the global variable `VAR` could become: 

```
opam config set --global vars.VAR VALUE
```

src/client/opamConfigCommand.ml
```diff
@@ -382,6 +382,88 @@ let variable gt v =
       "Variable %s not found"
       (OpamVariable.Full.to_string v)
 
+let set_config_field gt field value =
+  let open OpamStd.Op in
+  let open OpamStd.Option.Op in
+  match OpamStd.List.find_opt ((=) field @* fst) OpamFile.Config.fields with
+  | None ->
+    OpamConsole.error_and_exit `Bad_arguments
+      "Field %s don't exists" (OpamConsole.colorise `underline field)
```
AltGr: `doesn't exist` :)

AltGr:
Thanks!
Indeed, maybe things would be made clearer if we rename `set` to `setvar`, and `set-global` to `set-global-var` (or `setvar-global` ?) ; but we'll of course have to keep aliases for the older versions.
It's important to keep setting options and variables separate, for one thing to avoid typos on options, and also because a variable name might be conflictng with an option (e.g. `jobs`, very bad example which shows how much confusion this can create)

AltGr:
I kind of like the idea of an `opam config set var.XXX|config.XXX` (although `opam config set config.jobs` is a little redundant ...)

dbuenzli:
> It's important to keep setting options and variables separate,

Ok that difference what not entirely clear to me. 

But then again you can distiguish that with a flag and keep everything as it is now except deprecate `*-global` subcommands -- that way of doing really doesn't scale to multiple context and types. So basically you could have:

```
opam config set [--switch | --global ] [--var | --opt] NAME VALUE
opam config get [--switch | --global ] [--var | --opt] NAME 
opam config delete [--switch | --global ] [--var | --opt] NAME
```
with a default to `--var` and `--switch` to keep backward compat.

dbuenzli:
Yet another design is `{set,get,delete}-{var,opt} [--switch | --global ]` and deprecate all the existing `set*` functions. 

In the long term this way may be less error prone, with the design of the previous message I can see people writing `set jobs 4` when they wanted to write `set --opt jobs 4`.

rjbou:
I agree that it can be seen unwieldy, but I try to keep `opam config` backward compatible. `set/setglobal` and  `setopt` don't target the same modifications in file. That's why I used different naming in the manpage (field vs var), to explain what is modified by `setopt`.
While `config set var.xxx | config set --options`  is clearer, backward compatibility won't be as simple as setting default to switch and variable. `set` and `set-global` update or add the already present variable field in the switch/global config.

>  Ok that difference what not entirely clear to me.

`setopt` is intended to set [`config`](https://opam.ocaml.org/doc/Manual.html#config) file fields, while `set` and `set-global` are resp. intended to update [`global-variables`](https://opam.ocaml.org/doc/Manual.html#configfield-global-variables) and [`variables`](https://opam.ocaml.org/doc/Manual.html#switchconfigsection-variables) fields.

>  deprecate all the existing set* functions.

If we can deprecate them, the whole would be more coherent, ofc. Maybe for next LTS?

dbuenzli:
> I agree that it can be seen unwieldy, but I try to keep `opam config` backward compatible.

None of what I proposed so far makes `opam config` backwards incompatible. 

> `setopt` is intended to set [`config`](https://opam.ocaml.org/doc/Manual.html#config) file fields, > while `set` and `set-global` are resp. intended to update [`global-variables`](https://opam.ocaml.org/doc/Manual.html#configfield-global-variables) and [`variables`](https://opam.ocaml.org/doc/Manual.html#switchconfigsection-variables) fields.

There's also a [switch config](https://opam.ocaml.org/doc/Manual.html#switch-config) so what you propose should rather be called `setopt-global`, but then if you want to keep a bit of sanity for end-users I think [this](https://github.com/ocaml/opam/pull/3992#issuecomment-535075222) is the way to go.

src/client/opamConfigCommand.mli
rjbou: Constrained to `unlocked`

rjbou:
> None of what I proposed so far makes opam config backwards incompatible.

Let's take the problematic case: `config set` backward compatibility. Now, `config set var value` *updates*  `var` in switch config variables. If we have `config set --switch --var var value`, that overwrites the variable `var`, we loose previous behavior. One solution to that would be add a `+` mechanism, but it can't be the default for `config set --switch --var var value`.
The problem is just on the naming. What you proposed [here](https://github.com/ocaml/opam/pull/3992#issuecomment-535075222) could be ok, I didn't see the dash after `set` at first reading.

dbuenzli:
> Let's take the problematic case: `config set` backward compatibility. Now, `config set var value` _updates_ `var` in switch config variables. If we have `config set --switch --var var value`, that overwrites the variable `var`, we loose previous behavior.

No `config set var value` remains the same, as I said here [here](https://github.com/ocaml/opam/pull/3992#issuecomment-535073138) `--var` and `--switch` would be the default if unspecified. So the two invocations you mention are equal.

In any case now that `set` has been used for variables I think the best solution is to move to this [this](https://github.com/ocaml/opam/pull/3992#issuecomment-535075222). (Otherwise we could have gone the `set OPT` `set var.VAR` route which still supports name checking but I bet it's too late now).

src/client/opamConfigCommand.mli
```diff
@@ -52,6 +52,13 @@ val set: full_variable -> string option -> unit
 (** Sets or unsets global config variables *)
 val set_global: full_variable -> string option -> unit
 
+(** [set_config_field gt field value] sets global config [field] to [value] in
+    <opamroot>/config file. Modifiable fields are a subset of all defined file
+    in [OpamFile.Config.t]. If the given [value] is [None], [field] is reverted
+    to its initial value as defined in [OpamInitDefaults.init_config], to
+    default value otherwise ([OpamFile.Config.empty]).*)
+val set_config_field: [< unlocked ] global_state -> string -> string option -> unit
```
AltGr: Noooo we need a write lock to write!!

src/client/opamConfigCommand.ml
```diff
@@ -382,6 +382,87 @@ let variable gt v =
...
+        | None ->
+          if OpamConsole.confirm ~default:false
+              "Are you sure you want to revert %s value"
+              (OpamConsole.colorise `underline field)
+          then
+            set_default config
+          else config
+      in
+      let _, _gt =
+        OpamGlobalState.with_write_lock gt @@ fun gt ->
```
AltGr: Don't use lock upgrade.
Also, it's useless to lock just for the write (the files are individually locked already at the OpamFile layer when writing), what we need it to keep the lock for the whole time between load and write, so that we are sure not to interfere with concurrent writes.

src/client/opamCommands.ml
AltGr: Please don't use lock upgrade unless you have to (which should only happen when you normally don't need to lock but there is a specific case, like format upgrade). Take all the locks you need ASAP (always in the same order), then discard them ASAP once you don't need them. Lock upgrade can cause deadlocks.

src/client/opamConfigCommand.ml
rjbou: After some discussion, `OpamConfigCommand` function will ben updated to always take & returns the state

rjbou:
New version adopting the `opam config set-{opt,var} {switch,global}` mechanism.
##### `set-var var value`
It has the same usage as `config {set,unset,set-global,unset-global}`, modifies `global-variables` with `global` and `variables` with `switch`. If `value` is missing it removes `var`. As discussed with @AltGr, those fields can only be modified using this command

##### `set-opt {switch|global} field value` 
Modifier of a subset of config fields. Some fields are fixed (e.g. jobs), other appendable/modifiable. For the latter, the real value can be preceded by `+=`, `_=`(*), or `=`, to resp. add, remove, and overwrite (default). 

It uses the field parsers, so `value` need to be as in the file: for some field, extra quotation marks could be needed, or surrounding square brackets.

(*) First i wanted to use `-=`, but it is catched by cmdliner. `_=` is a temporary "solution"... I didn't dig into that yet

rjbou:
New new (last?) syntax for `set-opt`.

##### `set-opt {switch|global} field[(+=|-=|=)value]` 
Modifier of a subset of config fields. Some fields are fixed (e.g. jobs), other appendable/modifiable. For the latter, operators `+=`, `-=`, or `=` respectively add, remove, and overwrite. If only `field` is specified, it revert the value to its default.

It uses the field parsers, so `value` need to be as in the file: for some field, extra quotation marks could be needed, or surrounding square brackets.

AltGr:
Quick notes from today's discussion:
- handle setting vars with quotes (escaping issue)
- `config set-var switch foo xxx` may duplicate `foo` if already defined

Otherwise great!

PR #3999, opened 2019-09-30
Avoid copying extraneous files in opam-devel example

maroneze:
I just did `opam install opam-devel` to upgrade from 2.0.4 to 2.0.5, and in the end there was this message:
> You should not run it from there, please install the binaries to your PATH, e.g. with
>  `sudo cp <OPAMROOT>/4.05.0/lib/opam-devel/* /usr/local/bin`

I know it's just an example, but that command would actually copy the `META` and `dune-package` files as well. And since there is only a single binary there (`opam`), it seems better to just do:

       sudo cp <OPAMROOT>/4.05.0/lib/opam-devel/opam /usr/local/bin

Unless there is a use case in which `opam-devel` has more than one binary file. If so, please ignore this PR.

### DIFF
--- a/opam-devel.opam
+++ b/opam-devel.opam
@@ -28,7 +28,7 @@ depends: [
 ]
 post-messages: [
 "The development version of opam has been successfully compiled into %{lib}%/%{name}%. You should not run it from there, please install the binaries to your PATH, e.g. with
-    sudo cp %{lib}%/%{name}%/* /usr/local/bin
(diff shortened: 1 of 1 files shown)


AltGr:
Ah, indeed: there used to be a `opam-admin` there too (and no META or other dune artifacts). But this no longer makes much sense. Thanks!

There might be a need for a `.exe` here in _some_ cases but well... :)

PR #4002, opened 2019-10-07
admin: add constraint on depopts

### LINKED ISSUE #3994: `opam admin add-constraint` doesn't add constraints for optional dependencies
At the moment `opam admin add-constraint` only modifies the `depends` field of each opam file but this is not complete. Some packages might depend optionally to the one you want to constrain and thus require their `conflicts` field to be modified.

For instance `opam admin add-constraint lwt<3.0.0` should modify `tls.0.9.3` because it depends on lwt optionally:
```
depopts: [
  "lwt"
  [...]
]
conflicts: [
  "lwt" {<"2.4.8"}
  [...]
]
```
If the command would have handled optional dependencies then the opam would look like this after running the command (or something equivalent):
```
depopts: [
  "lwt"
  [...]
]
conflicts: [
  "lwt" {<"2.4.8"}
  "lwt" {>= "3.0.0"}
```

_Tested with opam 2.0.5_

rjbou:
fixes #3994

### DIFF
--- a/src/client/opamAdminCommand.ml
+++ b/src/client/opamAdminCommand.ml
@@ -974,16 +974,16 @@ let add_constraint_command =
     OpamArg.apply_global_options global_options;
     let repo_root = checked_repo_root () in
     let pkg_prefixes = OpamRepository.packages_with_prefixes repo_root in
-    let name, cstr = atom in
-    let cstr = match cstr with
+    let name, cstr_opt = atom in
+    let cstr = match cstr_opt with
       | Some (relop, v) ->
         OpamFormula.Atom
           (Constraint (relop, FString (OpamPackage.Version.to_string v)))
       | None ->
         OpamFormula.Empty
     in
-    let add_cstr nv n c =
-      let f = OpamFormula.ands [c; cstr] in
+    let add_cstr op cstr nv n c =
+      let f = op [ cstr; c] in
       match OpamFilter.simplify_extended_version_formula f with
       | Some f -> f
       | None -> (* conflicting constraint *)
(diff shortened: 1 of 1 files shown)


AltGr:
I am not sure what is the best approach to this. The proposed PR updates depopts just like dependencies, which makes sense, but as @kit-ty-kate pointed out, it could be more convenient and less error-prone to update conflicts instead.

The reason is that adding a constraint on a dependency in practice adds _conflicting versions_ to your package, while adding a constraint to a depopt merely disables the depopt (since 2.0 anyway).

So e.g. say you have a package with a depopt on `foo`, and you want to add the constraint `< "2.0"`: is it more likely that you just want to not consider `foo.2.0` a depopt, or that you want to guarantee that `foo`, if installed, is a lower version ?

Since opam only enforces scope up to what is declared, but doesn't hide stuff (i.e. your depends/depopts are guaranteed not to change from your package build to its removal, but you might be able to see any other package that you haven't declared a dependency on while building), and dependencies are commonly handled with `./configure` scripts that will detect the dependency without the version checks, my experience was that if we just ignore the depopt, the package will try to build _with_ `foo.2.0` and then fail.

This would strongly push towards the "add conflict" solution, esp. since `opam admin add-constraint` most common use-case is to introduce a new, incompatible version of `foo`. But maybe the context has changed a bit, since `dune` has replaced many `./configure` scripts already ?

rjbou:
Updated with conflcts

src/client/opamAdminCommand.ml
```diff
@@ -1005,13 +1005,38 @@ let add_constraint_command =
...
               | (n,c as atom) ->
-                if n = name then Atom (n, (add_cstr nv n c))
+                if n = name then Atom (n, (add_cstr cstr nv n c))
                 else Atom atom)
             deps0
         in
-        if deps <> deps0 then
+        let depopts0 = OpamFile.OPAM.depopts opam in
+        let conflicts0 = OpamFile.OPAM.conflicts opam in
+        let contains =
```
AltGr: `contains_name` ?

AltGr: Or even `contains name` actually, would be easier to read.

src/client/opamAdminCommand.ml
```diff
@@ -1005,13 +1005,38 @@ let add_constraint_command =
...
+          OpamFormula.fold_left (fun contains (n,_) ->
+              contains || n = name) false
+        in
+        let conflicts =
+          if contains depopts0 then
+            let icstr =
+              match cstr with
+              | OpamFormula.Atom (Constraint (relop, FString v)) ->
+                OpamFormula.Atom (Constraint (OpamFormula.neg_relop relop, FString v))
+              | _ -> OpamFormula.Empty
```
AltGr: Shouldn't this be an `assert false` ?
That or just use `OpamFormula.neg (fun (op,v) -> OpamFormula.neg_relop op, v) cstr` ?
(See `admin-scripts/depopts_to_conflicts.ml`)

src/client/opamAdminCommand.ml
```diff
@@ -1005,13 +1005,38 @@ let add_constraint_command =
...
+            in
+            if contains conflicts0 then
+              OpamFormula.map (function
+                  | (n,c as atom) ->
+                    if n = name then Atom (n, (add_cstr icstr nv n c))
```
AltGr: `add_cstr` creates an AND, here we would need to create an OR instead
(if we had `conflicts: foo < 3` and we add the constraint `foo < 4`, we then need to conflict with `foo (< 3 | >= 4)` — `(< 3 & >= 4)` would be the empty set and actually cancel the existing conflicts)

src/client/opamAdminCommand.ml
rjbou: better use the first `cstr` option definition to have only the two ways: a constraint is done → construct its neg, a constraint is released → remove package from conflicts (this case wasn't handled)

src/client/opamAdminCommand.ml
```diff
@@ -1005,13 +1005,43 @@ let add_constraint_command =
...
+                    | (n,c as atom) ->
+                      if n = name then Atom (n, (add_cstr OpamFormula.ors icstr nv n c))
+                      else Atom atom)
+                  conflicts0
+              else
+                OpamFormula.ors [ conflicts0; Atom (name, icstr) ]
+            | None ->
+              OpamFormula.map_formula (function
+                  | Atom (n,_) as f -> if n = name then Empty else f
+                  | f -> f) conflicts0
```
AltGr:
```suggestion
            | None -> conflicts0
```
As per the manpage: 
"If no version constraint is specified, the command will just simplify existing version constraints on dependencies to the named package."
So no reason to remove the conflict in this case.

AltGr:
LGTM apart from the small change above

rjbou:
updated!

PR #4004, opened 2019-10-08
Determine jobs number at launch (state config launch)

### LINKED ISSUE #3986: Number of jobs hardcoded at init
Opam can detect the number of processors on the machine, and uses `nprocs - 1` as the default for the `jobs` parameter.

However, this detection is done only once at `opam init`, and then coded in the config. This is a problem for e.g. docker images that get moved around (and can be generated on machines with many cores). Instead the value should not be set in the config by default, in which case a dynamic value is computed (as for `os-distribution`, etc.)

rjbou:
Fixes #3986

### DIFF
--- a/src/state/opamSysPoll.ml
+++ b/src/state/opamSysPoll.ml
@@ -158,3 +158,6 @@ let variables =
       "os-version", os_version_lazy;
       "os-family", os_family_lazy;
     ]
+
+let cores_lazy = lazy (OpamSystem.cpu_count ())
+let cores () = Lazy.force cores_lazy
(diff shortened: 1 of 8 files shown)


src/state/opamSysPoll.ml
```diff
@@ -158,3 +158,6 @@ let variables =
       "os-version", os_version_lazy;
       "os-family", os_family_lazy;
     ]
+
+let cores_lazy = lazy (OpamSystem.cpu_count ())
+let cores () = Lazy.force cores_lazy
```
AltGr: I prefer to make `cores_lazy` local in this case:
```
let cores =
  let cores_lazy = ... in
  fun () -> Lazy.force ...
```
(but this is fine; the idea is just to make it clear that cores_lazy won't be otherwise used)

rjbou: It is to have the same _pattern_ as other lazy values.

PR #4007, opened 2019-10-09
update: don't fail if dev source is not vcs, just set it as non dirty dev package

### LINKED ISSUE #3991: `opam update` stops working if an installed package's url changes from tarball to git
As I discovered today while moving packages around in the Solo5 staging repository, if a package is installed in an opam switch, and that package's `url` changes from a release tarball to a Git repository, `opam update` will start failing with this message:

```
$ opam update
[ERROR] Command "/usr/bin/git diff --no-ext-diff --quiet" failed:
"/usr/bin/git diff --no-ext-diff --quiet" exited with code 129
```

```
$ cat ~/.opam/log/log-20708-6d07ae.info 
# path        ~/.opam/solo5-4.09.0/.opam-switch/sources/solo5-bindings-hvt.0.6.2
# command     /usr/bin/git diff --no-ext-diff --quiet
# env-file    ~/.opam/log/log-20708-6d07ae.env
# output-file ~/.opam/log/log-20708-6d07ae.out
$ cat ~/.opam/log/log-20708-6d07ae.out 
Not a git repository
To compare two paths outside a working tree:
usage: git diff [--no-index] <path> <path>
```

I had to `strace` opam to find the path to the above log file, it would also be useful to give a pointer to this if commands such as this fail.

rjbou:
If from a repo, primary url backend changes from archive to vcs, on update, as the package is considered `dev`, `is dirty` check fails. 

Fixes #3991

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -580,8 +580,12 @@ let update
              in
              match OpamSwitchState.primary_url st nv with
              | Some { OpamUrl.backend = #OpamUrl.version_control as vc; _ } ->
-               OpamProcess.Job.run @@
-               OpamRepository.is_dirty { cache_url with OpamUrl.backend = vc }
+               (try
+                  OpamProcess.Job.run @@
+                  OpamRepository.is_dirty { cache_url with OpamUrl.backend = vc }
+                with OpamSystem.Process_error _ ->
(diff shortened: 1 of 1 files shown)


AltGr:
Thanks!

PR #4008, opened 2019-10-09
repo: fall back in case archive doesn't exist

rjbou:
When changing a repo url, the previous archive is discarded, but later opam tries extracted it (caching in `repo.repos_tmp`). Adds a fall back for that case.

### DIFF
--- a/src/state/opamRepositoryState.ml
+++ b/src/state/opamRepositoryState.ml
@@ -167,13 +167,20 @@ let load_repo repo repo_root =
     (t ());
   repo_def, opams
 
+let clean_repo_tmp tmp_dir =
+  if Lazy.is_val tmp_dir then
+    let d = Lazy.force tmp_dir in
+    OpamFilename.cleandir d;
+    OpamFilename.rmdir_cleanup d
+
+let remove_from_repos_tmp rt name =
+  OpamStd.Option.iter (fun tmp_dir ->
+      clean_repo_tmp tmp_dir;
+      Hashtbl.remove rt.repos_tmp name)
+    (Hashtbl.find_opt rt.repos_tmp name)
+
 let cleanup rt =
-  Hashtbl.iter (fun _ tmp_dir ->
-      if Lazy.is_val tmp_dir then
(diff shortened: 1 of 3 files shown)


src/state/opamRepositoryState.ml
```diff
@@ -214,8 +214,12 @@ let load lock_kind gt =
       then
         let tmp = lazy (
           let tmp_root = Lazy.force repos_tmp_root in
-          OpamFilename.extract_in tar tmp_root;
-          OpamFilename.Op.(tmp_root / OpamRepositoryName.to_string name)
+          if OpamFilename.exists tar then
+            (OpamFilename.extract_in tar tmp_root;
+             OpamFilename.Op.(tmp_root / OpamRepositoryName.to_string name))
+          else
```
AltGr: So IIUC, this will only be triggered if the tar file has been removed since the repo state was loaded (and there was no uncompressed root when it loaded). Where will the `Not_found` be caught ?

It  might be safer to better protect the functions that can remove the tar file (at worst, force a full repo state reload — I agree that this design with hidden state within lazies in the repo state may make correct updates more complicated, esp. with our finalisers... we would need to keep the old binding for cleanup, but not use it anymore when `get_root`/`get_repo_root` is called) ?

src/state/opamRepositoryState.ml
```diff
@@ -177,9 +177,9 @@ let cleanup rt =
   Hashtbl.clear rt.repos_tmp
 
 let get_root_raw root repos_tmp name =
-  match Hashtbl.find repos_tmp name with
-  | lazy repo_root -> repo_root
-  | exception Not_found -> OpamRepositoryPath.root root name
+  try
+    Lazy.force (Hashtbl.find repos_tmp name)
+  with Not_found -> OpamRepositoryPath.root root name
```
AltGr: Err this doesn't change anything right ?

src/state/opamRepositoryState.ml
rjbou: cleanup is now done on `set_url` call

PR #4014, opened 2019-10-16
1.2 upgrade fix

rjbou:
Config version was set to the current opam version, without updating the config file `format_version`, leading to an error in the `opam-version` field check (`2.1 > 2.0`). On upgrade, instead of setting the current version to config, set the min between current version and lastest compatible config version.

### DIFF
--- a/src/state/opamFormatUpgrade.ml
+++ b/src/state/opamFormatUpgrade.ml
@@ -1046,7 +1046,7 @@ let from_2_0_beta5_to_2_0 _ conf = conf
 
 let v2_1 = OpamVersion.of_string "2.1"
 
-let from_2_0_to_2_1 _ conf = conf
+let _from_2_0_to_2_1 _ conf = conf
 
 let lastest_compatible_switch_version = OpamFile.Switch_config.format_version
 let latest_version = v2_1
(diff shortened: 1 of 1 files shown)


src/state/opamFormatUpgrade.ml
```diff
@@ -1089,12 +1090,17 @@ let as_necessary global_lock root config =
     then
       let update_to v f config =
         if OpamVersion.compare config_version v < 0 then
-          let config = f root config |> OpamFile.Config.with_opam_version v in
-          (* save the current version to mitigate damage is the upgrade goes
-             wrong afterwards *)
-          OpamFile.Config.write (OpamPath.config root)
-            (OpamFile.Config.with_opam_version v config);
-          config
+        let v =
```
AltGr: * indentation
* wouldn't it be safer to just skip the `update_to v2_1`, which is a no-op anyways ? Maybe this should be an assert instead, I don't see the case where we would want to perform a (non no-op) upgrade but without updating the version number... seems dangerous

rjbou: It was on the same scheme than `2.0~beta5` no-op upgrade & the last compatible version as done on switches.
Concretely, doing this no-op upgrade, with the last compatible version, have the same result than not doing the upgrade. A case can be handled by this mechanism, is if a field is updated, but without breaking retro-compatibility... but i don't think we want to allow this case.

AltGr: yup! The original idea was that creating the file would yield the new version, but running with the older version wouldn't change anything. I don't think that was very useful, so we better avoid it now.

If in the future we want to add an update that doesn't break compat, and doesn't require a version bump... we can always be extra careful and change the checks at that point. It'd still be weird because in that case, how would we avoid running the upgrade every time ?

rjbou: As it is based on `OpamVersion.(compare OpamFile.Config.format_version config_version)`, and they remain the same with the latest compatible version, we don't enter the upgrade.

rjbou:
Updated

PR #4015, opened 2019-10-16
travis: add a stage to test opam1.2 upgrade

rjbou:
To merge after #4014

### DIFF
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -265,6 +265,7 @@ export OPAMYES=1
 export OCAMLRUNPARAM=b
 
 ( # Run subshell in bootstrap root env to build
+  echo -en "travis_fold:start:build\r"
   if [[ $OPAM_TEST -eq 1 ]] ; then
     export OPAMROOT=$OPAMBSROOT
     eval $(opam env)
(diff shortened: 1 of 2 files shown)


.travis-ci.sh
```diff
@@ -296,13 +301,59 @@ export OCAMLRUNPARAM=b
...
+  CACHE=$OPAM12DIR/cache
+  export OPAMROOT=$CACHE/root20
+  if [[ ! -f $CACHE/bin/opam ]]; then
+    cd $OPAM12DIR
+    echo -en "travis_fold:start:opam12\r"
+    opam init --bare default git+https://github.com/ocaml/opam-repository#8be4290a
+    opam switch create opam12 --empty
+    eval `opam env`
+    OPAMEDITOR="sed -i -e 's/printconf/\"printconf\"/'" opam pin -n --edit dose 3.4.2 << EOF
+EOF
```
AltGr: ??

AltGr: If you just want a closed stdin, it would be more idiomatic to use `CMD <&-`. For an empty (but not closed) channel, maybe you could prefer `CMD </dev/null`.

.travis-ci.sh
```diff
@@ -296,13 +301,59 @@ export OCAMLRUNPARAM=b
...
+    mkdir -p $CACHE/sources
+    cd $CACHE/sources
+    wget https://github.com/ocaml/opam/archive/1.2.2.tar.gz -O opam1.2.tar.gz
```
AltGr: Wouldn't it be OK to dl the binary here ? opam 1.2 isn't supposed to be patched anymore ?

rjbou: Why make it simple when we can make it complicated... \o/

.travis-ci.sh
rjbou: yes that's it, to not block travis with a waiting `enter` from the `--edit`

PR #4016, opened 2019-10-17
Factorize option_map and option_default in `core/opamProcess.ml`

nobrakal:
Pretty simple PR here, it reuses some definitions of `opamStd` and remove locally defined ones.

A little bit of self-promotion, this redundancy was detected while testing an (experimental) library called [`asak`](https://github.com/nobrakal/asak).


rjbou:
Thanks!

PR #4019, opened 2019-10-22
sandbox: add dune cache as rw

### LINKED ISSUE #4012: Enable read/write access to dune's cache directory when sandboxing.
Currently Opam's sandboxing prevent dune caching mechanism from working, as sandboxing removes write access to most of the filesystem.

The currently proposed solution would be to simply add the cache directory to the list of rw directiories in `sandbox.sh`. The value on Unix and OSX would be `$XDG_CACHE_HOME/dune` if set, `$HOME/.cache/dune` otherwise. Windows would be `$AppData/Local Settings/Cache`, although I'm not sure sandboxing works there.

rjbou:
Fixes #4012

### DIFF
--- a/src/state/shellscripts/bwrap.sh
+++ b/src/state/shellscripts/bwrap.sh
@@ -73,6 +73,12 @@ add_ccache_mount() {
   fi
 }
 
+add_dune_cache_mount() {
+  DUNE_CACHE=${XDG_CACHE_HOME:-$HOME/.cache}/dune
+  mkdir -p ${DUNE_CACHE}
+  add_mounts rw $DUNE_CACHE
+ }
+
 # This case-switch should remain identical between the different sandbox implems
 COMMAND="$1"; shift
 case "$COMMAND" in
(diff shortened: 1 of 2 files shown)


src/state/shellscripts/bwrap.sh
```diff
@@ -93,6 +100,7 @@ case "$COMMAND" in
         add_mounts rw "$OPAM_SWITCH_PREFIX"
         add_mounts ro "$OPAM_SWITCH_PREFIX/.opam-switch"
         add_mounts rw "$PWD"
+        add_dune_cache_mount
```
rjbou: is it needed for install ? I don't think but not sure

AltGr: In principle, the "install" command is not even supposed to write to the source directory (this is a good rule even for `make`). It's not always true, and many packages write e.g. log files though, so we keep it rw.

I think in this case it's good hygiene to keep the mount only for the build. Anything breaking would be suspicious, and we can investigate and decide to be less strict at that point. (Anyway, packages don't use `dune` with the `install:` field usually, since that's done through `.install` files !)

PR #4026, opened 2019-11-07
make cold, check that patch is not from BusyBox

### LINKED ISSUE #4021: make lib-ext fails patching cmdliner on alpine / busybox
```
# git describe
2.0.0-531-g2ab2a6e4

# make lib-ext
make -j -C src_ext lib-ext
make[1]: Entering directory '/tmp/opam/src_ext'
mkdir -p tmp-cmdliner
cd tmp-cmdliner && bunzip2 -c ../cmdliner.tbz | tar xf -
rm -f seq/src/seq.ml seq/src/seq.mli
rm -rf cmdliner
patching file src/cmdliner_docgen.ml
patching file _tags
Hunk 1 FAILED 1/1.
-true : bin_annot, safe_string
+true : bin_annot, safe_string, package(bytes), package(result)
 <src> : include
-<test> : include
patching file pkg/META
[...]
make[1]: *** [Makefile:173: cmdliner.stamp] Error 1
make[1]: Leaving directory '/tmp/opam/src_ext'
make: *** [Makefile:60: lib-ext] Error 2
```

Here's a `Dockerfile` that shows the problem:

```
FROM alpine:3.10
RUN apk update && apk upgrade
RUN apk add build-base bzip2 git tar curl ca-certificates openssl
RUN git clone git://github.com/ocaml/opam /tmp/opam
WORKDIR /tmp/opam
RUN make lib-ext
```

Running `make cold` instead of `make lib-ext` fails in the same way (but takes much longer).

rjbou:
Fixes #4021

### DIFF
--- a/configure
+++ b/configure
@@ -4828,6 +4828,10 @@ fi
 if test "x${COLD_CHECK}" = "xyes" ; then
   if test "x$PATCH"  = "x" ; then
     as_fn_error $? "You must have patch installed." "$LINENO" 5
+  else
+    if test "x`patch --version 2>&1 | grep BusyBox`" != "x" ; then
+      as_fn_error $? "You must have a standalone patch installed, not from BusyBox." "$LINENO" 5
+    fi
   fi
   if test "x$BUNZIP2"  = "x" ; then
     as_fn_error $? "You must have bunzip2 installed." "$LINENO" 5
--- a/configure.ac
+++ b/configure.ac
@@ -233,6 +233,10 @@ AC_CHECK_TOOL(BUNZIP2,bunzip2)
 if test "x${COLD_CHECK}" = "xyes" ; then
   if test "x$PATCH"  = "x" ; then
     AC_MSG_ERROR([You must have patch installed.])
+  else
+    if test "x`patch --version 2>&1 | grep BusyBox`" != "x" ; then
+      AC_MSG_ERROR([You must have a standalone patch installed, not from BusyBox.])
+    fi
   fi
   if test "x$BUNZIP2"  = "x" ; then
     AC_MSG_ERROR([You must have bunzip2 installed.])


XVilka:
Seems red on AppVeyor:
```
<><> Creating initial switch (ocaml-system>=4.02.3) <><><><><><><><><><><><><><>
Fatal error:
"cmd": permission denied.
[ERROR] No switch is currently set. Please use 'opam switch' to set or install a switch
'opam install -y -v ocamlfind' failed.
make: *** [Makefile:218: run-appveyor-test] Error 50
make: Leaving directory '/cygdrive/c/projects/opam'
Command exited with code 1
```

rjbou:
yes, this error is present since few prs now... a problem with the cygwin job

dra27:
This is the wrong fix - we should fix the patch file, not reject the tool! See #4028

rjbou:
Ok for #4028. But busybox patch contains unfixed bugs and doesn't seem to be an up-to-date project (see links in https://github.com/ocaml/opam/issues/4021#issuecomment-551180898). As there is the alternative to install `patch`, better rely on the latter.

dra27:
Neither of those bugs are a necessary reason to *require* another `patch` if we provide working patch files. If we know busybox `patch` is a lowest common denominator, it would seem better just to test with it, given that `busybox` is packaged in distros? The `configure` logic also prevents opam using it even if it is fixed. `busybox` had a release a few weeks ago, and it's embedded in a lot of systems - it's not exactly inactive!

rjbou:
I didn't intend to say that the project is inactive, just that the errors concerning patch has not been addressed. But if we can have busybox _compatible_  patches, it's better, I agree. As I said, i'm ok with dropping this PR in favor to #4028.

PR #4032, opened 2019-11-26
Use install command instead of cp and chmod

imbsky:
The install command seems easier and more suitable.
https://linux.die.net/man/1/install


rjbou:
Thanks!

imbsky:
Anytime!

PR #4028, opened 2019-11-19
Update the cmdliner patch for busybox patch

dra27:
Fixes #4021. I produced the original patch from a Git tree, which means a lot of the context differs from a release tarball - I've updated the patch and tested it with busybox patch (that might be an idea for CI...)

Doesn't affect 2.0 branch as it's still using cmdliner 1.0.2

PR #4034, opened 2019-11-28
src_ext patches mark 2 - with checks and everything

dra27:
Forgot to push this at the weekend - this supersedes #4028, but it also:

- tests the patches using busybox patch on Travis
- provides an automated script for simplifying the patches. This allows you to put Git patches in and get simple `diff -Naur` patches out. This would also deal with previous issues with BSD patch not renaming files (see #3617 and #3620)

Occasionally, this isn't enough: 878a712 attempted to change the EOL-at-EOFness of a file (for no reason) which was upsetting busybox, but without that final commit the CI now catches it.

This PR should at least eliminate surprises from these patches...


rjbou:
Thanks!

PR #4038, opened 2019-12-07
DirTrack: check both size and mtime for cached entries

hannesm:
the earlier code checked size OR mtime for equality, whereas it should be size AND mtime to avoid bad cache lookups

### DIFF
--- a/src/core/opamDirTrack.ml
+++ b/src/core/opamDirTrack.ml
@@ -60,7 +60,7 @@ let cached_digest =
   fun f size mtime ->
     try
       let csize, cmtime, digest = Hashtbl.find item_cache f in
-      if csize = size || mtime = cmtime then Digest.to_hex digest
+      if csize = size && mtime = cmtime then Digest.to_hex digest
       else raise Not_found
     with Not_found ->
       let digest = Digest.file f in


hannesm:
I ended up using an experimental branch of orb (https://github.com/rjbou/orb/pull/1) which:
- enables dirtrack
- creates a switch /tmp/orb-YYY-ZZZ
- installs an opam package
- exports the switch packages
- recods checksums0 from dirtrack for the given opam packages
- removes the switch
- create the same switch /tmp/orb-YYY-ZZZ
- import the above exported packages
- records checksums1 from dirtrack

now, the checksum1 and checksum0 were equal, although manually checking the md5 of the files in the old and new lead to different results (checksum1 contained wrong data). the byte size of the binaries was equal, but not their mtime.

in general, I'm not entirely convinced of this cache (i can generate a equally-sized and equal mtime file at any point in time, whose data is different), does it really lead to better performance in the general setting? in my experiments, I use each digest only once anyways.

rjbou:
I'm wondering if for `orb`, we shouldn't use a digest cache, but compute every time the digest?

rjbou:
Not that `orb` have in its opam file `opam-*` pinned to rbou/orb branch. For those specific ongoing work, maybe better do pull requests on this branch, like that it is quickly merged/avail, and later it will be one PR to merge to finalise first `orb` release.

hannesm:
@rjbou as commented above, if the cache is not useful, I'd be in favour of removing it entirely. If the cache is useful, this is a fix for it.

it is only this PR and #4040 on top of your orb branch (see https://github.com/rjbou/opam/compare/orb...hannesm:orb). imho this PR is worth to review and merge separately from the other(s).

AltGr:
Hm, I don't remember precisely, but indeed this is very suspicious (in particular the `||` that you fixed !)

For the use that opam has of this module, it would repeatedly scan the whole prefix between each package installs, so having such a cache made sense (assuming that no package install modified files while forging their `mtime` to go unnoticed...) ; re-computing all the hashes of all the prefix after each install would just not be practical.

Of course, now opam by default goes the full way of only using the size+mtime (with a correct `&&`, thankfully !), since it was still too slow (the first scan is after all the most costly, in particular on a HDD).

So the question is, does this middle-way "precise tracking" still make sense ? I don't think anyone uses it (@rjbou found out it had been broken for months, if not years, a while ago).
If not, it may be removed (together with the cache). If it still does, we should see if it is usable or not at all without the cache, and accordingly remove just the cache, or keep it as is and add an option to disable the cache for `orb`.

AltGr:
In any case, thanks for the fix :)

hannesm:
thanks. for (my branch of) `orb`, checking mtime and size solves the issue. :)

PR #4040, opened 2019-12-09
Include extra-files in switch export if `full` is provided, on import create an overlay directory with the file contents

hannesm:
Why I'm doing this? I worked on reproducible builds and tooling thereof, where it is very convenient to use opam export and import for rebuilding the same artifact. The current situation that an export contains only a subset of packages (and thus an import may be incomplete or using package metadata from elsewhere) does not fit well into the reproducible builds concept.

~~This PR embeds https://github.com/mirage/ocaml-base64 into opam, I'm happy to remove this and use a dependency for that (but I didn't know how to add depedencies for opam) -- or use a hex encoding if prefered.~~

//cc @rjbou

### DIFF
--- a/src/client/opamAction.ml
+++ b/src/client/opamAction.ml
@@ -368,31 +368,31 @@ let prepare_package_source st nv dir =
     try
       let extra_files_dir =
         OpamPath.Switch.extra_files_dir st.switch_global.root st.switch
-      and repos_roots = OpamRepositoryState.get_root st.switch_repos
-      in
-      List.iter (fun (base, hash) ->
-          let internal_extra_file =
-            let base = OpamFilename.Base.of_string (OpamHash.contents hash) in
-            OpamFilename.create extra_files_dir base
-          in
-          let src =
-            if OpamFilename.exists internal_extra_file then
-              internal_extra_file
-            else match OpamFile.OPAM.get_metadata_dir ~repos_roots opam with
-              | Some mdir ->
-                let files_dir = OpamFilename.Op.(mdir / "files") in
-                OpamFilename.create files_dir base
-              | None ->
-                failwith (Printf.sprintf "couldn't find %s"
-                            (OpamFilename.Base.to_string base))
-          in
+      and metadata_dir = OpamFile.OPAM.metadata_dir opam
+      and extra_files =
+        OpamFile.OPAM.get_extra_files
+          ~repos_roots:(OpamRepositoryState.get_root st.switch_repos)
+          opam
+      and copy_if_hash_matches (src, base, hash) =
           if OpamHash.check_file (OpamFilename.to_string src) hash then
             let dst = OpamFilename.create dir base in
             OpamFilename.copy ~src ~dst
           else
             failwith
-              (Printf.sprintf "Bad hash for %s" (OpamFilename.to_string src)))
-        (match OpamFile.OPAM.extra_files opam with None -> [] | Some xs -> xs);
+              (Printf.sprintf "Bad hash for %s" (OpamFilename.to_string src))
+      in
+      (match metadata_dir with
+       | None ->
+         (* lookup in switch-local hashmap overlay *)
+         List.iter (fun (base, hash) ->
+             let src =
+               let base = OpamFilename.Base.of_string (OpamHash.contents hash) in
+               OpamFilename.create extra_files_dir base
+             in
+             copy_if_hash_matches (src, base, hash))
+           (match OpamFile.OPAM.extra_files opam with None -> [] | Some xs -> xs)
+       | Some _ ->
+         List.iter copy_if_hash_matches extra_files);
       None
     with e -> Some e
   in
--- a/src/client/opamAction.ml
+++ b/src/client/opamAction.ml
@@ -366,29 +366,26 @@ let prepare_package_source st nv dir =
   in
   let check_extra_files =
     try
+      let extra_files_dir =
+        OpamPath.Switch.extra_files_dir st.switch_global.root st.switch
+      in
       List.iter (fun (base, hash) ->
           let dst = OpamFilename.create dir base in
-          let name = "x-extra-file-" ^ OpamHash.contents hash in
-          match OpamStd.String.Map.find_opt name opam.OpamFile.OPAM.extensions with
-          | Some (_, String (_, value)) ->
-            let value' = Base64.decode_string value in
-            let my = OpamHash.compute_from_string ~kind:(OpamHash.kind hash) value' in
-            if OpamHash.contents my = OpamHash.contents hash then
-              OpamFilename.write dst value'
-            else
-              failwith "Bad hash for inline extra-files"
-          | _ ->
-            let repos_roots = OpamRepositoryState.get_root st.switch_repos in
+          let repos_roots = OpamRepositoryState.get_root st.switch_repos in
+          let src =
             match OpamFile.OPAM.get_metadata_dir ~repos_roots opam with
-            | None -> assert false
+            | None ->
+              let base = OpamFilename.Base.of_string (OpamHash.contents hash) in
+              OpamFilename.create extra_files_dir base
             | Some mdir ->
               let files_dir = OpamFilename.Op.(mdir / "files") in
-              let src = OpamFilename.create files_dir base in
-              if OpamHash.check_file (OpamFilename.to_string src) hash then
-                OpamFilename.copy ~src ~dst
-              else
-                failwith
-                  (Printf.sprintf "Bad hash for %s" (OpamFilename.to_string src)))
+              OpamFilename.create files_dir base
+          in
+          if OpamHash.check_file (OpamFilename.to_string src) hash then
+            OpamFilename.copy ~src ~dst
+          else
+            failwith
+              (Printf.sprintf "Bad hash for %s" (OpamFilename.to_string src)))
         (match OpamFile.OPAM.extra_files opam with None -> [] | Some xs -> xs);
       None
     with e -> Some e
--- a/src/client/opamAction.ml
+++ b/src/client/opamAction.ml
@@ -368,20 +368,26 @@ let prepare_package_source st nv dir =
     try
       let extra_files_dir =
         OpamPath.Switch.extra_files_dir st.switch_global.root st.switch
+      and repos_roots = OpamRepositoryState.get_root st.switch_repos
       in
       List.iter (fun (base, hash) ->
-          let dst = OpamFilename.create dir base in
-          let repos_roots = OpamRepositoryState.get_root st.switch_repos in
+          let internal_extra_file =
+            let base = OpamFilename.Base.of_string (OpamHash.contents hash) in
+            OpamFilename.create extra_files_dir base
+          in
           let src =
-            match OpamFile.OPAM.get_metadata_dir ~repos_roots opam with
-            | None ->
-              let base = OpamFilename.Base.of_string (OpamHash.contents hash) in
-              OpamFilename.create extra_files_dir base
-            | Some mdir ->
-              let files_dir = OpamFilename.Op.(mdir / "files") in
-              OpamFilename.create files_dir base
+            if OpamFilename.exists internal_extra_file then
+              internal_extra_file
+            else match OpamFile.OPAM.get_metadata_dir ~repos_roots opam with
+              | Some mdir ->
+                let files_dir = OpamFilename.Op.(mdir / "files") in
+                OpamFilename.create files_dir base
+              | None ->
+                failwith (Printf.sprintf "couldn't find %s"
+                            (OpamFilename.Base.to_string base))
           in
           if OpamHash.check_file (OpamFilename.to_string src) hash then
+            let dst = OpamFilename.create dir base in
             OpamFilename.copy ~src ~dst
           else
             failwith
--- a/src/client/opamAction.ml
+++ b/src/client/opamAction.ml
@@ -365,36 +365,41 @@ let prepare_package_source st nv dir =
       (Done None) (OpamFile.OPAM.extra_sources opam)
   in
   let check_extra_files =
-    try
-      let extra_files_dir =
-        OpamPath.Switch.extra_files_dir st.switch_global.root st.switch
-      and metadata_dir = OpamFile.OPAM.metadata_dir opam
-      and extra_files =
+    let extra_files =
+      let extra_files =
         OpamFile.OPAM.get_extra_files
           ~repos_roots:(OpamRepositoryState.get_root st.switch_repos)
           opam
-      and copy_if_hash_matches (src, base, hash) =
-          if OpamHash.check_file (OpamFilename.to_string src) hash then
-            let dst = OpamFilename.create dir base in
-            OpamFilename.copy ~src ~dst
-          else
-            failwith
-              (Printf.sprintf "Bad hash for %s" (OpamFilename.to_string src))
       in
-      (match metadata_dir with
-       | None ->
-         (* lookup in switch-local hashmap overlay *)
-         List.iter (fun (base, hash) ->
-             let src =
-               let base = OpamFilename.Base.of_string (OpamHash.contents hash) in
-               OpamFilename.create extra_files_dir base
-             in
-             copy_if_hash_matches (src, base, hash))
-           (match OpamFile.OPAM.extra_files opam with None -> [] | Some xs -> xs)
-       | Some _ ->
-         List.iter copy_if_hash_matches extra_files);
-      None
-    with e -> Some e
+      if extra_files <> [] then extra_files else
+      match OpamFile.OPAM.extra_files opam with
+      | None -> []
+      | Some xs ->
+        (* lookup in switch-local hashmap overlay *)
+        let extra_files_dir =
+          OpamPath.Switch.extra_files_dir st.switch_global.root st.switch
+        in
+        OpamStd.List.filter_map (fun (base, hash) ->
+            let src =
+              OpamFilename.create extra_files_dir
+                (OpamFilename.Base.of_string (OpamHash.contents hash))
+            in
+            if OpamFilename.exists src then
+              Some (src, base, hash)
+            else None)
+          xs
+    in
+    let bad_hash =
+      OpamStd.List.filter_map (fun (src, base, hash) ->
+          if OpamHash.check_file (OpamFilename.to_string src) hash then
+            (OpamFilename.copy ~src ~dst:(OpamFilename.create dir base); None)
+          else
+            Some src) extra_files
+    in
+    if bad_hash = [] then None else
+      Some (Failure
+              (Printf.sprintf "Bad hash for %s"
+                 (OpamStd.Format.itemize OpamFilename.to_string bad_hash)));
   in
   OpamFilename.mkdir dir;
   get_extra_sources_job @@+ function Some _ as err -> Done err | None ->
--- a/src/client/opamAction.ml
+++ b/src/client/opamAction.ml
@@ -366,15 +366,30 @@ let prepare_package_source st nv dir =
   in
   let check_extra_files =
     try
-      List.iter (fun (src,base,hash) ->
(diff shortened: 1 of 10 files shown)


rjbou:
Also, I would change the title to "Switch export and import: include extra_files in embedded opams" or something like that. To include all installed opam file, there is the `--full` option, and in the PR, only embedded opam file have their extra file included.

src/client/opamSwitchCommand.ml
```diff
@@ -477,38 +477,51 @@ let import_t ?ask importfile t =
...
+              opam files
+        in
+        OpamPackage.Name.Map.add nv.name opam' acc
       | None -> acc)
     packages
     OpamPackage.Name.Map.empty
 
-let export ?(full=false) filename =
-  let switch = OpamStateConfig.get_switch () in
+let export rt ?(full=false) ?(switch = OpamStateConfig.get_switch ()) filename =
```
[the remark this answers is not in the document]
hannesm: that was leftover debugging/attempts to get it right i guess

src/client/opamSwitchCommand.ml
```diff
@@ -477,38 +477,51 @@ let import_t ?ask importfile t =
...
+        let opam' = match OpamFile.OPAM.get_extra_files ~repos_roots opam with
+          | [] -> opam
+          | files ->
+            (* read file and embed them into opam *)
+            List.fold_left (fun opam (file, _rel_file, hash) ->
+                if OpamFilename.exists file &&
+                   OpamHash.check_file (OpamFilename.to_string file) hash then
+                  let value = OpamFilename.read file in
+                  let value' = B64.encode_string value in
+                  let name = "x-" ^ OpamHash.contents hash in
```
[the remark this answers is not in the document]
hannesm: `x-extra-file-HASH` is it now, I'd prefer not to use a single field containing pairs, since that means I'd need to write more code. I can't use `x-filename` since filename is not necessarily an identifier.

hannesm:
thanks for the review, and the suggestion for the title. I managed (let's see what CI thinks) to use the extlib base64, and renamed the fields.

AltGr:
Thanks... indeed that is something that was missing, but I couldn't find a satisfying way to do it.

I must say that I am not really convinced by the choice of `x-*` fields with varying names, which goes somewhat against the existing conventions for opam files. Ideally, the feature could also be used to embed extra files in opam files even for normal uses... which it seems this PR would allow, but in an undocumented and inconvenient way.

Here are a few options I could think of (this is to open discussions, I am not sure about the best design for this...):

- limit the extra-files embedding to export/import:
  * in this case, I think the extra file content may be better outside of the `package "foo" {}` section, to preserve the format ; but then it seems tedious to recover the file at the right moment for non-pinned packages
  * an idea could be to just have a field with a `hash → contents` mapping, that gets unpacked to a temporary cache when the import is done. Then, similar to what is done in the proposition (and for `extra_sources:`), check there to find the file in `prepare_package_source`

- extend the opam file format:
  * the files could be in an extra section or field, to  not disrupt file formatting too much
  * to be more usable, it would be best if embedded text files could be put as raw text (we do already have the functions for safe escaping with `"""`)
  * an idea closer to what we have would be to simply overload the `checksum` field: besides `"md5=..."`, `"sha256=..."` etc., we could just add `"raw=..."` (or more reasonably `"""raw=..."""`) with the raw (properly escaped) file contents, and `"base64=..."`. This seems the least disruptive, but I am not sure about the readability of the resulting files.

What do you think ?

hannesm:
I'd first strive for a solution to limit the extra-files embedding for import and export. I'm keen to get to a state where an `opam export` produces a file that can `opam import` without any other sources of information (fine with me to download hashed tarballs or git commit ids). It is the minimal invasive solution I could come up with, that solves my issue.

I initially went for a `string` (no base64) encoding, but that didn't work (IIRC `\"` was mistranslated, the outputter already does some escaping), and thus decided to use base64 encoded strings instead.

>   limit the extra-files embedding to export/import:
>        in this case, I think the extra file content may be better outside of the package "foo" {} section, to preserve the format ; but then it seems tedious to recover the file at the right moment for non-pinned packages
>        an idea could be to just have a field with a hash → contents mapping, that gets unpacked to a temporary cache when the import is done. Then, similar to what is done in the proposition (and for extra_sources:), check there to find the file in prepare_package_source

that sounds good to me. Unfortunately I don't know the opam codebase well enough to implement this. would you be up for helping out? :)

hannesm:
To further add to the discussion, I [added preliminary functionality](https://github.com/hannesm/opam/commit/160f8666229eb25871893a74003928a0f872c6d0) which rewrites git urls by specific commits. This is crucial for reproducibility.

It makes me wonder where/how this "import/export" is used at the moment, or rather whether the changes I want for reproducibility should be hidden behind optional arguments or maybe even a different code path.

[Another change](https://github.com/hannesm/orb/commit/0b580034c6ef52b0b60c0934f3adac0cc79b4582#diff-c962676bdd25309306bc999585df2e63) (this time in orb) cleans up the environment before proceeding further. Opam already at initialisation time captures the environment, that's why I needed to do it before the opam library is loaded somehow. It feels a bit clunky. I'm interested what you think about making the minimizing the environment for opam a default? I.e. is there a reason to not do that? only having `HOME` and `PATH` may be a bit too radical.

rjbou:
For the last solution, there are several steps:

- [ ] add an `extra-files` field to switch export file
- [ ] add the directory where they will be stored in path
- [ ] in `OpamSwitchCommand.export`, fill the new fields with a scanning of the opam files to save
- [ ] in `OpamSwitchCommand.import_t`, write them (filename = hash ; content = content of the file) in the given directory
- [ ] in `OpamAction.prepare_source`, check & take if the file exists in the extra-files directory
- [ ] on `opam clean` empty extra-files directory

You can pick the two first steps in my repo, branch `exportXfiles`.

hannesm:
I did the steps asked for, apart from " on `opam clean` empty extra-files directory" since I don't know where I should cleanup things. My impression is that `OpamSwitchCommand.clear_switch` already cleans the `extra_files` directory in `OpamFilename.rmdir comp_dir`.

src/client/opamSwitchCommand.ml
```diff
@@ -372,6 +372,19 @@ let switch lock gt switch =
...
 
+  let extra_files = importfile.OpamFile.SwitchExport.extra_files in
+  let dir = OpamPath.Switch.extra_files_dir t.switch_global.root t.switch in
+  OpamHash.Map.iter (fun hash content ->
+      let value = Base64.decode_string content in
+      let my = OpamHash.compute_from_string ~kind:(OpamHash.kind hash) value in
+      if OpamHash.contents my = OpamHash.contents hash then
+        let dst =
+          let base = OpamFilename.Base.of_string (OpamHash.contents hash) in
+          OpamFilename.create dir base in
```
```suggestion
          OpamFilename.create dir base
        in
```
Just tiny style thing :)

src/client/opamSwitchCommand.ml
```diff
@@ -492,29 +505,69 @@ let read_overlays (read: package -> OpamFile.OPAM.t option) packages =
...
+                OpamHash.Map.add hash value' acc
+              else begin
+                OpamConsole.warning "Ignoring file %s with invalid hash"
+                  (OpamFilename.to_string file);
+                acc
+              end)
+            acc files)
+    packages data
+
+let export rt ?(full=false) ?(freeze = false) filename =
```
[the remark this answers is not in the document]
rjbou: Wondering, shouldn't freeze be introduced in the git commit / export pr ?

src/client/opamSwitchCommand.ml
```diff
@@ -492,29 +505,69 @@ let read_overlays (read: package -> OpamFile.OPAM.t option) packages =
...
   let export =
     OpamFilename.with_flock `Lock_none (OpamPath.Switch.lock root switch)
     @@ fun _ ->
     let selections = S.safe_read (OpamPath.Switch.selections root switch) in
     let overlays =
-      read_overlays (fun nv ->
-          OpamFileTools.read_opam
-            (OpamPath.Switch.Overlay.package root switch nv.name))
-        selections.sel_pinned
+        read_overlays (not freeze) (fun nv ->
```
rjbou: Why not use `full` ?

src/client/opamAction.ml
```diff
@@ -366,15 +366,27 @@ let prepare_package_source st nv dir =
...
+            match OpamFile.OPAM.get_metadata_dir ~repos_roots opam with
+            | None ->
+              let base = OpamFilename.Base.of_string (OpamHash.contents hash) in
+              OpamFilename.create extra_files_dir base
+            | Some mdir ->
+              let files_dir = OpamFilename.Op.(mdir / "files") in
+              OpamFilename.create files_dir base
+          in
+          if OpamHash.check_file (OpamFilename.to_string src) hash then
+            OpamFilename.copy ~src ~dst
```
rjbou: I don't really understand what you try to do here. From what I see, you create the hash-named extra-file you want to copy (in the internal `extra-files` directory or directly in the source or internal repo), then copy it.
In this source preparation function, shouldn't be: 
* if the extra-file exists in the `extra-file` internal, take it from there
* otherwise take it from the `metadata_dir` (you can use `OpamFile.OPAM.get_extra_files` to not duplicate code.

src/client/opamAction.ml
```diff
@@ -366,15 +366,27 @@ let prepare_package_source st nv dir =
...
     try
-      List.iter (fun (src,base,hash) ->
-          if not (OpamHash.check_file (OpamFilename.to_string src) hash) then
-            failwith
-              (Printf.sprintf "Bad hash for %s" (OpamFilename.to_string src))
+      let extra_files_dir =
+        OpamPath.Switch.extra_files_dir st.switch_global.root st.switch
+      in
+      List.iter (fun (base, hash) ->
+          let dst = OpamFilename.create dir base in
```
rjbou: If the check fails, `dir/base` is created and not deleted. Better create it at last moment.

src/client/opamAction.ml
```diff
@@ -366,15 +366,27 @@ let prepare_package_source st nv dir =
...
+          let repos_roots = OpamRepositoryState.get_root st.switch_repos in
```
rjbou: `repos_roots` doesn't change, it doesn't need to be inside the `iter`

src/client/opamSwitchCommand.ml
```diff
@@ -492,29 +505,69 @@ let read_overlays (read: package -> OpamFile.OPAM.t option) packages =
...
-      else overlays
+      else
+        overlays
+    in
+    let extra_files =
+      if freeze then
+        let repos_roots = OpamRepositoryState.get_root rt in
+        let data =
+          read_extra_files ~repos_roots (fun nv ->
+              OpamFileTools.read_opam
```
rjbou: It is possible to factorise read overlay opam function and read installed opam one. each one is used two times.

rjbou: It can even be the opposite: have two function `on_overlay_pinned_opam` and `on_installed_opam` that takes the function to call, e.g? `on_installed_opam @@ read_extra_files ~repos_roots`.

rjbou: _just a suggestion/thinking about it, i'd just be happy with the factorisation :)_

rjbou:
For the `opam clean` part, it is in the clean, by adding the remove of the `OpamPath.Switch.extra_files` directory in the switch cleaning. See [there](https://github.com/ocaml/opam/blob/master/src/client/opamCommands.ml#L3079)

src/client/opamAction.ml
hannesm: in 9d5d9d34089439e066abbd97618b9771f7a69c69 I revise the implementation. previously, opam used OpamFile.OPAM.get_extra_files, but this iterates only over the existing files in the `files` subdirectory. The new implementation iterates over the `extra_files` field in the opam file (and not over the existing files), and checks whether a file with this hash is in `extra_files` subdirectory, or if `files/<filename>` is present.

Does the revised implementation make more sense to you?

src/client/opamSwitchCommand.ml
hannesm: done in 4633e78e

src/client/opamSwitchCommand.ml
```diff
@@ -477,44 +491,79 @@ let import_t ?ask importfile t =
...
   let switch = OpamStateConfig.get_switch () in
   let root = OpamStateConfig.(!r.root_dir) in
   let export =
     OpamFilename.with_flock `Lock_none (OpamPath.Switch.lock root switch)
     @@ fun _ ->
     let selections = S.safe_read (OpamPath.Switch.selections root switch) in
-    let overlays =
-      read_overlays (fun nv ->
-          OpamFileTools.read_opam
+    let repos_roots = OpamRepositoryState.get_root rt in
```
rjbou: upcoming refactor proposal

rjbou: PR updated with the refactor, as discussed.

hannesm:
hi, thanks to @rjbou for refactoring, this works fine for me -- best be squashed and read as a single commit. It implements the top approach suggested by @AltGr in https://github.com/ocaml/opam/pull/4040#issuecomment-564620513

>  - limit the extra-files embedding to export/import:
>      - in this case, I think the extra file content may be better outside of the package "foo" {} section, to preserve the format ; but then it seems tedious to recover the file at the right moment for non-pinned packages
>      - an idea could be to just have a field with a hash → contents mapping, that gets unpacked to a temporary cache when the import is done. Then, similar to what is done in the proposition (and for extra_sources:), check there to find the file in prepare_package_source

About using `OpamFile.OPAM.get_extra_files`: if it is used directly in `prepare_package_sources`, the import won't work since `get_extra_files` returns the empty list if the `files/` subdirectory is not present. Another approach may be to modify `get_extra_sources` to use the switch-local hashmap -- but `get_extra_sources` has more call sites than `OpamAction.prepare_package_source`, and would need switch arguments.

Previously (before my last commit), each `extra_file` was first looked up in the switch-local hashmap, and uses the `files/yyy` as fallback. With 4329ecb7a6cc34d79e29e9683eb6ce8e0021b420, the behaviour changes slightly: either use all `extra_files` from `files/`, or if `metadata_dir` is absent, use the switch-local hashmap, over the behaviour of this PR.

Please let me know what you think, and how to progress. I'd be happy to get this merged before 2.1 (/cc @AltGr).

My testing of this PR was done as follows:
- pin opam-* to https://github.com/hannesm/opam.git#orb-partial
- pin orb to https://github.com/hannesm/orb.git#active
- opam repo add repro-test https://github.com/roburio/reproducible-testing-repo
- run `orb build --twice --repo=repro-test extra-file` (an opam package with an extra file)

rjbou:
When I was talking about using the `get_extra_files` function it was not to _only_ uses it, but more to avoid code duplication between `get_extra_files` code and the one in your first commits. The first check in the cache hashmap is needed to be kept here, and as you said, having each `files/*` files as a fallback is the way to go.

src/client/opamAction.ml
```diff
@@ -366,15 +366,33 @@ let prepare_package_source st nv dir =
...
+             let src =
+               let base = OpamFilename.Base.of_string (OpamHash.contents hash) in
+               OpamFilename.create extra_files_dir base
+             in
+             copy_if_hash_matches (src, base, hash))
+           (match OpamFile.OPAM.extra_files opam with None -> [] | Some xs -> xs)
+       | Some _ ->
+         List.iter copy_if_hash_matches extra_files);
       None
     with e -> Some e
```
rjbou: I was thinking about something like that : first get the cached ones then the already present in files (downloaded) ones, and merge the two lists, giving priority to cached ones
```suggestion
  let check_extra_files =
    try
      let cached_files =
        match OpamFile.OPAM.extra_files opam with
        | None -> []
        | Some xs ->
          (* lookup in switch-local hashmap overlay *)
          let extra_files_dir =
            OpamPath.Switch.extra_files_dir st.switch_global.root st.switch
          in
          OpamStd.List.filter_map (fun (base, hash) ->
              let src =
                let base = OpamFilename.Base.of_string (OpamHash.contents hash) in
                OpamFilename.create extra_files_dir base
              in
              if OpamFilename.exists src then
                Some (src, base, hash)
              else None)
            xs
      in
      let files =
        match
          OpamFile.OPAM.get_extra_files
            ~repos_roots:(OpamRepositoryState.get_root st.switch_repos)
            opam with
        | [] -> cached_files
        | fs ->
          OpamStd.List.filter_map (fun (_, base, hash) ->
              List.find_opt (fun (_, base', hash') ->
                  base <> base' && hash <> hash')
                cached_files) fs
      in
      let bad_hash =
        OpamStd.List.filter_map (fun (src, base, hash) ->
            if OpamHash.check_file (OpamFilename.to_string src) hash then
              (OpamFilename.copy ~src ~dst:(OpamFilename.create dir base); None)
            else
              Some src) files
      in
      if bad_hash <> [] then
        failwith
          (Printf.sprintf "Bad hash for %s"
             (OpamStd.Format.pretty_list (List.map OpamFilename.to_string bad_hash)));
      None
    with e -> Some e
```

hannesm: hmm, as I tried to elaborate in my message, I don't like the semantics "take either from cache or from `files/`", but really prefer to take them _all_ from `files/` -- or if files is not present, take them all from cache.

hannesm: but if you think it is better (or makes it more likely to get merged) to take files from either here or there, feel free to push your changes directly on the branch.

rjbou: Oh ok. That behavior is ok for me too. cf. last commit

hannesm: thanks :)

src/format/opamPath.ml
```diff
@@ -117,6 +117,10 @@ module Switch = struct
 
   let sources_dir t a = meta t a / "sources"
 
+  let extra_files_dir t a = sources_dir t a / "extra-files"
```
AltGr: We should choose another directory: this might conflict with a pinned package called "extra-files".

src/format/opamPath.ml
```diff
@@ -117,6 +117,10 @@ module Switch = struct
...
+
+  let extra_file t a h = extra_files_dir t a / OpamHash.contents h
```
AltGr:
```suggestion
  let extra_file t a h = extra_files_dir t a // OpamHash.contents h
```

src/format/opamPath.ml
AltGr:
```suggestion
  let extra_files_dir t a = meta t a / "extra-files-cache"
```

src/format/opamPath.mli
```diff
@@ -153,6 +153,10 @@ module Switch: sig
       $meta/sources/$name.$version/} *)
   val sources: t -> switch -> package -> dirname
 
+  val extra_files_dir: t -> switch -> dirname
+
+  val extra_file: t -> switch -> OpamHash.t -> dirname
+
```
AltGr:
```suggestion
  (** Temporary switch-local directory where a by-hash map of extra files may be stored.
      This is used for switch-imports: {i $meta/extra-files-cache} *)
  val extra_files_dir: t -> switch -> dirname

  (** Extra file with the given hash from the temporary switch-import cache:
      {i $meta/extra-files-cache/HASH} *)
  val extra_file: t -> switch -> OpamHash.t -> filename
```

src/client/opamSwitchCommand.ml
```diff
@@ -372,6 +372,20 @@ let switch lock gt switch =
...
+        let dst =
+          let base = OpamFilename.Base.of_string (OpamHash.contents hash) in
+          OpamFilename.create dir base
+        in
+          OpamFilename.write dst value
+      else
+        failwith "Bad hash for inline extra-files") extra_files;
+
   let import_sel = importfile.OpamFile.SwitchExport.selections in
   let import_opams = importfile.OpamFile.SwitchExport.overlays in
```
AltGr: The extra-files directory should be cleaned then after `resolve_and_apply` is done with it.

rjbou:
updated & rebased

PR #4049, opened 2019-12-20
opamFile extensions (x- fields) enhancements

hannesm:
this PR consists of two commits:
- add `remove_extension : t -> string -> t`
- change behaviour of `to_string_with_preserved_format` to not treat `x-` specially. I'm not sure why the code was around, is there a test / example for the previous behaviour @AltGr?

### DIFF
--- a/src/format/opamFile.ml
+++ b/src/format/opamFile.ml
@@ -2252,6 +2252,9 @@ module OPAMSyntax = struct
     if not (is_ext_field fld) then invalid_arg "OpamFile.OPAM.add_extension";
     {t with
      extensions = OpamStd.String.Map.add fld (pos_null,syn) t.extensions }
+  let remove_extension t fld =
+    if not (is_ext_field fld) then invalid_arg "OpamFile.OPAM.remove_extension";
+    {t with extensions = OpamStd.String.Map.remove fld t.extensions }
 
   let with_url url t =
     let format_errors =
--- a/src/format/opamFile.ml
+++ b/src/format/opamFile.ml
@@ -824,7 +824,9 @@ module Syntax = struct
                    OpamPrinter.items [f] :: strs
                  with Not_found -> strs
              with Not_found | OpamPp.Bad_format _ ->
-               if OpamStd.String.starts_with ~prefix:"x-" name then
+               if OpamStd.String.starts_with ~prefix:"x-" name &&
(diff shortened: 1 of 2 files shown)


src/format/opamFile.ml
```diff
@@ -823,10 +823,7 @@ module Syntax = struct
                    in
                    OpamPrinter.items [f] :: strs
                  with Not_found -> strs
-             with Not_found | OpamPp.Bad_format _ ->
-               if OpamStd.String.starts_with ~prefix:"x-" name then
-                 field_str (`Var name) :: strs
-               else strs)
+             with Not_found | OpamPp.Bad_format _ -> strs
```
rjbou:
```suggestion
             with Not_found | OpamPp.Bad_format _ -> strs)
```

rjbou:
It's to keep `x-` fields at their place, instead of moving them if they are in `rem`. There is no test for that function(ality), only manual ones ;).
Why do you want to remove the `x-` fields special case ?

hannesm:
@rjbou because otherwise even if I remove `x-` fields from the OPAM.t, they are retained in the output file when using `to_string_with_preserved_format`

src/format/opamFile.ml
```diff
@@ -823,10 +823,7 @@ module Syntax = struct
                    in
                    OpamPrinter.items [f] :: strs
                  with Not_found -> strs
-             with Not_found | OpamPp.Bad_format _ ->
-               if OpamStd.String.starts_with ~prefix:"x-" name then
-                 field_str (`Var name) :: strs
-               else strs)
+             with Not_found | OpamPp.Bad_format _ -> strs)
```
rjbou:
```suggestion
             with Not_found | OpamPp.Bad_format _ ->
               if OpamStd.String.starts_with ~prefix:"x-" name &&
                  List.find_opt (fun i -> it_ident i = `Var name)
                    syn_t.file_contents <> None then
                 field_str (`Var name) :: strs
               else strs)
```

rjbou:
In that case, better check is `x-` field is present in the given opamfile.

hannesm:
thanks @rjbou -- that is fine with me as well (see above, I just don't had any idea why the code was around, but thanks for explaining it), and I committed your patch :D

rjbou:
Cleaning the branch (squash of the 3 last commits) and it's a good to merge for me.

hannesm:
thanks, I squashed and fixed the commit message

PR #4051, opened 2020-01-06
Use pkg_add to install opam on OpenBSD

daimrod:
Built packages are preferred over building your own packages on OpenBSD.

See https://www.openbsd.org/faq/faq15.html

### DIFF
--- a/doc/pages/Install.md
+++ b/doc/pages/Install.md
@@ -124,12 +124,10 @@ urpmi opam
 
 #### OpenBSD
 
-Opam builds via sources fine on OpenBSD 5.6 or earlier, and is available in the
-ports and packages tree on OpenBSD 5.7 or higher.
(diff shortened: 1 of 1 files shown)


daimrod:
If so, the README should be updated, it says the tutorials and guides are generated from `doc/pages/`.

Thanks for the review.

rjbou:
Thanks!
It's the good place to update it!
Just a question / curiosity: why not keep the `Opam builds via sources fine on OpenBSD 5.6 or earlier` ?

daimrod:
OpenBSD 5.6 is from 2014 and the current OpenBSD version is 6.6. I
thought there was no point in keeping this.

rjbou:
fair enough :)

avsm:
Agreed with @daimrod -- and binary packages are the recommended installation mechanism on openbsd at any rate, since it works with system upgrades and such

PR #4052, opened 2020-01-06
Include the version number in "compilation failed" messages

Armael:
This is a minor improvement of the console logging when building a series of packages. If installation succeeds, the message is `* installed pkg.ver`, but if it fails, it is `the compilation of pkg failed at ...`.

This commits adds the version number to the failure message, for symmetry and to be more explicit.
(this way, one can start investigating the issue before waiting for the final message which does include the version)


rjbou:
Thanks!

PR #4064, opened 2020-01-14
Copy files using OCaml code instead of calling to cp or install

Armael:
This PR proposes to use pure ocaml code instead of calling to the `cp` or `install` programs in order to copy files around.
This is here restricted to invocations of `cp` that copy a single file, and invocations of `install`. Calls to `cp -r` are not modified.

The copy code itself has been taken as-is from dune (see `src/stdune/io.ml`).

I do realize that such a PR should be carefully reviewed, as it might result in subtle bugs if the new code differs from coreutils' `cp` on some corner-cases that occur in practice. 
Taking that into consideration, I still think it is a risk worth taking, considering the following:

## Performance

I benchmarked an invocation of `opam install` on the following [317 packages](https://gist.github.com/Armael/2f75cd0a3114e4c94d88c5d6bee460ff).
Both invocations run from a fresh ocaml 4.09.0 switch, with all the source archives already in the download cache. They run on a beefy server: 80 cores, `OPAMJOBS=20`, and the opam root is stored in RAM (in a tmpfs).

- before the PR: `7min 22s`
- after the PR: `2min 46`

Hence a speedup of more than 2x (!). 
Note that having a beefy machine is somewhat of an extreme situation (and is favorable for my benchmark); on a more realistic machine, the speedup will be lower as the CPU/disk will be more of a bottleneck. 

Still, I believe this benchmark illustrates that significant time is spent by opam on launching processes.

### DIFF
--- a/src/core/opamSystem.ml
+++ b/src/core/opamSystem.ml
@@ -175,6 +175,45 @@ let write file contents =
   output_string oc contents;
   close_out oc
 
+let setup_copy ?(chmod = fun x -> x) ~src ~dst () =
+  let ic = open_in_bin src in
+  let oc =
+    try
+      let perm =
+        (Unix.fstat (Unix.descr_of_in_channel ic)).st_perm |> chmod
+      in
+      open_out_gen
+        [ Open_wronly; Open_creat; Open_trunc; Open_binary ]
+        perm dst
+    with exn ->
+      OpamStd.Exn.finalise exn (fun () -> close_in ic)
+  in
+  (ic, oc)
+
+let copy_channels =
+  let buf_len = 4096 in
+  let buf = Bytes.create buf_len in
+  let rec loop ic oc =
+    match input ic buf 0 buf_len with
+    | 0 -> ()
+    | n ->
+      output oc buf 0 n;
+      loop ic oc
+  in
+  loop
+
+let copy_file_aux ?chmod ~src ~dst () =
+  let close_channels ic oc =
+    OpamStd.Exn.finally (fun () -> close_in ic) (fun () -> close_out oc) in
+  try
+    let ic, oc = setup_copy ?chmod ~src ~dst () in
+    OpamStd.Exn.finally (fun () -> close_channels ic oc)
+      (fun () -> copy_channels ic oc);
+  with Unix.Unix_error _ as e ->
+    (* Remove the partial destination file, if any. *)
+    (try Unix.unlink dst with Unix.Unix_error _ -> ());
+    internal_error "Cannot copy %s to %s (%s)." src dst (Printexc.to_string e)
+
 let chdir dir =
   try Unix.chdir dir
   with Unix.Unix_error _ -> raise (File_not_found dir)
@@ -532,7 +571,7 @@ let cygify f =
   else
     fun x -> x
 
-let copy_file_aux f src dst =
+let copy_file src dst =
   if (try Sys.is_directory src
       with Sys_error _ -> raise (File_not_found src))
   then internal_error "Cannot copy %s: it is a directory." src;
@@ -541,9 +580,8 @@ let copy_file_aux f src dst =
   if file_or_symlink_exists dst
   then remove_file dst;
   mkdir (Filename.dirname dst);
-  command ~verbose:(verbose_for_base_commands ()) ("cp"::(cygify f [src; dst]))
-
-let copy_file = copy_file_aux (get_cygpath_function ~command:"cp")
+  log "copy %s -> %s" src dst;
+  copy_file_aux ~src ~dst ()
 
 let copy_dir src dst =
   if Sys.file_exists dst then
@@ -687,7 +725,7 @@ let install ?(warning=default_install_warning) ?exec src dst =
               warning dst `Install_unknown;
               (dst, false)
         in
-        copy_file src dst;
+        copy_file_aux ~src ~dst ();
         if cygcheck then
           match OpamStd.Sys.is_cygwin_variant dst with
             `Native ->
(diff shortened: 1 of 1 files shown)


Armael:
(there may be other optimizations that would be possible along this line, e.g. with `cp -r`, that I might investigate, but I think this one is the most significant one)

Armael:
An alternative to copy-pasting code from dune would be to depend on [fileutils](https://github.com/gildor478/ocaml-fileutils) (which would also give us ocaml code for other of the coreutils utilities).

avsm:
Good idea, and benchmarks! Just a quick note that the dune license needs to be followed since the code came from there; just add https://github.com/ocaml/dune/blob/master/LICENSE.md to this file.

Armael:
Sorry, which file are you referring to?

rjbou:
Thanks for the proposal @Armael. Those benchs seems promising!

For the license file, we can see/fix that once the PR is almost ready for merge. I'm not in favor of including directly inside a existing file code from other project. Maybe better have them in another file, like that the header specifies exactly from where they originate, have right license, etc. One other thing is as this code will evolve in dune, opam won't necessarily have those update as the code is copied as is. 

On the code itself, we also need to check that it is cygwin/windows compliant (cc @dra27).

`fileutils` don't have a lot of dependencies, it is conceivable to add it as an opam dependencies. The same checks need to be done (portability, reliability).

mefyl:
The impact of this PR is greatly amplified by the binary cache, as the compilation time will be massively reduced :+1:

Armael:
@mefyl out of curiosity, what kind of binary cache are you referring to exactly? (but indeed, I agree with you, I'm just not aware of people actually using an opam binary cache)

kit-ty-kate:
I did a battery of tests and I can confirm that this PR increases performances dramatically. If it's of any interest for the opam devs here is this PR cherry-picked on top of opam 2.0.5 in case this PR is going to be backported to the 2.0 branch: https://github.com/kit-ty-kate/opam/commit/e77305d317d7fe1f417bf785ec42420758ac4cf1

This PR improves [opam-health-check](http://github.com/kit-ty-kate/opam-health-check)'s overall performances by 20% (from instance the setup I'm using to test OCaml 4.10 & OCaml 4.09 went from 15 hours to 12 hours with just this patch alone).

With @mefyl we came across this issue while trying to figure out why adding the dune binary cache to opam-health-check was not improving performances as much as expected, it turns out that for instance, with a full cache, a typical compilation of `core_kernel` takes < 2s while the installation phase done by opam takes more than a minute due to the large number of calls to `install`. With this PR it now takes just a couple of seconds.

dra27:
In essence, I'm of course of completely for this (especially as it has to be done that way on Windows anyway), but slightly playing devil's advocate for using the **system**'s command for installing files as being safer - can't the same performance improvement by made by grouping the files to be installed by directory and executableness? i.e. `install -m 755 file1 file2 file3 file4 file5 /home/user/.opam/default/bin/`?

mefyl:
> @mefyl out of curiosity, what kind of binary cache are you referring to exactly? (but indeed, I agree with you, I'm just not aware of people actually using an opam binary cache)

The one that is available since 2.0 but not yet advertised - see https://github.com/ocaml/dune/blob/master/doc/caching.rst

Armael:
I've refactored the code a bit to use facilities from `OpamStd.Exn` and make it a bit less copy-paste-y.

>slightly playing devil's advocate for using the system's command for installing files as being safer 

safer in what respect?

>can't the same performance improvement be made by grouping the files to be installed by directory and executableness?

That's not so easy, because in the general case, a `.install` file defines a mapping `from -> to` which is not 1-to-1 (one `from` file can be installed several times as different files, for instance), and does not necessarily follow the same directory structure in the source and in the target.
So we would have to cluster files, to regroup files with the same source and target directory (and mode...) to be copied with the same `install` invocation.
That seems a bit too complicated for its own good (and I don't know what the gains would be, since it would depend on how good the clustering can be in practice)...

src/core/opamSystem.ml
```diff
@@ -175,6 +175,44 @@ let write file contents =
...
+    try
+      let perm =
+        (Unix.fstat (Unix.descr_of_in_channel ic)).st_perm |> chmod
+      in
+      open_out_gen
+        [ Open_wronly; Open_creat; Open_trunc; Open_binary ]
+        perm dst
+    with exn ->
+      close_in ic;
+      raise exn
```
AltGr:
```suggestion
      OpamStd.Exn.finalise exn (fun () -> close_in ic)
```
This helps preserve stack traces

src/core/opamSystem.ml
```diff
@@ -175,6 +175,44 @@ let write file contents =
...
+  in
+  (ic, oc)
+
+let copy_channels =
+  let buf_len = 65536 in
```
AltGr: Isn't this a bit large ? Does it get much better perf than, say, a block size of 4096 ?

src/core/opamSystem.ml
```diff
@@ -175,6 +175,44 @@ let write file contents =
...
+      loop ic oc
+  in
+  loop
+
+let copy_file_aux ?chmod ~src ~dst () =
+  let close_channels ic oc =
+    OpamStd.Exn.finally (fun () -> close_in ic) (fun () -> close_out oc) in
+  try
+    let ic, oc = setup_copy ?chmod ~src ~dst () in
+    OpamStd.Exn.finally (fun () -> close_channels ic oc)
```
AltGr: It could be nice to delete the partial destination file in case of error, but not sure.

src/core/opamSystem.ml
Armael: No idea. I took the value from the dune implementation. I don't care enough to benchmark it, so I changed it to the more reasonable-looking 4096.

src/core/opamSystem.ml
Armael: fixed, thanks.

src/core/opamSystem.ml
Armael: sounds reasonable.

Armael:
I just pushed an updated version that hopefully accounts for the few comments.

Issue #4036, opened 2019-12-04
Feature request: Add the --download-only option to opam install


kit-ty-kate:
Sometimes when testing stuff or for other reasons, we might want to download just the archives we need before installing them.

Would it be possible to get this option added to the `opam install` command?

rjbou:
related to #2895

talex5:
This would also be really useful for ocaml-ci. We want to do the download step with opam having write access to a shared download cache, but perform the builds themselves (which contain untrusted code) with read-only access.

Issue #2895, opened 2017-03-28
missing functionality: ability to do `download`, `build` and `install` as separate steps


matejkosik:
The functionality that we (part of the Coq team) currently miss is the ability to do these steps:
- `download`
- `build`
- `install`

separately. At the moment, `install` command does all that (an more).

The reason is that we are using OPAM to distribute Coq packages.
At the moment we are investigating whether we can rely on `opam` command in our benchmarking machinery. Natually, we would like to measure only `build` phase (not `download`, not `install`).

Currently, they are mixed together even though the necessary input for each these actions is already clearly separate.

I am not sure if this qualifies as a missing feature because I am not sure whether this could be useful to other team (I can imagine that it could be useful to Ocaml team, @gasche what do you think?).

Let me know whether this qualifies or not.

avsm:
This would also be useful for the OPAM CI, which could do sandboxed builds without external network access if it had commands to prepare the build environment (for Git clone etc) separately from the build phase.

The lack of external network access, combined with OPAM2's sandboxed installation, would improve security and reproducibility of builds significantly.

dra27:
@avsm - we can sort of do this with the developer mode proposed in #2937. Though even that would require URL rewriting to local copies for Git clones.

@matejkosik - it's probably not of much use to OCaml itself, as it sits below opam (and we have enough circular dependencies to contend with in FlexDLL and, soon, Menhir!), but regardless I think this is a good idea.

AltGr:
Having a download-only option that fills the local cache would be fairly easy. However, building without installing would be tricky for several reasons: ① packages expect their dependencies to be installed before they get built, ② the build directories are in general generated fresh and wiped after the build (although there is `--keep-build-dir`/`--reuse-build-dir`).

One solution that I could suggest for benchmarking, though, is to use the hooks. These need to be configured by hand in `~/.opam/config`, and are documented [here](http://opam.ocaml.org/doc/2.0/Manual.html#configfield-pre-build-commands). For example, you could use:

    wrap-build-commands: [ "/usr/bin/time" "-o" "/tmp/timings" "-a" "-f" "%{package}%: %e %C" ]

and then gather all the data from `/tmp/timings`. You probably also want to disable parallel builds (use `-j 1` or `OPAMJOBS=1`)

PR #4071, opened 2020-01-21
Add a --download-only flag to "opam install"

Armael:
This implements #4036  (cc @kit-ty-kate ).

This is still WIP because even when using the option, we get the usual message:
```
  ∗ install dune            2.1.3
  ∗ install conf-gmp        1      [required by z3]
...
===== ∗ 9 =====
Do you want to continue? [Y/n]
```

It would be better to recall explicitly at this point that in fact, this will only fetch the sources of the packages. I'm not sure how to do it though, in part because this message is currently printed using the graph before the build and fetch actions are made explicit, so we only have install/remove/reinstall/(change?) actions.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -21,6 +21,7 @@ New option/command/subcommand are prefixed with ◈.
 ## Install
   * The stdout of `pre-` and `post-session` hooks is now propagated to the user [#4382 @AltGr - fix #4359]
   * `post-install` hooks are allowed to modify or remove installed files, the but not add new ones. Those changes are integrated in changes file [#4388 @lefessan]
+  * ◈ Add `--download-only` flag [#4071 @Armael @rjbou - fix #4036]
 
 ## Remove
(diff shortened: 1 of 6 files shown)


rjbou:
Thanks for the pr!

For that, there is two way to proceed:
* update [`request`](https://github.com/ocaml/opam/blob/master/src/format/opamTypes.mli#L275) type to handle `fetch` actions, but it is maybe too intrusive for just a download only option
* handle this case without calling all install mechanism, but by resolving the given request and downloading all its dependencies. A function can be added in `OpamClient`.

Another thing, on the behavior: If the package if already installed, install mechanism would end on a noop. In the meantime, an opam clean can have been launched, and its dependencies no more available locally. In this case, should it download them even if installed (and its deps) or assume that it is installed, no need to?

rjbou:
Also related to #2895

rjbou:
Finalised, just waiting got cli versioning PR to be merged

PR #4072, opened 2020-01-24
OpamSystem.t_resolve_command: use ocaml code for looking up commands in PATH

Armael:
...instead of calling the shell (on linux).

The motivation is performance. It turns out that the speedup of #4064 is due not only to less time being spent launching processes, but also less time being spent calling to the shell in order to resolve the command. Doing the resolution (using PATH) directly in ocaml should be faster.

Note that command resolution was already done in ocaml on Windows, so this PR mostly merges the (previously) windows-specific code with the linux code.

The PR also attempts to streamline a bit the handling of `Not_found` vs `Denied`. The behavior is mostly unchanged when looking up a fully specified path (I added a `Not_found` case for increased precision; previously, a non-existing binary would be reported as "denied").
When a command is looked up in `PATH`, I follow the shell behavior a bit more consistently (as [specified here](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap08.html) "Other environment variables) and skip binaries that do not have the right permissions. 
If no suitable binary is found, then `Not_found` is returned (the previous code was inconsistent: it would've returned `Denied` under windows, but `Not_found` under linux).

### DIFF
--- a/src/core/opamSystem.ml
+++ b/src/core/opamSystem.ml
@@ -356,45 +356,41 @@ let t_resolve_command =
       with e -> OpamStd.Exn.fatal e; false
   in
   let resolve ?dir env name =
-    if not (Filename.is_relative name) then (* absolute path *)
-      if check_perms name then `Cmd name else `Denied
-    else if is_external_cmd name then (* relative *)
+    if not (Filename.is_relative name) then begin
+      (* absolute path *)
+      if not (Sys.file_exists name) then `Not_found
+      else if not (check_perms name) then `Denied
+      else `Cmd name
+    end else if is_external_cmd name then begin
+      (* relative path *)
       let cmd = match dir with
         | None -> name
         | Some d -> Filename.concat d name
       in
-      if check_perms cmd then `Cmd cmd else `Denied
-    else (* bare command, lookup in PATH *)
-    if Sys.win32 then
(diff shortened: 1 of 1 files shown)


Armael:
(incidentally, this PR happens to be fixing the cygwin CI (!))

AltGr:
This is great, thanks a bunch :)

dra27:
Coming very late to this, for which apologies - this didn't fix Cygwin CI, it masked the problem, owing I think to a bug in the implementation. Beforehand, `command -v cmd` was correctly returning `/cygdrive/c/WINDOWS/system32/cmd` and the permissions check was failing, so an exception was raised. After this change, the function considers `/cygdrive/c/WINDOWS/system32` in `PATH` but because it fails the permissions check (as it was before) it continues on to the next item in PATH. Except it never finds `cmd` in any other directory so now resolving the command returns `` `Not_found`` instead of `` `Denied``.

Failing the permissions check is valid to skip on to the next directory in PATH, but it's incorrect (I think) to return `` `Not_found`` if _any_ of the directories did contain the command, just with the appropriate permissions.

Armael:
Ah, wow, that makes sense indeed. Sorry for introducing that trick bug!

PR #4073, opened 2020-01-24
unpin: don't keep unpinned package version if it exists in repo

### LINKED ISSUE #3630: `opam pin remove` behaviour
I think this used to work in the v2 betas but it seems it no longer. 

When you release you usually have the package you release as a `pin` on the current version `X`. You submit to the opam repo version `X+1`. Then when you `opam update` and have the update `X+1` from the OCaml opam repository, you want to remove the pin and recompile with `X+1`. The current behaviour when you `opam pin remove` is to recompile with `X` which is annoying if your package has many deps, since you have to follow with a recompile with `X+1`.

rjbou:
Fixes #3630 , see discussion there

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -1475,14 +1475,10 @@ module PIN = struct
   let unpin st ?(action=true) names =
     let pinned_before = st.pinned in
     let st = unpin st names in
-    let available = Lazy.force st.available_packages in
     let installed_unpinned = (pinned_before -- st.pinned) %% st.installed in
     if action && not (OpamPackage.Set.is_empty installed_unpinned) then
       let atoms =
-        OpamPackage.Set.fold (fun nv acc ->
-            if OpamPackage.Set.mem nv available then
-              (nv.name, Some (`Eq, nv.version)) :: acc
-            else (nv.name, None) :: acc)
+        OpamPackage.Set.fold (fun nv acc -> (nv.name, None) :: acc)
           installed_unpinned []
       in
       upgrade_t
--- a/master_changes.md
+++ b/master_changes.md
@@ -7,6 +7,9 @@ note.
 ## Switch
   * Fix Not_found with `opam switch create . --deps` [#4151 @AltGr]
 
+## Pin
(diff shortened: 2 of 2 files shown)


src/client/opamClient.ml
```diff
@@ -1433,14 +1433,10 @@ module PIN = struct
...
     let st = unpin st names in
-    let available = Lazy.force st.available_packages in
     let installed_unpinned = (pinned_before -- st.pinned) %% st.installed in
     if action && not (OpamPackage.Set.is_empty installed_unpinned) then
       let atoms =
-        OpamPackage.Set.fold (fun nv acc ->
-            if OpamPackage.Set.mem nv available then
-              (nv.name, Some (`Eq, nv.version)) :: acc
-            else (nv.name, None) :: acc)
+        OpamPackage.Set.fold (fun nv acc -> (nv.name, None) :: acc)
```
AltGr: I think this is the good approach, but I we should probably keep some check on availability (for _any_ version of the package, to remain consistent).

Otherwise, I imagine the following scenario could happen:
- package `foo` was unavailable, but you wanted to try anyway
- so you pinned it to change its constraints and managed to install it
- now you want to unpin it
→ without the check for availability, we would request `upgrade(foo)`, and the solver would fail ("no available package named 'foo'" — leaving you with your pinned package installed until you remove or upgrade all manually!)
→ with a check, ideally, we would remove `foo` from the constraint, and the `upgrade` would naturally remove it

However, we could still encounter the case where the solver fails due to more complex constraints. _Maybe_ this is a good case for a use of `best_effort` instead? It would upgrade the packages if possible, but guarantee consistency anyway by removing them if needed. Could there be unwanted side-effects to this approach ?

rjbou: FWIS, all those checks are already done in the `compute_ugrade` function: as `foo` is an orphan, it will be added to the remove set in the solver request.

AltGr: Ah indeed, `compute_upgrade` has a special handling for that case :+1:

PR #4074, opened 2020-01-24
env: add `--status` option that indicates by its return code if opam environment is synchronised

### LINKED ISSUE #3725: Feature request: command to show env status
I'm always having trouble with is keeping track of when I need to run `eval $(opam env)`. For example when cd-ing between a local switches. If there was a quick way to query opam about when I needed to update my env it would be easy to write a shell plugin that automatically reconfigures the env when needed.

For example a command like this would do the trick:

```
opam env --status
```

That returns `Env is up to date` or `Run 'eval $(opam env)'`

This might also be useful for dune to give a hint that you forgot to configure the env when your build fails.

Does this feature make sense? In that case I could try to put together some kind of pull request.

rjbou:
Fixes #3725


src/client/opamCommands.ml
```diff
@@ -1181,10 +1181,22 @@ let env =
     mk_flag ["revert"]
       "Output the environment with updates done by opam reverted instead."
   in
+  let status =
+    mk_flag ["status"]
+      "Fail if environment is not synchronised. In that case display not upt to date variables"
```
AltGr:
```suggestion
      "Exits with 0 if the environment is already up-to-date, 1 otherwise, after \
       printing the list of not up-to-date variables."
```

AltGr: I would also rename the option with `--check`, since it is a bit similar to `opam update --check`.

src/client/opamCommands.ml
```diff
@@ -1181,10 +1181,22 @@ let env =
...
       global_options shell sexp inplace_path set_opamroot set_opamswitch
-      revert =
+      revert status =
     apply_global_options global_options;
+    if status then
+      OpamGlobalState.with_ `Lock_none @@ fun gt ->
+      OpamSwitchState.with_ `Lock_none gt @@ fun st ->
+      if not (OpamEnv.status st) then
+        OpamStd.Sys.exit_because `False;
+      OpamSwitchState.drop st;
```
AltGr: `drop` is not needed within a `with_`

src/client/opamCommands.ml
```diff
@@ -1197,7 +1209,8 @@ let env =
        | Some sw ->
          OpamConfigCommand.env gt sw
            ~set_opamroot ~set_opamswitch
-           ~csh:(shell=SH_csh) ~sexp ~fish:(shell=SH_fish) ~inplace_path)
+           ~csh:(shell=SH_csh) ~sexp ~fish:(shell=SH_fish) ~inplace_path);
+      OpamGlobalState.drop gt
```
AltGr: unneeded

src/state/opamEnv.ml
```diff
@@ -639,14 +639,21 @@ let update_user_setup root ?dot_profile shell =
...
+
 let check_and_print_env_warning st =
   (* if you are trying to silence this warning,
      set the ~no_env_notice:true flag from OpamStateConfig,
      which is checked by (is_up_to_date st). *)
   if not (is_up_to_date st) &&
      (OpamFile.Config.switch st.switch_global.config = Some st.switch ||
       OpamStateConfig.(!r.switch_from <> `Command_line))
-  then
+      then
```
AltGr: ??

src/state/opamEnv.ml
```diff
@@ -639,14 +639,21 @@ let update_user_setup root ?dot_profile shell =
     OpamStd.Option.iter (fun f -> update_dot_profile root f shell) dot_profile
   )
 
+let status st =
+  OpamStateConfig.update ~no_env_notice:false ();
```
AltGr: Don't do side effect in a function that should be pure according to what the interface lets assume.
Either add an optional argument to `is_up_to_date`, or warn against `no_env_notice` in the .mli.

src/state/opamEnv.ml
```diff
@@ -639,14 +639,21 @@ let update_user_setup root ?dot_profile shell =
...
+  not
+    (OpamFile.Config.switch st.switch_global.config = Some st.switch ||
+     OpamStateConfig.(!r.switch_from <> `Command_line)) ||
```
AltGr: Why these conditions ? This would lead e.g. `opam env --check --switch foo` to always return 0, which is not what I would expect. Just `is_up_to_date` seems to be what you want here, maybe aside from the `no_env_notice` setting...

src/state/opamEnv.ml
rjbou: dunno ^^

src/state/opamEnv.ml
```diff
@@ -639,6 +639,8 @@ let update_user_setup root ?dot_profile shell =
     OpamStd.Option.iter (fun f -> update_dot_profile root f shell) dot_profile
   )
 
+let check st = is_up_to_date ~check:true st
```
AltGr: I would remove this function, it seems it's no longer needed (and `is_up_to_date` has a clearer name)

src/state/opamEnv.ml
```diff
@@ -280,8 +280,8 @@ let get_full
   let updates = u @ updates ~set_opamroot ~set_opamswitch ~force_path st in
   add env0 updates
 
-let is_up_to_date_raw updates =
-  OpamStateConfig.(!r.no_env_notice) ||
+let is_up_to_date_raw ?(check=false) updates =
```
AltGr: I would rename the argument to `skip` instead, and define
```
let is_up_to_date_raw ?(skip=OpamStateConfig.(!r.no_env_notice)) updates = ...
```
I believe it would make things clearer.

rjbou:
Updated!

PR #4081, opened 2020-02-02
[docs] Fixed git command in packaging docs

roddyyaga:
Currently [the docs](https://opam.ocaml.org/doc/Packaging.html) suggest to do
```
git tag -a 0.1
git push 0.1
```
I think the second command should be `git push origin 0.1` (`git push 0.1` tries to push to repository called 0.1).


rjbou:
Thanks!

PR #4086, opened 2020-02-13
repo: on add error, display also the parsed url

### LINKED ISSUE #4085: opam repo add gives errors for file target where it should not
This is tested with opam 2.0.5.

When I add a local repo to some switch as follows:
```
opam repo --this-switch add devel file:$HOME/Opam/devrepo
```
this works fine. But when I later add the same repo to a different switch I get the error message:
``` 
[ERROR] Repository devel is already set up and points to file:///Users/<me>/Opam/devrepo. To change that, use
        'opam repository set-url'.
```
This is not in line with the fact that I can add a git based repo to as many switches I want with e.g.:
```
opam repo add coq-released https://coq.inria.fr/opam/released
```
Also if I leave away the file url, it works fine:
```
opam repo --this-switch add devel
```
So I would say the behaviour is unexpected since adding the repo with explicitly giving an URL is a legitimate operation and also this is inconsistent between file:// and https://.

rjbou:
When `repo add name url` fails because of an already registered repo mismatching url, printing also the opam parsed resulting url helps to understand the error origin (missing `//`, etc.)
Fixes #4085

### DIFF
--- a/src/client/opamRepositoryCommand.ml
+++ b/src/client/opamRepositoryCommand.ml
@@ -64,18 +64,20 @@ let add rt name url trust_anchors =
     -> rt
   | Some r ->
     OpamConsole.error_and_exit `Bad_arguments
-      "Repository %s is already set up %s. To change that, use 'opam \
-       repository set-url'."
+      "Repository %s is already set up%s. To change that, use 'opam \
+       repository set-url %s %s'."
       (OpamRepositoryName.to_string name)
       (if r.repo_url <> url then
-         "and points to "^OpamUrl.to_string r.repo_url
+         " and points to "^OpamUrl.to_string r.repo_url
        else match r.repo_trust with
-         | None -> "without trust anchors"
+         | None -> " without trust anchors"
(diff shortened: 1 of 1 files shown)


MSoegtropIMC:
I think I would prefer if eroneous URLs would be rejected or one would get a warning the *first* time one does `opam repo add`. As far as I understand opam silently fixes the URL. If it does, one could give a warning like "The url %s is incorrect and has been modified to %s". I think what confused me is that the URL is consdiered OK once, but not in subsequent calls.

rjbou:
The url is not wrong itself (in that case it is another error that is raised), but its resolution can be different from what user wanted to. In your case, `file:$HOME/...` resolves to an `ssh` url `file:/home/<you>/...`. If on your first repository add your repository with that url, the update during that add would have failed.

MSoegtropIMC:
I thought I always used the same URL since I did this with scripts, but maybe I did some accidential change in between or possibly changed ~ with $HOME and and this change this way. I will test this.

PR #4087, opened 2020-02-13
Sandbox: fix some cache failures

### LINKED ISSUE #4068: Regression: Opam 2.0.6 fails any build with cache-related bwrap error [cache as link]
Possibly related to #4012, after updating from 4.0.5 to 4.0.6 I can no longer upgrade anything.  I consistently get, for packages I tried (`ppx_tools`, `dune`):

```text
### output ###
# bwrap: Can't mkdir /home/lis/.cache/dune: No such file or directory
```

Note that in my setup, `/home/lis/.cache` is a symbolic link to `/tmp/` (which is in RAM with `tmpfs`). If I remove directory `/tmp/dune` it does get re-created, despite what the above error implies.

Downgrading to 4.0.5 restores normal behavior, hence my suspicion that opam may be at fault.

```text
lis@titanium:~$ opam config report
# opam config report
# opam-version      2.0.6
# self-upgrade      no
# system            arch=x86_64 os=linux os-distribution=debian os-version=10.2
# jobs              7
# repositories      1 (http) (default repo at 4585683c)
# pinned            0
```

### LINKED ISSUE #4079: Sandbox script doesn't work with newer versions of ccache
[It seems](https://ccache.dev/releasenotes.html#_ccache_3_7) that ccache 3.7 changed the name of its `--print-config` argument to `--show-config`. This means that the sandboxing script is now broken if you use ccache with a customized `ccache_dir`. It seems that `--print-config`/`--show-config` is available as `-p` both before and after the rename -- so the easiest fix is probably to use that.

rjbou:
* ccache:
  * fix option name
  * handle `CCACHE_DIR` env variable in case `ccache` call fails
  * fix #4079 
* cache:
  * follow links of `~/.cache` & `~/.cache/dune` for bwrap call (maybe we just need to bind just `.cache`?)
  * is it needed on mac ?
  * fix #4068

### DIFF
--- a/src/state/shellscripts/bwrap.sh
+++ b/src/state/shellscripts/bwrap.sh
@@ -61,15 +61,17 @@ add_sys_mounts /usr /bin /lib /lib32 /lib64 /etc /opt /home /var
 # that remain writeable. ccache seems widespread in some Fedora systems.
 add_ccache_mount() {
   if command -v ccache > /dev/null; then
-      CCACHE_DIR=$HOME/.ccache
       ccache_dir_regex='cache_dir = (.*)$'
       local IFS=$'\n'
-      for f in $(ccache --print-config 2>/dev/null); do
+      for f in $(ccache -p 2>/dev/null); do
         if [[ $f =~ $ccache_dir_regex ]]; then
-          CCACHE_DIR=${BASH_REMATCH[1]}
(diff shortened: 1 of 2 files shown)


rjbou:
Tested on MacOs Catalina 10.15.3 , no need to update mac script

Issue #4091, opened 2020-02-19
Opam "cannot revert" some things "while removing", but it is unclear why


RalfJung:
On CI, we are seeing lots of logs like [this](https://gitlab.mpi-sws.org/iris/iris/-/jobs/56790):
```
 [WARNING] While removing coq-stdpp.dev.2020-02-11.3.e95c967d: cannot revert:
            - modifications of lib/coq/user-contrib/stdpp/zmap.vo
            - modifications of lib/coq/user-contrib/stdpp/zmap.v
            - modifications of lib/coq/user-contrib/stdpp/zmap.glob
            - modifications of lib/coq/user-contrib/stdpp/vector.vo
            - modifications of lib/coq/user-contrib/stdpp/vector.v
            - modifications of lib/coq/user-contrib/stdpp/vector.glob
            - modifications of lib/coq/user-contrib/stdpp/telescopes.vo
            - modifications of lib/coq/user-contrib/stdpp/telescopes.v
            - modifications of lib/coq/user-contrib/stdpp/telescopes.glob
            - modifications of lib/coq/user-contrib/stdpp/tactics.vo
            - modifications of lib/coq/user-contrib/stdpp/tactics.v
            - modifications of lib/coq/user-contrib/stdpp/tactics.glob
            - modifications of lib/coq/user-contrib/stdpp/strings.vo
            - modifications of lib/coq/user-contrib/stdpp/strings.v
            - modifications of lib/coq/user-contrib/stdpp/strings.glob
            - modifications of lib/coq/user-contrib/stdpp/stringmap.vo
            - modifications of lib/coq/user-contrib/stdpp/stringmap.v
            - modifications of lib/coq/user-contrib/stdpp/stringmap.glob
            - modifications of lib/coq/user-contrib/stdpp/streams.vo
            - modifications of lib/coq/user-contrib/stdpp/streams.v
            - modifications of lib/coq/user-contrib/stdpp/streams.glob
[...]
```
Unfortunately, it is entirely unclear (a) what exactly this means, and (b) why it happens. Does this mean it tried to remove the file but failed due to permissions? Does this mean it did not even try to remove the file? Does it mean it actually removed the file but something was odd?  Given this is a warning (and not a fatal error), I would assume the files actually do get removed appropriately, but I cannot tell. (This is happening on CI, so interactive debugging is not possible.) Also, nothing is ever touching those files besides opam (and the CI machinery creating/restoring the cache), so it is not clear what the cause of the issue is.

[...]

PR #4094, opened 2020-02-20
Tracking: log & force

rjbou:
* Add precision of kind of tracking, and gather together kinds
* propagate `--force` on removal to changes revert
Related to #4091

### DIFF
--- a/src/core/opamDirTrack.ml
+++ b/src/core/opamDirTrack.ml
@@ -224,9 +224,13 @@ let revert ?title ?(verbose=OpamConsole.verbose()) ?(force=false)
   if already <> [] then
     log ~level:2 "%sfiles %s were already removed" title
       (String.concat ", " (List.rev already));
-  if modified <> [] && verbose then
-    OpamConsole.warning "%snot removing files that changed since:\n%s" title
-      (OpamStd.Format.itemize (fun s -> s) (List.rev modified));
+  if modified <> [] then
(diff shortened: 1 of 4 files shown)


rjbou:
Also, for the moment, revert with `force` only removes added files that changed since install, but not those that have been modified by install, nor kind changed.
A remove can failssilently (log are only when verbose) et keep the package in an inconsistent state: if a file remain, rerunning an `opam remove` even with `--force` won't remove all files, with no notification.
Maybe just need to force logging when `--force` is given...

rjbou:
Updated following

src/core/opamDirTrack.ml
```diff
@@ -224,14 +224,40 @@ let revert ?title ?(verbose=OpamConsole.verbose()) ?(force=false)
   if already <> [] then
     log ~level:2 "%sfiles %s were already removed" title
       (String.concat ", " (List.rev already));
-  if modified <> [] && verbose then
-    OpamConsole.warning "%snot removing files that changed since:\n%s" title
-      (OpamStd.Format.itemize (fun s -> s) (List.rev modified));
+  if modified <> [] then
+    if OpamConsole.confirm ~default:false
+        "%snot removing files that changed since:\n%sRemove them anyway?" title
```
AltGr:
```suggestion
        "%sthese files have been modified since installation:\n%sRemove them anyway?" title
```

PR #4095, opened 2020-02-21
shell: fix zsh check interactive terminal

OCamlPro-mattiasdrp:
instead of -t 0 since it can mess up with some shells

(-t 0 checks if the shell is tty, not interactive, even though most of the times it's ok to do so it can fail with some shells)

### DIFF
--- a/src/state/opamEnv.ml
+++ b/src/state/opamEnv.ml
@@ -463,9 +463,11 @@ let source root shell f =
     Printf.sprintf "source %s >& /dev/null || true\n" (file f)
   | SH_fish ->
     Printf.sprintf "source %s > /dev/null 2> /dev/null; or true\n" (file f)
-  | SH_sh | SH_bash | SH_zsh ->
+  | SH_sh | SH_bash ->
     Printf.sprintf "test -r %s && . %s > /dev/null 2> /dev/null || true\n"
       (file f) (file f)
+  | SH_zsh ->
+    Printf.sprintf "[[ ! -r %s ]] || source %s" (file f) (file f)
 
 let if_interactive_script shell t e =
   let ielse else_opt = match else_opt with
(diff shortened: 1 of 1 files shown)


rjbou:
Thanks for the proposal!
Related zsh issue romkatv/powerlevel10k#516 , those hints are the good way to go imho

src/state/opamEnv.ml
```diff
@@ -463,18 +463,22 @@ let source root shell f =
     Printf.sprintf "source %s >& /dev/null || true\n" (file f)
   | SH_fish ->
     Printf.sprintf "source %s > /dev/null 2> /dev/null; or true\n" (file f)
-  | SH_sh | SH_bash | SH_zsh ->
+  | SH_sh | SH_bash ->
     Printf.sprintf "test -r %s && . %s > /dev/null 2> /dev/null || true\n"
       (file f) (file f)
+  | SH_zsh ->
+    Printf.sprintf "[[ ! -r %s ]] || source %s" (file f) (file f)
```
rjbou:
```suggestion
    Printf.sprintf "[[ ! -r %s ]] || source %s  > /dev/null 2> /dev/null" (file f) (file f)
```

There is missing redirections here.

src/state/opamEnv.ml
```diff
@@ -463,18 +463,22 @@ let source root shell f =
     Printf.sprintf "source %s >& /dev/null || true\n" (file f)
   | SH_fish ->
     Printf.sprintf "source %s > /dev/null 2> /dev/null; or true\n" (file f)
-  | SH_sh | SH_bash | SH_zsh ->
+  | SH_sh | SH_bash ->
     Printf.sprintf "test -r %s && . %s > /dev/null 2> /dev/null || true\n"
       (file f) (file f)
+  | SH_zsh ->
+    Printf.sprintf "[[ ! -r %s ]] || source %s  > /dev/null 2> /dev/null" (file f) (file f)
```
romkatv: Consider replacing this:

```zsh
source %s  > /dev/null 2> /dev/null
```

With this:

```zsh
source %s &> /dev/null
```

Just one space after `%s` instead of two and `&>` to redirect both stdout and stderr at once.

If you can drop the redirection, that would be even better. Hiding potential errors like that can make it very difficult to identify problems. If some parts of the sourced file can produce unwanted console output, those parts should have redirections rather than doing it wholesale.

rjbou: First we didn't use it as it is not POSIX compatible, but for `zsh` it can be considered. It is supported since the beginning or there is a min version ?
From what I see, it can misbehave in case of [multios](http://zsh.sourceforge.net/Doc/Release/Redirection.html#Multios), but we're not in that case.

On not dropping outputs, I think it was done in purpose at first, and those scripts evolved. After a quick look, all called outputable programs (usually opam) have a stderr redirection to `/dev/null`. Removing them can be considered, indeed. ping @AltGr

romkatv: `&>` has been supported in Zsh since the beginning. It doesn't misbehave when `multios` is on.

PR #4096, opened 2020-02-21
Use proper module names

rgrinberg:
get-git-version isn't a proper module name. This commit switches it into
a valid module name.

In 2.3.0, dune main validation more strict. So we ended up accidentally
breaking this. Opam should not rely on this.

### DIFF
--- a/src/client/dune
+++ b/src/client/dune
@@ -2,7 +2,7 @@
   (name        opam_client)
   (public_name opam-client)
   (synopsis    "OCaml Package Manager client and CLI library")
-  (modules     (:standard \ opamMain get-git-version))
+  (modules     (:standard \ opamMain get_git_version))
   (libraries   opam-state opam-solver re cmdliner)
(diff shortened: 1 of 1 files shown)


dra27:
It’s not a problem to change this in opam, but it feels like a bug in the new Dune interpretation - the file is _not_ an OCaml module, it’s a script so its file name shouldn’t matter. Shouldn’t (or even does?) Dune now just ignore .ml files which aren’t valid module names?

rgrinberg:
I thought it was a bit strange to allow non module names in the modules field.

Dune’s previous behaviour was to allow to exclude these files in the modules field, but if you forgot to do that, it would give a cryptic error when it would build an alias file out of them (for example)

The planned behaviour for dune is to just warn on such module names. So this change isn’t technically necessary.

I realize that this isn’t ideal, but I thought that it would be best to give clear errors/warnings early on even if some edge cases are less convenient.

PR #4097, opened 2020-02-23
Bump compat releases (for 4.09.0 with dune 2.3.0)

tuncer:
src_ext changes needed to fix building with 4.09.0 and dune 2.3.0. Additionally, where possible, URLs have been upgraded to https.

### DIFF
--- a/src_ext/Makefile.sources
+++ b/src_ext/Makefile.sources
@@ -1,5 +1,5 @@
-URL_cppo = https://github.com/mjambon/cppo/archive/v1.6.5.tar.gz
-MD5_cppo = 1cd25741d31417995b0973fe0b6f6c82
+URL_cppo = https://github.com/mjambon/cppo/archive/v1.6.6.tar.gz
+MD5_cppo = 2258f90f3b7f0190bda76b710df0af56
 
 $(call PKG_SAME,cppo)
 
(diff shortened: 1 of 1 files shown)


tuncer:
CI fails with 4.02.3 (https://travis-ci.org/ocaml/opam/jobs/654180813)

rjbou:
Thanks for the proposal!

It fails because [`cppo`](https://opam.ocaml.org/packages/cppo/) is not compatible with `ocaml <4.03`. For the moment, we keep opam compatible with `ocaml.4.02.3`.

tuncer:
How can we support 4.09 and newer, while also building on 4.02? Maybe check `ocaml -vnum` in `Makefile.sources`?

tuncer:
Alternatively, we could bump the minimum OCaml version, if that's acceptable. However, I suspect that would break compatibility promises, so no.

tuncer:
Is it correct that opam `master` or `2.0` should build on 4.02 and 4.10, or is the idea that one must keep a `[4.02..4.08]` switch around for that purpose? Or is the right way bootstraping opam from scratch, ignoring switches from `~/.opam` or any other global ocaml?

rjbou:
It is correct, on the idea, we want opam compatible with 4.02.3 → 4.10.0. Concretely, opam packages are ftm not 4.10.0 compatible (missing `stdlib-shims` for 4.10.0, and adding it via a variable `ocaml:version` is not possible —package vars are not resolved at this step).
There is no right way to install opam, it differs from several user workflows.

But here, the blocking point is only about `cppo.1.6.6` that is incompatible with `ocaml.4.02.3` and concerns only the opamless opam install. Keeping it to `1.6.5` would be ok. I just tested on a switch with `ocaml.4.10.0` & `cppo.1.6.5` with no error.

Bumping the minimum ocaml version should be discussed. With buster release, it is no more so blocking (even if it remains the default in oldstable), and most os/distro have their packages [> 4.02.3](https://repology.org/project/ocaml/versions).

tuncer:
Only cppo.1.6.6 includes https://github.com/ocaml-community/cppo/commit/2d93c68b9af5748fc8fa28605cabf2ec1445d130, which allows building with a newer dune. I think you're saying we can't resolve the situation by conditionally selecting lib-ext sources, correct? If that's the case, maybe I should try using `make cold` instead.

rjbou:
The `stdlib-shims` seems to be a local issue (mine), not seen as an issue during the 4.10 ecosystem testing. 

On having a conditional selecting lib-ext source, it depends on `make cold` purpose: just permitting to compile from scratch (in that case some changes are needed but it's possible), or also ensure to always have the same bundle.

tuncer:
I don't mind using `make cold`, if it works reliably and is a supported option. That said, are you saying we should try to make the `./configure && make lib-ext && make` route work with 4.02 up to 4.10? If so, what can I do to fix the issues?

rjbou:
lib-ext is used also for m`make cold`, that's why i'm talking about it.

And yes, we want to keep ftm `./configure && make lib-ext && make` work for ocaml 4.02.3 → 4.10.
After discussion with @AltGr, we have to remain unicity of sources of lib-ext, in order to always have the same bundle of version, for each opam version.
For 2.0 branch, we don't bump version unless it is needed, but we can on master, btu the [problem](https://travis-ci.org/ocaml/opam/jobs/654183335#L1173) is that cmdliner bump have a 4.03.0 restriction.

rjbou:
Closing this PR based on 2.0.

PR #4098, opened 2020-02-23
Bump compat releases (for 4.09.0 with dune 2.3.0) 

tuncer:
Same as #4097, but against master.

### DIFF
--- a/src_ext/Makefile.sources
+++ b/src_ext/Makefile.sources
@@ -43,8 +43,8 @@ MD5_opam-file-format = d7852cb63df0f442bed14ba2c5740135
 
 $(call PKG_SAME,opam-file-format)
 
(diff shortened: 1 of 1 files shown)


tuncer:
This diff is the same as #4097, but it fails to build `src_ext/cmdliner` for some reason, even though no such error happens in #4097. I suspect a difference between `master` and `2.0`, not sure what. Any ideas?

dra27:
2.0 shouldn’t have deps bumped until it’s absolutely necessary (cmdliner was already lagging)

tuncer:
It's what I had to change to make it work, but I know it's tricky to support a wide range of OCaml switches. @dra27, let's continue the discussion in #4097, if you don't mind.

dra27:
This is superseded by #4257

PR #4109, opened 2020-03-14
Update mccs

MisterDA:
Turns out mcss was updated for 2.0.6, but not on the master branch. Also, MSVC and G++ in C++17 mode won’t compile mccs without the second patch. I have opened a PR: https://github.com/AltGr/ocaml-mccs/pull/29.


rjbou:
Added a commit to your branch to fix the reworking commit error at hygiene stage.

AltGr:
Thanks!

PR #4111, opened 2020-03-15
ensure that Env.get is initialized lazily, not before .init functions are called

gasche:
(joint work with Armaël Guéneau)

OpamStd.Env.get is opam's entry point to access the process
environment. It caches the whole environment lazily, on the first
environment access. Most environment-accessing parts of the code are
written with carefully placed lazy thunks, to ensure that they are
computed on first need (when the .init functions are called) rather
than at module initialization. This PR fixes a few places that would
still force the environment at module-initialization time.

This is important for programs that would like to modify the
environment (using Unix.putenv) before calling opam-lib functions, and
ensure that opam functions do see the modified environment. Before
this commit, the environment was cached at module-loading time and
could never be modified by user code.

We had to change the definition of `OpamSys.home ()` slightly: this
function is forced at module-initialization time
(in OpamStateConfig.default) in a way that seems hard to remove, so
this function now makes a direct call to `Unix.getenv` instead of
going through the `Env` cache.

(This changes slightly its semantics as previously, on Windows,
environment lookup would happen modulo-case, so the user would be able
to put "Home" or "home" in their environment. We believe that this is
acceptable -- most Windows user do not have a HOME environment
variable at all anyway.)

We think it would be nice to have a testsuite item to ensure that the
environment remains loaded lazily (not before .init() functions
are called), but we are not completely sure where/how to write this
testcase, so would rather leave this part to someone else.

### DIFF
--- a/src/core/opamProcess.ml
+++ b/src/core/opamProcess.ml
@@ -572,9 +572,9 @@ let verbose_print_cmd p =
      else Printf.sprintf " (CWD=%s)" p.p_cwd)
 
 let verbose_print_out =
-  let pfx = OpamConsole.colorise `yellow "- " in
+  let pfx = lazy (OpamConsole.colorise `yellow "- ") in
   fun s ->
-    OpamConsole.msg "%s%s\n" pfx s
+    OpamConsole.msg "%s%s\n" (Lazy.force pfx) s
 
 (** Semi-synchronous printing of the output of a command *)
 let set_verbose_f, print_verbose_f, isset_verbose_f, stop_verbose_f =
--- a/src/core/opamStd.ml
+++ b/src/core/opamStd.ml
@@ -607,7 +607,6 @@ type warning_printer =
   {mutable warning : 'a . ('a, unit, string, unit) format4 -> 'a}
 let console = ref {warning = fun fmt -> Printf.ksprintf prerr_string fmt}
 
-
 module Env = struct
 
   (* Remove from a c-separated list of string the one with the given prefix *)
@@ -709,7 +708,7 @@ module OpamSys = struct
 
   let tty_in = Unix.isatty Unix.stdin
 
-  let default_columns =
+  let default_columns = lazy (
     let default = 16_000_000 in
     let cols =
       try int_of_string (Env.get "COLUMNS") with
@@ -717,6 +716,7 @@ module OpamSys = struct
       | Failure _ -> default
     in
     if cols > 0 then cols else default
+  )
 
   let get_terminal_columns () =
     let fallback = 80 in
@@ -757,15 +757,19 @@ module OpamSys = struct
       fun () ->
         if tty_out
         then win32_get_console_width ()
-        else default_columns
+        else Lazy.force default_columns
     else
       fun () ->
         if tty_out
         then Lazy.force !v
-        else default_columns
+        else Lazy.force default_columns
 
   let home =
-    let home = lazy (try Env.get "HOME" with Not_found -> Sys.getcwd ()) in
+    (* Note: we ask Unix.getenv instead of Env.get to avoid
+       forcing the environment in this function that is used
+       before the .init() functions are called -- see
+       OpamStateConfig.default. *)
+    let home = lazy (try Unix.getenv "HOME" with Not_found -> Sys.getcwd ()) in
     fun () -> Lazy.force home
 
   let etc () = "/etc"
--- a/src/core/opamSystem.ml
+++ b/src/core/opamSystem.ml
@@ -797,26 +797,31 @@ module Tar = struct
         Some (Printf.sprintf "Tar needs %s to extract the archive" cmd)
       else None)
 
-  let tar_cmd =
+  let tar_cmd = lazy (
     match OpamStd.Sys.os () with
     | OpamStd.Sys.OpenBSD -> "gtar"
     | _ -> "tar"
+  )
+
+  let cygpath_tar = lazy (
+    Lazy.force (get_cygpath_function ~command:(Lazy.force tar_cmd))
+  )
 
   let extract_command =
-    let f = get_cygpath_function ~command:tar_cmd in
     fun file ->
       OpamStd.Option.Op.(
         get_type file >>| fun typ ->
-        let f = Lazy.force f in
+        let f = Lazy.force cygpath_tar in
+        let tar_cmd = Lazy.force tar_cmd in
         let command c dir =
           make_command tar_cmd [ Printf.sprintf "xf%c" c ; f file; "-C" ; f dir ]
         in
         command (extract_option typ))
 
   let compress_command =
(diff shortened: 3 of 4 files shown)


gasche:
(cc @Armael)

hannesm:
I recently encountered a similar issue "opam captures the environment before I can run any code to `unsetenv` parts of the environment" (and managed to work around it in https://github.com/hannesm/orb/blob/active/src/orb_stub.ml which is executed before the opam loading :). This patch would help me to cleanup that code quite a bit (the extra module with lots of top-level statements would not be necessary at all if I understand correctly).

gasche:
Here is my testing methodology to check that this PR is effective (and it's also a debugging methodology to find where to test when it is not): I insert an `assert false` at the place where the environment is forced first:

```diff
diff --git a/src/core/opamStd.ml b/src/core/opamStd.ml
index a5cd1664..9c46dc9d 100644
--- a/src/core/opamStd.ml
+++ b/src/core/opamStd.ml
@@ -629,6 +629,7 @@ module Env = struct
 
   let list =
     let lazy_env = lazy (
+      assert false;
       let e = Unix.environment () in
       List.rev_map (fun s ->
           match OpamString.cut_at s '=' with
```

Without the PR (and with the assert false), many things break during the `make` build, and running the `opam` binary with `OCAMLRUNPARAM=b` gives the following backtrace:

```
Fatal error: exception File "opam/src/core/opamStd.ml", line 633, characters 6-12: Assertion failed
Raised at file "opam/src/core/opamStd.ml", line 633, characters 6-18
Called from file "camlinternalLazy.ml", line 31, characters 17-27
Re-raised at file "camlinternalLazy.ml", line 36, characters 4-11
Called from file "opam/src/core/opamStd.ml" (inlined), line 649, characters 28-37
Called from file "opam/src/core/opamStd.ml", line 716, characters 24-43
```

 This shows that the environment is forced by some toplevel expression in `opamStd`, too early. (Now one can iterate: fix this toplevel expression, repet the test).

With the PR (and the assert false), almost nothing breaks during `make` (only one call to `opam-installer`), and running the binary with `OCAMLRUNPARAM=b` gives the following trace:

```
File "opam/src/core/opamStd.ml", line 632, characters 6-12: Assertion failed
Fatal error: exception File "opam/src/core/opamStd.ml", line 632, characters 6-12: Assertion failed
Raised at file "camlinternalLazy.ml", line 35, characters 56-63
Called from file "camlinternalLazy.ml", line 31, characters 17-27
Re-raised at file "camlinternalLazy.ml", line 36, characters 4-11
Called from file "opam/src/core/opamStd.ml" (inlined), line 648, characters 28-37
Called from file "opam/src/core/opamStd.ml", line 697, characters 35-51
Called from file "opam/src/core/opamStd.ml", line 726, characters 8-89
Called from file "camlinternalLazy.ml", line 31, characters 17-27
Re-raised at file "camlinternalLazy.ml", line 36, characters 4-11
Called from file "opam/src/core/opamStd.ml" (inlined), line 1211, characters 44-71
Called from file "opam/src/core/opamStd.ml", line 1247, characters 16-59
Called from file "opam/src/core/opamStd.ml", line 79, characters 20-23
Called from file "list.ml", line 121, characters 24-34
Called from file "opam/src/core/opamStd.ml", line 78, characters 8-189
Called from file "opam/src/core/opamStd.ml", line 1279, characters 8-78
Called from file "opam/src/client/opamMain.ml", line 184, characters 32-64
Called from file "opam/src/client/opamMain.ml", line 245, characters 2-48
```

This is a good trace, as the environment is forced in `opamMain` and not before. (opam-lib users do not link `opamMain`, so they will not get the environment forced unless they call an `.init` function explicitly.)

Armael:
I feel like it is a bit ad-hoc to change the implementation of `home` because we know it is going to be called at module-initialization time. To me, it seems cleaner to instead change the code that queries the environment at initialization time in order to avoid forcing the lazy environment.

One solution would be to introduce a variant of `home` called e.g. `uncached_home` that contains the uncached query to the environment, and would used instead of `home` in places that need it at module initialization time.

gasche:
@Armael on your suggestion I tried your approach, and I gave up (for now) as it is more invasive than we thought.

There are several functions (I found at least `OpamFilename.Dir.of_string` and `OpamFilename.prettify_path` that use `Sys.home ()` and are called both during module-initialization and after. If I migrate them to `uncached_home`, I risk a performance regression. I could implement caching for each of those locally, but this makes the whole thing more complex. I could also propose `Sys.separately_cached_home` instead of `uncached_home`, but this seems overly complex to me. I suspect that keeping a white lie about `Sys.home ()` may be the best option.

rjbou:
Is it ok to: add a `set` function in `OpamEnv`, that forces `get`/`list` evaluation, and updates it. Like that changes are minimal (`OpamEnv` located), and when environment is retrieved, it is the updated one.
This can be applied only if opam library is used to update the environment. If in a program, both libraries need to take into account some env variables, there will be duplicated code (`Unix.putenv` & `OpamStd.Env.set`).

@gasche for test, there is no test using directly opam libs:
* `make test` do some basic install / reinstall tests (python script)
* `opam-rt` do some more advanced tests (ocaml program, calling opam externally). But it already depends on opam lib, it is possoble to add in it some opam lib tests.
* `make reftester` do some opam output check

gasche:
@rjbou but then opam has to deal with the fact that the value of an environment variable may change over time, which I think is part of what you tried to avoid (along with a performance hit) with the caching-once-and-for-all interface. (The current patchset shows a glimpse of this unpleasantness with `Sys.home ()` which may different from `Env.get "HOME"`, but this is just a small glimpse.)

The changes in the PR touch several parts of the codebase, but I think of them as bugfixes. My mental model (purely guess-based) is that many parts of the existing code where written with the idea of being lazy ( see for example https://github.com/ocaml/opam/pull/4111/files#diff-e870142f1e5369ee452b98457cf7369cL748-L765 or https://github.com/ocaml/opam/pull/4111/files#diff-e870142f1e5369ee452b98457cf7369cR806-R823 or https://github.com/ocaml/opam/pull/4111/files#diff-9956c59aca0fd1cd396e711776733833L418-L427 ), and some parts crept in over time that were in fact not lazy, and would force large part of the rest to be less lazy than initially hoped.

rjbou:
@gasche yes, i was answering/thinking to the impossibility to change/act on environment when using opam-lib. You come accross this problem now with home, but other variables in the future, for other uses, can be impacted. Having the changes in the `OpamStd.Env` with a `set` function can be solution (and we won't need to change specifically `OpamStd.Sys.home`), with a way to ensure that environment modification is only done once, before forcing `lazy_env`.

The PR itself is ok for me, `Sys.home` modification is an aceptable workaround ftm. Besides these changes are a prerequisite for such `set` function. Thanks for digguing!

AltGr:
Thanks a lot!

Issue #4100, opened 2020-02-25
`opam-format` failing due to warn-error on 4.10.0-rc2


mgree:
In https://travis-ci.com/mgree/opam-repository/jobs/291122531, I'm getting a failure from `opam-format` because of deprecation warnings and the build sets `-warn-error`.

It's slightly tricky to get the output of `opam config report`, unfortunately, because all of this is happening within CI.

rjbou:
I just tested on `4.10.0` & `4.10.0-rc2` switches, `opam-format` build prints well those warnings, but don't error on them. There is the `-warn-error A-3`, but in that case, it should have errored multiple times on the module deprecation...

In any case, this module deprecation will be removed.
Thanks for the report

rjbou:
Fixed by #4115

PR #4115, opened 2020-03-24
Makefile: add --profile=$(DUNE_PROFILE) to remaning invocations of dune

hannesm:
observed in https://github.com/ocaml/ocaml-ci-scripts/pull/258 that opam-format on 4.10 fails to install since warnings are treat as errors (see CI log at https://travis-ci.org/github/ocaml/ocaml-ci-scripts/jobs/666251349?utm_medium=notification&utm_source=github_status).

maybe this should be backported to the 2.0 branch and a 2.0.7 being released?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,11 +100,11 @@ ifneq ($(LIBINSTALL_DIR),)
 endif
 
 opam-devel.install: $(DUNE_DEP)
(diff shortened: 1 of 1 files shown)


rjbou:
Thanks! Yes it should backported. 
Fixes #4100

PR #4116, opened 2020-03-27
Option configuration, v2

rjbou:
This _new_ implementation mainly for UI evolution. The `set-opt` subcommand is replaced by the `option` subcommand:
* `opam config option : list modifiable options and their value (by default switch)
* `opam config option foo`: display `foo` value, available for all fields/sections, not only modifiable ones
* `opam config option foo<=|-=|+=>bar`: update `foo` value
* `opam config option foo=`: revert `foo` value

Options can be given to select scope `--global` or `--switch switch`. If not, `opam option` select by default switch, and other subcommands look at the presence of the field/section name in its internal list to determine the scope. 

`set-var` remain unchanged.

EDIT: update options

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -1195,35 +1195,41 @@ let config ?(option=false) () =
           `Ok ()
         with e -> print "read-state" "%s" (Printexc.to_string e); `Ok ())
     (* deprecated *)
-    | Some `set, [var; value] ->
-      OpamConsole.warning
-        "Subcommand set is deprecated. Use set-var switch %s %s instead."
-        var value;
-      OpamGlobalState.with_ `Lock_none @@ fun gt ->
-      OpamSwitchState.with_ `Lock_write gt @@ fun st ->
-      let _st = OpamConfigCommand.set_var_switch st var (Some value) in
-      `Ok ()
-    | Some `unset, [var] ->
-      OpamConsole.warning
-        "Subcommand set is deprecated. Use set-var switch %s instead." var;
-      OpamGlobalState.with_ `Lock_none @@ fun gt ->
-      OpamSwitchState.with_ `Lock_write gt @@ fun st ->
-      let _st = OpamConfigCommand.set_var_switch st var None in
-      `Ok ()
-    | Some `set_global, [var; value] ->
-      OpamConsole.warning
-        "Subcommand set-global is deprecated. Use set-var global %s %s instead."
-        var value;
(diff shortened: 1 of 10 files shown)


src/client/opamCommands.ml
```diff
@@ -851,14 +851,16 @@ let config ?(setopt=false) () =
...
      perform any variable expansion.";
-    "set-opt", `set_opt, ["[switch|global]"; "FIELD[(+=|-=|=)VALUE]"],
-    "Set the given opam configuration field in the global/switch configuration \
-     file.  If $(b,op VALUE) is omitted, $(b,FIELD) is reset to its default \
-     initial configuration, as after a fresh init (use `opam init \
+    "option", `option, ["<switch|global>"; "[FIELD[(=|+=|-=)VALUE]"],
+    "Global and switch configuration options settings. If no field is given, \
+     list all configurable fields. If a only a field name is given, display \
+     its content. Otherwise, set the given field in the global/switch \
+     configuration file. If $(b,VALUE) is ommited, $(b,FIELD) is reset to its \
```
AltGr: `set or update`
`omitted`
But I would say `empty` instead, since the docstring doesn't show it as optional

src/client/opamConfigCommand.mli
```diff
@@ -32,43 +32,80 @@ val ensure_env: 'a global_state -> switch -> unit
...
 val set_var_switch:
-  rw switch_state -> string -> string option -> rw switch_state
+  rw switch_state -> string -> whole_op -> rw switch_state
 
-(** Execute a command in a subshell, after variable expansion *)
-val exec:
-  [< unlocked ] global_state ->
-  ?set_opamroot:bool -> ?set_opamswitch:bool -> inplace_path:bool ->
-  string list -> unit
+(** List switch and/or global fields/sections and their value *)
```
AltGr, line 90: Usually, you should take a `'a state` rather than an explicitely `unlocked` state. Unless you specifically don't want a write lock!
(also, I am not fond of this kind of alignment in general, although it looks nice, because of the maintenance cost: if you modify or add something with a longer name, either you let it be unaligned, and the codes starts looking clumsy, or you have to realign all the lines, which takes time and adds noise to the patch)

AltGr: (my bad, the `unlocked` type was designed as a super-type of the `rw` type, so this is fine! It's just that in other places I took the habit to use the shorter `'a` instead)

PR #4117, opened 2020-03-30
opam_init: handle log_dir in arg and as a env variable

### LINKED ISSUE #4076: "opam install" and environment variables? [OPAMLOG dropped]

```
<include here the output of `opam config report`, if applicable>
```
I'm using opam v 2.0.6 (on both Debian Stretch and the most recent Ubuntu LTS), and I notice that the environment variable OPAMLOGS doesn't seem to work?  I searched the sources and couldn't find any mention of it?  And certainly it doesn't seem to have any effect -- logs appear, but in OPAMROOT/log.

rjbou:
fixes #4076

### DIFF
--- a/src/client/opamClientConfig.ml
+++ b/src/client/opamClientConfig.ml
@@ -174,9 +174,13 @@ let opam_init ?root_dir ?strict =
 
   (* (iii) load from env and options using OpamXxxConfig.init *)
   let log_dir =
+    OpamStd.Option.map OpamFilename.Dir.to_string @@
     if log_dir = None && initialised
-    then Some OpamFilename.(Dir.to_string (OpamPath.log root))
-    else None
+       && OpamStd.Config.env_string "LOGS" = None then
+      (* fixme: in order to not revert [OPAMLOGS] value,
+         we need to check it here *)
+      Some (OpamPath.log root)
+    else log_dir
   in
   (fun () -> ()) |>
   OpamStd.Config.initk ?log_dir |>


src/client/opamClientConfig.ml
```diff
@@ -174,9 +174,15 @@ let opam_init ?root_dir ?strict =
...
-    else None
+  OpamStd.Option.map OpamFilename.Dir.to_string @@
+    match log_dir with
+    | None ->
+      (* fixme: in order to not revert [OPAMLOGS] value,
+         we need to check it here *)
+      if initialised && OpamStd.Config.env_string "LOGS" = None then
+        Some (OpamPath.log root)
+      else None
+    | some -> some
```
AltGr: If I read this correctly, a shorter version of the patch is:
```
  let log_dir =
    if log_dir = None && initialised && OpamStd.Config.env_string "LOGS" = None
    then Some OpamFilename.(Dir.to_string (OpamPath.log root))
    else log_dir
```
So there are two changes:
- The added condition `OpamStd.Config.env_string "LOGS" = None`: a hack, but that should indeed work around the problem. Maybe we could use something like `logs_dir ++ env_string "LOGS"` instead for consistency, but not sure it would be much better, so that's OK.
- the `else` case, which indeed seemed like a mistake before (but since that argument is never used in opam... we couldn't notice)

rjbou: yes, it fixes a unnoticed missing `log_dir` arg handling...

For the `OPAMLOGS` env variable,  i choose to not get its value here, because its value is already retrieved in `OpamStd.Config.initk`. Like that, there is only one place in the code where the value is really get (and not just checked if empty). To disambiguate, we can add an `Config.defined_env` that does that check.

AltGr: No, your point makes sense, I don't think the change is needed. Thanks

PR #4122, opened 2020-04-01
Update Dune to 1.11.4

MisterDA:
Update Dune to the latest 1.x, after that the build slightly process changes.
For the patch, see https://github.com/ocaml/dune/pull/3325. If the executable path contains spaces, the build process stops. The bug also happens with the current bundled version of Dune.

### DIFF
--- a/src_ext/Makefile.sources
+++ b/src_ext/Makefile.sources
@@ -48,8 +48,8 @@ MD5_result = d3162dbc501a2af65c8c71e0866541da
 
 $(call PKG_SAME,result)
 
-URL_dune-local = https://github.com/ocaml/dune/releases/download/1.10.0/dune-1.10.0.tbz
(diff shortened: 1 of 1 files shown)


MisterDA:
Ok, so the patch does not always work. Please don’t merge while we figure this out.

MisterDA:
I decided to have just one commit updating Dune to the latest 1.x release since the bug fix is not working in all cases in Dune CI. There is ongoing discussion, so on this topic I think Opam should wait and see how Dune resolves the issue.
I did not update Dune to the latest 2.x release, since the build process has slightly changed, and Dune started rejecting jbuilder files which some of Opam dependencies still use.

dra27:
Subsumed by https://github.com/ocaml/opam/pull/4257/commits/ea34d473b38f3c13fd7b7cfeb8487d219dd8d309 in #4257 (with Dune 2.6 change to follow)

PR #4129, opened 2020-04-06
Revert "Add missing Cmdliner.Manpage.escape calls"

AltGr:
This reverts commit 888351728e1e28ddd13ee4de5828e348b6a0bbfa.

This patch makes the cmdliner escapes `$(i,xxx)` appear verbatim in the
man-page. Probably the solution is to replace the file_sep used, or use
another escaping function instead. @dra27 ?


rjbou:
Added  `OpamArg.dir_sep` that escapes windows dir sep

dra27:
Sorry for lag - yes, that fix of mine was pretty dumb, yours is much better! 🙂

PR #4132, opened 2020-04-08
Fix unused module

rjbou:
Revert adding `dune-profile`, it errors opam package install with opam
Fix the underlying errors and warnings
related #4100


hannesm:
oops, sorry for the bad PR, and thanks for the fix :)

PR #4138, opened 2020-04-10
(re) enable `var` and `option` switch load optimisations

rjbou:


src/client/opamConfigCommand.ml
```diff
@@ -518,26 +521,38 @@ let switch_allowed_fields, switch_allowed_sections =
...
   }
 
-let set_opt_switch_t ?inner st field value =
-  let switch_config =
-    set_opt ?inner field value (confset_switch st)
-  in
-  { st with switch_config }
-
-let set_opt_switch = set_opt_switch_t ~inner:false
+let with_switch lock gt st k =
```
AltGr, line 536: I would add a type annotation here to clarify the lock-kind of the switch state (might be necessary since we use phantom types, e.g. to ensure the consistency of the required lock and provided st). It would also clarify that the `st` argument is actually an `st_opt`.
```suggestion
let with_switch:
  'a lock -> 'b global_state -> 'a switch_state option
  -> (switch -> OpamFile.Switch_config.t -> 'c) -> 'c =
  fun lock gt st k ->
```

PR #4141, opened 2020-04-16
depext: enable command/output display only from verbose level 3

rjbou:


src/state/opamSysInteract.ml
```diff
@@ -376,7 +379,7 @@ let sudo_run_command cmd =
       ) else get_cmd cmd
     | _ -> get_cmd cmd
   in
-  match run_command_exit_code ~allow_stdin:true ~verbose:true cmd args with
+  match run_command_exit_code ~allow_stdin:true cmd args with
```
AltGr, line 382: Err, this needs to remain verbose (the user might be prompted for confirmation)

PR #4142, opened 2020-04-16
appveyor: add Visual Studio 2019 image

MisterDA:


dra27:
What's the motivation behind this change?

MisterDA:
Modernity, and it’s a good way to find bugs.

MisterDA:
> VS2017 instead of VS2019

Don’t you mean VS2019 instead of VS2017?
Sure, it would be better to use only one version of VS.

> Do you have an actual instance of a bug affecting VS 2019 or VS 2017 specifically?

Sadly some builds failed with a HTTP error, but https://ci.appveyor.com/project/AltGr/opam/builds/32224187/job/64mbt0qi7s2vaxam failed.
https://github.com/AltGr/ocaml-mccs/pull/29 was found with VS2019. It has been fixed in mccs upstream, and patched in Opam. It’s weird that this build failed, because it reminds me of this bug.

dra27:
Yes, I did mean that other way around! An http failure has nothing to do with the build image.

I certainly agree with switching to VS2019 testing only, yes!

MisterDA:
My goal here was to compare the two builds, show if one was exhibiting bugs that the other does not. I’d be completely fine with only VS2019, and I’ll change the commit when we’re sure that using both compilers 

> An http failure has nothing to do with the build image.

Yes but I don’t have the rights to re-launch the test. Perhaps I should have my own instance of AppVeyor builds for my Opam fork. I rebased against master to trigger the rebuild.

The fac that [Image: Visual Studio 2019; Environment: OCAML_PORT=msvc, DEP_MODE=lib-pkg, GIT_FOR_WINDOWS=1](https://ci.appveyor.com/project/AltGr/opam/builds/32250144/job/ouoe2va0q45oasow) fails puzzles me because I thought I had fixed this bug with https://github.com/ocaml/opam/pull/4109.

dra27:
Your goal would have been clearer with text in the PR 🙂 If you're testing something which doesn't require review, personal AppVeyor/Travis instances are useful, yes - I've [blogged about this for OCaml](http://www.dra27.uk/blog/platform/2017/09/27/ocaml-github-ci.html) in the past.

Your patch in #4109 is in the wrong place (sorry, in the Corona-pandemonium I missed that PR) - it should be in `mccs.common`.

MisterDA:
> Your goal would have been clearer with text in the PR :slightly_smiling_face: 

You’re right, sorry about that. Changed the commit to VS2019 only.

The build times are atrocious! I think that builds in the CI are not using parallelization: AppVeyor offers two cores, which Dune uses automatically, but the compilation OCaml and Opam can be parallelized with `make -j$(nproc)` if that’s not already the case.

> I've blogged about this for OCaml 

Thanks for the blog post! It’s clear and useful.

> it should be in mccs.common.

Oh, missed that. Weird :/ I opened a new PR moving the patch file.

dra27:
`make -j` doesn't work reliably for Windows OCaml prior to 4.09.0. It's hurting in this PR because you're changing the build image, but someone did some prior work on that to make sure it caches the compilers, so it only hurts when the cold compiler is moved forwards, or the AppVeyor is changed.

avsm:
This seems fine to merge now; removing labels. Just pinging @dra27 to confirm.

dra27:
I've discovered as part of #4338 that you can select a different image for each part of the matrix (see https://github.com/ocaml/opam/pull/4338/files#diff-180360612c6b8c4ed830919bbb4dd459R17) - we could therefore switch the default image to VS2019 but keep testing msvc32 on the VS2017 image (the other PR is using the VS 2019 because AppVeyor apparently some images and not others...!)

MisterDA:
> the other PR is using the VS 2019 because AppVeyor apparently some images and not others...!

I think there’s a verb missing here :wink: 

> switch the default image to VS2019 but keep testing msvc32 on the VS2017 image

we need to test msvc32 on VS2017 and VS2019, right?

rjbou:
ping @dra27

dra27:
Yes, sorry! All good - and merged.

PR #4144, opened 2020-04-18
depext: fail hard for BSD distributions that are not supported

hannesm:
earlier code failed soft by returning the empty set

the same function ends with the following pattern:
```OCaml
    | family ->
      Printf.ksprintf failwith
        "External dependency handling not supported for OS family '%s'."
        family
```

this PR unifies the behaviour for unsupported BSD distributions.

### DIFF
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -129,7 +129,10 @@ let packages_status packages =
             |> OpamSysPkg.Set.of_list
           in
           packages %% installed
-        | _ -> OpamSysPkg.Set.empty
+        | distribution ->
+          Printf.ksprintf failwith
+            "External dependency handling not supported for BSD distribution '%s'."
+            distribution
       in
       installed, packages -- installed, OpamSysPkg.Set.empty
     | "debian" ->


hannesm:
hmm, OTOH it seems like `netbsd` is actually being used with `depext` in opam-repository (conf-autoconf and others) -- now I'm no longer sure what the right thing to do is here..

src/state/opamSysInteract.ml
```diff
@@ -129,7 +129,10 @@ let packages_status packages =
             |> OpamSysPkg.Set.of_list
           in
           packages %% installed
-        | _ -> OpamSysPkg.Set.empty
+        | distribution ->
+          Printf.ksprintf failwith
+            "External dependency handling not supported for BSD distribution '%s'."
+            distribution
```
rjbou, line 135: It would better to have the global fail back, by restraining the `bds` pattern
```ocaml
    | "bsd" when spv OpamSysPoll.os_distribution = "freebsd" ->
      let installed =
        let installed =
          run_query_command "pkg" ["query"; "%n"]
          |> List.map OpamSysPkg.of_string
          |> OpamSysPkg.Set.of_list
        in
        packages %% installed
      in
      installed, packages -- installed, OpamSysPkg.Set.empty
    | "bsd" when spv OpamSysPoll.os_distribution = "openbsd" ->
      let installed =
        let installed =
          run_query_command "pkg_info" ["-mqP"]
          |> List.map OpamSysPkg.of_string
          |> OpamSysPkg.Set.of_list
        in
        packages %% installed
      in
      installed, packages -- installed, OpamSysPkg.Set.empty
```

rjbou:
Thanks for testing & the PR!
This makes me think that it would be better to have a name resolution that handle all these cases instead of this string pattern matching...

hannesm:
> name resolution that handle all these cases

yes, I still don't entirely understand the different `opamSysInteract` functions, but would expect them to handle the same variants. it'd be easier to check if you'd use (maybe polymorphic to have set inclusion) variants for os-family + distribution.

rjbou:
Yes, that's what I have in mind

rjbou:
I'll do that on another PR, for this one, just the code comment, and it's lgtm

rjbou:
closing this one, superseded by #4152

PR #4147, opened 2020-04-20
Remove dependency on `state` value for opam2nix use cases

### LINKED ISSUE #4030: Feedback: places where opam2nix implementation required reimplementing parts of opam
@avsm [invited me](https://discuss.ocaml.org/t/ann-opam2nix-v1/4741) to report on things which were difficult in [opam2nix](https://github.com/timbertson/opam2nix) (which uses opam-lib to automate building nix derivations for opam packages)

Broadly, the main thing I found I had to reimplement was anything which depended on opam switch state. I was able to build my own package universes, variable resolvers etc without too much trouble, but the state is very tied to opam's on-disk state, which is unsurprisingly ill-fitted to a system like nix which is stateless and where every package lives in its own prefix.

I don't think it'd be useful to make `state` generic enough to work for `opam2nix`, but perhaps there are ways to require a concrete state instance in less places.

For examples, here's some code in opam2nix which was largely copied from opam itself but couldn't be reused due to the use of state:

- [OpamSwitchState.get_conflicts](https://github.com/ocaml/opam/blob/c17c5cb2c1aa49a8e2026739714234039889bc14/src/state/opamSwitchState.ml#L479-L524) ([opam2nix version](https://github.com/timbertson/opam2nix/blob/1ffd0231c972bef5d741f178072593bee98c62da/src/select.ml#L83-L122)). This one was quite sneaky, because nothing forced me to implement it. I originally didn't implement this, and only found out about this missing code when diagnosing solves which picked conflicting packages. It's also a good example of something which could be decoupled from the switch state, my implementation only requires `lookup_var nv`, which in your case is `OpamPackageVar.resolve_switch ~package:nv st`.

- [OpamAction.prepare_package_build](https://github.com/ocaml/opam/blob/c17c5cb2c1aa49a8e2026739714234039889bc14/src/client/opamAction.ml#L243-L337) ([opam2nix version](https://github.com/timbertson/opam2nix/blob/1ffd0231c972bef5d741f178072593bee98c62da/src/invoke.ml#L200-L256)). Again `st` seems to be used just to resolve variables and opam files.

- [OpamPath.Switch file paths](https://github.com/ocaml/opam/blob/c35eb977d19457d3f13b31c1aa60de4fb210865d/src/format/opamPath.ml#L77-L139) ([opam2nix version](https://github.com/timbertson/opam2nix/blob/1ffd0231c972bef5d741f178072593bee98c62da/src/vars.ml#L40-L61), although fairly different). This one's not a big deal and might be too hard to bother detaching from state, it was mostly just reading the docs to replicate but would be neat if it could be shared.

And.. that's it actually, most of the rest of the code is necessarily different and makes use of opam internals where appropriate, so thanks :)

timbertson:
Fixes #4030


src/state/opamSwitchState.mli
```diff
@@ -54,6 +54,13 @@ val compute_available_packages:
   pinned:package_set -> opams:OpamFile.OPAM.t package_map ->
   package_set
 
+(** Raw function to compute the conflicts for all packages, given
+    the set of available packages and the corresponding opam files.
+    the `u_conflicts` field when building a universe manually. *)
```
avsm, line 59: Is there a missing line in this ocamldoc comment?  `the u_conflicts field <...>` seems to be incomplete.

src/state/opamSwitchState.ml
```diff
@@ -476,7 +476,7 @@ let conflicts_with st subset =
 let remove_conflicts st subset pkgs =
   pkgs -- conflicts_with st subset pkgs
 
-let get_conflicts st packages opams_map =
+let get_conflicts package_var packages opams_map =
```
rjbou, line 479: Usually, _raw_ functions are named `fname_t` or `fname_raw` (see e.g. in `OpamClient`).It would be better to call this function `get_conflicts_t` (and add it in the mli) and the new one `get_conflicts`.

rjbou: Same as `prepare_package_build` for the `package_var` name.

rjbou:
The context free switch paths is not only advantageous for `opam2nix`, but also for `opam-installer`:) cf. #4237

src/state/opamSwitchState.mli
timbertson: Not sure where that went, fixed thanks.

timbertson:
Thanks for the feedback, I've pushed updates to address them (as well as merge with master). The CI failures don't look related to my changes.

rjbou:
Yes, CI was broken on master, but it's ok (all green on my launches). Thanks!

PR #4152, opened 2020-04-22
depext: update

rjbou:
* Refactor `package_status`
* Add an internal type to handle unsupported os families, and add consistency
* Fix centos/rhel install commands
* Fix `config option` in error message
* Add macports commands

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -6,3 +6,5 @@ note.
 
 ## Depext
   * Fix performance issue of depext under Docker/debian [#4165 @AltGr]
+  * Refactor package status [#4152 @rjbou]
+  * Add Macport support [#4152 @rjbou]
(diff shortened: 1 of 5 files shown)


hannesm:
this is nice and superseeds #4144 -- but I wonder whether the variants introduced here and `OpamStd.Sys.os` should be unified?

rjbou:
It can't be unified with `OpamStd.Sys.os` (two different kind of information), but there is maybe something to do with `OpamSysPoll`. Here we refine the `OpamSysPoll.family` by grouping them regarding the system package manager.
In fact, all `OpamSysPoll.family/os/os_distribution/arch/...` can be rewritten to return variants, but as they are primarily used as string in variables, there isn't really that need. This _pattern matching_ constraint is really limited to `OpamSysInteract`

Issue #4154, opened 2020-04-22
Opam install script asks for user input, cannot use in docker container


lucaszanella:
https://raw.githubusercontent.com/ocaml/opam/master/shell/install.sh

As I understood by reading the script, there's no way to prevent it asking for user input. Thus I can't use this script to install opam in my docker container.

Shouldn't it have this feature?

lucaszanella:
Nevermind I updated the install script: https://github.com/ocaml/opam/pull/4155

rjbou:
`install.sh` is not intended to be used in scripts. It is meant to be interactive. Plus local install and a docker install are quite different. cf also #4136

PR #4155, opened 2020-04-22
adds docker support

lucaszanella:
I added an option for no interaction and also one for disabling sandbox after opam is installed
```
/install.sh --no-interaction --disable-sandboxing
```

Dockerfile example:

https://gist.github.com/lucaszanella/a3d9aa0db5066b158c0a5ebd2206209c


rjbou:
As mentioned in #4154, install script is not meant to initiate opam in scripts. Thanks for the pr

PR #4165, opened 2020-04-29
Fix performance issue of depext under Docker/debian

AltGr:


kit-ty-kate:
I did a quick test on my laptop for a start. Here what is the difference I get when I time both commands:
```
apt-cache search .*  14.91s user 0.03s system 99% cpu 14.945 total
apt-cache search .* --names-only  13.19s user 0.02s system 99% cpu 13.212 total
```

PR #4168, opened 2020-04-30
depexts: Make package system behaviour consistent accross platforms

kit-ty-kate:
The new depext system in opam 2.1 has several issues:
* `opam install -y` does not mean `apt-get` will be called with `-yy` anymore, which breaks every CI.
* It's behaviour regarding to whether or not package systems were to ask the user before installing packages has changed from opam-depext 2.0 (see https://github.com/ocaml/opam-depext/blob/2.0/depext.ml#L129)
* This behaviour is not consistent accross platfrom. Meaning some platforms will ask an input from their users by default, and others don't and install anything without any acknowledgement from the user.

What I've done here is the following:
* By default, always ask for user to interactively accept the installation. There are some differences from before: on Gentoo and Alpine, the user was never asked for input on opam-depext2.0.
* If `-y, --yes or OPAMYES=1` is given, never ask the user for any input. There is a minor difference from before on Archlinux (for sure) and BSDs (not quite sure because I don't have any BSD machine)
* The two exceptions are `homebrew` and `macports` which don't have any way of asking the user before installing the packages (at least that I could fine, I don't own any Mac).

Here are the man-pages for all the package systems by order in the code:
* Homebrew: https://docs.brew.sh/Manpage#install-options-formula
* Macports: https://guide.macports.org/chunked/using.html (I couldn't find better docs online)
* Apt-get: https://linux.die.net/man/8/apt-get
* Yum: https://linux.die.net/man/8/yum
* Pkg: https://www.freebsd.org/cgi/man.cgi?query=pkg-install
* Pkg_add: https://man.openbsd.org/pkg_add
* Pacman: https://www.archlinux.org/pacman/pacman.8.html
* Emerge: https://dev.gentoo.org/~zmedico/portage/doc/man/emerge.1.html
* Apk: https://git.alpinelinux.org/apk-tools/tree/doc/apk.8.scd
* Zypper: https://en.opensuse.org/SDB:Zypper_manual_%28plain%29
* https://man.openbsd.org/pkg_add

Can someone with a Mac confirm there is no way of asking for user input before installation a package? cc @mseri 

*Side-note: this PR seems to require https://github.com/ocaml/opam/pull/4152 to be rebased afterwards* cc @rjbou

### DIFF
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -240,6 +240,7 @@ let help_sections = [
        report` to know the current setting. See also option --criteria";
   `P "$(i,OPAMCUDFFILE file) save the cudf graph to \
       $(i,file)-actions-explicit.dot";
+  `P "$(i,OPAMDEPEXTYES) launch system package managers in non-interactive mode";
   `P "$(i,OPAMCURL) can be used to select a given 'curl' program. See \
       $(i,OPAMFETCH) for more options.";
   `P "$(i,OPAMDEBUG) see options `--debug' and `--debug-level'.";
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -474,7 +474,7 @@ let install_packages_commands_t sys_packages =
   | Macports ->
     ["port", "install"::packages], (* NOTE: Does not have any interactive mode *)
     None
-  | Openbsd -> ["pkg_add", yes ~no:["-I"] ["-i"] packages], None
+  | Openbsd -> ["pkg_add", yes ~no:["-i"] ["-I"] packages], None
   | Suse -> ["zypper", ("install"::yes ["--non-interactive"] packages)], None
 
 let install_packages_commands sys_packages =
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -440,7 +440,8 @@ let packages_status packages =
 
 let install_packages_commands_t sys_packages =
   let yes ?(no=[]) yes r =
-    if OpamCoreConfig.(!r.answer) = Some true then yes @ r else no @ r
+    if OpamStd.Config.env_bool "DEPEXTYES" = Some true then
+      yes @ r else no @ r
   in
   let packages =
     List.map OpamSysPkg.to_string (OpamSysPkg.Set.elements sys_packages)
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -439,14 +439,14 @@ let packages_status packages =
 (* Install *)
 
 let install_packages_commands_t sys_packages =
-  let yes opt r =
-    if OpamCoreConfig.(!r.answer) = Some true then opt @ r else r
+  let yes ?(no=[]) yes r =
+    if OpamCoreConfig.(!r.answer) = Some true then yes @ r else no @ r
   in
   let packages =
     List.map OpamSysPkg.to_string (OpamSysPkg.Set.elements sys_packages)
   in
   match family () with
-  | Alpine -> ["apk", "add"::packages], None
+  | Alpine -> ["apk", "add"::yes ~no:["-i"] [] packages], None
   | Arch -> ["pacman", "-S"::yes ["--noconfirm"] packages], None
   | Centos ->
     (* TODO: check if they all declare "rhel" as primary family *)
@@ -466,13 +466,16 @@ let install_packages_commands_t sys_packages =
                  |> OpamStd.String.Set.elements);
        "rpm", "-q"::"--whatprovides"::packages], None
   | Debian -> ["apt-get", "install"::yes ["-qq"; "-yy"] packages], None
-  | Freebsd -> ["pkg", "install"::packages], None
-  | Gentoo -> ["emerge", packages], None
+  | Freebsd -> ["pkg", "install"::yes ["-y"] packages], None
+  | Gentoo -> ["emerge", yes ~no:["-a"] [] packages], None
   | Homebrew ->
-    ["brew", "install"::packages],
+    ["brew", "install"::packages], (* NOTE: Does not have any interactive mode *)
     Some (["HOMEBREW_NO_AUTO_UPDATE","yes"])
+  | Macports ->
+    ["port", "install"::packages], (* NOTE: Does not have any interactive mode *)
+    None
+  | Openbsd -> ["pkg_add", yes ~no:["-I"] ["-i"] packages], None
   | Suse -> ["zypper", ("install"::yes ["--non-interactive"] packages)], None
-  | Openbsd -> ["pkg_add", packages], None
 
 let install_packages_commands sys_packages =
   fst (install_packages_commands_t sys_packages)
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -439,12 +439,15 @@ let packages_status packages =
 (* Install *)
 
 let install_packages_commands_t sys_packages =
+  let yes opt r =
+    if OpamCoreConfig.(!r.answer) = Some true then opt @ r else r
+  in
   let packages =
     List.map OpamSysPkg.to_string (OpamSysPkg.Set.elements sys_packages)
   in
   match family () with
   | Alpine -> ["apk", "add"::packages], None
-  | Arch -> ["pacman", "-S"::"--noconfirm"::packages], None
+  | Arch -> ["pacman", "-S"::yes ["--noconfirm"] packages], None
   | Centos ->
     (* TODO: check if they all declare "rhel" as primary family *)
     (* When opam-packages specify the epel-release package, usually it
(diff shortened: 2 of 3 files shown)


kit-ty-kate:
The Travis failures seem unrelated.

Could someone with a FreeBSD and OpenBSD on hand tell me if my reading of the manuals were accurate? @hannesm @avsm maybe?

rjbou:
depext used that mechanism to avoid the blocking system install request, and it was removed in depext integration. What we wanted to avoid is the fact that the system install is done without the user agreement. It can happen automatically when opam-yes is propagated to system install and the user did a sudo call few commands before.
We could add an depext-yes environment variable for CI, to propagate the `yes` to system install commands.

kit-ty-kate:
I feel like having two variables `OPAMYES`/`-y`/`--yes` + a special variable/option for depext seems redundant. People who use `-y` are already ready to have potentially breaking things happening to their system. Furthermore, the opam calls sudo and show the command that is going to be run, so users will be aware of that.
If you really feel like it's something users might go through then maybe a separate option can be used but it does not change this PR significantly anyway.

avsm:
If the behaviour does diverge from opam 2.0, then there really needs to be a clear way to run both reliably with the same command line.

For example, the `CI` environment variable is largely set by most CIs now, so we could patch opam 2.0's depext to use that for yes, and opam 2.1 could also automatically go into non-interactive mode if that is set.  Dune similarly uses the `CI` environment variable to cue non-interactive behaviour, so there is precedent here.

src/state/opamSysInteract.ml
```diff
@@ -320,28 +323,28 @@ let install_packages_commands s_packages =
...
       |> OpamSysPkg.Set.elements
       |> List.map OpamSysPkg.to_string);
      "rpm"::"-q"::"--whatprovides"::packages]
   | "bsd" ->
     if spv OpamSysPoll.os_distribution = "freebsd" then
-      ["pkg"::"install"::packages]
+      ["pkg"::"install"::yes ["-y"] packages]
     else
-      ["pkg_add"::packages]
+      ["pkg_add"::yes ~no:["-I"] ["-i"] packages]
```
avsm, line 339: This is correct for openbsd.  I suspect we need to distinguish `netbsd` and `openbsd` here, since their package managers are a different implementation now.

On OpenBSD, you can also add
```
     -a       Automated package installation; do not record packages as
              installed manually.
```

which marks it as an automatically installed package for the purposes of system cleanup.

dra27: Wouldn't that make any clean-up of the packages automatically remove it?

avsm: Not on OpenBSD -- it's just used to give a cleaner output in `pkg_info` (to only show packages you've chosen to install by default, and not their dependencies).

> pkg_add differentiates between packages specified on the command line, and packages installed automatically because of inter-dependencies: the first kind will be tagged as ‘installed manually’. The -a option is used internally by the ports(7) infrastructure and dpb(1) to handle dependencies.
>
> It is also possible to tweak the ‘installed manually’ status of a package after the fact. Running pkg_add on an already installed package will tag it as ‘installed manually’, even if it was already there as a dependency of something else, and doubling the -a option will remove the ‘installed manually’ tag from installed packages.
>
> pkg_info(1) can be used to show only manually-installed packages, and pkg_delete(1) can be used to remove dependencies when they are no longer needed.

dra27: That's what I mean with the last bit. How does pkg_delete know it's no longer needed, given that ports has no knowledge of opam's dependency on it?

dra27: Or is the point that `pkg_delete` never does that automatically, the user determines it?

avsm: The user determines it. It's only ever manually done by `pkg_delete -a`

hannesm: I've no idea about what `opam depext` semantics should be in this respect, but note that FreeBSD has a similar `-A` argument which could be passed.

hannesm:
@kit-ty-kate for FreeBSD, I can confirm that your `pkg install -y <pkg>` is fine.

src/state/opamSysInteract.ml
rjbou: About marking packages as automatically installed, we thought about proposing that  on the opam package removal, like that the system package is no more marked as manually installe din its system. But it can lead to a removal of the system package on the next cleanup, and if a package that depends on is installed on another switch, it can lead to a reinstall.

rjbou:
PR updated.
Added also an `OPAMCIDEPEXTYES` env var to trigger the yes to system package manager. Another solution would be to add an depext option in config, which is enable/disabled via `opam option depext-yes=trure|false`

kit-ty-kate:
I'm not sure why this new environment variable has `CI` in it's name. It seems useful to have this feature outside of a CI as well. Could we rename this `OPAMDEPEXTYES` instead?
I also think we should have an argument option to accompany this new environment variable. Something like `--depext-yes/-Y`, maybe? Maybe this new `-Y` option could also trigger `--yes/-y/OPAMYES`?

AltGr:
Thanks a lot @kit-ty-kate for taking the time to lookup the behaviours of all these different package managers.

Here it's quite hard to strike the right balance between asking the same thing too many times (from opam, sudo, the sysPM...) or worse, too little, before running potentially destructive system commands (e.g. `aptitude` has been infamous, in the past, for having `install` commands that could wipe 80% of your system because of conflicts); having a way to consistently abstract over sysPMs is a necessary first step here.

About that balance, we already provide a few settings, but it's most important to have the right defaults. This is what we have now:

```
% opam install conf-bmake  
  - install conf-bmake 1.0

    bmake

Let opam run your package manager to install the required system packages? [Y/n] y
+ /usr/bin/sudo "apt-get" "install" "bmake"
[sudo] password for lg: 
- Reading package lists...
- Building dependency tree...
[...]
- The following NEW packages will be installed:
-   bmake
- 0 upgraded, 1 newly installed, 0 to remove and 236 not upgraded.
- Need to get 0 B/289 kB of archives.
- After this operation, 509 kB of additional disk space will be used.

- Selecting previously unselected package bmake.
(Reading database ... 60%
(Reading database ... 777135 files and directories currently installed.)
- Preparing to unpack .../bmake_20160220-2+b1_amd64.deb ...
- Unpacking bmake (20160220-2+b1) ...
- Setting up bmake (20160220-2+b1) ...
- Processing triggers for man-db (2.8.5-2) ...

-> installed conf-bmake.1.0
Done.

- Preparing to unpack .../bmake_20160220-2+b1_amd64.deb ...
- Unpacking bmake (20160220-2+b1) ...
- Setting up bmake (20160220-2+b1) ...
- Processing triggers for man-db (2.8.5-2) ...

-> installed conf-bmake.1.0
Done.
```
Here we are possibly asked for confirmation:
- by opam, to confirm the actions (not here, because they match the request bit by bit)
- by opam, to allow sudo access (`Let opam run your package manager to install the required system packages?`)
- once by sudo (but could be skipped if already authorised)
- by `apt-get` (skipped here, because `apt-get`, like opam, doesn't ask for confirmation of installs when no extra dependencies get pulled in)

If no terminal is bound, we get the following instead:
```
% opam install conf-bmake <&-
  - install conf-bmake 1.0

    bmake

Let opam run your package manager to install the required system packages? [Y/n] n
[NOTE] Use 'opam config option global depext-run-installs=false' if you don't want to be prompted again.

    apt-get install bmake

You can retry with '--assume-depexts' to skip this check, or run 'opam config option global depext=false' to permanently disable handling of system packages altogether.
```
With `--yes`:
```
% opam install conf-bmake --yes <&-
  - install conf-bmake 1.0

    bmake

Let opam run your package manager to install the required system packages? [Y/n] y
+ /usr/bin/sudo "apt-get" "install" "bmake"
- Reading package lists...
[...]
- Unpacking bmake (20160220-2+b1) ...
- Setting up bmake (20160220-2+b1) ...

-> installed conf-bmake.1.0
Done.
```
(assuming `sudo` doesn't hang)

The important bit is that here, it works because of `apt-get`'s magic and the fact that no other dependency was needed. Nice to avoid asking the user one more time, but not a solution in general.

There is no reliable way to tell all package managers "don't ask unless you remove anything", so I believe that, in general, it is too risky to let `-yy` or similar slip through (the manpage for `apt-get --force-yes` is _very_ scary). So:
- either leave the responsibility to the sysPM
- or force it to prompt interactively, as this PR does, which seems more consistent and safer.

**I am absolutely against using `--yes` to bypass that, because I have seen at large people using it by default** (I would blame them, but that wouldn't help...)**, or worse, advising it in snippets to help others get started with opam.**

This is all fine for interactive mode though. What when we can't ask ?
In CI, it's perfectly fine to force yes everywhere — we don't even have a choice. In scripts... well it depends. The most common example I see is projects using OCaml that have a custom `make deps` target. We don't want that to allow bypassing the OS-level confirmation. I say in this case, if non-interactive, we should stick with the following failure:
```
- Do you want to continue? [Y/n] Abort.
[ERROR] System package install failed with exit code 1 at command:
            sudo apt-get install libgd-dev
```
In any case, if you don't allow passwordless sudo, this would have failed anyway earlier.

To conclude, I confirm that we should only ever allow bypassing it in CI or similiarly controlled environments. For that purpose, I don't think a command-line flag is suitable ; an internal option should be more than enough, and the proposed environment variable could do the job as well, as would the standard `CI` variable.

**I would go as far as to suggest an `opam init --ci` command — remember we need to disable sandboxing anyway for these environments ? — without even an option, command or environment variable to change the option afterwards** (bar editing the config files of course).

src/state/opamSysInteract.ml
```diff
@@ -358,12 +358,16 @@ let packages_status packages =
 (* Install *)
 
 let install_packages_commands sys_packages =
+  let yes ?(no=[]) yes r =
+    if OpamStd.Config.env_bool "CIDEPEXTYES" = Some true then
```
AltGr, line 362: A reason for not going through `OpamStateConfig` ?

src/state/opamSysInteract.ml
```diff
@@ -372,22 +376,22 @@ let install_packages_commands sys_packages =
...
+                 |> OpamStd.String.Set.elements);
        "rpm", "-q"::"--whatprovides"::packages]
-  | Debian -> ["apt-get", "install"::packages]
-  | Freebsd -> ["pkg", "install"::packages]
-  | Gentoo -> ["emerge", packages]
-  | Homebrew -> ["brew", "install"::packages]
-  | Macports -> ["port", "install"::packages]
-  | Openbsd -> ["pkg_add", packages]
-  | Suse -> ["zypper", ("install"::packages)]
+  | Debian -> ["apt-get", "install"::yes ["-qq"; "-yy"] packages]
```
AltGr, line 388: Another option here, which is less aggressive, is `DEBIAN_FRONTEND=noninteractive`
(note that this could also be manually set by the caller as a workaround).

src/client/opamArg.ml
```diff
@@ -240,6 +240,7 @@ let help_sections = [
        report` to know the current setting. See also option --criteria";
   `P "$(i,OPAMCUDFFILE file) save the cudf graph to \
       $(i,file)-actions-explicit.dot";
+  `P "$(i,OPAMCIDEPEXTYES) launch system package manager with `--yes` option.";
```
AltGr, line 243:
```suggestion
  `P "$(i,OPAMCIDEPEXTYES) launch system package managers in non-interactive mode";
```

src/state/opamSysInteract.ml
rjbou: When I did that, I had in mind to add a new _CI state_ that aggregates all CI options/environment variables. So, I didn't had it in `OpamStateConfig` to minimise the changes.

src/state/opamSysInteract.ml
rjbou: In #4200, I add the handling of giving environment variables to install commands.

AltGr:
Rebased

rjbou:
Updated the env var to `OPAMDEPEXTYES`.

rjbou:
Thanks!

PR #4180, opened 2020-05-05
Add master changelog and github action check

rjbou:
The github action checks that the file `master_changes.md` has been updated in the PR.
Some files are concerned by this check (travis, appveyor, etc.)

### DIFF
--- a/.github/scripts/changelog_checker
+++ b/.github/scripts/changelog_checker
@@ -1,24 +1,28 @@
-#!/bin/sh
+# This script check that the current master changelog has been updated by PR,
+# ignoring some internal files.
+# It is used byt the changelog_check github action.
+
+# (c) Copyright Raja Boujbel OCamlPro 2020
 
 set -ue
 
-UNTRACKED="
-appveyor_build.cmd
-appveyor.patch
-appveyor_test.sh
-appveyor.yml
-AUTHORS
-CONTRIBUTING.md
+IGNORE="
 .gitattributes
 .github
 .gitignore
-ISSUE_TEMPLATE.md
 .ocamlinit
 .ocp-indent
 .ocplint
-README.md
 .travis-ci.sh
 .travis.yml
+AUTHORS
+CONTRIBUTING.md
+ISSUE_TEMPLATE.md
+README.md
+appveyor.patch
+appveyor.yml
+appveyor_build.cmd
+appveyor_test.sh
 master_changes.md
 "
 changelog=master_changes.md
@@ -28,17 +32,15 @@ git fetch origin $GITHUB_BASE_REF --depth=1 --quiet
 echo "> base commit"
 git show origin/$GITHUB_BASE_REF --format=oneline -s
 
-git diff origin/$GITHUB_BASE_REF --name-status --diff-filter=AM > $diffile
+git diff origin/$GITHUB_BASE_REF --name-only --diff-filter=AMRCX > $diffile
 updated=0
-if [ -z "`grep -s $changelog $diffile`" ] ; then
-  updated=1
-fi
+grep -sq $changelog $diffile || updated=1
 
 echo "> all changes"
 cat $diffile
 
-for untrack in $UNTRACKED ; do
-  sed -i "/$untrack/d" $diffile
+for ign in $IGNORE ; do
+  sed -i "/^$ign/d" $diffile
 done
 
 echo "> kept changes"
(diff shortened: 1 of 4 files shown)


.github/scripts/changelog_checker
```diff
@@ -0,0 +1,58 @@
...
+changelog=master_changes.md
+diffile=/tmp/diff
+
+git fetch origin $GITHUB_BASE_REF --depth=1 --quiet
+echo "> base commit"
+git show origin/$GITHUB_BASE_REF --format=oneline -s
+
+git diff origin/$GITHUB_BASE_REF --name-status --diff-filter=AM > $diffile
+updated=0
+if [ -z "`grep -s $changelog $diffile`" ] ; then
```
AltGr, line 33: `if ! grep -sq $changelog $diffile; then ...`
Or even just `grep -sq $changelog $diffile || updated=1` actually, but that's less readavle

.github/workflows/changelog_check.yml
```diff
@@ -0,0 +1,12 @@
...
+on:
+  pull_request:
+    branches: master
+
+jobs:
+  diff:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - run: bash .github/scripts/changelog_checker
```
AltGr, line 12: The script says `#!/bin/sh`, why is `bash` needed here ?
(Ok, we probably don't care ;) )

AltGr: Ah my bad, this is specific for GH runners with a limited choice, not any Docker image like GL ones, so suggesting Alpine makes no sense here

.github/scripts/changelog_checker
```diff
@@ -0,0 +1,58 @@
+#!/bin/sh
```
dra27, line 1:
```suggestion
#!/usr/bin/env bash
```
(if it's going to run with bash, it would be better for that to be reflected in the header!)

.github/scripts/changelog_checker
```diff
@@ -0,0 +1,58 @@
...
+
+set -ue
+
+UNTRACKED="
```
dra27, line 5: It's slightly confusing terminology, given that "untracked" has special meaning for Git. Perhaps "INFRASTRUCTURE", or "INTERNAL"?

.github/scripts/changelog_checker
dra27, line 33:
```suggestion
if [[ -z $(grep -s $changelog $diffile) ]] ; then
```
is a bit bashier, given that it's using bash!

.github/scripts/changelog_checker
```diff
@@ -0,0 +1,58 @@
...
+updated=0
+if [ -z "`grep -s $changelog $diffile`" ] ; then
+  updated=1
+fi
+
+echo "> all changes"
+cat $diffile
+
+for untrack in $UNTRACKED ; do
+  sed -i "/$untrack/d" $diffile
```
dra27, line 41: This is risky - all those patterns are treated as though *$untrack* - shouldn't the patterns at least match at the start only?

dra27: ("risky" in terms of the utter evil of escaping the change logging checking action!)

.github/scripts/changelog_checker
```diff
@@ -0,0 +1,58 @@
...
+done
+
+echo "> kept changes"
+cat $diffile
+
+num_changes=`wc -l $diffile | cut -f 1 -d ' '`
+
+if [ $num_changes -ne 0 ] ; then
+  if [ $updated -eq 0 ] ; then
+    echo "[32mChangelog updated[0m"
```
dra27, line 51:
```suggestion
    echo "[32mChangelog updated[0m"
    echo -e "\033[32mChangelog updated\033[m"
```

.github/scripts/changelog_checker
```diff
@@ -0,0 +1,58 @@
...
+  else
+    echo "[31mPlease update changelog in master_changes.md[0m"
```
dra27, line 53:
```suggestion
    echo -e "\033[31mPlease update changelog in master_changes.md\033[m"
```

.github/scripts/changelog_checker
```diff
@@ -0,0 +1,58 @@
...
+    exit 1
+  fi
+else
+  echo "[33mCommited files not concerned by changelog[0m"
```
dra27, line 57:
```suggestion
  echo -e "\033[33mCommitted files not concerned by changelog\033[m"
```
(note spelling correction too)

.github/workflows/changelog_check.yml
rjbou: No, just a habit to begin my scripts with that shebang

.github/scripts/changelog_checker
rjbou: Thanks!

.github/scripts/changelog_checker
rjbou: Thanks!

rjbou:
`.gitattributes`  updated and no, the script is not executable, that's why I launch it via `bash`.

I dissociate the current changelog from the release one because they don't have the same purpose (neither the same format). The `master_changes.md` files contains all updates, so it contains _almost_ all PRs (unless it's infra/CI ones). It meant to be a draft material to construct the release changelog and the release note.
The current changelog check/update is meant to evolve, to facilitate user pr process, e.g. for small prs, when PR title & changelog update are exactly the same, using a comment directive, an action to update automatically the PR.
I'll open an issue to discuss about it.

PR #4184, opened 2020-05-06
Remove m4 from the list of recommended tools

kit-ty-kate:
With depext now builtin part of opam in opam 2.1 and `ocamlfind` (which uses `m4`) being less and less used, I'm not sure it feels necessary to have a warning recommending to install this package anymore:
```
[WARNING] Recommended dependencies -- most packages rely on these:
  - m4
```


avsm:
Good idea; needs a changes entry

PR #4191, opened 2020-05-11
Store depext exceptions in switch with `--assume-depexts` and `--no-depexts`

### LINKED ISSUE #4177: depext-bypass not always updated as expected
It seems that `opam install foo --no-depexts` doesn't properly store the exceptions:
```
 $ opam install conf-pkg-config --no-depexts
  - install conf-pkg-config 1.1

-> installed conf-pkg-config.1.1
Done.
# Run eval $(opam env) to update the current shell environment
$ opam upgrade --show
[WARNING] Opam package conf-pkg-config.1.1 depends on the following system package that can no longer be found: pkgconfig
```

rjbou:
& add `depext-bypass`  as a modifiable field in switch config
Fix #4177

### DIFF
--- a/src/client/opamSolution.ml
+++ b/src/client/opamSolution.ml
@@ -1139,6 +1139,11 @@ let apply ?ask t ~requested ?add_roots ?(assume_built=false) ?force_remove
           let depext_bypass =
             OpamSysPkg.Set.Op.(t.switch_config.depext_bypass ++ sys_packages)
           in
+          OpamConsole.note "Add system packages %s to switch 'depext-bypass' field"
+            (OpamStd.Format.pretty_list
+               (List.map (fun s ->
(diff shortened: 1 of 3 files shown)


src/client/opamSolution.ml
```diff
@@ -1133,6 +1133,20 @@ let apply ?ask t ~requested ?add_roots ?(assume_built=false) ?force_remove
           action_graph
       in
       let success = match r with | OK _ -> true | _ -> false in
+      let t =
+        if success && (OpamClientConfig.(!r.assume_depexts) ||
+                       OpamStateConfig.(!r.no_depexts)) then
```
rjbou, line 1138: We don't need to check `no_depexts` as `assume depexts` is defined as `assume_depexts || no_depexts`. But to remove it, we need to be sure that this implication remain always true.

rjbou:
Commits ported in #4194

PR #4200, opened 2020-05-13
depext: fix arch/brew, use regexp, environment variable handling, etc.

### LINKED ISSUE #4199: [2.1.0~alpha] arch depexts query commands incorrect

```
 opam config report
# opam config report
# opam-version         2.1.0~alpha
# self-upgrade         no
# system               arch=x86_64 os=linux os-distribution=arch os-version=rolling
# install-criteria     -removed,-count[version-lag,request],-count[version-lag,changed],-changed
# upgrade-criteria     -removed,-count[version-lag,solution],-new
# jobs                 4
# repositories         1 (http), 3 (version-controlled) (default repo at 12d69d0d)
# pinned               8 (git)
# current-switch       /home/dave/dev/mydir
```

Hi there :-), loving opam 2.1 so far!

I've noticed when modifying my switch state (installing a package etc) any depexts always require reinstall - if the state is not actually modified I get this warning:

```
$ opam install dune
[WARNING] Opam package conf-m4.1 depends on the following system package that can no longer be found: m4
[NOTE] Package dune is already installed (current version is 1.11.4).
```

I think the cause of this is the pacman commands used to see installed/available packages are incorrect. Currently `pacman -Q` is used for installed packages, and `pacman -Qe` is used for available packages (from https://github.com/ocaml/opam/blob/8f3cdddd2f62713ea1e8d4c376d7a22930ab6adb/src/state/opamSysInteract.ml#L146-L147). However I think this is actually looking at all installed packages (`pacman -Q`) and packages that are installed explicitly (i.e. not as dependencies of other packages) (`pacman -Qe`). e.g.:

```
$ pacman -Ss m4
core/m4 1.4.18-3 (base-devel) [installed]
    The GNU macro processor
... <other packages>

$ pacman -Q | grep m4
m4 1.4.18-3

$ pacman -Qe | grep m4

```

I believe the commands should be:
- `pacman -Qe` to get the installed package list
- `pacman -S --list` to get available packages.

If it helps at all, the output of `pacman -S --list` looks like this, where the first word 'core' is the name of the package repo:

```
$ pacman -S --list | grep m4 -C 3
core lvm2 2.02.187-1 [installed]
core lz4 1:1.9.2-2 [installed]
core lzo 2.10-3 [installed]
core m4 1.4.18-3 [installed]
core make 4.3-3 [installed]
core man-db 2.9.1-2 [installed]
core man-pages 5.06-2 [installed]
... <other matches>
```

or alternatively there's 
```
$ pacman -Sq --list | grep m4 -C 3
lvm2
lz4
lzo
m4
make
man-db
man-pages
... <other matches>
```

Thanks!

rjbou:
* arch: use `pacman -Ss <packages regexp>` to have only one exec call. fix #4199
* regexp: transform list string manipulation into regexp to treat sys package manager outputs
* add some example outputs
* fix homebrew package name check, allow `package` and `package@version`
* add environment variabe handling for `OpamSysInteract` commands (addition and overriding)
* force `LC_ALL=C` for query commands
* fix dry run for install commands

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -32,9 +32,15 @@ Possibly scripts breaking changes are prefixed with ✘
 
 
 ## Depext
+  * Fix arch query [#4200 @rjbou]
   * Fix performance issue of depext under Docker/debian [#4165 @AltGr]
-  * Refactor package status [#4152 @rjbou]
+  * Refactor package status [#4152 #4200 @rjbou]
+  * Add environment variables handling [#4200 @rjbou]
   * Add Macport support [#4152 @rjbou]
+  * Homebrew: add no auto update env var for install, accept `pkgname` and `pkgnam@version` on query [#4200 @rjbou]
+  * Force LC_ALL=C for query commands [#4200 @rjbou]
+  * Fix install command dryrun [#4200 @rjbou]
+
 
 
 ## Remove
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -118,7 +118,13 @@ let packages_status packages =
          (Re.replace ~all:true need_escape ~f:(fun g -> "\\"^Re.Group.get g 1))
          str_pkgs)
   in
-  let with_regexp ~re_installed ~re_pkg =
+  let with_regexp_sgl re_pkg =
+    List.fold_left (fun pkgs l ->
+        try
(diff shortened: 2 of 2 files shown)


rjbou:
It is possible to add that, yes

rjbou:
updated!

PR #4209, opened 2020-05-27
Fix path resolution with `file://` prefix

### LINKED ISSUE #4208: When pinning, local paths should be fully resolved
When pinning a local path, it looks like the path itself is not fully resolved before being registered. In particular, if it is a relative path, it is stored as is.

I ran `opam pin add foo .` while in the directory of the package I wanted to pin. Later on, I ran `opam upgrade` from my home directory: opam then started copying my entire home directory into the sources directory for the package.

I expected instead the path to be resolved to an absolute path at the time the pin was added.

rjbou:
Fixes #4208

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -19,7 +19,7 @@ New option are prefixed with ◈
   * ext: fix extlib url [#4248 @rjbou]
 
 ## Install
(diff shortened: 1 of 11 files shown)


rjbou:
Updated, file paths are resolved only when they are from cli
+ a lint error (65) to check that opam file urls are absolute

PR #4211, opened 2020-05-28
Assume built rewriting

rjbou:
Some fixes:
* handling of post dependencies (leading to an infinite look up of dependencies, or dependency unavailable)
* specific missing dep message

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -22,6 +22,7 @@ New option are prefixed with ◈
 ## Install
   * Add `_build` to rsync exclusion list [#4230 @rjbou - fix #4195]
   * Recursive opam file lookup: ignore `_build` [#4230 @rjbou]
+  * Assume-built fix & rewriting [#4211 @rjbou]
 
 ## Archives fetch
   * Copy instead of calling rsync when archives are in a local cache [#4270 @kit-ty-kate]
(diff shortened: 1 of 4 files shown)


lthls:
I've tried this PR, using the instructions on the wiki [here](https://github.com/ocaml/opam/wiki/How-to-test-an-opam-feature), with the compiler.

The setup part, in addition the setting `opam2` to use the experimental version with a distinguished root, includes checking out a compiler branch I'm interesting in testing and `cd`ing into it.
At this point the current directory is a fresh copy, not configured.

Then I ran the following commands:
```
$ opam2 sw cr test --empty
$ opam2 pin add . --inplace-build # I think using install would have worked too
# At this point, I start making changes to the compiler, recompiling, testing using the compiler's testsuite
# Then when I'm ready to test on opam:
$ make -j4 world.opt # just in case
$ opam install . --inplace-build --assume-built
```

The last command successfully ran, installing the compiler without recompiling it from scratch.

The main pain point is that the first `pin` command has to be run to configure the compiler correctly, but I didn't find an easy way to tell it to stop after configure so it has to build everything (Although I could have interrupted the process with `Ctrl+C`, I think).
Otherwise, this looks quite useful (especially on flambda2, where compiling the compiler can take quite long).

PR #4178, opened 2020-05-05
Update opam file with dune, remove `opam-%-install` target

rjbou:
`opam-%s.install` target was only used by opam files, there is no occurrence of their call in the opam source. Are they used in other projects ? @samoht @avsm @dra27

Fix #4173

PR #4229, opened 2020-06-10
Fix the opam-devel package + various fixes

kit-ty-kate:
Following the merge of https://github.com/ocaml/opam/pull/4178, the `opam-devel` package now doesn't install anything anymore (see the special `opam-devel.install` makefile rule). 1471a27 should fix that.

Otherwise the rest is more of a collection of various fixes that can be taken independently:
* 8d4ae0d is here because the `opam-%.install` rule is still used in the makefile (see the `installlib-%` rule for instance)
* 5429c09 fixes an issue that happens when an external archive is not available to download and the script tries to get it from the cache but the url does not exist anymore.
* 9f3dacf is a preventive fix in case someone changes the name of the opam binary (in the src/client/dune) it will not try to use the `opam` binary from the PATH (which, if it is not opam 2.1 already is going to hang up forever trying to get `opam lock --help=groff`
* 866fbc7 fixes an issue in the opam file as dune is not compatible with the `{build}` tag anymore (see https://github.com/ocaml/dune/issues/2147)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,6 +103,9 @@ opam-devel.install: $(DUNE_DEP)
 	$(DUNE) build $(DUNE_ARGS) -p opam opam.install
 	sed -e "s/bin:/libexec:/" opam.install > $@
 
+opam-%.install: $(DUNE_DEP)
+	$(DUNE) build $(DUNE_ARGS) -p opam-$* $@
+
 .PHONY: build-opam-installer
 build-opam-installer: $(DUNE_DEP) 
 	$(DUNE) build --profile=$(DUNE_PROFILE) $(DUNE_ARGS)$(DUNE_PROMOTE_ARG) opam-installer.install
--- a/master_changes.md
+++ b/master_changes.md
@@ -12,7 +12,7 @@ Possibly scripts breaking changes are prefixed with ✘
   * Fix atoms formula restriction with `--all` [#4221 @rjbou - fix #4218]
 
 ## Build
-  * Opam file build using dune, removal of opam-%.install makefile target [#4178 @rjbou - fix #4173]
+  * Opam file build using dune, removal of opam-%.install makefile target [#4178 @rjbou #4229 @kit-ty-kate - fix #4173]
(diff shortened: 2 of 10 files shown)


AltGr:
Rebased

doc/man/dune
```diff
@@ -1,7 +1,7 @@
 (rule
   (targets opam.1)
   (deps opam-topics.inc opam-admin-topics.inc)
-  (action (with-stdout-to %{targets} (run %{bin:opam} --help=groff))))
+  (action (with-stdout-to %{targets} (run ../../src/client/opamMain.exe --help=groff))))
```
dra27, line 4: This change breaks the mingw builds - what's it fixing elsewhere?

dra27:
Ah, looking at the description of the PR I see the answer to my question. `doc/man/dune` must use `%{bin:opam}` - the fix instead might be to a put a comment above `public_name opam` in `src/client/dune` noting that if it's changed that it needs to be changed elsewhere. To be honest, I'm not totally sure why we should be hardening the build system against external changes 🙂

doc/man/dune
kit-ty-kate: oh.. I thought dune would have dealt with that and replaced the path

kit-ty-kate:
The problem I was trying to solve arise when I was trying to remove the dummy opam.opam package/file and fix the opam-devel package without reintroducing a custom Makefile rule.
Since normal dune packages install their opam file in `%{lib}%/opam-devel/opam` (the same place opam is installed right now), I had to change the name of the binary. It seemed to worked locally (since I had opam 2.1 already installed) but I ran into an infinite loop when I tried to compile it in an environment that had opam 2.0.

But I agree this change was optional and since it fails with mingw I removed it from the PR

AltGr:
> Since normal dune packages install their opam file in `%{lib}%/opam-devel/opam` (the same place opam is installed right now)

:scream: ah I get it now...
100% dune's fault though :innocent:

rjbou:
Thanks!

PR #4234, opened 2020-06-16
build cold bootstrap with OCaml 4.10 to fix gcc10 compilation

avsm:
This is because gcc10 defaults to -fcommon which causes
compatibility carnage.  See ocaml/opam-repository#16583.

Reported by @Misterda

### DIFF
--- a/appveyor.patch
+++ b/appveyor.patch
@@ -14,8 +14,8 @@ index 1a350068..964a818c 100644
 --- a/src_ext/Makefile
 +++ b/src_ext/Makefile
 @@ -12,8 +12,8 @@ endif
- URL_ocaml = http://caml.inria.fr/pub/distrib/ocaml-4.07/ocaml-4.07.1.tar.gz
- MD5_ocaml = 0b180b273ce5cc2ac68f347c9b06d06f
+ URL_ocaml = http://caml.inria.fr/pub/distrib/ocaml-4.10/ocaml-4.10.0.tar.gz
+ MD5_ocaml = 34de3225e40d92a6a8b03f1626a01dd3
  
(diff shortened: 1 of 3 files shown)


avsm:
@dra27 instead of the appveyor patch for unreleased flexdll, can we just use your fork in src_ext/Makefile directly?

MisterDA:
I have been using flexdll master https://github.com/alainfrisch/flexdll/commit/bd636def70d941674275b2f4b6c13a34ba23f9c9 in my
experiments. I don’t think that using @dra27 flexdll fork is
good. In a great scheme of things, a patch that is needed for
Opam’bootstrap should be included upstream — and @dra27 has write
access to the repo.
It’s possible to download a tar.gz of a specific commit of a repo from GitHub
(as a matter of fact, of any git ref):
https://github.com/alainfrisch/flexdll/archive/bd636def70d941674275b2f4b6c13a34ba23f9c9.tar.gz.
I don’t see a major downside to this approach. We’re not tracking
'master' since we’re pinning a revision.
For other dependencies of Opam, it could be even better since we would
not have to carry patches that already have been merged in master.

avsm:
I'm referring to the Appveyor workaround - that should be reconciled with src_ext/Makefile. I'd prefer we didn't carry CI specific patches.

dra27:
Sorry for the lag. There will/should be a release of flexdll incorporating that patch prior to the release of OCaml 4.11 (as noted, it is now in flexdll master) which should make this moot. I agree with @MisterDA that pointing to my repo is a bad idea (especially as I may arbitrarily delete that branch when I next hoover my fork...)

PR #4235, opened 2020-06-18
Tag packages with missing depexts to the solver

AltGr:
to allow playing with criteria and seeing if we can solve the
mariadb/mysql dilemma.

With this patch, you can add a `-count[missing-depexts,solution]`
criterion (putting it before the main `version-lag` criteria would be a
bad idea though).

### DIFF
--- a/src/state/opamSwitchState.ml
+++ b/src/state/opamSwitchState.ml
@@ -858,6 +858,14 @@ let universe st
     | Some set -> set
     | None -> OpamPackage.Set.empty
   in
+  let missing_depexts =
+    OpamPackage.Map.fold (fun nv status acc ->
+        if OpamSysPkg.Set.is_empty status.OpamSysPkg.s_available
+        then acc
+        else OpamPackage.Set.add nv acc)
+      (Lazy.force st.sys_packages)
+      OpamPackage.Set.empty
+  in
   let u =
 {
   u_packages  = st.packages;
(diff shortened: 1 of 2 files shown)


rjbou:
Tested in simple case
```
$ opam show foo -f depends # bmake is installed in the system, but no llvm
"ocaml"  ("conf-bmake" | "conf-llvm")

$ ./opam install foo.0.4 --show
The following actions would be performed:
  ∗ install conf-llvm 10.0.0 [required by foo]
  ∗ install foo       0.4
===== ∗ 2 =====

    llvm-10-dev

$ ./opam install foo.0.4 --show '--criteria=-removed,-count[version-lag,request],-count[missing-depexts,solution],-count[version-lag,changed],-changed'
The following actions would be performed:
  ∗ install conf-bmake 1.0 [required by foo]
  ∗ install foo        0.4
===== ∗ 2 =====

File: src/state/opamSwitchState.ml [hunk 1]
```diff
@@ -858,6 +858,14 @@ let universe st
     | Some set -> set
     | None -> OpamPackage.Set.empty
   in
+  let missing_depexts =
+    OpamPackage.Map.fold (fun nv status acc ->
+        if OpamSysPkg.Set.is_empty status.OpamSysPkg.s_available
```
rjbou, line 863: Should'nt be if available and not found are empty?

rjbou: Not necessarily, as not found one aren't in the request (removed from available)

PR #4237, opened 2020-06-18
Switch name check

### LINKED ISSUE #4231: opam accepts switch names that create unusable switches
Hi,

Opam can create switches with some characters in their names that create unusable switches:

### colons

When a switch name contains a colon, `PATH` is extended with this switch's bin directory, but since this is a colon-separated variable (with no escape mechanism), it fails to find the right binaries:

```
% opam switch create test:a 4.09.1 
...
∗ installed base-bigarray.base
∗ installed base-threads.base
∗ installed base-unix.base
∗ installed ocaml-base-compiler.4.09.1
∗ installed ocaml-config.1
[ERROR] The compilation of ocaml failed at
        "/home/etienne/.opam/opam-init/hooks/sandbox.sh build ocaml
        /home/etienne/.opam/test:a/share/ocaml-config/gen_ocaml_config.ml 4.09.1
        ocaml".

#=== ERROR while compiling ocaml.4.09.1 =======================================#
# context     2.0.0 | linux/x86_64 | ocaml-base-compiler.4.09.1 | https://opam.ocaml.org/#9e7e7e45
# path        ~/.opam/test:a/.opam-switch/build/ocaml.4.09.1
# command     ~/.opam/opam-init/hooks/sandbox.sh build ocaml /home/etienne/.opam/test
:a/share/ocaml-config/gen_ocaml_config.ml 4.09.1 ocaml
# exit-code   1
# env-file    ~/.opam/log/ocaml-104246-ad6566.env
# output-file ~/.opam/log/ocaml-104246-ad6566.out
### output ###
# OCaml version mismatch: 4.08.1, expected 4.09.1
```

The reason for the error message is that it tries to find `ocaml` in `PATH`, but because there's a colon in there it cannot find it and falls back to my system compiler.

This happens in a local switch as well if the directory contains colons at any level.

### shell metacharacters (backticks, dollar sign)

When a switch name (or in the case of a local switch, a directory name) contains characters that are interpreted by the shell such as backticks or dollar signs, they are interpreted during the installation. For example, a switch created in a directory named `id` (backtick, i, d, backtick) will try to create a switch named by the output of this command, and a switch named `xmessage hello` (backtick, x, ..., o, backtick) will open popup window during switch creation. A similar behavior can be observed with dollar signs. This suggests that somewhere a command is built with double quotes and passed to the shell.

### Possible security concerns

Opam is primarily a development tool, so it's expected that it will run external code or install binaries. There might be scenarios where an attacker can control a directory name or a global switch name but I don't think that should be an important concern.

Thanks!

rjbou:
* Switch names (global or local directories) containing  `| '<' | '>' | '!' | '<backquote>' | '$' | '(' | ')' | ':'`, or empty are rejected. For windows compatibility, switch names that begin with `\w:/` & `\w:\` are allowed.  Empty switche names are forbidden.
  * Compliance regexp is `^(\w:/\\)?[^<>!$():]+$` (missing the backquote because of markdown) cf. [code](https://github.com/ocaml/opam/pull/4237/files#diff-69b74e9ec849a48241d3e03d0196540bR25)
* Update opam installer to not use an empty switch to retrieve needed paths
* Cherry-picked context-free switch paths from #4147 
* fix opam env `OPAMSWITCH` setting: if empty string, consider it as unset

fix #4231 

@rjbou edit: update content

### DIFF
--- a/src/format/opamSwitch.ml
+++ b/src/format/opamSwitch.ml
@@ -20,17 +20,18 @@ let is_external s =
 let external_dirname = "_opam"
 
 let check s =
-  let re =
-    Re.(compile @@
-        seq [
-          bol;
-          opt @@ seq [ wordc ; char ':'; set "/\\" ];
-          rep @@ diff any @@ set "<>!`$():";
-          eol
-        ])
-  in
-  (try ignore @@ Re.exec re s with Not_found ->
-     failwith (Printf.sprintf "Invalid character in switch name %S" s));
+  if String.compare s "" = 0 &&
+    let re =
+      Re.(compile @@
+          seq [
+            bol;
+            opt @@ seq [ wordc ; char ':'; set "/\\" ];
+            rep @@ diff any @@ set "<>!`$():";
+            eol
+          ])
+    in
+    (try ignore @@ Re.exec re s; true with Not_found -> false) then
+    failwith (Printf.sprintf "Invalid character in switch name %S" s);
   s
 
 let of_string s =
--- a/src/format/opamSwitch.ml
+++ b/src/format/opamSwitch.ml
@@ -19,12 +19,28 @@ let is_external s =
 
 let external_dirname = "_opam"
 
+let check s =
+  let re =
+    Re.(compile @@
+        seq [
+          bol;
+          opt @@ seq [ wordc ; char ':'; set "/\\" ];
+          rep @@ diff any @@ set "<>!`$():";
+          eol
+        ])
+  in
+  (try ignore @@ Re.exec re s with Not_found ->
+     failwith (Printf.sprintf "Invalid character in switch name %S" s));
+  s
+
 let of_string s =
+  check @@
   if is_external s then OpamFilename.Dir.(to_string (of_string s))
   else s
 
 let of_dirname d =
   let s = OpamFilename.Dir.to_string d in
+  check @@
   try
     let swdir = Unix.readlink (Filename.concat s external_dirname) in
     let swdir =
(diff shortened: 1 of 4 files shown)


rjbou:
Appveyor failure lightened something: should we allow local switches at system root, e.g. `/` or `C:\`?

src/format/opamSwitch.ml
```diff
@@ -19,12 +19,29 @@ let is_external s =
...
 let external_dirname = "_opam"
 
+let check s =
+  if String.compare s "" = 0 &&
+    let re =
+      Re.(compile @@
+          seq [
+            bol;
+            opt @@ seq [ wordc ; char ':'; set "/\\" ];
+            rep @@ diff any @@ set "<>!`$():";
```
AltGr, line 29:
```suggestion
            rep1 @@ diff any @@ set "<>!`$():";
```
Just making sure we refuse the empty string. This will also refuse Windows disk roots though (`E:\`)

rjbou: First I had `rep1`, but switched to `rep` to be able to parse root switches (`/` or `C:\`), and added the empty string check at the beginning to be sure to refuse empty switches.
Another solution would be to have the regexp : 
``
alt [ seq [bol; seq [ wordc ; char ':'; set "/\\" ]; eol]; (* accept root *)
       seq [bol;                                                                        (* accept non root paths *)
            opt @@ seq [ wordc ; char ':'; set "/\\" ];
            rep1 @@ diff any @@ set "<>!`$():";
             ...
```

AltGr: Ah ok, missed it, all good then :)

PR #4238, opened 2020-06-18
Add `opam install --depext-only` option

rjbou:
to install only external dependencies required by the install of targeted package(s).

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -1009,8 +1009,8 @@ let filter_unpinned_locally t atoms f =
          None))
     atoms
 
-let install_t t ?ask ?(ignore_conflicts=false) atoms add_to_roots
-    ~deps_only ~assume_built =
+let install_t t ?ask ?(ignore_conflicts=false) ?(depext_only=false)
+    atoms add_to_roots ~deps_only ~assume_built =
   log "INSTALL %a" (slog OpamFormula.string_of_atoms) atoms;
   let names = OpamPackage.Name.Set.of_list (List.rev_map fst atoms) in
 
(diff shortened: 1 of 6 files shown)


src/client/opamCommands.ml
```diff
@@ -1480,9 +1480,15 @@ let install =
        installed. If not, output the names of the missing dependencies to \
        stdout, and exits with 1."
   in
+  let depext_only =
+    mk_flag ["depext-only"]
+      "Install external dependencies required by $(i,PACKAGES) and their \
+       dependencies, and quit"
```
avsm, line 1486: Is it worth also forbidding `--deps-only --depext-only` with an error?

rjbou: It is actually handled, both option can be passed. Why do you think it should be forbidden?

avsm: Ah I see, so `--depext-only --deps-only` will install just the depexts for the dependencies, and not the main package?  It will also not install the opam packages for the dependencies.

That's worth clarifying in the help message if that's the case...

rjbou: Yes, in that case it only installs dependencies depexts. I'll update the doc.

src/client/opamClient.ml
```diff
@@ -1142,14 +1142,18 @@ let install_t t ?ask ?(ignore_conflicts=false) atoms add_to_roots
       OpamConsole.msg "%s"
         (OpamCudf.string_of_conflict t.packages
            (OpamSwitchState.unavailable_reason t) cs);
-      t, Conflicts cs
+      t, if depext_only then None else Some (Conflicts cs)
     | Success solution ->
```
AltGr, line 1146: I am not sure I understand this change... don't we want to report conflicts normally rather than silently ignore the request if there are any ?

src/client/opamCommands.ml
AltGr, line 1486:
```suggestion
      "Resolves the package installation normally, but only installs the required \
       system dependencies, without affecting the opam switch state or installing \
       opam packages"
```

src/client/opamClient.ml
AltGr: Ah indeed! Still don't completely understand why, but seems fine.

rjbou: In case of `depext_only`, we don't need to check the conflicting solution.

PR #4240, opened 2020-06-22
Add an optional solver relying on opam-0install-cudf

kit-ty-kate:
A month ago, @talex5 presented a new fast solver for opam in this discourse post: https://discuss.ocaml.org/t/ann-opam-0install-0-1-fast-opam-solver-for-ci-etc/5852

The main goal of the project was to give a fast, CI-oriented external solver without having to use opam.
However for my own CI needs, a builtin solver was more appropriate, thus I started working on a version of this solver which would use the CUDF interface instead of the opam API: opam-0install-cudf (version 0.2 available on opam)

So, after months of testing on [check.ocamllabs.io](http://check.ocamllabs.io) (aka. https://github.com/kit-ty-kate/opam-health-check), where this new solver replaced and outperformed the Z3 builtin solver, I believe the interface in the right one (*) and this is now ready to be included upsteam.

The one change slightly outside of the scope of this PR in the `OpamCudf` module, is here because during the testing phase when the solver wasn't ready yet, the solver would sometime say that there is no solution, but the Dose3 solver used to print the diagnostics succeeded. However the dose3 solution usually includes the removal of most of the packages in the current switch. So for users' sake I'd rather see this outcome be an hard error, asking the user to report it instead, just in case this happens with whichever solver the user is currently using.

~(*): The current PR relies on the next version of `opam-0install-cudf`: `opam-0install-cudf.0.3` which should include https://github.com/talex5/opam-0install-solver/pull/16 for this PR to compile.~
EDIT: `opam-0install-cudf.0.3` has been released a month ago.

### DIFF
--- a/src/solver/opamBuiltin0install.ml.real
+++ b/src/solver/opamBuiltin0install.ml.real
@@ -81,7 +81,10 @@ let reconstruct_universe universe selections =
     ) [] |>
   Cudf.load_universe
 
-let call ~criteria:_ ?timeout:_ (preamble, universe, request) =
+let call ~criteria ?timeout:_ (preamble, universe, request) =
+  if not (String.equal criteria default_criteria.crit_default) then begin
+    OpamConsole.warning "Custom CUDF criteria is not supported by the 0install solver";
+  end;
   let timer = OpamConsole.timer () in
   let pkgs, constraints = create_spec universe request in
   let context = Opam_0install_cudf.create ~constraints universe in
--- a/src/solver/opamBuiltin0install.ml.real
+++ b/src/solver/opamBuiltin0install.ml.real
@@ -1,0 +1,92 @@
+(**************************************************************************)
+(*                                                                        *)
+(*    Copyright 2017 OCamlPro                                             *)
+(*                                                                        *)
+(*  All rights reserved. This file is distributed under the terms of the  *)
+(*  GNU Lesser General Public License version 2.1, with the special       *)
+(*  exception on linking described in the file LICENSE.                   *)
+(*                                                                        *)
+(**************************************************************************)
+
+open OpamCudfSolverSig
+
+let log f = OpamConsole.log "0install" f
+
+let name = "builtin-0install"
+
+let ext = ref None
+
+let is_present () = true
+
+let command_name = None
+
+let default_criteria = {
+  crit_default = "";
+  crit_upgrade = "";
+  crit_fixup = "";
+  crit_best_effort_prefix = None;
+}
+
+let not_relop = function
+  | `Eq -> `Neq
+  | `Neq -> `Eq
+  | `Geq -> `Lt
+  | `Gt -> `Leq
+  | `Leq -> `Gt
+  | `Lt -> `Geq
+
+let if_not_already_in_spec f ((pkgs, _) as spec) ((pkgname, _) as vpkg) =
+  if not (List.exists (fun (pkg, _) -> String.equal pkg pkgname) pkgs)
+  then f spec vpkg
+  else spec
+
+let add_spec pkg req f_relop c (pkgs, constraints) =
+  let pkgs = (pkg, req) :: pkgs in
+  let constraints = match c with
+    | None -> constraints
+    | Some (relop, v) -> (pkg, (f_relop relop, v)) :: constraints
+  in
+  (pkgs, constraints)
+
+let essential spec (pkg, c) = add_spec pkg `Essential (fun x -> x) c spec
+let recommended spec (pkg, c) = add_spec pkg `Recommended (fun x -> x) c spec
+let restricts spec (pkg, c) = add_spec pkg `Restricts not_relop c spec
+
+let create_spec universe request =
+  let spec = ([], []) in
+  let spec = List.fold_left essential spec request.Cudf.install in
+  let spec = List.fold_left essential spec request.Cudf.upgrade in
+  let spec = List.fold_left restricts spec request.Cudf.remove in
+  Cudf.fold_packages_by_name (fun spec pkgname pkgs ->
+      match List.find_opt (fun pkg -> pkg.Cudf.installed) pkgs with
+      | Some {Cudf.keep = `Keep_version; version; _} -> essential spec (pkgname, Some (`Eq, version))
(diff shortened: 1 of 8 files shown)


src/solver/opamBuiltin0install.ml.real
```diff
@@ -0,0 +1,99 @@
+(**************************************************************************)
+(*                                                                        *)
+(*    Copyright 2017 OCamlPro                                             *)
```
avsm, line 3: This should be copyright @kit-ty-kate

src/solver/opamCudf.ml
```diff
@@ -1037,11 +1037,12 @@ let get_final_universe ~version_map univ req =
...
-    | Some {Algo.Diagnostic.result = Algo.Diagnostic.Success _; _}(*  -> *)
-      (* fail "inconsistent return value." *)
+    | Some {Algo.Diagnostic.result = Algo.Diagnostic.Success _; _}
     | None ->
-      (* External solver did not provide explanations, hopefully this will *)
-      check_request ~version_map univ req
+      raise (Solver_failure
+        "The current solver could not find a solution but dose3 could. \
+         This is probably a bug in the current solver. Please file a bug-report \
+         on the opam bug tracker: https://github.com/ocaml/opam/issues/")
```
AltGr, line 1045: I'll double-check, but I believe this actually disables display of conflicts _always_.

Note however that I have a pending patch that calls Dose's solver first all the time, since the cost is minimal when there is a solution, and the gain can be huge when there actually is none (I first did this when working on improving the conflict messages, but it turned out to be a more general improvement).

src/solver/opamBuiltin0install.ml.real
```diff
@@ -0,0 +1,99 @@
...
+let reconstruct_universe universe selections =
+  Opam_0install_cudf.packages_of_result selections |>
+  List.fold_left (fun pkgs (pkg, v) ->
+      let pkg = Cudf.lookup_package universe (pkg, v) in
+      {pkg with was_installed = pkg.installed; installed = true} :: pkgs
+    ) [] |>
+  Cudf.load_universe
+
+let call ~criteria:_ ?timeout:_ (preamble, universe, request) =
+  let t0 = Unix.gettimeofday () in
```
AltGr, line 85: Use `OpamConsole.timer ()`

src/solver/opamBuiltin0install.ml.real
```diff
@@ -0,0 +1,99 @@
...
+  | Ok selections ->
+    let universe = reconstruct_universe universe selections in
+    let t1 = Unix.gettimeofday () in
+    log "Solution found. Solve took %.2f s" (t1 -. t0);
+    (Some preamble, universe)
+  | Error _problem ->
+    let t1 = Unix.gettimeofday () in
+    log "No solution. Solve took %.2f s" (t1 -. t0);
+    (* Diagnostics disabled. Feel free to enable again for debug *)
+    (* print_endline (Opam_0install_cudf.diagnostics problem); *)
```
AltGr, line 98: `OpamConsole.log ~level:3 "%a" (slog Opam_0install_cudf.diagnostics) problem` ?

src/solver/opamBuiltin0install.ml.real
```diff
@@ -0,0 +1,99 @@
...
+let create_spec universe request =
+  let spec = ([], []) in
+  let spec = List.fold_left essential spec request.Cudf.install in
+  let spec = List.fold_left essential spec request.Cudf.upgrade in
+  let spec = List.fold_left restricts spec request.Cudf.remove in
+  Cudf.fold_packages_by_name (fun spec pkgname pkgs ->
+      match List.find_opt (fun pkg -> pkg.Cudf.installed) pkgs with
+      | Some {Cudf.keep = `Keep_version; version; _} -> essential spec (pkgname, Some (`Eq, version))
+      | Some {Cudf.keep = `Keep_package; _} -> essential spec (pkgname, None)
+      | Some {Cudf.keep = `Keep_feature; _} -> assert false (* TODO: Is this used? *)
```
AltGr, line 70: Not in opam no, we have no support for "features".

src/solver/opamBuiltin0install.ml.real
```diff
@@ -0,0 +1,99 @@
...
+
+let reconstruct_universe universe selections =
+  Opam_0install_cudf.packages_of_result selections |>
+  List.fold_left (fun pkgs (pkg, v) ->
+      let pkg = Cudf.lookup_package universe (pkg, v) in
+      {pkg with was_installed = pkg.installed; installed = true} :: pkgs
+    ) [] |>
+  Cudf.load_universe
+
+let call ~criteria:_ ?timeout:_ (preamble, universe, request) =
```
AltGr, line 84: Maybe print a warning if `~criteria <> default_criteria`, to avoid confusion due to the fact that criteria aren't supported by the solver ?

kit-ty-kate: Great idea! Done.

src/solver/opamBuiltin0install.ml.real
kit-ty-kate: Oh nice, I hadn't thought of that. Done.

src/solver/opamCudf.ml
kit-ty-kate: I'm not sure to understand. What do you mean by "disables display of conflicts *always*"?
The conflicts are being displayed correctly, I've never seen any issues here in several months of use. Is there a special case where this was correctly triggered?

kit-ty-kate:
I took care of the comments. Thanks a lot for the reviews.

> I'll check the build size, but probably we could even bundle it by default in our builds, to offer a good alternative in case of timeouts.

From what I can see the build size increase isn't significant. Building using opam:
* Size without this PR: 17,168Mo
* Size with 0install included: 17,738Mo

I'm not sure what you are using to reduce the size for the ready-to-use binaries but I suppose the difference shouldn't be significant either with that.

kit-ty-kate:
I rebased the branch against master and added an entry to the changelog.
I have a few improvements in the form of options for CI purpose in the works as well but I'll propose them independently once this PR gets merged.

rjbou:
Thanks!

PR #4242, opened 2020-06-23
Bootstrap Opam with OCaml 4.11.1

MisterDA:
This PR builds on https://github.com/ocaml/opam/pull/4234 to fix the little things that are missing and completely replaces 4.07.1 with 4.10.0 for bootstrapping.
Updating findlib to 1.8.1 is mandatory because 1.8.0 checks for the `-vmthread` flag of `ocamlc` which has been removed in 4.10.0.
EDIT: I removed the 'flexdll appveyor patch' in favor of using the latest flexdll commit available at the time of writing.
I have also updated and expanded Travis test matrix.

PR #4247, opened 2020-06-26
Cold boostrap back to 4.07.1

rjbou:
This reverts #4234 (1b0f4bacae7fe64b23b802fe04ae75d5c203ce24)
To integrate in #4242


avsm:
Offline discussion: we believe 4.09 to be safe to hop to before going all the way to 4.10

PR #4248, opened 2020-06-29
src-ext: fix extlib url

rjbou:
Seems that ygreg.org.ua is no more reachable and  the [package](https://opam.ocaml.org/packages/extlib/extlib.1.7.6/) redirects to ygrek.org.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -14,6 +14,7 @@ Possibly scripts breaking changes are prefixed with ✘
 ## Build
   * Opam file build using dune, removal of opam-%.install makefile target [#4178 @rjbou - fix #4173]
   * Use version var in opam file instead of equal current version number in opamlib dependencies [#4178 @rjbou]
+  * ext: fix extlib url [#4248 @rjbou]
 
 ## Install
   * Add `_build` to rsync exclusion list [#4230 @rjobou - fix #4195]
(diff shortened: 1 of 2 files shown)


MisterDA:
Why not switch to 1.7.7 and GitHub?
https://github.com/ygrek/ocaml-extlib/archive/1.7.7.tar.gz

dra27:
Assuming it's still working, there's an update-sources script in src_ext which should try to pull the URLs for the latest versions of all the lib-ext and lib-pkg packages from opam… the only tedium then is updating the patches, if necessary.

rjbou:
@MisterDA we rely on the package defined url.
@dra27 Thanks, I didn't saw the script, very helpful! I'm updating it to check that the package version is compliant with the 4.02.3 ocaml lower bound.
I'll merge this one to unlock the CI, and do another PR for the script update & version bump.

rjbou:
related discussion #4097 & PR #4098

dra27:
@rjbou - the script doesn't need to check that - the `src_ext` is meant to include patches to ensure 4.02.3 compatibility (if necessary). So after running the script, you then test `make lib-ext` and ensure that the patches work (usually, fortunately, it's a matter of deleting dune overlays). I'm just doing a PR for 4.09.1

rjbou:
ok, perfect! Thanks!

PR #4252, opened 2020-06-30
Disable recursive & subpath pinning

rjbou:
Partial revert of #3499, disable cli options `--recursive` and `--subpath`.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -50,6 +50,7 @@ New option are prefixed with ◈
 ## Pin
   * Don't keep unpinned package version if it exists in repo [#4073 @rjbou - fix #3630]
   * Fix path resolving when pinning with `file://` [#4209 @rjbou - fix #4208]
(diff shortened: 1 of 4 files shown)


master_changes.md
```diff
@@ -26,6 +26,7 @@ Possibly scripts breaking changes are prefixed with ✘
 
 ## Pin
   * Don't keep unpinned package version if it exists in repo [#4073 @rjbou - fix #3630]
+  * ✘ Disable recursive & subpath pinning [#4252 @rjbou]
```
avsm, line 29:
```suggestion
  * ✘ Disable recursive & subpath pinning (only present experimentally in opam 2.1alpha1) [#4252 @rjbou]
```

avsm:
Just one thing I noticed: the manual still reserves the "x-subpath" field, so that needs to be removed as well.

PR #4257, opened 2020-07-01
Advance cold compiler and infrastructure to OCaml 4.09.1

dra27:
This PR:
- Advanced the "cold" compiler to 4.09.1
- Updates the testing matrices to use 4.09.1 instead of 4.07.1 for testing
- Adds OCaml 4.10.0 to the Travis matrix
- Updates the src_exts

Dune is held to 1.11.4, but once this is merged I intend to do a separate PR to move forwards to Dune 2.0.


src_ext/patches/cmdliner.common/0001-Restore-4.02.3-support.patch
```diff
@@ -21,10 +21,11 @@ diff -Naur a/pkg/META b/pkg/META
 diff -Naur a/src/cmdliner.ml b/src/cmdliner.ml
 --- a/src/cmdliner.ml	2019-06-14 09:55:03.000000000 +0000
 +++ b/src/cmdliner.ml	2019-11-21 14:29:33.236981936 +0000
-@@ -4,11 +4,11 @@
+@@ -4,11 +4,12 @@
```
dra27, line 24: The changes in this patch are actually a consequence of Result 1.5 which (slightly annoyingly) sync'd the entire Result module for OCaml 4.08+ rather than just the type (annoying in the sense that I'd have made it version 2.0...)

PR #4260, opened 2020-07-07
Allow CI to download archives from opam archive cache

dra27:
The URL for the opam cache in `src_ext/Makefile` was incorrect. This PR also extends `make -C src_ext cache-archives` to use the opam cache if upstream tarballs aren't available (this was already the case for `make lib-ext`, but if `make -C src_ext cache-archives` fails then the CI build fails...)

### DIFF
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -116,18 +116,7 @@ wrap-install-commands: []
 wrap-remove-commands: []
 EOF
 
-    if [[ $COLD -eq 1 ]] ; then
-      if [ ! -x ~/local/bin/make ] ; then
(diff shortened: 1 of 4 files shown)


dra27:
This PR is running [in AppVeyor](https://ci.appveyor.com/project/dra27/opam/builds/33955568) with a commit which ensures that the caches are blown away to test that mode too.

dra27:
Hmm, 5th time lucky. The macOS Travis worker re-exposed an issue spotted in https://github.com/ocaml/opam/pull/3190, but which was incorrectly fixed by updating `make` instead. I've atoned for my past sin, and removed the stray `=` in `src_ext/Makefile`.

PR #4262, opened 2020-07-08
Update cold compiler to 4.09.1 (2.0 version)

dra27:
This cherry-picks:
 - `src_ext/update-sources.sh` changes from #3889
 - #4257 (except that only findlib and ocamlbuild are updated in `src_ext/Makefile.sources`)
 - #4260 (except with no change to the message in `shell/md5check.ml`)

### DIFF
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -116,18 +116,7 @@ wrap-install-commands: []
 wrap-remove-commands: []
 EOF
 
-    if [[ $COLD -eq 1 ]] ; then
-      if [ ! -x ~/local/bin/make ] ; then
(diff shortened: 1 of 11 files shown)


dra27:
This _should_ pass CI (fingers crossed) - the change to .travis.yml automatically invalidates the caches, but I have also checked that the new code in `.travis-ci.sh` correctly invalidates the cache.

The problem is that new PRs pull the master branch Travis cache (not sure why they don't pick up the 2.0 one - they do try it). The issue is not the opam root, it's the cached bootstrap `opam` in `~/local/bin/opam-bootstrap` which wasn't being reset. That should now be being done.

If this passes CI here, I'll cherry-pick 3cc78ea and b32c139 back to master.

PR #4264, opened 2020-07-08
Harden doc/man/dune

dra27:
Two fixes:

- Building the man pages should never inexplicably block the build
- There's a check to ensure that `%{bin:opam}` refers to something that's just been built and not something in `PATH`

cc @kit-ty-kate (replaces commit in #4229)


rjbou:
Thanks!

PR #4268, opened 2020-07-09
Depext: handle virtual packages on Debian

### LINKED ISSUE #4251: [opam 2.1~alpha] Debian virtual packages are not taken into account
Currently the `conf-gsl.1` package in `opam-repository has the following depext on debian:
```
["libgsl0-dev"] {os-family = "debian"}
```
The package used to install fine with opam 2.0.
However with opam 2.1 (master), the package won't install:
```
[ERROR] Package conf-gsl depends on the unavailable system package 'libgsl0-dev'. You can use `--no-depexts' to attempt installation anyway.
```
This is because `libgsl0-dev` is a virtual package (see https://www.debian.org/doc/debian-policy/ch-relationships.html#s-virtual for more informations about virtual packages) so given that opam tries to figure out which external packages are available and uses `apt-cache search .*` which does not really deal with virtual packages all that well (i.e. does not display them, but will give their respective "normal" package on demand), this fails.

One possible fix would be to use `apt-cache depends --no-pre-depends --no-depends --no-recommends --no-suggests --no-conflicts --no-breaks --no-replaces --no-enhances ".*"` instead of `apt-cache search --names-only ".*"`. It prints almost the same list of packages but the former also gives all the virtual packages (but between brackets).

rjbou:
Change querying method to `apt-cache showpkg` to retrieve available and virtual packages (and the concrete packages they provide).
This is the most efficient. Other tested methods:
- `apt-cache depends` for available (concrete and virtual) packages + `apt-cache search` for each virtual package -> 1.824s
- `apt-cache search` for available packages + `apt-cache showpkg` on all not found for virtual ones -> 0.235
- `apt-cache showpkg` for available (concrete and virtual) packages, and their mapping (+ regexp optim) -> 0.178s
For reference, on this run, the old version takes 0.151s.
Fix #4251

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -44,6 +44,7 @@ Possibly scripts breaking changes are prefixed with ✘
   * Homebrew: add no auto update env var for install, accept `pkgname` and `pkgnam@version` on query [#4200 @rjbou]
   * Force LC_ALL=C for query commands [#4200 @rjbou]
   * Fix install command dryrun [#4200 @rjbou]
+  * Handle debian virtual packages [#4267 @rjbou - fix #4251]
 
 
 ## Env
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -122,7 +122,6 @@ let family =
   ) in
   fun () -> Lazy.force family
 
-
 let packages_status packages =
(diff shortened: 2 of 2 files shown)


rjbou:
Superseded by #4269

PR #4270, opened 2020-07-10
Copy files instead of rsync when they are in a local cache

kit-ty-kate:
This PR is just to upstream a change I've been using in https://github.com/kit-ty-kate/opam-health-check.
This change improves slightly the time it takes to retrieve archives from a local cache by avoiding a rather unnecessary call to `rsync` from the cache to a temporary file (which is then going to be moved to the global download cache in `~/.opam/download-cache`), and replacing it by the `OpamSystem.copy_file` function improved in this wonderful PR: https://github.com/ocaml/opam/pull/4064

This PR approximatively improves the time it takes to retrieve archive by 0.02s per archive in cache on my test server. In my use-case (CI), opam is repetitively called and thus repetitively retrieve archive from the cache.
On a typical medium sized-example: `core.v0.14.0` for instance pulls 59 archives, so in total this PR make us gain 1.13s in this build. Multiply by 10_000 packages and this is now significant.

This was tested and timed with master + https://github.com/ocaml/opam/pull/4071 for simplicity and to avoid noise in the timing. (cc @Armael in case you want a simple-to-use commit to rebase #4071 you can use this one: https://github.com/kit-ty-kate/opam/commit/ac20d2f56fcd3befbb0032db6238137ac2962212)

Just to avoid confusion, I know I could directly move (or link) the entire cache to `~/.opam/download-cache` for an extra second of gain per 50 archives, but this is less composable and this PR is still an improvement nonetheless.

### DIFF
--- a/src/repository/opamRepository.ml
+++ b/src/repository/opamRepository.ml
@@ -80,9 +80,15 @@ let fetch_from_cache =
         ~quiet:true ~validate:false ~overwrite:true ~checksum
         url file
     | `rsync ->
-      (OpamLocal.rsync_file url file @@| function
-        | Result _ | Up_to_date _-> ()
-        | Not_available (s,l) -> raise (OpamDownload.Download_fail (s,l)))
+      begin match OpamUrl.local_file url with
+        | Some src ->
+          OpamFilename.copy ~src ~dst:file;
+          OpamProcess.Job.Op.Done ()
+        | None ->
+          (OpamLocal.rsync_file url file @@| function
+            | Result _ | Up_to_date _-> ()
+            | Not_available (s,l) -> raise (OpamDownload.Download_fail (s,l)))
+      end
     | #OpamUrl.version_control ->
       failwith "Version control not allowed as cache URL"
   in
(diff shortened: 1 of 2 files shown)


src/repository/opamRepository.ml
```diff
@@ -80,9 +80,17 @@ let fetch_from_cache =
...
         url file
     | `rsync ->
-      (OpamLocal.rsync_file url file @@| function
-        | Result _ | Up_to_date _-> ()
-        | Not_available (s,l) -> raise (OpamDownload.Download_fail (s,l)))
+      begin match OpamUrl.local_file url with
+        | Some src ->
+          let src = OpamFilename.to_string src in
+          let file = OpamFilename.to_string file in
+          OpamSystem.copy_file src file;
```
rjbou, line 87:
```suggestion
          OpamFilename.copy_file src file;
```
Better use `OpamFilename`, than `OpamSystem` (it's a redirection)

rjbou:
Minimal changes, but lgtm. thanks!

src/repository/opamRepository.ml
kit-ty-kate: Done. Thanks a lot, I wasn't aware this function existed.

PR #4273, opened 2020-07-14
Update Usage.md

robz:


rjbou:
@dra27 No, it's automatic.
@robz Thanks!

PR #4274, opened 2020-07-15
Use the uname syscall instead of calling the uname command

kit-ty-kate:
Opam uses the `uname` syscall to get hold of the `os`, `arch` and in some cases `os-version` variables in Unix environments.
On POSIX-compliant systems, the `uname` syscall is available which the `uname` command directly maps (for instance, in GNU coreutils: https://github.com/coreutils/coreutils/blob/master/src/uname.c)

The idea behind this PR is to get all these variables without having to call an external command and instead of the `uname` syscall.

This PR also fixes the value of `os-version` on FreeBSD, where it was previously gotten using `uname -U` (FreeBSD-only feature). However the value returned is rather odd for a version number:
```
$ uname -U
1201000
```
whereas the POSIX-compliant `uname -r` seems much more reasonable:
```
$ uname -r
12.1-RELEASE
```


hannesm:
> fixes the value of `os-version` on FreeBSD

see discussion in https://github.com/ocaml/opam/issues/3058

If I understand correctly, the goal of `os-version` is to be able to express "in version X the feature Y was implemented, thus the dependency (or behaviour) should change after X". I'm not aware of any opam package using this for FreeBSD, but from my intuition I'd favour the `uname -U` (which is a number that can be compared, changes whenever the ABI changes). The output of `uname -r` is imprecise, especially when not on a release.

dra27:
On the implementation: opam-core already has C stubs for Windows and https://github.com/ocaml/opam/pull/4265/commits/79c974026cc0a2a54dc0232794fdb7e5b545ac81 adds the possibility for these on Unix implementations (including Cygwin, which _doesn't_ use the Windows stubs). At least for now, it remains important that opam-core can be built without _needing_ C stubs, but it would be easy to generalise the `src/stubs/libacl` proposed in #4265 to be `src/stubs/unix`.

Out of interest, rather than automatic rejection, is there a specific motivation for converting this to a system call, or just hygiene?

kit-ty-kate:
> Out of interest, rather than automatic rejection, is there a specific motivation for converting this to a system call, or just hygiene?

Mostly hygene. Closing as it is too much of a hassle for the very little gain on cleanliness and performance

PR #4284, opened 2020-07-27
On init, check and propose to disable sandboxing if tool present & not working (docker, chroot, WLS1, etc.)

rjbou:
The check is done by calling the sandbox script with `sandbox <action> sh -c "echo SUCCESS >/tmp/t && cat /tmp/t"`, and checking that stdout is `SUCCESS`.
Disabling is done by removing from global config sandboxing wrappers, default ones, defined in `OpamInitDefaults.sandbox_wrappers`.

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -779,10 +779,17 @@ let update_with_init_config ?(overwrite=false) config init_config =
     (I.default_compiler init_config)
 
 let reinit ?(init_config=OpamInitDefaults.init_config()) ~interactive
-    ?dot_profile ?update_config ?env_hook ?completion ?inplace config shell =
+    ?dot_profile ?update_config ?env_hook ?completion ?inplace
+    ?(check_sandbox=true)
+    config shell =
   let root = OpamStateConfig.(!r.root_dir) in
   let config = update_with_init_config config init_config in
   let _all_ok = init_checks ~hard_fail_exn:false init_config in
+  let config =
+    if check_sandbox then
+      OpamAuxCommands.check_and_revert_sandboxing root config
+    else config
+  in
   OpamFile.Config.write (OpamPath.config root) config;
   let custom_init_scripts =
     let env v =
@@ -812,6 +819,7 @@ let init
     ~init_config ~interactive
     ?repo ?(bypass_checks=false)
     ?dot_profile ?update_config ?env_hook ?(completion=true)
+    ?(check_sandbox=true)
     shell =
   log "INIT %a"
     (slog @@ OpamStd.Option.to_string OpamRepositoryBackend.to_string) repo;
@@ -840,6 +848,11 @@ let init
         let config =
           update_with_init_config OpamFile.Config.empty init_config |>
           OpamFile.Config.with_repositories (List.map fst repos)
+        in
+        let config =
+          if check_sandbox then
+            OpamAuxCommands.check_and_revert_sandboxing root config
+          else config
         in
         OpamFile.Config.write config_f config;
 
--- a/.travis-ci.sh
+++ b/.travis-ci.sh
@@ -112,19 +112,6 @@ case "$TARGET" in
     git config --global gc.autoDetach false
 
   # Disable bubblewrap wrapping, it's not available within Docker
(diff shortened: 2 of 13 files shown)


src/client/opamClient.ml
```diff
@@ -752,6 +752,62 @@ let init_checks ?(hard_fail_exn=true) init_config =
...
+    let working =
+      try
+        OpamSystem.read_command_output ~allow_stdin:false (cmd @ test_cmd)
+        = ["SUCCESS"]
+      with e ->
+        (OpamConsole.error "Sandboxing is not working on your platform%s:\n%s"
+           (OpamStd.Option.to_string (fun os -> " "^os)
+              (OpamSysPoll.os_distribution ()))
+           (Printexc.to_string e);
+         not (OpamConsole.confirm ~default:false "Do you want to disable it?"))
```
avsm, line 793:
```suggestion
         not (OpamConsole.confirm ~default:false "Do you want to disable it? Note that this will result in less secure package builds, so please ensure that you have some other isolation mechanisms in place (such as running within a container or virtual machine)."))
```

This is a bit more verbose, but it at least makes the tradeoff clear to the user.

rjbou:
ping @mato?

src/client/opamClient.ml
```diff
@@ -752,6 +752,62 @@ let init_checks ?(hard_fail_exn=true) init_config =
...
+    let working =
+      try
+        OpamSystem.read_command_output ~allow_stdin:false (cmd @ test_cmd)
+        = ["SUCCESS"]
+      with e ->
+        (OpamConsole.error "Sandboxing is not working on your platform%s:\n%s"
+           (OpamStd.Option.to_string (fun os -> " "^os)
+              (OpamSysPoll.os_distribution ()))
+           (Printexc.to_string e);
+         not (OpamConsole.confirm ~default:false "Do you want to disable it? Note that this will result in less secure package builds, so please ensure that you have some other isolation mechanisms in place (such as running within a container or virtual machine)."))
```
AltGr, line 793: ouch my columns :grin:

src/client/opamClient.ml
```diff
@@ -752,6 +752,62 @@ let init_checks ?(hard_fail_exn=true) init_config =
   if hard_fail && hard_fail_exn then OpamStd.Sys.exit_because `Configuration_error
   else not (soft_fail || hard_fail)
 
+(* Check sandboxing script call. If it errors or unattended output, disable
```
AltGr, line 755: Not sure I would have chosen this module for this, since it's at the moment concerned with package-handling stuff and the switch state... but I don't have a much better proposition ?

src/client/opamClient.ml
```diff
@@ -752,6 +752,62 @@ let init_checks ?(hard_fail_exn=true) init_config =
...
+  let hook_var = OpamVariable.of_string "hooks" in
+  let hook_val = Some (OpamVariable.dirname (OpamPath.hooks_dir root)) in
+  let env = fun v ->
+    let fv = OpamVariable.Full.variable v in
+    if fv = hook_var then hook_val else
+      OpamStd.Option.Op.(OpamStd.Option.of_Not_found (List.assoc fv)
+                           OpamSysPoll.variables >>= Lazy.force)
+  in
+  match OpamFilter.commands env sdbx_wrappers with
+  | [] -> config
```
AltGr, line 781: Too bad we can't use the code of `OpamAction` here :/

AltGr (commented):
Seems good, see comments for fields where I'd have liked it differently, but you may not have had better options.

src/client/opamClient.ml
rjbou: ow, must had so much tabs at that time, that it was a 80col ^^

rjbou: ah, it was a github edited part

src/client/opamClient.ml
rjbou: moved to `OpamAuxCommand`

src/client/opamClient.ml
rjbou: Added `OpamEnv.hook_env`

PR #4294, opened 2020-07-31
Move src_ext Dune forwards to 2.6.2

dra27:
It's been possible to build opam with Dune 2.0 since its release (I think) - this PR moves the "lib-ext" copy - i.e. the "vendored" copy built if Dune available forwards from 1.11.4 to 2.6.2.

The only complexity for this is that Dune 2.6.2 has a minimum OCaml version of 4.07.0. I've worked around this by allowing `shell/bootstrap_ocaml.sh` to install the compiler to `src_ext/secondary`. So:

- When using `make compiler` / `make cold`, everything will just work because you just built OCaml 4.09.1
- When not using a "bootstrap" compiler, and it's the version is < 4.07.0, a bytecode-tools-only version of OCaml 4.09.1 is quickly assembled in `src_ext/secondary` and is used to build `src_ext/dune-local/dune.exe`. This Dune is then used as normal.


dra27:
Thanks, @rjbou! I don't think there's a pressing need for this on 2.0 branch - @avsm, is that correct?

PR #4298, opened 2020-08-03
Update opam-format to opam-file-format new type definition

rjbou:
~*To merge after `opam-file-format` new release, containing https://github.com/ocaml/opam-file-format/pull/24*~

See details in referenced PR

### DIFF
--- a/src_ext/Makefile.sources
+++ b/src_ext/Makefile.sources
@@ -38,8 +38,8 @@ MD5_mccs = 9c0038d0e945f742b9320a662566288b
 
 $(call PKG_SAME,mccs)
 
-URL_opam-file-format = https://github.com/ocaml/opam-file-format/archive/2.0.0.tar.gz
-MD5_opam-file-format = d7852cb63df0f442bed14ba2c5740135
+URL_opam-file-format = https://github.com/ocaml/opam-file-format/archive/2.1.1.tar.gz
+MD5_opam-file-format = c1949c5dff062c754d0642c7eded794f
 
 $(call PKG_SAME,opam-file-format)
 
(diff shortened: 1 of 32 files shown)


src_ext/Makefile.sources
```diff
@@ -38,8 +38,8 @@ MD5_mccs = 9c0038d0e945f742b9320a662566288b
 
 $(call PKG_SAME,mccs)
 
-URL_opam-file-format = https://github.com/ocaml/opam-file-format/archive/2.0.0.tar.gz
-MD5_opam-file-format = d7852cb63df0f442bed14ba2c5740135
+URL_opam-file-format = https://github.com/rjbou/opam-file-format/archive/endpos.tar.gz
+MD5_opam-file-format = 88c7beb3bde643a3aad04df0f7bbca38
```
rjbou, line 42:
```suggestion
MD5_opam-file-format = 9eee1a6cd5083510d4a7b193f7875db7
```

rjbou:
@kit-ty-kate replying to https://github.com/ocaml/opam/pull/4302#issuecomment-715655074 in this pr, as it is the origin.
> Btw changes some API bits used in dune-release. Here is the change I had to make to dune-release to get it to work with the new API: https://github.com/ocamllabs/dune-release/compare/master...kit-ty-kate:opam-2.1
> Is there a way to get some kind of backward compatible API? I feel like it might be annoying down the road for users if it ends up having a split of incompatible API.

In the `opam-file-format` library there is a backward compatible api, the legacy one, which is deprecated, and the new one, that can be used with `Module.FullPos`. cf. pr ocaml/opam-file-format#24.
In opam, it is not conceivable t have this kind of backward compatibility (maybe for `OpamFile.OPAM`, but even there, too much implication/duplication, need to check). But there is functions in `opam-file-format` that permit the translation from the old type def to the new one, cf. [OpamParser.mli](https://github.com/rjbou/opam-file-format/blob/endpos/src/opamParser.mli#L75). They could be integrated in `opam-format` itself, to ease their use & visibility.

PR #4300, opened 2020-08-04
Fix pin kind consistency

rjbou:
* With `opam pin target`, when opam file is not versioned and at root, vcs-pin the package instead of path-pin
* With `opam pin add nv target`, take opam file even if not versioned


rjbou:
The second point have an unattended side effect: if there is local modification and you want to use an opam file at version `hash` in git history, it is not possible.

PR #4301, opened 2020-08-05
`opam pin --with-version'

rjbou:
Add `--with-version` pin option to permit to specify the version of the pinned package.
```
$ opam pin add git+https://github.com/ocaml/opam --with-version 3.0 -n -y
This will pin the following packages: opam-client, opam-core, opam-devel, opam-format, opam-installer, opam-repository, opam-solver, opam-state. Continue?
[Y/n] y
opam-client is now pinned to git+https://github.com/ocaml/opam (version 3.0)
opam-core is now pinned to git+https://github.com/ocaml/opam (version 3.0)
opam-devel is now pinned to git+https://github.com/ocaml/opam (version 3.0)
opam-format is now pinned to git+https://github.com/ocaml/opam (version 3.0)
opam-installer is now pinned to git+https://github.com/ocaml/opam (version 3.0)
opam-repository is now pinned to git+https://github.com/ocaml/opam (version 3.0)
opam-solver is now pinned to git+https://github.com/ocaml/opam (version 3.0)
opam-state is now pinned to git+https://github.com/ocaml/opam (version 3.0)

$ opam pin
opam-client.3.0      (uninstalled)  git  git+https://github.com/ocaml/opam
opam-core.3.0        (uninstalled)  git  git+https://github.com/ocaml/opam
opam-devel.3.0       (uninstalled)  git  git+https://github.com/ocaml/opam
opam-format.3.0      (uninstalled)  git  git+https://github.com/ocaml/opam
opam-installer.3.0   (uninstalled)  git  git+https://github.com/ocaml/opam
opam-repository.3.0  (uninstalled)  git  git+https://github.com/ocaml/opam
opam-solver.3.0      (uninstalled)  git  git+https://github.com/ocaml/opam
opam-state.3.0       (uninstalled)  git  git+https://github.com/ocaml/opam
```

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -2875,8 +2875,9 @@ let pin ?(unpin_only=false) cli =
       "Set the pinning version to $(b,VERSION) for named $(b,PACKAGES) or \
        retrieved packagesi fro $(b,TARGET). It has priority to any other \
        version specification (opam file version field, $(i,name.vers) \
-       argument). It is equivalent to editing pinned package opam file to set \
-       its version."
+       argument)). With version pin, package source is version pin one and \
+       package version is the one specified with this option. It is equivalent \
+       to editing pinned package opam file to set its version."
       Arg.(some package_version) None
   in
   let guess_names kind ~recurse ?subpath url k =
@@ -3123,7 +3124,11 @@ let pin ?(unpin_only=false) cli =
     | `add_wtarget (n, target) ->
       (match (fst package) n with
        | `Ok (name,version) ->
-         let pin = pin_target kind target in
+         let pin =
+           match pin_target kind target, with_version with
+           | `Version v, Some v' -> `Source_version (v, v')
+           | p, _ -> p
+         in
          let version = OpamStd.Option.Op.(with_version ++ version) in
          OpamGlobalState.with_ `Lock_none @@ fun gt ->
          OpamSwitchState.with_ `Lock_write gt @@ fun st ->
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -2808,7 +2808,7 @@ let pin ?(unpin_only=false) cli =
     `P "If $(i,PACKAGE) has the form $(i,name.version), the pinned package \
         will be considered as version $(i,version) by opam. Beware that this \
         doesn't relate with the version of the source actually used for the \
-        package.";
+        package. See also $(i,--with-version) option.";
     `P "The default subcommand is $(i,list) if there are no further arguments, \
         and $(i,add) otherwise if unambiguous.";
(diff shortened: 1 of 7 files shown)


src/client/opamCommands.ml
```diff
@@ -2860,6 +2860,14 @@ let pin ?(unpin_only=false) () =
     mk_flag ["dev-repo"] "Pin to the upstream package source for the latest \
                           development version"
   in
+  let with_version =
+    mk_opt ["with-version"] "VERSION"
+      "Set $(b,PACKAGES) version to $(b,VERSION). If $(b,PACKAGES) contains a \
+       version, it is ignored. With version pin, pinned package source are \
+       taken from the version pin and the package version is the specified \
+       with this option."
+      Arg.(some package_version) None
```
AltGr, line 2869: This does not make it totally clear to me:
- does it force the version of all pinnings ?
- does is affect the source used for pinning, or the version it is pinned as, or both ?
- how does it interact with `opam pin foo.VERSION` ? I guess it should have lower priority or be an error, unless their meaning are different ; it seems to me this option has the same purpose, but the point is to scale with pinning URLs directly

Note that at the moment, weird stuff like `opam pin foo.2 3` is allowed, and simply ignores the `2`, so it's a version pin similar to `opam pin foo 3`. Maybe it should be an error ? You _could_ want to pin the package definition of version `2`, while pretending it's version `3` though ; that would no longer qualify as a version pinning however.

rjbou: >  does it force the version of all pinnings ?

yes all pinned packages of the current command

>    does is affect the source used for pinning, or the version it is pinned as, or both ?

It affects only the version of the pinned packages, it is similar to `opam pin <whatever sort of pinning> --edit` with a version change at edit stage.

>  version clash

In this implementation, the `--with-version` version is the one taken (highest priority), regardless of `foo.VERSION`, `foo version`, etc. On `opam pin foo 2 --with-version 3`, the code of version `2` is taken, and the package pinned with version `3` (cf. new [Source_version](https://github.com/ocaml/opam/pull/4301/files#diff-fbfddeb93a7d76d66f9bfd9b6cfbfec1R3096)).

AltGr: I am interested in the answers, but my point was that they should be more obvious from the commands/doc ;)
As for `opam pin foo.V1 --with-version V2`, it would seem more obvious to me to choose `V1`, since it's syntactlically "more specific" than `V2`. Of course, with a single package pinning, this makes no sense, and it's the main case... so maybe an error is still best.

rjbou: Funny, I had another (obvious) thought process about `opam pin foo.V1 --with-version V2`: as user specified the version with an additional version, `V2` has priority.
For the doc :+1:

rjbou: Updated

PR #4302, opened 2020-08-05
Opam file with preserved format more precise

### LINKED ISSUE #3993: `opam admin add-constraint` does too much reformatting
I tried to use `opam admin add-constraint` and I found out that apart from adding the correct constraint it also reformat unrelated lines in the `depends` field.
This makes it a bit impractical to use.

For example `opam admin add-constraint dune>=1.11` changes `hardcaml.v0.12.0`, `lwt.4.2.1` and other similar files in the following manner:
```diff
diff --git a/packages/hardcaml/hardcaml.v0.12.0/opam b/packages/hardcaml/hardcaml.v0.12.0/opam
index ca3c8171bf..87f513d55c 100644
--- a/packages/hardcaml/hardcaml.v0.12.0/opam
+++ b/packages/hardcaml/hardcaml.v0.12.0/opam
@@ -10,13 +10,13 @@ build: [
   ["dune" "build" "-p" name "-j" jobs]
 ]
 depends: [
-  "ocaml"            {>= "4.07.0"}
-  "base"             {>= "v0.12" & < "v0.13"}
-  "ppx_jane"         {>= "v0.12" & < "v0.13"}
-  "stdio"            {>= "v0.12" & < "v0.13"}
+  "ocaml" {>= "4.07.0"}
+  "base" {>= "v0.12" & < "v0.13"}
+  "ppx_jane" {>= "v0.12" & < "v0.13"}
+  "stdio" {>= "v0.12" & < "v0.13"}
   "topological_sort" {>= "v0.12" & < "v0.13"}
-  "dune"             {>= "1.5.1"}
-  "zarith"           {>= "1.5"}
+  "dune" {>= "1.11"}
+  "zarith" {>= "1.5"}
 ]
 synopsis: "RTL Hardware Design in OCaml"
 description: "
```
```diff
diff --git a/packages/lwt/lwt.4.2.1/opam b/packages/lwt/lwt.4.2.1/opam
index 6c4fb5e040..431af99ad7 100644
--- a/packages/lwt/lwt.4.2.1/opam
+++ b/packages/lwt/lwt.4.2.1/opam
@@ -20,12 +20,11 @@ dev-repo: "git+https://github.com/ocsigen/lwt.git"
 
 depends: [
   "cppo" {build & >= "1.1.0"}
-  "dune"
-  "mmap"  # mmap is needed as long as Lwt supports OCaml < 4.06.0.
+  "dune" {>= "1.11"}
+  "mmap"
   "ocaml" {>= "4.02.0"}
-  "result" # result is needed as long as Lwt supports OCaml 4.02.
-  "seq" # seq is needed as long as Lwt supports OCaml < 4.07.0.
-
+  "result"
+  "seq"
   "bisect_ppx" {dev & >= "1.3.0"}
   "ocamlfind" {dev & >= "1.7.3-1"}
 ]
```
This would require to modify `OpamFile.Syntax.to_string_with_preserved_format` by detecting which elements changed and only modifying the constraints bloc for those elements only, leaving the rest of the line untouched.

_Tested with opam 2.0.5_

rjbou:
Fixes #3993. To not merge before #4298
To take the same example, with `opam admin add-constraint dune>=1.11` : 
```diff
diff --git a/packages/hardcaml/hardcaml.v0.12.0/opam b/packages/hardcaml/hardcaml.v0.12.0/opam
index ca3c8171bf..87f513d55c 100644
--- a/packages/hardcaml/hardcaml.v0.12.0/opam
+++ b/packages/hardcaml/hardcaml.v0.12.0/opam
@@ -10,13 +10,13 @@ build: [
   ["dune" "build" "-p" name "-j" jobs]
 ]
 depends: [
-  "ocaml"            {>= "4.07.0"}
-  "base"             {>= "v0.12" & < "v0.13"}
-  "ppx_jane"         {>= "v0.12" & < "v0.13"}
-  "stdio"            {>= "v0.12" & < "v0.13"}
+  "ocaml" {>= "4.07.0"}
+  "base" {>= "v0.12" & < "v0.13"}
+  "ppx_jane" {>= "v0.12" & < "v0.13"}
+  "stdio" {>= "v0.12" & < "v0.13"}
   "topological_sort" {>= "v0.12" & < "v0.13"}
-  "dune"             {>= "1.5.1"}
-  "zarith"           {>= "1.5"}
+  "dune" {>= "1.11"}
+  "zarith" {>= "1.5"}
 ]
 synopsis: "RTL Hardware Design in OCaml"
 description: "
diff --git a/packages/lwt/lwt.4.2.1/opam b/packages/lwt/lwt.4.2.1/opam
index 6c4fb5e040..431af99ad7 100644
--- a/packages/lwt/lwt.4.2.1/opam
+++ b/packages/lwt/lwt.4.2.1/opam
@@ -20,12 +20,11 @@ dev-repo: "git+https://github.com/ocsigen/lwt.git"
 
 depends: [
   "cppo" {build & >= "1.1.0"}
-  "dune"
-  "mmap"  # mmap is needed as long as Lwt supports OCaml < 4.06.0.
+  "dune" {>= "1.11"}
+  "mmap"
   "ocaml" {>= "4.02.0"}
-  "result" # result is needed as long as Lwt supports OCaml 4.02.
-  "seq" # seq is needed as long as Lwt supports OCaml < 4.07.0.
-
+  "result"
+  "seq"
   "bisect_ppx" {dev & >= "1.3.0"}
   "ocamlfind" {dev & >= "1.7.3-1"}
 ]
```
to
```diff
diff --git a/packages/hardcaml/hardcaml.v0.12.0/opam b/packages/hardcaml/hardcaml.v0.12.0/opam
index ca3c8171bf..83cdbee304 100644
--- a/packages/hardcaml/hardcaml.v0.12.0/opam
+++ b/packages/hardcaml/hardcaml.v0.12.0/opam
@@ -15,7 +15,7 @@ depends: [
   "ppx_jane"         {>= "v0.12" & < "v0.13"}
   "stdio"            {>= "v0.12" & < "v0.13"}
   "topological_sort" {>= "v0.12" & < "v0.13"}
-  "dune"             {>= "1.5.1"}
+  "dune" {>= "1.11"}
   "zarith"           {>= "1.5"}
 ]
 synopsis: "RTL Hardware Design in OCaml"
diff --git a/packages/lwt/lwt.4.2.1/opam b/packages/lwt/lwt.4.2.1/opam
index 6c4fb5e040..91781964df 100644
--- a/packages/lwt/lwt.4.2.1/opam
+++ b/packages/lwt/lwt.4.2.1/opam
@@ -20,7 +20,7 @@ dev-repo: "git+https://github.com/ocsigen/lwt.git"
 
 depends: [
   "cppo" {build & >= "1.1.0"}
-  "dune"
+  "dune" {>= "1.11"}
   "mmap"  # mmap is needed as long as Lwt supports OCaml < 4.06.0.
   "ocaml" {>= "4.02.0"}
   "result" # result is needed as long as Lwt supports OCaml 4.02.
```

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -97,6 +97,7 @@ New option/command/subcommand are prefixed with ◈.
 
 # Opam file format
   * Update opam-format lib to opam-file-format end position and new type definition [#4298 @rjbou]
+  * `with_preserved_format` preserves in fields also, don't drop comments, etc. [#4302 @rjbou - fix #3993]
 
 ## Solver
   * Fix missing conflict message when trying to remove required packages [#4362 @AltGr]
--- a/src/format/opamFile.ml
+++ b/src/format/opamFile.ml
@@ -790,29 +790,115 @@ module Syntax = struct
         in
         aux [0] str
       in
-      let pos_index pos =
-        let (li, col) = pos.start in
-        lines_index.(li - 1) + col
-      in
-      let field_str ident =
-        let rec aux = function
-          | it1 :: r when it_ident it1 = ident ->
-            let start = pos_index it1.pos in
-            let stop = match r with
-              | it2 :: _ -> pos_index it2.pos - 1
-              | [] ->
-                let len = ref (String.length str) in
-                while str.[!len - 1] = '\n' do decr len done;
-                !len
-            in
-            String.sub str start (stop - start)
-          | _ :: r -> aux r
-          | [] -> raise Not_found
+      let pos_index (li,col) = lines_index.(li - 1) + col in
+      let extract start stop = String.sub str start (stop - start) in
+      let value_list_str lastpos vlst vlst_raw =
+        let extract_pos start stop = extract (pos_index start) (pos_index stop) in
+        let def_blank blank = OpamStd.Option.default "\n  " blank in
+        let find_split f =
+          let rec aux p = function
+            | x::r when f x -> Some (p, x, r)
+            | p::r -> aux (Some p) r
+            | [] -> None
+          in
+          aux None
         in
-        aux syn_file.file_contents
+        let full_vlst_raw = vlst_raw in
+        let rec aux lastpos blank acc vlst vlst_raw =
+          match vlst, vlst_raw with
+          | v::r, vraw :: rraw when OpamPrinter.value_equals v vraw ->
+            let blank = extract lastpos (pos_index vraw.pos.start) in
+            let str = extract_pos vraw.pos.start vraw.pos.stop in
+            let new_v = blank ^ str in
+            let blank = Some blank in
+            let lastpos = pos_index vraw.pos.stop in
+            aux lastpos blank (new_v :: acc) r rraw
+          | v::r , _ ->
+            (match find_split (OpamPrinter.value_equals v) full_vlst_raw with
+             | Some (pvraw, vraw, rraw) ->
+               let str = extract_pos vraw.pos.start vraw.pos.stop in
+               let blank, lastpos =
+                 if pos_index vraw.pos.start - lastpos <= 0 then
+                   def_blank blank, lastpos
+                 else
+                   (let start = match pvraw with
+                       | Some pvraw -> pos_index pvraw.pos.stop
+                       | None -> lastpos
+                    in
+                    let stop = pos_index vraw.pos.start in
+                    extract start stop),
+                   pos_index vraw.pos.stop
+               in
+               let new_v = blank ^ str in
+               let blank = Some blank in
+               aux lastpos blank (new_v :: acc) r rraw
+             | None ->
+               let blank, rraw, lastpos =
+                 match vlst_raw with
+                 | vraw :: rraw ->
+                   let blank = extract lastpos (pos_index vraw.pos.start) in
+                   let rraw, lastpos =
+                     if OpamStd.List.find_opt
+                         (OpamPrinter.value_equals vraw) vlst <> None then
+                       vlst_raw, lastpos
+                     else
+                       rraw, pos_index vraw.pos.stop
+                   in
+                   blank, rraw, lastpos
+                 | [] -> def_blank blank, vlst_raw, lastpos
+               in
+               let new_v = blank ^ (OpamPrinter.value v) in
+               let blank = Some blank in
+               aux lastpos blank (new_v :: acc) r rraw)
+          | [], _ ->  acc
+        in
+        aux lastpos None [] vlst vlst_raw
       in
-      let rem, strs =
-        List.fold_left (fun (rem, strs) item ->
+      let item_var_str name field =
+        let field_raw =
+          List.find (fun i -> it_ident i = `Var name) syn_file.file_contents
+        in
+        match field.pelem with
+        | Variable (n, { pelem = List { pelem = full_vlst;_}; _}) when n.pelem = name ->
+          let full_vlst_raw, full_vlst_raw_pos =
+            match field_raw.pelem with
+            | Variable (_, {pelem = List vlst_raw;  pos}) -> vlst_raw.pelem, pos
+            | _ -> raise Not_found
+          in
+          (* aux *)
+          let item_var_str =
+            let lastpos = pos_index full_vlst_raw_pos.start +1 in
+            let final_list = value_list_str lastpos full_vlst full_vlst_raw in
+            String.concat "" (List.rev final_list)
+          in
+          let beginning =
+            let start = pos_index field_raw.pos.start in
+            let stop = pos_index full_vlst_raw_pos.start +1 in
+            extract start stop
+          in
+          let ending =
+            let start = pos_index (List.hd (List.rev full_vlst_raw)).pos.stop in
+            let stop = pos_index full_vlst_raw_pos.stop in
+            extract start stop
+          in
+          beginning ^ item_var_str ^ ending
+        | _ -> OpamPrinter.items [field]
+      in
+      (* Fields *)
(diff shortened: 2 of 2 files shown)


kit-ty-kate:
I've used this branch to make https://github.com/ocaml/opam-repository/pull/17428 and it worked perfectly aside from one weird bug:
* For `gremlin.0.1.1`, `opam admin add-constraint "conduit-lwt-unix<3.0.0"` results in the `author` field being moved from line 3 to the end of the file and be renamed `authors` (with an s)

Aside from that it was almost perfect (a few spaces were removed for the line being modified but the rest did not move and it feels so good!

Thanks for this! :tada:

mseri:
Very neat job indeed.

Just a question remakr: we used to split conflicts on multiple lines otherwise the constraints were not correctly interpreted. Is this still an issue? If so, the tool is not splitting the conflicts on different lines, so that will need to be addressed as well

mseri:
Please ignore my comment, I tested it locally and the `|` in the conflicts filters seems to work perfectly fine. 
I think I got confused with an old unrelated issue (most likely https://github.com/ocaml/opam-repository/issues/15441).

rjbou:
Thanks for the tests!
@kit-ty-kate For the authors, I think I know from where it comes, but need to check first. For the spaces removed, it's normal. There (almost) no way to determine how the person wrote it's opam file, if it is organized by columns or just one space. So it is the default padding. The only `magic` that I introduced, is for the beginning of line (and newline) padding: it is always the same as previous line one. Thinking about it, if it is only a change in the filter, maybe i can keep the same padding as before... Need to check, this piece of code was written a months ago :)

kit-ty-kate:
I haven't tried but would that also erase potential comments put at the end of the line being updated?

rjbou:
From what I remember, it should...

kit-ty-kate:
This PR seems to have an unexpected effect on `dune-release`. In https://github.com/ocaml/opam-repository/pull/17481 I was using a dune-release compiled on a switch where opam has this PR merged in and the dune comment on the first line wasn't removed as it is per usual.

I would guess they need some kind of way in the API to say "remove this comment" or something like that now with this change. cc @NathanReb

kit-ty-kate:
Btw changes some API bits used in dune-release. Here is the change I had to make to `dune-release` to get it to work with the new API: https://github.com/ocamllabs/dune-release/compare/master...kit-ty-kate:opam-2.1
Is there a way to get some kind of backward compatible API? I feel like it might be annoying down the road for users if it ends up having a split of incompatible API. cppo and cie is good and all but adding a dependency just for basically one line change seems a bit excessive.

rjbou:
Comments: As the new version is more precise, it doesn't remove comments as before. It is indeed possible, in the function `with_preserved_format` to add a _remove all comments_ optional argument, but it won't be available from cli, only via api.

kit-ty-kate:
> Comments: As the new version is more precise, it doesn't remove comments as before. It is indeed possible, in the function `with_preserved_format` to add a _remove all comments_ optional argument, but it won't be available from cli, only via api.

dune-release only needs the API so it would be perfect!

rjbou:
Ok, i'll add it then. Not that it will also remove all authors comments, not only dune ones.

NathanReb:
Following an offline discussion with @kit-ty-kate it seems that the ideal behaviour in dune-release would be to remove only the dune comment.

Any thoughts on this?

If we agree that's the case then it's probably better to let dune-release filter out that comment. That will preserve other comments and allow dune-release code to be compatible with both 2.1 and 2.0 without conditional builds based on the opam version.

NathanReb:
Ah nevermind my comment, the API already changes in a breaking way! In that case then yeah indeed, being able to drop all comments seems fine, it will preserve dune-release's behaviour between opam 2.0 and 2.1. We'll have to maintain compat between the two versions anyway!

rjbou:
Thinking about it, It is more consistent for dune-release to remove the the dune comment than for opam to remove all comment just for one targeted line.
Anyway, it could be useful to have another function that clear all comments (and not difficult to do).

NathanReb:
If I understood this correclty, `OpamFile.OPAM.write_with_preserved_format` drops all comments and that's what dune-release currently uses. Offering a way to clear all comments in the new opam API would be nice as I think we should focus on preserving the current behaviour first. Does that sound reasonable to you?

If we want to change dune-release to include other comments we can take care of it later on and indeed this will probably require dune-release to strip the dune comment itself as I don't think it makes sense for opam to allow modifying comments in the `OpamFile` API nor to be aware of this particular comment.

rjbou:
The origin of this PR is that `write_with_preserved_format` was not precise enough on what is _preserved_. (cf. example in issue #4302) . As it is more precise, trying to modify only the concerned lines instead of fields, it is already a behaviour change. Keeping the comment is a side effect of trying to have the minimal change when writing.

rjbou:
And yes, if it's ok for `dune-release` to use another function to drop comments, I'd prefer than removing them from  new format, or adding an option to preserved format function. @AltGr an opinion on that ?

rjbou:
[no comment](https://github.com/ocaml/opam-file-format/pull/32) function ping @NathanReb @kit-ty-kate

rjbou:
In fact, I don't think that the `authors` issue should be fixed here. As the field in a legacy one (it should be `authors` according the [manual](https://opam.ocaml.org/doc/Manual.html#opamfield-authors)), we shouldn't keep a non accurate field.
There is a fallback to transform `author` into `authors`, I guess to not block for such a non critical field.

rjbou:
Good to go for me

PR #4303, opened 2020-08-05
Git empty repository (just initialised) error message

### LINKED ISSUE #3905: Confusing error messages related to bad VCS states
Using
```$ opam --version
2.0.4
```

I tried to `opam install .` a WIP project, only to have it fail with:

```
[ERROR] Could not synchronize /home/me/.opam/default/.opam-switch/sources/project-name from "git+file:///home/me/projects/project-name#master":
        "/usr/bin/git fetch -q file:///home/me/projects/project-name +master:refs/remotes/opam-ref-master" exited with code 128 "fatal: couldn't find remote ref master"
[ERROR] The sources of the following couldn't be obtained, aborting:
          - project-name.~dev
```

this means nothing to me. Fortunately, i recalled that opam implicitly ties itself in to the state of your git if you happen to have a `.git` dir in the directory. So i made sure that all the changes were committed, and then the install went forward fine.

A message along the following lines would be much more helpful and user friendly:

```
[ERROR] Project project-name has a `.git` file but opam could not synchronize with the `master` branch. 

Commit all changes to `master` and run the command again.
```

I could easily see this kind of thing causing a newbie to give up on our ecosystem, so I think it's important to get ironed out.
 
I would be happy to help improve the state of this, if someone want's to point me at the right block of code.

rjbou:
fix #3905

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -47,6 +47,9 @@ New option/command/subcommand are prefixed with ◈.
 ## Repository management
   *
 
+## VCS
+  * Add error message in case git repo is empty [#4303 @rjbou - fix #3905]
+
 ## Build
   * src-ext: bump topkg to 1.0.2 and dune to 2.6.2, with a second compiler built in case main one is < 4.07.0 (dune restriction) [#4294 @dra27]
   * Bump to 2.1.0~alpha3 version [#4299 @rjbou]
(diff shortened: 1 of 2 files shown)


AltGr:
LGTM!

PR #4314, opened 2020-08-17
Opam root light upgrade

rjbou:
When a read lock is acquired, upgrade config file _on-the-fly_ instead of the blocking process.

### DIFF
--- a/src/state/opamFormatUpgrade.ml
+++ b/src/state/opamFormatUpgrade.ml
@@ -1052,40 +1052,55 @@ let from_2_0_to_2_1 _ conf = conf
 
 let latest_version = OpamFile.Config.format_version
 
+let latest_hard_upgrade = v2_0
+
 let as_necessary global_lock root config =
   let config_version = OpamFile.Config.opam_version config in
   let cmp =
     OpamVersion.(compare OpamFile.Config.format_version config_version)
   in
-  if cmp = 0 then ()
-  else if cmp < 0 then
-    if OpamFormatConfig.(!r.skip_version_checks) then () else
+  if cmp = 0 then config else
+  if cmp < 0 then
+    if OpamFormatConfig.(!r.skip_version_checks) then config else
       OpamConsole.error_and_exit `Configuration_error
         "%s reports a newer opam version, aborting."
         (OpamFilename.Dir.to_string root)
   else
-  if OpamVersion.compare config_version OpamFile.Config.format_version >= 0 then ()
+  if OpamVersion.compare config_version OpamFile.Config.format_version >= 0 then
+    config
   else
+  let on_the_fly, file_lock =
+    if OpamVersion.compare config_version latest_hard_upgrade >= 0
+    && OpamSystem.get_lock_flag global_lock <> `Lock_write then
+      (log "On the fly config upgrade, from %s to %s"
+         (OpamVersion.to_string config_version)
+         (OpamVersion.to_string latest_version);
+       true, `Lock_read)
+    else
+      false, `Lock_write
+  in
   let is_dev = OpamVersion.git () <> None in
-  OpamConsole.formatted_msg
-    "This %sversion of opam requires an update to the layout of %s \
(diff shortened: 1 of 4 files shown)


src/state/opamFormatUpgrade.ml
```diff
@@ -1052,49 +1052,65 @@ let from_2_0_to_2_1 _ conf = conf
...
   let is_dev = OpamVersion.git () <> None in
-  OpamConsole.formatted_msg
-    "This %sversion of opam requires an update to the layout of %s \
-     from version %s to version %s, which can't be reverted.\n\
-     You may want to back it up before going further.\n"
-    (if is_dev then "development " else "")
-    (OpamFilename.Dir.to_string root)
-    (OpamVersion.to_string config_version)
-    (OpamVersion.to_string latest_version);
+  if not on_the_fly then
```
rjbou, line 1083: Add a message for the light upgrade?

src/state/opamFormatUpgrade.ml
```diff
@@ -1052,49 +1052,65 @@ let from_2_0_to_2_1 _ conf = conf
 
 let latest_version = OpamFile.Config.format_version
 
+let latest_light_upgrade = v2_0
```
AltGr, line 1055: Shouldn't it actually be called `latest_hard_upgrade` ?

src/state/opamFormatUpgrade.ml
```diff
@@ -1052,49 +1052,65 @@ let from_2_0_to_2_1 _ conf = conf
...
+       You may want to back it up before going further.\n"
+      (if is_dev then "development " else "")
+      (OpamFilename.Dir.to_string root)
+      (OpamVersion.to_string config_version)
+      (OpamVersion.to_string latest_version);
   let dontblock =
     (* Deadlock until one is killed in interactive mode, but abort in batch *)
     if OpamStd.Sys.tty_out then None else Some true
   in
   try
```
AltGr, line 1096: The code seems ok but a bit fragile ; I would have found it safer to have two completely separate sections for `current` → `latest hard upgrade` if necessary, and within the `with_flock_upgrade`, then `latest  hard` → `latest`, with a final write of the conf file only if possible.

src/state/opamGlobalState.ml
```diff
@@ -44,7 +44,12 @@ let load lock_kind =
   let has_root = OpamFilename.exists_dir root in
   let global_lock =
     if has_root then
-      OpamFilename.flock `Lock_read (OpamPath.lock root)
+      let lock = (* needed for on-the-fly upgrade config file *)
+        match lock_kind with
+        | `Lock_none | `Lock_read -> `Lock_read
```
AltGr, line 49: Wow, aren't you forcing a read-lock for all commands here ?
Why would you need this ?

src/state/opamFormatUpgrade.ml
AltGr: (maybe the light updates should still be called whithin the `with_flock_upgrade` if you are doing it; my point is that I wouldn't expect `update_to v2_0       from_2_0_beta5_to_2_0` to get called within code reachable when `on_the_fly`, even if we expect it to be a no-op in this case)

AltGr (changes requested):

src/state/opamGlobalState.ml
rjbou: No it's only the `global_lock` that is used for the upgrade, the _real_ conf lock is [`config_lock`](https://github.com/ocaml/opam/pull/4314/files/1ddb5b1d3994cca66fdea87b09d9a1ea7f6c8b8f#diff-927760da3264767b18aaab24a8795803L56)

AltGr: ah right, of course

src/state/opamFormatUpgrade.ml
rjbou: Should more understandable now :)

PR #4316, opened 2020-08-18
Implement CLI versioning

dra27:
Implements the [CLI Versioning spec](https://github.com/ocaml/opam/wiki/Spec-for-opam-CLI-versioning)

Fundamentally, this PR allows the user/author to select backwards-compatibility with the opam 2.0 CLI by setting `OPAMCLI` to `2.0` and forwards-compatibility with future releases of the opam client by passing `--cli=2.1` _anywhere_ on the command line.

This is implemented by increasing the pre-processing which is done to `Sys.argv` before `Cmdliner.Term.eval` is invoked. Various things to note:

- While refactoring, I noticed that there's no reason for not supporting `opam adm` as an abbreviation for `opam admin` so this is now implemented (see `OpamCommands.is_admin_subcommand`).
- I've converted `OpamConsole.log` to use `Format.fprintf` instead of `Printf.fprintf` which has been on my TODO-list for a while. This actually improves the Windows story ever-so-slightly and it turns out with only tiny alterations to a couple of messages which included `@` signs. In future, it will allow the formatting to be applied using tags which will be better in terms of the code and also better for Windows (since `Format` will do the heavy lifting instead of my horrendous VT100 parser). I've done it here because I wanted to be able to debug the CLI parsing process, and its hard to set `OpamCoreConfig.!r.debug_level` _before_ you've parsed the command-line!
- opam already included pre-processing of `Sys.argv` in order to allow `opam --yes plugin args` to mean "automatically build and install plugin and then call it with args". I've extracted this pre-processing and included the `--cli` processing with it.
- I have changed the processing of `--yes` to allow *all* commands to benefit from it. This means that `opam --yes list` now works, where before it was a syntax error. This is necessary to allow plugins to become built-in commands compatibly. At the moment this has a slight change of semantics in that `opam --yes lock` now means `opam lock --yes` where before the plugin would not have been invoked with `--yes`. I think this is acceptable and consistent to do universally, rather than tying that handling to CLI version (although we could...).
- `--yes` is transformed - that is, `opam --yes list` is still handled by Cmdliner, it just sees `opam list --yes` instead. `--cli` is completely extracted since it affects the generation of the Cmdliner parser - `--cli` itself is included in the parser only so that it's included in `--help` (in particular, this is why parsing the prefix `--cl` is important or `opam --cli=2.0 --cl=2.1` would be a contradiction).

### DIFF
--- a/doc/index.html
+++ b/doc/index.html
@@ -302,6 +302,8 @@ td.sublib {
   <td>All sub-commands of the "opam admin" command</td></tr>
 <tr><th><a href="opam-client/OpamCommands">opamCommands.ml</a></th>
   <td>Opam CLI commands and their handlers as Cmdliner terms</td></tr>
+<tr><th><a href="opam-client/OpamCLIVersion">opamCLIVersion.ml</a></th>
+  <td>Functions for the CLI versioning</td></tr>
 
 <tr><td colspan="2" class="sublib">Main opam CLI</td></tr>
 <tr><th><a href="opam-client/OpamMain">opamMain.ml</a></th>
--- a/doc/pages/FAQ.md
+++ b/doc/pages/FAQ.md
@@ -411,3 +411,27 @@ you could lose your opam configuration at each reboot.
 You can use the [`nfsopam`](https://github.com/UnixJunkie/nfsopam) script to
 have the best of both worlds: persistence of NFS directories and fast operations
 of local directories.
+
+---
+
+#### 🐫  What does the `--cli` option do? Should I be using it everywhere?
+
+`--cli` was introduced in opam 2.1 to deal with changes in the command line
+interface between releases. It tells opam to interpret the command line as a
+specific version, in particular it means that new options or options which
+have had their meaning altered will not be available, or will be behave as they
+did in that version. It only affects the command-line - it does not, for
+example, stop a root from being upgraded from an older version to the current
+version.
+
+We recommend using it in scripts (and programs which call opam) since they can
+then expect to work seamlessly with future versions of the opam client. It's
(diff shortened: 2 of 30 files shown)


dra27:
I'm now working on actually doing something the information, and will push more!

rjbou:
Rebased!

AltGr:
Great work, thanks!
I have a couple questions:
- you remove the individual commands from `OpamCommands`; I don't mind, but did you ensure they weren't useful when calling as lib?
- did you check the changelog to make sure `OPAMCLI=2.0` works as expected on all current cli changes?

Also, while I agree with the clearer `opam --yes plugin args`, in this case it would seem more obvious to me that `--yes` only applies to `opam` and not to `plugin`. Also, it seems that `opam plugin --yes` no longer passes the `--yes` option to opam, as I believe was the case (I don't remember precisely where it was useful, but I think some scripts used it to e.g. allow format upgrade; @avsm might remember more). I am just being overly prudent about potential breakages ;)

dra27:
Catching up, and answering @AltGr's questions:

- I'm not sure _why_ anyone would want to use the commands and, if they did then they need to handle CLI versioning now - so I went with "it needs to be broken, rather than silently wrong" idea!
- Note that this PR didn't implement any of the actual changes - it's just the handling. So I have a big TODO list of the things which need addressing (I posted it in Slack before vacation). However, it's also the case that things which don't work as before are bugs, so are fixable in a 2.1 point release (obviously, we should aim to deal with them before release!)
- I don't think `opam plugin --yes` ever passed the `--yes` option to opam before?! The scanning for the auxiliary command happened earlier than that, I think?

PR #4317, opened 2020-08-19
CLI help updates

dra27:
Best checked commit-by-commit:

- Use `Cmdliner.Manpage.s_` constants instead of hard-coded section names (small tidy-up, although possibly one "bug" found)
- Aliases put the `--help` and `--version` options in `OPTIONS` instead of `COMMON OPTIONS`
- `opam lock --help` was missing all the common information (`ENVIRONMENT`, `BUGS`, etc.)

There are two other things to consider/fix:
- We call the section `ENVIRONMENT VARIABLES` (`OpamArg.help_sections`) but the standard name is `ENVIRONMENT` ([`Cmdliner.Manpage.s_environment`](https://erratique.ch/software/cmdliner/doc/Cmdliner.Manpage.html#VALs_environment)) - any reason not to change this?
- opam lock has a section `What is changed in the lock file?` - this should be in capitals and re-worded. Perhaps `LOCK PROCEDURE`, perhaps?

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -3,6 +3,9 @@ note.
 Possibly scripts breaking changes are prefixed with ✘.
 New option/command/subcommand are prefixed with ◈.
 
+## Global CLI
+  * --help/--version documented in wrong section for aliases [#4317 @dra27]
+
 ## Init
   *
 
--- a/master_changes.md
+++ b/master_changes.md
@@ -5,6 +5,7 @@ New option/command/subcommand are prefixed with ◈.
 
 ## Global CLI
   * --help/--version documented in wrong section for aliases [#4317 @dra27]
+  * opam lock --help missing common information {#4317 @dra27]
 
 ## Init
   *
--- a/src/client/opamAdminCommand.ml
+++ b/src/client/opamAdminCommand.ml
@@ -28,7 +28,7 @@ let admin_command_doc =
   "Tools for repository administrators"
 
 let admin_command_man = [
-  `S "DESCRIPTION";
+  `S Manpage.s_description;
   `P (Printf.sprintf
        "This command can perform various actions on repositories in the opam \
         format. It is expected to be run from the root of a repository, i.e. a \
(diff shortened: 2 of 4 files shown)


dra27:
1e5c401 FWIW should go to 2.0 as well.

rjbou:
On the environment variable, no reason to not switch to `Cmdliner.Manpage.s_environment`.
For opam lock, it is not really the procedure, but what contains the lock file regarding the regular opam file, maybe something like `LOCK FILE CONTENT`?

master_changes.md
```diff
@@ -3,6 +3,10 @@ note.
 Possibly scripts breaking changes are prefixed with ✘.
 New option/command/subcommand are prefixed with ◈.
 
+## Global CLI
+  * --help/--version documented in wrong section for aliases
+  * opam lock --help missing common information
```
rjbou, line 8:
```suggestion
  * --help/--version documented in wrong section for aliases [#4317 @dra27]
  * opam lock --help missing common information [#4317 @dra27]
  * Use `Cmdliner.Manpage.s_*` instead of constants  [#4317 @dra27]
```

dra27: Oops - updated!

dra27:
How about `LOCK FILE UPDATES`? "CONTENT" to me implied that it completely writes the file, but I (think?!) the point is that it makes changes to it, right?

rjbou:
ok with that, or a more specific `LOCK FILE UPDATED FIELDS`, to not let think that it is updates on the lock file process, but the fields that are updated from the opam file.

rjbou (approved):

AltGr:
> We call the section ENVIRONMENT VARIABLES (OpamArg.help_sections) but the standard name is ENVIRONMENT (Cmdliner.Manpage.s_environment) - any reason not to change this?

Only reason I could see is the completion scripts that parse the manpage output; but they don't use this section so it should be fine.

PR #4322, opened 2020-08-19
Fix --columns advertised options

dra27:
#3931 allows `<field>` to refer directly to a raw opam field and so added an entry to `OpamListCommand.field_names`. This list is also used for the `--columns` argument to `opam list` and `opam admin list` which still requires the trailing colon at all times.

This PR removes the spurious `<field>` from the help for `--columns` - this is correct (I think) since `--raw` can't be used here as `--columns` is a list of fields.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -22,7 +22,7 @@ New option/command/subcommand are prefixed with ◈.
   *
 
 ## List
-  *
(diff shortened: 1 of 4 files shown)


rjbou (approved):
Thanks!

PR #4324, opened 2020-08-19
Consistently have -n as alias of --no-action

dra27:
`opam pin` has `-n` as an alias of `--no-action` (cf. `-n` meaning `--no-setup` in `opam init` as well)


rjbou (approved):
Thanks!

Issue #4325, opened 2020-08-20
`.opam-switch/environment` file format is ambiguous and can lead to errors


lefessan:
Currently, the `.opam-switch/environment` file is wrongly parsed by `opam` when the value of a variable is empty:

If my file is 
```
CPP     =               Updated\ by\ package\ ocaml-base-compiler
```
`CPP` will wrongly be set to `Updated by package ocaml-base-compiler`. This will prevent creating another switch after `eval $(opam env)` because ocaml will fail to compile with a wrong CPP var.

An immediate fix should be to stop writing these "Updated by package $name" comments. A long-term fix should be to use a more standard format.

dra27:
Which version of opam?

dra27:
This rang a bell, and some digging reveals that I hit this years ago and fixed it in https://github.com/dra27/opam/commit/246b518fdf79f4c7dd76e475fd45e553195a43db, however that commit seems to have got lost in my myriad rebasing during the opam 2.0 cycle. My solution in that was to encode `""` as `@` and `"@"` as `\@`. The lexer was updated to recognise the single token `@` as meaning `""` - no handling is needed for `\@` since that's dealt with automatically by the line lexer already.

dra27:
I agree that the format could be changed; the comment is useful though.

PR #4326, opened 2020-08-20
Handle empty environment variable updates

dra27:
This is apparently a piece of lost history on my part, but it's resurfaced in #4325. Empty strings cause corrupt environment files, since two tabs are written in a row which are of course lexed as a single piece of whitespace.

For better or worse, my solution in 2016 was to write an empty string as `@`. The ambiguity with the string `"@"` is resolved by writing that as `\@` (since the format already allows `\` to escape _any_ character.

This can go on 2.0 with one caveat - it would mean that any environment updates which intentionally set a value to be `@` will now be empty. That seems unlikely, but it may be reason enough to fix this for 2.1 only.

### DIFF
--- a/src/format/opamFile.ml
+++ b/src/format/opamFile.ml
@@ -294,7 +294,12 @@ module LinesBase = struct
       aux acc (i-1) in
     aux ([],0) (len - 1)
 
-  let escape_spaces str =
+  let escape_spaces = function
+  | "" ->
+      "@"
+  | "@" ->
+      "\\@"
+  | str ->
     let len = String.length str in
     match find_escapes str len with
     | [], _ -> str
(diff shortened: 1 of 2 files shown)


AltGr (commented):
Change seems fine;
one thing I am still not very comfortable about is that we don't support _unsetting_ variables (this is particularly true for `opam env --revert`), and we set them to the empty string instead. It has more or less worked until now, and fixing it would need specific variable-unsetting code for all supported shells with all the pitfalls that involves, which is why it's still like that.

The reason I mention it here is that maybe we should provision a syntax for that as well (at first when @rjbou told me about this PR I thought that's what the `@` syntax was about)

dra27:
Windows doesn't support empty environment variables, so for the most part I think it's better to (try to) head towards a world where we treat "unset" and "empty" as the same thing.

dra27:
Has this been cherry-picked to `master`?

rjbou:
Nop, not yet

PR #4328, opened 2020-08-24
Optimise package name comparison

### LINKED ISSUE #4245: Performance analysis: opam show and opam switch list-available
Hi,

I noticed that some commands like `opam show $pkg` or `opam switch list-available` take a lot of time. This happens in 2.0.0 and in 2bf062d2c as well.

I first had a look at what it is doing using strace and found that it is opening the same files many times, in particular, switch-state
files for all switches (global and local ones). I have 55 switches on my machine, and each is read about 50 times as well, so it might be a quadratic behavior. (I was a bit surprised that local switches are included in the output of this command)

<details>
<summary>strace output</summary>

    % strace -c -e openat opam show dune
    ...
     time     seconds  usecs/call     calls    errors syscall
    ------ ----------- ----------- --------- --------- ----------------
    100.00    0.004137           1      3721       385 openat

    % strace -e openat opam show dune 2>&1 |grep ^openat|sort|uniq -c|sort -n
    ... 55 lines of :
    56 openat(AT_FDCWD, "path/to/_opam/.opam-switch/switch-state", O_RDONLY) = 4
</details>

I confirmed using perf and flamegraph that dealing with these files (in particular parsing them) is taking an important time in the whole command run.

<details>
<summary>flamegraph</summary>

    % perf record -g --call-graph=dwarf opam show dune
    % perf script | stackcollapse-perf.pl | stackcollapse-recursive.pl | flamegraph.pl > flame-opam-show.svg

flame-opam-show_000

In the screenshot, I highlighted in purple the `safe_read` function. It's visible in several places, but the `output_table` -> `version_color` -> `read_opt` branch alone takes 36% of the whole running time.

The two towers on the left are external processes: `apt-cache` and `lsb-release`, which I suppose are related to depext being part of opam now.
</details>

As a quick fix, it is possible to set a global cache for file reading and it fixes the problem. However, it could significantly increase
opam's memory usage, and is not correct since in other cases, opam might want to re-read the file (in other words, this would need a cache invalidation strategy). A more local solution would be necessary, for example by parsing the file once and passing the parsed value down.

This eliminates most of the file parsing, and pushes the `version_color` branch under 2%. The next thing that is visible in the profile is that a lot of time is spent in comparison functions for package names and package versions (note that it is also visible in the original version as well).

<details>
<summary>flamegraph with read cache highlighting comparison functions</summary>

flame-opam-compare

</details>

Comparing package names uses the following function:

https://github.com/ocaml/opam/blob/2bf062d2c755b2316ecad7198810c155c406ace5/src/format/opamPackage.ml#L86-L89

This allocates and copies the two strings, which puts some pressure on the GC. A non-allocating alternative is todo a case-insensitive comparison on chars (see [this example in base](https://github.com/janestreet/base/blob/be8d141e711ee8ae3e92690498a5e17c4f3f98ea/src/string.ml#L518-L546)).

Comparing package versions is a bit more complex. `skip_while_from` is a hot spot, and might be microoptimized; also there are some `String.sub` (which allocates and copies) calls that might be replaced by index manipulation (or a substring abstraction). More importantly, it might be useful to store version numbers as a parsed representation on which an efficient comparison function can be worked, rather than strings.

Apologies for the meta-issue; we can open individual issues and close this issue or move it to the wiki if needed.

Thanks!

AltGr:
Removing unneeded string allocations can possibly have a big impact due
to the large number of set/map operations we have.

Fixes #4245, thanks for the suggestion

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -74,7 +74,7 @@ New option/command/subcommand are prefixed with ◈.
   * Don't penalise packages with more recent 'hidden-versions' [#4312 @AltGr]
 
 ## Internal
-  *
+  * Optimise package name comparison [#4328 @AltGr - fix #4245]
 
 ## Test
   *
--- a/src/core/opamStd.ml
+++ b/src/core/opamStd.ml
@@ -639,6 +639,26 @@ module OpamString = struct
     for i = 0 to String.length s - 1 do acc := f !acc s.[i] done;
     !acc
 
+  let compare_case s1 s2 =
+    let l1 = String.length s1 and l2 = String.length s2 in
+    let len = min l1 l2 in
+    let rec aux i =
+      if i < len then
+        let c1 = s1.[i] and c2 = s2.[i] in
+        match Char.compare (Char.lowercase_ascii c1) (Char.lowercase_ascii c2)
+        with
+        | 0 ->
+          (match Char.compare c1 c2 with
+           | 0 -> aux (i+1)
+           | c -> c)
+        | c -> c
+      else
+        if l1 < l2 then -1
+        else if l1 > l2 then 1
+        else 0
+    in
+    aux 0
+
 end
 
 type warning_printer =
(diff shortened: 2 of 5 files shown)


AltGr:
(neither tested nor benched)

rjbou (approved):

PR #4329, opened 2020-08-24
Provide all functions in the client library

AltGr:
the `main.ml` should really just be one trivial call


AltGr:
(doesn't show well in the GH log, but I basically moved one file and added a new .mli)

rjbou:
rebased

PR #4334, opened 2020-08-27
Fix to output a correct note by changing to the --global flag instead

smorimoto:
This is probably wrong.


smorimoto:
I haven't written changelog yet because I didn't know where to add.

rjbou:
Thanks! You can add it in the *Option* section

rjbou (approved):

smorimoto:
I just updated it!

PR #4341, opened 2020-08-31
A bunch of small optimisations to reduce the startup cost

AltGr:
* make repo/switch loading a bit faster with more caching and better
  cache usage
* avoid loading it altogether when possible for 'opam exec' calls


AltGr:
Appveyor tests passed before I edited the changelog :man_shrugging:

PR #4347, opened 2020-09-10
Add note about Windows

yawaramin:
Currently, there is no mention of Windows on the install page. Add a note saying we are looking for help with Opam on Windows.

### DIFF
--- a/doc/pages/Install.md
+++ b/doc/pages/Install.md
@@ -40,6 +40,8 @@ We provide pre-compiled binaries for:
 - OSX (intel 64 bits)
 (other platforms are available using the other methods below)
 
+Note: We are looking for help packaging Opam for Windows.
+
 If you don't like scripts, you can just pick your download
 [here](https://github.com/ocaml/opam/releases), put it in your PATH as
 `opam`, and set it as executable, e.g.
@@ -175,10 +177,6 @@ for Opam usage.
 
 [![badge](https://repology.org/badge/version-for-repo/ubuntu_20_04/opam.svg)](https://repology.org/project/opam/versions)
 
-#### Windows
-
-We are looking for help packaging Opam on Windows. If you are interested, please get in touch!
-
 ##### Versions 18.04 and newer
 There is a [ppa](https://launchpad.net/~avsm/+archive/ubuntu/ppa) available that contains the current stable version of `opam`.
 ```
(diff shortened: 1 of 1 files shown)


doc/pages/Install.md
```diff
@@ -40,6 +40,8 @@ We provide pre-compiled binaries for:
 - OSX (intel 64 bits)
 (other platforms are available using the other methods below)
 
+Note: We are looking for help packaging Opam for Windows.
+
```
dra27, line 44: We're not "looking" in the sense that no one is doing it at the moment: the core team's goal is merging (in terms of requirements) with https://fdopen.github.io/opam-repository-mingw/ (which is a binary packaging of opam).

Perhaps:
```suggestion
- We do not at present provide an official Windows distribution of opam, but please see [this separately maintained distribution](https://fdopen.github.io/opam-repository-mingw/)
(other platforms are available using the other methods below)

```

doc/pages/Install.md
```diff
@@ -186,6 +188,10 @@ apt install opam
 ##### Versions older than 18.04
 Use the binary distribution. Instructions provided at https://opam.ocaml.org/doc/Install.html#Binary-distribution
 
+#### Windows
+
+We are looking for help packaging Opam on Windows. If you are interested, please get in touch!
+
```
dra27, line 194: I don't think this mention is necessary, or potentially:
```suggestion
#### Windows

Full support for Windows is planned for opam 2.2, and we expect to provide an opam package in [Chocolatey](https://chocolatey.org/) and [winget](https://docs.microsoft.com/en-us/windows/package-manager/). If you'd like to help out, please get in touch!
```

dra27 (commented):
Out of curiosity, is there a discussion which has triggered this patch? There are instructions on https://ocaml.org/docs/install.html for OCaml itself (including opam).

I agree there's no harm in mentioning Windows more explicitly on opam's install page at https://opam.ocaml.org/doc/Install.html, but the premise of this wording isn't quite right: development contributions are welcome, but upstream opam isn't ready for packaging on Windows yet!

yawaramin:
@dra27 thank you for the helpful suggestions! The reason I sent this PR is that I've noticed that newcomers to the ecosystem think that Opam doesn't support Windows because they see no mention of it on the Install page. i hope that mentioning it will encourage people to help out (if possible).

dra27 (approved):
LGTM, thanks!

Issue #4344, opened 2020-09-07
Attempting to install an unreleased package


CraigFe:
Given the following set of Opam files in a local directory:

```markdown
# dependency.opam
opam-version: "2.0"
version: "0.1.0"

# unreleased.opam
opam-version: "2.0"
depends: [ "dependency" {= version} ]
```

Attempting to run a local install will fail:

```
❭ opam install --deps-only ./
The following dependencies couldn't be met:
  - unreleased → dependency < 0.1.0
      no matching version

No solution found, exiting
```

Pinning both packages to a common version such as `dev` fixes the problem (unfortunately, I don't think it's possible to do that in a single `opam` invocation). My questions are:

 - should it be possible to install a local `.opam` file without that file needing to be released in the repository? This is an issue when a new package has recently been added to a monorepo, but no release cycle has occurred yet. Perhaps `opam` should automatically pin such packages to `dev` when it encounters them?

- regardless, the error message for this install is a bit opaque. Could it be improved to something like "No version of `unreleased` is available to install"?

CC: @pascutto

AltGr:
There is no problem with installing an unreleased package; let's first look at your error:
here, you specified in `unreleased.opam` a dependency to `dependency` at the same version, but without specifying the version of `unreleased` itself. Without any specific information to guess the version from, opam silently assumed version `~dev`, so the constraint was resolved to `"dependency" {= "~dev"}`. The version constraint appearing in the error message is unfortunately slightly different (hence the confusion) due to intermediate rewritings, but basically equivalent.

> Pinning both packages to a common version such as dev fixes the problem (unfortunately, I don't think it's possible to do that in a single opam invocation)

`opam pin --with-version` has just been added in #4301 for that purpose :)

> should it be possible to install a local .opam file without that file needing to be released in the repository? This is an issue when a new package has recently been added to a monorepo, but no release cycle has occurred yet. Perhaps opam should automatically pin such packages to dev when it encounters them?

It sure is: just use `opam pin <file>`; you can also use `opam pin --edit` to interactively edit the opam file.

CraigFe:
> `opam pin --with-version` has just been added in #4301 for that purpose :)

Neat, thanks!

> Without any specific information to guess the version from, opam silently assumed version `~dev`, so the constraint was resolved to `"dependency" {= "~dev"}`.

> It sure is: just use opam pin <file>; you can also use opam pin --edit to interactively edit the opam file.

[...]

Issue #4171, opened 2020-05-01
[2.1.0~alpha] Solver failures are too verbose


kit-ty-kate:
Using opam master (https://github.com/ocaml/opam/commit/3def1a6d1ff102aaa608f5e9170783b96f55069d), when opam cannot find a solution to install a package, the breakdown of why the package can't be installed became too verbose compared to previous opam versions.
For instance when trying to install `pgocaml.4.2` on `ocaml-variants.4.11.0+trunk`, the error message used to be the following:
```
The following dependencies couldn't be met:
  - pgocaml -> ppx_deriving >= 4.0 -> ocaml < 4.11.0
      base of this switch (use `--unlock-base' to force)

No solution found, exiting
```
With opam 2.1.0~alpha, the error message became:
```
The following dependencies couldn't be met:
  - pgocaml -> ppx_deriving >= 4.0 -> ocaml < 4.11.0 -> ocaml-base-compiler (!= 3.07+1 & != 3.09.3 & != 3.11.0 & != 4.02.1 & != 4.07.1)
      conflict with the base packages of this switch
  - pgocaml -> ppx_deriving >= 4.0 -> ocaml < 4.11.0 -> ocaml-variants < 4.10.2~
      incompatible with the switch invariant ["ocaml-variants" {= "4.11.0+trunk"}] (use `--update-invariant' to force)
  - ocaml-variants -> ocaml = 4.11.0
Your request can't be satisfied:
  - No available version of ocaml satisfies the constraints
  - No available version of ocaml-variants satisfies the constraints
  - ocaml-variants (!= 4.00.0+debug-runtime & < 4.00.1+annot | >= 4.00.1+french & != 4.01.0+PIC & != 4.01.0+open-types & < 4.02.1+32bit | = 4.02.1+modular-implicits | >= 4.02.1+musl+static & != 4.02.3+buckle-1 & < 4.03.0+32bit |
>= 4.03.0+beta1+flambda & != 4.04.0+32bit & < 4.04.0+afl | = 4.04.0+beta1+flambda | >= 4.04.0+beta2+flambda & < 4.04.0+fPIC | >= 4.04.0+fp & < 4.04.0+spacetime | >= 4.04.1+32bit & != 4.04.1+fp & != 4.04.2+32bit & != 4.04.2+fp &
[...]

Issue #2229, opened 2015-06-29
Wrong formula in conflict message


AltGr:
```
$ mkdir -p /tmp/p; echo 'opam-version: "1.2" depends: "ocp-indent" { >= "2" | < "0.1" }' >/tmp/p/opam
$ opam pin add test-conflicts.1 /tmp/p
[...]
The following dependencies couldn't be met:
  - test-conflicts -> ocp-indent (< 0.1 & >= 2)
Your request can't be satisfied:
  - ocp-indent<0.1 is not available because the package is pinned to /home/lg/ocamlpro/ocp-indent, version 1.5.1.
  - ocp-indent>=2 is not available because the package is pinned to /home/lg/ocamlpro/ocp-indent, version 1.5.1.
```

The message refers to `(< 0.1 & >= 2)` while it was an `|` in the original constraint. Otherwise it's fine (in particular the two reasons following). [This](https://github.com/ocaml/opam/blob/c93132e707258a19281ad67a66495ba65751a080/src/solver/opamCudf.ml#L276-L348) is probably the function in cause ; but it's quite complex already (it gathers all dependency chains returned by Dose in order to aggregate them for a concise message).
It may be difficult to retrieve the original logical operator, as Dose returns a list of chains without connection between them.

kit-ty-kate:
This has been fixed at some point in the past 9 years. As of 2.3.0 opam outputs either:
```
[ERROR] Package conflict!
  * Missing dependency:
    - test-conflicts → ocp-indent >= 2
    no matching version

No solution found, exiting
```
or
```
[ERROR] Package conflict!
  * Missing dependency:
    - test-conflicts → ocp-indent >= 1.8
    not available because the package is pinned to version 1.5.1

No solution found, exiting
```
Depending on the version of ocp-indent installed.

PR #4349, opened 2020-09-10
Rewrite the conflict explanation extraction mechanism

AltGr:
Finally! This needs more testing, so please everyone gather your
extreme conflict messages and send them to me; but this is very
promising.

### DIFF
--- a/tests/reftests/conflict-badversion.test
+++ b/tests/reftests/conflict-badversion.test
@@ -14,31 +14,9 @@ Faking installation of ocaml.4.02.3
 Faking installation of base-ocamlbuild.base
 Done.
 ### opam install core'>'112.17.00 core'<'112.17.00 --show
-The following dependencies couldn't be met:
-  - ocaml = 4.02.3 (from ocaml-base-compiler = 4.02.3)
-  - ocaml-base-compiler (< 3.07+1 | >= 3.08.2 & != 3.09.0 & != 3.09.2 & != 3.11.0 & < 3.12.1) (from core (> 112.17.00 & < 112.17.00 & > 112.17.00 & < 112.17.00 & > 112.17.00 & < 112.17.00 & > 112.17.00 & < 112.17.00 & > 112.17.00 & < 112.17.00 & > 112.17.00 & < 112.17.00 & > 112.17.00 & < 112.17.00 & > 112.17.00 & < 112.17.00) → custom_printf < 111.21.00 → ocaml < 4.02.0)
-      incompatible with the switch invariant ["ocaml-base-compiler" {= "4.02.3"}] (use `--update-invariant' to force)
-  - ocaml-base-compiler (< 3.07+1 | >= 3.08.2 & != 3.09.0 & != 3.09.2 & != 3.11.0 & < 3.12.1) (from core (> 112.17.00 & < 112.17.00 & > 112.17.00 & < 112.17.00 & > 112.17.00 & < 112.17.00 & > 112.17.00 & < 112.17.00 & > 112.17.00 & < 112.17.00 & > 112.17.00 & < 112.17.00 & > 112.17.00 & < 112.17.00 & > 112.17.00 & < 112.17.00) → core_kernel = 111.21.00 → ocaml < 4.02.1)
-      incompatible with the switch invariant ["ocaml-base-compiler" {= "4.02.3"}] (use `--update-invariant' to force)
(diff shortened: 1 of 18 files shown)


kit-ty-kate:
Really amazing! I tried it quickly locally and I have a few comments.

Before:
```
$ opam install extlib-compat
  - ocaml-variants → ocaml = 4.11.0
  - extlib-compat → ocaml < 4.07.0 → ocaml-variants < 4.06.2~
      incompatible with the switch invariant ["ocaml-variants" {= "4.11.0+trunk"}] (use `--update-invariant' to force)
  - extlib-compat → ocaml < 4.07.0 → ocaml-base-compiler (>= 3.07+2 & != 3.09.2 & != 3.10.2 & != 4.01.0 & != 4.02.2 & != 4.04.1 & != 4.06.0 & < 4.07.0)
      conflict with the base packages of this switch
Your request can't be satisfied:
  - No available version of ocaml satisfies the constraints
  - No available version of ocaml-variants satisfies the constraints
  - ocaml-variants (!= 4.00.1+BER & != 4.03.0+beta1+flambda & < 4.03.0+statistical-memprof | >= 4.04.0+BER & != 4.05.0+beta3 & < 4.06.0+32bit) is in conflict with ocaml-base-compiler

No solution found, exiting
```
After:
```
$ opam install extlib-compat
[ERROR] Package conflict!
    - (invariant) → ocaml-variants = 4.11.0+trunk → ocaml = 4.11.0
    - extlib-compat → ocaml < 4.07.0
  * No agreement on the version of ocaml-variants:
    - (invariant) → ocaml-variants = 4.11.0+trunk
    - extlib-compat → ocaml < 4.05.0 → ocaml-variants < 3.09.2~

No solution found, exiting
```
I feel like the second bullet point could be removed for an even shorter conflict message. However I suppose this might be needed for much difficult constraints.

I'm also not sure this is a good idea to tell users about `--update-invariant`. What about:
```
You might want to use another switch or update your current switch invariant.
```
or something like that? I'm not sure what would be the best sentence here but I'm sure we can find something.
In particular I don't think users using the "one version of the compiler per switch" scheme need that. I think most users will continue to use this and maybe some will have a `default` switch, however this one is already vague enough so `--update-invariant` wouldn't be necessary.

I also think this could be said only once as supposed to twice.

kit-ty-kate:
Related issue: https://github.com/ocaml/opam/issues/4171

tests/reftests/conflict-badversion.test
```diff
@@ -14,31 +14,9 @@ Faking installation of ocaml.4.02.3
...
-Your request can't be satisfied:
-  - No available version of core satisfies the constraints
-  - No available version of ocaml satisfies the constraints
-  - No available version of ocaml-base-compiler satisfies the constraints
-  - ocaml-base-compiler (< 3.07+1 | = 3.08.2 | = 3.08.4 | = 3.09.1 | >= 3.10.0 & < 3.10.2 | = 3.11.1 | = 3.12.0 | = 4.02.3) is in conflict with ocaml-variants
-  - ocaml-base-compiler = 4.02.3 is in conflict with ocaml-system
+[ERROR] Package conflict!
+  * No agreement on the version of core:
+    - core != 112.17.00
+    - core != 112.17.00
```
samoht, line 20: what does it mean? :-)

AltGr: The message is off, but there is actually little we can easily do about it :/ ; my opinion is that in this case the error is clear anyway so it's still OK (but yeah, we should at least filter out duplicate lines).

As for the explanation why, this is a different instance of #2229: we are working on a package graph where we can't really distinguish between conjunction and disjunction between edges. In this case, we have `core<112.17.00 & core>112.17.00`, which actually resolves to `false`, but without knowing about the `&` we wrongly reduce it as an `|`, which reduces to `core != 112.17.00`.
I'll check if I can find a way to detect such cases, and skip the "formula reduction" when they happen.

samoht: ha yes I was talking about the duplication -- but you are right, it's a bit hard to suggest something more actionnable to the user here.

AltGr:
> I feel like the second bullet point could be removed for an even shorter conflict message. However I suppose this might be needed for much difficult constraints.

I totally agree... but I couldn't find a generic way to determine that the first message is strictly better and supersedes the second... I already have quite a few heuristics to remove (lots of) spurious or secondary messages, but here the reason why the first is better is not totally clear, especially without resorting to independent knowledge of the package graph. We have a similar case [here](https://github.com/ocaml/opam/pull/4349/files#diff-22e649fc2cd1e978c6d687459fc93646R38-R45), where actually the "missing dependency" is extra information too.

In your case, there is another message issue though: a `ocaml-variants < 3.09.2~` is appearing; this one would require some work, basically when no package version matches a constraint after conversion to cudf, we encode it as `less than min existing version` (this was raised in #4344 as well). I need to check if we can find a way to cure this.

> I'm also not sure this is a good idea to tell users about --update-invariant

Indeed, it might help if you're stuck but may not be the best advice in general. I at least removed the duplication, but am open to removing or changing the message as well (with your suggestion, unless something better comes up ?)

Thanks!

PR #4362, opened 2020-09-15
Fix missing conflict message when trying to remove required packages

AltGr:
We were missing some roots when the chain of dependencies ended up of
length 0!


rjbou (approved):

rjbou:
Thanks!

PR #4371, opened 2020-09-23
Update opam files

AltGr:
They were referring to the `version` variable already anyway, so I
guess there is no point in keeping them locked to the very obsolete
1.2: you'll probably prefer to use the bootstrap scripts if yiy are in
this case anyway?


rjbou (approved):
no objection, lgtm

PR #4376, opened 2020-10-02
Use dune to run reftests

emillon:
The existing `reftests.ml` are responsible for discovering the tests and running them, and a Makefile action performs the diff.

This splits the work a bit more so that it plays better with dune (and becomes compatible with `dune promote`):

- a `gen` executable reads the tests directory and outputs rules to execute tests. In addition, it determines the versions of
  opam-repository that are necessary, and outputs rules to download and unpack them (these will be shared across runs, and between tests that are based on the same version).
- a `run` executable executes a single test with enough context passed as arguments: the path to the development `opam` binary, the test file, and the opam repository contents.
- the generated rules are checked in the repository as `dune.inc` because dune needs to know them statically, but they are checked: if a test file is added, `dune runtest` will compute the new rules and they can be promoted into `dune.inc`.


emillon:
Some remarks:
- the `camelus` test fails. According to @rjbou that's expected, but I'd like to know a bit more so that we determine what's the best course of action: promote the result, fix it, skip it, or make the test more reliable?
- I added a lock so that tests are not executed concurrently for now. Running them concurrently seem to work, but they change the camelus test behavior as well, so I'd like to make sure this is the right thing to do before removing the lock.
- there's a weird behavior with dune in watch mode sometimes, when the test fails it's not restarted when the files change. restarting dune works though. That might be a bug in dune with directory targets, I'd like to investigate before merging in any case.

emillon:
I added some aliases so that the tests can be run individually:

- each test has its own alias (e.g. `dune build @reftest-conflict-camlp4`)
- all tests are attached to `@reftest` (`dune build @reftest` runs all)
- this alias is attached to `@runtest` too (so `dune runtest` will run these)

Would you like a Makefile target in addition @rjbou or are you ok with the dune aliases?

rjbou:
Dune aliases are sufficient. Thanks!

PR #4379, opened 2020-10-06
Add `opam update --depexts` to replace `opam depext -u`

AltGr:
This is not strictly an opam command, but it's useful to have here. Closes

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -1812,6 +1812,11 @@ let update cli =
     mk_flag ~cli cli_original ["development"]
       "Update development packages (skipping repositories unless \
        $(b,--repositories) is also specified)." in
+  let depexts_only =
+    mk_flag ~cli (cli_from cli2_1) ["depexts"]
+      "Request the system package manager to update its databases (skipping \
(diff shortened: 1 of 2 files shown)


src/client/opamCommands.ml
```diff
@@ -1809,6 +1809,11 @@ let update cli =
     mk_flag ["development"]
       "Update development packages (skipping repositories unless \
        $(b,--repositories) is also specified)." in
+  let depexts_only =
+    mk_flag ["depexts"]
+      "Request the system package manager to update its databases (skipping \
+       all opam packages, unless $(b,--development) or $(b,--repositories) is \
+       also specified). This generally requires $(b,sudo) rights." in
```
dra27, line 1816: Just a side-note that this will want gating on the CLI version for >= 2.1

rjbou:
Note: `OpamSysInteract.update` update calls weren't checked nor updated, just imported as is from `depext` plugin.

rjbou (approved):

rjbou:
Queued until CLI versioning PR merged, otherwise, lgtm! thanks

PR #4384, opened 2020-10-08
'opam admin cache': use the configured caches for update

### LINKED ISSUE #4352: Opam admin cache add --secondary-cache flag
This affects opam 2.0.7.

Given that many artifact URLs become obsolete over time, users requiring an offline clone of the opam repository need a way to fetch missing artifacts from other opam repository caches.  

Many source URLs in the opam repository are becoming obsolete.  For example, bitbucket has stopped serving archives of mercurial repositories.  Running `opam admin cache` will never fully populate these entries even though the artifacts are still present on opam.ocaml.org in the cache.

I would like `opam admin cache` to add a flag `--secondary-cache` that names the url of an opam cache directory to use when directly fetching an artifact fails.  This way, artifacts can be located when a URL cannot be resolved or a checksum is wrong.

AltGr:
Caches are taken from the local `repo` file `archive-mirrors` field, then from 
the global opam configuration, if any.

Interestingly, the former (only) was already used by `opam admin add-hashes`, 
but not by `opam admin cache`.

Closes #4352

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -67,7 +67,7 @@ New option/command/subcommand are prefixed with ◈.
   *
 
 ## Admin
-  *
+  * Use the archive caches when running `opam admin cache` [#4384 @AltGr - fix #4352]
 
 ## Opam installer
   *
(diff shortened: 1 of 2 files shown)


rjbou (approved):
Thanks!

PR #4385, opened 2020-10-09
Cli versioning

rjbou:
Implementing CLI versioning, handling of the different CLIs from [spec](https://github.com/ocaml/opam/wiki/Spec-for-opam-CLI-versioning).
It is only handled for subcommands and flags, not principal commands. A note is added in the manpage depending of flag  validity and requested cli. When a not valid flag is used, a message is also printed to the user with cli indication, etc.

_See discussion below for more precisions_

Summary of new `[N]` and deprecated `[D]` options in 2.1 (diff of manpages from 2.0.7 & master).

#### opam commands
  * [x] `lock` [N]
  * [x] `option` [N]

#### opam globals options
  * ~`--cli` [N]~

#### package build options
  * [x] `--assume-depexts` [N]
  * [ ] ~`--locked` (behaviour change)~
  * [x] `--lock-suffix` [N]
  * [x] `--no-depexts` [N]
  * [x] `--unlock-base` [D]
  * [x] `--update-invariant` [N]

#### opam admin add-hashes
  * [x] `--packages` [N]

#### opam config
  * [x] `set` [D]
  * [x] `unset` [D] 
  * [x] `set-global` [D]
  * [x] `unset-global` [D]
  * [x] `var` [D]
  * [x] `exec` [D]

#### opam env
  * [x] `--check` [N]

#### opam info/show
  * [x] `--all-versions` [N]
  * [x] `--just-file` [N]
  * [x] `--sort` [N]
  * [x] `--file` [D]

#### opam install
  * [x] `--check` [N]
  * [x] `--depext-only` [N]
  * [x] `--ignore-conflicts` [N]

#### opam pin
  * [x] `scan` [N]
  * [x] `--normalise` [N]
  * [x] `--with-version` [N]

#### opam switch
  * [x] `invariant` [N]
  * [x] `set-invariant` [N]
  * [x] `set-base` [D]
  * [x] `install` [D]
  * [x] `--force` [N]
  * [x] `--formula` [N]
  * [x] `--freeze` [N]
  * [x] `--no-action` [N]

#### opam unpin
  * [x] `--normalise` [N]
  * [x] `--with-version` [N]

#### opam upgrade
  * [x] `--installed` [N]

#### opam var
  * [x] `--global` [N]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -176,7 +176,7 @@ checker:
 tests: $(DUNE_DEP)
 	$(DUNE) build --profile=$(DUNE_PROFILE) $(DUNE_ARGS) opam.install src/tools/opam_check.exe tests/patcher.exe
 	$(DUNE) build --profile=$(DUNE_PROFILE) $(DUNE_ARGS) doc/man/opam-topics.inc doc/man/opam-admin-topics.inc
-	$(DUNE) runtest --force --no-buffer --profile=$(DUNE_PROFILE) $(DUNE_ARGS) src/ tests/
+	OPAMCLI=2.0 $(DUNE) runtest --force --no-buffer --profile=$(DUNE_PROFILE) $(DUNE_ARGS) src/ tests/
 
 .PHONY: crowbar
 # only run the quickcheck-style tests, not very covering
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -1256,10 +1256,6 @@ let make_command_alias cmd ?(options="") name =
     ~docs:"COMMAND ALIASES" ~sdocs:global_option_section
     ~doc ~man
 
-let term_info title ~doc ~man =
-  let man = man @ help_sections in
-  Term.info ~sdocs:global_option_section ~docs:Manpage.s_commands ~doc ~man title
-
 let arg_list name doc kind =
   let doc = Arg.info ~docv:name ~doc [] in
   Arg.(value & pos_all kind [] & doc)
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -183,6 +183,7 @@ let create_build_options
   }
 
 let apply_build_options b =
+  let open OpamStd.Option.Op in
   let flag f = if f then Some true else None in
   OpamRepositoryConfig.update
     (* ?download_tool:(OpamTypes.arg list * dl_tool_kind) Lazy.t *)
@@ -193,16 +194,16 @@ let apply_build_options b =
     ();
   OpamStateConfig.update
     (* ?root: -- handled globally *)
-    ?jobs:OpamStd.Option.Op.(b.jobs >>| fun j -> lazy j)
+    ?jobs:(b.jobs >>| fun j -> lazy j)
     (* ?dl_jobs:int *)
     (* ?no_base_packages:(flag o.no_base_packages) -- handled globally *)
     ?build_test:(flag b.build_test)
     ?build_doc:(flag b.build_doc)
     ?dryrun:(flag b.dryrun)
-    ?makecmd:OpamStd.Option.Op.(b.make >>| fun m -> lazy m)
+    ?makecmd:(b.make >>| fun m -> lazy m)
     ?ignore_constraints_on:
-      OpamStd.Option.Op.(b.ignore_constraints_on >>|
-                         OpamPackage.Name.Set.of_list)
+      (b.ignore_constraints_on >>|
+       OpamPackage.Name.Set.of_list)
     ?unlock_base:(flag b.unlock_base)
     ?locked:(if b.locked then Some (Some b.lock_suffix) else None)
     ?no_depexts:(flag b.no_depexts)
@@ -800,6 +801,20 @@ let mk_opt_all ?section ?vopt ?(default=[]) flags value doc kind =
 let mk_tristate_opt ?section flags value doc =
   let doc = Arg.info ?docs:section ~docv:value ~doc flags in
   Arg.(value & opt (some (enum when_enum)) None & doc)
+
+let mk_vflag ?section default flags =
+  let flags =
+    List.map (fun (content, flag, doc) ->
+        content, Arg.info ?docs:section flag ~doc) flags
+  in
+  Arg.(value & vflag default flags)
+
+let mk_vflag_all ?section ?(default=[]) flags =
+  let flags =
+    List.map (fun (content, flag, doc) ->
+        content, Arg.info ?docs:section flag ~doc) flags
+  in
+  Arg.(value & vflag_all default flags)
 
 type 'a subcommand = string * 'a * string list * string
 
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -841,6 +841,16 @@ module Mk : sig
     cli:OpamCLIVersion.t -> ?defaults:(string * string) list ->
     'a subcommands -> Manpage.block list
 
+  type command = unit Term.t * Term.info
+
+  val mk_command:
+    OpamCLIVersion.t -> validity -> string -> doc:string ->
+    man:Manpage.block list -> (unit -> unit) Term.t -> command
+
+  val mk_command_ret:
+    OpamCLIVersion.t -> validity -> string -> doc:string ->
+    man:Manpage.block list -> (unit -> unit Term.ret) Term.t -> command
+
 end = struct
 
   let cli2_0 = OpamCLIVersion.of_string "2.0"
@@ -1195,6 +1205,34 @@ end = struct
                     exe usage)
       | _ ->
         `Error (true, Printf.sprintf "Invalid %s subcommand" command)
+
+  (* Commands *)
+
+  let term_info title ~doc ~man =
+    let man = man @ help_sections in
+    Term.info ~sdocs:global_option_section ~docs:Manpage.s_commands ~doc ~man title
+
+  type command = unit Term.t * Term.info
+
+  let mk_command cli validity name ~doc ~man cmd =
+    let doc = update_doc_w_cli doc ~cli validity in
+    let info = term_info name ~doc ~man in
+    let check =
+      check_cli_validity cli validity () [name]
+      |> Term.const
+      |> Term.ret
+    in
+    Term.(cmd $ check), info
+
+  let mk_command_ret cli validity name ~doc ~man cmd =
+    let doc = update_doc_w_cli doc ~cli validity in
+    let info = term_info name ~doc ~man in
+    let check =
+      check_cli_validity cli validity () [name]
+      |> Term.const
+      |> Term.ret
+    in
+    Term.(ret (cmd $ check)), info
 
 end
 
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -778,304 +778,400 @@ let opamlist_columns =
 
 (* CLI version helpers *)
 
-type flag_validity =
-  | Valid_since of OpamCLIVersion.t
-  | Deprecated of OpamCLIVersion.t * OpamCLIVersion.t * string option
-  (* since * removal * use instead *)
+(* Module Mk defines Cmdliner optional argument function-helpers, with the cli
+   version. *)
 
-let cli2_0 = OpamCLIVersion.of_string "2.0"
-let cli2_1 = OpamCLIVersion.of_string "2.1"
-let cli2_3 = OpamCLIVersion.of_string "2.3"
-let cli3_0 = OpamCLIVersion.of_string "3.0"
-let valid_cli_legacy = Valid_since cli2_0
+module Mk : sig
 
-let update_doc_w_cli doc ~cli = function
-  | Valid_since c ->
-    if cli < c then
-      Printf.sprintf "(Since $(b,%s)) %s"
-        (OpamCLIVersion.to_string c) doc
-    else doc
-  | Deprecated (since, removal, instead) ->
-    if cli < since then
-      Printf.sprintf "(Deprecated from $(b,%s), removed at $(b,%s)%s) %s"
+  type validity
+
+  val cli_from: OpamCLIVersion.t -> validity
+  val cli_between:
+    OpamCLIVersion.t -> ?replaced:string -> OpamCLIVersion.t -> validity
+  val cli_original: validity
+
+  val cli2_0: OpamCLIVersion.t
+  val cli2_1: OpamCLIVersion.t
+
+  val mk_flag:
+    cli:OpamCLIVersion.t -> validity -> ?section:string -> string list ->
+    string -> bool Term.t
+
+  val mk_opt:
+    cli:OpamCLIVersion.t -> validity -> ?section:string -> ?vopt:'a ->
+    string list -> string -> string -> 'a Arg.converter -> 'a -> 'a Term.t
+
+  val mk_opt_all:
+    cli:OpamCLIVersion.t -> validity -> ?section:string -> ?vopt:'a ->
+    ?default:'a list -> string list -> string -> string -> 'a Arg.converter ->
+    'a list Term.t
+
+  val mk_vflag:
+    cli:OpamCLIVersion.t -> ?section:string -> 'a ->
+    (validity * 'a * string list * string) list -> 'a Term.t
+
+  val mk_vflag_all:
+    cli:OpamCLIVersion.t -> ?section:string -> ?default:'a list ->
+    (validity * 'a * string list * string) list -> 'a list Term.t
+
+  val mk_tristate_opt:
+    cli:OpamCLIVersion.t -> validity -> ?section:string -> string list ->
+    string -> string -> [> `Always | `Auto | `Never ] option Term.t
+
+  type 'a subcommand = validity * string * 'a * string list * string
+  type 'a subcommands = 'a subcommand list
+
+  val mk_subcommands:
+    cli:OpamCLIVersion.t -> 'a subcommands ->
+    'a option Term.t * string list Term.t
+
+  val mk_subcommands_with_default:
+    cli:OpamCLIVersion.t -> 'a default subcommands ->
+    'a option Term.t * string list Term.t
+
+  val bad_subcommand:
+    cli:OpamCLIVersion.t -> 'a default subcommands ->
+    (string * 'a option * string list) -> 'b Term.ret
+
+  val mk_subdoc :
+    cli:OpamCLIVersion.t -> ?defaults:(string * string) list ->
+    'a subcommands -> Manpage.block list
+
+end = struct
+
+  let cli2_0 = OpamCLIVersion.of_string "2.0"
+  let cli2_1 = OpamCLIVersion.of_string "2.1"
+
+  type 'b validity_and_content = {
+    valid: OpamCLIVersion.t;
+    removed: (OpamCLIVersion.t * string option) option;
+    content: 'b;
+  }
+
+  type 'a contented_validity =
+    [ `Valid of 'a | `Removed of 'a ] validity_and_content
+
+  type validity = unit validity_and_content
+
+  let elem_of_vr = function `Valid e | `Removed e -> e
+
+  let contented_validity (validity:validity) content : 'a contented_validity =
+    match validity.removed with
+    | None -> { validity with content = `Valid content}
+    | Some _ -> { validity with content = `Removed content}
+
+  let cli_from valid = { valid ; removed = None; content = () }
+  let cli_between since ?replaced removal =
+    if since >= removal then
+      OpamConsole.error_and_exit `Internal_error
+        "An option can't be added in %s and removed in %s"
         (OpamCLIVersion.to_string since)
-        (OpamCLIVersion.to_string removal)
-        (match instead with
-         | Some instead ->
-           Printf.sprintf ", see $(i,%s)" instead
-         | None -> "")
-        doc
-    else if cli < removal then
-      Printf.sprintf
-        "Deprecated since $(b,%s), will be removed at $(b,%s)%s"
-        (OpamCLIVersion.to_string since)
-        (OpamCLIVersion.to_string removal)
-        (match instead with
-         | Some instead ->
-           Printf.sprintf ", use $(i,%s) instead." instead
-         | None -> ".")
+        (OpamCLIVersion.to_string removal);
+    { valid = since ; removed = Some (removal, replaced); content = () }
+  let cli_original = cli_from cli2_0
+
+  let string_of_cli_option cli =
+    if cli = cli2_0 then
+      Printf.sprintf "set %s" (OpamConsole.colorise `bold "OPAMCLI=2.0")
     else
-      Printf.sprintf "Removed since $(b,%s)%s"
-        (OpamCLIVersion.to_string removal)
-        (match instead with
-         | Some instead ->
-           Printf.sprintf ", use $(i,%s) instead." instead
-         | None -> ".")
+      Printf.sprintf "use --cli %s"
+        (OpamConsole.colorise `bold (OpamCLIVersion.to_string cli))
 
-let get_long_form flags =
-  List.fold_left (fun (lgth,long) f ->
-      let flgth =  String.length f in
-      if flgth > lgth then (flgth, f) else (lgth, long))
-    (0,"") flags |> snd
+  let update_doc_w_cli doc ~cli = function
+    | { valid = c ; removed = None; _} ->
+      if cli < c then
+        Printf.sprintf "(Since $(b,%s)) %s"
+          (OpamCLIVersion.to_string c) doc
+      else doc
+    | { removed = Some (since, instead); _} ->
+      if cli < since then doc else
+        Printf.sprintf "Removed since $(b,%s)%s"
+          (OpamCLIVersion.to_string since)
+          (match instead with
+           | Some instead ->
+             Printf.sprintf ", use $(i,%s) instead." instead
+           | None -> ".")
 
-let error_new_flag cli valid_since flags =
-  let flag = get_long_form flags in
-  let msg =
-    Printf.sprintf
-      "%s is only available since version %s of the opam CLI, \
-       but version %s has been requested."
-      flag (OpamCLIVersion.to_string valid_since)
-      (OpamCLIVersion.to_string cli)
-  in
-  `Error (false, msg)
+  (* Error messages *)
+  let get_long_form flags =
+    List.fold_left (fun (lgth,long) f ->
+        let flgth =  String.length f in
+        if flgth > lgth then (flgth, f) else (lgth, long))
+      (0,"") flags |> snd
 
-let error_rm_flag cli removal instead flags =
-  let flag = get_long_form flags in
-  let msg =
-    Printf.sprintf
-      "%s was removed in version %s of the opam CLI, \
-       but version %s has been requested.%s"
-      flag (OpamCLIVersion.to_string removal)
-      (OpamCLIVersion.to_string cli)
-      (OpamStd.Option.to_string (fun ist ->
-           Printf.sprintf  " Use %s instead."
-             (OpamConsole.colorise `bold ist))
-          instead)
-  in
-  `Error (false, msg)
-
-let ok_deprecated_flag cli since removal instead flags =
-  let flag = get_long_form flags in
-  OpamConsole.warning
-    "%s is deprecated in %s (since %s), it will be removed in %s.%s"
-    flag
-    (OpamCLIVersion.to_string cli)
-    (OpamCLIVersion.to_string since)
-    (OpamCLIVersion.to_string removal)
-    (OpamStd.Option.to_string (fun ist ->
-         Printf.sprintf  " Use %s instead."
-           (OpamConsole.colorise `bold ist))
-        instead)
-
-(* Cli version check *)
-let cond_new cli c = cli < c
-let cond_removed cli removal =
-  cli >= removal || removal <= OpamCLIVersion.current
-let cond_deprecated cli removal = cli < removal
-
-let check_cli_validity cli validity ?cond elem flags =
-  let cond = OpamStd.Option.default (fun x -> x) cond in
-  match validity with
-  | Valid_since c when cond (cond_new cli c) ->
-    error_new_flag cli c flags
-  | Deprecated (_, removal, instead) when cond (cond_removed cli removal) ->
-    error_rm_flag cli removal instead flags
-  | Deprecated (since, removal, instead)
-    when cond (cond_deprecated cli removal) ->
-    (ok_deprecated_flag cli since removal instead flags;
-     `Ok elem)
-  | _ -> `Ok elem
-
-let term_cli_check ~check arg =
-  Term.(ret ((const check) $ (Arg.value arg)))
-
-(* Arguments constructors *)
-
-let mk_flag ~cli validity ?section flags doc =
-  let doc = update_doc_w_cli doc ~cli validity in
-  let doc = Arg.info ?docs:section ~doc flags in
-  let check elem =
-    check_cli_validity cli validity ~cond:(fun c -> c && elem) elem flags
-  in
-  term_cli_check ~check Arg.(flag & doc)
-
-let mk_opt ~cli validity ?section ?vopt flags value doc kind default =
-  let doc = update_doc_w_cli doc ~cli validity in
-  let doc = Arg.info ?docs:section ~docv:value ~doc flags in
-  let check elem =
-    check_cli_validity cli validity
-      ~cond:(fun c -> c && default != elem) elem flags
-  in
-  term_cli_check ~check Arg.(opt ?vopt kind default & doc)
-
-let mk_opt_all ~cli validity ?section ?vopt ?(default=[])
-    flags value doc kind =
-  let doc = update_doc_w_cli doc ~cli validity in
-  let doc = Arg.info ?docs:section ~docv:value ~doc flags in
-  let check elem =
-    check_cli_validity cli validity
-      ~cond:(fun c -> c && default != elem) elem flags
-  in
-  term_cli_check ~check Arg.(opt_all ?vopt kind default & doc)
-
-let mk_vflag ~cli ?section default flags =
-  let info_flags =
-    List.map (fun (validity, content, flag, doc) ->
-        let doc = update_doc_w_cli doc ~cli validity in
-        content, Arg.info ?docs:section flag ~doc)
-      flags
-  in
-  let check elem =
-    match
-      OpamStd.List.find_opt (fun (_, content, _, _) -> content = elem) flags
-    with
-    | Some (validity, _, flags, _) ->
-      check_cli_validity cli validity elem flags
-    | None -> `Ok elem
-  in
-  term_cli_check ~check Arg.(vflag default info_flags)
-
-let mk_vflag_all ~cli ?section ?(default=[]) flags =
-  let info_flags =
-    List.map (fun (validity, content, flag, doc) ->
-        let doc = update_doc_w_cli doc ~cli validity in
-        content, Arg.info ?docs:section flag ~doc)
-      flags
-  in
-  let check selected =
-    let neww, rm =
-      List.fold_left (fun (neww,rm) elem ->
-          match OpamStd.List.find_opt (fun (_, content, _, _) ->
-              content = elem) flags with
-          | Some (Valid_since c, _, flags, _) when cond_new cli c ->
-            (flags, c)::neww, rm
-          | Some (Deprecated (_, removal, instead), _, flags, _)
-            when cond_removed cli removal ->
-            neww, (flags, (removal, instead))::rm
-          | Some (Deprecated (since, removal, instead), _, flags, _)
-            when cond_deprecated cli removal ->
-            ok_deprecated_flag cli since removal instead flags;
-            neww,rm
-          | _ -> neww,rm) ([],[]) selected
+  let newer_flag_error cli valid_since flags =
+    let flag = get_long_form flags in
+    let msg =
+      Printf.sprintf
+        "%s is only available since version %s of the opam CLI, \
+         but version %s has been requested."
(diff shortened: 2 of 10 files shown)


AltGr:
Wow, pretty heavy infrastructure work... Seems pretty clean given the scope :)
Two minor comments:
- it would be nice to add (ref)tests for the different kinds of arguments (flags, vflags, positional parameters, commands, etc.) and the different cases (below version where it was introduced, deprecated, just fine, etc.) Having the corresponding output in the git log also serves as a good advertisement for the PR ;)
- you may have overlooked the case of a flag that was introduced strictly after 2.0, but is already obsolete or removed; this might happen in the future. Or maybe I just missed something ?

rjbou:
> flag that was introduced strictly after 2.0, but is already obsolete or removed

If a flag is introduced in 3.0, deprecated in 3.1 and removed in 3.2 :
*  using opam 3.0 -> no error
*  using opam > 3.1 -> clear deprecation/error message 
*  using opam > 3.2 with a cli < 3.2 -> clear error message as the flag is no more implemented

Some notes on the implementation:
* opam option are not concerned by this PR (`lock`, `option`)
* In the message, there is no differentiation between a subcommand, an option, etc.
* Format changes are not handled neither

+1 for reftests!!!

src/client/opamArg.ml
```diff
@@ -782,28 +774,263 @@ let opamlist_columns =
...
+type flag_validity =
+  | Valid_since of OpamCLIVersion.t
+  | Deprecated of OpamCLIVersion.t * OpamCLIVersion.t * string option
+  (* since * removal * use instead *)
+
+let cli2_0 = OpamCLIVersion.of_string "2.0"
+let cli2_1 = OpamCLIVersion.of_string "2.1"
+let cli2_3 = OpamCLIVersion.of_string "2.3"
+let cli3_0 = OpamCLIVersion.of_string "3.0"
+let valid_cli_legacy = Valid_since cli2_0
```
rjbou, line 788: These are dropped here, but maybe there is a better place to put them. I thought about directly in `OpamCLIVersion`, but it would be quite more verbose to write/read.

src/client/opamArg.ml
```diff
@@ -782,28 +774,263 @@ let opamlist_columns =
...
+  term_cli_check ~check Arg.(opt (some (enum when_enum)) None & doc)
+
+(*
+let mk_pos ?rev p content default ~docv ~doc =
+Arg.(value & pos ?rev p content default & info ~docv ~doc [])
+
+let mk_pos ~cli ~validity ?rev p content default ~docv ~doc =
+  let doc = update_doc_w_cli doc ~cli validity in
+  Arg.(value & pos ?rev p content default & info ~docv ~doc [])
+*)
```
rjbou, line 1024: I didn't neither treated positional arguments, as they are not prone to change.

src/client/opamArg.mli
```diff
@@ -16,17 +16,54 @@ open Cmdliner
...
+type flag_validity =
+  (* New flag since version [OpamCLIVersion.t] *)
+  | Valid_since of OpamCLIVersion.t
+  (* since * removal * instead
+     Option to remove, deprecated [since], to be removed at [removal],
+     and a hint to use another option or command [instead] *)
+  | Deprecated of
+      OpamCLIVersion.t (* since *)
+      * OpamCLIVersion.t (* removal *)
+      * string option (* use instead *)
```
rjbou, line 30: This naming makes a lot of sense to me, but maybe not for everyone :) If it is not clear, it is possible to renames it. It should be something easy to understand as  it is mandatory from this PR for every flag/option/subcommand handling.

dra27:
Finally looking properly at this - many thanks for taking the heavy lifting of this task list! A few things are not working quite as I envisaged in the spec, but I expect this to be tweaking, rather than a wholesale problem!

The idea with the CLI versioning is that we only deprecate features, rather than commands. So, taking `opam install --unlock-base` / `--update-invariant` as an example. Suppose I have a switch with `["ocaml-base-compiler" {= "4.10.0"}]` for an invariant. The following should happen in each case:

- `opam install ocaml.4.10.1`: advertise `--update-invariant` (this is already correct)
- `OPAMCLI=2.0 opam install ocaml.4.10.1`: advertise `--unlock-base` (i.e. the 2.0 behaviour) (this is a tweak to the previous message based on the CLI version)
- `opam install ocaml.4.10.1 --unlock-base`: should display `[ERROR] --unlock-base was removed in opam 2.1. Use --update-invariant instead, or set OPAMCLI=2.0.` (i.e. the warning presently displayed should be a hard error returning an exit code for a bad command line). Note that for 2.0 removed flag, we advertise setting `OPAMCLI=2.0`, for a 2.1 removed feature in future we'd say `use --cli=2.1`. The point is we didn't "deprecate" `--unlock-base` we have actually removed and the user must change - but, as with Dune variables, etc., we retain a nice message rather than brutally saying "Unrecognised command line option".
- `OPAMCLI=2.0 opam install ocaml.4.10.1 --unlock-base` should proceed without error - at the moment it displays a warning about `--unlock-base` which shouldn't be there. i.e. it's OK to use 2.0 features without warning as long as you specified the version you were after.
- `OPAMCLI=2.0 opam install ocaml.4.10.1 --update-invariant`: `[ERROR] --update-invariant was added in opam 2.1 and is not available for CLI version 2.0`. At present even `--update-invariant` is always display a warning `[WARNING] unlock-base is deprecated in 2.1 (since 2.1), it will be removed in 2.3. Use --update-invariant instead.` The main point here is that new 2.1 command line features should not be available if you selected the 2.0 CLI explicitly.
- `OPAMCLI=2.0 opam install --help` - the help message for `--unlock-base` doesn't need to mention opam 2.3 - the flag is gone in 2.1.
- `opam install --help` - the help for `--unlock-base` should say "Removed in 2.1; use --update-invariant instead". We can choose to remove the help message completely at any point but we don't need to say when. I would say probably that we get rid of flags like that at 3.0.

rjbou:
>   `OPAMCLI=2.0 opam install --help` - the help message for `--unlock-base` doesn't need to mention opam 2.3 - the flag is gone in 2.1.

No mention of removal if help called with 2.0 (can be changed).

>  `OPAMCLI=2.0 opam install ocaml.4.10.1`: advertise `--unlock-base` (i.e. the 2.0 behaviour) (this is a tweak to the previous message based on the CLI version)

As it is a message in the code, and not in option management, a specific behavior need to be called when the message is displayed. The problem here is that we don't have the information about the requested cli version in all modules, only on opam-client & those depending on ones.

rjbou:
There is some options that was removed in 2.0 betas and reintroduced in 2.0 (cf. #3685, for non erroneous message). In cli versioning it means that it was introduced in 2.0 and removed in 2.0, that errors. It is possible to remove them completely in 2.1, but some behaviour can't be reintroduced as it was not even present in 2.0. It concerns `no-aspcud`, `alias-of` and `no-autoinstall`.

src/client/opamArg.ml
```diff
@@ -782,62 +774,414 @@ let opamlist_columns =
...
+        (OpamCLIVersion.to_string since)
+        (OpamCLIVersion.to_string removal);
+    { valid = since ; removed = Some (removal, replaced); content = `Nothing }
+  let cli_legacy = introduced_in cli2_0
+
+  let string_of_cli_option cli =
+    if cli = cli2_0 then
+      Printf.sprintf "set %s" (OpamConsole.colorise `bold "OPAMCLI=2.0")
+    else
+      Printf.sprintf "use --cli %s"
```
dra27, line 876:
```suggestion
      Printf.sprintf "use --cli=%s"
```
(my personal preference is to use the `=` for long forms!)

src/client/opamArg.ml
```diff
@@ -782,62 +774,414 @@ let opamlist_columns =
...
+
+  let update_doc_w_cli doc ~cli = function
+    | { valid = c ; removed = None; _} ->
+      if cli < c then
+        Printf.sprintf "(Since $(b,%s)) %s"
+          (OpamCLIVersion.to_string c) doc
+      else doc
+    | { removed = Some (since, instead); _} ->
+      if cli < since then doc else
+        Printf.sprintf "Removed since $(b,%s)%s"
```
dra27, line 887: I think
```suggestion
        Printf.sprintf "Removed in $(b,%s)%s"
```
reads better, given that it's not the word deprecated. It's removed from the CLI, even if the support to do it hasn't been removed from opam.

src/client/opamArg.ml
```diff
@@ -782,62 +774,414 @@ let opamlist_columns =
...
+        let flgth =  String.length f in
+        if flgth > lgth then (flgth, f) else (lgth, long))
+      (0,"") flags |> snd
+
+  let error_new_flag cli valid_since flags =
+    let flag = get_long_form flags in
+    let msg =
+      Printf.sprintf
+        "%s is only available since version %s of the opam CLI, \
+         but version %s has been requested."
```
dra27, line 905: Again, possible different wording (which I think is similar to Dune's):
```suggestion
        "%s was added in version %s of the opam CLI, \
         but version %s has been requested, which is older."
```

src/client/opamArg.ml
```diff
@@ -782,62 +774,414 @@ let opamlist_columns =
...
+  let mk_vflag_all ~cli ?section ?(default=[]) flags =
+    let flags = List.map (fun (v,c,f,d) -> contented_validity v c, f, d) flags in
+    let info_flags =
+      List.map (fun (validity, flag, doc) ->
+          let doc = update_doc_w_cli doc ~cli validity in
+          validity.content, Arg.info ?docs:section flag ~doc)
+        flags
+    in
+    let check selected =
+      let neww, rm =
```
dra27, line 1005: Possible `neww` → `newer_cli` and `rm` → `older_cli`?

src/client/opamArg.ml
```diff
@@ -782,62 +774,414 @@ let opamlist_columns =
...
+             Printf.sprintf ", use $(i,%s) instead." instead
+           | None -> ".")
+
+  let get_long_form flags =
+    List.fold_left (fun (lgth,long) f ->
+        let flgth =  String.length f in
+        if flgth > lgth then (flgth, f) else (lgth, long))
+      (0,"") flags |> snd
+
+  let error_new_flag cli valid_since flags =
```
dra27, line 900: Possibly `error_newer_flag` (i.e. that the flag is newer than the request CLI)

src/client/opamArg.ml
```diff
@@ -782,62 +774,414 @@ let opamlist_columns =
...
+    let msg =
+      Printf.sprintf
+        "%s is only available since version %s of the opam CLI, \
+         but version %s has been requested."
+        flag (OpamCLIVersion.to_string valid_since)
+        (OpamCLIVersion.to_string cli)
+    in
+    `Error (false, msg)
+
+  let error_rm_flag cli removal instead flags =
```
dra27, line 911: As with the other, possibly `error_older_flag` - i.e. it's older than the requested CLI, not that it's removed.

src/client/opamArg.ml
```diff
@@ -782,62 +774,414 @@ let opamlist_columns =
...
+        let options, clis = List.split neww in
+        let msg =
+          Printf.sprintf
+            "%s are only available since respectively versions %s of the opam \
+             CLI, but version %s has been requested."
+            (OpamStd.Format.pretty_list (List.map get_long_form options) )
+            (OpamStd.Format.pretty_list
+               (List.map OpamCLIVersion.to_string clis))
+            (OpamCLIVersion.to_string cli)
+        in
```
dra27, line 1033: I'd alter this message to look for the maximum CLI and instead display:
```
            "%s can only be used with at least version %s of the opam \
             CLI, but version %s has been requested."
```
i.e. tell the user what they need for this specific combination, rather than for each individual flag (especially as if we display "respectively", it's actually nicer to group them by version, which starts to become a huge function for an error message!)

src/client/opamArg.ml
```diff
@@ -782,62 +774,414 @@ let opamlist_columns =
...
+        `Error (false,msg)
+      | [], _::_->
+        let options, clis = List.split rm in
+        let clis = List.split clis |> fst in
+        let msg =
+          Printf.sprintf
+            "%s were removed in, respectively, %s versions of the opam CLI, \
+             but version %s has been requested."
```
dra27, line 1041: Likewise here,

```
            "%s were all removed by version %s of the opam CLI, \
             but version %s has been requested."
```
(this one _might_ be worth special-casing to display "were all _in_ version" if they all have the same version)

src/client/opamArg.ml
```diff
@@ -782,62 +774,414 @@ let opamlist_columns =
...
+             CLI,and %s were removed in respectively versions %s, \
+             but version %s has been requested."
+            (OpamStd.Format.pretty_list (List.map get_long_form news))
+            (OpamStd.Format.pretty_list
+               (List.map OpamCLIVersion.to_string nclis))
+            (OpamStd.Format.pretty_list (List.map get_long_form rms))
+            (OpamStd.Format.pretty_list
+               (List.map OpamCLIVersion.to_string rclis))
+            (OpamCLIVersion.to_string cli)
+        in
```
dra27, line 1064: I think this needs to be a little more complicated (sadly!): if _any_ of the options don't exist any more (i.e. have no alternate flag name) then this CLI isn't valid in any version and the message is:

```
            "This combination of options is not possible: %s require
             at least version %s of the opam CLI but %s were all \
             removed by version %s of the opam CLI!"
```
(possibly with the same "removed by"/"removed in" distinction, if you wanted). I wouldn't mention the requested version of the CLI - in this case, there's no version.

If all of `rms` were in fact renamed, then the message is:
```
            "This combination of options is not possible: %s require
             at least version %s of the opam CLI and the newer %s \
             flags must be used for %s respectively!"
```
where the last parameters are the list of old options and the list they were renamed (which I think is in `snd rclis`?)

src/client/opamArg.ml
dra27: Agreed!

src/client/opamArg.ml
```diff
@@ -782,62 +774,414 @@ let opamlist_columns =
...
+                    exe usage)
+        else
+          `Error (false, Printf.sprintf "%s: Too many arguments.\nUsage: %s\n"
+                    exe usage)
+      | _ ->
+        `Error (true, Printf.sprintf "Invalid %s subcommand" command)
+
+end
+
+include Mk
```
dra27, line 1184: Does it need to be a module, therefore?

src/client/opamArg.ml
```diff
@@ -894,39 +1214,36 @@ let nonempty_arg_list name doc kind =
   Arg.(non_empty & pos_all kind [] & doc)
 
 (* Common flags *)
-let print_short_flag =
-  mk_flag ["s";"short"]
+let print_short_flag ?(validity=cli_legacy) cli =
```
dra27, line 1217: It makes the diff even bigger, but I don't think `validity` should be optional - it runs a risk if `--short` is added elsewhere that ~validity gets forgotten and suddenly a new flag claims to be a 2.0 flag.

src/client/opamArg.ml
```diff
@@ -894,39 +1214,36 @@ let nonempty_arg_list name doc kind =
...
+  mk_flag ~cli validity ["s";"short"]
     "Output raw lists of names, one per line, skipping any details."
 
-let installed_roots_flag =
-  mk_flag ["installed-roots"] "Display only the installed roots."
-
-let shell_opt =
+let shell_opt ?(validity=cli_legacy) cli =
```
dra27, line 1221: Similarly here

src/client/opamArg.ml
```diff
@@ -1027,35 +1344,34 @@ let global_options =
...
-    mk_opt ~section ["root"]
+    mk_opt ~cli cli_legacy ~section ["root"]
       "ROOT" "Use $(docv) as the current root path. \
               This is equivalent to setting $(b,\\$OPAMROOT) to $(i,ROOT)."
       Arg.(some dirname) None in
-  let d_no_aspcud =
-    mk_flag ~section ["no-aspcud"]
-    "Deprecated."
+  let d_no_aspcud = Term.pure false
+(*     mk_flag ~cli (removed_since cli2_0 cli2_0) ~section ["no-aspcud"] "" *)
```
dra27, line 1366: I think this can be restored just to display the deprecation message in 2.0!

src/client/opamArg.ml
```diff
@@ -1279,16 +1601,16 @@ let assume_built =
...
-  mk_flag ["recursive"]
+  mk_flag ~cli cli_legacy ["recursive"]
     "Allow recursive lookups of (b,*.opam) files. Cf. $(i,--subpath) also."
 *)
 
-let subpath = Term.const None
+let subpath _cli = Term.const None
 (*
-  mk_opt ["subpath"] "PATH"
+  mk_opt ~cli cli_legacy ["subpath"] "PATH"
```
dra27, line 1613: Is it worth an extra note that this isn't `cli_legacy` really?

src/client/opamArg.ml
dra27: The alternative for all of these would be for the default to be `OpamCLIVersion.current` - although I still think I'd prefer it to be explicit everywhere, I think we're more likely to notice that an option has _disappeared_ in an older syntax than that it's magically _appeared_!

src/client/opamCommands.ml
```diff
@@ -2404,65 +2408,61 @@ let switch cli =
...
-  let d_alias_of =
-    mk_opt ["A";"alias-of"]
-      "COMP"
-      "This option is deprecated."
-      Arg.(some string) None
+  let d_alias_of = Term.pure None
+(*
+    mk_opt ~cli (removed_since cli2_0 cli2_0 ~replaced:"opam switch <switch-name> <compiler>")
+      ["A";"alias-of"] "COMP" "" Arg.(some string) None
+*)
```
dra27, line 2456: Similarly to `--no-aspcud`, this can be restored just to make the 2.0 behaviour of a warning consistent!

src/client/opamCommands.ml
```diff
@@ -2404,65 +2408,61 @@ let switch cli =
...
   in
-  let d_no_autoinstall =
-    mk_flag ["no-autoinstall"]
-      "This option is deprecated."
+  let d_no_autoinstall = Term.pure false
+(*     mk_flag ~cli (removed_since cli2_0 cli2_0) ["no-autoinstall"] "" *)
```
dra27, line 2459: and here

dra27 (commented):
This looks basically wonderful - some notes and comments on the API, but there's nothing which looks major:

* [ ] For `--no-aspcud`, `--just-file` (old behaviour), `--no-autoinstall`, `--alias-of`, the behaviour has changed from "ignore the flag" to "syntax error", so I'd keep `OpamArg.deprecated_option` for 2.0 so that it's ignored in OPAMCLI=2.0 (for compatibility) and is an error in opam 2.1+
* [ ] I don't follow the need for the polymorphic variant on the content field in `OpamArg.Mk`, unless there's a weird interaction with Cmdliner that I'm not yet seeing?
* [ ] `OPAMCLI=2.0 opam option` should display an error, `OPAMCLI opam option --global` does, I guess it's just some threading of cli version for the default behaviour of a new command?
* [x] Given that all the uses of, say, `OpamArg.print_short_flag` have had to be updated to include `cli`, I'd make `validity` non-optional just to be really sure that if it gets added elsewhere that adding the version isn't forgotten.
* [ ] It's not this PR which has done it, but we've lost the description for the opam admin command somewhere along the line:
```
COMMANDS
       admin


       clean
           Cleans up opam caches
```

src/client/opamArg.mli
```diff
@@ -37,22 +80,24 @@ val escape_path: string -> string
 (** {2 Flags} *)
 
 (** --short *)
-val print_short_flag: bool Term.t
-
-(** --installed-root *)
-val installed_roots_flag: bool Term.t
+val print_short_flag:
+  ?validity:validity -> OpamCLIVersion.t -> bool Term.t
```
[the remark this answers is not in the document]
rjbou: It could... Or removing from all functions. There is no confusing other `OpamCLIVersion.t` parameter that needs labeling.

src/client/opamArg.ml
```diff
@@ -782,62 +774,414 @@ let opamlist_columns =
...
+
+  type 'b validity_and_content = {
+    valid: OpamCLIVersion.t;
+    removed: (OpamCLIVersion.t * string option) option;
+    content: 'b;
+  }
+
+
+  type 'a contented_validity =
+    [ `Valid of 'a | `Removed of 'a ] validity_and_content
```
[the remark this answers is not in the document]
rjbou: cf. answer below

src/client/opamArg.ml
```diff
@@ -782,62 +774,414 @@ let opamlist_columns =
...
+
+  type validity = [ `Nothing ] validity_and_content
+
+  let elem_of_vr = function `Valid e | `Removed e -> e
+
+  let contented_validity (validity:validity) content : 'a contented_validity =
+    match validity.removed with
+    | None -> { validity with content = `Valid content}
+    | Some _ -> { validity with content = `Removed content}
```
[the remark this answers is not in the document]
rjbou: They are used in inner check, like in `mk_vflag`, to retrieve the good element. There is the necessity to dissociate a valid from a removed content.
In
```
    let check elem =
      match
        OpamStd.List.find_opt (fun (validity, _, _) ->
            validity.content = elem) 
          flags
      with
      ...
```
It permit to retrieve the good element validity to check cli against. For example, for `switch`, both `"create"` and `"install` return an `'install`, and there is now way to know which one is given via cli as `cmdliner` returns an `'install`. Here, `"create"` returns a `Valid 'install` and `"install"` a `Removed 'install`.

src/client/opamArg.ml
rjbou: (UI, and good error message is always a long part :))

src/client/opamArg.ml
rjbou: It is not programmatically needed. I wanted to enclose all internal cli mechanism into the module, in order to forbid use of validity types and functions, and access them only via exposed ones `mk_*` functions.

src/client/opamArg.ml
rjbou: I kept it optional to simplify changes, but yes, it can introduce errors.

src/client/opamArg.ml
rjbou: cf. above answer

src/client/opamArg.mli
```diff
@@ -241,21 +290,16 @@ val make_command_alias:
...
+  cli:OpamCLIVersion.t ->
   ?defaults:(string * string) list -> 'a subcommands -> Manpage.block list
 (** [mk_subdoc cmds] is the documentation block for [cmds]. *)
 
-(** {2 Misc} *)
-
-val deprecated_option: 'a -> 'a -> string -> string option -> unit
-(** [deprecated_option option default name instead] displays a message if
-    [option] if set to its non [default] value. [instead], if present, is the new
-    option/command to launch *)
```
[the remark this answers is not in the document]
rjbou: Replaced by simple messages, no need of the deprecated deprecation function

Makefile
```diff
@@ -168,9 +168,9 @@ checker:
 
 .PHONY: tests tests-local tests-git
 tests: $(DUNE_DEP)
-	$(DUNE) build --profile=$(DUNE_PROFILE) $(DUNE_ARGS) opam.install src/tools/opam_check.exe tests/patcher.exe
-	$(DUNE) build --profile=$(DUNE_PROFILE) $(DUNE_ARGS) doc/man/opam-topics.inc doc/man/opam-admin-topics.inc
-	$(DUNE) runtest --force --no-buffer --profile=$(DUNE_PROFILE) $(DUNE_ARGS) src/ tests/
+	OPAMCLI=2.0 $(DUNE) build --profile=$(DUNE_PROFILE) $(DUNE_ARGS) opam.install src/tools/opam_check.exe tests/patcher.exe
+	OPAMCLI=2.0 $(DUNE) build --profile=$(DUNE_PROFILE) $(DUNE_ARGS) doc/man/opam-topics.inc doc/man/opam-admin-topics.inc
+	OPAMCLI=2.0 $(DUNE) runtest --force --no-buffer --profile=$(DUNE_PROFILE) $(DUNE_ARGS) src/ tests/
```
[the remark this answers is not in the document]
dra27: You've marked as resolved but not answered my question! 🙂

dra27:
Are there still some un-pushed commits? You've ticked "--no-aspcud, --just-file (old behaviour), --no-autoinstall, --alias-of", for example, but those CLI options aren't yet restored? Similarly for `?validity` changing to `~validity`.

rjbou:
All comments that are marked as resolved but not present in the PR. I'll post a response to the main post when I'll finish. The pushed commits are only from _suggestions_ comments.

Makefile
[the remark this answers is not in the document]
rjbou: I know :) It's because the answer (and modifications) is in unpushed commits (at that time).

rjbou:
Thanks for the detailed review!
So so so...  On resolved comments : some are closed with no comment (so it was simply integrated), or with a comment (integrated and commented). remaining one is a discussion.
Some global answers

> * For `--no-aspcud`, `--just-file` (old behaviour), `--no-autoinstall`, `--alias-of`, the behaviour has changed from "ignore the flag" to "syntax error", so I'd keep `OpamArg.deprecated_option` for 2.0 so that it's ignored in OPAMCLI=2.0 (for compatibility) and is an error in opam 2.1+

I kept just messaghes printing to not reinttroduce the function. For `--file` I reimplemented a consistent (imho) old behaviro, so it is handled by the new system

> I don't follow the need for the polymorphic variant on the content field in `OpamArg.Mk`, unless there's a weird interaction with Cmdliner that I'm not yet seeing?

In fact, there is no need of a polymorphic variant, a simple variant is sufficient. I'm not against having a variant instead of polymoprhic variant, it's just a new type to declare.
On the need of specifying `Valid` or `Remove`, cf. https://github.com/ocaml/opam/pull/4385#discussion_r530614862

> `OPAMCLI=2.0 opam option` should display an error, `OPAMCLI opam option --global` does, I guess it's just some threading of cli version for the default behaviour of a new command?

Not yet implemented, it needs another mechanism (this one handles only options & flags). But with all this all done, it should be not so hard to add.

> Given that all the uses of, say, `OpamArg.print_short_flag` have had to be updated to include `cli`, I'd make `validity` non-optional just to be really sure that if it gets added elsewhere that adding the version isn't forgotten.

Agree on it, updated accordingly.

> It's not this PR which has done it, but we've lost the description for the opam admin command somewhere along the line

It was removed with changes in 97c22520

rjbou:
Commands added!

src/client/opamArg.ml
```diff
@@ -782,62 +779,465 @@ let opamlist_columns =
...
+  type validity = unit validity_and_content
+
+  let elem_of_vr = function `Valid e | `Removed e -> e
+
+  let contented_validity (validity:validity) content : 'a contented_validity =
+    match validity.removed with
+    | None -> { validity with content = `Valid content}
+    | Some _ -> { validity with content = `Removed content}
+
+  let cli_from valid = { valid ; removed = None; content = () }
```
rjbou, line 878: This can also be implemented for a more simple use as
```
val cli_from: string -> valitidy
let cli_from version = 
  try
    let valid = OpamCLIVersion.of_string version in
    if OpamCLIVersion.is_supported valid then
      { valid; removed = None; content = () }
    else
      OpamConsole.error_and_exit `Internal_error "not supported"
  with Failure _ -> OpamConsole.error_and_exit `Internal_error "not a valid cli"
```
but it will require to define newer version as supported (for example when we know that we want to deprecate an option in a future opam cli version), and we loose a strong check at compile time replaced by a runtime failure.

dra27:
> > It's not this PR which has done it, but we've lost the description for the opam admin command somewhere along the line
> 
> It was removed with changes in [97c2252](https://github.com/ocaml/opam/commit/97c225207b545021434f509e0d64c1306aecac49)

Oops - duly amending from "~~we've~~ lost the description" to "**I** lost the description" 😊

src/client/opamArg.mli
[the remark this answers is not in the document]
dra27: Shucks! (it does indeed look fine without a deprecated deprecation function)

src/client/opamArg.ml
```diff
@@ -38,22 +38,18 @@ type global_options = {
...
     git_version debug debug_level verbose quiet color opt_switch yes strict
     opt_root external_solver use_internal_solver
     cudf_file solver_preferences best_effort safe_mode json no_auto_upgrade
     working_dir ignore_pin_depends
     d_no_aspcud _ =
-  deprecated_option d_no_aspcud false "no-aspcud" None;
+  if d_no_aspcud then
+    OpamConsole.warning
+      "Option %s is deprecated, ignoring it. \
+       Use instead 'opam switch <switch-name> <compiler>'"
```
dra27, line 51: This message isn't correct! There was no extra message before.

src/client/opamArg.ml
```diff
@@ -1027,35 +1396,34 @@ let global_options =
...
   let root =
-    mk_opt ~section ["root"]
+    mk_opt ~cli cli_original ~section ["root"]
       "ROOT" "Use $(docv) as the current root path. \
               This is equivalent to setting $(b,\\$OPAMROOT) to $(i,ROOT)."
       Arg.(some dirname) None in
   let d_no_aspcud =
-    mk_flag ~section ["no-aspcud"]
-    "Deprecated."
+    mk_flag ~cli (cli_between cli2_0 cli2_2) ~section ["no-aspcud"] "Deprecated"
```
dra27, line 1418:
```suggestion
    mk_flag ~cli (cli_between cli2_0 cli2_1) ~section ["no-aspcud"] "Deprecated"
```

src/client/opamCommands.ml
```diff
@@ -2406,65 +2407,67 @@ let switch cli =
...
        command."
   in
   (* Deprecated options *)
   let d_alias_of =
-    mk_opt ["A";"alias-of"]
-      "COMP"
-      "This option is deprecated."
-      Arg.(some string) None
+    mk_opt ~cli (cli_between cli2_1 cli2_2)
+      ["A";"alias-of"] "COMP" "Deprecated" Arg.(some string) None
```
dra27, line 2453:
```suggestion
    mk_opt ~cli (cli_between cli2_0 cli2_1)
      ["A";"alias-of"] "COMP" "Deprecated." Arg.(some string) None
```

src/client/opamCommands.ml
```diff
@@ -2406,65 +2407,67 @@ let switch cli =
...
   in
   let d_no_autoinstall =
-    mk_flag ["no-autoinstall"]
-      "This option is deprecated."
+      mk_flag ~cli (cli_between cli2_1 cli2_2) ["no-autoinstall"] "Deprecated"
```
dra27, line 2456:
```suggestion
      mk_flag ~cli (cli_between cli2_0 cli2_1) ["no-autoinstall"] "Deprecated."
```

src/client/opamArg.mli
```diff
@@ -16,17 +16,64 @@ open Cmdliner
...
 
-val mk_flag: ?section:string -> string list -> string -> bool Term.t
+(** {3 CLI versioning} *)
+
+(* Type of the validity of a flag *)
+type validity
+
+val cli2_0: OpamCLIVersion.t
+val cli2_1: OpamCLIVersion.t
+val cli2_2: OpamCLIVersion.t
```
dra27, line 26:
```suggestion
```
This appeared for some of the options which were deprecated in opam 2.0, but I don't think it's needed at all?

src/client/opamCommands.ml
```diff
@@ -705,55 +704,57 @@ let show cli =
...
+    mk_flag ~cli cli_original ["list-files"]
       "List the files installed by the package. Equivalent to \
        $(b,--field=installed-files), and only available for installed \
        packages"
   in
   let file =
-    mk_opt ["file"] "FILE" "DEPRECATED: see $(i,--just-file)"
-      Arg.(some existing_filename_or_dash) None
+    mk_opt ~cli (cli_between cli2_0 cli2_1 ~replaced:"--just-file")
+      ["file"] "FILE" ""
```
dra27, line 724:
```suggestion
      ["file"] "FILE" "DEPRECATED: use an explicit path argument as package instead. \
              Get package information from the given FILE instead of from \
              known packages. This implies $(b,--raw) unless $(b,--fields) is \
              used. Only raw opam-file fields can be queried."
```
(lifted out of #3729)

src/client/opamCommands.ml
```diff
@@ -3789,6 +3793,8 @@ let admin =
   Term.(ret (const (`Error (true, doc)))),
   Term.info "admin"
```
dra27, line 3794: Since you happen to have a diff nearby, we can correct my sins from #4316 here:

```suggestion
  Term.info "admin" ~doc:OpamAdminCommand.admin_command_doc
```

src/client/opamArg.ml
```diff
@@ -782,62 +779,465 @@ let opamlist_columns =
...
+  type validity
+
+  val cli_from: OpamCLIVersion.t -> validity
+  val cli_between:
+    OpamCLIVersion.t -> ?replaced:string -> OpamCLIVersion.t -> validity
+  val cli_original: validity
+
+  val cli2_0: OpamCLIVersion.t
+  val cli2_1: OpamCLIVersion.t
+  val cli2_2: OpamCLIVersion.t
```
dra27, line 798:
```suggestion
```
(I don't think this is needed?)

src/client/opamArg.ml
```diff
@@ -782,62 +779,465 @@ let opamlist_columns =
...
+
+  val mk_command_ret:
+    OpamCLIVersion.t -> validity -> string -> doc:string ->
+    man:Manpage.block list -> (unit -> unit Term.ret) Term.t -> command
+
+end = struct
+
+  let cli2_0 = OpamCLIVersion.of_string "2.0"
+  let cli2_1 = OpamCLIVersion.of_string "2.1"
+  let cli2_2 = OpamCLIVersion.of_string "2.2"
```
dra27, line 858:
```suggestion
```
(see above, and below!)

dra27 (commented):
A few minor nits (all as code suggestions). Why did the restored `--no-aspcud` (`opam-install`) and `--alias-of`/`--no-autoinstall` (`opam-switch`) end up referencing the 2.2 cli? I was expecting for those to be CLI errors in 2.1. It's just 2.0 that wants to display the message that they're being ignored (if, for example, there are still scripts using `--no-ascpud` - it's OK that they need to switch to using OPAMCLI=2.0, but that's the only change 2.0 scripts should have to have added).

The implementation for subcommands looks great (and is working as I anticipated too), thanks.

A few doc string updates, but otherwise I think this is ready to go in - thanks again for the considerable amount of work on it!

src/client/opamArg.ml
rjbou: There was, in the [`deprecated_option`](https://github.com/ocaml/opam/pull/4385/files#diff-e674376a5ba688da9b6f11571be146af2e7bbde7c583630c6e8d4fc15c575035L41) function

src/client/opamArg.ml
rjbou: see below

src/client/opamArg.ml
rjbou: see below

src/client/opamCommands.ml
rjbou:
```suggestion
      ["file"] "FILE" "Get package information from the given FILE instead of from \
              known packages. This implies $(b,--raw) unless $(b,--fields) is \
              used. Only raw opam-file fields can be queried."
```

src/client/opamCommands.ml
rjbou: Those option were removed in opam 2.0 and reintroduced with a deprecation notice in 2.1. (cf #3685). We could set theme as available in cli2.0 and removed in 2.1, from the cli point of view, but it is not correct according the real code, where the option results on an error with opam 2.0.x series, and a deprecation message since the 2.1.0~alpha.

src/client/opamCommands.ml
rjbou: see above

src/client/opamArg.ml
dra27: There was an optional parameter, but it wasn't used for _this_ option. opam 2.0 just gives:

```
[WARNING] Option no-aspcud is deprecated, ignoring it.
```

rjbou: oups, bad copy-pasting. I thought you were talking globally, not for this message particularly.

src/client/opamCommands.ml
dra27: Those three options were reintroduced in 2.0.2 (see https://github.com/ocaml/opam/pull/3669). So while 2.0.0 and 2.0.1 don't have them, the latest service release of 2.0 does. I think it's OK to change that behaviour to the better "gone in 2.1" (it's a beta rather than an RC)

src/client/opamArg.ml
rjbou, line 51:
```suggestion
    OpamConsole.warning
      "Option %s is deprecated, ignoring it.'"
```

src/client/opamCommands.ml
rjbou: ok, lgtm

src/client/opamCommands.ml
rjbou: The deprecation/removal notice will be printed by cli versioning mechanism.  Do we need another deprecated notice in its doc ?

dra27: The message I'd put in there is the one displayed for `OPAMCLI=2.0` which simply matches what opam 2.0 displays at the moment

rjbou: oh ok

dra27:
The AppVeyor failure can be ignored - the Ubuntu test failure appears to be because of... CLI versioning 😊

Apart from that, I think is ready to go in?

PR #4387, opened 2020-10-11
Provide opam during build

lefessan:
With the latest version of `opam`, `opam-bin` fails because an internal call (to `opam show` to get the `opam` file corresponding to the built package) fails because of the global lock (it happens only when creating a switch).

This PR makes `opam` create an `opam` description file in the build directory called `$OPAMROOT/$SWITCH/.opam-switch/build/$PACKAGE.opam` that can be used by hooks, and `opam-bin` in particular, avoiding the recursive call to `opam`.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -17,6 +17,7 @@ New option/command/subcommand are prefixed with ◈.
 
 ## Install
   * The stdout of `pre-` and `post-session` hooks is now propagated to the user [#4382 @AltGr - fix #4359]
+  * A PACKAGE.opam file is provided during build in the build/ directory [@lefessan]
 
 ## Remove
   *
--- a/master_changes.md
+++ b/master_changes.md
@@ -17,7 +17,7 @@ New option/command/subcommand are prefixed with ◈.
 
 ## Install
   * The stdout of `pre-` and `post-session` hooks is now propagated to the user [#4382 @AltGr - fix #4359]
-  * A PACKAGE.opam file is provided during build in the build/ directory [@lefessan]
+  * A `PACKAGE.opam` file is provided during build in the `build/` directory [#4387 @lefessan]
 
 ## Remove
   *
--- a/src/client/opamAction.ml
+++ b/src/client/opamAction.ml
@@ -398,6 +398,9 @@ let prepare_package_source st nv dir =
                  (OpamStd.Format.itemize OpamFilename.to_string bad_hash)));
   in
   OpamFilename.mkdir dir;
+  OpamFile.OPAM.write
+    (OpamFile.make (OpamPath.Switch.build_opam ~build_dir:dir))
+    opam;
   get_extra_sources_job @@+ function Some _ as err -> Done err | None ->
     check_extra_files |> function Some _ as err -> Done err | None ->
       let opam = OpamSwitchState.opam st nv in
--- a/src/client/opamAction.ml
+++ b/src/client/opamAction.ml
@@ -399,7 +399,7 @@ let prepare_package_source st nv dir =
   in
   OpamFilename.mkdir dir;
   OpamFile.OPAM.write
-    (OpamFile.make (OpamPath.Switch.build_opam ~build_dir:dir))
+    (OpamPath.Switch.build_opam st.switch_global.root st.switch nv)
     opam;
   get_extra_sources_job @@+ function Some _ as err -> Done err | None ->
     check_extra_files |> function Some _ as err -> Done err | None ->
--- a/src/client/opamSolution.ml
+++ b/src/client/opamSolution.ml
@@ -718,8 +718,10 @@ let parallel_apply t ~requested ?add_roots ~assume_built ?(force_remove=false)
           when not (OpamPackage.has_name t.pinned nv.name) ->
           let build_dir =
             OpamPath.Switch.build t.switch_global.root t.switch nv in
-          if not OpamClientConfig.(!r.keep_build_dir) then
-            OpamFilename.rmdir build_dir
+          if not OpamClientConfig.(!r.keep_build_dir) then begin
+            OpamFilename.rmdir build_dir;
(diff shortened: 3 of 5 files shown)


src/client/opamSolution.ml
```diff
@@ -718,8 +718,12 @@ let parallel_apply t ~requested ?add_roots ~assume_built ?(force_remove=false)
           when not (OpamPackage.has_name t.pinned nv.name) ->
           let build_dir =
             OpamPath.Switch.build t.switch_global.root t.switch nv in
-          if not OpamClientConfig.(!r.keep_build_dir) then
-            OpamFilename.rmdir build_dir
+          if not OpamClientConfig.(!r.keep_build_dir) then begin
+            OpamFilename.rmdir build_dir;
+            OpamFilename.remove
```
AltGr, line 723: Please define this in `OpamPath` rather than hardcoding the string operations here.

AltGr (changes requested):
Thanks!

A few notes:
- please avoid hardcoding specific paths at several places, it's not maintainable; we use the `OpamPath` module to centralise such definitions. Especially here since it's expected as an interface with plugins.
- maybe a cleaner interface would be to provide the `opam` file location as a variable to the hook.
- with a variable, since we should already have the opam file stored somewhere, maybe we could even provide its location directly and avoid the extra file write?

src/client/opamSolution.ml
lefessan: Done

lefessan:
I have done the fix using `OpamPath`.
For the hook variable, it was not clear to me how to find the location of the `opam` description file from the place where hook variables are defined. 
Moreover, using such a variable means creating an incompatibility between the current and new versions of `opam`, since the plugin would have to put a hook description that would depend on the version of `opam`. From that point of view, I would prefer the env. variable or the (current with this PR) implicit file location (`opam-bin` checks for the existence of the file, and calls `opam` if the file does not exist, which is compatible with new and old `opam` versions).

src/format/opamPath.ml
```diff
@@ -102,6 +102,9 @@ module Switch = struct
 
   let build t a nv = build_dir t a / OpamPackage.to_string nv
 
+  let build_opam ~build_dir =
+    OpamFilename.of_string ( OpamFilename.Dir.to_string build_dir ^ ".opam" )
```
AltGr, line 106:
```suggestion
  let build_opam t a nv = build_dir t a /- OpamPackage.to_string nv ^ ".opam"
```

src/format/opamPath.mli
```diff
@@ -117,6 +117,10 @@ module Switch: sig
       {i $meta/build/$packages} *)
   val build: t -> switch -> package -> dirname
 
+  (** Temporary copy of the opam description of the package during
+      the build/install process *)
+  val build_opam: build_dir:dirname -> filename
```
AltGr, line 122:
```suggestion
  val build_opam: t -> switch -> package -> OpamFile.OPAM.t OpamFile.t
```

src/client/opamAction.ml
```diff
@@ -398,6 +398,9 @@ let prepare_package_source st nv dir =
                  (OpamStd.Format.itemize OpamFilename.to_string bad_hash)));
   in
   OpamFilename.mkdir dir;
+  OpamFile.OPAM.write
+    (OpamFile.make (OpamPath.Switch.build_opam ~build_dir:dir))
+    opam;
```
AltGr, line 403:
```suggestion
  OpamFile.OPAM.write
    (OpamPath.Switch.build_opam st.switch_global.root st.switch nv)
    opam;
```

src/client/opamSolution.ml
```diff
@@ -718,8 +718,10 @@ let parallel_apply t ~requested ?add_roots ~assume_built ?(force_remove=false)
           when not (OpamPackage.has_name t.pinned nv.name) ->
           let build_dir =
             OpamPath.Switch.build t.switch_global.root t.switch nv in
-          if not OpamClientConfig.(!r.keep_build_dir) then
-            OpamFilename.rmdir build_dir
+          if not OpamClientConfig.(!r.keep_build_dir) then begin
+            OpamFilename.rmdir build_dir;
+            OpamFilename.remove ( OpamPath.Switch.build_opam ~build_dir )
```
AltGr, line 723:
```suggestion
            OpamFilename.remove (OpamFile.filename (OpamPath.Switch.build_opam t.switch_global.root t.switch nv))
```

AltGr (commented):
Suggested changes mostly for style, but there is a reason for not computing the path relative to `build_dir`, for cases where e.g. `--inplace-build` was used and `build_dir` can be outside of the opam-controlled dirs.

src/format/opamPath.ml
lefessan: That would create the file within the packages sources, no ? I think it's a bit dangerous since it can badly interact with the build/install process (for example, `opam-bin` installs everything that is in the source tree of its binary packages, so it would install the opam file too).

src/format/opamPath.mli
lefessan: I used the `build_dir` here because it was available at the place where `build_opam` is called. I don't like recomputing something that was already computed...

master_changes.md
```diff
@@ -17,6 +17,7 @@ New option/command/subcommand are prefixed with ◈.
 
 ## Install
   * The stdout of `pre-` and `post-session` hooks is now propagated to the user [#4382 @AltGr - fix #4359]
+  * A PACKAGE.opam file is provided during build in the build/ directory [@lefessan]
```
rjbou, line 20:
```suggestion
  * A PACKAGE.opam file is provided during build in the build/ directory [#4387 @lefessan]
```

rjbou:
Closing as #4402 merged. Thanks for the PR!

PR #4388, opened 2020-10-11
Allow `post-install-commands` hooks to modify/remove installed files

lefessan:
This modification is necessary because `opam-bin` uses the `post-install-commands` to share files between switches using hard links. Unfortunately, without this PR, such actions modify the digests of the files just after their recording, so that `opam remove` will complain that the installed files have been modified.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -17,6 +17,7 @@ New option/command/subcommand are prefixed with ◈.
 
 ## Install
   * The stdout of `pre-` and `post-session` hooks is now propagated to the user [#4382 @AltGr - fix #4359]
+  * `post-install` hooks are allowed to modify or remove installed files, but not add new ones (@lefessan)
 
 ## Remove
   *
--- a/master_changes.md
+++ b/master_changes.md
@@ -17,7 +17,7 @@ New option/command/subcommand are prefixed with ◈.
 
 ## Install
   * The stdout of `pre-` and `post-session` hooks is now propagated to the user [#4382 @AltGr - fix #4359]
-  * `post-install` hooks are allowed to modify or remove installed files, but not add new ones (@lefessan)
+  * `post-install` hooks are allowed to modify or remove installed files, the but not add new ones. Those changes are integrated in changes file [#4388 @lefessan]
 
 ## Remove
   *
--- a/src/core/opamDirTrack.ml
+++ b/src/core/opamDirTrack.ml
@@ -268,3 +268,32 @@ let revert ?title ?(verbose=OpamConsole.verbose()) ?(force=false)
                     (OpamStd.Option.to_string (fun pr ->
                          if pr then "[hash] " else "[tms] ") pre) ^ x) lf))
           cannot))
+
+let update prefix t =
+  let removed = ref [] in
+  let update_digest file digest =
+    match
+      let filename = Filename.concat prefix file in
+      let precise = is_precise_digest digest in
+      item_digest ( item_of_filename ~precise filename )
+    with
+    | exception Unix.Unix_error ( ENOENT, _, _) ->
+      removed := file :: !removed;
+      digest
+    | exception _exn -> digest
+    | digest -> digest
+  in
+  let t =
+    SM.mapi (fun file change ->
+        match change with
+        | Added digest -> Added (update_digest file digest)
+        | Removed -> Removed
+        | Contents_changed digest ->
+          Contents_changed (update_digest file digest)
+        | Perm_changed digest -> Perm_changed (update_digest file digest)
+        | Kind_changed digest -> Kind_changed (update_digest file digest)
+      ) t
+  in
+  List.fold_left (fun t file ->
+      SM.remove file t
+    ) t !removed
--- a/src/core/opamDirTrack.ml
+++ b/src/core/opamDirTrack.ml
@@ -271,6 +271,7 @@ let revert ?title ?(verbose=OpamConsole.verbose()) ?(force=false)
 
 let update prefix t =
   let removed = ref [] in
+  let prefix = OpamFilename.Dir.to_string prefix in
   let update_digest file digest =
     match
       let filename = Filename.concat prefix file in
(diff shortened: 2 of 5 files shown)


AltGr:
LGTM, thanks!

hannesm:
I'm a bit sceptical this is a good idea - is there an overview which phases and hooks are allowed to do what? At least, the [documentation](https://opam.ocaml.org/doc/Manual.html#configfield-post-build-commands) should be adjusted:
```
The post-install-commands hook also has access to an extra variable installed-files which expands to the list of files and directories added or modified during the installation of the package. Note that this hook is run after the scan for installed files is done, so any additional installed files won't be recorded and must be taken care of by a pre-remove-commands hook.
```

wouldn't such a  post-install and a corresponding pre-remove work for `opam-bin`?

lefessan:
> wouldn't such a post-install and a corresponding pre-remove work for `opam-bin`?

`opam-bin` already adds files during the `post-install` hook that it later removes using the `pre-remove` hook. 
But in this particular case, it is not about adding new files, but changing files that have just been installed. Even if `opam-bin` records its changes, it won't be able to restore the original files, so `opam` will keep complaining that the files that were installed were either removed or modified. Keeping a copy of the file would defeat the purpose which is to save space on the disk by sharing installed files between switches.

Note that this PR does not allow a `post-install` hook to add new files in the changes, only to modify/remove some of them.

avsm:
Lets not be hasty about adding in hook features this late in opam 2.1's release cycle without careful justification, please. I think it's important to understand the power of these hooks:

- does this hook facility bypass the sandboxing process?
- can a post-install hook modify files installed by other packages, or just those under its original control?

lefessan:
> Lets not be hasty about adding in hook features this late in opam 2.1's release cycle without careful justification, please. I think it's important to understand the power of these hooks:
> 
> * does this hook facility bypass the sandboxing process?
> * can a post-install hook modify files installed by other packages, or just those under its original control?

Hooks completely bypass the sandboxing process : it is already possible for post-install hooks (and other hooks) to modify files anywhere on the system.

AltGr:
Just clarifying two points:
- sandboxing is itself a hook, so command-wrapping hooks can be put inside or outside the sandbox by customising the order of definition in ~/.opam/config; pre/post hooks are not sandboxed.
- this PR really doesn't allow hooks to do anything they couldn't: it just actually handles the case where the `post-install` hook modifies files in a more consistent way.

The latter explains I have no objection, since the PR does not change existing features but improves handling for a specific case of an existing feature.

master_changes.md
```diff
@@ -17,6 +17,7 @@ New option/command/subcommand are prefixed with ◈.
 
 ## Install
   * The stdout of `pre-` and `post-session` hooks is now propagated to the user [#4382 @AltGr - fix #4359]
+  * `post-install` hooks are allowed to modify or remove installed files, but not add new ones (@lefessan)
```
rjbou, line 20:
```suggestion
  * `post-install` hooks are allowed to modify or remove installed files, but not add new ones [#4388 @lefessan]
```

hannesm:
thanks for the explanation @AltGr.

avsm:
Thanks for the explanations. Could we get the Changes entry to be more accurate then?  Specifically, it appears that opam 2.1 is now _tracking_ the results of ~/.opam changes made during the post-install hook execution and registering those with opam.  Are there any other implications of making this change (e.g. to `opam list --show-files`?)

lefessan:
> Thanks for the explanations. Could we get the Changes entry to be more accurate then? Specifically, it appears that opam 2.1 is now _tracking_ the results of ~/.opam changes made during the post-install hook execution and registering those with opam. Are there any other implications of making this change (e.g. to `opam list --show-files`?)

There is already an entry in the Changes that is added with the PR. Note that only the files that were installed by the package are checked for modification, other files added/modified by the `post-install` hook are not, and should be removed/restored by the plugin using the `pre-remove` hook.

lefessan:
> Are there any other implications of making this change (e.g. to `opam list --show-files`?)

Users that do not use `opam-bin` usually have no `post-install` hook, so nothing is changed for them (no additional cost). For those using a `post-install` hook, the PR does not increase the power of what the hook can do, it just prevents them from corrupting opam changes files.

AltGr:
Just one last tidbit, the last sentence in the manual [here](http://opam.ocaml.org/doc/Manual.html#configfield-post-install-commands) should be updated to reflect the change. The doc doesn't become incorrect, but just precise something along the lines of "Modified installed files will be handled correctly by opam"

Source: https://github.com/ocaml/opam/blob/master/doc/pages/Manual.md

lefessan:
> Just one last tidbit, the last sentence in the manual [here](http://opam.ocaml.org/doc/Manual.html#configfield-post-install-commands) should be updated to reflect the change. The doc doesn't become incorrect, but just precise something along the lines of "Modified installed files will be handled correctly by opam"
> 
> Source: https://github.com/ocaml/opam/blob/master/doc/pages/Manual.md

Indeed, I have modified the Manual file in this new version.

master_changes.md
rjbou, line 20:
```suggestion
  * `post-install` hooks are allowed to modify or remove installed files, the but not add new ones. Those changes are integrated in changes file [#4388 @lefessan]
```

src/core/opamDirTrack.ml
```diff
@@ -268,3 +268,32 @@ let revert ?title ?(verbose=OpamConsole.verbose()) ?(force=false)
                     (OpamStd.Option.to_string (fun pr ->
                          if pr then "[hash] " else "[tms] ") pre) ^ x) lf))
           cannot))
+
+let update prefix t =
+  let removed = ref [] in
```
rjbou, line 273:
```suggestion
  let removed = ref [] in
  let prefix = OpamFilename.Dir.to_string prefix in
```

rjbou (changes requested):
Small changes, but LGTM!
In order to keep consistency with function inf `OpamDirTrack`, `update` should take a `dirname` and not a `string`.

rjbou (approved):

PR #4395, opened 2020-10-16
Use dune "expected" convention for patcher test

emillon:
When a there's a file named `x.expected` next to a test named `x.exe`, dune will use expect semantics:

- `x.exe` will get run as part of the test suite
- its output will be saved to a temporary file
- this file is compared to `x.expected`
- promotion works (if the output changes, `dune promote` will update `x.expected`)


dra27 (approved):
Nice - puts responsibility on finding `diff` to Dune as well!

PR #4396, opened 2020-10-18
Depexts: Add support for NetBSD, DragonFlyBSD (+ fix OpenBSD, FreeBSD and Gentoo)

kit-ty-kate:
This PR:
* Adds support for DragonFlyBSD, which use virtually the same package manager as FreeBSD (`pkg`). For our purpose they behave exactly the same way and have the same required commands so I just merged the two together like it's done for centos/fedora/...
* Adds support for NetBSD, it seems like `pkgin` is the default package manager now, the old `pkg_add` is not even configured by default. It is to be noted that OpenBSD's and NetBSD's `pkg_info`/`pkg_add` are two very different software, even if they have the same name.
* Fixes OpenBSD, FreeBSD and Gentoo with regards to package paths. Packages on those systems come from their "ports collection" system and categorize their packages to prevent name conflicts.
  - On OpenBSD and Gentoo, packages can be installed using the fullpath name or the short one, and fail or ask the user if the short one is requested but another package with the same name exists
  - On FreeBSD it's the same thing, however a prefix or suffix is added to the package name in case of a conflict. For instance, both `lang/elm` and `mail/elm` exists but `lang/elm` becomes `hs-elm` when "packetized".

Many packages in opam-repository right now have either one of the form (long names or short ones) and as it stands with opam 2.1 the long ones would not be detected on gentoo on freebsd and the short ones would not be detected on openbsd.

*Side note: I only tested individual commands on those systems, not opam directly. More testing is needed if we want to ensure this works correctly.*

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -46,7 +46,8 @@ New option/command/subcommand are prefixed with ◈.
   *
 
 ## External dependencies
-  *
+  * Add support for NetBSD and DragonFlyBSD [#4396 @kit-ty-kate]
+  * Fix OpenBSD, FreeBSD and Gentoo: Allow short names and full name paths for ports-based systems [#4396 @kit-ty-kate]
 
 ## Sandbox
   *
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -106,10 +106,14 @@ let family =
       | "amzn" | "centos" | "fedora" | "mageia" | "oraclelinux" | "ol"
       | "rhel" -> Centos
       | "archlinux" | "arch" -> Arch
-      | "bsd" when OpamSysPoll.os_distribution () = Some "freebsd" ->
-        Freebsd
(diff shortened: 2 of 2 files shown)


rjbou (approved):
Thanks for the fix and new platform! It's a good to go for me!
Agree that it need some tests, if some netbsd/freebsd/openbsd/gentoo people can try this PR, indication are [here](https://github.com/ocaml/opam/wiki/How-to-test-an-opam-feature).
@kit-ty-kate juste some nitpicking :) I try to keep the type and its pattern matching alphabetically ordered, to find more easily a platform when needed, can you move `Netbsd` before `Openbsd`. I'd should add a comment about that in the code, something like
```
(* Please keep this alphabetically ordered, in the type definition,
    and in pattern matching *)
type families =
  | Alpine
```

rjbou:
`Unbound value String.rindex_opt` for ocaml 4.02.3 (travis)

PR #4398, opened 2020-10-19
opam-solver: Require opam-0install-cudf >= 0.4

kit-ty-kate:
[opam-0install-cudf.0.4](https://github.com/ocaml-opam/opam-0install-solver/releases/tag/v0.4) fixed a rather big issue with the way the solver treated conflicts and the previous version (0.3) should be avoided by end users.


kit-ty-kate:
Would it be possible get that in opam 2.1.0~beta4? @rjbou

PR #4399, opened 2020-10-20
Parallel pin source fetching

### LINKED ISSUE #4315: pin-depends pinning speed
I recently tried to `opam install` a local package with every dependency pinned via the `pin-depends` field in its opam file and the `opam install` is significantly slowed down by the initial pinning process.

I tried the 2.1 alpha hoping it might have been improved by recent changes but it seems to have no effect.

Do you have any idea why this takes so long and/or if this can be easily improved? I suspect that cloning the repositories is what takes up most of the time but I also imagine running those in parallel could improve the situation.

rjbou:
Fix #4315
Using the same example, it took 4m32 (with download-only option), and now with parallel and reuse of source dir, it takes only 1m32.
It also applies to `pin scan` arguments, and url pin.
cc/ @NathanReb

### DIFF
--- a/src/client/opamPinCommand.ml
+++ b/src/client/opamPinCommand.ml
@@ -366,6 +366,43 @@ let default_version st name =
   try OpamPackage.version (OpamSwitchState.get_package st name)
   with Not_found -> OpamPackage.Version.of_string "~dev"
 
+let fetch_all_pins st pins =
+  let root = st.switch_global.root in
+  let fetched =
+    let cache_dir =
+      OpamRepositoryPath.download_cache OpamStateConfig.(!r.root_dir)
+    in
+    let command (name, url, subpath) =
+      let srcdir =
+        OpamPath.Switch.pinned_package root st.switch
+          (OpamPackage.Name.of_string name)
+      in
+      OpamProcess.Job.Op.(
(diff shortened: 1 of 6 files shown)


src/client/opamPinCommand.ml
```diff
@@ -366,6 +366,43 @@ let default_version st name =
...
+    List.fold_left (fun (err,ok) result ->
+        let name, url, subpath, result = result in
+        match result with
+        | Not_available _ -> (* clean dir ? *)
+          (name, url, subpath)::err, ok
+        | _ -> err, (url, subpath)::ok)
+      ([],[]) fetched
+  in
+  if errored = []
+  || OpamConsole.confirm "Could not retrieve%s\nContinue?"
```
AltGr, line 396: We may want to be more clear about the implications of saying "yes" here ?
Also, be careful that `yes` is the default and may be forced with `--yes` misled people.

AltGr (commented):
Seems all good to me, just one `confirm` call that may not be clear enough to the user.

src/client/opamPinCommand.ml
rjbou: Maybe adding something like "You may reach an inconsistent state because of those not pinned packages"?

rjbou: updated

PR #4402, opened 2020-10-21
New package variable: `opamfile`

rjbou:
superseeds #4387

### DIFF
--- a/src/state/opamPackageVar.ml
+++ b/src/state/opamPackageVar.ml
@@ -46,6 +46,7 @@ let package_variable_names = [
   "dev",       "True if this is a development package";
   "build-id",  "A hash identifying the precise package version with all its \
                 dependencies";
+  "opamfile", "Path of the curent opam file";
 ]
 
 let predefined_depends_variables =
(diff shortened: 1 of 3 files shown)


src/state/opamPackageVar.ml
```diff
@@ -305,6 +306,19 @@ let resolve st ?opam:opam_arg ?(local=OpamVariable.Map.empty) v =
...
+    | "opamfile-loc", Some opam ->
+      let repos_roots reponame =
+        match Hashtbl.find st.switch_repos.repos_tmp reponame with
+        | lazy repo_root -> repo_root
+        | exception Not_found ->
+          OpamRepositoryPath.root st.switch_global.root reponame
+      in
+      OpamFile.OPAM.get_metadata_dir ~repos_roots opam
+      |> OpamStd.Option.map (fun d ->
+          OpamFilename.Op.(d//"opam")
```
AltGr, line 318: Hm, does this work for pinned packages ? Can't the file be named differently?

AltGr (commented):
Thanks! Some concerns about the generality of implementation, and maybe we could find a better name ? (`-loc` doesn't sound straightforward to me... Why not just `opamfile` or `opam-file` ?)

src/state/opamPackageVar.ml
rjbou: for pinned packages, it's stored in the `overlay` dir, in the [file `opam`](https://github.com/ocaml/opam/blob/master/src/format/opamPath.ml#L254), and from repo, its the package dir with name `opam`.

rjbou:
ok for me to rename it. `opamfile` can be misleading with the content of the opam file... first I thought about `opamfile-path`

lefessan:
I think `opamfile` is ok, nobody would expect the content of the file as an argument to a hook script, no ?

src/state/opamPackageVar.ml
AltGr: aah right, it'll never get back to the original srcdir, we always have a mirror, got it! :+1:
maybe worth adding a comment ?

rjbou: It's always there, in the current run of opam (repo dir are tmp ones), yes. In fact, it's the code of `OpamRepositoryState.get_root`, that is not reachable from `OpamPackageVar`.
+1 for the comment

Issue #4406, opened 2020-10-28
Option to refresh sandbox.sh file


ChrisVine:
`opam upgrade` recently started failing for me.  The proximate cause of this seems to have been ccache moving its default cache
directory from $HOME/.ccache to $HOME/.cache/ccache in its latest version (version 4.0), coupled with the sandbox.sh file installed by opam being out of date.  This caused the sandboxing issue referred to [here](https://discuss.ocaml.org/t/ocaml-migrate-parsetree-update-fails/6663) to arise.

In the end I resolved this by copying src/state/shellscripts/bwrap.sh from the latest opam tarball to $HOME/.opam/opam-init/hooks/sandbox.sh by hand: moving sandbox.sh and running `opam init` again does not appear to cause it to be refreshed automatically.

I suggest `opam init` should have an option available to it to refresh sandbox.sh.  Better error reporting where a sandboxing error like this occurs would also be helpful, so that the user can more easily diagnose the cause.

rjbou:
Related to #4361

Issue #4361, opened 2020-09-15
macOS sandbox file is not updated when opam is updated


MSoegtropIMC:
As far as I can tell the macOS sandbox script (https://github.com/ocaml/opam/blob/master/src/state/shellscripts/sandbox_exec.sh) is copied to `.opam/opam-init/hooks` during `opam init`. If opam is updated, this file is not touched. This has the effect that people have recent versions of opam with a not working sandbox. This might depend on the way how opam is installed and/or updated.

A notable effect of this is that any project build with `remake` does not build in case the sandbox file is too old. I meanwhile check besides the version of opam some properties of the sandbox script in my meta install scripts.

rjbou:
In order to update sandbox scripts, a `reinit` is needed : `opam init --reinit -ni`. It is advised in all release blog posts (at the bottom).

Is the problem occurs even with a reinit?

MSoegtropIMC:
> Is the problem occurs even with a reinit?

I can't tell immediately - it was an issue reported by a user of my opam based Coq meta build scripts. I will contact the user and try to clarify this.

Is there a way to check which version of opam created or last updated the `.opam` folder, so that I could check this in my scripts and automatically suggest and run the opam reinit?

I can definitely check if the sandbox script is suitable and suggest the reinit in case it is not.

rjbou:
It is possible if there is changes at opam root (cf. new field `opam-version` at the top of the `.opam/config`), but it won't help you about sandbox scripts.
To be sure, by default at each new version of opam, advise to run the reinit command. It also can be run outside any opam new release, to be sure to regenerate scripts. You can also see `opam init` options to customize the reinit.

rjbou:
Note that there is a [fix](https://github.com/ocaml/opam/pull/4092) of `-ni` option in 2.1.0~alpha.

rjbou:
Something that may help you for integration: in `2.1.0~alpha`, we added a hash of sandbox script, to avoid overwriting user's modifications (cf. #4020). That hash (md5) is stored in `.opam/opam-init/hooks/sandbox.sh.hash`.
You can check that hash with the opam repo one at given version (e.g. for 2.1.0~beta on linux https://github.com/ocaml/opam/blob/2.1.0-beta/src/state/shellscripts/bwrap.sh), and advice a reinit if it differ from the opam version you need. Keep in mind that hashes can differ because of local user modification.

MSoegtropIMC:
[...]

PR #4414, opened 2020-11-04
Sandbox updates checker

rjbou:
Add in sandbox script a tes to check if the current opam generated the sandbox script. Display a message and fail in case it differs.
The check is done only regarding opam available in PATH ("system-wide" installed).

fixe #4361 #4406


rjbou:
I'm not sure if it should fail, but if not, there is no way to show the message to user as it is embedded in sandbox script. It is done this way to avoid checking it at each opam action command launch.
example of failure:
```
opam install ago
  ∗ install ago 0.4

⬇ retrieved ago.0.4  (cached)
[ERROR] The compilation of ago.0.4 failed at "jbuilder build -p ago -j 7".

#=== ERROR while compiling ago.0.4 ============================================#
# context     2.1.0~beta4 | linux/x86_64 |  | git+file:///home/rjbou/ocamlpro/opam-repository#master
# path        ~/.opam/switch/.opam-switch/build/ago.0.4
# command     ~/.opam/opam-init/hooks/sandbox.sh build jbuilder build -p ago -j 7
# exit-code   2
# env-file    ~/.opam/log/ago-1006599-0bf4e0.env
# output-file ~/.opam/log/ago-1006599-0bf4e0.out
### output ###
# Sandbox script is out of date. To update it, please run
#     opam init --reinit -ni



│ λ build ago 0.4
└─ 
╶─ No changes have been performed
```

src/state/dune
```diff
@@ -13,8 +13,10 @@
 (rule
   (targets opamScript.ml)
   (deps    ../../shell/crunch.ml (glob_files shellscripts/*.*sh))
-  (action  (with-stdout-to %{targets} (run ocaml %{deps}))))
-
+  (action  (with-stdout-to %{targets}
+  (progn
+    (run ocaml %{deps})
+    (run sed -i -e "s/$SCRIPTOPAMVERSION\\b/%{read-lines:../version}/g" %{targets})))))
```
dra27, line 19: `sed -i` isn't portable (macOS) and while unlikely "2.1.0" is itself a regex! This wants doing with an OCaml script using `Re` (it's tedious, but not too bad).

src/state/shellscripts/bwrap.sh
```diff
@@ -11,6 +11,13 @@ if ! command -v bwrap >/dev/null; then
     exit 10
 fi
 
+OPAMVERSION=`opam --version`
+if [ "$OPAMVERSION" != "$SCRIPTOPAMVERSION" ]; then
```
dra27, line 15: Wouldn't it be better to use a key which itself isn't a shell variable? Like `@SCRIPTOPAMVERSION@` or something?

dra27 (commented):
I think this scheme needs changing (and, as ever, I'm sorry for entering with that so late on). This will cause every single point release to tell users to `--reinit`, whether we've updated the sandbox script or not.

I think this scheme might work:

- opam needs to embed the hash of the crunched script and have a cli option which will return it
- The script should no longer have a shebang - that and `SCRIPT_HASH=...` should be written by opam followed by the actual script.

On the plusside, it removes the need for doing `sed` in the Dune files 🙂

src/state/shellscripts/bwrap.sh
```diff
@@ -11,6 +11,13 @@ if ! command -v bwrap >/dev/null; then
     exit 10
 fi
 
+OPAMVERSION=`opam --version`
```
kit-ty-kate, line 14:
```suggestion
local OPAMVERSION=`opam --version`
```

AltGr: uuh... calling opam from here ??
Also we have no guarantee that the calling process is actually the same version as `opam` looked up in the PATH here ?

kit-ty-kate, line 14: _(putting back @AltGr that I somehow deleted when deleting my original comment that opened this block)_

uuh... calling opam from here ??
Also we have no guarantee that the calling process is actually the same version as opam looked up in the PATH here ?

src/state/dune
AltGr: The `sed -i %{targets}` within `with-stdout-to %{targets}` seems highly suspicious to me anyway :roll_eyes:

A cleaner approach would be to rename `bwrap.sh` to `bwrap.sh.in`, and have a separate rule for the sed from the latter to the former (I don't remember if `glob_files` will handle it correctly though).

kit-ty-kate (changes requested):
This PR wouldn't have caught things like https://github.com/ocaml/opam/issues/6405. I think the only way to do it correctly is to do the check we already do as part of `opam init --reinit -ni` but everytime prior to running `opam install` and things that might call the sandbox (i'm not sure where is best to put that though)

PR #4427, opened 2020-11-11
Print shell-appropriate eval command on `opam init`

freevoid:
This change is similar in spirit to https://github.com/ocaml/opam/pull/3274/commits/46eab70fd9318057c785a2bcfb31e8184f9fa492 (an open pull request marked as obsolete) with a slightly different implementation.

I'm new to OCaml and been following the instructions to install `opam`, during `opam init` it prints a shell-dependent line that will be added, but incorrectly prints `eval $(opam env)` (I use `fish` so that should be `eval (opam env)`):

```
In normal operation, opam only alters files within ~/.opam.

  However, to best integrate with your system, some environment variables
  should be set. If you allow it to, this initialisation step will update
  your fish configuration by adding the following line to ~/.config/fish/config.fish:

    source /Users/name/.opam/opam-init/init.fish > /dev/null 2> /dev/null; or true

  Otherwise, every time you want to access your opam installation, you will
  need to run:

    eval $(opam env)

  You can always re-run this setup with 'opam init' later.
```

Let me know if the change is not idiomatic, since I just installed `opam` and `ocaml` and have no previous experience with either.

### DIFF
--- a/src/state/opamEnv.ml
+++ b/src/state/opamEnv.ml
@@ -335,6 +335,24 @@ let is_up_to_date ?skip st =
   is_up_to_date_raw ?skip
     (updates ~set_opamroot:false ~set_opamswitch:false ~force_path:false st)
 
+(** Returns shell-appropriate statement to evaluate [cmd]. *)
+let shell_eval_invocation shell cmd =
+  match shell with
+  | SH_fish ->
+    Printf.sprintf "eval (%s)" cmd
+  | SH_csh ->
+    Printf.sprintf "eval `%s`" cmd
+  | _ ->
+    Printf.sprintf "eval $(%s)" cmd
+
+(** Returns "opam env" invocation string together with optional root and switch
+    overrides *)
+let opam_env_invocation ?root ?switch ?(set_opamswitch=false) () =
+  let root = OpamStd.Option.map_default (Printf.sprintf " --root=%s") "" root in
+  let switch = OpamStd.Option.map_default (Printf.sprintf " --switch=%s") "" switch in
+  let setswitch = if set_opamswitch then " --set-switch" else "" in
+  Printf.sprintf "opam env%s%s%s" root switch setswitch
+
 let eval_string gt ?(set_opamswitch=false) switch =
   let root =
     let opamroot_cur = OpamFilename.Dir.to_string gt.root in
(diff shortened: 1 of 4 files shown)


dra27:
Thanks for picking up the pieces of #3274! It's a better idea than I did in that PR to split the `eval` into two functions, but I would split it differently - I would have one function which works out the `opam env` command required (so takes all the extra parameters in to account) and then another command which takes that and turns it into the correct `eval` command.

I think you then want `opam_env_invocation`. That wants to be `let opam_env_invocation : ?root ?switch ?(set_opamswitch=false) () = (* if root is Some x, convert it to --root=x, as eval_string does, etc. *)` and essentially works out `suffix` in your current patch but with the `opam env` part too. `eval_string` would use that function instead of working out the `--root=` and `--switch`, etc. itself.

 You then want a new function `shell_eval_invocation` which takes parameters `shell` and `cmd` (and is basically your current `eval_string_simple`). I think this addresses @AltGr's https://github.com/ocaml/opam/pull/3274#issuecomment-378984654 nicely since it's then not necessary to add `--shell` to the other commands.

If you want, you could also use those functions for a version of https://github.com/ocaml/opam/pull/3274/commits/b4b3fca9f9b96a74431dc99304baea19e3bfe77e (which tries to the use the correct "guessed" eval string in help text output too)

freevoid:
Hey David, thanks for the detailed comment!

I've tried to express your suggestions in 8340e6e, hopefully I didn't misunderstand.

In terms of using the functions for `opamCommands.ml`, I think I could export the `shell_eval_invocation` function and do something in the spirit of `OpamEnv.(shell_eval_invocation shell "opam config env" |> Manpage.escape)`. If it sounds about right, I will go ahead and make the changes.

I see how b4b3fca used `Manpage` and `ManSwitch` to call different flavors of `eval_string`, I think we can keep it a bit simpler and just use `shell_eval_invocation` directly, so that `eval \\$(opam env --switch=SWITCH --set-switch\\)` becomes `OpamEnv.(shell_eval_invocation shell "opam env --switch=SWITCH --set-switch" |> Manpage.escape)`.

The problem this PR tries to address is specifically "`eval $(foo)`" messages from opam, so maybe we could address spelled out "--switch=SWITCH" and "--set-switch" in help messages separately.

src/state/opamEnv.ml
```diff
@@ -335,6 +335,32 @@ let is_up_to_date ?skip st =
...
+  let root =
+    match root with
+    | None -> ""
+    | Some r -> Printf.sprintf " --root=%s" r
+  in
+  let switch =
+    match switch with
+    | None -> ""
+    | Some s -> Printf.sprintf " --switch=%s" s
+  in
```
dra27, line 360: Slightly more concise as:
```suggestion
  let root = OpamStd.Option.map_default (Printf.sprintf " --root=%s") "" r in
  let switch = OpamStd.Option.map_default (Printf.sprintf " --switch=%s") "" s in
```

src/state/opamEnv.ml
```diff
@@ -344,12 +370,13 @@ let eval_string gt ?(set_opamswitch=false) switch =
...
       ) in
     if opamroot_cur <> opamroot_env then
-      Printf.sprintf " --root=%s" opamroot_cur
+      Some opamroot_cur
     else
-      "" in
+      None
+  in
   let switch =
     match switch with
```
dra27, line 378: This can be refactored slightly more clearly as:
```ocaml
let f sw =
  let sw_cur = ... (* etc. code from the `Some sw` case *)
in
OpamStd.Option.replace f switch
```

dra27 (commented):
My apologies for not responding to your message a bit more quickly, but thank you for rebasing and updating this in the meantime. The implementation is exactly what I had in mind - the suggestions below are just possibly to use `OpamStd.Option` to reduce some of the `match`es on options.

This is basically ready to merge - thanks for taking on my stale PR and improving it!

src/state/opamEnv.ml
freevoid: Thanks, applied (r and s converted to "root" and "switch")

src/state/opamEnv.ml
freevoid: Thanks, refactored

freevoid:
Thank you for the review!
I've squashed the commits a bit as their number started to grow unnecessarily

dra27:
Lovely, thank you!

dra27:
(The mingw64 failure is nothing to do this PR)

PR #4429, opened 2020-11-13
Be more robust w.r.t. new caches updates

AltGr:
Technically speaking, users should use `--readonly`, but it's nicer of us just
to skip the cache update and go on anyway since no functionnality will be
impacted.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -91,6 +91,7 @@ New option/command/subcommand are prefixed with ◈.
   *
 
 ## Internal
+  * Be more robust w.r.t. new caches updates [#4429 @AltGr - fix #4354]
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
Does this also work for `opam config var prefix` ? `fiat-p256.0.2.3` fails with this patch due to opam failing during this command.

kit-ty-kate:
Same failure and same command in `mirage-crypto.0.8.7`

rjbou:
This commit was merged in #4467, but as @kit-ty-kate mentionned, the issue is still here.

rjbou:
Closing, the issue is still open for tracking the issue

PR #4430, opened 2020-11-13
Internal caches: use size checks from Marshal

AltGr:
it seems in some cases (large files) there can be a bit of padding, and the
`Marshal` module should already detect size issues, no need do duplicate the 
work.


dra27 (approved):
This looks good to me - the check is indeed unnecessary, `Marshal` of course cannot guarantee the type of the resulting value, but it does fail properly (i.e. without segfault) if the marshalled data is corrupt.

Is there an issue closed by this, or was it just spotted while working on opam?

rjbou (approved):

rjbou:
Thanks!

PR #4435, opened 2020-11-18
Update opam repository man doc

### LINKED ISSUE #4381: Failure to remove the last repository in a switch
On `2.1.0~alpha3`, it's not possible to remove the only repository in a switch, but the user messages and exit codes do not suggest as such.

```
❭ mkdir /tmp/opam-no-remove-default
❭ cd /tmp/opam-no-remove-default
❭ opam switch create ./ --empty
❭ opam repo
[NOTE] These are the repositories in use by the current switch. Use '--all' to see all configured repositories.

<><> Repository configuration for switch /tmp/opam-no-remove-default ><><><><><>
 1 default https://opam.ocaml.org

❭ opam repo remove default
Repositories removed from the selections of switch /tmp/opam-no-remove-default. Use '--all' to forget about them altogether.

❭ opam repo
[NOTE] These are the repositories in use by the current switch. Use '--all' to see all configured repositories.

<><> Repository configuration for switch /tmp/opam-no-remove-default ><><><><><>
 1 default https://opam.ocaml.org
```

I'd expect `opam repo remove default` to either fail with an appropriate error message or to actually remove the `default` repository from the switch.

rjbou:
fix #4381

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -2027,10 +2027,10 @@ let repository cli =
      enable a repository only to install its switches, you may be \
      looking for $(b,opam switch create --repositories=REPOS).";
     cli_original, "remove", `remove, ["NAME..."],
-    "Unselects the given repositories so that they will not be used to get \
-     package definitions anymore. With $(b,--all), makes opam forget about \
-     these repositories completely. When last repository is removed, it is \
-     removed and  replaced by global default repository ";
+    "Removes the given repositories from the selected scopes (the default scope is the current switch, \
+     see $(b,--this-switch)). Repositories are removed completely with $(b,--all-switches). \
+     If the last repository is removed in the current switch scope, then the selection of \
+     repositories is restored to the default selection ($(b,opam list --set-default)).";
     cli_original, "set-repos", `set_repos, ["NAME..."],
     "Explicitly selects the list of repositories to look up package \
      definitions from, in the specified priority order (overriding previous \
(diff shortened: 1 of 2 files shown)


src/client/opamCommands.ml
```diff
@@ -1944,7 +1944,8 @@ let repository cli =
     "remove", `remove, ["NAME..."],
     "Unselects the given repositories so that they will not be used to get \
      package definitions anymore. With $(b,--all), makes opam forget about \
```
dra27, line 1946:
```suggestion
    "Removes the given repositories from the selected scopes (the default scope is the current switch, \
     see $(b,--this-switch)). Repositories are removed completely with $(b,--all-switches). \
```

src/client/opamCommands.ml
```diff
@@ -1944,7 +1944,8 @@ let repository cli =
...
-     these repositories completely.";
+     these repositories completely. When last repository is removed, it is \
+     removed and  replaced by global default repository ";
```
dra27, line 1948:
```suggestion
     If the last repository is removed in the current switch scope, then the selection of \
     repositories is restored to the default selection ($(b,opam list --set-default)).";
```

dra27 (commented):
I've re-worded the message text, but I don't think is sufficient - the 2.0 behaviour here is really surprising. On my opam root, which has two default repositories, this means that if you remove the last repo from a switch's selection then you suddenly have two repos added!

I think this also be displayed as a warning when it happens - i.e. if `OpamSwitchState.update_repositories` has restored the list to the default then the "Repositories removed from the selections of switch %s." message should also be "There are no repositories selected, so the selections have been restored to the default selection."

XVilka (approved):

XVilka:
Sorry for unsolicited approval, mistakenly did a review in a wrong browser tab 🤦 Feel free to remove and sorry for the noise.

dra27:
Hah, not to worry @XVilka - it appears I can delete your comment, but not the review!

kit-ty-kate (approved):

PR #4434, opened 2020-11-18
2.0: 4.12 and Dune 2 compatibility

dra27:
~~Wants double-checking with 4.12.0~alpha1 - I've seen some failures with cppo, but hopefully they were one-offs.~~

Fixes:
- https://github.com/ocaml/opam/pull/4425#issuecomment-729676670 - we wind extlib forwards to 1.7.7 and apply the patch for 4.12
- Update cppo to 1.6.6 in order to have `dune` instead of `jbuild` files for Dune 2 compatibility.
- cppo patches for 4.12 and 4.02.3
- Update re to 1.9.0 and result to 1.4 in order to have `dune` files instead of `jbuild` files for Dune 2 compatibility

PR #4437, opened 2020-11-18
4.12 compatibility

dra27:
Pulls the relevant bits from #4434 through.

I've also switched `src_ext/` to use `vendored_dirs` since this is better than the dance done with cppo. That pulls Dune forward to 1.11 (goodness!).

There're still warnings with 4.12, so this isn't quite finished - please feel free to push more commits, I just didn't want us accidentally to fix 2.0 branch and not master!


kit-ty-kate:
Added to the 2.1.0~rc milestone as this require slight change in the API (the warnings invoking this change can be hidden but I feel like it's cleaner this way). Feel free to remove it from the milestone if you think it does not belong to it.

dra27:
It definitely needs to be in the milestone - we can't release not supporting the latest OCaml 🙂

dra27:
I'm more inclined to suppress the warning than change those APIs, though - possibly it could be done in opam-client, but definitely not core and repository, at least at this stage.

master_changes.md
```diff
@@ -81,6 +81,10 @@ New option/command/subcommand are prefixed with ◈.
   * Restrict to opam-file-format 2.1.2 [#4495 @rjbou]
   * Switch to newer version of MCCS (based on newer GLPK) for src_ext [#4559 @AltGr]
   * Bump dune version to 2.8.2 [#4592 @AltGr]
+  * Update opam file to 2.0 [#4371 @AltGr]
+  * Makefile: Add rule `custom-libinstall` for `opam-custom-install` use [#4401 @AltGr]
+  * opam-client lib: fix dependency constraints for cmdliner, extlib [#4410 @AltGr]
+  * 4.12 compatibility [#4437 @dra27]
```
rjbou, line 87: Some of changes are already in beta4 one

kit-ty-kate: Fixed

kit-ty-kate:
> I'm more inclined to suppress the warning than change those APIs, though - possibly it could be done in opam-client, but definitely not core and repository, at least at this stage.

Fair point. Fixed for `core`. `repository`'s API isn't change, only the implementation so it's fine to change.

kit-ty-kate:
Almost all green… Does anyone know what's wrong with the cold check?

dra27:
Ahem - 3/5 jobs failing on AppVeyor is not almost all green 😉

dune
```diff
@@ -1 +1 @@
-(ignored_subdirs (bootstrap))
+(dirs :standard \ bootstrap)
```
dra27, line 1: I assumed these two would behave equivalently, but perhaps this should in fact be `(data_only_dirs bootstrap)`

kit-ty-kate: let's try ^^

kit-ty-kate: it didn't work :(

dra27: Your latest commit looks like it should work, thanks - I shouldn't have ever switched that to `vendored_dirs`... the point is that it's supposed to ignore any of the pkg build directories (because they'll be dirty) and it should find the packages in `bootstrap/` because that's the system root.

kit-ty-kate:
appveyor didn't start :sob: Rebasing..

kit-ty-kate:
All green!

kit-ty-kate (approved):

src/client/opamConfigCommand.mli
```diff
@@ -21,9 +21,9 @@ open OpamStateTypes
     default). [inplace_path] changes how the PATH variable is updated when there
     is already an opam entry: either at the same rank, or pushed in front. *)
 val env:
-  'a global_state -> switch ->
   ?set_opamroot:bool -> ?set_opamswitch:bool ->
   csh:bool -> sexp:bool -> fish:bool -> inplace_path:bool ->
+  'a global_state -> switch ->
```
dra27, line 26: Given that it's just the one place, can we go down the path of greatest caution and just suppress warning 16 in the ML as well?

kit-ty-kate: Done

dra27:
Thanks for the testing & finishing it off, @kit-ty-kate!

PR #4438, opened 2020-11-18
Some fixes in install.sh

### LINKED ISSUE #3927: install.sh should allow use of ~ in install path
Run `shell/install.sh` and, when prompted for the install directory, type `~/.local/bin` (or any other directory that exists under your home dir) and note that the installer claims that the directory doesn't exist:

```
## Downloading opam 2.0.5 for linux on x86_64...
## Downloaded.
## Where should it be installed ? [/usr/local/bin] ~/.local/bin 
## ~/.local/bin does not exist. Create ? [Y/n] n
## Where should it be installed ? [/usr/local/bin] /home/cjs/.local/bin
## opam 2.0.5 installed to /home/cjs/.local/bin
## Run this script again with '--restore ' to revert.
```

rjbou:
* Add `--dowload-only` in help message
* Resolve `~` for bindir using `eval` (fix #3927)

### DIFF
--- a/shell/install.sh
+++ b/shell/install.sh
@@ -378,6 +378,19 @@ while true; do
 
     if [ -d "$BINDIR" ]; then break
     else
+        if [ "${BINDIR#\~/}" != "$BINDIR" ] ; then
+            RES_BINDIR="$HOME/${BINDIR#\~/}"
+            printf "## '$BINDIR' resolves to '$RES_BINDIR', do you confirm [Y/n] "
+            read R
+            case "$R" in
+                ""|"y"|"Y"|"yes")
+                   BINDIR="$RES_BINDIR"
+                   if [ -d "$BINDIR" ]; then break; fi
+                   ;;
+                *)
+                   ;;
+            esac
+        fi
         printf "## $BINDIR does not exist. Create ? [Y/n] "
         read R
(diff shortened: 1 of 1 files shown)


shell/install.sh
```diff
@@ -291,6 +292,19 @@ while true; do
...
+            read R
+            case "$R" in
+                ""|"y"|"Y"|"yes")
+                   BINDIR="$RES_BINDIR"
+                   if [ -d "$BINDIR" ]; then break; fi
+                   ;;
+                *)
+                   ;;
+            esac
+        fi
```
dra27, line 307: This wouldn't allow just `~`, but I think that's OK?
```suggestion
        if [ "${BINDIR#\~/}" != "$BINDIR" ] ; then
            RES_BINDIR="$HOME/${BINDIR#\~/}"
            printf "## '$BINDIR' resolves to '$RES_BINDIR', do you confirm [Y/n] "
            read R
            case "$R" in
                ""|"y"|"Y"|"yes")
                   BINDIR="$RES_BINDIR"
                   if [ -d "$BINDIR" ]; then break; fi
                   ;;
                *)
                   ;;
            esac
        fi
```

rjbou:
Tested, resolves well

PR #4441, opened 2020-11-19
Treat os-family=ubuntu as os-family=debian

alan-j-hu:
At least on Linux Mint 20, the value of `os-family` is `ubuntu`. This caused `opam depext` to fail for me, as `ubuntu` is not one of the handled cases. This PR treats `os-family=ubuntu` as `os-family=debian`.


dra27 (approved):
LGTM, thanks!

rjbou (approved):

kit-ty-kate (approved):

rjbou:
Thanks!

Issue #3964, opened 2019-08-20
opam source --dev-repo: add --shallow-clone option


aantron:
Lwt has a script for fetching the code of all of its opam revdeps: https://github.com/ocsigen/lwt/blob/master/src/util/fetch-dependees.py

It would save a lot of time while running this script, if there was an `opam source --dev-repo --shallow-clone` option, which passed `--depth 1` on the git command line.

PR #4442, opened 2020-11-19
Use --depth=1 for Git repos

dra27:
Triggered from [this Discuss comment](https://discuss.ocaml.org/t/removing-fat-compiler-images-on-hub-docker-com-r-ocaml-opam2/6658/6). This PR acts as a tracking "issue" for further discussion. I don't think in the repo backends we _need_ the Git history for the repository being checked out - we just need it to be possible to update it?

So this adds `--depth=1` to the git-fetch call. This reduces opam-repository as a Git clone, for example, from ~95MiB storage to ~15MiB.

But there may be other consequences...

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -73,7 +73,7 @@ users)
   *
 
 ## Repository management
-  *
+  * Pass --depth=1 to git-fetch in the Git repo backend [#4442 @dra27]
 
 ## VCS
   *
--- a/src/repository/opamGit.ml
+++ b/src/repository/opamGit.ml
@@ -93,7 +93,7 @@ module VCS : OpamVCS.VCS = struct
           OpamFilename.write alternates
             (OpamFilename.Dir.to_string (cache / "objects")))
       global_cache;
-    git repo_root [ "fetch" ; "-q"; origin; "--update-shallow"; refspec ]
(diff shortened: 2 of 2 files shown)


rjbou:
related to #3964

dra27:
Indeed - it's the overlap with git pins which concerns me 🙂

rjbou:
Yup! I take a quick look up and it should be ok, but i'll confirm that by a closer look up (and I saw some maybe enhancement to do).

AltGr:
Don't remember exactly — sorry —, but I think I tried that a while ago and it was broken. There are many operations that are not allowed on shallow clones ; this would also probably fail to leverage the shared git object cache. And compatibility with older versions of `git` should also be taken into account.

avsm:
git 1.9 and greater have mostly fixed the shallow clone issues -- we will need a git version check to pass --depth I suspect for older distros (centos 7)

dra27:
Isn't there already a fallback if the command fails?

AltGr:
We might want to try and merge this after 2.1 is out to be sure it is tested thoroughly by the time 2.2 gets in beta :)

dra27:
Another hardening would be to check the version of Git the first time it's executed - every version of Git ever released will have supported `--version`

kit-ty-kate (approved):

PR #4447, opened 2020-11-24
Allow git fetch to ask for password

Blaisorblade:
Problem: `git fetch` can ask for a password, but that fails without `allow_stdin`. Tentative fix: pass `allow_stdin`.
- [ ] Untested as `opam` fails to compile, so take it as pseudocode.
- [ ] Please update `master_changes.md` file with your changes.

### DIFF
--- a/src/repository/opamGit.ml
+++ b/src/repository/opamGit.ml
@@ -29,8 +29,8 @@ module VCS : OpamVCS.VCS = struct
     (* If the ?env arg is restored here, then the caching for the Cygwin-ness
        of git will need to change, as altering PATH could select a different
        Git *)
-    fun ?verbose ?stdout args ->
-      OpamSystem.make_command ~dir ?verbose ?stdout "git" args
+    fun ?allow_stdin ?verbose ?stdout args ->
+      OpamSystem.make_command ~dir ?verbose ?stdout ?allow_stdin "git" args
 
   let init repo_root repo_url =
     OpamFilename.mkdir repo_root;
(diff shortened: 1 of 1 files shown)


Blaisorblade:
The testing failure was due to https://github.com/ocaml/opam/issues/4448; backporting to 2.0 fails because 2.0 doesn't build on dune 2.7.1.

AltGr:
At the very least you will need to enforce `download_jobs=1` for this to be a viable option. It's actually annoying that we can't let `git` understand that it shouldn't ask for password just by providing it with a closed `stdin`... but we haven't found a workaround for that.

Blaisorblade:
FWIW, the status here is that I don’t know what `download_jobs=1` means.

dra27:
> FWIW, the status here is that I don’t know what `download_jobs=1` means (I should probably check).

You'd want to set `OPAMDOWNLOADJOBS=1` before running, otherwise you could end up with multiple jobs in parallel prompting.

Git does have `core.askPass`, so this, for example, errors properly: `git -c core.askPass=echo https://github.com/dra27/my-secret-hidden-repo`, but we shouldn't go doing that by default because it also stops my perfectly valid and non-interactive credential manager from jumping in with a token!

Another option might be to open the Git process with a pipe to stdin and write a few newlines to it. I'm more focused on wanting it to fail in these cases, not ever prompt for a password...

dra27:
We discussed this PR this morning. Thanks for your work on this, however our conclusion is that this isn't the direction to go in. We'd rather aim to ensure (as far as `git`, etc. allow) that opam doesn't _block_ on these commands - i.e. instructing the commands being called that they're not permitted to be interactive.

PR #4463, opened 2020-12-02
Github actions: linux & macos

rjbou:
Divided in several jobs:
 * cache for ocaml system compilation
 * cache for src_ext archives retrieving
 * build on linux 4.0.2.3 -> 4.11.1 & macos 4.11.1
 * test (opam-rt) on linux/macos 4.11.1 with an embedded compiler build cache & an opam-rt cache
 * upgrade 1.2 -> 2.0 test on linux 4.11.1, with an embedded cache for opam 1.2 root
 * make cold test

Travis is not yet disabled, those scripts need to be stress tested.

Jobs are imitating travis ones, ie:
* all linux build to check compiler compliance
* on macos build to check macos compliance
* one test job per platform (we need to introduce here further reftests and disable make tests from main build jobs)
* one make cold job on linux, to test make cold (need on macos also ?)

And:
* builds with fail fast, no need to build them all if there is a compilation issue
* test/cold/upgrade jobs depends on build jobs success, as there is no fail fast between jobs (and no need to run jobs if in the main there is a compilation issue)

### DIFF
--- a/.github/scripts/hygiene.sh
+++ b/.github/scripts/hygiene.sh
@@ -1,0 +1,146 @@
+#!/bin/bash -xue
+
+PR_COMMIT_RANGE=
+if [ "$GITHUB_EVENT_NAME" = "pull_request" ]; then
+  PR_COMMIT_RANGE="$GITHUB_REF_SHA...$GITHUB_SHA"
+fi
+CI_BRANCH=${GITHUB_REF##*/}
+
+
+echo "PR_COMMIT_RANGE=$PR_COMMIT_RANGE"
+echo "GITHUB_SHA=$GITHUB_SHA"
+if [[ $GITHUB_EVENT_NAME = 'pull_request' ]] ; then
+  FETCH_HEAD=$(git rev-parse FETCH_HEAD)
+  echo "FETCH_HEAD=$FETCH_HEAD"
+else
+  FETCH_HEAD=$GITHUB_SHA
+fi
+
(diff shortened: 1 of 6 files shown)


dra27:
I wonder if it's worth breaking the YAML file up into a few more workflows - for example, having "Build / ", "Caches / ", "Tests / " (which could include the change log checker too)? It's only for readability of the Actions files!

I think it would be worth merging this (the AppVeyor failure is unrelated) and reviewing the scripts "after-the-fact" just so that this starts being tested on PRs?

rjbou:
It's double-edged to split the yaml (and yes, i agree, it's not an easy readable one). It's not possible to include a workflow from another one, so we will need to duplicate all the common yaml (cache jobs, environment variables). We can't use actions to simplify as it is not yes possible to call an action inside an action (and the yaml subset of actions is less expressive). And there is no support of yaml aliases ftm. There is already a lot of duplication in the main yaml, duplicating on several ones will be more error-prone (especially for caches keys, as it is not possible to delete a cache we need to be sure that it will be rebuild when one of its component changes).

dra27:
That's annoying! Ah well - in which case, let's leave it as it is... other CIs are available, longer term 🙂

rjbou:
It's really an on-developement project github actions. There was a certain amount of things I wanted to do that were not possible, and paths give an headache if you don't use them the straight way (bash don't help neither :)).

rjbou:
don't wait for travis & appeyor, as they are not concerned, merging

PR #4470, opened 2020-12-09
Upgrade to ocamlgraph 2.0.0, dune 2.7, dose3 6.0.2

kit-ty-kate:


dra27:
It's not necessarily a problem, but is the upgrading of opam's dune files to 2.0 necessary to fix something?

kit-ty-kate:
currently ocamlgraph 2.0.0 requires dune 2.0 and dose3 6.0.2 requires dune 2.7

dra27:
Ah, handy! The ocamlgraph one doesn't matter (because you could install 1.8.8) but the Dose3 need for 2.7 is a handy excuse to jump with it 🙂

kit-ty-kate:
dose3 6.0.2 also requires ocamlgraph >= 2.0.0, so the two packages are linked

dra27:
All the upgrades 🙂

kit-ty-kate:
arf I forgot dose3 6.0 now depends on camlzip and camlbz2... I think this brings too many requirements for opam, I'll talk with them to try to see if they can make it optional

cheese:
Hope for this. Otherwise opam has to be removed from Fedora official repo.

dra27:
@cheese - this might be worth opening a separate issue for. These dependency upgrades will only be for opam 2.2+ There is a problem that in its present packaging the dose3 6.x no longer builds on Windows.

Compilation support for ocamlgraph 2.0.0 would be very easy to back-port, but the dose3 6.x problem looks like it will require upstream assistance for us to adopt it all. It would certainly be ironic for a an ocaml package which is exclusively used by opam to be the cause of its removal from Fedora 😣

cheese:
> @cheese - this might be worth opening a separate issue for. These dependency upgrades will only be for opam 2.2+ There is a problem that in its present packaging the dose3 6.x no longer builds on Windows.

I just make a comment to subscribe the conversation. It is not an emergency. It is OK to get it done at any time before 2022Q2.

kit-ty-kate:
Since the issue is with dose, please follow https://gitlab.com/irill/dose3/-/issues/15 instead and its two possible fixes:
- https://gitlab.com/irill/dose3/-/merge_requests/23
- https://gitlab.com/irill/dose3/-/merge_requests/22

kit-ty-kate:
Superseded by https://github.com/ocaml/opam/pull/4760

kit-ty-kate:
@cheese #4760 has now been merged in master. The next major release will have it

cheese:
> @cheese #4760 has now been merged in master. The next major release will have it

Thanks! Building of 8267b40fc succeeded on Fedora 35 (Rawhide). Hope for the release of 2.2.0!

Issue #4468, opened 2020-12-09
Github actions todo list


rjbou:
* [ ] Add reftest run in `test` job for macos
* [ ] Add `cold` job for macos
* [ ] Add windows jobs
* [ ] Cache lib-ext build with key os, ocaml version, & src_ext

PR #4475, opened 2020-12-14
Fix MacOS upgrade CI test using OS-specific opam 1.2 cache

### LINKED ISSUE #4474: MacOS upgrade CI test fails when executed first on clean opam 1.2 cache
A recently added CI test (introduced in #4463) failed on my pull-request branch:
https://github.com/freevoid/opam/runs/1540409759?check_suite_focus=true

The test in question is `upgrade (macos-latest, 4.11.1)` and the error is due to the following behavior:

 * There is a cache for opam 1.2.2 binary (that is used to test the upgrade command) and this cache is *not keyed by `runner.os`*
 * The logic in `main.sh` fetches opam 1.2.2 binary using hardcoded URL `wget https://github.com/ocaml/opam/releases/download/1.2.2/opam-1.2.2-x86_64-Linux -O /Users/runner/.cache/opam1.2/cache/bin/opam`
 * The upgrade test is executed for Linux and MacOS and the 1.2.2 cache is shared
 * If MacOS action is executed *first* and there is no cached opam 1.2.2 binary, the error happens while trying to execute `opam init` using the Linux version of the binary
 * If Linux action is executed first, however, the action succeeds and the MacOS action picks up the cache and executes successfully too (because if the cached opam 1.2 root is present, old binary is not executed, or, to be precise, it is only executed to show the 1.2 version as silently fails to do so while continuing the test).

The exempt from the logs:

```
++ eval echo '~/.cache/opam1.2/cache'
+++ echo /Users/runner/.cache/opam1.2/cache
+ OPAM12CACHE=/Users/runner/.cache/opam1.2/cache
+ OPAM12=/Users/runner/.cache/opam1.2/cache/bin/opam
+ [[ ! -f /Users/runner/.cache/opam1.2/cache/bin/opam ]]
+ mkdir -p /Users/runner/.cache/opam1.2/cache/bin
+ wget https://github.com/ocaml/opam/releases/download/1.2.2/opam-1.2.2-x86_64-Linux -O /Users/runner/.cache/opam1.2/cache/bin/opam
--2020-12-11 23:37:52--  https://github.com/ocaml/opam/releases/download/1.2.2/opam-1.2.2-x86_64-Linux
Resolving github.com (github.com)... 140.82.112.4
...
HTTP request sent, awaiting response... 200 OK
Length: 6265225 (6.0M) [application/octet-stream]
Saving to: ‘/Users/runner/.cache/opam1.2/cache/bin/opam’

     0K .......... .......... .......... .......... ..........  0% 3.16M 2s
...
  6100K .......... ........                                   100%  159M=0.2s

2020-12-11 23:37:52 (32.5 MB/s) - ‘/Users/runner/.cache/opam1.2/cache/bin/opam’ saved [6265225/6265225]

+ chmod +x /Users/runner/.cache/opam1.2/cache/bin/opam
+ export OPAMROOT=/tmp/opamroot
+ OPAMROOT=/tmp/opamroot
+ rm -rf /tmp/opamroot
+ [[ ! -d /Users/runner/.cache/opam1.2/cache/root ]]
+ /Users/runner/.cache/opam1.2/cache/bin/opam init
.github/scripts/main.sh: line 113: /Users/runner/.cache/opam1.2/cache/bin/opam: cannot execute binary file
Error: Process completed with exit code 126.
```

freevoid:
This PR is aimed to fix #4474 by

 1) using `$RUNNER_OS` to build a OS-specific URL for opam 1.2 binary (albeit only Darwin and Linux releases are available so Windows is not supported)
 2) using `runner.os` when building a cache key for opam 1.2 root (so that MacOS will not try to re-use the linux binary and vice versa depending on which action ran first when the cache was empty)

Note that since the test still passes if the cache was built previously, we might be fine with just implementing 1, but this will lead to not-deterministic test behavior and can bring sorrow down the line.

### DIFF
--- a/.github/scripts/main.sh
+++ b/.github/scripts/main.sh
@@ -104,7 +104,12 @@ if [ $OPAM_UPGRADE -eq 1 ]; then
   OPAM12=$OPAM12CACHE/bin/opam
   if [[ ! -f $OPAM12 ]]; then
     mkdir -p $OPAM12CACHE/bin
-    wget https://github.com/ocaml/opam/releases/download/1.2.2/opam-1.2.2-x86_64-Linux -O $OPAM12
+
+    os="Linux"
+    if [ "$RUNNER_OS" = "macOS" ]; then
+      os="Darwin"
+    fi
+    wget "https://github.com/ocaml/opam/releases/download/1.2.2/opam-1.2.2-x86_64-$os" -O $OPAM12
     chmod +x $OPAM12
   fi
   export OPAMROOT=/tmp/opamroot
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -199,7 +199,7 @@ jobs:
       uses: actions/cache@v2
       with:
         path: ${{ env.OPAM12CACHE }}
-        key: opam1.2-root
+        key: ${{ runner.os }}-opam1.2-root
     - name: ocaml ${{ matrix.ocamlv }} cache
       id: ocaml-cache
       uses: actions/cache@v2
--- a/master_changes.md
+++ b/master_changes.md
@@ -128,6 +128,7 @@ New option/command/subcommand are prefixed with ◈.
   * patcher: fix local [#4467 @AltGr]
   * Add github actions [#4463 @rjbou]
   * Add reftests to github actions [#4467 @rjbou]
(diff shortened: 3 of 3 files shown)


rjbou:
Thanks! Seems that a commit disappeared, somewhere I've already added the runner in the key.

rjbou:
related #4468

PR #4482, opened 2020-12-18
init: fix non-existing file check in csh

gahr:
This fixes an issue whereas disabling the hook script by answering "no"
to the 2nd question of "opam init" you end with an init script that
tries to source the non-existing env hook script.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -15,6 +15,7 @@ New option/command/subcommand are prefixed with ◈.
   * Fix sandbox check with not yet set opam environment variables [#4370 @rjbou - fix #4368]
   * Sandboxing check: use configured temp dir and cleanup afterwards [#4467 @AltGr]
   * Print shell-appropriate eval command on `opam init` [#4427 @freevoid]
+  * Fix non-existing file check in csh [#4482 @gahr]
 
 ## Config Upgrade
   *
--- a/src/state/opamEnv.ml
+++ b/src/state/opamEnv.ml
@@ -472,18 +472,18 @@ let env_hook_script shell =
     (env_hook_script_base shell)
 
 let source root shell f =
-  let file f = OpamFilename.to_string (OpamPath.init root // f) in
+  let fname = OpamFilename.to_string (OpamPath.init root // f) in
   match shell with
   | SH_csh ->
-    Printf.sprintf "if ( -f %s ) source %s >& /dev/null\n" (file f) (file f)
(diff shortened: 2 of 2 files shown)


gahr:
I guess the failure in appveyor is not related to my change

```
C:/projects/opam/bootstrap/ocaml/lib/ocaml/caml/misc.h:409:18: error: static declaration of ‘caml_snprintf’ follows non-static declaration
  409 | #define snprintf caml_snprintf
      |                  ^~~~~~~~~~~~~
C:/projects/opam/bootstrap/ocaml/lib/ocaml/caml/misc.h:408:12: note: previous declaration of ‘caml_snprintf’ was here
  408 | extern int caml_snprintf(char * buf, size_t size, const char * format, ...);
      |            ^~~~~~~~~~~~~
```

rjbou:
Thanks for the PR.
On the issue you mention, do you have a failing log ?
Because it's not important that the script is written or no. In the `init.sh` (that is sourced in `*shrc`), there is a [test](https://github.com/ocaml/opam/blob/master/src/state/opamEnv.ml#L482) on the availability of the script to source.
A `sh` generated one gives:
```
$ cat /tmp/opamroot/opam-init/init.sh
if [ -t 0 ]; then
  test -r /tmp/opamroot/opam-init/complete.sh && . /tmp/opamroot/opam-init/complete.sh > /dev/null 2> /dev/null || true

  test -r /tmp/opamroot/opam-init/env_hook.sh && . /tmp/opamroot/opam-init/env_hook.sh > /dev/null 2> /dev/null || true
fi

test -r /tmp/opamroot/opam-init/variables.sh && . /tmp/opamroot/opam-init/variables.sh > /dev/null 2> /dev/null || true
```

gahr:
Ah thanks, I hadn't noticed the attemp to handle a non-existing file in init.(c)sh. That attempt is broken for csh, though.

https://github.com/ocaml/opam/blob/8ce5e420f8007fd8d7ad56fb6293af6f5469a432/src/state/opamEnv.ml#L478.
However, according to csh's manpage, "An error in a source at any level terminates all nested source commands."

It tries to handle the error using `|| true`, but in csh (citing the man page) "An error in a source at any level terminates all nested source commands." This causes an error in the startup file `~/.cshrc`.

I'll update the PR to properly check for the file existence in `init.csh`.

gahr:
@rjbou need anything from my side?

dra27:
Thanks!

rjbou:
Oops, forgot to see the new version. lgtm, thanks @gahr!

PR #4483, opened 2020-12-21
Restore mingw CI testing

dra27:
mingw-w64 8.0.0 requires some changes in OCaml. These will be in 4.11.2 and 4.12.0, but in the meantime this PR sorts out the AppVeyor workers to eliminate the failures on PRs.

The bump of FlexDLL 0.39 is not necessary to restore the support, but it has the benefit of clearing the AppVeyor caches and rebuilding the compilers!


dra27:
Spectrum is Green!

rjbou:
Thanks!

PR #4492, opened 2021-01-06
Set OPAMCLI=2.0 by default during compilation/install/remove phases

### LINKED ISSUE #4491: [opam-2.1~beta4] Automatically add OPAMCLI=2.0 when package has opam-version: "2.0"
Just an idea. This might be a blocker for beta4.
Idea brought up in https://github.com/ocaml/opam-repository/pull/17927#issuecomment-755055186

kit-ty-kate:
People who want to use the 2.1 features should be using `--cli=2.1` or `build-env: OPAMCLI = "2.1"`

Tested successfully using the following opam files:
```
opam-version: "2.0"
build: ["env"]
install: ["env"]
remove: ["env"]
```
```
opam-version: "2.0"
build-env: [OPAMCLI = "2.1"]
build: ["env"]
install: ["env"]
remove: ["env"]
```

Fixes #4491

### DIFF
--- a/tests/reftests/cli-versioning.test
+++ b/tests/reftests/cli-versioning.test
@@ -52,7 +52,7 @@ The following actions would be performed:
 ===== ∗ 6 =====
 ### opam install ocaml.4.10.0 --unlock-base --show
 opam: unlock-base was removed in version 2.1 of the opam CLI, but version 2.1 has been requested. Use --update-invariant instead or set OPAMCLI environment variable to 2.0.
-### #opam option use mk_command_ret
+### # opam option uses mk_command_ret
 ### opam option foo
 [ERROR] No option named 'foo' found. Use 'opam option [--global]' to list them
 ### OPAMCLI=2.0 opam option foo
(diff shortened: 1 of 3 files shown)


kit-ty-kate:
I've been trying to add tests for this, but I'm unsure where I should put it. Any idea?

rjbou:
I've added a test in cli versioning one, can you check that it is exhaustive, what you had in mind to test ?
I'm not completely convinced with those test as they need an opam system installed with cli versioning, I'd prefer not to depend on that (even if it will be the usual case).

dra27:
LGTM, thank you! 

Note that `OPAMCLI=2.1` should never go in a released file (apart from anything else, it generates a warning). All new package descriptions should expect to use `--cli=2.1`

dra27:
I haven't reviewed the reftests PR yet - do they use the system-installed `opam` at any point rather than the one just built?

tests/reftests/cli-versioning.test
```diff
@@ -61,8 +61,153 @@ opam: option was added in version 2.1 of the opam CLI, but version 2.0 has been
...
 opam: lock was added in version 2.1 of the opam CLI, but version 2.0 has been requested, which is older.
+### # Check for build test env
+### # Note: you must have an installed opam with cli version enabled to pass these tests
+### mkdir opams
+### <opams/ok-var-2-0.opam>
+opam-version: "2.0"
+build: [ "opam" "config" "var" "prefix" "--readonly" ]
+### <opams/ok-var-2-0-cli.opam>
+opam-version: "2.0"
+build: [ "opam" "config" "var" "prefix" "--cli=2.0" "--readonly" ]
```
kit-ty-kate, line 77:
```suggestion
### <opams/ok-var-2-0.opam>
opam-version: "2.0"
build: ["sh" "-c" "env | grep -q '^OPAMCLI=2.0$'"]
install: ["sh" "-c" "env | grep -q '^OPAMCLI=2.0$'"]
remove: ["sh" "-c" "env | grep -q '^OPAMCLI=2.0$'"]
### <opams/ok-var-2-0-cli.opam>
opam-version: "2.0"
build-env: [OPAMCLI = "2.1"]
build: ["sh" "-c" "env | grep -q '^OPAMCLI=2.0$'"]
install: ["sh" "-c" "env | grep -q '^OPAMCLI=2.0$'"]
remove: ["sh" "-c" "env | grep -q '^OPAMCLI=2.0$'"]
```

> I've added a test in cli versioning one, can you check that it is exhaustive, what you had in mind to test ?
> I'm not completely convinced with those test as they need an opam system installed with cli versioning, I'd prefer not to depend on that (even if it will be the usual case).

I had more in mind something like that, just to make sure that the environment variable is correctly set.

rjbou:
@dra27 I don't think. In reftests, it's the newly compiled opam that is used to launch command to test, a substitution is done. But in  test added in this PR, I want to test an opam call inside an opam package, and these call use the system installed opam. The requirement for the installed opam is to have cli versioning working because test output is based on its error messages.

tests/reftests/cli-versioning.test
rjbou: Is it enough to test the presence of the variable in env ? As it is set for opam calls, I tested opam calls. In the other hand, it permit to not rely on a system installed opam requirement...

kit-ty-kate:
I've changed the tests. Feel free to add more if you feel like it should but as is I think this should be good to merge.

AltGr:
This seems sensible; my worry is that we really don't want to encourage calls to `opam` from within opam packages.

kit-ty-kate:
> my worry is that we really don't want to encourage calls to `opam` from within opam packages.

I think it is way too late to worry about this right now. People started using it this way years ago.
This PR is about adding a sane default to opam cli system and allow hundreds of packages in opam-repository to compile again with opam.2.1.0~beta4 (coming today?)

avsm:
I agree with @kit-ty-kate -- it's use is common (even though we want to discourage it).  I'm in favour of this change.

avsm:
Just one note: this definitely needs to be documented in the manual somewhere.

kit-ty-kate:
Good point. There was no section documenting the default environment variables so I've documented this in https://github.com/ocaml/opam/pull/4496. (separately because I'd rather avoid blocking this PR and the release on documentation)

dra27:
I also agree - the stable door is definitely off its hinges! CLI versioning and configuration information in the build are orthogonal features (even if CLI versioning does make _using_ the opam command in build scripts less brittle)

dra27:
Thanks both for the idea and the PR, @kit-ty-kate!

Issue #4422, opened 2020-11-08
[optimization] Only check for changes on the filesystem when necessary


kit-ty-kate:
Currently opam check for changes in the `.opam` after every single package installation. This produces a lot of IO.
However when the package being installed does not have any `install` field and sandbox is on this is totally unnecessary.

~Two solution could a considered:~
1) ~If we decide that the sandbox must be enabled, here is what is required to get this working internally:~
* [ ] ~a way for opam to know if the sandbox is enabled (currently the sandbox is simply a `wrap-*-command`)~
* [ ] ~a way for opam to ensure the sandbox script is unmodified~

~Since this requires the sandbox to be there, only Linux and MacOS would be eligible to this.~

2) Another solution would be to disregard whether or not the sandbox is enabled and do this anyway as a package trying to cross the sandbox is a packaging issue and would fail anyway on systems with sandbox enabled.

*Original idea by @Armael*

kit-ty-kate:
The first solution already goes out of the window when considering parallel builds. Parallel installs would detect things installed during the build phase of the potential faulty package and mess things up anyway.

The second solution of simply disabling scanning of `opamroot` when a package has no `install` field is still possible

dra27:
Fixed by #4494

PR #4494, opened 2021-01-08
Make install faster by not scanning the whole switch when no install field is present

kit-ty-kate:
Implements https://github.com/ocaml/opam/issues/4422

Currently opam scans the whole switch directory recursively twice when installing a package (per package). This introduces a huge strain on the IO of the machine. This is especially visible under high load (e.g. in CI)

This PR makes so that opam only scan the files that are going to be installed (if they exist already, this is rare) if the package does not have any `install` field. If the package does have an `install` field the old behaviour is used.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -26,6 +26,7 @@ New option/command/subcommand are prefixed with ◈.
   * Don't patch twice [#4529 @rjbou]
   * With `--deps-only`, set dependencies as root packages [#4964 @rjbou - fix #4502]
   * Keep global lock only if root format upgrade is performed [#4612 @rjbou - fix #4597]
+  * Improve installation times by only tracking files listed in `.install` instead of the whole switch prefix when there are no `install:` instructions (and no preinstall commands) [#4494 @kit-ty-kate @rjbou - fix #4422]
 
 ## Remove
   *
--- a/src/core/opamDirTrack.mli
+++ b/src/core/opamDirTrack.mli
@@ -41,6 +41,13 @@ val track:
   OpamFilename.Dir.t -> ?except:OpamFilename.Base.Set.t ->
   (unit -> 'a OpamProcess.job) -> ('a * t) OpamProcess.job
 
+(** [track_files prefix paths ?except job] as [track] wraps a job to track
+    changes for a predefined list of [paths] (files and directories).
+    [paths] are relative to [prefix]. *)
+val track_files:
+  prefix:OpamFilename.Dir.t -> string list -> ?except:OpamFilename.Base.Set.t ->
+  (unit -> 'a OpamProcess.job) -> ('a * t) OpamProcess.job
+
 (** Removes the added and kind-changed items unless their contents changed and
     [force] isn't set, and prints warnings for other changes unless [verbose] is
     set to [false]. Ignores non-existing files.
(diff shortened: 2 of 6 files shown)


kit-ty-kate:
This PR is now ready for review. cc @AltGr

AltGr (commented):
This is more complicated than what I would have thought at first; I was initially pretty confused by `track_files` comments saying it is "similar to [track]", while the latter is a wrapper around the execution of a job and the former just an atomic function that expects the results of a previous run as argument; there is nothing wrong with that, but what were the reasons not to go for the more straight-forward extraction of the list of files from `.install`, followed by the run of a version of `track` that is limited to a given set of files ? It would be useful to document more precisely the usage of `track_files` in `opamDirTrack.mli`.

Another remark, it seems the behaviour on `pre-install` hooks is changed, since they are no longer tracked. Have you studied if that had any consequences in the wild ? Though the only other option would be to disable the optimised tracking as soon as any `pre-install` hook is defined, I guess... The tracking behaviour for `post-install` is already specific and detailed [here](http://opam.ocaml.org/doc/Manual.html#configfield-post-install-commands).

I am otherwise happy with the code and this will be nice to have!

rjbou:
Just to keep a trace of the test (can't be added in reftest):
```
$ cat dotty/dot.install
bin: [ "script" ]
$ cat dotty/dot.opam
opam-version: "2.0"
build: [ "sh" "-c" "echo hellow > script%{exe}%" ]
$ opam pin -yn ./dotty
$ opam install dot #with win executable name
<><> Synchronising pinned packages ><><><><><><><><><><><><><><><><><><><><><><>
[dot.0.1] synchronised (no changes)

  ∗ install dot 0.1*

[WARNING] .install file is missing .exe extension for script
[WARNING] Automatically adding .exe to <switchprefix>/build/dot.0.1/script.exe
∗ installed dot.0.1
Done.
```

master_changes.md
```diff
@@ -21,6 +21,7 @@ New option/command/subcommand are prefixed with ◈.
   * Don't patch twice [#4529 @rjbou]
   * With `--deps-only`, set dependencies as root packages [#4964 @rjbou - fix #4502]
   * Keep global lock only if root format upgrade is performed [#4612 @rjbou - fix #4597]
+  * Only track `.install` content when install field is empty instead of whole switch directory, makes install faster [#4494 @kit-ty-kate @rjbou - fix #4422]
```
AltGr, line 24: I can't parse this sentence :rofl: 
```suggestion
  * Improve installation times by only tracking files listed in `.install`  instead of the whole switch prefix when there are no `install:` instructions [#4494 @kit-ty-kate @rjbou - fix #4422]
```

src/core/opamDirTrack.mli
```diff
@@ -41,6 +41,11 @@ val track:
   OpamFilename.Dir.t -> ?except:OpamFilename.Base.Set.t ->
   (unit -> 'a OpamProcess.job) -> ('a * t) OpamProcess.job
 
+val track_t:
+  [< `Top of OpamFilename.Dir.t | `Specific of OpamFilename.Dir.t * string list ] ->
```
AltGr, line 45: It's generally considered better practice for interfaces such as this one to just export two functions rather than complexify the arguments (even if they will both call into the same code, of course).

AltGr (approved):
Just a teeny bit of nitpicking, otherwise all good! ;)

PR #4497, opened 2021-01-12
Restrict to opam-file-format 2.1.2

rjbou:


dra27 (approved):
LGTM! The module link order (ocaml/opam-file-format#41) is a nuisance, but it has a clear workaround for opam-repository (install Dune!), so I don't think we need to release it hot on the heels of 2.1.2.

kit-ty-kate:
Rebased on top of master (there was a conflict in the changelog). I'll merge it once it passes CI

kit-ty-kate:
All green. Thanks!

PR #4500, opened 2021-01-14
Fix opam-devel's tests on platforms without openssl, GNU-diff and a system-wide ocaml

kit-ty-kate:
On some systems, opam's tests might be called by opam itself.
However, opam seems to hide parts of the `PATH` that looks like an opam `bin` directory.

To counteract that the tests need to copy `ocamlc` and `ocamlopt` directly before calling the local opam binary.

`openssl` and GNU-`diff` was also missing from the list of test dependencies. For instance those are not installed by default on Alpine.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -60,7 +60,7 @@ New option/command/subcommand are prefixed with ◈.
   *
 
 ## Build
-  *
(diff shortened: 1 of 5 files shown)


kit-ty-kate:
There seems to also be an issue when compiled inside opam 2.1, I'll have a look

rjbou:
Thanks!

PR #4506, opened 2021-01-21
Mention pre-built opam for Arm Macs in INSTALL.md

eth-arm:
Opam 2.0.7 binaries for `arm64-macos` was added to `install.sh` in commit 037c31036a0f94d43fdc91856830fb248e63e8b3.


rjbou:
Thanks!

PR #4509, opened 2021-01-25
depext: macport variants handling

### LINKED ISSUE #4297: Depext: Issues with variants in MacPorts
I added MacPorts solutions to a few opam conf packages. See: (ocaml/opam-repository#16912)

One issues is that depext doesn't seem to handle variants like "gtk3 +quartz" in a consistent manner. When printing what to install, gtk3 and +quartz are listed on separate lines, but when sorting packages alphabetically it seems to treat them as one. I guess this is just not handled at all explicitly - maybe it should.

It somehow works as is, but it looks like it might break any day by a change.

(See https://github.com/ocaml-opam/opam-depext/issues/128)

rjbou:
Fixes #4297, see discussion there.
~Variants availability is not part of the PR as it is time consuming: one call per system package to check available variants~ One call is done now

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -49,6 +49,7 @@ New option/command/subcommand are prefixed with ◈.
   * Fix `features` parser [#4507 @rjbou]
 
 ## External dependencies
+  * Handle macport variants [#4509 @rjbou - fix #4297]
 
 ## Sandbox
   * Fix the conflict with the environment variable name used by dune [#4535 @smorimoto - fix ocaml/dune#4166]
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -388,26 +388,53 @@ let packages_status packages =
     in
     compute_sets sys_installed
   | Macports ->
+    let variants_map, packages =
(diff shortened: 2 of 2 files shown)


kit-ty-kate:
just in case, cc @MSoegtropIMC who reported the issue

MSoegtropIMC:
Thank's! Is the assumption that opam's CI does test this or should I run a few tests with this opam version manually? It would take me a little while but I think I could setup a Coq Platform branch which takes an arbitrary opam branch. This would make it much easier for me to test this.

kit-ty-kate:
> Thank's! Is the assumption that opam's CI does test this or should I run a few tests with this opam version manually?

Unless I'm mistaken I don't think opam's CI has any tests for macports, only homebrew. AFAIK github actions don't ship with macports

rjbou:
> It would take me a little while but I think I could setup a Coq Platform branch which takes an arbitrary opam branch.

If it is difficult, I can test in the next days on Kate's mac

rjbou:
With `port search --name --line --regex <pkg regex>` for availability checl, depext time down to 1.5s, but still too high

PR #4514, opened 2021-01-28
On install action, reinstall if test/doc dependencies are not present

### LINKED ISSUE #4507: `with-test` deps not installed after second call to opam install
In a local clone of an opam package, I ran a first install in a local package with just `opam install .`.

After this, I realized I had not installed the test dependencies, so I ran the following:
```
$ opam install . --with-test
[NOTE] Ignoring uncommitted changes in /home/javier/code/tyxml (`--working-dir' not active).
[tyxml.dev] no changes from git+file:///home/javier/code/tyxml#master
[NOTE] Package tyxml is already installed (current version is dev).
```

I was surprised that the test dependencies were not installed. @Khady pointed out that I could use `opam reinstall . --with-test`, and that installed the test dependencies successfully.

I guess the current behavior of `install` is expected, but as a user I assumed that calling `install` a second time with a different flag, would install any packages that are missing.

rjbou:
fix  #4507

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -16,7 +16,7 @@ New option/command/subcommand are prefixed with ◈.
   *
 
 ## Install
-  *
+  * Check if with-test & with-doc depends are installed to launch a reinstall [#4513 @rjbou - fix #4507]
 
 ## Remove
   *
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -1103,6 +1103,29 @@ let install_t t ?ask ?(ignore_conflicts=false) ?(depext_only=false)
     if assume_built then OpamPackage.Set.of_list pkg_skip
     else Lazy.force t.reinstall %% OpamPackage.Set.of_list pkg_skip
   in
+  let pkg_skip, pkg_reinstall =
+    if OpamStateConfig.(!r.build_test || !r.build_doc) then
+      let to_reinstall =
+        List.fold_left (fun reinstall nv ->
+            let deps_installed =
+              OpamSwitchState.opam t nv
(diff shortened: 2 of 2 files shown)


rjbou:
Current state : it filter according `with-test` & `with-doc`, not if they are combined with other variables as `build`, `post`, `dev`.

avsm:
Dev meeting: Pushing this out to 2.1.1, we can make sure the right variables are all accounted for. There is a workaround for `opam reinstall` right now that we can release note.

AltGr (commented):
Hm, this should make the issue less pregnant, but isn't a complete solution: indeed, if you `opam install foo --with-doc` and doc-deps were missing, it will install them, and then recompile `foo` (you don't even need to add to `reinstall` in this case, the dependency change will trigger it already ; just removing from `pkg_skip` would have been enough) : you get the doc for `foo` installed as expected. But if the doc dependencies happened to be installed already, you will still get a `nothing to do` without a recompilation with docs enabled.

We don't really have a way to fix this since we don't record (atm) the flags that were set when `foo` was initially installed, so the only options would be to always reinstall when `--with-x` is set (that sounds dangerous ; but it does actually make complete sense in the case of tests ? Maybe after a question like for `opam reinstall <not-installed-pkg>` ?) ; or just advise to use `reinstall` instead ?

kit-ty-kate:
I think this should be closed in favour of https://github.com/ocaml/opam/pull/6209 which fixes the root cause

rjbou:
Issues are not the same, this one is not specific to pinned packages, but more generally to packages with predefined variables that can be specified via cli.

PR #4515, opened 2021-01-29
[WIP] Make opam compatible with dose3 6.0

kit-ty-kate:
Dose_algo.Depsolver.check_request_using actually never used its ~criteria argument(!!) :exploding_head:

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ easier to already have a working opam installation in this case, so you can do
 it as a second step.
 
 * Make sure to have ocamlfind, ocamlgraph, cmdliner >= 0.9.8, cudf >= 0.7,
-  dose3 >= 5, re >= 1.5.0, opam-file-format installed. Or run `opam install
+  dose3 >= 6.0, re >= 1.5.0, opam-file-format installed. Or run `opam install
   . --deps-only` if you already have a working instance. Re-run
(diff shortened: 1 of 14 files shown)


kit-ty-kate:
@abate I've tried dose3 6.0.2 with this branch and I couldn't get opam to work properly. We use dose to show to the user why a package cannot be installed (when it is the case).
However with this branch we do not get any answers (@AltGr any idea how to debug this?).

With dose 5.0.1-1:
```
$ opam install configurator.v0.11.0
[ERROR] Package conflict!
    - (invariant) → ocaml-variants = 4.12.0+trunk → ocaml = 4.12.0
    - configurator >= v0.11.0 → base < v0.14 → ocaml < 4.06.0

No solution found, exiting
```
With dose 6.0.2:
```
$ opam install configurator.v0.11.0
[ERROR] Package conflict!
Sorry, no solution found: there seems to be a problem with your request.

No solution found, exiting
```

Do yo have any idea what has changed in the code?

abate:
maybe we can generate the cudfs and compare them. this should give us an idea I think

kit-ty-kate:
I found the culprit. I made a PR here: https://gitlab.com/irill/dose3/-/merge_requests/18

PR #4525, opened 2021-02-05
Allow to upgrade to a hidden-version package if a hidden-version package is already installed

kit-ty-kate:
Currently if for instance `ocaml-base-compiler.4.12.0~beta1` (flagged with `hidden-version) is installed, calling `opam upgrade` results in:
```
$ opam upgrade
Everything as up-to-date as possible (run with --verbose to show unavailable upgrades).

The following packages are not being upgraded because the new versions conflict with other installed packages:
  - ocaml.4.13.0
However, you may "opam upgrade" these packages explicitly, which will ask permission to downgrade or uninstall the conflicting packages.
Nothing to do.
$ opam switch invariant
["ocaml-base-compiler"]
```
With this patch it should be upgraded to `ocaml-base-compiler.4.12.0~beta2` in this case.

cc @dra27 @AltGr

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -187,8 +187,8 @@ let compute_upgrade_t
              if OpamPackage.Set.exists
                  (fun nv ->
                     OpamFormula.check atom nv &&
-                    not (OpamFile.OPAM.has_flag Pkgflag_AvoidVersion
-                           (OpamSwitchState.opam t nv)))
+                    (not (OpamFile.OPAM.has_flag Pkgflag_AvoidVersion (OpamSwitchState.opam t nv)) ||
+                     OpamSwitchState.can_upgrade_to_avoid_version (OpamPackage.name nv) t))
                  (Lazy.force t.available_packages)
              then atom
              else (n, None)
--- a/src/state/opamSwitchState.ml
+++ b/src/state/opamSwitchState.ml
@@ -815,6 +815,12 @@ let get_conflicts st packages opams_map =
     (fun package -> OpamPackageVar.resolve_switch ~package st)
     packages opams_map
 
+let can_upgrade_to_avoid_version name st =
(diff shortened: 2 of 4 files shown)


kit-ty-kate:
Successfully tested and now ready for review.

src/state/opamSwitchState.ml
```diff
@@ -917,9 +917,19 @@ let universe st
...
   in
+  let has_hidden_version opam =
+    OpamFile.OPAM.has_flag Pkgflag_HiddenVersion opam
+  in
+  let hidden_version_already_installed name =
+    OpamPackage.Set.exists (fun pkg ->
+        OpamPackage.Name.equal (OpamPackage.name pkg) name &&
+        has_hidden_version (OpamPackage.Map.find pkg st.opams)
+      ) st.installed
+  in
```
rjbou, line 928: Why not merge those two functions in one with a single `opam` argument ?

src/client/opamClient.ml
```diff
@@ -184,11 +184,21 @@ let compute_upgrade_t
...
+             let has_hidden_version nv =
+               OpamFile.OPAM.has_flag Pkgflag_HiddenVersion
+                 (OpamSwitchState.opam t nv)
+             in
+             let hidden_version_already_installed pkg =
+               OpamPackage.Set.exists (fun nv ->
+                   OpamPackage.Name.equal (OpamPackage.name nv) pkg &&
+                   has_hidden_version nv
+                 ) t.installed
+             in
```
rjbou, line 196: If it is used in 2 places, it is possible to add a function `OpamSwitchState.has_hidden_version_and_installed` for ex. Like that if a change need to be done, it would be only in one place

rjbou (approved):
lgtm, just some code style comments :)

rjbou:
Thanks!

PR #4477, opened 2020-12-15
Ensure we don't advertise upgrades to hidden versions

AltGr:
Closes #4432 (together with defining unreleased versions of `ocaml` as
hidden).

PR #4527, opened 2021-02-05
Rename hidden-version to avoid-version

dra27:
As per https://github.com/ocaml/opam/wiki/Dev-Meeting-2020.11.27#hidden-version, but apparently I forgot to open the issue to track it!

There were two other points:

- Post-process solutions to see if packages have changed to be hidden and warn (with a default of not installing)
- ocaml.4.12.0 needs to be tagged hidden-version

The second point I think has already been addressed addressed by https://github.com/ocaml/opam/pull/4477. I'll open an issue to track the other.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -47,6 +47,7 @@ New option/command/subcommand are prefixed with ◈.
 
 ## Opamfile
   * Fix `features` parser [#4507 @rjbou]
+  * Rename `hidden-version` to `avoid-version` [#4527 @dra27]
 
 ## External dependencies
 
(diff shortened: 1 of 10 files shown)


kit-ty-kate:
Slight note: if this is to be merge, I think we should merge it and release beta5 before the final release of 4.12.0 next week as it would be annoying for users to have to rebuild their switch twice

dra27:
It's the opposite, isn't it - it's better to merge now that 4.12.0 is released, since no users should have moved off the affected packages?

Issue #4400, opened 2020-10-20
"Ctrl-C" during installation quits opam but the build continues


RalfJung:
>   If your issue concerns a package not building, please report to
>   https://github.com/ocaml/opam-repository/issues or to the package maintainer
>   unless you are confident it is an issue in the opam tool itself.

I just noticed that when I cancel opam via Ctrl-C during the build of something that needs a long time to build, opam quits immediately -- but the build actually continues in the background, still taking up CPU and memory.

I would expect Ctrl-C to also cancel the running build.

```
# opam config report
# opam-version      2.0.7 
# self-upgrade      no
# system            arch=x86_64 os=linux os-distribution=debian os-version=testing
# solver            builtin-mccs+glpk
# install-criteria  -removed,-count[version-lag,request],-count[version-lag,changed],-changed
# upgrade-criteria  -removed,-count[version-lag,solution],-new
# jobs              8
# repositories      4 (http), 3 (local), 2 (version-controlled) (default repo at 6bbebb34)
# pinned            1 (git)
# current-switch    /home/r/Dokumente/Unisachen/perennial/coq
```

rjbou:
Opam relay signal, but if the process don't catch them, opam can't really do something... See discussion in #3373

RalfJung:
After the Ctrl-C, I saw a `make` process somewhere, sending that a SIGTERM worked fine. `make` reacts to Ctrl-C so I assume SIGINT would also have worked.

So, the signal is getting lost somewhere before reaching the actual build process. Maybe it is swallowed by bubblewrap?

AltGr:
That should absolutely not be the case in general, so I would be extremely interested if you could provide a reproducible case, or at least more details on your build and how it happened; what often happens is that a process (typically, it's difficult to handle C-c properly in OCaml) eats the signal. But yes, probably not in the case of `make`...

In fact, in any case, opam is expected to [wait](https://github.com/ocaml/opam/blob/master/src/core/opamParallel.ml#L217-L218) for the child processes to exit.

So indeed bubblewrap may be the one causing trouble here, they have e.g. [this open issue](https://github.com/containers/bubblewrap/issues/369)

RalfJung:
> That should absolutely not be the case in general, so I would be extremely interested if you could provide a reproducible case,

```
opam pin add -nq coq-perennial.dev git+https://github.com/mit-pdos/perennial#coq/tested
opam install coq-perennial
# wait a minute
# Ctrl-C
[...]

PR #4530, opened 2021-02-07
Kill builds on Ctrl-C with bubblewrap

kit-ty-kate:
Per comment in https://github.com/ocaml/opam/issues/4400#issuecomment-774725510, I don't think we need to worry about supporting older version of bubblewrap.

- Only users on Debian 9 (EOL since July 2020) will have to upgrade bubblewrap
- Also, turns out opam already required bubblewrap >= 0.1.7

Fixes https://github.com/ocaml/opam/issues/4400

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -51,7 +51,8 @@ New option/command/subcommand are prefixed with ◈.
 ## External dependencies
 
 ## Sandbox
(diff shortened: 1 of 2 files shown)


avsm (approved):
Bumped version requirement looks good to me.

rjbou:
Thanks!

Issue #2976, opened 2017-06-23
support for semver version constraint operator


whitequark:
Many people prefer [semantic versioning](http://semver.org/) (although perhaps fewer in the opam ecosystem than elsewhere). Even when not followed entirely faithfully, semver-style version constraints greatly reduce the number of completely unconstrained dependency edges (which I believe currently are the majority in the repository) and thus downstream breakage and manual work for opam admin staff after a new major release of a popular package is released.

I propose opam to support the semver version constraint operator, `~`. The semantics are as follows:
  * `~ x.y.z` where `x > 0` expands to `>= x.y.z & < x.y+1.0`
  * `~ x.y` where `x > 0` expands to `>= x.y & < x+1.0`
  * `~ x` where `x > 0` expands to `>= x & < x+1`
  * `~ 0.y.z` expands to `>= 0.y.z & < 0.y+1.0`
  * `~ 0.y` expands to `>= 0.y & < 0.y+1`
  * everything else is an error.

You may notice that the behavior with major version 0 is not quite what the semver specification would naively imply (it says that at `0.y.z` anything may change at any time). This deviation is intentional and legal (semver does not define any such constraint operator), and provides a meaningful way for a package still in major version 0 to release bugfixes.

You may also notice that there is no mention of alphanumeric versions (alpha, beta, etc). As far as I know these basically do not get used in opam repository so it seems pointless to drag in their (considerably complex and, I think, inexpressible through simple expansion) semantics.

whitequark:
Duplicate of #2102 (although I suggest different / possibly cleaner semantics), feel free to close.

AltGr:
Sounds good, and thanks for suggesting this.
I would really like it if the new operator was defined on any allowed version, rather than throwing errors: this needs some generalisations of the above definition, but maybe we can manage it.

For the first case, for example, in opam terms it would be more consistent to have

```
{ ~ x.y.z }  ⇒  { >= x.y.z & < x.y+1 }
```

I think this is better because of the tilde semantics in our [total ordering](http://opam.ocaml.org/doc/2.0/Manual.html#version-ordering): `1.2.0~beta` would match `~ 1.1.3` (⇒ `>= 1.1.3 & < 1.2.0`) otherwise. This also generalises the `0.x.y` cases.

I don't know though how to handle the 2nd/6th cases:

```
{ ~ x.y } ⇒ { >= x.y & < x+1.0 } if x > 0
[...]

PR #4532, opened 2021-02-08
New operator '~' for semantic versioning

Keryan-dev:
This follows #2976 as an attempt to implement the '~' operator. It requires another patch of opam-file-format, coming along.

Given the difference between this operator and the others, I tried to avoid it to go too deep in the opam internals. Thus, most of the changes are contained in [opamFormat.ml](https://github.com/ocaml/opam/blob/master/src/format/opamFormat.ml) where occurrences of the new operator are caught and rewrote from `~ v` to ` >= v & < next(v)` with the behavior of `next` defined [in the issue](https://github.com/ocaml/opam/issues/2976#issuecomment-311049761).

I open the PR as a draft, as I feel the changes might need some more work to be satisfactory. Comments and reviews are most welcome.

### DIFF
--- a/src/format/opamFormat.ml
+++ b/src/format/opamFormat.ml
@@ -306,7 +306,13 @@ module V = struct
         | Int i -> FString (string_of_int i)
         | Ident _ -> FIdent (parse ~pos:v.pos filter_ident v)
         | Group g -> parse_filter ~pos:v.pos g.pelem
-        | Relop (op,e,f) -> FOp (aux e, op.pelem, aux f)
+        | Relop (op,e,f) ->
+           (match op.pelem with
+            | `Sem -> (* rewrite `Sem case *)
+               FAnd (FOp (aux e, `Geq, aux f),
+                     FOp (aux e, `Lt, FNext (aux f)))
+            | `Eq | `Neq | `Gt | `Geq | `Lt | `Leq as op ->
+               FOp (aux e, op, aux f))
         | Pfxop ({ pelem = `Not; _}, e) -> FNot (aux e)
         | Pfxop ({ pelem = `Defined; _}, e) -> FDefined (aux e)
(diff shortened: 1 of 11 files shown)


dra27:
Notes from dev meeting:
- The fact that older opam would silently interpret this as `>=` means we don't feel this can be reliably used in `opam` files
- However, it would be _very_ useful to be able to use it with `--formula`
- Conclusion is to look at having this properly (in opam files as well) in opam 3.0

src/format/opamFormat.ml
```diff
@@ -306,7 +306,13 @@ module V = struct
...
         | Ident _ -> FIdent (parse ~pos:v.pos filter_ident v)
         | Group g -> parse_filter ~pos:v.pos g.pelem
-        | Relop (op,e,f) -> FOp (aux e, op.pelem, aux f)
+        | Relop (op,e,f) ->
+           (match op.pelem with
+            | `Sem -> (* rewrite `Sem case *)
+               FAnd (FOp (aux e, `Geq, aux f),
+                     FOp (aux e, `Lt, FNext (aux f)))
+            | `Eq | `Neq | `Gt | `Geq | `Lt | `Leq as op ->
+               FOp (aux e, op, aux f))
```
AltGr, line 315: Easier to read with a flattened match ?
```suggestion
        | Relop ({pelem = `Sem; _}, e, f) -> (* rewrite `Sem case *)
           FAnd (FOp (aux e, `Geq, aux f),
                 FOp (aux e, `Lt, FNext (aux f)))
        | Relop ({pelem = `Eq | `Neq | `Gt | `Geq | `Lt | `Leq as op; _}, e, f) ->
           FOp (aux e, op, aux f))
```

AltGr (approved):
This deserves some tests in tests/reftests, but otherwise LGTM!

PR #4534, opened 2021-02-08
Fix Cudf preprocessing

AltGr:
This bug didn't impact correctness, but it did make it less efficient
(in particular, it could make higher versions of OPAMDIGDEPTH perform less 
trimming !)


rjbou:
Thanks!

PR #4535, opened 2021-02-08
Fix the conflict with the environment variable name used by dune

smorimoto:
As the title says. This is a bug that was found while investigating https://github.com/ocaml/dune/issues/4166.

Related issue: https://github.com/ocaml/opam/issues/4068

### DIFF
--- a/src/state/shellscripts/sandbox_exec.sh
+++ b/src/state/shellscripts/sandbox_exec.sh
@@ -49,9 +49,9 @@ add_ccache_mount() {
 }
 
 add_dune_cache_mount() {
-  DUNE_CACHE=${XDG_CACHE_HOME:-$HOME/.cache}/dune
-  mkdir -p "${DUNE_CACHE}"
-  add_mounts rw "$DUNE_CACHE"
+  dune_cache=${XDG_CACHE_HOME:-$HOME/.cache}/dune
+  mkdir -p "${dune_cache}"
+  add_mounts rw "$dune_cache"
  }
 
 # When using opam variable that must be defined at action time, add them also
--- a/src/state/shellscripts/sandbox_exec.sh
+++ b/src/state/shellscripts/sandbox_exec.sh
@@ -49,10 +49,10 @@ add_ccache_mount() {
 }
 
 add_dune_cache_mount() {
-  dune_cache=${XDG_CACHE_HOME:-$HOME/.cache}/dune
+  local dune_cache=${XDG_CACHE_HOME:-$HOME/.cache}/dune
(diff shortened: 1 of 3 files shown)


src/state/shellscripts/sandbox_exec.sh
```diff
@@ -49,9 +49,9 @@ add_ccache_mount() {
 }
 
 add_dune_cache_mount() {
-  DUNE_CACHE=${XDG_CACHE_HOME:-$HOME/.cache}/dune
-  mkdir -p "${DUNE_CACHE}"
-  add_mounts rw "$DUNE_CACHE"
+  dune_cache=${XDG_CACHE_HOME:-$HOME/.cache}/dune
```
kit-ty-kate, line 52:
```suggestion
  local dune_cache=${XDG_CACHE_HOME:-$HOME/.cache}/dune
```

smorimoto: Do we need to do this for non-macOS system script? (e.g. `src/state/shellscripts/bwrap.sh`)

rjbou: It's cleaner yes

smorimoto: Done.

rjbou (approved):

rjbou:
Thanks!

rjbou:
PR to backport in 2.0.9

smorimoto:
Thank you, too! Is there a timeline for the 2.0.9 release btw?

rjbou:
The list of backported PRs/fixes is done on the fly, regarding the needs. (and no date set for 2.0.9 yet).

smorimoto:
Got it, 2.1.0 doesn't seem to be released as early as I thought, so it would be helpful if 2.0.9 be released as soon as possible. (Because once put this change into the stable release, setup-ocaml can use the dune cache to speed up the entire workflow a couple of times.)

rjbou:
Ok, noted!

smorimoto:
But definitely there is no rush at all. Please release it at your convenience. Thanks for your effort on opam :)

rjbou:
Until the release, you can change the sandbox file. It is in `~/.opam/opam-init/hooks/sandbox.sh` (or sed it if in stateless scripty).

smorimoto:
Indeed, I don't really like doing that kind of hack on the action side, but let me create another branch and check it if it works as expected.

smorimoto:
OK, I replaced the file and it works as expected!

PR #4538, opened 2021-02-10
Backport CUDF trimming to the 2.0 branch

kit-ty-kate:
WIP


dra27:
Had we reached a conclusion that this won't work and that rebuilding opam 2.0 with GLPK 5 is the best we can do for now?

PR #4545, opened 2021-02-12
Fix build from source when a dune-project file is presented in the parent directory

### LINKED ISSUE #4537: Cannot build opam from sources, when 'dune-project' file is presented in the parent directory
Hi, I've been trying to build `opam` from sources and stumbled upon this problem.

To reproduce:
```
git clone https://github.com/ocaml/opam.git
touch dune-project
cd opam && ./configure && make opam
```

The following error occurs:
```
src_ext/dune-local/dune.exe build --profile=release   --promote-install-files opam-installer.install opam.install
Entering directory '/experiments'
File "dune-project", line 1, characters 0-0:
Error: Invalid first line, expected: (lang <lang> <version>)
Makefile:128: recipe for target 'build-opam' failed
make: *** [build-opam] Error 1
```

Environment: Ubuntu 18.04 (actually docker container), OCaml 4.0.5

kit-ty-kate:
Fixes #4537

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,11 +2,13 @@ ifeq ($(findstring clean,$(MAKECMDGOALS)),)
 -include Makefile.config
 endif
 
+# TODO: Replace --profile=$(DUNE_PROFILE) by --release when we require on dune >= 2.5
+
 all: opam opam-installer
 	@
 
 admin:
-	$(DUNE) build --profile=$(DUNE_PROFILE) $(DUNE_ARGS) opam-admin.install
+	$(DUNE) build --profile=$(DUNE_PROFILE) --root . $(DUNE_ARGS) opam-admin.install
 
 DUNE_PROMOTE_ARG =
 DUNE_PROMOTE_ARG += --promote-install-files
@@ -21,7 +23,7 @@ ifeq ($(DUNE),)
 else
   DUNE_EXE=
   # NB make does not export the PATH update in Makefile.config to $(shell ...)
-  ifeq ($(shell PATH='$(PATH)' $(DUNE) build --help=plain 2>/dev/null \
+  ifeq ($(shell PATH='$(PATH)' $(DUNE) build --root . --help=plain 2>/dev/null \
                   | grep -F -- '$(DUNE_PROMOTE_ARG) '),)
     DUNE_PROMOTE_ARG =
   endif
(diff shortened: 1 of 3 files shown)


rjbou (approved):
Thanks!

rjbou:
Just a question, is there a reason to not use `DUNE_ARGS` to add it one place instead on each `dune` call ?

dra27:
> Just a question, is there a reason to not use `DUNE_ARGS` to add it one place instead on each `dune` call ?

🌴 You don't want the user to be able to change it - `DUNE_ARGS` is for extra arguments, not mandatory ones. It could go in another variable like `DUNE_CMD = $(DUNE) --profile=$(...) --root=.` but we already didn't do that for `--profile`

Makefile
```diff
@@ -2,11 +2,13 @@ ifeq ($(findstring clean,$(MAKECMDGOALS)),)
 -include Makefile.config
 endif
 
+# TODO: Replace --profile=$(DUNE_PROFILE) by --release when we require on dune >= 2.5
+
```
dra27, line 6: This would need to be slightly more complicated - the idea was that you can say `make DUNE_PROFILE=dev`

doc/Makefile
```diff
@@ -14,21 +14,8 @@ endif
...
 
-ifndef OPAM
-  OPAM = $(DUNE) exec --no-print-directory --profile=$(DUNE_PROFILE) -- opam
-  OPAM_DEP := $(shell $(DUNE) exec -- which opam)
-endif
-
-TOPICS  = $(shell $(OPAM) help topics)
-TOPICS_ADMIN = cache filter index lint list upgrade
-
 HELPFMT = --help=groff
```
dra27, line 17: This can go too

Makefile
kit-ty-kate: we could still have something like that:
```
ifeq ($(DUNE_PROFILE),release)
DUNE_PROFILE_ARG=--release
else
DUNE_PROFILE_ARG=--profile=$(DUNE_PROFILE)
endif

doc/Makefile
kit-ty-kate: done

Makefile
dra27: Indeed, that was my slightly more complicated 🙂

dra27:
Good to go when CI’s ready

rjbou:
@AltGr can you take a look (and update PR) to be sure to not broke the new dune tests makefiles :innocent:

dra27 (approved):
Rebased and I (throughly) checked all the dune calls - good to go, as long as CI agrees.

PR #4548, opened 2021-02-12
Detect available packages from homebrew

kit-ty-kate:
Even if it makes less sense than on Debian or other distributions, having opam detect which packages are available in homebrew might be still useful:
- users might have local brew remotes
- this helps the solver decide to avoid certain packages (e.g. camlp4 which isn't available in homebrew anymore)
- this leaves the door open to easily tell opam that some packages have been installed manually (using `opam option depext-bypass` / `--assume-depexts` / `--no-depexts`)

cc @patricoferris who got the issue with `camlp4` on the upcoming macOS CI

### DIFF
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -618,6 +618,11 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
     in
     compute_sets sys_installed
   | Homebrew ->
+    let sys_available =
+      run_query_command "brew" ["formulae"]
+      |> List.map OpamSysPkg.of_string
+      |> OpamSysPkg.Set.of_list
+    in
     (* accept 'pkgname' and 'pkgname@version'
        exampe output
        >openssl@1.1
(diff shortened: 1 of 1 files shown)


rjbou (approved):
Thanks!

patricoferris:
Thanks @kit-ty-kate! Unfortunately it seems that the command `brew formulae` doesn't also list the set of aliases (or there's no way to make it include those). It uses this [items script](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/items.sh) with [these parameters](https://github.com/Homebrew/brew/blob/master/Library/Homebrew/cmd/formulae.sh). This is unfortunate because some alias names are used in opam files for example: 
  - [Sqlite3 in opam](https://github.com/ocaml/opam-repository/blob/master/packages/conf-sqlite3/conf-sqlite3.1/opam#L26) is an [alias in Homebrew](https://github.com/Homebrew/homebrew-core/blob/master/Aliases/sqlite3). 
  
I was testing this in macOS CI and came across it whilst installing [current_ocluster](https://opam.ocaml.org/packages/current_ocluster/): 

```
[ERROR] Package conflict!
  * Missing dependency:
    - current_ocluster → current >= 0.3 → sqlite3 → conf-sqlite3
```

But a brew search reveals: 

```sh
$ brew search sqlite3
==> Formulae
sqlite3 ✔
```

The problem with `brew search` is that it takes quite a while (I think it hits the Homebrew API for searching). The maintainers of [Homebrew discussed aliases a long time ago](https://github.com/Homebrew/brew/issues/30). I can't seem to find any PRs that added any kind of support for this.

kit-ty-kate:
https://github.com/ocaml/opam/pull/4801 also shows us that this PR in its current state does not handle casks

PR #4556, opened 2021-02-22
archlinux: Always upgrade all the installed packages when installing a new package

kit-ty-kate:
e.g. when libc is upgraded and `opam update --depexts && opam install conf-gtksourceview3` is called, pacman will not upgrade the underlying libc and will result in a link failure similar to `undefined reference to 'stat64@GLIBC_2.33'`

Backported to `opam-depext` in https://github.com/ocaml-opam/opam-depext/pull/138

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -50,6 +50,7 @@ New option/command/subcommand are prefixed with ◈.
 
 ## External dependencies
   * Handle macport variants [#4509 @rjbou - fix #4297]
+  * Always upgrade all the installed packages when installing a new package on Archlinux [#4556 @kit-ty-kate]
 
 ## Sandbox
   * Fix the conflict with the environment variable name used by dune [#4535 @smorimoto - fix ocaml/dune#4166]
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
Example: https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/22202f5f9e690ac85c4fa313e36d27c8485d74d7/variant/distributions,archlinux,coqide.8.13.1

rjbou:
Do you know why the update operation doesn't update correctly ?

kit-ty-kate:
Which update operation?

rjbou:
The one called by `opam update --depexts` : `pacman -Sy`

kit-ty-kate:
It does update correctly. `pacman -Sy` is the same as `apt update`, i.e. only update the internal database of packages.
The issue is that on Debian every packages recusively depending on libc will be updated to include a constraint `libc>=2.33`, but archlinux maintainers do not do that. They expect users to always have an up-to-date system when a new package is installed (given one version of the package database, ofc)

rjbou (approved):

PR #4563, opened 2021-02-25
Add `--unsafe-depext-yes' 

### LINKED ISSUE #4321: Command-line option needed for OPAMDEPEXTYES
Picking up @kit-ty-kate's comment in https://github.com/ocaml/opam/pull/4168#issuecomment-627326960 while sorting out CLI versioning.

We shouldn't have settings which can only be triggered by environment variables. I'm not convinced that `OPAMDEPEXTYES=yes OPAMYES=yes` is something a user would ever _need_ to do (i.e. opam is interactive but the package manager is not)?

A simple enough alternative - used by some of those package managers already - would be to allow `-y`/`--yes` to be repeatable so `opam -yy` means yes to everything (package manager included). That seems superior to `-Y` for the same reason as not having `OPAMDEPEXTYES` and `OPAMYES`. The repetition can be encoded in the environment by adding `always` and `never` to the allowed values for `OPAMYES`?

rjbou:
For system package managers & change `OPAMDEPEXTYES' to `OPAMUNSAFEDEPEXTYES'
fix #4321

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -7,7 +7,7 @@ New option/command/subcommand are prefixed with ◈.
   *
 
 ## Global CLI
-  *
+  * ◈ Add `--unsafe-depext-yes' for system package managers and change `OPAMDEPEXTYES' to `OPAMUNSAFEDEPEXTYES' [#4563 @rjbou - fix #4321]
 
 ## Init
   *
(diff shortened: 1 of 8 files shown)


avsm (approved):
Looks good to me. Once this is merged, we'll need to update ocaml-ci/opam-repo-ci to use the new env var.  This could also use a note in the CHANGES file that the variable that used to work in the previous betas does not work now.

rjbou:
In the release notes / blog post too. Also we shouldn't forget to specify that it will be removed in next release.

rjbou:
Superseed by #4582

PR #4564, opened 2021-02-25
With `--deps-only`, set direct dependencies as root by default

### LINKED ISSUE #4502: `opam install --deps` does not mark as root
I am not certain it should, but it may be a bit confusing that `opam install . --deps` followed by `opam remove -a` will remove what was just installed. `--set-root` does not work either, which it definitely should.

rjbou:
fixes #4502

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -1214,22 +1214,39 @@ let install_t t ?ask ?(ignore_conflicts=false) ?(depext_only=false)
         (OpamCudf.string_of_conflicts t.packages
            (OpamSwitchState.unavailable_reason t) cs);
       t, if depext_only then None else Some (Conflicts cs)
-    | Success solution ->
+    | Success full_solution ->
       let solution =
         if deps_only then
           OpamSolver.filter_solution (fun nv ->
               not (OpamPackage.Name.Set.mem nv.name names))
-            solution
-        else solution in
+            full_solution
+        else full_solution in
       if depext_only then
         (OpamSolution.install_depexts ~force_depext:true ~confirm:false t
            (OpamSolver.all_packages solution)), None
       else
       let add_roots =
-        OpamStd.Option.map (function
-            | true -> names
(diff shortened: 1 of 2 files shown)


src/client/opamClient.ml
```diff
@@ -1214,22 +1214,53 @@ let install_t t ?ask ?(ignore_conflicts=false) ?(depext_only=false)
...
+            let open OpamPackage.Set.Op in
+            (installed -- to_change) ++ t.installed_roots
+          in
+          let t = { t with installed_roots } in
+          t, Some depends_to_install
+        else
+          t, OpamStd.Option.map (function
+              | true -> names
+              | false -> OpamPackage.Name.Set.empty)
+            add_to_roots
```
AltGr, line 1263: Shorter proposition:

```suggestion      let add_roots =
        if deps_only && add_to_roots <> Some false then
          let requested_deps =
            OpamPackage.Set.fold (fun nv acc ->
                OpamFormula.ors [
                  OpamPackageVar.all_depends t (OpamSwitchState.opam t nv)
                    ~depopts:false ~build:true ~post:false;
                  acc
                ])
              (OpamPackage.packages_of_names
                 (OpamSolver.all_packages full_solution)
                 names)
              OpamFormula.Empty
          in
          Some (OpamPackage.names_of_packages
                  (OpamFormula.packages
                     (OpamSolver.all_packages solution) requested_deps))
        else
          OpamStd.Option.map (function
              | true -> names
              | false -> OpamPackage.Name.Set.empty)
            add_to_roots
```

This also respects `--unset-root`, and evaluates the formula instead of extracting raw package names without filtering.
I don't think updating the state is needed, since the `apply` function already takes care of that.

AltGr: (note that `--unset-root --deps-only` still won't unroot already installed packages; I am not sure it is worth considering)

rjbou: Indeed...

rjbou: Integrated, thanks!

Issue #4565, opened 2021-02-25
Colorize URLs in opam info output


aantron:
For example, the following is only part of the output of one `info` command &mdash; and it fills most of my terminal window. It would save a bit of frustration if some of the human-oriented links (homepage, bug-reports) were colored, say, blue, as they can be difficult to pick out when working quickly. Some of the other output is already colored, so this shouldn't be much trouble.

I also noticed that there is no output corresponding to the `doc` field, which seems pretty important. I suppose I should open a separate issue about that.

```
$ opam info bisect_ppx
[SNIP]
<><> Version-specific details <><><><><><><><><><><><><><><><><><><><><><><><><>
version       2.6.0
repository    default
url.src:      "https://github.com/aantron/bisect_ppx/archive/2.6.0.tar.gz"
url.checksum: "md5=1e87162003525cd853631dae5b4aea66"
homepage:     "https://github.com/aantron/bisect_ppx"
bug-reports:  "https://github.com/aantron/bisect_ppx/issues"
dev-repo:     "git+https://github.com/aantron/bisect_ppx.git"
authors:      "Xavier Clerc <bisect@x9c.fr>"
              "Leonid Rozenberg <leonidr@gmail.com>"
              "Anton Bachin <antonbachin@yahoo.com>"
maintainer:   "Anton Bachin <antonbachin@yahoo.com>" "Leonid Rozenberg
              <leonidr@gmail.com>"
license:      "MIT"
depends:      "base-unix"
              "cmdliner" {>= "1.0.0"}
              "dune" {>= "2.7.0"}
              "ocaml" {>= "4.02.0"}
              "ppxlib" {>= "0.21.0"}
              "ocamlformat" {with-test & = "0.16.0"}
synopsis      Code coverage for OCaml
description
          Bisect_ppx helps you test thoroughly. It is a small preprocessor
          that inserts instrumentation at places in your code, such as
          if-then-else and
          match expressions. After you run tests, Bisect_ppx gives a nice HTML
          report
          showing which places were visited and which were missed.
          Usage is simple - add package bisect_ppx when building tests, run your
          tests,
          then run the Bisect_ppx report tool on the generated visitation files.
```

dra27:
The idea is still potentially useful, though two points as it's causing pain right now:

- We added opam support to Google - searching for `opam bisect_ppx` gives a page with coloured (and clickable!) links
- Including the `doc` field by default would be a good idea, you can also get it with `opam show bisect_ppx -f doc`

PR #4568, opened 2021-02-26
Clickable links in opam show

dra27:
A piece of sleeplessness-induced silliness looking at #4565 - if a field in opam show parses as a URL, turn the field marker into a clickable hyperlink.

Not for 2.1 (if at all, but the function might be useful, even if not used in opam show)

### DIFF
--- a/src/client/opamListCommand.ml
+++ b/src/client/opamListCommand.ml
@@ -760,7 +760,18 @@ let info st ~fields ~raw ~where ?normalise ?(show_empty=false)
       List.fold_left (fun acc item ->
           let contents = detail_printer ?normalise ~sort st nv item in
           if show_empty || contents <> "" then
-            [ OpamConsole.colorise `blue (string_of_field ~raw item); contents ]
+            let name = string_of_field ~raw item in
+            let name_url =
+              try
+                let basic_contents = detail_printer ~prettify:true ?normalise ~sort st nv item in
+                let contents_url = OpamUrl.parse basic_contents in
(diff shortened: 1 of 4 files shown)


src/client/opamListCommand.ml
```diff
@@ -760,7 +760,18 @@ let info st ~fields ~raw ~where ?normalise ?(show_empty=false)
       List.fold_left (fun acc item ->
           let contents = detail_printer ?normalise ~sort st nv item in
           if show_empty || contents <> "" then
-            [ OpamConsole.colorise `blue (string_of_field ~raw item); contents ]
+            let name = string_of_field ~raw item in
+            let name_url =
+              try
+                let basic_contents = detail_printer ~prettify:true ?normalise ~sort st nv item in
+                let contents_url = OpamUrl.parse basic_contents in
+                if OpamUrl.(contents_url.transport = "http" || contents_url.transport = "https") then
```
rjbou, line 768: A reason to not use the `backend` field ?

src/core/opamConsole.ml
```diff
@@ -442,6 +442,12 @@ let carriage_delete =
   else
     carriage_delete_unix
 
+let url ~ref ~label =
+  if not (color ()) || Sys.win32 && get_win32_console_shim `stdout Mode = Shim then
```
avsm, line 446: Might be worth a comment linking to the handy https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda which describes the escape codes for Linux

PR #4570, opened 2021-02-26
config report: fix not found and compilers variables display

rjbou:
Instead of  `compiler_packages` or invariant (which is not necessarily a compiler), ~it looks for installed packages with compiler flag and display their variables.~ it displays variables of installed compilers (packages with compiler flag) and their installed dependencies.

cc/ @dra27

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -12,8 +12,9 @@ New option/command/subcommand are prefixed with ◈.
 ## Init
   *
 
-## Config Upgrade
-  *
+## Config report
+  * Fix `Not_found` (config file) error [#4570 @rjbou]
+  * Print variables of installed compilers and their (installed) dependencies [#4570 @rjbou]
 
 ## Install
   * Don't patch twice [#4529 @rjbou]
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -1319,19 +1319,21 @@ let config cli =
           print "current-switch" "%s"
             (OpamSwitch.to_string state.switch);
           let process nv =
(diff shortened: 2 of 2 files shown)


dra27:
It appears to be doing what I want, but not according to your description...
```
dra@thor:~/opam$ ./opam config report
# opam config report
# opam-version         2.1.0~beta4 (fc8680c46a88613ae4bcbb3af8255baaf07f24d6)
# self-upgrade         no
# system               arch=x86_64 os=linux os-distribution=ubuntu os-version=20.10
# jobs                 7
# repositories         1 (http), 4 (version-controlled) (default repo at d8869936)
# pinned               0
# current-switch       test-pinning-4.12
# ocaml:native         true
# ocaml:native-tools   true
# ocaml:native-dynlink true
# ocaml:stubsdir       /home/dra/.opam/test-pinning-4.12/lib/ocaml/stublibs:/home/dra/.opam/test-pinning-4.12/lib/ocaml
# ocaml:preinstalled   false
# ocaml:compiler       4.12.0
dra@thor:~/opam$ ./opam show ocaml --field=flags

```
The `ocaml` doesn't have the `compiler` flag set, but its variables are being displayed?

dra27:
I'm wondering, though, if it would be better to display _both_ the variables of any packages in the switch invariant _and_ any flagged `compiler`

rjbou:
because I forgot an *and their installed dependencies*

dra27:
So it does! In which case, it all LGTM 🙂

rjbou:
informations updated!

Issue #4503, opened 2021-01-20
On breaking compat for the CLI for 2.1


Drup:
Hi,

I recently migrated to the last version of `opam-git`, and noticed that opam now breaks compat with the CLI. In particular `opam config var` is not available anymore, and `opam var` must be used instead.

This is just a head's up that it break at the very least the following tools (and probably more):
- merlin
- ocp-browser
- tuareg
- dune

thanksfully this can be rolledback using the `OPAMCLI` variable, but I'm not sure the change is worth it.

------------

```
# opam config report
# opam-version         2.1.0~beta4 (8ce5e420f8007fd8d7ad56fb6293af6f5469a432) 
# self-upgrade         no
# system               arch=x86_64 os=linux os-distribution=arch os-version=rolling
# solver               builtin-mccs+glpk
# install-criteria     -removed,-count[hidden-version,changed],-count[version-lag,request],-count[version-lag,changed],-count[missing-depexts,changed],-changed
# upgrade-criteria     -removed,-count[hidden-version,changed],-count[version-lag,solution],-count[missing-depexts,changed],-new
# jobs                 3
# repositories         2 (http), 1 (local), 4 (version-controlled) (default repo at 1aa37053)
# pinned               4 (git)
# current-switch       main
# read-state           Not_found

```

hannesm:
Thanks for opening this issue, indeed in `opam-repository` there are several occasions of `opam config var` (and these are as well present in some "cross-compilation" settings for MirageOS: mirage-solo5, mirage-crypto, bigstringaf, ... are invoking `opam config var prefix` as well). I'd as well be in favour to keep the `opam config var` (maybe it could emit a deprecation warning on stderr?)

dra27:
@hannesm - opam-repository should not be affected, since opam itself sets `OPAMCLI=2.0` for package builds (see #4992). Scripts which intend to use 2.1 features/CLI syntax should specify `--cli=2.1`

hannesm:
@dra27 I'm not sure what the story is. Do you mean that with opam 2.1, the following will be run with `OPAMCLI=2.0`: `opam install mirage-crypto`, while `opam pin add mirage-crypto --dev` will use the 2.1 cli version? OR is it the case that opam will always behave as 2.0, but only if `--cli=2.1` is passed it'll unlock the new behaviour?

Makes me slightly puzzled and afraid, versioning in command line behaviour...

dra27:
[...]

PR #4575, opened 2021-03-02
Introducing default cli

rjbou:
Some deprecated options need to still be accepted (cf. #4503). The default cli is here to allow them when used with no specified cli (not from environment or from option `--cli`). Opam prints then a removal warning notice.
The default version is set to 2.0, it will be upgraded at each major release.
Default cli options are ftm `opam config` deprecated commands.

Example with `config var`:
```
$ opam config var prefix
[WARNING] var is deprecated since version 2.1 of the opam CLI, but version 2.1 has been requested. Use opam var instead or set OPAMCLI environment variable to 2.0.
.opam/switch

$ opam config var prefix --cli 2.0
.opam/switch

$ opam config var prefix --cli 2.1
opam: var was removed in version 2.1 of the opam CLI, but version 2.1 has been requested. Use opam var instead or set OPAMCLI environment variable to 2.0.

$ OPAMCLI="2.0" opam config var prefix
.opam/switch

$ OPAMCLI="2.1" opam config var prefix
opam: var was removed in version 2.1 of the opam CLI, but version 2.1 has been requested. Use opam var instead or set OPAMCLI environment variable to 2.0.
```

### DIFF
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -847,11 +847,11 @@ module Mk : sig
   type command = unit Term.t * Term.info
 
   val mk_command:
-    OpamCLIVersion.t -> validity -> string -> doc:string ->
+    cli:OpamCLIVersion.t -> validity -> string -> doc:string ->
     man:Manpage.block list -> (unit -> unit) Term.t -> command
 
   val mk_command_ret:
-    OpamCLIVersion.t -> validity -> string -> doc:string ->
+    cli:OpamCLIVersion.t -> validity -> string -> doc:string ->
     man:Manpage.block list -> (unit -> unit Term.ret) Term.t -> command
 
 end = struct
@@ -1217,7 +1217,7 @@ end = struct
 
   type command = unit Term.t * Term.info
 
-  let mk_command cli validity name ~doc ~man cmd =
+  let mk_command ~cli validity name ~doc ~man cmd =
     let doc = update_doc_w_cli doc ~cli validity in
     let info = term_info name ~doc ~man in
     let check =
@@ -1227,7 +1227,7 @@ end = struct
     in
     Term.(cmd $ check), info
 
-  let mk_command_ret cli validity name ~doc ~man cmd =
+  let mk_command_ret ~cli validity name ~doc ~man cmd =
     let doc = update_doc_w_cli doc ~cli validity in
     let info = term_info name ~doc ~man in
     let check =
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -822,9 +822,9 @@ module Mk : sig
     cli:OpamCLIVersion.Sourced.t -> ?section:string -> ?default:'a list ->
     (validity * 'a * string list * string) list -> 'a list Term.t
 
-  val mk_tristate_opt:
+  val mk_state_opt:
     cli:OpamCLIVersion.Sourced.t -> validity -> ?section:string -> string list ->
-    string -> string -> [> `Always | `Auto | `Never ] option Term.t
+    string -> (string * 'a) list -> string -> 'a option Term.t
 
   type 'a subcommand = validity * string * 'a * string list * string
   type 'a subcommands = 'a subcommand list
@@ -1130,11 +1130,11 @@ end = struct
     let default = List.map (fun x -> `Valid x) default in
     term_cli_check ~check Arg.(vflag_all default info_flags)
 
-  let mk_tristate_opt ~cli validity ?section flags value doc =
+  let mk_state_opt ~cli validity ?section flags value state doc =
     let doc = update_doc_w_cli doc ~cli validity in
     let doc = Arg.info ?docs:section ~docv:value ~doc flags in
     let check elem = check_cli_validity cli validity elem flags in
-    term_cli_check ~check Arg.(opt (some (enum when_enum)) None & doc)
+    term_cli_check ~check Arg.(opt (some (enum state)) None & doc)
 
   (* Subcommands *)
   type 'a subcommand = validity * string * 'a * string list * string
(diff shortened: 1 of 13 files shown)


src/client/opamArg.ml
```diff
@@ -922,38 +928,55 @@ end = struct
         "%s was added in version %s of the opam CLI, \
          but version %s has been requested, which is older."
         flag (OpamCLIVersion.to_string valid_since)
-        (OpamCLIVersion.to_string cli)
+        (string_of_sourced_cli cli)
     in
     `Error (false, msg)
 
+  let previsouly_str removal instead =
```
dra27, line 935:
```suggestion
  let previously_str removal instead =
```

src/client/opamArg.ml
```diff
@@ -922,38 +928,55 @@ end = struct
...
-        (let previous =
-           string_of_cli_option (OpamCLIVersion.previous removal)
-         in
-         match instead with
-         | Some ist ->
-           Printf.sprintf  ". Use %s instead or %s"
-             (OpamConsole.colorise `bold ist) previous
-         | None -> Printf.sprintf ", %s"  previous)
+        (string_of_sourced_cli cli)
+        (previsouly_str removal instead)
```
dra27, line 953:
```suggestion
        (previously_str removal instead)
```

src/client/opamArg.ml
```diff
@@ -922,38 +928,55 @@ end = struct
...
     `Error (false, msg)
 
+  let deprecated_warning cli removal instead flags =
+    let flag = get_long_form flags in
+    OpamConsole.warning
+      "%s is deprecated since version %s of the opam CLI, \
+       but version %s has been requested%s."
+      flag (OpamCLIVersion.to_string removal)
+      (string_of_sourced_cli cli)
+      (previsouly_str removal instead)
```
dra27, line 964: `snd cli` must be `` `Default``, so `fst cli` must be `OpamCLIVersion.current` so there's no need to display it in the message. In fact, the `cli` parameter isn't needed, beyond possibly asserting that ``snd cli = `Default``?

```suggestion
  let deprecated_warning cli removal instead flags =
    let flag = get_long_form flags in
    OpamConsole.warning
      "%s was deprecated in version %s of the opam CLI%s."
      flag (OpamCLIVersion.to_string removal)
      (previously_str removal instead)
```

tests/reftests/cli-versioning.test
```diff
@@ -10,7 +10,8 @@ The following actions would be performed:
 ### OPAMCLI=2.0 opam install --show --assume-depexts base-unix
 opam: assume-depexts was added in version 2.1 of the opam CLI, but version 2.0 has been requested, which is older.
 ### opam config set cli version
-opam: set was removed in version 2.1 of the opam CLI, but version 2.1 has been requested. Use opam var instead or set OPAMCLI environment variable to 2.0.
+[WARNING] set is deprecated since version 2.1 of the opam CLI, but version 2.1 has been requested. Use opam var instead or set OPAMCLI environment variable to 2.0.
```
dra27, line 13:
```suggestion
[WARNING] set was deprecated in version 2.1 of the opam CLI. Use opam var instead or set OPAMCLI environment variable to 2.0.
```

src/client/opamCLIVersion.ml
```diff
@@ -27,17 +27,16 @@ let of_string s =
   | _ -> failwith "OpamVersion.CLI.of_string"
 
 let current = of_string @@ OpamVersion.(to_string current_nopatch)
+let default = (2,0)
```
dra27, line 30:
```suggestion
let default = (fst current, 0)
```
?

dra27 (commented):
Lovely, thanks! The only changes are the misspelling of `previously_str` and the deprecation warning can be re-worded. Semantically it's doing exactly what's needed, and I very much like `OpamCLIVersion.default`. Perhaps it could be automatically the major version of opam?

src/client/opamCLIVersion.ml
rjbou: I tend to agree with this change. Having it directly related permit the automatic update of the default as major version (no more need to handles it). But this automation makes the change invisible, as opposed to a PR announcing the default cli change, in order to let package maintainers and users update their scripts.

dra27: From dev meeting: this check could be in the release scripts instead (i.e. release checks that the default cli matches major version)

release/Makefile
```diff
@@ -21,6 +21,8 @@ out/opam-full-$(VERSION).tar.gz:
 	mkdir -p out
 	cd out && curl -OfL $(FULL_ARCHIVE_URL) || { \
 	  git clone $(GIT_URL) -b $(TAG) --depth 1 opam-full-$(VERSION); \
+	  test "`sed -n "s/^let default.*(\([0-9]\),\([0-9]\))/\1.\2/p" src/client/opamCLIVersion.ml`" \
+	    = "`echo $VERSION | cut -f 1-2 -d '.'`" ; \
```
rjbou, line 25: ping @AltGr

dra27, line 25: Awesome! 😀 Slightly more hardened:
```suggestion
	  test "`sed -n "/let *default *=/s/.*(\([0-9]*\)[^0-9]*\([0-9]*\).*/\1.\2/p" src/client/opamCLIVersion.ml`" \
	    = "`echo $VERSION | cut -f 1-2 -d '.'`" ; \
```
It's worth a comment in `opamCLIVersion.ml` that the line is analysed (cf. [things like this](https://github.com/ocaml/ocaml/blob/trunk/tools/pre-commit-githook#L16-L18)).

Is this release script used to build the 2.0.x releases as well, or do we use the release script on the 2.0 branch?

rjbou: Good question... @AltGr ?

rjbou: Moved as a ci check, on master push

PR #4580, opened 2021-03-04
pin version: don't error if archive opam file is malformed

rjbou:
cc/ @kit-ty-kate


kit-ty-kate:
In my opinion the check itself shouldn't be there as the version field the archive is often either not present or outright false.

rjbou:
In that case, there is (ftm) no way to determine which source it is, and we will need to download it (extract archive) each time.

PR #4581, opened 2021-03-04
Deprecate using cli version already deprecated `build-doc', `build-test', and `make'

rjbou:


dra27:
This doesn't appear to be working correctly:

```
$ ./opam install bitmasks --build-doc
opam: build-doc was removed in version 2.1 of the opam CLI, but version 2.1 has been requested. Use --with-doc instead or set OPAMCLI environment variable to 2.0.
```

dra27:
Ouch, there's a more serious problem as well (not necessarily introduced with this PR, but in CLI versioning): `OPAMBUILDDOC=true ./opam install bitmasks` silently "works" to select the `with-doc` dependencies of `bitmasks`.

rjbou:
There is no handling of environment variables in cli versioning, only cli.

dra27:
This is an oversight/error for options which have been removed - it shouldn't be that `--build-doc` is banned but you can quietly persuade opam to do it with an "old" environment variable!

rjbou:
I think at that time, it was removed from doc, but kept handled to not break scripts. We can set up a new guidelines for environment variables deprecation.

rjbou:
Queued, to rebase on cli versioning environment variables edition

dra27 (approved):
LGTM!

PR #4582, opened 2021-03-04
Introduce Confirm level

rjbou:


dra27:
From dev meeting:
- Introduce `OPAMCONFIRMLEVEL` to take the value and have `OPAMYES` and `OPAMNO` remain booleans only
- Need to remove `OPAMDEPEXTYES` either here or subsequently

src/client/opamArg.ml
```diff
@@ -442,9 +454,16 @@ let create_global_options
   let debug_level = OpamStd.Option.Op.(
       debug_level >>+ fun () -> if debug then Some 1 else None
     ) in
+  let answer =
+    match yes, confirm_level with
+    | None, None -> None
+    | _ , Some c -> Some c
+    | Some true, None -> Some `all_yes
+    | Some false, None -> Some `all_no
+  in
```
rjbou, line 463: priorities over cli: confirm level > yes/no, no yes/no simultaneously

src/core/opamCoreConfig.ml
```diff
@@ -157,11 +159,19 @@ let initk k =
         | true -> Some `Extended
         | false -> None)
     ) in
+  let answer =
+    match E.confirmlevel (), E.yes (), E.no () with
+    | Some c, _,  _ -> Some c
+    | _, Some true, _ -> Some `all_yes
+    | _, _, Some true -> Some `all_no
+    | _ -> None
+(*
```
rjbou, line 168: priority over environment variables: confim level > yes > no

src/client/opamArg.ml
```diff
@@ -23,6 +23,14 @@ let when_enum =
   [ "always", `Always; "never", `Never; "auto", `Auto ]
   |> List.map (fun (s,v) -> cli_original, s, v)
 
+let confirm_enum =
+  [
+    cli_from cli2_1, "ask", `ask;
+    cli_from cli2_1, "yes", `all_yes;
+    cli_from cli2_1, "no", `all_no;
```
dra27, line 30:
```suggestion
    cli_from cli2_1, "no", `all_no;
    cli_from cli2_1, "yes", `all_yes;
```
Cosmetically, it means the orders are decreasing order of safety:
- Ask user (i.e. no automatic answer)
- No
- Yes
- Unsafe-yes

src/core/opamCoreConfig.ml
dra27: It's worth having that in the docstrings too, I think.

src/client/opamCliMain.ml
```diff
@@ -132,8 +132,12 @@ let check_and_run_external_commands () =
...
     (* No such command, check if there is a matching plugin *)
     let command = OpamPath.plugin_prefix ^ name in
-    let answer = if yes then Some true else OpamCoreConfig.E.yes () in
-    OpamCoreConfig.init ~answer ();
+    let answer =
+      if yes then Some `all_yes else
+        OpamStd.Option.Op.(OpamCoreConfig.E.yes ()
+                           >>| (fun b -> if b then `all_yes else `all_no))
+    in
+    OpamCoreConfig.init ?answer ();
```
dra27, line 140: Shouldn't this take `OPAMCONFIRMLEVEL` and `OPAMNO` into account as well?

dra27: Even if you can't say `opam --confirm-level=yes plugin blah`, I think it would be correct to expect `OPAMCONFIRMLEVEL=yes opam plugin blah` to install the plugin automatically and `OPAMNO=true opam plugin blah` not to?

src/client/opamArg.ml
dra27: Hmm - I think it's inconsistent that you can't give `--yes` and `--no` at the same time, yet you _can_ do `--yes --confirm-level=no`.

Overall, I think you should be able to give `--yes`, `--no` and `--confirm-level` as many times as you like - it's simply the rightmost one which wins. The rationale for that is that you can always add it in a script:

e.g.

```bash
opam_install="opam install --confirm-level=yes"
# .., many lines of bash...
# OK, risking it here...
$opam_install --confirm-level=unsafe-yes ...
```

However, it's late in the day to allow this - I might propose that for 2.2! For now, I think `--confirm-level` shouldn't be permitted with `--yes` or `--no`... what do you think?

dra27 (commented):
Priority of the environment variables looks good to me - a couple of questions about the handling of `OPAMNO`/`OPAMCONFIRMLEVEL` for plugins and a CLI question on conflicting options.

src/client/opamCliMain.ml
rjbou: It should yes. For that we need to have `OPAMCONFIRMLEVEL` and `OPAMNO` out of the environment variables cli versioning system, and initialised at [launch](https://github.com/ocaml/opam/blob/master/src/client/opamArg.ml#L42).

src/client/opamArg.ml
```diff
@@ -1119,9 +1138,22 @@ let global_options cli =
...
+      cli_from cli2_1, Some false, ["no"],
+      "Answer no to all opam yes/no questions without prompting. \
+       This is equivalent to setting $(b,\\$OPAMNO) to \"true\".";
+    ]
+  in
+  let confirm_level =
+    mk_enum_opt ~cli (cli_from cli2_1) ~section ["confirm-level"] "LEVEL"
+      confirm_enum
+      (Printf.sprintf "Confirmation level, $(docv) must be %s. \
+                       This is equivalent to setting $(b, \\$OPAMYES)`."
```
kit-ty-kate, line 1154: To me, this line means that `OPAMYES=unsafe` is a thing, but it's not

src/client/opamArg.ml
rjbou: Ah it is not the same kind of flag (simple flag & flag with argument), it is not possible to mix them and check their order of appearance (using cmdliner api). It is only possible between yes/no. 
Will update to
* yes/no take the last in command line*
* take the confirm-level in command line
* if both defined take the confirm level value

src/client/opamCliMain.ml
rjbou: In fact, it is possible here to initialise environment variables, as we have cli, and it's a piece of code that will just install or launch the plugin. I think that it is reasonable that plugins inherit of environment variable set?

src/client/opamArg.ml
dra27: Sounds good - that would also be allowing `--confirm-level` to appear multiple times (and take the last one?)

src/client/opamCliMain.ml
dra27: Agreed!

AltGr (approved):
LGTM
for `mk_enum_opt_all` I'll take your word on it :grimacing:

dra27 (commented):
Some message change suggestions - I think it's clearer to state exactly how the env vars are interpreted

src/client/opamArgTools.mli
```diff
@@ -56,6 +56,10 @@ val mk_enum_opt:
   cli:OpamCLIVersion.Sourced.t -> validity -> ?section:string -> string list ->
   string -> (validity * string * 'a) list -> string -> 'a option Term.t
 
+val mk_enum_opt_all:
```
rjbou, line 59: add some doc

master_changes.md
```diff
@@ -14,6 +14,10 @@ New option/command/subcommand are prefixed with ◈.
   * Add cli versioning for enums of flags with predefined enums [#4606 @rjbou]
   * Ensure the symlink for a plugin is maintained on each invocation [#4621 @dra27 - partially fixes #4619]
   * Clearer messages about using --cli and OPAMCLI [#4655 @dra27]
+  * Add cli versioning for enums of flags with predefined enums [#4626 @rjbou]
+  * ◈ Add `--confirm-level` and `OPAMCONFIRMLEVEL` [#4582 @rjbou - fix #4168]
```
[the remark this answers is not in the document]
rjbou: unicode powa! :)

src/core/opamCoreConfig.ml
```diff
@@ -157,11 +159,21 @@ let initk k =
         | true -> Some `Extended
         | false -> None)
     ) in
+  let answer =
+    (* priorities: confirm level > yes > no *)
+    match E.confirmlevel (), E.yes (), E.no () with
+    | Some c, _,  _ -> Some c
+    | _, Some true, _ -> Some `all_yes
+    | _, Some false, _
```
dra27, line 167:
```suggestion
    | _, Some false, _ -> None
```
At present, `OPAMYES=false opam ...` still prompts, where this would still answer no.

Issue #4448, opened 2020-11-24
opam master can't run commands on ocaml < 4.08.0 and Xcode 12


Blaisorblade:
On mac os x 10.15, opam master fails to run openssl, or sw_vers, while opam 2.0.7 works fine, and openssl and sw_vers are both available:

```
$ git describe
2.1.0-beta2-121-g51c160be
$ git branch
  2.0
  allow-git-fetch-pwd
  allow-git-fetch-pwd-backport
* master
$ ./configure && make lib-ext && make
$ dune exec --profile=release  --promote-install-files -- opam repo list
Fatal error:
"openssl": permission denied.
```
The error persists after make install.

The programs work.
```
$ which openssl
/usr/bin/openssl
$ ls -l $(which openssl )
-rwxr-xr-x 1 root wheel 510656 Sep 22 02:30 /usr/bin/openssl
$ openssl
OpenSSL> version
LibreSSL 2.8.3
OpenSSL> exit
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.15.7
BuildVersion:	19H15
```

opam config report output, as requested:
```
$ dune exec --profile=release  --promote-install-files -- opam config report
# opam config report
# opam-version         2.1.0~beta4 (51c160be9a018b69ee2ba6cc38c483565bf00550)
# self-upgrade         no
Fatal error:
"sw_vers": permission denied.
```

rjbou:
In both case, it shouldn't error with a simple `permission denied`.
For `openssl`, opam should fallback to internal sha computing. Could you share the stack trace of `opam repo list` ?
The [code](https://github.com/ocaml/opam/blob/master/src/core/opamSystem.ml#L382) for permission check can help to locate the problem.

Blaisorblade:
> Could you share the stack trace of `opam repo list` ?

I'm afraid I have no idea how

> The code for permission check can help to locate the problem.

Shouldn't that code use https://caml.inria.fr/pub/docs/manual-ocaml/libref/Unix.html#VALaccess instead? In C, `access` is implemented by a separate kernel API, which is not guaranteed to have the implementation you use.

Blaisorblade:
Looking at 41d44f9687cffbea9a9762b93c15880a7eb55b02 and https://github.com/ocaml/ocaml/issues/7640, I see why you reimplement `PATH` lookup, but that shouldn't prevent you from using `access`(2).

rjbou:
[...]

PR #4585, opened 2021-03-05
Make of OCaml 4.08 the new minimal requirement

kit-ty-kate:
OCaml 4.08 is required to get `opam-0install` out of the box inside opam by default.
https://github.com/ocaml/opam/issues/4448 also shows that old compilers are not well tested and can results in bugs hard to debug.
This proposal makes OCaml 4.08 the new minimal requirement for opam.

This brings a few caveats however. Mainly that users of the opam libraries will now require OCaml >= 4.08 if they wish to use the latest version of the opam libraries.


kit-ty-kate:
Superseded by #4775

PR #4588, opened 2021-03-08
Fix `opam exec` on Windows

AltGr:
split out from #4504

Expose `OpamProcess.create_process_env`, which has the correct way of calling 
cygwin executables from a Windows native opam (without unwanted expansion of 
arguments, etc.)

And use that for `opam exec` (instead of `Unix.execv`) so that it has the 
correct behaviour.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -7,7 +7,7 @@ New option/command/subcommand are prefixed with ◈.
   *
 
 ## Global CLI
-  *
+  * Fix `opam exec` on native Windows when calling cygwin executables (@AltGr)
 
 ## Init
   *
--- a/master_changes.md
+++ b/master_changes.md
@@ -7,7 +7,7 @@ New option/command/subcommand are prefixed with ◈.
   *
 
 ## Global CLI
-  * Fix `opam exec` on native Windows when calling cygwin executables (@AltGr)
+  * Fix `opam exec` on native Windows when calling cygwin executables [#4588 @AltGr]
 
 ## Init
   *
--- a/src/core/opamProcess.ml
+++ b/src/core/opamProcess.ml
@@ -310,6 +310,13 @@ let set_resolve_command =
     resolve_command_fn := resolve_command
 let resolve_command cmd = !resolve_command_fn cmd
 
+let create_process_env cmd args env cin cout cerr =
+  let resolve cmd = OpamStd.Option.default cmd @@ resolve_command cmd in
+  if Sys.win32 && OpamStd.Sys.is_cygwin_variant (resolve cmd) = `Cygwin then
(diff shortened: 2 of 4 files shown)


src/core/opamProcess.mli
```diff
@@ -214,3 +214,10 @@ type 'a job = 'a Job.Op.job
 (**/**)
 val set_resolve_command :
   (?env:string array -> ?dir:string -> string -> string option) -> unit
+
+(** Like Unix.create_process_env, but with correct escaping of arguments when
+    invoking cygwin executables *)
```
dra27, line 219:
```suggestion
    invoking cygwin executables from a native Windows executable. *)
```

src/core/opamProcess.ml
```diff
@@ -310,6 +310,13 @@ let set_resolve_command =
     resolve_command_fn := resolve_command
 let resolve_command cmd = !resolve_command_fn cmd
 
+let create_process_env cmd args env cin cout cerr =
+  let resolve cmd = OpamStd.Option.default cmd @@ resolve_command cmd in
+  if Sys.win32 && OpamStd.Sys.is_cygwin_variant (resolve cmd) = `Cygwin then
+    cygwin_create_process_env cmd args env cin cout cerr
+  else
+    Unix.create_process_env cmd args env cin cout cerr
```
dra27, line 318: FWIW:
```suggestion
let create_process_env =
  if Sys.win32 then
    fun cmd ->
      let resolved_cmd = resolve_command cmd in
      if OpamStd.(Option.map_default Sys.is_cygwin_variant `Native resolved_cmd) = `Cygwin then
        cygwin_create_process_env cmd
      else
        Unix.create_process_env
  else
    Unix.create_process_env
```
The `Sys.win32` hoist is a personal meta-programming foible of mine (i.e. everything based on `Sys.win32` should be done once, as the value can't change!) but getting partial application `OpamProcess.create_process_env "foo"` to do the resolve immediately is useful/better.

dra27 (commented):
Couple of suggestions...

src/core/opamProcess.mli
dra27: (and with apologies for changing my mind on the wording _again_!)

master_changes.md
```diff
@@ -7,7 +7,7 @@ New option/command/subcommand are prefixed with ◈.
   *
 
 ## Global CLI
-  *
+  * Fix `opam exec` on native Windows when calling cygwin executables (@AltGr)
```
rjbou, line 10:
```suggestion
  * Fix `opam exec` on native Windows when calling cygwin executables [#4588 @AltGr]
```

rjbou:
Should be fixed now

PR #4589, opened 2021-03-08
Sandbox script: change handling of TMPDIR

AltGr:
(also: port improvements from bwrap to sandbox_exec, and fix a few
bugs, in particular for detection of the sandbox and newer dune)

The `bwrap` script now bind-mounts the existing `$TMPDIR` read-only,
allowing proper communication from outside the sandbox to the inside;
and re-creates a tmpfs someplace else, re-defining `TMPDIR`.
It might break scripts that accessed `/tmp` directly (instead of
`TMPDIR`) though, since it won't be writeable from within the sandbox.

The previous behaviour was to bind-mount $TMPDIR to /tmp within the
sandbox, and re-set TMPDIR accordingly ; however, this meant that absolute
paths within TMPDIR could be incompatible between the outside and the
inside of the sandbox.

This became a visible problem with dune 2.8, which redefines TMPDIR to a
subdirectory (e.g. `/tmp/buildXXXX.dune`). If you happen to be running
tests with an OPAMROOT within $TMPDIR, your opam root will no longer be
accessible, and packages that run opam commands from within the sandbox
(which is discouraged, but well...) will break.

In our tests, run through dune, OPAMROOT gets created in
`/tmp/buildXXXX.dune/OPAM`, then packages get installed within a sandbox,
this is relocated to `/tmp/OPAM` but `OPAMROOT` and `OPAM_SWITCH_PREFIX`
remained set to below `/tmp/buildXXXX.dune`, and things broke.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -63,6 +63,10 @@ New option/command/subcommand are prefixed with ◈.
 ## Sandbox
   * Fix the conflict with the environment variable name used by dune [#4535 @smorimoto - fix ocaml/dune#4166]
   * Kill builds on Ctrl-C with bubblewrap [#4530 @kit-ty-kate - fix #4400]
+  * Linux: mount existing TMPDIR read-only, re-bind `$TMPDIR` to a separate tmpfs [#4589 @AltGr]
+  * Fix the sandbox check [#4589 @AltGr]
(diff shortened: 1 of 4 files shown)


rjbou:
which one is to merge, #4561 or this one ?

PR #4591, opened 2021-03-08
Fix and improve some depext messages

AltGr:


src/client/opamSolution.ml
```diff
@@ -386,12 +386,14 @@ let parallel_apply t
       (if bypass <> !bypass_ref then
          (let spkgs = OpamSysPkg.Set.Op.(bypass -- !bypass_ref) in
           OpamConsole.note
-            "The bypass of system package %s has been registered in this switch. You \
-             can use `opam option depext-bypass-=%s' to revert."
+            "The bypass of system package%s %s has been registered in this \
+             switch. You can use `opam option depext-bypass-=%s' to revert."
```
dra27, line 390: Possible clearer wording (for an already fairly complicated concept!):
```suggestion
            "System package%s %s %s bypassed in this \
             switch. You can use `opam option depext-bypass-=%s' to revert."
             (* ... *)
             (if OpamSysPkg.Set.cardinal skpgs > 1 then "are" else "is")
```
A further possibility might be to explain what bypassing means - i.e. describe it in terms of ignoring the depext?

dra27 (commented):
We can possibly make the message even clearer (although it's an advanced function!)

master_changes.md
```diff
@@ -52,6 +52,7 @@ New option/command/subcommand are prefixed with ◈.
 ## External dependencies
   * Handle macport variants [#4509 @rjbou - fix #4297]
   * Always upgrade all the installed packages when installing a new package on Archlinux [#4556 @kit-ty-kate]
+  * Improve messages to hint that answering `no` doesn't abort installation [@AltGr]
```
rjbou, line 55:
```suggestion
  * Improve messages to hint that answering `no` doesn't abort installation [#4591 @AltGr]
```

src/client/opamSolution.ml
AltGr: Does that still convey the idea that the change to the switch was permanently recorded ? Hm, I guess the fact that there is a way to revert it speaks for itself.

dra27: I think it's unclear in `System package%s %s %s` but the user will see either `System package foo is` or `System packages foo and bar are` both of which imply permanence (and, as you say, so does the revert instruction!)

dra27:
Is the message going to be updated (I'm not sure why I'm pinged for review?)

rjbou:
Pinged for review according last meeting notes :)

dra27:
Ah, I must have missed the “@dra27 to look at it after @AltGr pushes changes” 🤣

dra27 (approved):
LGTM, thanks!

PR #4592, opened 2021-03-08
Bump bundled dune version to 2.8.2

AltGr:
it has some fixes we require for not hanging CI in some situations.


rjbou:
to rebase after #4593 merge

Issue #4597, opened 2021-03-15
Compilation of some packages hanging!


AltGr:
Thanks @avsm for finding this, it seems it is a very recent regression.

```
$ cd /tmp
$ git clone https://github.com/avsm/debug-opam-dune-hang
$ opam switch create debug-opam-dune-hang/ -v --debug --yes
[...]
02:25.329  FILE(opam)             Wrote /tmp/debug-opam-dune-hang/_opam/.opam-switch/packages/dune-configurator.2.8.4/opam in 0.000s
02:25.329  PARALLEL               Job 44612439 finished
Processing 46/51: [bigstringaf: dune build]
02:25.329  PARALLEL               Starting job 268152395 (worker 2/24 -/1 -/5): λ mirage-crypto.0.8.10
02:25.329  SYSTEM                 rmdir /tmp/debug-opam-dune-hang/_opam/.opam-switch/build/mirage-crypto.0.8.10
02:25.335  ACTION                 prepare_package_source: mirage-crypto.0.8.10 at /tmp/debug-opam-dune-hang/_opam/.opam-switch/build/mirage-crypto.0.8.10
02:25.335  PARALLEL               Next task in job 268152395: $HOME/.opam/opam-init/hooks/sandbox.sh build dune build -p mirage-crypto -j 24
Processing 47/51: [bigstringaf: dune build] [mirage-crypto: dune build]
<hangs here>
```
Disabling the sandbox doesn't seem to help.

AltGr:
It seems that `mirage-crypto` builds with `dune` that calls `make` in `freestanding` that calls `opam config var prefix` which hangs ; killing the latter releases the build.

dra27:
Simpler repro: `opam switch create hangs ocaml-system mirage-crypto bigstringaf`. The issue only occurs during switch creation (i.e. `opam switch create works ocaml-system && opam install mirage-crypto bigstringaf` works).

The fault's been present since the first beta - it was introduced by c6aaa6074f2a4f2a187b60198c54865bd8785cb9 in #4314.

AltGr:
Thanks for the bisecting!

PR #4599, opened 2021-03-15
Downgrade global write lock after upgrading format

dra27:
Tentative fix for #4597 - is it strictly necessary to retain the lock at all after loading the global config?

This definitely needs a shiny new reftest - "all" that's required is a dummy package which calls `opam var prefix` as part of its build and to create a switch with that package.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -26,6 +26,7 @@ New option/command/subcommand are prefixed with ◈.
 ## Switch
   * Don't exclude base packages from rebuilds (made some sense in opam 2.0
     with base packages but doesn't make sense with 2.1 switch invariants) [#4569 @dra27]
+  * Don't hog the global write lock during switch creation [dra27 - fix #4597]
 
 ## Pin
   * Don't look for lock files for pin depends [#4511 @rjbou - fix #4505]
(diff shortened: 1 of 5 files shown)


AltGr:
Test should be fairly easy to add, see the [short documentation](https://github.com/ocaml/opam/blob/master/tests/reftests/run.ml#L12-L36) and examples in `legacy-local.git`.

I may actually be in favor of completely reverting #4597 for now... if I know one thing about locks it's that their flow must really be kept as straight-forward as possible. I was already uncomfortable with adding `with_flock_upgrade` (you need to be really sure that you can't possibly hold a lock someone else might be waiting on before calling that, or at least that they can't be concerned with what you are upgrading... but since these things are all over the place...)
So the fix is probably correct, but it seems too complex to me to be really sure...

rjbou:
Looking at #4314, there was already a [write lock upgrade](https://github.com/ocaml/opam/pull/4314/files#diff-c1c5115aa2b63bf0a31ca269d682c57d4ab22bccd50f86ca4c91b6dd354fe3efL1083), but it was located in the upgrade function. With the new on the fly mechanism, it is no more needed to always take a write lock. This temporary write upgrade lock can be reintroduced temporarily in `OpamFormatUpgrade.as_necessary`.

rjbou:
In fact, it shouldn't be needed to release the lock, as `global_lock` is only used with `OpamFilename.with_flock_upgrade f` that restores previous lock level after `f`. In previous version, it was restored to a write lock, so it should be more blocking. I need to take a closer look at these mechanism. Lock upgrading/downgrading is really touchy!

rjbou:
Superseded by #4612.
Thanks!

PR #4601, opened 2021-03-16
Update lib-ext extlib (required for OCaml 4.12 builds)

AltGr:


kit-ty-kate:
sadly extlib 1.7.8 breaks dose3 < 6.0. In opam-repository we went the route of patching 1.7.7 while the revdeps catch up https://github.com/ocaml/opam-repository/pull/18041

See https://github.com/ocaml/opam/pull/4470 for the full upgrade needed to bump to extlib 1.7.8 and https://github.com/ocaml/opam/pull/4437 for the middle ground.

AltGr:
Ah I knew I was missing something :sweat_smile:

PR #4606, opened 2021-03-17
Cli version - environment variable edition

rjbou:
New opam environment variable handling. They are defined per lib, and defined and loaded only at opam client start (no more on library loading). There is no more direct access, each environment variable have its associated field in its lib config, or can be accessed throw `OpamXxxConfig.Env.<var>` function.
When defined in `OpamArg`, the cli version is required, to handle cli versioning of environment variable in a slightly different way than options: if not defined in current cli, a warning is printed and it is ignored.


rjbou:
Remaining:
* [ ] preinit env vars (`OPAMYES`, `OPAMSELFUPGRADE`)
* [ ] admin command env vars (`OPAMNOAGGREGATE`)
* [ ] add missing env vars doc
* [ ] specify env vars deprecation
* [ ] specialise env var cli warning messages?
* [ ] add test

PR #4607, opened 2021-03-17
Introduce 'default-invariant' besides 'default-compiler'

AltGr:
(todo: add tests)

compared to 2.0, this just adds the new 'default-invariant', and keeps
similar semantics for 'default-compiler', which is only used for 'opam
init' (but must now be compatible with the invariant).

previously the 2.1 branch completely replaced the older meaning of
'default-compiler', using it as 'default-invariant' instead, but the order
of priority was lost.

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -373,8 +373,14 @@ let init cli =
            ~check_sandbox:(not no_sandboxing)
            (OpamFile.Config.safe_read config_f) shell;
       else
-        OpamEnv.setup root ~interactive ~dot_profile ?update_config ?env_hook
-          ?completion ~inplace shell
+        (if not interactive &&
+            update_config <> Some true && completion <> Some true && env_hook <> Some true then
+           OpamConsole.msg
+             "Opam was already initialised. If you want to set it up again, \
+              use `--interactive', `--reinit', or choose a different \
+              `--root'.\n";
+         OpamEnv.setup root ~interactive ~dot_profile ?update_config ?env_hook
+           ?completion ~inplace shell)
     else
     let init_config =
       get_init_config ~no_sandboxing
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -2538,15 +2538,14 @@ let switch cli =
     apply_build_options build_options;
     let invariant_arg ?repos rt args =
       match args, packages, formula, empty with
-      | [], None, None, false ->
-        OpamFile.Config.default_invariant rt.repos_global.config
+      | [], None, None, false -> None
       | _::_ as packages, None, None, false ->
-        OpamSwitchCommand.guess_compiler_invariant ?repos rt packages
+        Some (OpamSwitchCommand.guess_compiler_invariant ?repos rt packages)
       | [], Some atoms, None, false ->
         let atoms = List.map (fun p -> Atom p) atoms in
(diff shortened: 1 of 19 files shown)


AltGr:
Ok, so apparently this works... but the perf is not acceptable at the moment :/
(>10s on a high-end PC)

AltGr:
Ok, tests added, and now reasonably performant

src/client/opamInitDefaults.ml
```diff
@@ -19,6 +19,14 @@ let repository_url = {
 }
 
 let default_compiler =
+  OpamFormula.ors [
+    OpamFormula.Atom (OpamPackage.Name.of_string "ocaml-system",
+                      OpamFormula.Empty);
```
dra27, line 24: What happens if the user has an old system compiler (e.g. they happen to be on a really old Ubuntu)? Does it quietly ignore it and move on to ocaml-base-compiler, or should we duplicate the `>= "4.05"` constraint here too?

src/client/opamCommands.ml
```diff
@@ -394,22 +403,32 @@ let init cli =
...
+       | comp -> " - initially with "^ (OpamFormula.string_of_atoms comp));
     let (), st =
       try
         OpamSwitchCommand.create
           gt ~rt ~invariant ~update_config:true (OpamSwitch.of_string name) @@
-        (fun st -> (), OpamSwitchCommand.install_compiler st)
+        (fun st ->
+           (),
+           OpamSwitchCommand.install_compiler st
+             ~additional_installs:default_compiler)
```
dra27, line 431: This causes an interactive prompt in `OpamSwitchCommand.install_compiler`

dra27 (commented):
This looks great, apart from the interactive prompt at `opam init`.

What happens if there's no installable compiler (e.g. because the user points at an empty repo or something like that)? Is it a reasonable error message or does opam explode? 🙂

src/client/opamInitDefaults.ml
AltGr: The former: this is in the tests here: https://github.com/ocaml/opam/pull/4607/files#diff-b3f6abdb5ab3eb6c5612e441c0709542022442b9cc6f47b5eae75bc5aa85799cR36-R54

src/client/opamCommands.ml
AltGr: Ah, well spotted, indeed that's unfortunate.

dra27:
Hopefully that fixes the testsuite on AppVeyor. The problem was twofold:

- The msvc test is done in Visual Studio 2017, although the relevant difference is that AppVeyor uses Server 2016 for that image instead of Server 2019. Server 2019 has a Windows built bsdtar in `C:\Windows\system32\tar.exe`. This meant that on Windows 10 1803+ / Server 2019 a Windows-formatted path ended up being passed to a Windows tar, so it worked, but more by luck than judgement, because...
- The fix was to identify that we're invoking tar and turn on cygpath conversion if it turns out we're calling the Cygwin version of tar. This fixes Server 2016, but then we break Server 2019! It turns out we get bitten by https://github.com/ocaml/ocaml/issues/7662. At present, executables in `C:\Windows\system32` always win. I've solved this by resolving the command first - for now, on Windows. It's likely that OCaml's Unix library will get fixed.

AltGr:
ow thanks !

PR #4609, opened 2021-03-19
MacOS sandbox: fix realpath use

AltGr:
Oops, had this fix lying around on the mac but it got lost in the final reftests PR.


AltGr:
Sandboxing issue gone, but the MacOS CI still has a solver problem on opam-rt :thinking:

dra27:
That seems to get us to _just_ the solver timeout!

PR #4610, opened 2021-03-19
Github CI : (macos) fixes

AltGr:
and add some debug to trace the weird solver timeout error

### DIFF
--- a/.github/scripts/main.sh
+++ b/.github/scripts/main.sh
@@ -84,7 +84,7 @@ export OCAMLRUNPARAM=b
       git reset --hard origin/master
     fi
 
-    test -d _opam || opam switch create . --empty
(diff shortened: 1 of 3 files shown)


.github/scripts/main.sh
```diff
@@ -87,8 +87,8 @@ export OCAMLRUNPARAM=b
     test -d _opam || opam switch create . --empty
     eval $(opam env)
     opam pin --kind=path $GITHUB_WORKSPACE --yes --no-action
-    opam pin . -yn
-    opam install opam-rt --deps-only
+    opam pin . -yn --ignore-pin-depends
```
rjbou, line 90: It is already tested on current branch

dra27: The issue is the `--empty` 3 lines above... if ocaml-system is selected (which is available from the cache) then this proceeds

PR #4615, opened 2021-03-26
Don't connect stdin to sub-processes

dra27:
Split off from #3897. @AltGr agreed this was worth doing in https://github.com/ocaml/opam/pull/3897#pullrequestreview-348824024, but it's too close to the 2.1 release for a change which isn't a critical part of that fix, so this can be merged post-2.1


dra27:
Indeed - cross-referencing https://github.com/ocaml/opam/pull/4447.

PR #4616, opened 2021-03-26
src_ext / cold updates

dra27:
- cold compiler to 4.12.0 (allows bootstrapping on Cygwin64)
- src_exts updated to latest versions (apart from dose3 and extlib)
- Tidy-up an old bit of jbuilder needed code in the vendoring system
- Add quite a few `-w -3` so the lib-ext build is quieter
- Update two CI builds to latest point release


dra27:
How embarrassing. I appear to have revealed the OS I didn't test this on...

src_ext/patches/ocamlgraph/restore-4.02-support.patch
```diff
@@ -0,0 +1,23 @@
...
+ }
+ 
+diff -Naur a/src/dune b/src/dune
+--- a/src/dune	2020-10-02 13:47:57.000000000 +0000
++++ b/src/dune	2021-03-26 12:33:07.160324274 +0000
+@@ -3,6 +3,7 @@
+ (library
+  (name graph)
+  (public_name ocamlgraph)
++ (flags -w -3)
```
kit-ty-kate, line 20: shouldn't this be `(flags :standard -w -3)` instead?

kit-ty-kate: (same for the rest of the `(flags -w -3)` additions)

dra27: Oops, yes!

src_ext/Makefile.packages
```diff
@@ -41,9 +41,13 @@ dose3.pkgbuild: findlib.pkgbuild ocamlbuild.pkgbuild cppo.pkgbuild extlib.pkgbui
...
 mccs.pkgbuild: dune-local.pkgbuild cudf.pkgbuild findlib.pkgbuild
 
+stdlib-shims.pkgbuild: dune-local.pkgbuild
+
+dune-local.pkgbuild: findlib.pkgbuild
+
 opam-file-format-pkg-build:
 	dune build @install -p opam-file-format
-	dune install -p opam-file-format
+	dune install -p opam-file-format opam-file-format
```
kit-ty-kate, line 50: I'm curious, are these changes really required? Isn't that already implied by the `-p <pkg>` argument?

dra27: It turns out they're not (I assumed so, too). You get an error with ocamlgraph complaining that `ocamlgraph_gtk.install` hasn't been built.

kit-ty-kate (approved):
apart from my last (very minor) comment, if CI is happy this seems all good.

src_ext/Makefile.packages
dra27: I put that in for the other just for belt-and-braces!

kit-ty-kate: I see, interesting

PR #4620, opened 2021-03-31
Clean-up our opam files

dra27:
A few house-keeping amendments:

- Our own opam files should really pass `opam lint` 🙈 (re-wrote the `synopsis` fields not to have `opam` as the first word)
- `opam-format` should pedantically have an explicit dependency on `re`
- `opam.opam` and `opam-admin.opam` changed from being empty files to be minimally populated (they pass both opam lint and opam dune-lint)

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -119,6 +119,7 @@ New option/command/subcommand are prefixed with ◈.
   * Fix temporary file with a too long name causing errors on Windows [#4590 @AltGr]
   * CLI: Add flag deprecation and replacement helper [#4595 @rjbou]
   * Win32 Console: fix VT100 support [#3897 @dra27]
+  * Tidied the opam files [#4620 @dra27]
 
 ## Test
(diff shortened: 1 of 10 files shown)


rjbou (approved):
Thanks!

dra27:
I figured we didn't need `version` in `opam.opam` because it's never pinned or used? Essentially, opam-dune-lint doesn't like empty opam files (although I think we may have fixed that) and having populated with `opam-version: "2.0"` it seemed best to make it pass opam lint.

If we do put a version in there, `opam-admin.opam` wants one too?

rjbou:
`opam-admin.opam` don't use the variable `version`, that seems to [confuse](https://github.com/ocaml/opam/runs/2244158321?check_suite_focus=true) opam for `opam.opam`:
```
opam install . --deps-only --yes
  - opam → opam-client < 2.0~alpha5
      no matching version
```

dra27:
Oh, I see now!

Issue #4619, opened 2021-03-30
Invalidate plugins at root upgrade


dra27:
This `Dockerfile` fails at the final step:
```
FROM ocaml/opam:debian-ocaml-4.12
WORKDIR /src
RUN opam install dune
RUN sudo chown opam /src
RUN opam exec -- dune init project foo
RUN echo > dune-project '(lang dune 2.8)\n(generate_opam_files true)(source (github foo/bar))(package (name foo))'
RUN opam exec -- dune build
RUN opam install opam-state.2.0.8
RUN opam install opam-dune-lint --deps-only
RUN opam pin -yn opam-dune-lint 'https://github.com/ocurrent/opam-dune-lint.git#790f4d552d37cfeb6c13c5d1d543ef7da576ad1d'
RUN opam dune-lint
RUN sudo rm /usr/bin/opam && sudo ln /usr/bin/opam-2.1 /usr/bin/opam
RUN opam update
RUN opam dune-lint
```

The reason is that `opam-dune-lint` was compiled with `opam-state` 2.0.x and then the root gets upgraded. https://github.com/ocurrent/opam-dune-lint/pull/26 proposes fixing its dependencies to require `opam-state` at least matching the opam version. This fixes manually installing `opam-state` 2.0.x in a 2.1 root by forcing the plugin to be uninstalled (cf. https://github.com/ocurrent/opam-dune-lint/pull/26#issuecomment-810030923), but I think we need to do _something_ about the upgrade case too.

A simple option would be to delete all the symlinks in `plugins/bin` as part of the format upgrade, but that doesn't work in this case where the plugin is installed in the current switch. More complex solutions might include a _validity_ check on the dependencies of a plugin (possibly just once after upgrade?) - if the pin is changed to `bf451c8cf2d4f2be3c7d979b613c79588b55a172`, then `opam-dune-lint` definitely needs recompiling after the root upgrade (`opam update` will catch that) but it would also be detectable that the current installation is invalid so `opam dune-lint` could refuse to run the plugin without reinstallation?

Thoughts, @AltGr? It will affect any plugin which uses the opam libraries which, ideally, should be all of them!

dra27:
(NB - this should _resolved_ before the release candidate but that resolution may include "won't fix")

AltGr:
[...]

PR #4621, opened 2021-03-31
Reinstall plugin package if the symlink is missing

dra27:
This is a partial fix for #4619 (the other part would be to remove the plugin symlinks at format upgrade).

This tweaks the plugin detection slightly so that if the symlink is missing, the command is still resolved, but package will be reinstalled in the current switch. This will recreate the symlink and also apply any changes to dependencies (e.g. after an opam client upgrade).

It's demonstrated by this slightly convoluted Dockerfile - commenting out the final `rm` command shows the existing error.

```
WORKDIR /src
RUN sudo chown opam /src
RUN git clone https://github.com/dra27/opam.git
WORKDIR /src/opam
RUN opam pin add -n dose3 5.0.1-1
RUN git checkout reinstall-plugins ; opam install dose3 opam-file-format mccs cmdliner
WORKDIR /src/test
RUN sudo chown opam /src/test
RUN opam install dune
RUN opam exec -- dune init project foo
RUN echo > dune-project '(lang dune 2.8)\n(generate_opam_files true)(source (github foo/bar))(package (name foo))'
RUN opam exec -- dune build
RUN opam install opam-state.2.0.8
RUN opam install opam-dune-lint --deps-only
RUN opam pin -yn opam-dune-lint 'https://github.com/ocurrent/opam-dune-lint.git#bf451c8cf2d4f2be3c7d979b613c79588b55a172'
RUN opam dune-lint
#RUN sudo rm /usr/bin/opam && sudo ln /usr/bin/opam-2.1 /usr/bin/opam
RUN sudo rm /usr/bin/opam && sudo cp /src/opam/opam /usr/bin/opam
RUN opam update
RUN rm ~/.opam/plugins/bin/*
RUN opam dune-lint
```

### DIFF
--- a/src/client/opamCliMain.ml
+++ b/src/client/opamCliMain.ml
@@ -138,6 +138,9 @@ let check_and_run_external_commands () =
     let root_dir = OpamStateConfig.opamroot () in
     let has_init = OpamStateConfig.load_defaults root_dir <> None in
     let plugins_bin = OpamPath.plugins_bin root_dir in
+    let plugin_symlink_present =
+      OpamFilename.is_symlink (OpamPath.plugin_bin root_dir (OpamPackage.Name.of_string name))
+    in
     let env =
       if has_init then
         let updates =
@@ -154,11 +157,11 @@ let check_and_run_external_commands () =
         Unix.environment ()
     in
     match OpamSystem.resolve_command ~env command with
-    | Some command ->
+    | Some command when plugin_symlink_present ->
       let argv = Array.of_list (command :: args) in
       raise (OpamStd.Sys.Exec (command, argv, env))
     | None when not has_init -> (cli, argv)
-    | None ->
+    | cmd ->
       (* Look for a corresponding package *)
       match OpamStateConfig.get_switch_opt () with
       | None -> (cli, argv)
@@ -183,7 +186,7 @@ let check_and_run_external_commands () =
         in
         let installed = OpamPackage.Set.inter plugins st.installed in
         if OpamPackage.Set.is_empty candidates then (cli, argv)
-        else if not OpamPackage.Set.(is_empty installed) then
+        else if not OpamPackage.Set.(is_empty installed) && cmd = None then
           (OpamConsole.error
              "Plugin %s is already installed, but no %s command was found.\n\
               Try upgrading, and report to the package maintainer if \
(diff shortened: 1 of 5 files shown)


src/client/opamCliMain.ml
```diff
@@ -206,18 +208,28 @@ let check_and_run_external_commands () =
         then
           let nv =
             try
-              OpamPackage.max_version plugins
-                (OpamPackage.Name.of_string prefixed_name)
+              (* If the command was resolved, attempt to find the package to reinstall. *)
+              if cmd = None then
+                raise Not_found
+              else
+                OpamPackage.package_of_name installed (OpamPackage.Name.of_string prefixed_name)
```
AltGr, line 215: Won't this cause problems if the currently installed version is no longer available ? This could happen e.g. with `available:` depending on `opam-version` after an opam root upgrade, where an upgrade of the plugin becomes mandatory, and cause an error during resolution, requiring a manual `opam upgrade`.
I don't remember why this code used to force the latest version (within availability) though

AltGr (approved):
Nice!

src/client/opamCliMain.ml
dra27: It turns out it doesn't, but I think that's a bug - the availability check is not done for `opam reinstall`!

The alternative would be to upgrade the package by name instead?

dra27:
Need to double-check the `reinstall` question above, but otherwise this is good to go.

src/client/opamCliMain.ml
rjbou: The `candidates` package already have the availability check, so if a plugin is no more available it won't be in `installed` and its latest version will be "newly" installed:
```
$ opam install opam-depext.1.1.3 #without opam < 2.1 condition
$ opam update # reintroduce opam < 2.1 condition
$ rm <link>
$ opam depext
Opam plugin "depext" is not installed. Install it on the current switch? [Y/n] y
opam-depext.1.1.5 is not installed. Install it? [Y/n] y
  ↗ upgrade opam-depext 1.1.3 to 1.1.5

⬇ retrieved opam-depext.1.1.5  (cached)
⊘ removed   opam-depext.1.1.3
∗ installed opam-depext.1.1.5
Done.
```

On the availability check for reinstall, this (specific) case is handled by the `orphans` mechanism, and can't be reinstalled:
```
$ opam install opam-depext.1.1.3 #without opam < 2.1 condition
$ opam update # reintroduce opam < 2.1 condition
$ opam reinstall opam-depext
[ERROR] Sorry, these packages are no longer available from the repositories: opam-depext (= 1.1.3)
```

rjbou (approved):

rjbou:
A rebase and good to merge for me!

dra27:
I agree - the `reinstall` question is already fixed - IIRC I tested on an old opam!

PR #4626, opened 2021-04-06
Cli versioning for enums in flags with enum

rjbou:
like `--color [always|never|auto]`

### DIFF
--- a/src/client/opamArgTools.ml
+++ b/src/client/opamArgTools.ml
@@ -298,14 +298,31 @@ let mk_vflag_all ~cli ?section ?(default=[]) flags =
   let default = List.map (fun x -> `Valid x) default in
   term_cli_check ~check Arg.(vflag_all default info_flags)
 
-let mk_state_opt ~cli validity ?section flags value state doc =
+let string_of_enum enum =
+  Arg.doc_alts_enum (List.map (fun (_, s, v) -> s,v) enum)
+
+let mk_enum_opt ~cli validity ?section flags value states doc =
   let doc = update_doc_w_cli doc ~cli validity in
   let doc = Arg.info ?docs:section ~docv:value ~doc flags in
   let check elem =
-    check_cli_validity cli validity ~cond:(fun c -> c && elem <> None)
-      elem flags
+    (* first check validity of flag *)
+    let flag_validity =
+      check_cli_validity cli validity ~cond:(fun c -> c && elem <> None)
+        elem (`Flags flags)
+    in
+    (* then check validity of the argument *)
+    match flag_validity with
+    | `Ok (Some elem) ->
+      let validity, str, _ = List.find (fun (_,_,v) -> v = elem) states in
+      check_cli_validity cli validity (Some elem)
+        (`Verbatim
+           (Printf.sprintf "the %s option for %s"
+              (OpamConsole.colorise `bold str)
+              (OpamConsole.colorise `bold ("--"^get_long_form flags))))
+    | _ -> flag_validity
   in
-  term_cli_check ~check Arg.(opt (some (enum state)) None & doc)
+  let states = List.map (fun (_, s, v) -> s,v) states in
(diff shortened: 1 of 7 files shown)


rjbou:
Should we keep `OpamArg.mk_tristate_opt` as the legacy  enum with `[ 'Always | 'Never | 'Auto ]` ?

src/client/opamArgTools.ml
```diff
@@ -289,11 +289,28 @@ let mk_vflag_all ~cli ?section ?(default=[]) flags =
...
+    let flag_validity =
+      check_cli_validity cli validity ~cond:(fun c -> c && elem <> None)
+        elem flags
+    in
+    (* then check validity of the argument *)
+    match flag_validity with
+    | `Ok (Some elem) ->
+      let validity, str, _ = List.find (fun (_,_,v) -> v = elem) states in
+      check_cli_validity cli validity (Some elem)
+        [ Printf.sprintf "%s with %s" (get_long_form flags) str ]
```
dra27, line 309: The sentence reads oddly to me (I deprecated `--shell=bash` 😂 - `opam: shell with bash was ...` is initially a bit confusing). Various options:
- Put the two names in "bold" (like `OPAMCLI` and the version later in the message)
- Display `--%s=%s`
- `the %s option for --%s` (with both parts in bold)

dra27 (commented):
I think a tweak to the message displayed is worth doing.

Incidentally, the same tweak could possibly be applied to the existing message:
```
dra@thor:~/opam-2$ ./opam --cli=2.1 install ocamlclean --unlock-base
opam: unlock-base was removed in version 2.1 of the opam CLI, but version 2.1 has been requested. Use --update-invariant instead or set OPAMCLI environment variable to 2.0.
```
I think it would be better for the display to be `--unlock-base` (and coloured bold in the same way as `--update-invariant` is)

rjbou:
Initially, there is no `--` added because the error/warning function is the one used for flags, commands, and subcommands. I'll check if there is a simple way to add it.

dra27:
Indeed - it's correct for commands subcommands - although flags could do with it as well (e.g. the `--unlock-base` example)

src/client/opamArgTools.ml
rjbou: Finally the third one

rjbou:
*et hop!* updated

src/client/opamArgTools.ml
```diff
@@ -77,16 +77,23 @@ let get_long_form flags =
       if flgth > lgth then (flgth, f) else (lgth, long))
     (0,"") flags |> snd
 
-let newer_flag_msg cli valid_since elem =
+let string_of_target = function
+   | `Flags flags ->
+     OpamConsole.colorise `bold
+       "--"^get_long_form flags
```
dra27, line 83:
```suggestion
       ("--"^get_long_form flags)
```
🙂

src/client/opamArgTools.ml
```diff
@@ -279,21 +290,39 @@ let mk_vflag_all ~cli ?section ?(default=[]) flags =
...
+        elem (`Flags flags)
+    in
+    (* then check validity of the argument *)
+    match flag_validity with
+    | `Ok (Some elem) ->
+      let validity, str, _ = List.find (fun (_,_,v) -> v = elem) states in
+      check_cli_validity cli validity (Some elem)
+        (`Verbatim
+           (Printf.sprintf "the %s option for --%s"
+              (OpamConsole.colorise `bold str)
```
dra27, line 320:
```suggestion
           (Printf.sprintf "the %s option for %s"
              (OpamConsole.colorise `bold ("--" ^ str))
```

dra27 (commented):
Tweaking what gets the bold attribute, but the changes the otherwise look lovely, thanks!

dra27 (approved):
LGTM!

PR #4627, opened 2021-04-06
Fix bug with Cudf pre-processing leading to some packages unnecessarily removed

### LINKED ISSUE #4624: 2.1.0~beta4: solver confusion
I have a very strange behavior with the solver:

```
$ opam install optint.0.1.0
  - remove    git-unix                3.3.3          [uses git]
  - remove    git-cohttp-mirage       3.3.3          [uses git]
  - remove    git-cohttp-unix         3.3.3          [uses git]
  - remove    git-cohttp              3.3.3          [uses git]
  - remove    git                     3.3.3          [conflicts with optint]
  - remove    carton-git              0.4.0          [uses carton]
  - remove    carton-lwt              0.4.0          [conflicts with optint]
  - remove    carton                  0.4.0          [conflicts with optint]
  - remove    decompress              1.3.0          [conflicts with optint]
  - remove    checkseum               0.3.0          [conflicts with optint]
  - recompile dune                    2.8.2          [uses carton]
  - upgrade   optint                  0.0.4 to 0.1.0
[..]
```
and
```
$ opam install git.3.3.3 optint.0.1.0
[NOTE] Package git is already installed (current version is 3.3.3).
  - recompile dune                    2.8.2          [upstream or system changes]
  - upgrade   optint                  0.0.4 to 0.1.0
[..]
```

Here is my configuration:
```
# opam config report
# opam-version         2.1.0~beta4 (5582bf9c975d93b0bee4b7b18c6e7b0f90338317)
# self-upgrade         no
# system               arch=x86_64 os=macos os-distribution=homebrew os-version=11.2.3
# install-criteria     -removed,-count[hidden-version,changed],-count[version-lag,request],-count[version-lag,changed],-count[missing-depexts,changed],-changed
# upgrade-criteria     -removed,-count[hidden-version,changed],-count[version-lag,solution],-count[missing-depexts,changed],-new
# jobs                 7
# repositories         1 (http), 4 (version-controlled) (default repo at ad4dd344)
# pinned               0
# current-switch       /Users/thomas/git/irmin
# read-state           Not_found
```

And the corresponding CUDF files:

[optint.cudf](https://github.com/ocaml/opam/files/6242533/optint.cudf.txt)
[optint+git.cudf](https://github.com/ocaml/opam/files/6242534/optint%2Bgit.cudf.txt)

AltGr:
Closes #4624

The issue was that, after pre-processing, we were only considering the 
dependency cone of required packages, plus already installed packages.

It happens this wasn't enough, because reverse dependencies may need to be 
upgraded: in this case, we wouldn't see the new version, and so just remove
the package.

However, once you do that, these newly included package versions may need to 
bring in new dependencies; we could just add their full depdendencies to the
set of packages we need to consider, but that would end up severely reducing
the scope of the trimming;

so, going further, we split the computation of dependencies into strict 
requirements, on the requested packages, bringing constraints that can always
be enforced ; and simple dependencies, computed for all other packages that
need to be considered.

For example, if requiring A, where A.1 depends on X>3 but A.2 doesn't, X won't 
be a strong requirement of the request, hence installed reverse dependencies
of A could bring in lower versions of X.

This patch also automatically enables "simple" trimming for maximisation 
requests, because the simplified trimming won't remove any coinstallable 
packages.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -110,7 +110,7 @@ New option/command/subcommand are prefixed with ◈.
   *
 
 ## Solver
-  * Fix Cudf preprocessing [#4534 @AltGr]
+  * Fix Cudf preprocessing [#4534 #4627 @AltGr - fix #4624]
   * Allow to upgrade to a hidden-version package if a hidden-version package is already installed [#4525 @kit-ty-kate]
 
 ## Client
--- a/src/solver/opamCudf.ml
+++ b/src/solver/opamCudf.ml
@@ -422,50 +422,96 @@ type conflict =
 module Map = OpamStd.Map.Make(Package)
 module Set = OpamStd.Set.Make(Package)
 
+let strong_and_weak_deps u deps =
+  (* strong deps are mandatory (constraint appearing in the top conjunction)
+     weak deps correspond to optional occurrences of a package, as part of a
+     disjunction: e.g. in (A>=4 & (B | A<5)), A>=4 is strong, and the other two
+     are weak. In the end we want to retain B and A>=4. *)
+  List.fold_left (fun (strong_deps, weak_deps) l ->
+      let names =
+        List.fold_left (fun acc (n, _) ->
+            OpamStd.String.Map.add n Set.empty acc)
+          OpamStd.String.Map.empty l
+      in
+      let set =
+        List.fold_left (fun acc (n, cstr) ->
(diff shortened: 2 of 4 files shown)


rjbou (approved):
Thanks!

PR #4631, opened 2021-04-13
Add support for a few select criteria useful to CI to the 0install solver

kit-ty-kate:
- `+count[version-lag,solution]` tells the solver to always choose the oldest version available instead of the latest (useful to test lower-bound constraints)
- `+removed` tells the solver to not bother trying to keep installed packages. This is useful for CI where you might want consecutive runs of opam to be reproducible.


kit-ty-kate:
For context, in my original custom branch for opam-health-check, I used to use a custom environment variable that would be only used by the opam-0install solver. However I'm not sure this approach is very scaleable as I'm adding more options. Criterias are already there and should satisfy most options I would want from the solver in the future, so I switched to using them in the process of upstreaming.

AltGr (approved):

rjbou:
Thanks!

PR #4632, opened 2021-04-13
Always avoid merge conflicts in master_changes.md

kit-ty-kate:
Simple quality of life improvement.
Inspired by https://twitter.com/dbrgn/status/1381547785116344323

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -19,3 +19,6 @@ tests/patcher-test.reference text eol=lf
 
 # Actual binary files
 *.pdf binary
+
+# Avoid conflicts in master_changes.md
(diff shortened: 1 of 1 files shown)


AltGr:
Sounds like a great idea :)
Even if we sometimes end up with duplicate lines, those should be easy to resolve upon release.

rjbou:
A good helper to have, thanks!

>  Even if we sometimes end up with duplicate lines, those should be easy to resolve upon release.

Definitively,  `master_changes.md` file is here to help write release note/changes, and always contains more information, some additional lines is harmless

PR #4635, opened 2021-04-15
lint W59/E60: don't check vcs upstream

rjbou:
Highlighted in https://github.com/ocurrent/opam-repo-ci/issues/93

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -55,7 +55,9 @@ New option/command/subcommand are prefixed with ◈.
   *
 
 ## Lint
-  * fix W59 & E60 with conf flag handling (no url required) [#4550 @rjbou - fix #4549]
+  * Fix W59 & E60 with conf flag handling (no url required) [#4550 @rjbou - fix #4549]
+  * Fix W59 & E60 with VCS urls, don't check upstream if url has VCS backend [#4635 @rjbou]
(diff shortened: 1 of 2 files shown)


dra27:
What's actually checked with an rsync URL?

dra27:
From dev meeting: the test needs to be http or rsync pointing to a file (rather than a directory)

dra27:
This could potentially have a test for whether a checksum has been added when there _shouldn't_ be one

kit-ty-kate (approved):
Tested successfully and looks good to me overall. I think the details about rsync can be figured out later, it's less critical.

dra27:
> I think the details about rsync can be figured out later, it's less critical.

The rsync file/directory part should be figured out now (we shouldn’t close one bug by creating another), but I agree the extra checksum check can wait. Regardless, please can issues be opened to track anything which isn’t fixed now 🙂

rjbou:
Just finished both, no need of new issue :)
In fact the good check is too look is it is an archive url, regardless of the backend

rjbou:
reviewed mechanism and add E67 for url present and no archive url

kit-ty-kate:
I've tested a number of cases with it and it seems good to go for me.

PR #4638, opened 2021-04-20
Error-free opam root changes

### LINKED ISSUE #4636: Introduce opam-root-version
Discussed on 1 Apr, introduce `opam-root-version` for the root and switch config files which is checked for taking out write locks. `opam-version` is checked for taking out read locks. So after root upgrade, `$OPAMROOT/config` would contain:

```
opam-version: "2.0"
opam-root-version: "2.1"
```

opam-state 2.0.x should be updated to internally recognise `opam-root-version` (i.e. no change in API in `OpamFile`) and ignore fields it doesn't recognise. The changes we've made for 2.1 mean that it should "just work" to have readable roots to 2.0.

As always, a given version of opam would never attempt to read a file with `opam-version` greater than itself.

rjbou:
Permit a non blocking call from older versions of opam binary or lib.

On files, it adds `opam-root-version` field in root `config` file to discriminate root version and opam file version. The field `opam-version` is the `opam-file-format` version of the file. In `OpamFile` modules, `format_version` field is set on some modules to specify the file format version (eg. new field), and `opam_file_format` the `opam-file-format` version. An internal `OpamFile.Config.root_version` permit to determine the current root version of the binary. it need to be bumped at each `opam_file_format` bump (regardless of the file), according opam binary version. 
`opam-root-version` is checked for all states loading, and for most state files reading. Also for root format upgrade mechanism.
The check/printing on the `opam-version` field is updated too. It must be printed on all files, and it is checked that it will be printed (`OpamFormat.opam_version`).

On format upgrade, latest compatible hard upgrade is still 2.0~beta5, but intermediate roots (from 2.1~alpha's) are set as hard upgrade: all files with `opam-version: 2.1` downgraded to `2.0`.

On state files should no more use `OpamFile` to read them outside state loading, there is functions to safely read them in `OpamStateConfig`. It concerns `config`, `switch-config`, `switch-state`, and `repos-config`.

Fixes #4636

### DIFF
--- a/src/client/opamCliMain.ml
+++ b/src/client/opamCliMain.ml
@@ -302,8 +302,8 @@ let rec main_catch_all f =
   | OpamFormatUpgrade.Upgrade_done conf ->
     main_catch_all @@ fun () ->
     OpamConsole.header_msg "Rerunning init and update";
-    OpamClient.reinit ~interactive:true ~update_config:false conf
-      (OpamStd.Sys.guess_shell_compat ());
+    OpamClient.reinit ~interactive:true ~update_config:false ~bypass_checks:true
+      conf (OpamStd.Sys.guess_shell_compat ());
     OpamConsole.msg
       "Update done, please now retry your command.\n";
     exit (OpamStd.Sys.get_exit_code `Aborted)
--- a/src/client/opamCliMain.ml
+++ b/src/client/opamCliMain.ml
@@ -141,7 +141,7 @@ let check_and_run_external_commands () =
     OpamFormatConfig.init ();
     let root_dir = OpamStateConfig.opamroot () in
     let has_init, root_upgraded =
-      match OpamStateConfig.load_defaults root_dir with
+      match OpamStateConfig.load_defaults ~lock_kind:`Lock_read root_dir with
       | None -> (false, false)
       | Some config ->
           let root_upgraded =
--- a/src/client/opamCliMain.ml
+++ b/src/client/opamCliMain.ml
@@ -146,10 +146,8 @@ let check_and_run_external_commands () =
       | Some config ->
         let root_upgraded =
           let cmp =
-            match  OpamFile.Config.opam_root_version config with
-            | Some root_version ->
-              OpamVersion.compare OpamFile.Config.root_version root_version
-            | None -> 1 (* older opam root *)
+            OpamVersion.compare OpamFile.Config.root_version
+              (OpamFile.Config.opam_root_version config)
           in
           if cmp < 0 then
             OpamConsole.error_and_exit `Configuration_error
--- a/src/client/opamCliMain.ml
+++ b/src/client/opamCliMain.ml
@@ -144,19 +144,21 @@ let check_and_run_external_commands () =
       match OpamStateConfig.load_defaults ~lock_kind:`Lock_read root_dir with
       | None -> (false, false)
       | Some config ->
-          let root_upgraded =
-            let config_version = OpamFile.Config.opam_version config in
-            let cmp =
-              OpamVersion.(compare OpamFile.Config.format_version config_version)
-            in
-            if cmp < 0 then
-              OpamConsole.error_and_exit `Configuration_error
-                "%s reports a newer opam version, aborting."
-                 (OpamFilename.Dir.to_string root_dir)
-            else
-              cmp = 0
+        let root_upgraded =
+          let cmp =
+            match  OpamFile.Config.opam_root_version config with
+            | Some root_version ->
+              OpamVersion.compare OpamFile.Config.root_version root_version
+            | None -> 1 (* older opam root *)
           in
-            (true, root_upgraded)
+          if cmp < 0 then
+            OpamConsole.error_and_exit `Configuration_error
+              "%s reports a newer opam version, aborting."
+              (OpamFilename.Dir.to_string root_dir)
+          else
+            cmp = 0
+        in
+        (true, root_upgraded)
     in
     let plugins_bin = OpamPath.plugins_bin root_dir in
     let plugin_symlink_present =
--- a/src/format/opamFile.ml
+++ b/src/format/opamFile.ml
@@ -1179,12 +1179,12 @@ end
 module ConfigSyntax = struct
 
   let internal = "config"
-  (* This version is used as a marker for the whole opam root, so it's not
-     strictly speaking the actual format of the config file *)
   let format_version = OpamVersion.of_string "2.1"
+  let root_version = OpamVersion.of_string "2.1~rc"
 
   type t = {
     opam_version : opam_version;
+    opam_root_version: opam_version option;
     repositories : repository_name list;
     installed_switches : switch list;
     switch : switch option;
@@ -1208,6 +1208,7 @@ module ConfigSyntax = struct
   }
 
   let opam_version t = t.opam_version
+  let opam_root_version t = t.opam_root_version
   let repositories t = t.repositories
   let installed_switches t = t.installed_switches
   let switch t = t.switch
@@ -1237,6 +1238,8 @@ module ConfigSyntax = struct
 
 
   let with_opam_version opam_version t = { t with opam_version }
+  let with_opam_root_version opam_root_version t =
+    { t with opam_root_version = Some opam_root_version }
   let with_repositories repositories t = { t with repositories }
   let with_installed_switches installed_switches t =
     { t with installed_switches }
@@ -1273,6 +1276,7 @@ module ConfigSyntax = struct
 
   let empty = {
     opam_version = format_version;
+    opam_root_version = None;
     repositories = [];
     installed_switches = [];
     switch = None;
@@ -1308,6 +1312,9 @@ module ConfigSyntax = struct
     [
       "opam-version", Pp.ppacc
         with_opam_version opam_version
+        (Pp.V.string -| Pp.of_module "opam-version" (module OpamVersion));
+      "opam-root-version", Pp.ppacc_opt
+        with_opam_root_version opam_root_version
         (Pp.V.string -| Pp.of_module "opam-version" (module OpamVersion));
       "repositories", Pp.ppacc
         with_repositories repositories
@@ -1416,6 +1423,19 @@ end
 module Config = struct
   include ConfigSyntax
   include SyntaxFile(ConfigSyntax)
+
+  let raw_root_version f =
+    let open OpamParserTypes.FullPos in
+    try
+      let opamfile = OpamParser.file (OpamFilename.to_string (filename f)) in
+      Some (OpamStd.List.find_map (function
+          | { pelem = Variable ({ pelem = "opam-root-version"; _},
+                                {pelem = String version; _}); _} ->
+            Some (OpamVersion.of_string version)
+          | _ -> None)
+          opamfile.file_contents)
+    with
+    | Sys_error _ | Not_found -> None
 end
 
 module InitConfigSyntax = struct
--- a/src/format/opamFile.ml
+++ b/src/format/opamFile.ml
@@ -1227,9 +1227,11 @@ module ConfigSyntax = struct
   let file_format_version = OpamVersion.of_string "2.0"
   let root_version = OpamVersion.of_string "2.1~rc"
 
+  let default_old_root_version = OpamVersion.of_string "2.1~~previous"
+
   type t = {
     opam_version : opam_version;
-    opam_root_version: opam_version option;
+    opam_root_version: opam_version;
     repositories : repository_name list;
     installed_switches : switch list;
     switch : switch option;
@@ -1254,6 +1256,9 @@ module ConfigSyntax = struct
 
   let opam_version t = t.opam_version
   let opam_root_version t = t.opam_root_version
+  let opam_root_version_opt t =
+    if OpamVersion.compare t.opam_root_version default_old_root_version = 0 then
+      None else Some t.opam_root_version
   let repositories t = t.repositories
   let installed_switches t = t.installed_switches
   let switch t = t.switch
@@ -1283,8 +1288,7 @@ module ConfigSyntax = struct
 
 
   let with_opam_version opam_version t = { t with opam_version }
-  let with_opam_root_version opam_root_version t =
-    { t with opam_root_version = Some opam_root_version }
+  let with_opam_root_version opam_root_version t = { t with opam_root_version }
   let with_repositories repositories t = { t with repositories }
   let with_installed_switches installed_switches t =
     { t with installed_switches }
@@ -1321,7 +1325,7 @@ module ConfigSyntax = struct
 
   let empty = {
     opam_version = file_format_version;
-    opam_root_version = None;
+    opam_root_version = default_old_root_version;
     repositories = [];
     installed_switches = [];
     switch = None;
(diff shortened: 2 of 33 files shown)


src/format/opamFile.ml
```diff
@@ -1170,12 +1170,12 @@ end
 module ConfigSyntax = struct
 
   let internal = "config"
-  (* This version is used as a marker for the whole opam root, so it's not
-     strictly speaking the actual format of the config file *)
   let format_version = OpamVersion.of_string "2.1"
```
dra27, line 1173: Shouldn't this now be `"2.0"` or do you want to release the field and then change that and opam-state separately?

rjbou:
* [ ] Another file had it format version updated, `switch-export`
* [ ] Add switch invariant inference in the format upgrade, or update the check. For the moment if an invariant is kept empty it is always recomputed.

src/state/opamGlobalState.mli
```diff
@@ -27,6 +27,9 @@ val all_installed: 'a global_state -> package_set
 
 val switches: 'a global_state -> switch list
 
+(* Fold over switches, using switch selections. Switch selection file
+   [switch-state] is loaded only as read-only, no further check given opam root
+   version is done. *)
```
dra27, line 32:
```suggestion
(** Fold over switches, using switch selections. Switch selection file
   [switch-state] is loaded only read-only; no further checks are done on the opam root
   version. *)
```
Do I follow that correctly that it doesn't check `$OPAMROOT/config` for opam-root-version, it just trusts that the switch-state is valid?

src/state/opamStateConfig.ml
```diff
@@ -190,14 +190,109 @@ let opamroot ?root_dir () =
...
+
+module Repos = struct
+  let safe_read ?lock_kind gt =
+    load_if_possible ?lock_kind gt
+      OpamFile.Repos_config.(safe_read, NoError.safe_read)
+      (OpamPath.repos_config gt.root)
+end
 
 let local_switch_exists root switch =
+  (* we don't use safe loading function to avoid several errors display *)
```
dra27, line 287:
```suggestion
  (* we don't use safe loading function to avoid errors displaying *)
```

src/state/opamStateConfig.ml
```diff
@@ -190,14 +190,109 @@ let opamroot ?root_dir () =
...
   OpamPath.Switch.switch_config root switch |>
-  OpamFile.Switch_config.read_opt |> function
+  OpamFile.Switch_config.NoError.read_opt |> function
   | None -> false
   | Some conf -> conf.OpamFile.Switch_config.opam_root = Some root
+(*
+  let conf = safe_load_switch_config_t ~lock_kind:`Lock_read root switch in
+  conf.OpamFile.Switch_config.opam_root = Some root
+*)
```
dra27, line 295: ?

src/state/opamStateConfig.mli
```diff
@@ -109,3 +111,39 @@ val local_switch_exists: OpamFilename.Dir.t -> switch -> bool
 (** Resolves the switch if it is a link to a global switch in the given root
     (return unchanged otherwise) *)
 val resolve_local_switch: OpamFilename.Dir.t -> switch -> switch
+
+(** Given opam root and binary version, the opam root can be loaded only for read-only actions. *)
```
dra27, line 115:
```suggestion
(** Returns [true] if the opam root can only be loaded read-only by the current binary. *)
```
is that correct?

src/state/opamSwitchState.ml
```diff
@@ -392,11 +398,12 @@ let load lock_kind gt rt switch =
...
-        switch_config.OpamFile.Switch_config.opam_version
-        (OpamVersion.of_string "2.1")
-       >= 0
-    || switch_config.OpamFile.Switch_config.invariant <> OpamFormula.Empty
+  (* tweak this check, but with what ? *)
+    if (match OpamFile.Config.opam_root_version gt.config with
+        | None -> false
+        | Some c ->
+          OpamVersion.compare c (OpamVersion.of_string "2.1.0~rc") >= 0)
+    && switch_config.OpamFile.Switch_config.invariant <> OpamFormula.Empty
```
dra27, line 406: This should still be `||`, right - we don't need to recompute the switch invariant if there is one there?

src/state/opamSwitchState.ml
```diff
@@ -392,11 +398,12 @@ let load lock_kind gt rt switch =
     else switch_config
   in
   let switch_config, switch_invariant =
-    if OpamVersion.compare
-        switch_config.OpamFile.Switch_config.opam_version
-        (OpamVersion.of_string "2.1")
-       >= 0
-    || switch_config.OpamFile.Switch_config.invariant <> OpamFormula.Empty
+  (* tweak this check, but with what ? *)
```
dra27, line 401: At the moment, `format_version` in `OpamFile.Switch_config` is no longer used - was the plan still to include that in the switch config?

I don't think we _need_ to (the `opam-root-version` in the root config file is enough), but wouldn't `format_version` be the `opam-config-version` field proposed [in the wiki](https://github.com/ocaml/opam/wiki/opam-root-compatibilities#fields-to-set-up)?

src/state/opamGlobalState.ml
```diff
@@ -57,6 +57,11 @@ let load lock_kind =
       "Opam has not been initialised, please run `opam init'";
   let config_lock = OpamFilename.flock lock_kind (OpamPath.config_lock root) in
   let config = load_config lock_kind global_lock root in
+  if OpamStateConfig.more_recent config && lock_kind <> `Lock_write then
+    log "Opam root version (%s) if greater than binary one (%s), load anyway"
```
dra27, line 61:
```suggestion
    log "root version (%s) is greater than running binary's (%s); load with best-effort (read-only)"
```

src/state/opamGlobalState.ml
```diff
@@ -154,6 +160,9 @@ let drop gt =
   let _ = unlock gt in ()
 
 let with_write_lock ?dontblock gt f =
+  if OpamStateConfig.is_readonly_opamroot gt then
+    OpamConsole.error_and_exit `Configuration_error
+      "Global state try to get write loçck whil opam root is readonly";
```
dra27, line 165:
```suggestion
      "The opam root has been upgraded by a newer version of opam-state and cannot be written to";
```

dra27: I would like to be heading towards a world where only opam-client uses `OpamConsole` - can we turn this into an exception which is caught in the client instead? (thinking of users of opam-state itself)

src/state/opamRepositoryState.ml
```diff
@@ -147,9 +147,12 @@ let get_repo_root rt repo =
 let load lock_kind gt =
   log "LOAD-REPOSITORY-STATE %@ %a" (slog OpamFilename.Dir.to_string) gt.root;
   let lock = OpamFilename.flock lock_kind (OpamPath.repos_lock gt.root) in
-  let repos_map =
-    OpamFile.Repos_config.safe_read (OpamPath.repos_config gt.root)
-  in
+  let repos_map = OpamStateConfig.Repos.safe_read ~lock_kind gt in
+  if OpamStateConfig.is_readonly_opamroot gt then
+    log "Opam root version (%s) if greater than binary one (%s), load anyway"
```
dra27, line 152:
```suggestion
    log "root version (%s) is greater than running binary's (%s); load with best-effort (read-only)"
```

src/state/opamRepositoryState.ml
```diff
@@ -265,6 +268,9 @@ let drop ?cleanup rt =
   let _ = unlock ?cleanup rt in ()
 
 let with_write_lock ?dontblock rt f =
+  if OpamStateConfig.is_readonly_opamroot rt.repos_global then
+    OpamConsole.error_and_exit `Configuration_error
+      "Global state try to get write loçck whil opam root is readonly";
```
dra27, line 273: Same comments as for `OpamGlobalState.with_write_lock`

src/state/opamStateConfig.ml
```diff
@@ -190,14 +190,109 @@ let opamroot ?root_dir () =
...
+    (OpamFile.Config.opam_root_version gt.config)
+
+let is_readonly_opamroot ?lock_kind gt =
+  is_readonly_opamroot_t ?lock_kind gt <> Some false
+
+let load_if_possible_raw ?lock_kind version (read,read_wo_err) f =
+  match is_readonly_opamroot_raw ?lock_kind version with
+  | None ->
+    OpamConsole.error_and_exit `Configuration_error
+      "Write lock and root more recent, abort!"
```
dra27, line 220: As for `OpamGlobalState.with_write_lock` and `OpamRepositoryState.with_write_lock`

src/state/opamSwitchState.ml
```diff
@@ -245,7 +246,12 @@ let load lock_kind gt rt switch =
   let lock =
     OpamFilename.flock lock_kind (OpamPath.Switch.lock gt.root switch)
   in
-  let switch_config = load_switch_config gt switch in
+  let switch_config = load_switch_config ~lock_kind gt switch in
+  if OpamStateConfig.is_readonly_opamroot gt then
+    log "Opam root version (%s) if greater than binary one (%s), load anyway"
```
dra27, line 251:
```suggestion
    log "root version (%s) is greater than running binary's (%s); load with best-effort (read-only)"
```

src/state/opamSwitchState.ml
```diff
@@ -631,6 +631,9 @@ let drop st =
   let _ = unlock st in ()
 
 let with_write_lock ?dontblock st f =
+  if OpamStateConfig.is_readonly_opamroot st.switch_global then
+    OpamConsole.error_and_exit `Configuration_error
+      "Global state try to get write loçck whil opam root is readonly";
```
dra27, line 636: As for all three previous ones 🙂

dra27 (commented):
This is looking great, thanks! Various comments (sometimes on comments) and suggestions.

Do you have a "story" for being able to do this on 2.0 without API changes?

src/client/opamCliMain.ml
```diff
@@ -137,7 +137,8 @@ let check_and_run_external_commands () =
     OpamFormatConfig.init ();
     let root_dir = OpamStateConfig.opamroot () in
     let has_init, root_upgraded =
-      match OpamStateConfig.load_defaults root_dir with
+      (* XXX do we still need this check ? *)
```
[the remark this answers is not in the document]
rjbou: Indeed, as states are loaded with a non lock, better have this check.
Updated with opam root.

src/format/opamFile.ml
```diff
@@ -1170,12 +1194,13 @@ end
...
+  let file_format_version = OpamVersion.of_string "2.0"
+  let root_version = OpamVersion.of_string "2.1.0~rc"
```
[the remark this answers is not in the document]
rjbou: I didn't set it up to current in case we have several updates in a single release., we want to be able to have `2.2.0~alpha.2` for ex

src/state/opamStateConfig.ml
rjbou: some remaining outdated code :)

src/state/opamSwitchState.ml
rjbou: fixed with invariant being an option, if undefined it is clearly at `None`

src/state/opamSwitchState.ml
rjbou: We don't need to have them in the file, we can always retrieve the information with config's `opam-root-version`. But internally it is still useful to have a quick lookup and a better following of the version updates.

src/state/opamGlobalState.ml
rjbou: After discussion, moving all exceptions in opam-client is post poned. For plugins, opam exception can be catched.

src/format/opamFile.ml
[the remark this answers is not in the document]
dra27: That's a very good point! In which case, does it just want to drop the `.0` (so be `2.1~rc`) - we might do `2.1~rc2`, etc. but I think we can commit that we'd _never_ do `2.1.1`?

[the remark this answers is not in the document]
rjbou: it should never happen, yes! The need to have a format change in minor version is in case of a bug fix that need this change, but I see it as another argument to do more frequent releases :D

src/format/opamFile.ml
```diff
@@ -1781,15 +1854,18 @@ module Switch_configSyntax = struct
...
   include Switch_configSyntax
   include SyntaxFile(Switch_configSyntax)
+  module BestEffort = MakeBestEffort(Switch_configSyntax)
 end
 
 module SwitchSelectionsSyntax = struct
 
   let internal = "switch-state"
   let format_version = OpamVersion.of_string "2.0"
+  let file_format_version = OpamVersion.of_string "2.0"
```
dra27, line 1868: Shouldn't this then be used on L1888 below?

src/format/opamFile.ml
```diff
@@ -1040,6 +1040,40 @@ module SyntaxFile(X: SyntaxFileArg) : IO_FILE with type t := X.t = struct
 
 end
 
+(* Error less reading for forward compatibility of opam roots *)
+module type BestEffortArg = sig
+  include SyntaxFileArg
+  val file_format_version: OpamVersion.t
```
dra27, line 1046: This one is unused

src/format/opamFile.ml
rjbou: well spotted!

src/format/opamFile.ml
```diff
@@ -1040,6 +1040,42 @@ module SyntaxFile(X: SyntaxFileArg) : IO_FILE with type t := X.t = struct
 
 end
 
+(* Error less reading for forward compatibility of opam roots *)
+module type BestEffortArg = sig
+  include SyntaxFileArg
+  (* This attribute can be deleted when 4.02 is ditched *)
+  [@@@ocaml.warning "-32"]
+  val file_format_version: OpamVersion.t [@@ocaml.warning "-32"]
+  val pp_cond:
```
AltGr, line 1049: Some doc would be welcome here :sweat_smile:

src/state/opamRepositoryState.ml
```diff
@@ -147,9 +147,13 @@ let get_repo_root rt repo =
 let load lock_kind gt =
   log "LOAD-REPOSITORY-STATE %@ %a" (slog OpamFilename.Dir.to_string) gt.root;
   let lock = OpamFilename.flock lock_kind (OpamPath.repos_lock gt.root) in
-  let repos_map =
-    OpamFile.Repos_config.safe_read (OpamPath.repos_config gt.root)
-  in
+  let repos_map = OpamStateConfig.Repos.safe_read ~lock_kind gt in
+  if OpamStateConfig.is_readonly_opamroot gt then
```
AltGr, line 151: hmm that confused me, maybe `is_readonly_opamroot` could be better named. Stg like `is_newer_than_self` ?

AltGr (approved):

src/format/opamFile.ml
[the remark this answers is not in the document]
rjbou: Updated!

rjbou:
All green, all conversations resolved, some commits to squash and it's good to merge!

PR #4639, opened 2021-04-21
Bump to opam-file-format 2.1.3

### LINKED ISSUE #4394: Enforce `opam-version` as first non-comment line for > 2.0
At present, `opam-version` is used to control the fields available in an `opam` file but it can't control the lexer. In particular, this will be limiting for adding additional operators to the constraint language.

- [x] opam-file-format / opam 2.1 need updating to reject an opam file with `opam-version: "2.1"` or greater where `opam-version` is not the first line, ignoring lines consisting of just whitespace/comments.
- [x] opam 2.0.8 should cease parsing opam files after `opam-version` is encountered (so what a 2.0 lexer would regard as syntax errors are not reported)

dra27:
Needs opam-file-format 2.1.3 to be released first (and `src_ext/Makefile.sources` updated with it).

This then automatically closes #4394. If you put `opam-version: "2.1"` as the second line of an opam file instead of:

```
    error  2: File format error: unsupported or missing file format version; should be 2.0 or older
```

you now get the correct:

```
    error  2: File format error at line 2, column 0: Parse error
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,8 +7,8 @@ env:
   OPAMBSROOT: ~/.cache/.opam.cached
   OPAM12CACHE: ~/.cache/opam1.2/cache
   # This should be identical to the value in appveyor.yml
-  OPAM_TEST_REPO_SHA: 75ff83837547183f000e1c18d42260d7d6cc5fbc
-  OPAM_REPO_SHA: 75ff83837547183f000e1c18d42260d7d6cc5fbc
+  OPAM_TEST_REPO_SHA: b02110b549a0b5275806ce27444fabac71093a0e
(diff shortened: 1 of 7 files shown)


dra27:
Updated to reflect the changes in ocaml/opam-file-format#44 - `OpamFile.SyntaxFile` knows the expected format version of the file it's parsing, and can therefore process the `#` sentinel group returned by an older opam-file-format when parsing an opam file newer than itself.

I tested this by changing `OpamBaseParser.version` to default to opam 1.9 and then added syntax errors in an opam file with `opam-version: "2.0"` and it's working _beautifully_

rjbou:
Just a note: now (in root version PR) we have the exception `Bad_version` in `OpamFormat`, it is raised by `check_opam_version` (the same function that prints the error above)

dra27:
This one's finally ready!

PR #4641, opened 2021-04-21
Delete plugin links on root upgrade

dra27:
This is the last part of #4619, which it turns out is a bit simpler and less dirty than I feared.

This `Dockerfile` demonstrates the problem:

```Dockerfile
WORKDIR /src/test
RUN sudo chown opam /src/test
RUN opam pin add -n dose3 5.0.1-1 ; opam install dune dose3 opam-file-format mccs cmdliner ; opam install opam-dune-lint --deps-only
RUN opam exec -- dune init project foo
RUN echo > dune-project '(lang dune 2.8)\n(generate_opam_files true)(source (github foo/bar))(package (name foo))'
RUN opam exec -- dune build
WORKDIR /src
RUN sudo chown opam /src
#RUN git clone https://github.com/dra27/opam.git -b clear-plugins
#Comment out the line below and uncomment the line above to get the fix
RUN git clone https://github.com/ocaml/opam.git
WORKDIR /src/opam
WORKDIR /src/test
RUN opam install opam-state.2.0.8
RUN opam pin -yn opam-dune-lint 'https://github.com/ocurrent/opam-dune-lint.git#bf451c8cf2d4f2be3c7d979b613c79588b55a172'
RUN opam dune-lint
RUN sudo rm /usr/bin/opam && sudo cp /src/opam/opam /usr/bin/opam
RUN opam update
RUN opam dune-lint
```
and swapping the cloned repo over fixes the problem.

### DIFF
--- a/src/state/opamFormatUpgrade.ml
+++ b/src/state/opamFormatUpgrade.ml
@@ -1098,14 +1098,31 @@ let as_necessary requested_lock global_lock root config =
     |> List.partition (fun (v,_) ->
         OpamVersion.compare v latest_hard_upgrade <= 0)
   in
+  let erase_plugin_links root =
+    let root = OpamPath.plugins_bin root in
+    if OpamFilename.exists_dir root then begin
+      let root = OpamFilename.Dir.to_string root in
+      let files = Sys.readdir root in
+      let erase_link file =
+        (* Sys.file_exists will be false for a dangling symlink *)
+        let file = Filename.concat root file in
+        if not (Sys.file_exists file && Sys.is_directory file) then
+          try Unix.unlink file
+          with Unix.Unix_error _ -> ()
+      in
+      Array.iter erase_link files
+    end
+  in
   let light config =
     let config =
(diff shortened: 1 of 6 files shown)


src/state/opamFormatUpgrade.ml
```diff
@@ -1098,14 +1098,28 @@ let as_necessary requested_lock global_lock root config =
     |> List.partition (fun (v,_) ->
         OpamVersion.compare v latest_hard_upgrade <= 0)
   in
+  let erase_plugin_links root =
+    let root = OpamFilename.Dir.to_string (OpamPath.plugins_bin root) in
+    let files = Sys.readdir root in
+    let erase_link file =
+      (* Sys.file_exists will be false for a dangling symlink *)
```
dra27, line 1105: This surprised me, @altgr, looking at `OpamSystem.files_with_links`, I think it doesn't return dangling symlinks. Is that a bug? If it is, then it can be fixed and used here instead.

AltGr (approved):

PR #4642, opened 2021-04-22
Allow multiple state caches

### LINKED ISSUE #4554: opam-state 2.0.8 and opam 2.0.8 are competing to rebuild the repository cache
Reproduction case:

```
$ opam --version
2.0.8
$ export OPAMDEBUG=1
$ opam install opam-0install opam-state.2.0.8
[..]
$ opam 0install dune
00:00.004  SYSTEM                 mkdir /tmp/opam-xxx-33243
00:00.000  GSTATE                 LOAD-GLOBAL-STATE @ /home/opam/.opam
00:00.000  RSTATE                 LOAD-REPOSITORY-STATE @ /home/opam/.opam
00:00.001  RSTATE                 Bad cache: incompatible magic string "18F028B2" (expected "3C5DCD19").
00:00.002  RSTATE                 Invalid cache, removing
00:00.002  SYSTEM                 rm /home/opam/.opam/repo/state.cache
00:00.006  RSTATE                 No cache found
[...]
$ opam info dune
00:00.001  GSTATE                 LOAD-GLOBAL-STATE @ /home/opam/.opam
00:00.002  RSTATE                 LOAD-REPOSITORY-STATE @ /home/opam/.opam
00:00.004  RSTATE                 Bad cache: incompatible magic string "3C5DCD19" (expected "18F028B2").
00:00.004  RSTATE                 Invalid cache, removing
00:00.004  SYSTEM                 rm /home/opam/.opam/repo/state.cache
00:00.009  RSTATE                 No cache found
[...]
```
This means that every invocation of tools linked with `opam-state` takes a 20s hit on startup to rebuild these cache.

dra27:
At present `$OPAMROOT/repo/state.cache` will be reset if the magic string does not match the version of opam-state / opam. This is tedious when developing and also visible to users as opam-state has a slightly different version from an opam binary.

The fix is quite simple: include the magic string in the filename as well, and tolerate having multiple cache files. `state.cache` is only ever written if something changed, and all that happens now is that `$OPAMROOT/repo/*.cache` is erased prior to writing the new cache (since other opam-state / opam binaries actually must regenerate the file).

As the cache files are all reset at `opam update`, even developers shouldn't see too many of them accumulating.

Fixes #4554. It can't directly be backported to 2.0.9, because 2.0.8 and earlier don't delete *.cache on a write. The "algorithm" could be adapted where opam 2.0.9 always erases its own cache of `state.cache` is newer, but I'm not convinced this is worth it. opam 2.1 does not require this kind of change, because even if opam 2.0.9 can _read_ an opam root, it's never going to write to it, so the cache can't be invalidated.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -111,7 +111,9 @@ New option/command/subcommand are prefixed with ◈.
   *
 
 ## State
-  *
+  * Rename state.cache to include the OpamVersion.magic() string. All .cache files are deleted if any
+    cache file is written to, allowing multiple versions of the library to co-exist without constantly
+    regenerating it [#4642 @dra27 - fix #4554]
 
 # Opam file format
   *
--- a/src/format/opamPath.ml
+++ b/src/format/opamPath.ml
@@ -27,7 +27,7 @@ let init_config_files () =
 
 let state_cache_dir t = t / "repo"
 
-let state_cache t = state_cache_dir t // "state.cache"
+let state_cache t = state_cache_dir t // Printf.sprintf "state-%s.cache" (OpamVersion.magic ())
 
 let lock t = t // "lock"
 
--- a/src/format/opamPath.ml
+++ b/src/format/opamPath.ml
@@ -25,7 +25,9 @@ let init_config_files () =
     OpamFilename.Dir.of_string (OpamStd.Sys.home ()) // ".opamrc";
   ]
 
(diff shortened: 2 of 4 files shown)


dra27:
This is speculatively on the 2.1.0~rc milestone - we might remove it in the morning.

AltGr (approved):

PR #4643, opened 2021-04-22
Build opam with the builtin 0install solver whenever possible

kit-ty-kate:
This effectively enables the `builtin-0install` solver in opam, when:
* `opam-0install-cudf` is installed and the compiler version is recent enough (>= 4.08)
* using the vendored ocaml compiler (e.g. `make cold`)

`--without-0install-solver` can be passed to the configure script to disable it, the same way `mccs` can be disabled.


dra27:
Closed in favour of #4646

PR #4646, opened 2021-04-24
Add a --with-0install-solver option to the configure script to enable the 'builtin-0install' solver

kit-ty-kate:
Same as https://github.com/ocaml/opam/pull/4643 but does not build the 0install solver by default without explicitly passing `--with-0install-solver` (unless it is already installed)

cc @dra27


src_ext/Makefile
```diff
@@ -33,12 +33,16 @@ MD5_PKG_$(1) = $(MD5_$(1))
 endef
 
 SRC_EXTS = cppo extlib re cmdliner ocamlgraph cudf dose3 opam-file-format result seq stdlib-shims
-PKG_EXTS = $(SRC_EXTS) dune-local findlib ocamlbuild topkg mccs
+PKG_EXTS = $(SRC_EXTS) dune-local findlib ocamlbuild topkg mccs opam-0install-cudf 0install-solver
```
dra27, line 36: FWIW, this isn't needed now - it'll pick it up through the SRC_EXTS below

dra27 (commented):
This is what I had in mind, yes! You should be able to do `make CONFIGURE_ARGS=--with-0install-solver cold` - it's up to you if you'd like to add a utility `cold-with-0install` target to do that explictly?

src_ext/Makefile
kit-ty-kate: fixed! TIL Make variable are lazy ^^

kit-ty-kate:
> This is what I had in mind, yes! You should be able to do `make CONFIGURE_ARGS=--with-0install-solver cold` - it's up to you if you'd like to add a utility `cold-with-0install` target to do that explictly?

I think I'll do with `make CONFIGURE_ARGS=--with-0install-solver cold` given a special target can be added separately later and the Makefile already has a `cold-%` target which I'm not sure what to make of.

avsm:
This is great! What's the downside to having it included by default with `make cold`? There's very little downside given that it's pure OCaml code being compiled into the binary.

dra27:
Altering the configuration produced by `make cold` days before we flip to rc seems too hi-octane to me - it's _supposed_ to be disabled by default, yes!

Plan is to enable (building) it by default after 2.1 branches.

kit-ty-kate:
For testing purpose, here is a dockerfile with most of the main different combinations in which you can build opam (comment/uncomment the combination you want to test):
```
FROM ocaml/opam:archlinux-ocaml-4.08
RUN git clone git://github.com/kit-ty-kate/opam.git -b with-0install-solver-option
WORKDIR opam

# does not have builtin-0install
#RUN opam install . && sudo cp $(opam var opam-devel:lib)/opam /usr/bin/opam

# has builtin-0install
#RUN opam install . opam-0install-cudf && sudo cp $(opam var opam-devel:lib)/opam /usr/bin/opam

# does not have builtin-0install
#RUN make cold && sudo cp ./opam /usr/bin/opam

# does not have builtin-0install
#RUN make CONFIGURE_ARGS=--without-0install-solver cold && sudo cp ./opam /usr/bin/opam

# has builtin-0install
RUN make CONFIGURE_ARGS=--with-0install-solver cold && sudo cp ./opam /usr/bin/opam

RUN rm -rf ~/.opam*
RUN sudo pacman -S --noconfirm ocaml
RUN opam init -yn --disable-sandboxing
RUN opam config report
RUN opam option solver=builtin-0install
RUN opam config report
```

avsm:
Enabling it by default in the 2.1 branches after some more testing sounds good to me. Agree it shouldn't affect the trees this close to 2.1.0RC1!

PR #4650, opened 2021-04-28
CI on 2.0 Branch

dra27:
- Constrain < 6 for dose3 (as on master)
- Update point releases to 4.10.2 and 4.11.2 (as on master)
- Add 4.12.0 (as on master!)
- Switch off Travis (you guessed it, as on master!)


dra27:
Travis was actually working (apart from the missing dose3 constraint), but it seems a waste of energy to be running both pipelines.

PR #4655, opened 2021-05-10
Clearer warnings and messages about --cli / OPAMCLI

dra27:
- Swap the two paragraphs in the explanation of CLI versioning - i.e. tell the user _what_ to do first and then _why_ to do it second
- More specific message in the help for `--cli` about `OPAMCLI` - they're not entirely equivalent, so we don't claim it.
- Bossier message when `OPAMCLI` is set to anything other than `2.0`: e.g. "`OPAMCLI` should only ever be set to `2.0` - use '`--cli=2.1`' instead."
- Bossier (and clearer) message when `--cli=2.0` is given: "`--cli=2.0` cannot be understood by opam 2.0; set `OPAMCLI` to `2.0` instead"

cc @avsm

### DIFF
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -1109,9 +1109,9 @@ let global_options cli =
       "Use the command-line interface syntax and semantics of $(docv). \
        Intended for any persistent use of opam (scripts, blog posts, etc.), \
        any version of opam in the same MAJOR series will behave as for the \
-       specified MINOR release. The flag was not available in opam 2.0, so for \
-       2.0, use $(b,\\$OPAMCLI). This is equivalent to setting $(b,\\$OPAMCLI) \
-       to $(i,MAJOR.MINOR)."
+       specified MINOR release. The flag was not available in opam 2.0, so to \
+       select the 2.0 CLI, set the $(b,OPAMCLI) environment variable to \
+       $(i,2.0) instead of using this parameter."
       Arg.string (OpamCLIVersion.to_string OpamCLIVersion.current) in
   let switch =
     mk_opt ~cli cli_original ~section ["switch"]
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -321,16 +321,16 @@ let help_sections cli =
           "Although opam only supports roots ($(i,~%s.opam%s)) for the current \
            version, it does provide backwards compatibility for its \
            command-line interface." dir_sep dir_sep);
-    `P "The command-line version is selected by using the `--cli' option or \
-        the $(i,OPAMCLI) environment variable. `--cli' may be specified more\
-        than once, where the last instance takes precedence. $(i,OPAMCLI) is \
-        only inspected if `--cli' is not given.";
     `P "Since CLI version support was only added in opam 2.1, use $(i,OPAMCLI) \
         to select 2.0 support (as opam 2.0 will just ignore it), \
         and `--cli=2.1' for 2.1 later versions, since an environment variable \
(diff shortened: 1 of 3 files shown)


avsm (approved):
As the source of the confusion, this wording reads well to me ;-)

src/client/opamArg.ml
```diff
@@ -321,16 +321,16 @@ let help_sections cli =
           "Although opam only supports roots ($(i,~%s.opam%s)) for the current \
            version, it does provide backwards compatibility for its \
            command-line interface." dir_sep dir_sep);
-    `P "The command-line version is selected by using the `--cli' option or \
-        the $(i,OPAMCLI) environment variable. `--cli' may be specified more\
-        than once, where the last instance takes precedence. $(i,OPAMCLI) is \
-        only inspected if `--cli' is not given.";
     `P "Since CLI version support was only added in opam 2.1, use $(i,OPAMCLI) \
         to select 2.0 support (as opam 2.0 will just ignore it), \
         and `--cli=2.1' for 2.1 later versions, since an environment variable \
```
avsm, line 326:
```suggestion
        and `--cli=2.1' for 2.1 (or later) versions, since an environment variable \
```

src/client/opamCliMain.ml
```diff
@@ -110,9 +110,11 @@ let check_and_run_external_commands () =
             if OpamCLIVersion.is_supported cli then
               let () =
                 if OpamCLIVersion.(cli >= (2, 1)) then
+                  let flag = "--cli=" ^ OpamCLIVersion.(to_string cli) in
                   OpamConsole.warning
-                    "Setting OPAMCLI is brittle - consider using the \
-                     '--cli <major>.<minor>' flag."
+                    "OPAMCLI should only ever be set to %s - use '%s' instead."
```
avsm, line 115: Is this really a necessary warning? In Dockerfiles for example, it would be quite hard to weave through a choice of "--cli" or "ENV OPAMCLI=2.0" without programmatic generation. Just setting OPAMCLI=2.1 is both reliable and easy in that environment -- what's brittle about environment variables if you control the environment?

dra27: The main reason was hygiene - you should control the _CLI_ version using the _CLI_! You never truly control the environment, or you at least need to be paranoid at _every_ call site to the opam binary that `OPAMCLI` is correct - the aim is to nudge all scripts (and all programs) to specify `--cli` with the command being executed (like a type annotation!). The secondary reason was to strongly discourage exporting `OPAMCLI` in `.bashrc`.

src/client/opamArg.ml
dra27: Ta!

src/client/opamCliMain.ml
dra27: For scripts, the expectation was that you'd do something like `opam='opam --cli=2.1' at the top of your script (or equivalent in better languages!) (cf. https://github.com/ocaml/opam/wiki/Spec-for-opam-CLI-versioning#summary)

avsm: Got it, thanks. I'd thought that was a hard error, but setting OPAMCLI=2.0 is actually only a warning as well. That does make it easier to transition via env variables "in case of scripting emergency"

dra27: Definitely - the only hard-error you should get is explicitly requesting 2.0 and trying to use a 2.1 feature or the sort of vice-versa (requesting 2.1 and trying to use something from 2.0 which has been removed/renamed from the CLI)

rjbou (approved):
Thanks!

PR #4658, opened 2021-05-11
Speedup installation when the package has no conflict

kit-ty-kate:
e.g. when calling `opam install utop` on a 4.12 switch, opam spends 4s in Dose3 before calling the actual solver

Partly fixes an issue detected by @talex5 in https://github.com/ocurrent/docker-base-images/pull/102#issuecomment-837221525

### DIFF
--- a/src/solver/opamCudf.ml
+++ b/src/solver/opamCudf.ml
@@ -1413,11 +1413,7 @@ let actions_of_diff (install, remove) =
 
 let resolve ~extern ~version_map universe request =
   log "resolve request=%a" (slog string_of_request) request;
-  let resp =
-    match check_request ~version_map universe request with
-    | Success _ when extern -> get_final_universe ~version_map universe request
-    | resp -> resp
-  in
+  let resp = get_final_universe ~version_map universe request in
   let cleanup univ =
     Cudf.remove_package univ opam_invariant_package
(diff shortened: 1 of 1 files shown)


AltGr:
This change would prevent all conflict messages from being printed, replacing them by "solver error": it's not a good idea as is.

The other option would be to revert to what we had in 2.0: first call the external solver, then call `check_request` (the dose3 SAT) to get the conflict information.

The change was made for a reason though, so we need to examine the balance here. My impression is that the order in 2.1 is better for the user in most cases, but here again, I hadn't hit such cases where `check_request` had a significant cost as of now.
- with the 2.0 order (external then `check_request`):
  * in case of success, you only get the (generally long, in comparison) external solver time without additional penalty
  * in case of conflict, you get a (generally quite long) external solving time before it realises there is no solution, before we run `check_request` to (quickly) get the conflicts
- with the current 2.1 order (`check_request` first):
  * in case of success, you _always_ get the penalty of the added `check_request` call. In general it is negligible compared to the solving time, but that may be different in problematic cases and with the 0install solver, as demonstrated in thereferenced issue.
  * in case of conflict, you detect it right away and print the details without calling the external solver at all

In a perfect world we would call both in parallel :)

PR #4663, opened 2021-05-14
Scrub OPAM* for variables added since 2.0

dra27:
Packages build with OPAMCLI=2.0. If the user sets an opam 2.1-specific environment variable, this is passed through to the package and will cause a warning to be displayed in the build log if the package calls opam.

All new environment variables are now scrubbed from the build environment.


rjbou (approved):

dra27:
I messed up the rebase on 21 May which the #4672 has now caught, but only the merge! Fixed by direct push in 5db90e1.

PR #4666, opened 2021-05-17
Build release tags as release tarballs [2.0 version]

dra27:
As for #4665, but without the new API in `OpamVersion` and with a more devilish PR number 😈

### DIFF
--- a/src/client/dune
+++ b/src/client/dune
@@ -2,7 +2,7 @@
   (name        opam_client)
   (public_name opam-client)
   (synopsis    "OCaml Package Manager client and CLI library")
-  (modules     (:standard \ opamMain opamManifest get-git-version))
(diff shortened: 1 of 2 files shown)


src/client/dune
```diff
@@ -2,7 +2,7 @@
   (name        opam_client)
   (public_name opam-client)
   (synopsis    "OCaml Package Manager client and CLI library")
-  (modules     (:standard \ opamMain opamManifest get-git-version))
+  (modules     (:standard \ opamMain opamManifest get_git_version))
```
dra27, line 5: As this script was being altered, I pulled in the change from https://github.com/ocaml/opam/pull/4096.

rjbou (approved):

Issue #4657, opened 2021-05-11
[master] opam-devel does not install the opam binary anymore


kit-ty-kate:
Using master (https://github.com/ocaml/opam/commit/a70ed4f98a9f8bc462908dd12298281f8ea93c7d) installing opam-devel:
```
$ opam install .
[...]
∗ installed opam-devel.2.1.0~beta4
Done.

<><> opam-devel.2.1.0~beta4 installed successfully ><><><><><><><><><><><><><><>
=> The development version of opam has been successfully compiled into /home/kit_ty_kate/.opam/default/lib/opam-devel.
   You should not run it from there, please install the binaries to your PATH, e.g. with
       sudo cp /home/kit_ty_kate/.opam/default/lib/opam-devel/opam /usr/local/bin

   If you just want to give it a try without altering your current installation, you could use instead:
       alias opam2="OPAMROOT=~/.opam2 /home/kit_ty_kate/.opam/default/lib/opam-devel/opam"
```
However `/home/kit_ty_kate/.opam/default/lib/opam-devel/opam` is an (executable?!) opam file and not a binary:
```
$ ls -lh /home/kit_ty_kate/.opam/default/lib/opam-devel/opam
-rwxr-xr-x 1 kit_ty_kate kit_ty_kate 741 May 11 13:51 /home/kit_ty_kate/.opam/default/lib/opam-devel/opam
$ cat /home/kit_ty_kate/.opam/default/lib/opam-devel/opam
opam-version: "2.0"
version: "2.1.0~beta4"
synopsis: "Meta-package for Dune"
[...]
```
Its content is the same as `opam.opam` at a70ed4f98a9f8bc462908dd12298281f8ea93c7d

dra27:
This is an inconsistency of opam-devel.install (which has been present for a long time, probably since the conversion to jbuilder) combined with a regression caused by #4494.

I'll sort out a fix to the opam-devel.install part - would you be able to look into the regression from #4494? The issue is that there an `opam` file mentioned in both the `lib` and `libexec` sections. opam should install the file twice which is what it was before #4494. I'm wondering if a list of files is being processed in the wrong order?

There's possibly an additional bug in the case of overwriting to check - the execute bit should have been cleared on overwriting it with the opam file.

PR #4667, opened 2021-05-17
Fix installation order when processing .install files

dra27:
#4657 exposed no fewer than 3 bugs! 🐛

- `OpamSystem.install` doesn't the file permissions correctly if the file already exists (since `open_out_gen` only uses the mode if it actually creates the file). The first commit here fixes that and also the case of a symlink being overwritten which is handled by `OpamSystem.copy_file` but wasn't being handled by `OpamSystem.install`.
- `opam-devel.install` installs a file called `opam` twice to `lib/opam-devel` (fixed in #4664)
- The updates in #4494 changed the order in which sections were processed - the second commit here reverses (excuse pun!) that slip

NB `List.flatten` blows the stack w.r.t. the number of lists (which is fundamentally small) and the length of the longest sub-list - this would already blow the stack in an earlier `List.map` call, so I convinced myself that a crazy opam .install file with gazillions of entries is already exploding.

### DIFF
--- a/src/client/opamAction.ml
+++ b/src/client/opamAction.ml
@@ -82,8 +82,8 @@ let preprocess_dot_install_t st nv build_dir =
       in
       [dir, inst]
     in
-    dir_and_install @
-    List.map (fun (base, dst) ->
+    dir_and_install @ List.rev @@
+    List.rev_map (fun (base, dst) ->
         let (base, append) =
           if exec &&
              not (OpamFilename.exists (OpamFilename.create build_dir base.c))
@@ -158,7 +158,10 @@ let preprocess_dot_install_t st nv build_dir =
   ]
   in
 
-  let to_install = List.map install_files to_install in
+  let files_and_installs =
+    List.fold_left (fun acc toi -> List.rev_append (install_files toi) acc)
+      files_and_installs to_install
+  in
 
   (* misc *)
   let misc_files =
@@ -179,7 +182,8 @@ let preprocess_dot_install_t st nv build_dir =
         in
         (file, inst)) (I.misc install)
   in
-  List.flatten ((files_and_installs :: to_install) @ [misc_files])
+
+  List.rev_append files_and_installs misc_files
 
 (* Returns function to install package files from [.install] *)
 let preprocess_dot_install st nv build_dir =
--- a/src/client/opamAction.ml
+++ b/src/client/opamAction.ml
@@ -27,8 +27,6 @@ let preprocess_dot_install_t st nv build_dir =
   let switch_prefix = OpamPath.Switch.root root st.switch in
   let file_wo_prefix f = OpamFilename.remove_prefix switch_prefix f in
 
-  let files_and_installs = [] in
-
   let name = nv.name in
   let install_f = OpamPath.Builddir.install build_dir nv in
   let install = OpamFile.Dot_install.safe_read install_f in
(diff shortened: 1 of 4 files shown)


src/core/opamSystem.ml
```diff
@@ -177,18 +177,29 @@ let write file contents =
...
-      let perm =
-        (Unix.fstat (Unix.descr_of_in_channel ic)).st_perm |> chmod
-      in
+  try
+    let perm =
+      (Unix.fstat (Unix.descr_of_in_channel ic)).st_perm |> chmod
+    in
+    let () =
+      try if Unix.((lstat dst).st_kind <> S_REG) then
+            remove_file dst
```
AltGr, line 186: This is the right behaviour, but now I am wondering what happens if a package overwrites a directory installed by another package with a file :thinking: 
It shouldn't be worse than just overriding files, and since this `DirTrack` is done outside of this, it should correctly detect the deleted files; but this has consequences with #4494, since that only handles lists of files expected to be installed: it should also track anything *below* in case it encounters a directory right ?

That seems pretty minor, though.

AltGr (approved):
Seems good! See my comment on interrogations about special cases for #4494 though.

kit-ty-kate (commented):
My original PR in https://github.com/ocaml/opam/pull/4494 was significantly changed so none of this is my code (this also explains why i didn't see these issues sooner) and I can't really review it properly, but from afar it looks ok.

src/core/opamSystem.ml
dra27: I'm not sure what happens at the higher level, but `OpamSystem.install` will fail if the destination is a directory?

src/client/opamAction.ml
```diff
@@ -84,8 +82,8 @@ let preprocess_dot_install_t st nv build_dir =
       in
       [dir, inst]
     in
-    dir_and_install @
-    List.map (fun (base, dst) ->
+    dir_and_install @ List.rev @@
+    List.rev_map (fun (base, dst) ->
```
rjbou, line 86:
```suggestion
    dir_and_install @
    List.map (fun (base, dst) ->
```
I went into *let's tail rec everything* mode, but maybe this is too much optimisation? It handles the case where there is a huge amount of files in a given field field in the `.install` file

src/core/opamSystem.ml
rjbou: Several kind of files are written in the install process:
* the `pkg.config` file that is package specific by name
* the files in `.install` file that are written in `prefix/pkg/xxx`
* the misc files in `.install` that need confirmation from user to be installed, as they are absolute paths
Other install/copy are done via `install` field commands not opam instructions

src/client/opamAction.ml
dra27: I think it's a good idea, actually - thank you for doing it!

src/core/opamSystem.ml
rjbou: > I'm not sure what happens at the higher level, but OpamSystem.install will fail if the destination is a directory?

indeed, it is the first check
```
let install ?(warning=default_install_warning) ?exec src dst =
  if Sys.is_directory src then
    internal_error "Cannot install %s: it is a directory." src;
  if (try Sys.is_directory dst with Sys_error _ -> false) then
    internal_error "Cannot install to %s: it is a directory." dst;
```

PR #4672, opened 2021-05-19
Harsher CI

dra27:
The builds of opam in CI should be done in dev mode.

My initial implementation simply used `DUNE_PROFILE` in the environment, but that's spawned https://github.com/ocaml/dune/issues/4632, so I've instead added it to `Makefile.config` after `configure` is run. The two places where `DUNE_PROFILE` is used in the build system both set it with `?=` (to allow an environment override) and also both read `Makefile.config`, so this will work. I also dealt with a TODO comment from #4545, in order to prevent any future accidental breakage of this.


.github/scripts/main.sh
```diff
@@ -20,6 +20,7 @@ export OCAMLRUNPARAM=b
   fi
 
   ./configure --prefix ~/local --with-mccs
+  echo 'DUNE_PROFILE=dev' >> Makefile.config
```
dra27, line 23: One possibility might to do this _only_ when `"$OPAM_TEST" != "1"` - i.e. test the build with the dev profile but test the release profile version with opam-rt. @rjbou?

dra27: Discussed this morning, and yes we'll do this!

PR #4676, opened 2021-05-21
Add support for non-interactive mode in macports

kit-ty-kate:
The default behaviour as I understand it is that if only one package is being installed it will not prompt for anything (same behaviour as apt) but will prompt if dependencies are required


dra27:
Eventually grepping the code, I managed to find some reasonable docs on it! The option appears to have been added in 2017 as part of macports 2.4 and it did look from a cursory glance as though it's semantics for `port install` is to answer yes for everything.

Do we need to worry about older versions of port than 2.4? @pmetzger @mseri?

mseri:
I am always on the latest version. Imho anything working with 2.6 and 2.7 would be perfect

pmetzger:
> Do we need to worry about older versions of port than 2.4?

No. MacPorts doesn't support old old versions.

PR #4678, opened 2021-05-24
Improve depext unavailable messages from the solver

dra27:
Well, hopefully improve. At the moment, if you install a conf package directly, you get the depext availability message from _before_ solving:

```
dra@thor:~/opam$ opam install conf-capnproto
[ERROR] Package conf-capnproto depends on the unavailable system package
        'capnproto libcapnp-dev'. You can use `--no-depexts' to attempt
        installation anyway.
```
although the message is not properly formatted. If, as is more likely, the conf package is pulled in via a dependency, the message comes _after_ solving and is a little more intimidating:

```
dra@thor:~/opam$ opam install capnp-rpc-unix
[ERROR] Package conflict!
  * Missing dependency:
    - capnp-rpc-unix → capnp-rpc-lwt < 0.2 → conf-capnproto
    depends on the unavailable system package 'capnproto libcapnp-dev'. Use
      `--assume-depexts' to attempt installation anyway.
  * Missing dependency:
    - capnp-rpc-unix → capnp-rpc-lwt >= 0.2 → conf-capnproto
    depends on the unavailable system package 'capnproto libcapnp-dev'. Use
      `--assume-depexts' to attempt installation anyway.
  * Missing dependency:
    - capnp-rpc-unix → capnp-rpc-net < 0.6.0 → conf-capnproto
    depends on the unavailable system package 'capnproto libcapnp-dev'. Use
      `--assume-depexts' to attempt installation anyway.
  * Missing dependency:
    - capnp-rpc-unix → capnp-rpc-net = 0.6.0 → conf-capnproto
    depends on the unavailable system package 'capnproto libcapnp-dev'. Use
      `--assume-depexts' to attempt installation anyway.
  * Missing dependency:
    - capnp-rpc-unix → capnp-rpc-net = 0.7.0 → conf-capnproto
    depends on the unavailable system package 'capnproto libcapnp-dev'. Use
      `--assume-depexts' to attempt installation anyway.
  * Missing dependency:
    - capnp-rpc-unix → capnp-rpc-net = 0.8.0 → conf-capnproto
    depends on the unavailable system package 'capnproto libcapnp-dev'. Use
      `--assume-depexts' to attempt installation anyway.
  * Missing dependency:
    - capnp-rpc-unix → capnp-rpc-net = 0.9.0 → conf-capnproto
    depends on the unavailable system package 'capnproto libcapnp-dev'. Use
      `--assume-depexts' to attempt installation anyway.
  * Missing dependency:
[...]

Issue #4669, opened 2021-05-18
Depexts detection does not work on CentOS when EPEL is involved


kit-ty-kate:
Detected in https://github.com/ocaml/opam-repository/pull/18704

The way things work in CentOS wrt opam is that when EPEL is needed, we install it first (through `epel-release`) then we install the package required. However the new depexts detection in opam does not find the package since it is not yet available (as it is in the EPEL repository) and thus mark it as unavailable.

I'm not entirely sure how to deal with that here..

dra27:
Hmm. Depending on epel-release as a depext may work for CentOS, but that already doesn't work for RHEL (you're supposed to install the epel-release rpm explicitly to enable it, IIRC).

This can't be fixed in depext detection (at least quickly) - it's fundamentally about _configuring_ the package manager which is not what this is for. Essentially the OS package manager has not been "correctly" configured (it'd be like having the wrong apt repo configuration)

Practical thoughts, though:
- Change the default invariant to pull in `"epel-release"` on CentOS - i.e. effectively have `opam init` request to set that up - so we "kick" users to get EPEL enabled before it's definitely needed. However, I'm not convinced we _should_ do this - there's a reason EPEL exists, and doing this semi-automatically rather violates it.
- See what needs doing to get opam itself into EPEL. According to https://src.fedoraproject.org/rpms/opam, the opam package is orphaned. If the main way to install opam is via EPEL then most users are going to get it by default!
- Improve the error handling (possibly for Alpine's benefit too?) to make better suggestions when the packages can't be found - perhaps adding a concept of enabled repos to depext? It feels like a rabbit-hole in something that's already quite complicated, though!

dra27:
Referencing all the relevant parts of this. This part, pulled in from https://github.com/ocaml-opam/opam-depext/pull/77, is not working (the special-casing needs to be lifted, executed - i.e. `epel-release` installed first, and then followed by a fresh query for the available package list for it to work):
https://github.com/ocaml/opam/blob/3cd0829538b90191bcd1c4eaa76c00d3d987b4ea/src/state/opamSysInteract.ml#L531-L547
There's also the original issues in https://github.com/ocaml-opam/opam-depext/issues/70 and https://github.com/ocaml-opam/opam-depext/issues/76.

I'm rather on @avsm's side in https://github.com/ocaml-opam/opam-depext/pull/77#issuecomment-326548524 - I'm not convinced opam-depext should ever have been doing this.

PR #4679, opened 2021-05-24
Display a message on CentOS if EPEL isn't enabled and depexts aren't available

dra27:
Provides the workaround for #4669 discussed last week. This PR temporarily sits on #4678, as it uses some of the refactorings - only https://github.com/ocaml/opam/commit/34d5cafba78d626e565bfe72cb8e1ab4ea6c4188 is new.

With this patch, on CentOS without EPEL enabled you now see:

```
dra@thor:~/opam$ opam install conf-capnproto
[ERROR] Package conf-capnproto depends on the unavailable system package
        'capnproto'. You can use `--no-depexts' to attempt installation anyway.
        On CentOS/RHEL, many packages may assume that the Extra Packages for
        Enterprise Linux (EPEL) repository has been enabled. This is typically
        done by installing the 'epel-release' package. Please see
        https://fedoraproject.org/wiki/EPEL for more information.
```

Most of the change is refactoring the existing API to separate the computation of what's happened from formatting the message (so that `OpamClient` can semantically what happened).

I've only applied this to `OpamClient.install_t` and `OpamSolution.check_availability` - I'm not sure if some of the other functions in `OpamClient` want this or not.

### DIFF
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -689,3 +689,17 @@ let update () =
   | Some (cmd, args) ->
     try sudo_run_command cmd args
     with Failure msg -> failwith ("System package update " ^ msg)
+
+let repo_enablers () =
+  if family () <> Centos then None else
+  let (needed, _) =
+    packages_status (OpamSysPkg.raw_set
+                       (OpamStd.String.Set.singleton "epel-release"))
+  in
+  if OpamSysPkg.Set.is_empty needed then None
+  else
+    Some
+      "On CentOS/RHEL, many packages may assume that the Extra Packages for \
+       Enterprise Linux (EPEL) repository has been enabled. \
+       This is typically done by installing the 'epel-release' package. \
+       Please see https://fedoraproject.org/wiki/EPEL for more information"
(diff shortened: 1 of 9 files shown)


src/state/opamSysInteract.ml
```diff
@@ -621,3 +621,18 @@ let update () =
...
+    let (needed, _) =
+      packages_status (OpamSysPkg.raw_set (OpamStd.String.Set.singleton "epel-release"))
+    in
+    if OpamSysPkg.Set.is_empty needed then
+      None
+    else
+      Some "On CentOS/RHEL, many packages may assume that the Extra Packages \
+            for Enterprise Linux (EPEL) repository has been enabled. This is \
+            typically done by installing the 'epel-release' package. Please \
+            see https://fedoraproject.org/wiki/EPEL for more information"
```
avsm, line 636: We're going to need a number of these workarounds for different operating systems (e.g. Alpine's edgecommunity, Brew's custom taps and so on).  Would it be feasible to move this into a package `{message}` entry somehow rather than hardcoding Fedora-specific logic in the client?  Obviously, other opam-repositories other than OCaml's may not make this assumption as well.

dra27: The logic here is being embedded in the section of the opam client which deals with the logic for specific package managers, so I think it is in the right place for now?

It's not feasible between now and releasing 2.1, no. At the moment, for EPEL specifically, this is fixing a regression in behaviour because of a poor decision in 2017 to handle this in opam-depext! All this PR is doing (the actual new code is only a tiny bit on an error message) is displaying this message if it thinks it _might_ be helpful. We should definitely follow this up with better (and generic) tagging, or some such, of depexts, though.

My understanding is that the Alpine repos work much better because the error message from apk clearly explains what needs doing (i.e. this PR brings dnf up to apk's standard 🙂). ppas, custom taps, etc. would be a _completely_ new feature - EPEL is an official repository system (as, again, if I understand correctly, edgecommunity is an official repo, it's just not necessarily enabled by default?).

dra27:
(rebased to pick up a fix in #4678)

PR #4683, opened 2021-05-27
OPAMCONFIRMLEVEL overrides --yes and --no, etc.

dra27:
Fixes #4582. It's possible that this can be done in a different order - I've had the global options record where the answer came from (to allow the environment to override it)


kit-ty-kate:
I think `--yes --confirm-level=unsafe-yes` (via command line options) should also result in `unsafe-yes`. Does this patch also do that?

Could we also add tests for the combination of both? It's something that is quite wildly used in CI environment

dra27:
@kit-ty-kate - that's already the case before this PR. The environment variables work together as you expect and the CLI flags work together as you expect - just not both at once!

rjbou (approved):

PR #4696, opened 2021-06-02
Link archives in cache when several checksum are given

rjbou:
Add link for other checksum on downloading, if possible. `OpamSystem.link` fallback is to copy, so for some systems, it can lead to multiples copies of a single archive/file.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -60,6 +60,9 @@ users)
   * W68: add warning for missing license field [#4766 @kit-ty-kate - partial fix #4598]
   * W62: use the spdx_licenses library to check for valid licenses. This allows to use compound expressions such as "MIT AND (GPL-2.0-only OR LGPL-2.0-only)", as well as user defined licenses e.g. "LicenseRef-my-custom-license" [#4768 @kit-ty-kate - fixes #4598]
 
+## Repository
+  * When several checksums are specified, instead of adding in the cache only the archive by first checksum, name by best one and link others to this archive [#4696 rjbou]
+
 ## Lock
   *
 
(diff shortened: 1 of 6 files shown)


master_changes.md
```diff
@@ -82,6 +82,9 @@ New option/command/subcommand are prefixed with ◈.
   * Add E67 checksum specified with non archive url [#4635 @rjbou]
   * Disable subpath warning E63,W64 [#4638 @rjbou]
 
+## Repository
+  * When several checksume are specified, instead of creating add in the cache only the first checksum, add best one and link others [#XXX rjbou]
```
AltGr, line 86: You should specify that this concerns `opam admin`

rjbou: It is for admin and for regular cache also (called from OpamRepository.pull_from_upstream)

kit-ty-kate (approved):
Rebased and added a new `OpamHash.sort_best_to_weakest_hash_strength` function to help making the code a bit clearer. Feel free to remove or rename the function if you think of a better one.

LGTM

rjbou:
Thanks!

src/core/opamHash.mli
```diff
@@ -32,6 +32,9 @@ val of_string_opt: string -> t option
     "md5/d4/d41d8cd98f00b204e9800998ecf8427e", as a list *)
 val to_path: t -> string list
 
+(** Sorts the list from best hash (SHA512, ...) to weakest (..., MD5) *)
+val sort : t list -> t list
```
kit-ty-kate, line 36: It feels a bit surprising to have a reverse sort being called sort tbh but I guess that’s ok in opam lib. We’re not a generalist library after all.

rjbou: That's the idea :) Like @AltGr said, its a sort from the less weak to the most weak :D

kit-ty-kate: adjugé vendu! :D

kit-ty-kate:
Thanks!

PR #4699, opened 2021-06-03
depext: ignore `--yes` on asking to run install commands

### LINKED ISSUE #4680: Integrated depext needs to treat sudo operations as unsafe
At present, unsafe confirmation is only used to suppress package manager prompts, however, the spec for integrating depext requires default confirmation of package manager operations (i.e. confirmation prior to a sudo call).

Discussed on [21.05.2021](https://github.com/ocaml/opam/wiki/2021-Developer-Meetings#21-may-2021) and should be easy to change once #4582 is merged, just tracking it so we don't forget!

rjbou:
Fix #4680


dra27 (approved):
LGTM!

PR #4706, opened 2021-06-10
Fix sys-ocaml-* variables with no system compiler

dra27:
`/usr/bin/sh: ocamlc: command not found` returned for `sys-ocaml-arch`, `sys-ocaml-cc` and `sys-ocaml-libc` if there's no system compiler


dra27:
Drive-by note in this that three variables are not included in the format-upgrade. I wonder if that might be a good idea?

rjbou:
Hard format upgrade does a reinit (and then it will be included), and for light format upgrade, we encourage people to do a reinit for each new binary/version used (to update sandbox scripts for example).

dra27:
It was only because I happened to notice the special-case for `sys-ocaml-version`

rjbou:
Indeed, the reinit don't update `eval-variables` field as it is already defined (it takes init default ones onlys if not defined).

Issue #4705, opened 2021-06-09
Space character in parent directory breaks many commands


ryanartecona:
I get an error like the following on a number of opam subcommands (switch create, install, upgrade).
```
sh: -c: line 0: syntax error near unexpected token `('
sh: -c: line 0: `rm -rf /Users/ryanartecona/Dropbox (Personal)/Code/Web/ryanartecona.github.io/example/_opam/.opam-switch/build/ocaml-system.4.12.0'
Switch initialisation failed: clean up? ('n' will leave the
switch partially installed) [Y/n] n
Cannot remove /Users/ryanartecona/Dropbox (Personal)/Code/Web/ryanartecona.github.io/example/_opam/.opam-switch/build/ocaml-system.4.12.0 (error 2).
```

This seems to be caused by `OpamSystem.remove_dir` running an `rm -rf dir` command without quoting the dir. With a space anywhere in the path, the shell splits the argument into 2 (and in my case chokes on a `(` that follows the space).

---

```
> opam config report
# opam config report
# opam-version      2.0.8 
# self-upgrade      no
# system            arch=x86_64 os=macos os-distribution=homebrew os-version=10.15.7
# solver            builtin-mccs+glpk
# install-criteria  -removed,-count[version-lag,request],-count[version-lag,changed],-changed
# upgrade-criteria  -removed,-count[version-lag,solution],-new
# jobs              4
# repositories      1 (http) (default repo at c8553cab)
# pinned            0
# current-switch    /Users/ryanartecona/Dropbox (Personal)/Code/Web/ryanartecona.github.io/example```

dra27:
As @kit-ty-kate notes in the PR, this has been present for a loooong time, but having spotted it it will be dealt with for 2.0.9 and 2.1.0, thanks!

PR #4707, opened 2021-06-10
Add missing shell quoting to support space and special shell characters in switch directory path

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/4705
Introduced in opam 1.1.0~rc https://github.com/ocaml/opam/commit/fb108835e5182ee17620a724aa3f552822e04052

### DIFF
--- a/src/core/opamSystem.ml
+++ b/src/core/opamSystem.ml
@@ -90,7 +90,7 @@ let rm_command =
 let remove_dir dir =
   log "rmdir %s" dir;
   if Sys.file_exists dir then (
-    let err = Sys.command (Printf.sprintf "%s %s" rm_command dir) in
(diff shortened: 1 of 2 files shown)


src/core/opamSystem.ml
```diff
@@ -88,9 +88,9 @@ let rm_command =
     "rm -rf"
 
 let remove_dir dir =
-  log "rmdir %s" dir;
+  log "rmdir %s" (Filename.quote dir);
```
dra27, line 91: FWIW, this one's not necessary - it's just a logging line

kit-ty-kate: yeah but i figured it's better to show what's really going to be done

PR #4715, opened 2021-06-16
Format upgrade fix: remove missing switches from config file

### LINKED ISSUE #4713: The layout update should not fail even if the installed-switches are incorrect
If you run `rm -rf _opam` insetad of `opam switch remove .` to delete a local switch, opam cannot update the config properly. If you try to update the layout under these conditions, you will get an error like the following, but it should not occur or it should be more friendly.

```
~
❯ opam update --upgrade
This version of opam requires an update to the layout of /home/smorimoto/.opam from version 2.1~alpha to version
2.1~rc, which can't be reverted.
Perform the update and continue? [Y/n] y
Fatal error:
Sys_error("/home/smorimoto/src/github.com/REDACTED/REDACTED/_opam/.opam-switch/switch-config: No such file or directory")
```

rjbou:
* Fix format upgrade when there is missing local switches in the config file
* Fix not recorded local switch handling, with format upgrade
* Set opam root version to 2.1
* Update tests with 2.1, local switch (recorded and not)
fix #4713

### DIFF
--- a/src/state/opamFormatUpgrade.ml
+++ b/src/state/opamFormatUpgrade.ml
@@ -1133,16 +1133,6 @@ let as_necessary requested_lock global_lock root config =
   let cmp = OpamVersion.(compare OpamFile.Config.root_version root_version) in
   if cmp <= 0 then config (* newer or same *) else
   let is_intermdiate_root = List.mem root_version intermediate_roots in
-  let need_hard_upg =
-    OpamVersion.compare root_version latest_hard_upgrade < 0
-    || is_intermdiate_root
-  in
-  let on_the_fly, global_lock_kind =
-    if not need_hard_upg && requested_lock <> `Lock_write then
-      true, `Lock_read
-    else
-      false, `Lock_write
-  in
   (* to generalise *)
   let intermediates = [
     v2_1_alpha,  from_2_0_to_2_1_alpha;
@@ -1169,6 +1159,13 @@ let as_necessary requested_lock global_lock root config =
       |> List.filter (fun (v,_) -> OpamVersion.compare root_version v < 0)
       |> List.partition (fun (v,_) ->
           OpamVersion.compare v latest_hard_upgrade <= 0)
+  in
+  let need_hard_upg = hard_upg <> [] in
+  let on_the_fly, global_lock_kind =
+    if not need_hard_upg && requested_lock <> `Lock_write then
+      true, `Lock_read
+    else
+      false, `Lock_write
   in
   let erase_plugin_links root =
     let plugins_bin = OpamPath.plugins_bin root in
--- a/src/state/opamFormatUpgrade.ml
+++ b/src/state/opamFormatUpgrade.ml
@@ -1211,6 +1211,7 @@ let as_necessary requested_lock global_lock root config =
             | _ -> "they no longer exist");
          config
   in
+  if hard_upg = [] && light_upg = [] then config (* no upgrade to do *) else
   let is_dev = OpamVersion.is_dev_version () in
   log "%s config upgrade, from %s to %s"
     (if on_the_fly then "On-the-fly" else
--- a/src/state/opamFormatUpgrade.ml
+++ b/src/state/opamFormatUpgrade.ml
@@ -1054,7 +1054,7 @@ let v2_1_alpha2 = OpamVersion.of_string "2.1~alpha2"
 (* config & sw config downgrade opam-version to 2.0 and add opam root version *)
 let v2_1_rc = OpamVersion.of_string "2.1~rc"
 
-let _v2_1 = OpamVersion.of_string "2.1"
+let v2_1 = OpamVersion.of_string "2.1"
 
 let from_2_0_to_2_1_alpha _ conf = conf
 
@@ -1094,7 +1094,9 @@ let from_2_1_alpha2_to_v2_1_rc root conf =
   downgrade_2_1_switches root conf;
   conf
 
-let from_2_0_to_v2_1_rc _ conf = conf
+let from_2_1_rc_to_v2_1 _ conf = conf
+
+let from_2_0_to_v2_1 _ conf = conf
 
 let latest_version = OpamFile.Config.root_version
 
@@ -1102,7 +1104,7 @@ let latest_hard_upgrade = (* to *) v2_0_beta5
 
 (* intermediates roots that need an hard upgrade *)
 let intermediate_roots = [
-  v2_1_alpha; v2_1_alpha2
+  v2_1_alpha; v2_1_alpha2; v2_1_rc
 ]
 
 let remove_missing_switches root conf =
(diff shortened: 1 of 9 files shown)


src/state/opamFormatUpgrade.ml
```diff
@@ -1210,6 +1219,24 @@ let as_necessary requested_lock global_lock root config =
...
+    let tostr sws =
+      OpamStd.Format.pretty_list
+        (List.map (OpamSwitch.to_string @> OpamConsole.colorise `bold) sws)
+    in
+    let switches = function [_] -> "" | _ -> "es" in
+    if global <> [] then
+      OpamConsole.warning "Broken global switch%s %s"
+        (switches global) (tostr global);
+    if not on_the_fly && local <> []then
+      OpamConsole.note "Removing local switch%s %s as it no longer exists"
```
dra27, line 1236: The plural form changes the sentence slightly: "as _they_ no longer _exist_"

src/state/opamFormatUpgrade.ml
```diff
@@ -1210,6 +1219,24 @@ let as_necessary requested_lock global_lock root config =
...
+    let local, global =
+      List.partition OpamSwitch.is_external missing_switches
+    in
+    let tostr sws =
+      OpamStd.Format.pretty_list
+        (List.map (OpamSwitch.to_string @> OpamConsole.colorise `bold) sws)
+    in
+    let switches = function [_] -> "" | _ -> "es" in
+    if global <> [] then
+      OpamConsole.warning "Broken global switch%s %s"
```
dra27, line 1233: The message in the reftests looks strange because it creates "Broken global switch missing". These switches are also removed, though, so could the sentence be in the same form as for local switches (I agree it's a warning - the user can reasonably delete local switches by accident, but manually deleting a global switch is naughty!).

"Removing references to broken global switch%s %s as it/they no longer exists/exist"?

dra27 (commented):
Possible message tweaks

src/state/opamFormatUpgrade.ml
rjbou: I intentionally didn't use the same wording of 'removal' because this is printed even in light upgrades (non blocking), users won't understand why a switch will be removed when they just want to install a package for example. But yes, better have two different sentences : 
* light: _Broken global switch%s %s_ , or use the same as in https://github.com/OCamlPro/opam/blob/master/src/state/opamSwitchState.ml#L30 _Removing local switch %s as it no longer exists_
* hard: _Removing references to broken global switch%s %s as it/they no longer exists/exist_ or the same as local ones
?

src/state/opamFormatUpgrade.ml
```diff
@@ -1189,6 +1228,33 @@ let as_necessary requested_lock global_lock root config =
...
+    in
+    (if not on_the_fly && global <> [] then
+       let switches, existence = sentences global in
+       OpamConsole.warning "Removing global switch%s %s as %s"
+         switches (tostr global) existence);
+    (if not on_the_fly && local <> [] then
+       let switches, existence = sentences local in
+       OpamConsole.note "Removing local switch%s %s as %s"
+         switches (tostr local) existence);
+    config
```
AltGr, line 1255: No need for a message, it's silently handled in the normal (no update) case (it's an expected workflow that you just `rm -rf` source dirs).

src/state/opamFormatUpgrade.ml
```diff
@@ -1054,10 +1054,26 @@ let v2_1_alpha2 = OpamVersion.of_string "2.1~alpha2"
 (* config & sw config downgrade opam-version to 2.0 and add opam root version *)
 let v2_1_rc = OpamVersion.of_string "2.1~rc"
 
-let _v2_1 = OpamVersion.of_string "2.1"
+let v2_1 = OpamVersion.of_string "2.1"
 
 let from_2_0_to_2_1_alpha _ conf = conf
 
+let get_unrecorded_local_switch conf =
```
AltGr, line 1061: I am not sure about this: format upgrade is not expected to depend on CWD ?

The proper way to do it would be to do the update lazily but it's probably not practical...

Another point: it's forbidden to touch an external switch if its opamroot doesn't match the current one, so that must be checked.

AltGr (changes requested):
Nice, thanks!

I am not sure about the unrecorded switches part though, we'd need to scan the whole disk for it to be consistent.

src/state/opamStateConfig.ml
```diff
@@ -281,12 +281,44 @@ module Repos = struct
...
-  OpamPath.Switch.switch_config root switch |>
-  OpamFile.Switch_config.BestEffort.read_opt |> function
+  let f = OpamPath.Switch.switch_config root switch in
+  match OpamFile.Switch_config.BestEffort.read_opt f with
   | None -> false
   | Some conf -> conf.OpamFile.Switch_config.opam_root = Some root
+  | exception (OpamPp.Bad_version _ as e) ->
+    match OpamFile.Config.raw_root_version (OpamPath.config root) with
+    | None -> raise e
+    | Some _ ->
```
rjbou, line 315: Should it be constrained to root version `2.1~rc/2.1` ?

dra27: I think so - doesn't this become a problem in future if it gets `opam-root-version: "2.2"`?

rjbou: With a `2.2` opam root, the local switch upgrade won't be triggered, yes.

dra27: Ah, of course, the whole thing is in the exception for an incorrect version.

src/state/opamStateConfig.ml
```diff
@@ -281,12 +281,44 @@ module Repos = struct
...
+    { opamfile with
+      file_contents =
+        List.map (fun item ->
+            match item.pelem with
+            | Variable (({pelem = "opam-version"; _} as opam_v),
+                        ({pelem = String "2.1"; _} as v)) ->
+              { item with
+                pelem = Variable ({opam_v with pelem = "opam-version"},
+                                  {v with pelem = String "2.0"})}
+            | _ -> item) opamfile.file_contents}
```
AltGr, line 299:
```suggestion
            | _ -> item)
          opamfile.file_contents }
```

src/state/opamStateConfig.ml
AltGr: Since this fix is only necessary for intermediate dev versions, I'd say we remove it from 2.2 or at least 2.3: it'll keep the upgrade code more maintainable and we can reasonably assume anyone who tried one of the affected versions switched to the release in the meantime ?

src/state/opamStateConfig.mli
```diff
@@ -137,3 +137,9 @@ module Repos : sig
   val safe_read:
     ?lock_kind: 'a lock -> 'b global_state -> OpamFile.Repos_config.t
 end
+
+(* Raw read an switch config to downgrade its [opam-version] from 2.1 to 2.0.
+   It is necessary to handle opam root and switch upgrade from 2.1
+   intermediates roots to 2.1. *)
```
AltGr, line 143: I'd add more context here to explain _why_ it is necessary ? E.g. "This allows a workaround for a bug in versions 2.1~xx which wrongly updated the declared switch versions, requiring that we fix it during [OpamFormatUpdate] from these specific intermediate versions"

tests/reftests/run.ml
```diff
@@ -476,6 +484,65 @@ let run_test ?(vars=[]) ~opam t =
           List.fold_left
             (fun vars (v, r) -> (v, r) :: List.filter (fun (w, _) -> v <> w) vars)
             vars bindings
+        | Cat files ->
+          let print_opamfile header file =
```
AltGr, line 488: This probably deserves to be a top-level function :grin:

AltGr (approved):
Only minor (mostly style and code doc) comments

src/state/opamStateConfig.ml
rjbou: it's unregistered local switch, they can stay around for a while without noticing

tests/reftests/run.ml
rjbou: it will be removed from `run.ml` to be an outside program

PR #4719, opened 2021-06-18
Sync the behaviour of the macOS sandbox script with Linux's

kit-ty-kate:
I'm not sure what was the reason behind having /tmp rw on macos but read-only on linux.
In opam 2.0 it seem to have been the invert at first (rw on linux and ro on macos) however it was synchronized to rw in both in 2.0.8 in see https://github.com/ocaml/opam/pull/3742.
In opam 2.1~rc a linux-only change was done which made it ro: https://github.com/ocaml/opam/pull/4589, but it was never ported to macos.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -66,7 +66,7 @@ Prefixes used to help generate release notes, changes, and blog posts:
   * Fix bypass-check handling on reinit [#4750 @rjbou] [#4763 @rjbou] [2.1.0~rc2 #4750 #4756]
 
 ## Sandbox
-  *
+  * Sync the behaviour of the macOS sandbox script with Linux's: /tmp is now ready-only [#4719 @kit-ty-kate]
 
 ## Repository management
   *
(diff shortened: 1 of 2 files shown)


rjbou (approved):

dra27:
Hold off on this one - I think we need the opposite fix. POSIX mandates that `/tmp` should be read-write for all applications and that `TMPDIR` if set points to a directory which is read-write an application. It doesn't say that if `TMPDIR` is set you should not use `/tmp`, so it's looking like we need to sync the Linux sandbox script, rather than the other way around.

AltGr:
IIRC our discussions, we decided the safer option was to keep disallowing writes to `/tmp` — even if we break POSIX ; the few packages directly writing to `/tmp` should be fixed already, or shouldn't be too hard to patch.

The alternatives are just not fully satisfactory

AltGr:
(_ergo_, I am fine with merging this!)

PR #4725, opened 2021-06-23
Bump version to 2.2.0~alpha~dev

dra27:
Version bumped to `2.2.0~alpha~dev` (which is versionly less than `2.2.0~alpha`)

`install.sh` obviously left alone - we'll need to revise this further for the gap between the 2.1.0 release and the 2.2.0~alpha release.


dra27:
Hah - we probably want to put another substitution in that test at some point!

PR #4731, opened 2021-06-25
avoid capturing the Unix.environment at top-level

hannesm:
Hi,

we develop an application that modifies the environment variables and then calls into opam-client. Adding support for depext was easy, but on Debian we need to set DEBIAN_FRONTEND to `noninteractive` to avoid having to customize the time zone.

This PR avoids in OpamSystem that at the top level, Unix.environment is executed and used (before we're able to set the DEBIAN_FRONTEND environment variable). Instead the `OpamStd.Env.list` (lazily) is used.

IMHO this should be part of 2.1. There are still other calls to `Unix.environment`, but none are executed at the top level :)

Best,

@reynir and @hannesm

### DIFF
--- a/src/core/opamSystem.ml
+++ b/src/core/opamSystem.ml
@@ -354,8 +354,8 @@ let real_path p =
 
 type command = string list
 
-let default_env =
-  Unix.environment ()
+let default_env () =
+  OpamStd.Env.list () |> List.map (fun (var, v) -> var^"="^v) |> Array.of_list
 
 let env_var env var =
   let len = Array.length env in
@@ -456,7 +456,8 @@ let t_resolve_command =
       else
         `Denied
   in
-  fun ?(env=default_env) ?dir name ->
+  fun ?env ?dir name ->
+    let env = match env with None -> default_env () | Some e -> e in
     resolve env ?dir name
 
 let resolve_command ?env ?dir name =
@@ -497,9 +498,10 @@ let print_stats () =
 let log_file ?dir name = temp_file ?dir (OpamStd.Option.default "log" name)
 
 let make_command
-    ?verbose ?(env=default_env) ?name ?text ?metadata ?allow_stdin ?stdout
+    ?verbose ?env ?name ?text ?metadata ?allow_stdin ?stdout
     ?dir ?(resolve_path=true)
     cmd args =
+  let env = match env with None -> default_env () | Some e -> e in
   let name = log_file name in
   let verbose =
     OpamStd.Option.default OpamCoreConfig.(!r.verbose_level >= 2) verbose
@@ -520,7 +522,8 @@ let make_command
   | `Denied -> permission_denied cmd
 
 let run_process
-    ?verbose ?(env=default_env) ~name ?metadata ?stdout ?allow_stdin command =
+    ?verbose ?env ~name ?metadata ?stdout ?allow_stdin command =
+  let env = match env with None -> default_env () | Some e -> e in
   let chrono = OpamConsole.timer () in
   runs := command :: !runs;
   match command with


dra27:
Ah, it wasn't as cleanly on `master` as I thought - please could you rebase it? A quick glance suggests that the behaviour is not a regression over 2.0, so assuming that's correct, it's definitely not going in 2.1.0!

hannesm:
This is a continuation of the work in #4111 (on master & 2.1, not part of 2.0). Since 2.1 integrated depext (which uses OpamSystem), this broke the behaviour that #4111 achieved.

The PR in here could be simpler by not using `OpamStd.Env`, but instead changing `OpamSystem.default_env` signature from `string array` to `unit -> string array` (which would arguably simpler to reason about), see e69cb179.

I guess @gasche @Armael (in case they're interested in using depext) would be interested in this PR.

hannesm:
The underlying issue is: I try to install `conf-pkg-config` in a nearly empty ubuntu 20.04 container with `--confirm-level=unsafe_yes`. The expectation is success, the result is that this `opam install` never finishes since it waits for input on command line.

Reproduction case:
```
# docker run -it --rm ubuntu:20.04

$ apt update
$ apt install -y --no-install-recommends --no-install-suggests git make gcc g++ ca-certificates libc-dev unzip patch wget
$ wget https://github.com/ocaml/opam/releases/download/2.1.0-rc2/opam-2.1.0-rc2-x86_64-linu
$ chmod +x opam-2.1.0-rc2-x86_64-linux
$ ./opam-2.1.0-rc2-x86_64-linux init --disable-sandboxing --bare
$ ./opam-2.1.0-rc2-x86_64-linux switch create default --empty
$ eval $(./opam-2.1.0-rc2-x86_64-linux env --switch=default)

$ env # check the output, eventually remove things via "unset"
OPAM_SWITCH_PREFIX=/root/.opam/default
HOME=/root
PATH=/root/.opam/default/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
_=/usr/bin/env

$ ./opam-2.1.0-rc2-x86_64-linux install --confirm-level=unsafe-yes conf-pkg-config
tput: No value for $TERM and no -T specified
[WARNING] Running as root is not recommended
  - install conf-pkg-config 2

    pkg-config

Let opam run your package manager to install the required system packages?
(answer 'n' for other options) [Y/n] y
+ /usr/bin/apt-get "install" "-qq" "-yy" "pkg-config"
- debconf: delaying package configuration, since apt-utils is not installed
- Selecting previously unselected package liblocale-gettext-perl.
(Reading database ... 10559 files and directories currently installed.)
- Preparing to unpack .../00-liblocale-gettext-perl_1.07-4_amd64.deb ...
- Unpacking liblocale-gettext-perl (1.07-4) ...
- Selecting previously unselected package libglib2.0-0:amd64.
- Preparing to unpack .../01-libglib2.0-0_2.64.6-1~ubuntu20.04.3_amd64.deb ...
- Unpacking libglib2.0-0:amd64 (2.64.6-1~ubuntu20.04.3) ...
- Selecting previously unselected package libglib2.0-data.
- Preparing to unpack .../02-libglib2.0-data_2.64.6-1~ubuntu20.04.3_all.deb ...
- Unpacking libglib2.0-data (2.64.6-1~ubuntu20.04.3) ...
- Selecting previously unselected package tzdata.
- Preparing to unpack .../03-tzdata_2021a-0ubuntu0.20.04_all.deb ...
- Unpacking tzdata (2021a-0ubuntu0.20.04) ...
- Selecting previously unselected package libicu66:amd64.
- Preparing to unpack .../04-libicu66_66.1-2ubuntu2_amd64.deb ...
- Unpacking libicu66:amd64 (66.1-2ubuntu2) ...
- Selecting previously unselected package libxml2:amd64.
- Preparing to unpack .../05-libxml2_2.9.10+dfsg-5ubuntu0.20.04.1_amd64.deb ...
- Unpacking libxml2:amd64 (2.9.10+dfsg-5ubuntu0.20.04.1) ...
- Selecting previously unselected package shared-mime-info.
- Preparing to unpack .../06-shared-mime-info_1.15-1_amd64.deb ...
- Unpacking shared-mime-info (1.15-1) ...
- Selecting previously unselected package xdg-user-dirs.
- Preparing to unpack .../07-xdg-user-dirs_0.17-2ubuntu1_amd64.deb ...
- Unpacking xdg-user-dirs (0.17-2ubuntu1) ...
- Selecting previously unselected package xz-utils.
- Preparing to unpack .../08-xz-utils_5.2.4-1ubuntu1_amd64.deb ...
- Unpacking xz-utils (5.2.4-1ubuntu1) ...
- Selecting previously unselected package libdpkg-perl.
- Preparing to unpack .../09-libdpkg-perl_1.19.7ubuntu3_all.deb ...
- Unpacking libdpkg-perl (1.19.7ubuntu3) ...
- Selecting previously unselected package libfile-fcntllock-perl.
- Preparing to unpack .../10-libfile-fcntllock-perl_0.22-3build4_amd64.deb ...
- Unpacking libfile-fcntllock-perl (0.22-3build4) ...
- Selecting previously unselected package pkg-config.
- Preparing to unpack .../11-pkg-config_0.29.1-0ubuntu4_amd64.deb ...
- Unpacking pkg-config (0.29.1-0ubuntu4) ...
- Setting up xdg-user-dirs (0.17-2ubuntu1) ...
- Setting up libfile-fcntllock-perl (0.22-3build4) ...
- Setting up libglib2.0-0:amd64 (2.64.6-1~ubuntu20.04.3) ...
- No schema files found: doing nothing.
- Setting up tzdata (2021a-0ubuntu0.20.04) ...
- debconf: unable to initialize frontend: Dialog
- debconf: (TERM is not set, so the dialog frontend is not usable.)
- debconf: falling back to frontend: Readline
- Configuring tzdata
- ------------------
- 
- Please select the geographic area in which you live. Subsequent configuration
- questions will narrow this down by presenting a list of cities, representing
- the time zones in which they are located.
- 
-   1. Africa      4. Australia  7. Atlantic  10. Pacific  13. Etc
-   2. America     5. Arctic     8. Europe    11. SystemV
-   3. Antarctica  6. Asia       9. Indian    12. US
- Geographic area: 
...
```

And here it waits for input. If DEBIAN_FRONTEND is set to `noninteractive` in the shell, there won't be an interaction (should this be done in depext (if confirm-level is unsafe_yes)?). As mentioned in the original issue, we don't use opam as a tool, but as a library (in https://github.com/roburio/orb) -- where we (due to the "let default_env = Unix.environment ()" at top level) can't `Unix.putenv "DEBIAN_FRONTEND" "noninteractive"` before the environment in opam-core is captured.

dra27:
I really have no problem with this change for `master` and so 2.2+, with the accompanying considerable testing which will go a new release. For 2.1, the behaviour has been the current one since the first alpha release. It is still possible (or at least it should be) to do 2.0 depext handling (i.e. via the plugin) - i.e. it's not a _regression_ here, it's an additional feature that's wanted. In orb's case, what was it doing for depext prior to switching to the 2.1 libs?

However, `--confirm-level=unsafe-yes` is _supposed_ to be non-interactive, so the fact apt is blocking is a _depext_ bug - I've proposed a fix to that in #4735.

I have also added orb to the [release process](https://github.com/ocaml/opam/wiki/Release-process) check list for the future!

hannesm:
Thanks for your comment @dra27. Answers inline:

> For 2.1, the behaviour has been the current one since the first alpha release.

Ok, and my point is that _only since #4111 got merged_, the behaviour is now very unintuitive "sometimes environment Y is used, sometimes environment Z".

> It is still possible (or at least it should be) to do 2.0 depext handling (i.e. via the plugin)

Maybe. Is this a supported and tested use case? Is it desirable to use it? Is it broken in the same way (i.e. is OpamSystem system, and thus the environment at startup)? Is it really worth going down that route?

> A quick glance suggests that the behaviour is not a regression over 2.0
> it's not a regression here, it's an additional feature that's wanted

Well, so 2.0 does not include #4111, thus setting environment variables before the opam library startup was pretty undefined / not doable. Since #4111 got merged, a clearer semantics came into existance. But, as you can see in this PR, there is a bug (missing treatment of depext and other things that are executed via OpamSystem without custom `vars`) in #4111, which only came to existance in #3975 (merged after #4111).

> In orb's case, what was it doing for depext prior to switching to the 2.1 libs?

None, orb is still unreleased, and can only work with 2.1+ (it requires #4040 and #4055 (and maybe others I forgot)).

Your fix in #4735 makes opam more robust in non-interactive mode (i.e. CI systems), this is great! :) It does not help, as outlined above, for reproducible builds, where being in control of the environment variables is crucial.

If you still think "the 2.1 HEAD is fine", I'd invite you to rephrase the changes entry `* Ensure that environment is initialized lazily, not before init functions are called [#4111 @gasche]` -- maybe `* Ensure that for some operations of the opam library the environment is initialized lazily, others (OpamSystem and its callers) still use a cached version of the environment at startup (i.e. before anything could execute `unsetenv` or `putenv`)`. My opinion is that e69cb17 is a bugfix that delays the Unix.environment call (and clears the semantics to the ones described in the current CHANGES entry), which is rather safe to merge -- even at this stage of development -- at a cost of several calls to `Unix.environment` (which e9d11d0 addresses directly by using `OpamStd.Env`).

Armael:
Like @hannesm I would also see this change as a bugfix for #4111, that would be really nice to have in 2.1 (also for our work with @gasche).

My current understanding is that the changes proposed here are not observable if you use opam+depext the "normal way" (as a binary), and make it usable as a library when one wants to be in control of the environment---which was indeed the goal behind #4111 and only partially achieved. I doubt anyone has been relying on the post-#4111 inconsistent behavior while using opam(+depext) as a library, especially considering that using opam as a library seems already like a pretty niche usecase.

Without this patch, opam-as-a-library would be left in a somewhat inconsistent state where it's mostly possible to control the environment but not always, which seems awkward.

kit-ty-kate:
We're reluctant about adding that during an rc phase because it hasn't been tested at any point during the alpha/beta/rc and it is not a critical fix nor a regression. However we're happy to merge this at part of the 2.1.1 planned right after the 2.1.0 release (we had a couple of critical bugs during the rc2 that pushed back the final release once again, sorry about that)

reynir:
I think it is not a great idea to intentionally change the semantics of opam-as-a-library in a patch release.

Armael:
I think that's totally fair :-). Sorry if my message read too much like pushing for immediate inclusion, I understand that everyone is busy with the release.

rjbou:
Changed base branch to master, and it will be backported into 2.1.1.

rjbou:
I can't force push to update the PR with the rebase on master.

hannesm:
I opened #4789 which is this PR rebased to master. I changed the based to 2.1 here (we use this branch in a custom opam-repository (https://git.robur.io/robur/unikernel-repo) to actually use orb (https://builds.robur.coop)).

rjbou:
Closed in favor of #4789

Issue #4732, opened 2021-06-27
SSL certificate has expired for 'opam.ocaml.org'?


wiremoons:
Trying to run `opam init` and get error:

```
opam init
[NOTE] Will configure from built-in defaults.
Checking for available remotes: rsync and local, git.
  - you won't be able to use mercurial repositories unless you install the hg command on your system.
  - you won't be able to use darcs repositories unless you install the darcs command on your system.


<><> Fetching repository information ><><><><><><><><><><><><><><><><><><><>  🐫 
[ERROR] Could not update repository "default": OpamDownload.Download_fail(_, "Curl failed: \"/usr/bin/curl --write-out %{http_code}\\\\n --retry 3 --retry-delay 2
        --user-agent opam/2.0.8 -L -o /private/var/folders/dv/nf3vf1xj46lckqgbjm12l9yw0000gn/T/opam-13783-e7966f/index.tar.gz.part https://opam.ocaml.org/index.tar.gz\"
        exited with code 60")
[ERROR] Initial download of repository failed
```

If try to use `curl` directly:
```
curl https://opam.ocaml.org/index.tar.gz
curl: (60) SSL certificate problem: certificate has expired
More details here: https://curl.haxx.se/docs/sslcerts.html

curl failed to verify the legitimacy of the server and therefore could not
establish a secure connection to it. To learn more about this situation and
how to fix it, please visit the web page mentioned above.
```

Visiting the web site: https://opam.ocaml.org/ also get a browser warning as certificate has expired, with message:
```
This server could not prove that it is opam.ocaml.org; its security certificate expired in the last day.
```

mt-caret:
For those who have encountered the same issue, here's a temporary fix which seems to work.
```
OPAMFETCH='curl --insecure --write-out %{http_code}\n --user-agent opam/2.0.8 -L -o %{out}% -- %{url}%' opam init```

dra27:
The certificate's been renewed. Sorry for the inconvenience - the problem should be permanently resolved in the next few weeks!

PR #4733, opened 2021-06-27
add a way to skip the certificate check

mt-caret:
Currently, there is no easy way to skip the default certificate check to circumvent issues with the certificate for opam servers (like https://github.com/ocaml/opam/issues/4732). This PR makes it possible to add an environment variable to fix this.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -61,7 +61,7 @@ Prefixes used to help generate release notes, changes, and blog posts:
   *
 
 ## Repository management
-  *
+  * [NEW] Add OPAMSKIPCERTIFICATECHECK for skipping SSL certificate checking [#4733 @mt-caret] [2.1.0~rc2 #4720]
 
 ## VCS
   *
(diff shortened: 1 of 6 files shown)


mt-caret:
Unfortunately, I can't set up an environment where I can compile or test this due to https://github.com/ocaml/opam/issues/4732

mt-caret:
Built from source in an empty Ubuntu 21.04 container and ran `OPAMSKIPCERTIFICATECHECK=true opam init` to verify that it indeed does as advertised.

dra27:
Thanks for looking at this: there are quite a few technical issues with this PR, but I don't think we should do this at all. `https://opam.ocaml.org` is the _default_ opam remote. The repository is served over both http and https, so the permanently workaround if you never want to be affected by certificate issues is to specify the http remote, i.e. `opam init http://opam.ocaml.org` or initialise securely with a git remote `opam init git+https://github.com/ocaml/opam-repository`.

dra27:
(That information should definitely be in our FAQ!)

mt-caret:
> Thanks for looking at this: there are quite a few technical issues with this PR, but I don't think we should do this at all. `https://opam.ocaml.org` is the _default_ opam remote. The repository is served over both http and https, so the permanently workaround if you never want to be affected by certificate issues is to specify the http remote, i.e. `opam init http://opam.ocaml.org` or initialise securely with a git remote `opam init git+https://github.com/ocaml/opam-repository`.

Ah, I didn't know that was an option. Thanks for the information!

PR #4736, opened 2021-06-29
deprecated syntax removed

vzaliva:
According to https://fishshell.com/docs/current/index.html#id4 
"Previous versions of fish also allowed specifying this as ^DESTINATION, but that made another character special so it was deprecated and will be removed in the future."

fish version 3.3.0 by default gives an error on old syntax

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -104,7 +104,7 @@ Prefixes used to help generate release notes, changes, and blog posts:
   * Fix ocaml link (http -> https) [#4729 @rjbou]
 
 ## Shell
-  *
+  * fish: fix deprecated redirection syntax `^` [#4736 @vzaliva]
 
 ## Doc
   *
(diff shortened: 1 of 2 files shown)


rjbou:
Thanks for the update. Does `2>` is full valid syntax for fish < 3.3.0 ?

vzaliva:
I am a relatively new fish user and not sure about older versions. However, from the comment in their documentation, it seems it was supported all along with ^ being just a shortcut which they deprecated.

rjbou (approved):

rjbou:
tested with fish 3.1.2, and seems ok.

smorimoto:
This seems worth backporting to 2.0.9 and 2.1.0 as well. In fact, I found this after updating fish today and reinit didn't fix the error, so I got here.

kit-ty-kate:
I just opened https://github.com/ocaml/opam/pull/4743 and https://github.com/ocaml/opam/pull/4743 to backport to the older branches.

TonalidadeHidrica:
@kit-ty-kate Thanks for the backport!  Could you just fix your comment "4743 and 4743" to "4743 and 4744" for easy reference?

PR #4742, opened 2021-07-02
Ignore new `opam-version` field in .changes files

dra27:
Possibly change for 2.0.9 - the .changes files in 2.1 now have `opam-version: "2.0"` at the top, which causes a lot of warnings to displayed unconditionally when reading those files from a plugin compiled with opam-format.2.0.9

I haven't checked for other file formats affected by this.

### DIFF
--- a/src/format/opamFile.ml
+++ b/src/format/opamFile.ml
@@ -3138,6 +3138,7 @@ module ChangesSyntax = struct
                   Pp.of_pair "digest" (digest_of_string, string_of_digest))))
 
   let fields = [
+    "opam-version", Pp.ppacc_ignore;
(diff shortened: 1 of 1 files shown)


dra27:
I think `Switch_configSyntax` and `InitConfigSyntax` should call `check_opam_version` too? There's already a check on `opam-version` in `Dot_configSyntax`, so I'm not sure if it needs to call `check_opam_version` too or be changed slightly?

rjbou:
It is not strictly needed for version handling of updated opam files with 2.1. The consistency of calling `check_opam_version` for all files was introduced in 2.1 with the new module value `file_format`.

PR #4744, opened 2021-07-03
Backport #4736 to the 2.0 branch

kit-ty-kate:
Attempt at backporting https://github.com/ocaml/opam/pull/4736 to the 2.0 branch
@rjbou I'm not sure how you usually do, but just in case this is useful.. ^^


dra27:
There's another possible fix to slip into 2.0.9, so this might just get in there, rather than 2.0.10... let's see next week when @rjbou and @AltGr are back from holiday!

rjbou:
Thanks!

PR #4747, opened 2021-07-08
opamSpdxList: add LGPL-3.0 linking exception

mseri:
Seen on https://github.com/ocaml/opam-repository/pull/19010

- [x] Update `master_changes.md` file with your changes.

### DIFF
--- a/src/state/opamSpdxList.ml
+++ b/src/state/opamSpdxList.ml
@@ -436,4 +436,5 @@ let exceptions = OpamStd.String.Set.of_list @@ List.map String.lowercase_ascii @
   "GCC-exception-2.0";
   "mif-exception";
   "OCaml-LGPL-linking-exception";
+  "LGPL-3.0-linking-exception";
 ]
--- a/master_changes.md
+++ b/master_changes.md
@@ -46,7 +46,7 @@ Prefixes used to help generate release notes, changes, and blog posts:
   *
 
 ## Lint
-  *
+  * Added support for LGPL-3 linking exception
 
 ## Lock
   *
--- a/master_changes.md
+++ b/master_changes.md
@@ -46,7 +46,7 @@ Prefixes used to help generate release notes, changes, and blog posts:
   *
 
 ## Lint
-  * Added support for LGPL-3 linking exception
+  * Add support for LGPL-3 linking exception [#4747 @mseri]
 
 ## Lock
   *


dra27:
Given the relevance to OCaml, I wonder if it's worth taking this opportunity to add a lint warning (or even error) when the exception and licence combination is invalid (cf. https://github.com/ocaml/opam-repository/pull/19010#discussion_r666229334).

opam-repository features these uses of `OCaml-LGPL-linking-exception`:
`LGPL-2.0-only WITH OCaml-LGPL-linking-exception`
`LGPL-2.1-only WITH OCaml-LGPL-linking-exception`
`LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception`
`LGPL-2.1 WITH OCaml-LGPL-linking-exception`
`LGPL-3.0-only WITH OCaml-LGPL-linking-exception`
`LGPL-3.0-or-later WITH OCaml-LGPL-linking-exception`

The exception has existed with both GNU Library GPL v2.0 and GNU Lesser GPL v2.1 (clause 2 is identical in both and, although I'm not lawyer, clause 6 does not look materially different), so I think the `LPGL-2.0-only`, `LGPL-2.1-only` and `LGPL 2.1` uses are clearly valid.

With even more emphasis on not being a lawyer, I would assume that `LGPL-2.1-or-later` requires a legal common-sense application of the exception to LPGL 3.0 (i.e. the fact that the clause numberings are different does not invalidate the intent of the exception).

However, that's quite clearly not the case for `LGPL-3.0-only` and `LGPL-3.0-or-later` so it would be a good service to the community if we rejected them!

src/state/opamSpdxList.ml
```diff
@@ -436,4 +436,5 @@ let exceptions = OpamStd.String.Set.of_list @@ List.map String.lowercase_ascii @
   "GCC-exception-2.0";
   "mif-exception";
   "OCaml-LGPL-linking-exception";
+  "LGPL-3.0-linking-exception";
```
kit-ty-kate, line 439: This file is autogenerated so this isn't quite the right fix (there is a comment in the dune file)
However https://github.com/ocaml/opam/pull/4768 should fix that for good hopefully.

kit-ty-kate:
Superseded by #4768. Thanks though!

PR #4760, opened 2021-07-22
Update to dose >= 6.1

kit-ty-kate:
Superseds #4470


README.md
```diff
@@ -37,7 +37,7 @@ easier to already have a working opam installation in this case, so you can do
 it as a second step.
 
 * Make sure to have ocamlfind, ocamlgraph, cmdliner >= 0.9.8, cudf >= 0.7,
-  dose3 >= 5, re >= 1.5.0, opam-file-format installed. Or run `opam install
+  dose3 >= 6.1, re >= 1.5.0, opam-file-format installed. Or run `opam install
```
dra27, line 40:
```suggestion
  dose3 >= 6.1, re >= 1.9.0, opam-file-format installed. Or run `opam install
```
(while we here!)

opam-client.opam
```diff
@@ -29,7 +29,7 @@ depends: [
   "ocaml" {>= "4.02.3"}
   "opam-state" {= version}
   "opam-solver" {= version}
-  "extlib" {>= "1.7.3" & < "1.7.8"}
+  "base64" {>= "3.1.0"}
```
dra27, line 32: `cudf` depends on extlib, so I think we need to either depend on extlib >= 1.7.8 or conflict extlib < 1.7.8

dra27 (commented):
I've pushed a commit from the weekend which should unblock most of the builds. This also needs the 4.02 support to be removed to succeed completely.

opam-client.opam
kit-ty-kate: but opam does not depend on it explicitly so removing it from the dependency list is the correct thing to do in my opinion.

dra27: opam-client _does_ depend on it - it requires extlib either being not present at all or being 1.7.8 or later, as otherwise base64 cannot link. I agree that `depends` feels weird for this, so perhaps `conflicts` is the best, therefore.

It's slightly moot, given that dose3 already has `extlib > 1.7.8`

kit-ty-kate: Ah, right. Then a `conflicts` is appropriate in my opinion.

dra27: There's a similar one still floating around for extlib-compat, although I didn't double-check that that's needed (I think that's because you get a conflict on Windows and possibly macOS over Uchar vs UChar otherwise)

kit-ty-kate: Conflicts added.

README.md
kit-ty-kate: Done.

dra27:
We’re dropping 4.02 support for 2.2.0 - I’ve been polishing the branch this afternoon!

dra27:
Although IIRC you said there are noticeable improvements to some conflict messages with Dose 6, so I’m guessing we will want to put this in 2.1.1 and possibly 2.0.10?

kit-ty-kate:
Yeah we can always remove it in a later PR. It wasn't much work and allows this PR to be merged and tested ASAP.

PR #4761, opened 2021-07-22
Make opam partially compilable on stock macOS/FreeBSD

kit-ty-kate:
FreeBSD and macOS do not have md5sum by default but `./configure && make lib-ext && make` works fine with this patch at least on macOS

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -76,6 +76,7 @@ Prefixes used to help generate release notes, changes, and blog posts:
 
 ## Build
   * Bump src_exts and fix build compat with Dune 2.9.0 [#4752 @dra27]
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
I'm not sure how to test this in CI. The github actions seem to have coreutils installed somehow, even though it's not listed in https://github.com/actions/virtual-environments/blob/main/images/macos/macos-10.15-Readme.md

kit-ty-kate:
Actually i know why it worked.. `make lib-ext` is permitted even if you haven't called the configure script, it works without md5sum with the shell/md5check.ml script. I'll open a different PR

Issue #4598, opened 2021-03-15
opam lint should raise a warning when the license field is missing


kit-ty-kate:
Mentioned in https://github.com/ocaml/opam-repository/issues/18343

It would be also nice to have (if there is not already one) an escape hatch for more complex licenses that don't fit in the SPDX standard. I would suggest something like `license: "custom:<custom license name>"`

dra27:
Cross-referencing https://github.com/ocaml/opam/issues/4686#issuecomment-855776339

dra27:
Do we definitely want an escape hatch - having gone to the trouble of getting things like the OCaml linking exception text ratified and added to SPDX (see https://github.com/ocaml/opam/pull/2224), is it too bad for us to make it a lint warning with no escape?

kit-ty-kate:
I'm currently making an SPDX parser to add to a separate PR to #4766. It turns out the SPDX format already has a way to add custom licenses, e.g. `LicenseRef-my-special-license` (https://spdx.github.io/spdx-spec/appendix-V-using-SPDX-short-identifiers-in-source-files/).

However I prefer to split the two features, as currently it's not such a big deal to not have support for custom licenses. We already ignore the warnings in opam-repo-ci when needed and we can have both at different times and even backport one if it's really necessary. The SPDX parser feature is larger chunk of work so I also prefer to not slow down the addition of feature actively asked by users.

dra27:
All good - I definitely agree with getting the missing check on license straight in, too

PR #4766, opened 2021-07-23
opam lint: Warn if the license field is not present

kit-ty-kate:
Partially fixes https://github.com/ocaml/opam/issues/4598

### DIFF
--- a/.github/scripts/solvers.sh
+++ b/.github/scripts/solvers.sh
@@ -3,6 +3,7 @@
 . .github/scripts/preamble.sh
 
 export OPAMYES=1
+export OPAMCONFIRMLEVEL=unsafe-yes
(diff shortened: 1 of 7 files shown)


.github/scripts/solvers.sh
```diff
@@ -3,6 +3,7 @@
 . .github/scripts/preamble.sh
 
 export OPAMYES=1
```
dra27, line 5: A very small nit, but as we set `OPAMCONFIRMLEVEL`, shall we stop setting `OPAMYES`?

kit-ty-kate: Added to the dose-7 PR: https://github.com/ocaml/opam/pull/4760/commits/05533008c535999fdb479cb05319f3f98fc8d86e

PR #4768, opened 2021-07-26
opam lint: Add support for the full SPDX license expression when checking the license

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/4598 and https://github.com/ocaml/opam/pull/4747

Implementation details at https://github.com/kit-ty-kate/spdx_licenses

Currently during linting, opam checks that the license field is a valid SPDX license. However:
* the list was outdated (as shown by #4747)
* the SPDX syntax itself isn't supported
* there is no support for custom licenses

The best one can do is to use the list format to say un implicit AND/OR (e.g. `license: ["MIT" "ISC"]`) and simply ignore the warning when the license isn't supported.

This PR fixes all of those issues by allowing the complete SPDX license syntax: https://spdx.github.io/spdx-spec/appendix-IV-SPDX-license-expressions/
The syntax supports:
* custom licenses (e.g. `LicenseRef-custom-name`)
* AND, OR, parenthesis
* ...

Compared to the current state, apart from the new feature added, this changes:
* the way the `WITH` operator is parsed. Currently opam parses it in a case-insensitive manner, but the SPDX standard requires it to be case-sensitive (see https://spdx.github.io/spdx-spec/appendix-IV-SPDX-license-expressions/#case-sensitivity). So licenses such as `LGPL-2.1-or-later with OCaml-LGPL-linking-exception` will have to be renamed `LGPL-2.1-or-later WITH OCaml-LGPL-linking-exception` to fit the standard.


kit-ty-kate:
This PR requires #4770 to be merged first. I've updated the description for a summary of the changes. This is ready to review.

Current checklist for merging this includes:
* [ ] #4770 merged
* [ ] `spdx_licenses.1.0.0` is released

AltGr (approved):
This is great, thanks !

PR #4775, opened 2021-07-29
Switch minimum OCaml version to 4.08

dra27:
Continues from #4770. opam-core, opam-format and opam-installer are now tested separately where only the libraries are built.

CI is still quite painful because the ocaml-secondary-compiler, however, I have an overhaul PR ready to go which sorts out the caching for that.

### DIFF
--- a/opam-core.opam
+++ b/opam-core.opam
@@ -26,7 +26,7 @@ build: [
   ["dune" "build" "-p" name "-j" jobs]
 ]
 depends: [
-  "ocaml" {>= "4.03.0"}
+  "ocaml" {(>= "4.03.0" & os != "win32") | >= "4.06.0"}
   "base-unix"
   "base-bigarray"
   "ocamlgraph"
--- a/opam-state.opam
+++ b/opam-state.opam
@@ -26,7 +26,7 @@ build: [
   ["dune" "build" "-p" name "-j" jobs]
 ]
 depends: [
-  "ocaml" {>= "4.03.0"}
+  "ocaml" {>= "4.08.0"}
   "opam-repository" {= version}
   "re" {>= "1.9.0"}
   "dune" {>= "2.0.0"}
--- a/.github/scripts/legacy.sh
+++ b/.github/scripts/legacy.sh
@@ -1,0 +1,16 @@
+#!/bin/bash -xue
+
+. .github/scripts/preamble.sh
+
+export OPAMYES=1
+export OCAMLRUNPARAM=b
+
+opam init default git+$OPAM_REPO#$OPAM_REPO_SHA
+eval $(opam env)
+OPAMCLI=2.0 opam reinstall ocaml-system --unlock-base
+opam upgrade
+for package in core format installer ; do
+  opam pin add -yn opam-$package .
+done
+opam install opam-core opam-format opam-installer --deps-only
+dune build --profile=dev --only-packages opam-core,opam-format,opam-installer @install
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -108,14 +108,14 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest ]
-        ocamlv: [ 4.03.0, 4.04.2, 4.05.0, 4.06.1, 4.07.1, 4.08.1, 4.09.1, 4.10.2, 4.11.2, 4.12.0 ]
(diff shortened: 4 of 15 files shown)


opam-state.opam
```diff
@@ -26,7 +26,7 @@ build: [
   ["dune" "build" "-p" name "-j" jobs]
 ]
 depends: [
-  "ocaml" {>= "4.03.0"}
+  "ocaml" {>= "4.08.0"}
```
kit-ty-kate, line 29: It is to be noted that notable plugins, such as `dune-release` and `opam-publish`, use `opam-state`.

Not a blocker ofc but it's probably good to ping the devs. cc @NathanReb

opam-core.opam
```diff
@@ -26,7 +26,7 @@ build: [
   ["dune" "build" "-p" name "-j" jobs]
 ]
 depends: [
-  "ocaml" {>= "4.03.0"}
+  "ocaml" {>= "4.03.0" & os != "win32" | >= "4.06.0"}
```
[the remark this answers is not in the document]
dra27: Boolean algebra 😱 (no problem to add, obviously 😉)

opam-state.opam
dra27: Indeed - I'd checked those through before proposing it. opam-publish is already pulled towards 4.08+ by depending on cohttp via ocaml-github and dune-release was already 4.06+ (and I think @emillon agreed it would shift on @NathanReb's behalf!)

NathanReb: I confirm there's no problem on my side for bumping minimum requirements on OCaml, to the contrary I'm rather glad about it!

kit-ty-kate (approved):

kit-ty-kate:
cc @hannesm just in case. I'm not sure which part of the opam library you are using

hannesm:
From my development perspective (orb, conex), this is not an issue

With a packagers view, I find this kind of surprising: according to https://repology.org/project/ocaml/versions there's still a huge amount of stable distributions that ship OCaml 4.05.0. Raising the minimum OCaml version will exclude any of these to upgrade opam (if they rely on their packaging scripts and packages to build binary packages).

This will make it much harder to bootstrap (upgrade to) opam-after-2-1-0 on various distributions / platforms. I fail to see the reasoning behind raising the minimum requirement.

kit-ty-kate:
> With a packagers view, I find this kind of surprising: according to https://repology.org/project/ocaml/versions there's still a huge amount of stable distributions that ship OCaml 4.05.0. Raising the minimum OCaml version will exclude any of these to upgrade opam (if they rely on their packaging scripts and packages to build binary packages).

The repology link shows that all of the distribution that ship OCaml < 4.08 are older version that won't update to opam >= 2.2.0 anyway (Debian Stable, Ubuntu OldLTS, CentOS, Alpine 3.10 (the latest being 3.14), …). The only on I personally know that doesn't match this pattern is FreeBSD which is still in 4.05, however there is a current ongoing effort to port to 4.11: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=250408

> This will make it much harder to bootstrap (upgrade to) opam-after-2-1-0 on various distributions / platforms. I fail to see the reasoning behind raising the minimum requirement.

To me, most people either:
- use the opam version shipped with their distribution. But usually don't care too much about being up-to-date (unless you're runnig Archlinux or something)
- Or use the opam release binaries (e.g. via the `install.sh` script) if they want to be more up-to-date

The people that really impacts it are people who compile opam from source, but for those we have `make cold` already which will compile an appropriate version of the compiler and compiles opam with that. Note that distributions can also use `make cold` if they need to.

The reasoning being raising the minimum requirement is that of developement simplicity: Everytime we upgrade the vendored libraries, we have to patch it to make it compatible with older versions of the compiler, we also lacked access to very good features in the compiler (e.g. exception matching from 4.03, access to the `Result`, `Option` modules from `Stdlib`, …). This all induces stress and slows down the development of opam.

hannesm:
@kit-ty-kate thanks for your reply.

dra27:
Good to go after CI comes back. Just to add that the bump to 4.03 was the big de-stresser (we finally can guarantee `String.lowercase_ascii`!) - the useful bump to 4.08 for opam-state onwards is that we can make the 0install solver a mandatory dependency.

PR #4777, opened 2021-07-29
Use enabled_if to control the Windows stubs

dra27:
Still draft as I need to check in more detail whether some of the compiler tests in `configure` either need to be removed entirely or converted to be called from `dune`.


rjbou:
rebased

dra27:
Fixed in #5381

PR #4782, opened 2021-08-03
Doc improvements

kit-ty-kate:
The change of the advised way to call `install.sh` was proposed in https://github.com/ocaml/platform-blog/pull/77 and includes:
* `<()` is a bash syntax, so people using this command in `sh` or `csh` will see the command fail
* `curl` without `-fS` will output things in stdout when something goes wrong (404, and other network failure) and can cause random shell commands being executed. This is what https://brew.sh/ recommend for instance.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -5,6 +5,10 @@ Prefixes used to help generate release notes, changes, and blog posts:
 * ◈ New option/command/subcommand
 * [BUG] for bug fixes
 * [NEW] for new features (not a command itself)
+* [API] api updates 🕮
+If there is changes in the API (new non optional argument, function renamed or
+moved, etc.), please update the _API updates_ part (it helps opam library
+users)
 
 ## Version
   *
@@ -92,7 +96,7 @@ Prefixes used to help generate release notes, changes, and blog posts:
 ## State
   *
 
-# Opam file format
+## Opam file format
   *
 
 ## Solver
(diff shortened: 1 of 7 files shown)


rjbou:
+ update of readme badges, release instructions, changelog for api follow-up

master_changes.md
```diff
@@ -120,7 +124,15 @@ Prefixes used to help generate release notes, changes, and blog posts:
   * fish: fix deprecated redirection syntax `^` [#4736 @vzaliva]
 
 ## Doc
-  *
+  * Standardise `macOs` use [#4782 @kit-ty-kate]
```
dra27, line 127:
```suggestion
  * Standardise `macOS` use [#4782 @kit-ty-kate]
```
🙂

master_changes.md
```diff
@@ -5,6 +5,10 @@ Prefixes used to help generate release notes, changes, and blog posts:
 * ◈ New option/command/subcommand
 * [BUG] for bug fixes
 * [NEW] for new features (not a command itself)
+* 🕮i api update
```
dra27, line 8: What's the emoji bit for, or is my browser being strange?

I'm not utterly convinced that this is a good idea, but let's see how it goes!

dra27 (commented):
One, or possibly two, typos but otherwise LGTM!

master_changes.md
rjbou: héhé, it is just a book for _library_ , maybe we should have the more intelligible `[API]`. The idea is to point API breakage/updates that can affect library users

PR #4784, opened 2021-08-05
Add default TMPDIR value for sandbox checker

### LINKED ISSUE #4783: Upgrading opam to 2.1.0 touches files in my CWD, but it shouldn't
I'm trying to upgrade to opam 2.1 from 2.1~beta-something (I think it was beta4). It does something with bwrap and tries to create something in my current workdir (`/home/kakadu/asp/typerex-lint`): see ` "bwrap: Can't mkdir /home/kakadu/asp/typerex-lint: No such file or directory"` in the output below. Why? Is it intended? Should I be worried?

```
$ bash -c "sh <(curl -fsSL https://raw.githubusercontent.com/ocaml/opam/master/shell/install.sh) --version 2.1.0"
## Downloading opam 2.1.0 for linux on x86_64...
## Downloaded.
## Where should it be installed ? [/home/kakadu/bin] 
## opam 2.1.0 installed to /home/kakadu/bin
## Converting the opam root format & updating
This version of opam requires an update to the layout of /media/work/.opam from version 2.1~alpha2 to version 2.1, which can't be reverted.

Perform the update and continue? [Y/n] Y
Format upgrade done.

<><> Rerunning init and update ><><><><><><><><><><><><><><><><><><><><><><><><>

[ERROR] Sandboxing is not working on your platform ubuntu:
        "/media/work/.opam/opam-init/hooks/sandbox.sh build sh -c echo SUCCESS >$TMPDIR/opam-sandbox-check-out && cat $TMPDIR/opam-sandbox-check-out; rm -f $TMPDIR/opam-sandbox-check-out"
        exited with code 1 "bwrap: Can't mkdir /home/kakadu/asp/typerex-lint: No such file or directory"
Do you want to disable it?  Note that this will result in less secure package builds, so please ensure that you have some other isolation mechanisms in place (such as running within a
container or virtual machine). [y/N] 
```

My config 
```
$  cat ~/.opam/config
opam-version: "2.0"
opam-root-version: "2.1"
repositories: "default"
installed-switches: [
  "memtrace"
  "4.11.2+fp+flambda"
  "4.10.1+flambda"
  "4.11.2+32bit"
  "4.11.1+BER"
  "4.12.0+fp+flambda"
  "4.10.1+fp+flambda"
  "4.07.1+fp+flambda"
]
switch: "4.10.1+flambda"
jobs: 3
download-jobs: 3
eval-variables: [
  sys-ocaml-version
  ["ocamlc" "-vnum"]
  "OCaml version present on your system independently of opam, if any"
]
default-compiler: [
  "ocaml-system" {>= "4.01.0"}
  "ocaml-base-compiler"
]
depext: true
depext-run-installs: true
depext-cannot-install: false
```

The `bwrap` is installed system-wide, my CWD seems not to be pinned

rjbou:


src/client/opamAuxCommands.ml
```diff
@@ -504,7 +504,7 @@ let check_and_revert_sandboxing root config =
   | [] -> config
   | cmd::_ ->
     (* All the provided sandboxing scripts are expected to define [TMPDIR] *)
-    let test_file = "$TMPDIR/opam-sandbox-check-out" in
+    let test_file = "${TMPDIR:-/tmp}/opam-sandbox-check-out" in
```
kit-ty-kate, line 507: The default sandbox forbids writes to `/tmp` so this would fail anyway.

kit-ty-kate (changes requested):

src/client/opamAuxCommands.ml
dra27: Yes, but I think it's a good hardening regardless - at least there's no possible situation where we attempt to write to `/`

dra27:
I've unlinked it from #4783 as it doesn't fix it, but I think it's a good idea regardless - it looks better to be sure that even with a faulty sandbox script we don't try to write and erase a file in `/`!

dra27:
Are you sure you want to close this one, @rjbou?

Issue #4751, opened 2021-07-14
Problems with .opam/download-cache and packages which depend on git information


Alizter:
I have a problem with the download-cache folder of opam. To replicate the behaviour try the following:

1. Get the coq dev packages: `opam repo add coq-core-dev https://coq.inria.fr/opam/core-dev`
2. Create a switch and install coq.8.13.dev
3. Now create another switch and install coq.dev
4. The installation of coq will fail due to some confusion with the git history in download-cache.

I know that this is not a problem with the coq packaging, because if you delete the download-cache folder between steps 2 and 3 it will work. For some reason, opam is not doing something right with the git stuff here.

```
# opam config report
# opam-version         2.1.0~rc2 
# self-upgrade         no
# system               arch=x86_64 os=linux os-distribution=ubuntu os-version=18.04
# solver               builtin-mccs+glpk
# install-criteria     -removed,-count[avoid-version,changed],-count[version-lag,request],-count[version-lag,changed],-count[missing-depexts,changed],-changed
# upgrade-criteria     -removed,-count[avoid-version,changed],-count[version-lag,solution],-count[missing-depexts,changed],-new
# jobs                 15
# repositories         4 (http) (default repo at 8c58ed60)
# pinned               1 (version)
# current-switch       coq.dev
# ocaml:native         true
# ocaml:native-tools   true
# ocaml:native-dynlink true
# ocaml:stubsdir       /mnt/sdd1/.opam/coq.dev/lib/ocaml/stublibs:/mnt/sdd1/.opam/coq.dev/lib/ocaml
# ocaml:preinstalled   false
# ocaml:compiler       4.12.0

```

kit-ty-kate:
would you be able to post the error messages for the error?

Alizter:
```
<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
[ERROR] The installation of coq failed at "make install".

#=== ERROR while installing coq.dev ===========================================#
# context     2.1.0~rc2 | linux/x86_64 |  | pinned(git+https://github.com/coq/coq.git#master#f684f34e)
# path        /mnt/sdd1/.opam/coq.dev/.opam-switch/build/coq.dev
# command     /mnt/sdd1/.opam/opam-init/hooks/sandbox.sh install make install
# exit-code   2
# env-file    /mnt/sdd1/.opam/log/coq-19150-d81af9.env
# output-file /mnt/sdd1/.opam/log/coq-19150-d81af9.out
[...]

PR #4787, opened 2021-08-06
Run sandbox checker in temp directory

dra27:
The sandbox will attempt to mount the current working directory which may fail (and which we don't necessarily have any business doing).

### DIFF
--- a/src/core/opamSystem.mli
+++ b/src/core/opamSystem.mli
@@ -31,8 +31,12 @@ exception Internal_error of string
 (** Raise [Internal_error] *)
 val internal_error: ('a, unit, string, 'b) format4 -> 'a
 
-(** [with_tmp_dir fn] executes [fn] in a tempory directory *)
+(** [with_tmp_dir fn] executes [fn] creates a temporary directory and
+    passes its name to [fn]. The directory is alwasy removed on completion. *)
 val with_tmp_dir: (string -> 'a) -> 'a
+
+(** [in_tmp_dir fn] executes [fn] in a temporary directory. *)
+val in_tmp_dir: (unit -> 'a) -> 'a
 
 (** Runs a job with a temp dir that is cleaned up afterwards *)
 val with_tmp_dir_job: (string -> 'a OpamProcess.job) -> 'a OpamProcess.job
--- a/master_changes.md
+++ b/master_changes.md
@@ -20,7 +20,7 @@ users)
   *
 
 ## Init
-  *
+  * Run the sandbox check in the temporary directory [#4787 @dra27 - fix #4783]
 
 ## Config report
   *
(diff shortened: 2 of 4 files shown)


rjbou (approved):

dra27:
Revised: we're supposed to be avoiding `/tmp`, so making it the rw CWD seems a bit weird! `OpamSystem.in_tmp_dir` added which does what `OpamSystem.with_tmp_dir` claimed to do - it creates a temporary directory and then actually calls the function in it.

As a result, it's possible to simplify the sandbox check, as `OpamSystem.in_tmp_dir` also cleans up the directory, so I changed the check to use `tee` instead. I also verified that it fails by changing the path to `/tmp/check-write` 🙂

kit-ty-kate:
The underlying issue in https://github.com/ocaml/opam/issues/4783 is that `/media` is not mounted by the sandbox
I think a better fix would be to simply fix the sandbox to mount `/` readonly instead of selecting which directory need to be ro-mounted.

This would also make the linux sandbox match the macOS sandbox (which does not have this distinction).
I believe this would also fix https://github.com/ocaml/opam/issues/4751 at the same time.

dra27:
That's the underlying _cause_, it's not the underlying _issue_. Given that opam is doing nothing in the CWD, it definitely has no business at all mounting it in the sandbox, much less mounting it RW (when the sandbox is called normally, the CWD is already the build directory in an opam root). See also the concern for handling of symlinks in `add_mounts` vs `add_sys_mounts` in the sandbox script.

In #4751, the problem is that _opam root_ is not accessible, which does still need fixing.

src/core/opamSystem.mli
```diff
@@ -31,9 +31,13 @@ exception Internal_error of string
 (** Raise [Internal_error] *)
 val internal_error: ('a, unit, string, 'b) format4 -> 'a
 
-(** [with_tmp_dir fn] executes [fn] in a tempory directory *)
+(** [with_tmp_dir fn] executes [fn] creates a temporary directory and
+    passes its name to [fn]. The directory is alwasy removed on completion. *)
 val with_tmp_dir: (string -> 'a) -> 'a
 
+(** [in_tmp_dir fn] executes [fn] in a temporary directory. *)
+val in_tmp_dir: (unit -> 'a) -> 'a
```
rjbou, line 39: TIny thing, wondering if it should takes unit as argument or the temporary created directory name ?
```suggestion
val in_tmp_dir: (string -> 'a) -> 'a
```

rjbou:
double approval : checked, working perfectly. Thanks!

src/core/opamSystem.mli
dra27: I wondered, but then I'd have had to do more than just use `in_dir` (and I didn't need the directory name) so I figured that if a function did need the actual directory then it could use `Sys.getcwd` 🙂

AltGr (approved):

rjbou:
Thanks!

PR #4789, opened 2021-08-07
opamSystem: avoid calling Unix.environment at top level

hannesm:
this is #4731 rebased onto master.


kit-ty-kate (approved):

rjbou:
Thanks!

rjbou:
For discussion, see #4731

Issue #4759, opened 2021-07-21
opam 2.1 and virtual packages


kit-ty-kate:
The situation with virtual packages isn't consistent accross distributions.
- Debian-based distributions' are supported since https://github.com/ocaml/opam/pull/4269 and provide a sometimes vital role (e.g. providing a common package name for a package with several versions or several variants)
- Some distributions do not have any (as far as i know at least): e.g. alpine, …
- For some other distribution, virtual packages are not supported: e.g. RHEL-based distributions (incl. OpenSUSE), Homebrew (see #4548), Archlinux (see https://github.com/ocaml/opam-repository/issues/18540), …
- Some distributions have a mixed story where their versions of that are usually simply aliases and are somewhat supported depending on the distribution. e.g. Gentoo, FreeBSD, (maybe homebrew is in this category?), …

The issue can be annoying for distributions that rely on them heavily (e.g. Fedora) and where opam 2.1 also supports the detection of available packages (see ~sys_available) in https://github.com/ocaml/opam/blob/17adc3b7670a7707a7b2ca15d8e9065f1642ee86/src/state/opamSysInteract.ml#L134

kit-ty-kate:
xref https://github.com/ocaml/opam-repository/pull/19102

kit-ty-kate:
Nevermind, archlinux have them too: https://github.com/ocaml/opam-repository/issues/18540

ddeclerck:
Following our discussion in [ocaml/opam-repository#22548](https://github.com/ocaml/opam-repository/pull/22548) :
on OpenSUSE, you can use the options `--provides` (with `--match-exact`) to search for a virtual package, or simply `zypper what-provides` ; however this requires to actually specify a virtual package name in the query, while from my understanding opam calls zypper without any package name, so as to list all available and installed packages. I'm still investigating.

PR #4791, opened 2021-08-08
depexts: Disable the detection of available packages on RHEL-based distributions

kit-ty-kate:
#4790 shows that `yum list` is an unreliable command to use and no options seems to be available to make it so.
#4759 also shows that RHEL-based distributions have virtual packages that opam is not taking into account.

For these two reasons I propose we simply remove the use of `yum list` in favour of `rpm -qa` which seems way more reliable and leaps faster (~0.1s vs. ~2s!!).

Sadly we incidently disable the detection of available packages and thus revert to the behaviour of the opam-depext plugin.
IMO this is not a big deal as only a few distributions that we support actually have it disabled as well due to the method to acquire the data being either incomplete or too tedious/unreliable. This is the case currently (in 2.1.0) for FreeBSD, Gentoo, Homebrew, NetBSD and OpenBSD

### DIFF
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -286,30 +286,16 @@ let packages_status packages =
     in
     compute_sets sys_installed ~sys_available
   | Centos ->
-    (* XXX /!\ only checked on centos XXX *)
-    let lines = run_query_command "yum" ["-q"; "-C"; "list"] in
-    (* -C to retrieve from cache, no update but still quite long, 1,5 sec *)
-    (* Return a list of installed packages then available ones:
-       >Installed Packages
-       >foo.arch    version   repo
-       >Available Packages
-       >bar.arch    version   repo
+    (* Output format:
+       >crypto-policies
+       >python3-pip-wheel
     *)
-    let sys_installed, sys_available, _ =
-      List.fold_left (fun (inst,avail,part) -> function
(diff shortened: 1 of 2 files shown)


bikallem:
I just tried the `fix-2` branch and it seems to work, i.e. `opam update`, `opam install` seems to work correctly even in the presence of a corrupted yum repository.

src/state/opamSysInteract.ml
```diff
@@ -286,30 +286,9 @@ let packages_status packages =
...
-            (match part, OpamStd.String.split l '.' with
-             | `installed, pkg::_ ->
-               pkg +++ inst, avail, part
-             | `available, pkg::_ ->
-               inst, pkg +++ avail, part
-             | _ -> (* shouldn't happen *) inst, avail, part))
-        OpamSysPkg.Set.(empty, empty, `preamble) lines
-    in
-    compute_sets sys_installed ~sys_available
+    let lines = run_query_command "rpm" ["-qa"; "--qf"; "%{NAME}\\n"] in
```
rjbou, line 289: Could you provide an output format example, is the same form than the other ones in the file (`>` are used as bol, to highlight spaces)

rjbou:
Good to go for me (with comment), it just need to be updated on github action depext test once finished.

rjbou:
Thanks!

PR #4795, opened 2021-08-11
Sandbox: Always mount every directories under / (gets rid of OPAM_USER_PATH_RO)

kit-ty-kate:
Alternative fix for https://github.com/ocaml/opam/issues/4783 and https://github.com/ocaml/opam/issues/4751.

This also helps make Linux and macOS sandbox behave the same: by default `sandbox-exec` mounts the whole file system (including external disks) readonly and the `OPAM_USER_PATH_RO` related code did not do anything.

I don't think there is any need to restrict access to non-standard directories since it is all read-only.
This also helps users on NixOS for instance as well as people using several disks/mountpoints to host opam as highlighted by the two issues above.


rjbou (approved):

src/state/shellscripts/bwrap.sh
```diff
@@ -53,11 +54,21 @@ add_sys_mounts() {
...
 
-# remove some unusual paths (/nix/stored and /rw/usrlocal )
-# use OPAM_USER_PATH_RO variable to add them
-# the OPAM_USER_PATH_RO format is the same as PATH
-# ie: export OPAM_USER_PATH_RO=/nix/store:/rw/usrlocal
-add_sys_mounts /usr /bin /lib /lib32 /lib64 /etc /opt /home /var /tmp
+# NOTE: The list of moint-points to avoid is always to be sync'd with the list at the top of the file
+# It is due to a limitation of bubblewrap that we have to mount the directories one by one
+# See https://github.com/containers/bubblewrap/issues/413
+for dir in $(ls -a /); do
```
AltGr, line 60: Using `ls` in shell scripts is generally discouraged, there may be lots of output options enabled locally.
I would go with something like:
```bash
for dir in /*; do
    if [ ! -d "$dir" ]; then continue; fi
    case "$dir" in
        /proc|/dev|/run|/opam-tmp) ;;
        *) add_sys_mounts "$dir";;
    esac
done
```

The difference is that this wouldn't mount hidden root directories... could that be an issue ?

AltGr (commented):
This will probably be better and easier to maintain in the long run :+1: 

Do we want to leave access to `/sys` ? (Real question, I am not sure of the implications)

src/state/shellscripts/bwrap.sh
kit-ty-kate: > there may be lots of output options enabled locally

Unless the user is purposfully using `source sandbox-script.sh` or called using `bash -l`/other non-interactive/login options, user changes (`/etc/profile`, `~/.bashrc`, …) are disabled when calling scripts so this should never happen.

kit-ty-kate: Actually I a solution to get the best of both world (use `/*` and get the hidden files): `shopt -s dotglob`.
I'll fix that right now.

kit-ty-kate:
> Do we want to leave access to `/sys` ? (Real question, I am not sure of the implications)

No idea either but that's a fair point and I think since it was also disabled before, it's more prudent to keep it that way. Thanks!

All the concerns and comments have been addressed I think. Everything should be green.

PR #4796, opened 2021-08-12
Make sure that `--best-effort` only installs root package versions that where requested

LasseBlaauwbroek:
**NOTE**: This PR is almost certainly wrong. Somebody knowledgeable should take a look

When calling `opam install --best-effort my-package.version`, opam creates a cudf file, in which `my-package` gets assigned the property `opam-query`. The solver then optimizes to install as many packages marked as `opam-query` as possible. However, Opam marks every version of `my-package` with `opam-query`, even though we request a specific version. Therefore, sometimes Opam installs versions that have not been requested. This behavior can be observed with
```
opam install coq-graph-theory.0.8 --best-effort
```
In this case, Opam will propose to install `coq-graph-theory.0.7`, which is wrong.

This PR attempts to fix this by only marking packages that actually satisfy the request. There are a lot of side effects to this change, and I'm sure that I resolved some of them wrong. Please suggest how to do this better.


src/client/opamClient.ml
```diff
@@ -38,7 +38,7 @@ let orphans ?changes ?(transitive=false) t =
   let allnames = OpamPackage.names_of_packages all in
   let invalidated = Lazy.force t.invalidated in
   let universe =
-    OpamSwitchState.universe t ~requested:OpamPackage.Name.Set.empty Reinstall
+    OpamSwitchState.universe t ~requested:[] Reinstall
```
kit-ty-kate, line 41: I'm not too~ knowledgable but it seems to me that `OpamPackage.Set` is a better choice to replace `OpamPackage.Name.Set` (in place of a list). Could you change it that way?

LasseBlaauwbroek: How about this?

kit-ty-kate: that looks like what i had in mind. Thanks a lot!

src/client/opamClient.ml
```diff
@@ -1449,7 +1451,7 @@ let reinstall_t t ?ask ?(force=false) ~assume_built atoms =
   let t, full_orphans, orphan_versions = check_conflicts t atoms in
 
   let requested =
-    OpamPackage.Name.Set.of_list (List.rev_map fst atoms) in
```
AltGr, line 1452: `rev_map` was only here because we don't care about order, no need to `List.rev` we are converting to a set anyway :)

src/client/opamClient.ml
```diff
@@ -169,6 +169,7 @@ let update_dev_packages_t ?(only_installed=false) atoms t =
 let compute_upgrade_t
     ?(strict_upgrade=true) ?(auto_install=false) ?(only_installed=false)
     ~all atoms t =
+  let packages = OpamFormula.packages_of_atoms t.packages (List.rev atoms) in
```
AltGr, line 172: No need to `List.rev`, see my comment below

src/client/opamClient.ml
```diff
@@ -1120,6 +1122,7 @@ let filter_unpinned_locally t atoms f =
 let install_t t ?ask ?(ignore_conflicts=false) ?(depext_only=false)
     ?(download_only=false) atoms add_to_roots ~deps_only ~assume_built =
   log "INSTALL %a" (slog OpamFormula.string_of_atoms) atoms;
+  let packages = OpamFormula.packages_of_atoms t.packages (List.rev atoms) in
```
AltGr, line 1125: same here

src/client/opamSwitchCommand.ml
```diff
@@ -483,7 +483,7 @@ let import_t ?ask importfile t =
     let add_roots = OpamPackage.names_of_packages import_sel.sel_roots in
 
     OpamSolution.resolve_and_apply ?ask t Import
-      ~requested:(OpamPackage.Name.Set.of_list @@ List.map fst to_import)
+      ~requested:(OpamPackage.Set.union (to_install %% available) unavailable_version)
```
AltGr, line 486:
```suggestion
      ~requested:((to_install %% available) ++ unavailable_version)
```

src/state/opamSwitchState.ml
```diff
@@ -849,8 +849,9 @@ let universe st
     ?reinstall
     ~requested
     user_action =
+  let names = OpamPackage.names_of_packages requested in
   let requested_allpkgs =
-    OpamPackage.packages_of_names st.packages requested
+    OpamPackage.packages_of_names st.packages names
```
AltGr, line 854: This is patched in a reasonable and conservative way...

But this makes me wonder: it would seem consistent to use `requested` here directly instead of `requested_allpkgs`. This would potentially change the semantics at some point, but could be expected to lead to the same results...

A visible change I could imagine is when you change the installed version of a package, _e.g._ with `opam install foo.2` when `foo.1` is installed: in the current implementation (and behaviour preserved by the proposed patch), `foo.1` is a member of `requested_allpkgs`. This affects:
- the values of `with-test`, `with-doc`: this is harmless, the action on `foo.1` can only be a removal
- the triggering of pending reinstallation within the dependency cone of `foo.1`

The second point _could_ have visible effects in some cases: the idea is not to trigger pending recompilations (see [`opam reinstall --list-pending`](http://opam.ocaml.org/doc/man/opam-reinstall.html)) if you `opam install` something unrelated. If `foo.2` depends on the package needing reinstallation, opam definitely needs to do that reinstallation before compiling it ; but if it's only `foo.1`, that will just be removed, I guess it can be skipped.

All well considered, I think this can be simplified by just using `requested` everywhere instead of `requested_allpkgs` :)

AltGr (commented):
Good job, thanks!

Please see my additional comments, but this is OK to merge.

rjbou:
ping @LasseBlaauwbroek

LasseBlaauwbroek:
Sorry for the delay, my attention moved away for this area for a while. I expect I'll work in the Opam area again in about a month. Unless you need this urgently, I'd like to fix this then. Or, if you like, you can take over this PR.

rjbou:
Thanks for the update!

src/client/opamClient.ml
rjbou:
```suggestion
  let packages = OpamFormula.packages_of_atoms t.packages atoms in
```

rjbou:
Rebased

rjbou:
Thanks @LasseBlaauwbroek !

LasseBlaauwbroek:
Thanks for finishing this @rjbou, and sorry that I couldn't get to this in a reasonable amount of time.

PR #4797, opened 2021-08-12
Relax the macOS sandbox (fixes #4389)

### LINKED ISSUE #4389: sandbox too strict on macos
The sandbox script on macos denies to properly run various commands in the test phase, as an example `security find-certificate -a -p /System/Library/Keychains/SystemRootCertificates.keychain` does not output the trust anchors as expected -- observed in https://travis-ci.org/github/ocaml/opam-repository/jobs/734785206

How to reproduce: on a macos system, `opam pin -n ca-certs https://github.com/mirage/ca-certs.git#v0.1.1 && opam install -t ca-certs`. The expected result is that all tests are passing (which is the case if you execute `dune runtest` from the above repository). The observed result are failures since the above mentioned command executed successfully (exit status = 0), but did not produce any output.

kit-ty-kate:
This is an attempt at fixing https://github.com/ocaml/opam/issues/4389
I've used `dtrace` on `/usr/bin/security` but I still can't find the exact point where something is written in that directory.
It looks related to something in `/private/var/db/mds`, probably a symlink to `/private/var/folders`.

In any case i've looked around and realized that macports had similar things so maybe it makes sense(?). They also have other things but I'm not sure we should allow even more: https://github.com/macports/macports-base/blob/2c6fc24ddd1d6961afa83c5b35be12224b6850f6/src/port1.0/portsandbox.tcl#L92

All in all I have no idea what this directory is supposed to be in the context of `/usr/bin/security`. Their commit adding it doesn't say anything either: https://github.com/macports/macports-base/commit/e3eceea305409fe06787242b4c498e6300f5a9b9

Several blog posts in the wild are also trying to understand what's up with this directory:
* http://www.magnusviri.com/what-is-var-folders.html
* https://www.swiftforensics.com/2017/04/the-mystery-of-varfolders-on-osx.html

`hier(7)` says it's "per-user temporary files and caches", so allowing writes like macports does doesn't seem right to me.

However, I found out (late into writing this PR even ^^") that if we forbid reads into that directory, `/usr/bin/security` won't try to write in it and still succeeds.
This is most likely a bug in macOS itself but there we are!

cc @hannesm


AltGr (approved):

dra27:
What's up with the macOS test?

kit-ty-kate:
I've sent a bug report to Apple btw.

rjbou:
> What's up with the macOS test?

Found it!
```
+ ${BASEDIR}/OPAM/opam-init/hooks/sandbox.sh "build" "sh" "-c" "echo SUCCESS | tee check-write"
- shell-init: error retrieving current directory: getcwd: cannot access parent directories: Operation not permitted
- SUCCESS
```

rjbou:
So sandbox check is failing, and it is removed from config file.

rjbou:
From tests, seems that this changes disable sandbox.

kit-ty-kate:
Well, time to unearth this and do an emergency point release tomorrow i guess https://discuss.ocaml.org/t/cant-install-async-0-17-0-on-macos-15-4/16468/3

The current version of the patch doesn't work but i'll push the one that does.

kit-ty-kate:
Coping the new commit description:

    Allow the macOS sandbox to write in the `/var/folders/` and `/var/db/mds/` directories
    
    It is required by some of macOS core tools such as security(1) and
    previous TMPDIR stored in /var/folders/*/*/T/ have been seen accessed by
    xcode in the wild on occasion.
    
    While the /var/folders/*/*/C/ directory contain things that are kept
    indefinitely, these files are in theory understood by applications to be
    writeable by anyone and thus not trusted.
    
    The mds subdirectory is a cache used to search for files in the system.
    
    Both of these are part of other tools that use sandbox-exec such as
    MacPorts. See for example:
    
    https://github.com/macports/macports-base/blob/2c6fc24ddd1d6961afa83c5b35be12224b6850f6/src/port1.0/portsandbox.tcl#L92

kit-ty-kate:
The ocaml-benchmark failure comes from https://github.com/ocaml/infrastructure/issues/179. Ignoring.

PR #4801, opened 2021-08-13
depexts/brew: Add support for casks and full-names

### LINKED ISSUE #4800: Opam keep reinstalling conf-libX11
Hello,

Since the upgrade to opam 2.1, each time I run `opam upgrade`, it wants to reinstall conf-libX11.

It seems there is a problem in the check of the presence of xquartz (I am on a mac). Because it is already installed:

```
brew install xquartz
Updating Homebrew...
==> Auto-updated Homebrew!
Updated 1 tap (homebrew/cask).
==> Updated Casks
Updated 1 cask.

Warning: Cask 'xquartz' is already installed.

To re-install xquartz, run:
  brew reinstall xquartz
```
And opam still want me to install it.

I tried reinstalling xquartz but this does not solve the problem.

`opam config report`
```
# opam config report
# opam-version         2.1.0
# self-upgrade         no
# system               arch=x86_64 os=macos os-distribution=homebrew os-version=10.14.6
# jobs                 7
# repositories         2 (http) (default repo at 12141eb0)
# pinned               0
# current-switch       4.10.1
# ocaml:native         true
# ocaml:native-tools   true
# ocaml:native-dynlink true
# ocaml:stubsdir       /Users/fdagnat/.opam/4.10.1/lib/ocaml/stublibs:/Users/fdagnat/.opam/4.10.1/lib/ocaml
# ocaml:preinstalled   false
# ocaml:compiler       4.10.1
```

kit-ty-kate:
Fixes #4800

### DIFF
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -427,6 +427,7 @@ let packages_status packages =
        exampe output
        >openssl@1.1
        >bmake
+       >koekeishiya/formulae/skhd
     *)
     let sys_installed =
       run_query_command "brew" ["list"; "--full-name"]
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -429,12 +429,18 @@ let packages_status packages =
        >bmake
     *)
     let sys_installed =
-      run_query_command "brew" ["list"; "--formula"]
+      run_query_command "brew" ["list"; "--full-name"]
       |> List.fold_left (fun res s ->
           List.fold_left (fun res spkg ->
+              let parse_fullname pkg =
+                match List.rev (String.split_on_char '/' pkg) with
+                | [] -> assert false (* split_on_char is guaranteed to never return [] *)
+                | [pkg] -> [pkg]
+                | simple_name::_ -> [pkg; simple_name]
+              in
               match OpamStd.String.cut_at spkg '@' with
-              | Some (n,_v) -> n::spkg::res
-              | None -> spkg::res)
+              | Some (n,_v) -> parse_fullname n@parse_fullname spkg@res
+              | None -> parse_fullname spkg@res)
             res (OpamStd.String.split s ' ')) []
       |> List.map OpamSysPkg.of_string
       |> OpamSysPkg.Set.of_list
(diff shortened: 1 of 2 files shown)


fdagnat:
I would gladly test your patch but I do not how to do it. From documentation, I would infer that I need a magic invocation close to `opam pin add opam-devel --dev` but I suspect this one only upgrade to the trunk version.

Could you give me the spell please?

Thank you

kit-ty-kate:
```
cd /tmp && git clone git://github.com/ocaml/opam.git && cd opam && make cold
```
The binary is then available as `./opam` (which is a symlink to `_build/default/src/client/opamMain.exe`)
This won't require you to install any dependencies or do anything with the opam already installed.

fdagnat:
Ouch, after a long compilation time I get an error:

```
File "_none_", line 1:   
Error: The external function `unix_spawn' is not available
make[1]: *** [build-opam] Error 1
make: *** [cold] Error 2
```

kit-ty-kate:
(oops i forgot to get my branch in the command above, I've updated it)

uh, that a weird one :/ I'm not able to produce this on an macOS/arm64. Could you give a bit more of log? (the full one would be good, but at least the last command to see where we are)

fdagnat:
I had already changed the git repo...

[log.txt](https://github.com/ocaml/opam/files/6983527/log.txt)

kit-ty-kate:
Ah I see the issue. I was able to reproduce locally, thanks for your help. It's an issue with the cold target when you also have bits of older compiler (in this case your 4.10.1 switch).

You can take it back from where you left it with:
```
env PATH="$(pwd)/bootstrap/ocaml/bin:$PATH" CAML_LD_LIBRARY_PATH= make
```

fdagnat:
Hello,

Sorry for the week end break. The compilation finished gracefully using your command. The compiled opam does not ask to install xquartz. So it seems to be solved.

Thank you

src/state/opamSysInteract.ml
```diff
@@ -429,12 +429,18 @@ let packages_status packages =
        >bmake
     *)
     let sys_installed =
-      run_query_command "brew" ["list"; "--formula"]
+      run_query_command "brew" ["list"; "--full-name"]
```
rjbou, line 432: Can you add/update the example output above

rjbou (approved):

src/state/opamSysInteract.ml
kit-ty-kate: it is up-to-date

kit-ty-kate: ok, thinking about it maybe i can add the `path/to/cask` form too

kit-ty-kate: Done

rjbou:
Thanks!

PR #4803, opened 2021-08-14
Make sure OpamListCommand.filter Depends_on recursive only returns a subset of base

LasseBlaauwbroek:
It seems that `OpamListCommand.filter ~base:packages` is supposed to only return a subset of the supplied packages. For recursive dependency search, this does not happen, however. As far as I can tell, this is not visible from the command line, because supplied patterns are filtered separately. However, for people who use Opam as a library, this is still unexpected behavior.


kit-ty-kate:
Fixed in https://github.com/ocaml/opam/pull/4999 with some added performance improvements. Thanks though!

Issue #4446, opened 2020-11-24
opam list --recursive --depends-on <pkg> should not include the package itself


kit-ty-kate:
```
$ opam list --recursive --depends-on labrys.0.1 --all-versions -s
labrys.0.1
```
This seems a bit counter-intuitive.

AltGr:
Indeed; the same is true for `--required-by`

PR #4804, opened 2021-08-14
Make sure that recursive dependencies are not reflexive

LasseBlaauwbroek:
`opam list --recursive --recursive --required-by=package` is reflexive (it shows the package itself). This does not seem correct, since a package does not require itself. Note that without `--recursive`, the command is not reflexive.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -38,7 +38,7 @@ users)
   *
 
 ## List
-  *
+  * Do not list the package itself when running `--depends-on` and `--required-by` with `--required` [#4804 @LasseBlaauwbroek]
 
 ## Show
   *
--- a/src/client/opamListCommand.ml
+++ b/src/client/opamListCommand.ml
@@ -226,10 +226,11 @@ let apply_selector ~base st = function
       | Depends_on _ -> OpamSolver.reverse_dependencies
       | _ -> assert false
     in
-    deps_fun ~depopts:tog.depopts ~build:tog.build ~post:tog.post
+    let packages = packages_of_atoms st atoms in
+    let deps = deps_fun ~depopts:tog.depopts ~build:tog.build ~post:tog.post
       ~installed:false ~unavailable:true
-      (get_universe st tog)
-      (packages_of_atoms st atoms)
+      (get_universe st tog) packages in
+    OpamPackage.Set.diff deps packages
   | Required_by (tog, atoms) ->
     atom_dependencies st tog atoms |>
     OpamFormula.packages base


kit-ty-kate:
Thanks! fixes #4446

kit-ty-kate:
The tests fail and it might be related. Could you have a look at it? e.g.
```
diff --git a/tests/reftests/legacy-git.test b/tests/reftests/legacy-git.out
index a629f71..63d38d2 100644
--- a/tests/reftests/legacy-git.test
+++ b/tests/reftests/legacy-git.out
@@ -747,20 +747,16 @@ Done.
 ocaml.system
 P1.1
 ### opam remove P1
-The following actions will be performed:
-  - remove P1 1
-
--> removed   P1.1
-Done.
+Nothing to do.
 ### opam list -is --columns=package
 ocaml.system
+P1.1
 ### : LIST :
 ### opam list -A
 # Packages matching: any
 # Name # Installed # Synopsis
 ocaml  system
-P1     --          A very useful package
+P1     1           A very useful package
 P2     --          An other very useful package
 P3     --          Testing version names
 P4     --          Testing transitive closure
 ```

LasseBlaauwbroek:
My guess is that some functions rely on the reflexive behavior (for example, in order to calculate package removal). I changed it so that the non-reflexive behavior only occurs in the `list` command.

kit-ty-kate (approved):

AltGr:
Thanks! Indeed this was inconsistent.

A note of caution however: this is incorrect with multiple packages, e.g. `--required-by=a,b` should still return `a` if `b` depends on `a`.

LasseBlaauwbroek:
I guess this is true, although it depends a bit on how one reads the documentation of `--required-by` and `--depends-on`.

The only way I can think of fixing this is to run a separate query for each of the requested atoms and then take the union of the results. Would that be acceptable, or is that too slow?

AltGr:
> I guess this is true, although it depends a bit on how one reads the documentation of --required-by and --depends-on.

Hm, that doc probably needs some effort then. The idea is that `--depends foo,bar` is not the same as `--depends foo --depends bar`, because the latter wouldn't need an extra syntax.

It's not a very common pattern, so I guess the solution you propose would be fine, dependency queries are not that expensive; and I can't think of an other one :)

kit-ty-kate:
Superseded by https://github.com/ocaml/opam/pull/5075 which supports the `--required-by=a,b` case.
Thanks a lot for getting the ball rolling!

PR #4805, opened 2021-08-14
Remove memoization from `OpamSolverConfig.best_effort`

LasseBlaauwbroek:
Because this function is memoized, it becomes impossible to change the
best-effort setting after the function has been called. The ability to change it
is important for library users that want to make multiple solver calls with
different settings.

The performance implications of this should be minimal, because normally this
function is only called once, and if best_effort is disabled, the `check ()`
function will be short-circuited.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -174,6 +174,7 @@ users)
   * [BUG] handle converted variables correctly when no_undef_expand is true [#4811 @timbertson]
   * [BUG] check Unix.has_symlink before using Unix.symlink [#4962 @jonahbeckford]
   * OpamCudf: provide machine-readable information on conflicts caused by cycles [#4039 @gasche]
+  * Remove memoization from `best_effort ()` to allow for multiple different settings during the same session (useful for libaray users) [#4805 @LasseBlaauwbroek]
 
 ## Test
   * Update crowbar with compare functions [#4918 @rjbou]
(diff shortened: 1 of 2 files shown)


kit-ty-kate (approved):

AltGr (commented):
Indeed, I see the problem; I would keep the `lazy` though, and just move the `    !r.best_effort &&` outside of its body and into the function, if that still works for you ?

The purpose would just be to avoid multiple printings of the warning.

kit-ty-kate:
> Indeed, I see the problem; I would keep the `lazy` though, and just move the ` !r.best_effort &&` outside of its body and into the function, if that still works for you ?
> 
> The purpose would just be to avoid multiple printings of the warning.

We can’t keep the lazy because the whole computation relies on config variables that can also change.
However I did the next best thing and used a boolean reference to avoid the multiple printing of warnings.

Good to go once CI is green

kit-ty-kate:
Oops, nevermind I just realized they were all lazy values to begin with. Changing this in a second

kit-ty-kate:
> Oops, nevermind I just realized they were all lazy values to begin with. Changing this in a second

Actually nevermind again... They’re references on lazy values so they can still change.

AltGr:
ah what I had in mind was simpler:
```ocaml
let best_effort =
  let r = lazy ( ... ) in
  fun () -> !r.best_effort && Lazy.force r
```
but indeed your version will also work if the function is run a first time with unitialised best-effort criteria; it's not supposed to happen but who knows what lib-users can do :)

kit-ty-kate:
Thanks!

PR #4815, opened 2021-08-30
Support LOCALAPPDATA as HOME fallback for Windows

jonahbeckford:
Splitting up https://github.com/ocaml/opam/pull/4813

> For a24d269 ("Support USERPROFILE as HOME fallback for Windows"), there's prior art in dra27@ed44880. The choice of directory (user documents) in that one is bad - in fact, my current intention is to address this "properly" as part of Support ~/.config/opam (per XDG base dir spec) #3766 in opam 2.2. My issue with falling back to %USERPROFILE% is that it's not a directory which we should be writing in (my enterprise setup, for example, uses both redirected folders and roaming profiles).

For continuity this initial PR is as-is from the original PR, but I'll make a change to address the valid enterprise concern.

### DIFF
--- a/src/state/opamStateConfig.ml
+++ b/src/state/opamStateConfig.ml
@@ -68,7 +68,11 @@ type t = {
 
 let default = {
   root_dir = OpamFilename.(
-      concat_and_resolve (Dir.of_string (OpamStd.Sys.home ())) ".opam"
+      if Sys.win32 then
+        (* CSIDL_LOCAL_APPDATA = 0x1c *)
+        concat_and_resolve (Dir.of_string (OpamStubs.(shGetFolderPath 0x1c SHGFP_TYPE_CURRENT))) "opam"
+      else
+        concat_and_resolve (Dir.of_string (OpamStd.Sys.home ())) ".opam"
     );
   current_switch = None;
   switch_from = `Default;


dra27:
Thanks! I _think_ the best default is `LOCALAPPDATA` (so `opam init` into `%LOCALAPPDATA%\.opam` by default), but I'm open to other suggestions.

jonahbeckford:
Definitely agree with using LOCALAPPDATA. Switched it:

```
dev-Debug@DESKTOP-CV21S6H:/z/source/opam$ env HOME= OPAMROOT= OPAMSWITCH= OPAM_SWITCH_PREFIX= \
   dune exec src/client/opamMain.exe -- config report --verbose --debug-level 2
00:00.015  CLI                    Parsing CLI version 2.2
# opam config report
# opam-version         2.2.0~alpha~dev (7e69fb4f4e2cbdb6e295dd439e331592b2c6b929)
# self-upgrade         no
# system               arch=x86_64 os=win32 os-distribution=win32 os-version=10.0.22000
00:00.025  GSTATE                 LOAD-GLOBAL-STATE @ C:\Users\beckf\AppData\Local\.opam
[ERROR] Opam has not been initialised, please run `opam init'
# read-state           OpamStd.OpamSys.Exit(50)
```

dra27:
Hmm, I've had a more thorough think and check (and paged back in some of my very old and not yet merged patches too!). `OpamStd.Sys.home` is used in:
 - `OpamFilename.Dir.of_string` to interpret the meaning of `"~\"`
 - `OpamFilename.prettify_path` and `OpamProcess.make_info` to display instances of `%HOME%` as `~`
 - `OpamStd.Sys.guess_dot_profile` to locate the shell profile files
 - `OpamPath.init_config_files` to locate `.opamrc`
 - `OpamStateConfig.default` for the default `root_dir`

I fear I've led you down the garden path with the `LOCALAPPDATA` part as that is correct for `OpamStateConfig.default` case, but not good for everything else!

As I said before, we're intending to support XDG properly in opam 2.2 which should sort out the opam root part. Switching to `LOCALAPPDATA` will also break the existing users of diskuv.

So, with apologies, I think your original suggestion of `USERPROFILE` is correct! 🙂 That said, I think writing to `%USERPROFILE%\.opam` is unfortunate - it's up to you before opam 2.2 is released if you want to have a Windows-specific check in `OpamStateConfig.default` to check if `HOME` is actually defined and specifically use `LOCALAPPDATA` there. This patch would do that, but I don't want to merge that in opam right now, as it conflicts with the XDG plans:

```diff
diff --git a/src/state/opamStateConfig.ml b/src/state/opamStateConfig.ml
index 765652764..31890c335 100644
--- a/src/state/opamStateConfig.ml
+++ b/src/state/opamStateConfig.ml
@@ -68,7 +68,14 @@ type t = {

 let default = {
   root_dir = OpamFilename.(
-      concat_and_resolve (Dir.of_string (OpamStd.Sys.home ())) ".opam"
+      let dir =
+        if Sys.win32 && OpamStd.Env.getopt "HOME" = None then
+          (* CSIDL_LOCAL_APPDATA = 0x1c *)
+          Dir.of_string (OpamStubs.(shGetFolderPath 0x1c SHGFP_TYPE_CURRENT))
+        else
+          Dir.of_string (OpamStd.Sys.home ())
+      in
+      concat_and_resolve  ".opam"
     );
   current_switch = None;
   switch_from = `Default;
```

(for the local app data case, it could even be improved further by writing to `%LOCALAPPDATA%\opam` (i.e. no Unix-like `.`))

jonahbeckford:
Ah I see.

I don't want the Diskuv OCaml distribution to be an island; that would cause Windows developers to have a poor experience interoperating with other Windows distributions, and force continual patches for OCaml packages that rely on the 2.2+ "standard" way. So it is better to take the backwards-incompatible hit on OPAMROOT right now.

What I'll do is:
* make a diskuv-only patch so that `root_dir` appears as `%LOCALAPPDATA%\opam`
* abandon this specific patch
* communicate now that a breaking change is coming for standards compliance

All that is required is that we agree that `%LOCALAPPDATA%\opam` will be the `root_dir` in some future version of Opam. Can we agree on `root_dir`?

(I'm glad this is getting teased out early!)

dra27:
Yes - let's definitely converge on `<LOCALAPPDATA>\opam` as the preferred location. I think it's better that on Windows `<LOCALAPPDATA>\opam` is the default (my patch above does that only if `%HOME%` is undefined), but that creates a potential issue for all previous versions. However, the changes to support XDG mean that the default opam root will actually the first directory found of:

- `<LOCALAPPDATA>\opam` (native Windows only)
- `$XDG_CONFIG_HOME/opam` (everything except Windows)
- `$HOME/.opam` (all platforms)

But if no root is found, then `opam init` will always _create_ the first one (i.e. `<LOCALAPPDATA>\opam` or `$XDG_CONFIG_HOME/opam`). That scheme retains compatibility with any users who have already init'd with `%HOME%\.opam` - it just requires a bit more care in `OpamClient.init` when computing `root` (`root` and `root_empty` will want to be computed simultaneously)

jonahbeckford:
Thanks for figuring that all out! I've updated the PR with the code I'll be using in the Windows distribution until the future Opam version is ready; it is based off of your patch.

I'll close/abandon this PR once I've given people a chance to comment in the https://discuss.ocaml.org/ thread; no more than a day or two.

PR #4817, opened 2021-08-30
Fallback to copying on symlink-less MSYS2

jonahbeckford:
Splitting up https://github.com/ocaml/opam/pull/4813:

> For 92397ac ("Allow MSYS2 to prefer copying over symlinking") I think the code duplication can be reduced if is_cygwin_variant and is_msys2_variant are merged into a single function get_windows_executable_variant : string -> [ `Native | `Cygwin of [`Cygwin | `Msys2] | `Tainted of [`Cygwin | `Msys2] ]. The old is_cygwin_variant and the new is_msys2_variant can then be defined in terms of that function. While it's being added, it would be good to thread the tainting support through to OpamSystem.install (the point of `CygLinked is to warn for the situation where you've accidentally linked with a Cygwin or MSYS2 compiled DLL which usually implies bindings are wrong somewhere). I was initially confused by the title of the commit, as we must definitely prefer symlinking over copying (if it's available and works), vs requiring symlinking (I completely agree with you that requiring developer mode is an unacceptable barrier), but I think that the actual rsync change works correctly if native symlinking is enabled. The longer-term plan for this (I say longer, it's possible this'll happen in 2.2 or 2.3) is to use ocaml-tar and implement the cp command directly. Cygwin/MSYS2' cp and tar commands are somewhat belligerently staying portable where symlinks are concerned - Windows symlinks do not require the file to exist, what they do require is for the type of the target to be known, which cp does (trivially) and tar can (slightly less-trivially). The only thing which is very hard to copy reliably is a dangling/broken symlink. However, I think your rsync solution is a very good stop-gap!

For continuity this initial PR is as-is from the original PR, except the new title and a change in the commit description:

> Default MSYS2 installations "support" symlinking by copying. This
commit adds a two-phase rsync (copy first without symlinks, then
copy symlinks) so the MSYS2 copying behavior works.

### DIFF
--- a/src/core/opamStd.ml
+++ b/src/core/opamStd.ml
@@ -1004,7 +1004,7 @@ module OpamSys = struct
       (fun f -> try f () with _ -> ())
       !registered_at_exit
 
-  let is_cygwin_variant =
+  let get_windows_executable_variant =
     if Sys.win32 then
       let results = Hashtbl.create 17 in
       let requires_cygwin name =
@@ -1013,19 +1013,26 @@ module OpamSys = struct
         let rec f a =
           match input_line c with
           | x ->
-              (* Treat MSYS2's variant of `cygwin1.dll` called `msys-2.0.dll` equivalently.
-                 Confer https://www.msys2.org/wiki/How-does-MSYS2-differ-from-Cygwin/ *)
-              let tx = String.trim x in
-              if (OpamString.ends_with ~suffix:"cygwin1.dll" tx ||
-                  OpamString.ends_with ~suffix:"msys-2.0.dll" tx) then
-                if OpamString.starts_with ~prefix:"  " x then
-                  f `Cygwin
-                else if a <> `Cygwin then
-                  f `CygLinked
-                else
-                  f a
+            (* Treat MSYS2's variant of `cygwin1.dll` called `msys-2.0.dll` equivalently.
+               Confer https://www.msys2.org/wiki/How-does-MSYS2-differ-from-Cygwin/ *)
+            let tx = String.trim x in
+            if (OpamString.ends_with ~suffix:"cygwin1.dll" tx ||
+                OpamString.ends_with ~suffix:"msys-2.0.dll" tx) then
+              if OpamString.starts_with ~prefix:"  " x then
+                f `Cygwin
+              else if a = `Native then
+                f (`Tainted `Cygwin)
               else
                 f a
+            else if OpamString.ends_with ~suffix:"msys-2.0.dll" tx then
+              if OpamString.starts_with ~prefix:"  " x then
+                f `Msys2
+              else if a = `Native then
+                f (`Tainted `Msys2)
+              else
+                f a
+            else
+              f a
           | exception e ->
               Unix.close_process_full process |> ignore;
               fatal e;
@@ -1046,6 +1053,13 @@ module OpamSys = struct
               result
     else
       fun _ -> `Native
+
+  let is_cygwin_variant cmd =
+    match get_windows_executable_variant cmd with
+    | `Native -> `Native
+    | `Cygwin
+    | `Msys2 -> `Cygwin
+    | `Tainted _ -> `CygLinked
 
   exception Exit of int
   exception Exec of string * string array * string array
--- a/src/core/opamStd.mli
+++ b/src/core/opamStd.mli
@@ -455,12 +455,26 @@ module Sys : sig
   val split_path_variable: ?clean:bool -> string -> string list
 
   (** For native Windows builds, returns [`Cygwin] if the command is a Cygwin-
-      compiled executable, [`CygLinked] if the command links to a library which is
-      itself Cygwin-compiled or [`Native] otherwise.
+      compiled executable, [`Msys2] if the command is a MSYS2-compiled
+      executable, and [`Tainted of [ `Msys2 | `Cygwin ]] if the command links
+      to a library which is itself Cygwin- or MSYS2-compiled, or [`Native]
+      otherwise.
 
       Note that this returns [`Native] on a Cygwin-build of opam!
 
-      Both cygcheck and an unqualified command will be resolved using the current PATH. *)
+      Both cygcheck and an unqualified command will be resolved using the
+      current PATH. *)
+  val get_windows_executable_variant:
+    string -> [ `Native | `Cygwin | `Tainted of [ `Msys2 | `Cygwin] | `Msys2 ]
+
+  (** For native Windows builds, returns [`Cygwin] if the command is a Cygwin-
+      or Msys2- compiled executable, and [`CygLinked] if the command links to a
+      library which is itself Cygwin/Msys2-compiled, or [`Native] otherwise.
+
+      Note that this returns [`Native] on a Cygwin-build of opam!
+
+      Both cygcheck and an unqualified command will be resolved using the
+      current PATH. *)
   val is_cygwin_variant: string -> [ `Native | `Cygwin | `CygLinked ]
 
   (** {3 Exit handling} *)
--- a/src/core/opamSystem.ml
+++ b/src/core/opamSystem.ml
@@ -10,7 +10,8 @@
 (**************************************************************************)
 
 type install_warning =
-  [ `Add_exe | `Install_dll | `Install_script | `Install_unknown | `Cygwin | `Cygwin_libraries ]
+  [ `Add_exe | `Install_dll | `Install_script | `Install_unknown
+  | `Cygwin | `Msys2 | `Tainted of [`Msys2 | `Cygwin] | `Cygwin_libraries ]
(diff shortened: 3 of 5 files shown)


jonahbeckford:
1. Switched to `get_windows_executable_variant`.
2. I kept `is_cygwin_variant` and marked it deprecated. It is not used in any current ocaml/opam code but perhaps packages outside of Opam rely on it.

src/core/opamProcess.ml
```diff
@@ -312,9 +312,10 @@ let create_process_env =
   if Sys.win32 then
     fun cmd ->
       let resolved_cmd = resolve_command cmd in
-      if OpamStd.(Option.map_default Sys.is_cygwin_variant `Native resolved_cmd) = `Cygwin then
+      match OpamStd.(Option.map_default Sys.get_windows_executable_variant `Native resolved_cmd) with
+      | `Cygwin | `Msys2 ->
         cygwin_create_process_env cmd
-      else
+      | _ ->
```
dra27, line 318: This diff can be reverted (see below)

src/core/opamProcess.ml
```diff
@@ -436,9 +437,10 @@ let create ?info_file ?env_file ?(allow_stdin=true) ?stdout_file ?stderr_file ?e
       else
         cmd, args in
     let create_process, cmd, args =
-      if Sys.win32 && OpamStd.Sys.is_cygwin_variant cmd = `Cygwin then
+      match OpamStd.Sys.get_windows_executable_variant cmd with
+      | `Cygwin | `Msys2 ->
         cygwin_create_process_env, cmd, args
-      else
+      | _ ->
```
dra27, line 443: This diff can be reverted (see below)

src/core/opamStd.ml
```diff
@@ -995,19 +995,27 @@ module OpamSys = struct
...
-                else if a <> `Cygwin then
+                else if a <> `Cygwin && a <> `Msys2 then
+                  f `CygLinked
+                else
+                  f a
+              else if OpamString.ends_with ~suffix:"msys-2.0.dll" tx then
+                if OpamString.starts_with ~prefix:"  " x then
+                  f `Msys2
+                else if a <> `Cygwin && a <> `Msys2 then
                   f `CygLinked
```
dra27, line 1019:
```suggestion
                else if a = `Native then
                  f (`Tainted `Cygwin)
                else
                  f a
              else if OpamString.ends_with ~suffix:"msys-2.0.dll" tx then
                if OpamString.starts_with ~prefix:"  " x then
                  f `Msys2
                else if a = `Native then
                  f (`Tainted `Msys2)
```

dra27: It's OK not to handle the very bizarre case of an executable which has managed to link with both msys2 and cygwin-based DLLs!

src/core/opamStd.ml
```diff
@@ -995,19 +995,27 @@ module OpamSys = struct
       (fun f -> try f () with _ -> ())
       !registered_at_exit
 
-  let is_cygwin_variant =
+  let get_windows_executable_variant =
     if Sys.win32 then
       let results = Hashtbl.create 17 in
-      let requires_cygwin name =
+      let requires_cygwin_or_msys2 name =
```
dra27, line 1001: The renaming of this function doesn't add anything - please could it go back to `requires_cygwin`?

src/core/opamStd.ml
```diff
@@ -1022,18 +1030,25 @@ module OpamSys = struct
...
               result
     else
       fun _ -> `Native
 
+  let is_cygwin_variant cmd =
+    match get_windows_executable_variant cmd with
+    | `Native -> `Native
+    | `Cygwin -> `Cygwin
+    | `CygLinked -> `CygLinked
+    | `Msys2 -> `Cygwin
```
dra27, line 1050:
```suggestion
    | `Native -> `Native
    | `Cygwin
    | `Msys2 -> `Cygwin
    | `Tainted _ -> `CygLinked
```

src/core/opamStd.mli
```diff
@@ -452,13 +452,27 @@ module Sys : sig
   val split_path_variable: ?clean:bool -> string -> string list
 
   (** For native Windows builds, returns [`Cygwin] if the command is a Cygwin-
-      compiled executable, [`CygLinked] if the command links to a library which is
-      itself Cygwin-compiled or [`Native] otherwise.
+      compiled executable, [`Msys2] if the command is a MSYS2-compiled executable,
+      and [`CygLinked] if the command links to a library which is itself
```
dra27, line 456:
```suggestion
      and [`Tainted of [ `Msys2 | `Cygwin ]] if the command links to a library which is itself
```

src/core/opamStd.mli
```diff
@@ -452,13 +452,27 @@ module Sys : sig
...
+      Cygwin- or MSYS2-compiled, or [`Native] otherwise.
 
       Note that this returns [`Native] on a Cygwin-build of opam!
 
+      Both cygcheck and an unqualified command will be resolved using the current PATH. *)
+  val get_windows_executable_variant: string -> [ `Native | `Cygwin | `CygLinked | `Msys2 ]
```
dra27, line 462:
```suggestion
  val get_windows_executable_variant: string -> [ `Native | `Cygwin | `Tainted of [ `Msys2 | `Cygwin] | `Msys2 ]
```

src/core/opamStd.mli
```diff
@@ -452,13 +452,27 @@ module Sys : sig
...
+      Cygwin-compiled, or [`Native] otherwise.
+
+      Note that this returns [`Native] on a Cygwin-build of opam!
+
+      MSYS2-compiled executables will return [`Cygwin], but you should be using
+      [get_windows_executable_variant] instead. This function is deprecated.
+
       Both cygcheck and an unqualified command will be resolved using the current PATH. *)
   val is_cygwin_variant: string -> [ `Native | `Cygwin | `CygLinked ]
+  [@@@ocaml.alert deprecated "Please use get_windows_executable_variant instead"]
```
dra27, line 475:
```suggestion
  (** For native Windows builds, returns [`Cygwin] if the command is a Cygwin- or Msys2-
      compiled executable, and [`CygLinked] if the command links to a library which is itself
      Cygwin/Msys2-compiled, or [`Native] otherwise.

      Note that this returns [`Native] on a Cygwin-build of opam!

      Both cygcheck and an unqualified command will be resolved using the current PATH. *)
  val is_cygwin_variant: string -> [ `Native | `Cygwin | `CygLinked ]
```

src/core/opamSystem.ml
```diff
@@ -477,10 +477,12 @@ let apply_cygpath name =
...
-      lazy (if OpamStd.(Option.map_default Sys.is_cygwin_variant `Native (resolve_command command)) = `Cygwin then
-              apply_cygpath
-            else
-              fun x -> x)
+      lazy (
+        match OpamStd.(Option.map_default Sys.get_windows_executable_variant `Native (resolve_command command)) with
+        | `Cygwin | `Msys2 ->
+          apply_cygpath
+        | _ ->
+          fun x -> x)
```
dra27, line 485: This diff can be removed

src/core/opamSystem.ml
```diff
@@ -754,10 +795,10 @@ let install ?(warning=default_install_warning) ?exec src dst =
...
       if cygcheck then
-        match OpamStd.Sys.is_cygwin_variant dst with
+        match OpamStd.Sys.get_windows_executable_variant dst with
           `Native ->
             ()
-        | `Cygwin ->
+        | `Cygwin | `Msys2 ->
             warning dst `Cygwin
         | `CygLinked ->
             warning dst `Cygwin_libraries
```
dra27, line 804:
```suggestion
        (`Cygwin | `Msys2 | `Tainted _) as code ->
            warning dst code
```
with corresponding updated constructors and messages in the `default_install_warning` function.

dra27 (commented):
Thanks for the update to this - sorry for the number of comments, though most of them are in fact reverts!

Consider Ubuntu and Debian: most of the time we can just treat as Debian in terms of handling and then very occasionally we care that it's specifically Ubuntu. I think the Msys2/Cygwin API should be the same - most of the time we simply need to know that it's Cygwin-like and we can call that Cygwin. Then only the specific place where you want some handling that's _just_ for Msys2 do we actually match on that. It makes the diff smaller too 🙂

I haven't fleshed out the change to `OpamSystem.default_install_warning` - I think there should be Msys2-specific messages for that.

jonahbeckford:
The suggested changes were made.

dra27 (approved):
Apologies for the delay approving this (and thanks to @rjbou for rebasing it!). Thanks for this - the rsync support will either get extended to the Cygwin-side at some point or the whole thing will get replaced with Windows-aware OCaml versions but, in the meantime, this is a great stability improvement!

kit-ty-kate:
Thanks a lot!

PR #4825, opened 2021-09-02
Fallback on dnf if yum does not exist on RHEL-based systems

kit-ty-kate:
On some settings, some RHEL-based systems do not come with `yum` installed by default.
Encountered by @vsiles on CentOS (Stream) and minimal docker images for Oraclelinux 8 https://github.com/avsm/ocaml-dockerfile/pull/35

### DIFF
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -136,7 +136,7 @@ let yum_cmd = lazy begin
   else if OpamSystem.resolve_command "dnf" <> None then
     "dnf"
   else
-    failwith "Could not find either yum or dnf to install external dependencies"
+    raise (OpamSystem.Command_not_found "yum or dnf")
 end
 
 let packages_status packages =
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -130,6 +130,14 @@ let family =
   ) in
   fun () -> Lazy.force family
 
+let yum_cmd = lazy begin
+  if OpamSystem.resolve_command "yum" <> None then
+    "yum"
+  else if OpamSystem.resolve_command "dnf" <> None then
+    "dnf"
+  else
+    failwith "Could not find either yum or dnf to install external dependencies"
+end
 
 let packages_status packages =
   let (+++) pkg set = OpamSysPkg.Set.add (OpamSysPkg.of_string pkg) set in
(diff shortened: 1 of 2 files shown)


vsiles:
Version info:
```
$ cat /etc/os-release
NAME="CentOS Stream"
VERSION="8"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="8"
PLATFORM_ID="platform:el8"
PRETTY_NAME="CentOS Stream 8"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:8"
HOME_URL="https://centos.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_SUPPORT_PRODUCT="Red Hat Enterprise Linux 8"
REDHAT_SUPPORT_PRODUCT_VERSION="CentOS Stream"
```

dra27:
It's slightly strange to use the deprecated command but fallback to the replacement 🙂 I haven't dug into the full details, but for RHEL 8+, `dnf` is the preferred package manager interface

kit-ty-kate:
> It's slightly strange to use the deprecated command but fallback to the replacement 🙂 I haven't dug into the full details, but for RHEL 8+, `dnf` is the preferred package manager interface

The main reasoning is that it does not imply too much change compared to before and that it forces us to detect if we use non-backward compatible options quicker.

bikallem:
Perhaps related with this PR #4791 ? We are proposing the use of `rpm` there but here it seems we are proposing the use of `dnf`? Wouldn't PR #4971 make this PR redundant or conflict with each other?

kit-ty-kate:
> Perhaps related with this PR #4791 ? We are proposing the use of `rpm` there but here it seems we are proposing the use of `dnf`? Wouldn't PR #4971 make this PR redundant or conflict with each other?

`dnf` and `rpm` are two very different programs. The latter critically doesn't know how to download packages from the package repository. The former calls the latter to do the low-level stuff.

For opam:
* `dnf`/`yum` is used to install packages
* `rpm` will be used to detect locally installed packages

src/state/opamSysInteract.ml
```diff
@@ -130,6 +130,14 @@ let family =
   ) in
   fun () -> Lazy.force family
 
+let yum_cmd = lazy begin
+  if OpamSystem.resolve_command "yum" <> None then
+    "yum"
+  else if OpamSystem.resolve_command "dnf" <> None then
+    "dnf"
+  else
+    failwith "Could not find either yum or dnf to install external dependencies"
```
rjbou, line 139:
```suggestion
    raise (OpamSystem.command_not_found "yum or dnf")
```

rjbou (approved):

rjbou:
needs a rebase too

src/state/opamSysInteract.ml
```diff
@@ -130,6 +130,14 @@ let family =
   ) in
   fun () -> Lazy.force family
 
+let yum_cmd = lazy begin
```
rjbou, line 133: Just cosmetic and personal appreciation: why not `yum_cmd: unit -> string` with an inner lazy/force?

kit-ty-kate: I’m not sure why it would bring over just `string lazy_t`

kit-ty-kate:
Rebased

PR #4827, opened 2021-09-05
Fix Permission Denied on Win32 install operation

jonahbeckford:


dra27:
Apologies for the delay looking at this! GitHub still doesn't let one do code suggestions anywhere in the file, so I've pushed a commit with my suggestions, but that commit may still be removed. My comments:
- `Unix.chmod` is an emulation on Windows - it only controls the readonly bit, so I'd suggest `0o666` as that more accurately reflects what will happen (Dune does the same, although that might be down to shared authorship!)
- In `setup_copy`, we should preserve the `cp`-like semantics, which are inode preserving (i.e. it truncates if possible). To this end, instead of always deleting the target on Windows, I instead moved the `chmod` change around the opening of the file. `open_out_gen` doesn't give very rich information, so instead of `open_out_gen` followed later by `Unix.descr_of_out_channel`, I switched to `Unix.openfile` and use `Unix.out_channel_of_descr`. This is safe for Windows as binary mode is wanted and no attempt will be made to change that.
- The `remove_file` function could do with the same logic, so I added it there too (propagates to `copy_file` as well)

Issue #4411, opened 2020-10-31
archive-mirrors in the repo file takes over archive-mirrors from the opam config


kit-ty-kate:
https://github.com/ocaml/opam-repository/pull/17355#issuecomment-717479996

AltGr:
Indeed, ignoring customisation matters, it made sense at the time to try the more specific cache (just this repo) before the more general one. Changing the order wouldn't be difficult, but we need to make sure there won't be other downsides.

A reasonable possibility would be to actually have two fields in `opamroot/config`: _e.g._ `global-cache` and `fallback-cache`, or something like that.

AltGr:
The change would otherwise be as trivial as changing the order on this line: https://github.com/ocaml/opam/blob/4141b47048880f90352a630beb2be367b62b284f/src/state/opamUpdate.ml#L501

PR #4830, opened 2021-09-07
Set the priority of user-set archive-mirrors higher than the repositories'

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/4411

I can't find any good reason someone would have a custom cache but want the repository one (even a custom local one) to be taken first.

The one annoying issue with this is that this changes behaviour based on the version of opam used by the user, however we can always announce it before the release (or rather before the change in opam-repository), so it should be ok.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -88,6 +88,8 @@ users)
 
 ## Repository
   * Don't display global message when `this-switch` is given [#4899 @rjbou - fix #4889]
+  * Set the priority of user-set archive-mirrors higher than the repositories'.
+    This allows opam-repository to use the default opam.ocaml.org cache and be more resilient to changed/force-pushed or unavailable archives. [#4830 @kit-ty-kate - fixes #4411]
 
 ## VCS
   * Pass --depth=1 to git-fetch in the Git repo backend [#4442 @dra27]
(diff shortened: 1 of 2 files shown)


dra27:
The reasoning for the original behaviour is mentioned in https://github.com/ocaml/opam/issues/4411#issuecomment-720333288. Discussing this morning, we think the only downside is the roundtrip for a 404 when trying a global cache which doesn't have a file where the repo cache then does (so the downside is small).

Sadly `archive-mirrors` isn't itself filtered, but the (negative) feedback when `archive-mirrors` was added to `git+https://github.com/ocaml/opam-repository.git` in https://github.com/ocaml/opam-repository/pull/17355 (and was reverted in https://github.com/ocaml/opam-repository/pull/17552) was for CI systems, which should more readily be in a position to update to the required version of `opam` if the change causes problems. I agree that an announcement is probably sufficient on the opam-repository PR which makes the change.

kit-ty-kate:
> The one annoying issue with this is that this changes behaviour based on the version of opam used by the user, however we can always announce it before the release (or rather before the change in opam-repository), so it should be ok.

Actually there is a way to do it (or not) based on the opam version: https://github.com/ocaml/opam-repository/pull/20130
So this isn’t an issue anymore.

kit-ty-kate:
@AltGr approved via DM.

PR #4833, opened 2021-09-09
depexts on archlinux: virtual packages detection

### LINKED ISSUE #4759: opam 2.1 and virtual packages

rjbou:
partially fix #4759
Use `pacman -Si` to get all package information

### DIFF
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -306,6 +306,8 @@ let packages_status packages =
          >Optional Deps   : python-setuptools
          >                  python-pip
          >[...]
+         
+         Format partially described in https://archlinux.org/pacman/PKGBUILD.5.html
       *)
       (* Discard stderr to not have it pollute output. Plus, exit code is the
          number of packages not found. *)
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -192,6 +192,27 @@ let packages_status packages =
         |> OpamSysPkg.Set.add (OpamSysPkg.of_string short_name)
       ) OpamSysPkg.Set.empty l
   in
+  let compute_sets_with_virtual get_avail_w_virtuals get_installed  =
+    let sys_available, sys_provides = get_avail_w_virtuals () in
+    let need_inst_check =
+      OpamSysPkg.Map.fold (fun cp vps set ->
+          if OpamSysPkg.Set.(is_empty (inter vps packages)) then set else
+            OpamSysPkg.Set.add cp set)
+        sys_provides packages
+    in
+    let str_need_inst_check = to_string_list need_inst_check in
+    let sys_installed = get_installed str_need_inst_check in
+    let sys_installed =
+      (* Resolve installed "provides" packages;
+         assumes provides are not recursive *)
+      OpamSysPkg.Set.fold (fun p acc ->
+          match OpamSysPkg.Map.find_opt p sys_provides with
+          | None -> acc
+          | Some ps -> OpamSysPkg.Set.union acc ps)
+        sys_installed sys_installed
+    in
+    compute_sets sys_installed ~sys_available
+  in
   match family () with
   | Alpine ->
     (* Output format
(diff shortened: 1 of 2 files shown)


src/state/opamSysInteract.ml
```diff
@@ -264,27 +285,72 @@ let packages_status packages =
...
+                rep1 @@ alt [alnum; punct];
+                char '/';
+                group @@ rep1 @@ alt [alnum; punct];
+                space;
+                rep1 any;
+                str "[installed]";
+                eol
+              ])
+      in
+      run_query_command "pacman" ["-Ss" ; names_re ~str_pkgs ()]
```
kit-ty-kate, line 350: why not using `pacman -Qi` ? It would avoid us to maintain two separate parsers for archlinux

rjbou: From what i know, `pacman -Qi` display only installed packages, so we can't use it to detect available ones.
Thinking about it would simplify the parser to use `pacman -Qs` instead.

kit-ty-kate: We already have `pacman -Si` for the available ones and `pacman -Ss` only displays installed packages too so using `pacman -Qi` would simplify everything

kit-ty-kate: We can keep `pacman -Ss` for now. It can always be changed later if needed.

src/state/opamSysInteract.ml
```diff
@@ -192,6 +192,27 @@ let packages_status packages =
         |> OpamSysPkg.Set.add (OpamSysPkg.of_string short_name)
       ) OpamSysPkg.Set.empty l
   in
+  let compute_sets_with_virtual get_avail_w_virtuals get_installed  =
```
kit-ty-kate, line 195:
```suggestion
  (* TODO: use it in the Debian section *)
  let compute_sets_with_virtual get_avail_w_virtuals get_installed  =
```

kit-ty-kate (approved):
Looks good overall. Thanks a lot! ✨

src/state/opamSysInteract.ml
rjbou: Changed to `-Qs`, faster, and as we want only installed packages status.

kit-ty-kate: mh, -Qs doesn't handle virtual packages though so if used it would have to be without arguments

kit-ty-kate (changes requested):
see previous comment about the -Qs change

src/state/opamSysInteract.ml
rjbou: It doesn't return the provided package name, but recognises it
```
# pacman -Qs zeromq
local/zeromq 4.3.4-1
    Fast messaging system built on sockets. C and C++ bindings. aka 0MQ, ZMQ.
# pacman -Qs libzmq.so
local/zeromq 4.3.4-1
    Fast messaging system built on sockets. C and C++ bindings. aka 0MQ, ZMQ.
```

kit-ty-kate: oh, nevermind my testing had issues. Looks alright

PR #4834, opened 2021-09-09
Fix error 67 in presence of an archive without extention

kit-ty-kate:
Archives without exentions exist in the wild and giving an error just because the extention isn't explicit doesn't feel right (especially an error)
With this change, we assume that any `http` or `rsync` source is an archive.

Examples:
```
Error in wget.0.1.0:
              error 67: Checksum specified with a non archive url: "https://github.com/samoht/ocaml-wget/tarball/0.1.0 - md5=d1e2f51b61b8d8791ccd7bc2a18d98dc"
```
```
Error in patoline.0.1:
              error 67: Checksum specified with a non archive url: "https://opam.ocaml.org/cache/md5/2f/2f7fe33ef6dff305e9a2407696c7a799 - md5=2f7fe33ef6dff305e9a2407696c7a799"
```

### DIFF
--- a/src/state/opamFileTools.ml
+++ b/src/state/opamFileTools.ml
@@ -296,7 +296,7 @@ let t_lint ?check_extra_files ?(check_upstream=false) ?(all=false) t =
     t.url >>| OpamFile.URL.url >>| (fun u ->
         match u.OpamUrl.backend with
         | #OpamUrl.version_control -> false
-        | _ -> OpamSystem.is_archive (OpamUrl.base_url u))
+        | `http | `rsync -> true)
   in
   let check_upstream =
     check_upstream &&


dra27:
Not sure - I’m both those examples the linter is showing that the package should be fixed - an extension can be added to the wget link (which also gets rid of the redirect) and the main source of a package shouldn’t be pointing to the cache?!

mseri:
Many packages have disappeared, especially after bitbucket and ocamlforge shutting down. They still exist in the cache, and are unmaintained or with maintainers that are still planning what to do and where to move them. By not allowing to use the cache, will have to effectively delete those packages from the opam repository, even if they are functional or still have some users

mseri:
Anyways, we were just discussing with @kit-ty-kate that we can likely filter the error at the level of the opam CI. If it works fine, it will avoid making this change

kit-ty-kate:
Superseded by #4960

PR #4840, opened 2021-09-16
Handle empty environment variable updates

dra27:
In 2.0.8 (#4326), but not cherry-pick back to master so missing in 2.1


rjbou:
Thanks!

PR #4844, opened 2021-09-23
Cygpath the PATH from opam env

jonahbeckford:
When using a native Windows opam.exe the PATH from `opam env` is in Windows format.

PowerShell output from `opam env`: `$env:PATH = 'Z:\source\opam\_opam\bin;...'`
CMD.exe output from `opam env`: `SET PATH=Z:\source\opam\_opam\bin;...`

With this change both MSYS2 and Cygwin will Unix-ize the PATH with essentially `cygpath --path $PATH`.

Cygwin output from `opam env`: `PATH='/cygdrive/c/Users/beckf/AppData/Local/opam/diskuv-boot-DO-NOT-DELETE/bin:...'`
MSYS2 output from `opam env`: `PATH='/z/source/opam/_opam/bin:...'`

`cygpath --path` on an already Unix PATH is idempotent, so a Cygwin opam.exe will still work after this change. That is, the following says "MATCHES" on both Cygwin and MSYS2:
```sh
if [[ "$(cygpath --path '$PATH')" = $(cygpath --path "$(cygpath --path '$PATH')") ]]; then echo MATCHES; fi
```

---

This PR gets `eval $(opam env)` working from a native Windows opam.exe into a MSYS2 shell. However the same command in a Cygwin shell chokes with `-bash: $'\r': command not found` because native Windows opam.exe prints carriage returns (which MSYS2 eval can handle, but not Cygwin). I don't actually know a clean fix for that, and I suspect removing carriage returns when in MSYS2/Cygwin probably belongs in another PR.

### DIFF
--- a/src/core/opamSystem.ml
+++ b/src/core/opamSystem.ml
@@ -506,7 +506,7 @@ let get_cygpath_function =
 let apply_cygpath_path_transform path =
   let r =
     OpamProcess.run
-      (OpamProcess.command ~name:(temp_file "command") ~verbose:false "cygpath" ["--path"; path])
+      (OpamProcess.command ~name:(temp_file "command") ~verbose:false "cygpath" ["--path"; "--"; path])
   in
   OpamProcess.cleanup ~force:true r;
   if OpamProcess.is_success r then
--- a/src/core/opamSystem.ml
+++ b/src/core/opamSystem.ml
@@ -503,6 +503,28 @@ let get_cygpath_function =
     let f = Lazy.from_val (fun x -> x) in
     fun ~command:_ -> f
 
+let apply_cygpath_path_transform path =
+  let r =
+    OpamProcess.run
+      (OpamProcess.command ~name:(temp_file "command") ~verbose:false "cygpath" ["--path"; path])
+  in
+  OpamProcess.cleanup ~force:true r;
+  if OpamProcess.is_success r then
+    List.hd r.OpamProcess.r_stdout
+  else
+    OpamConsole.error_and_exit `Internal_error "Could not apply cygpath --path to %s" path
+
+let get_cygpath_path_transform =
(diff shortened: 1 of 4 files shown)


jonahbeckford:
Rebased.

src/core/opamSystem.ml
```diff
@@ -503,6 +503,28 @@ let get_cygpath_function =
     let f = Lazy.from_val (fun x -> x) in
     fun ~command:_ -> f
 
+let apply_cygpath_path_transform path =
+  let r =
+    OpamProcess.run
+      (OpamProcess.command ~name:(temp_file "command") ~verbose:false "cygpath" ["--path"; path])
```
dra27, line 509:
```suggestion
      (OpamProcess.command ~name:(temp_file "command") ~verbose:false "cygpath" ["--path"; "--"; path])
```

dra27 (commented):
I'm a bit uneasy about this one, to be honest - mostly, I think if you want to use a Cygwin or MSYS2 shell, then it's better to use a Cygwin or MSYS2 `opam`, but our cross-compilation and tooling story isn't good enough for me to insist on that yet!

A thought for the `\r` issue - I expect this can be fixed by detecting that stdout is not a tty (i.e. you're inside `$(eval )`) and detecting the parent as being a Cygwin-ish shell (it's in the same class of shim that the OCaml toplevel uses to display colour in mintty, etc.). That depends on your other PR and also on a further bit of work on top of that which I've done which gives the full path to the parent process, so perhaps it would be worth opening an issue to track that.

This can be merged once the `"--"` part is added, thanks!

dra27 (approved):
Good to merge when CI finishes, thanks!

kit-ty-kate:
CI is green. Thanks!

PR #4847, opened 2021-09-28
Replace linear search with Array.exists

rgrinberg:
Stop scanning once the result is found. Also simplifies the code a bit.


rjbou:
Thanks!

PR #4859, opened 2021-10-06
Add Software ID in opam file

rjbou:
Add Software Heritage ID  field `swhid` in url section, and in opam show, and fallback support

### DIFF
--- a/src/client/opamListCommand.ml
+++ b/src/client/opamListCommand.ml
@@ -541,9 +541,23 @@ let detail_printer ?prettify ?normalise ?(sort=false) st nv =
     +! ""
   | Raw_field f | Field f ->
     (try
-       OpamStd.List.assoc String.equal f
-         (OpamFile.OPAM.to_list (get_opam st nv))
-       |> mini_field_printer ?prettify ?normalise
+       let open OpamStd.Option.Op in
+       let opam = get_opam st nv in
+       let value =
+         match f with
+         | "url.swhid" ->
+           (match OpamFile.(OPAM.url opam >>= URL.swhid) with
+            | Some f -> OpamTypesBase.nullify_pos (String (OpamSWHID.to_string f))
+            | None -> raise Not_found)
+         | "url.mirrors" ->
+           OpamFile.(OPAM.with_url_opt
+                       (OPAM.url opam >>| URL.with_swhid_opt None)
+                       opam
+                     |> OPAM.to_list
+                     |> OpamStd.List.assoc String.equal f)
+         | _ -> OpamStd.List.assoc String.equal f (OpamFile.OPAM.to_list opam)
+       in
+       mini_field_printer ?prettify ?normalise value
      with Not_found -> "")
   | Installed_version ->
     (try OpamPackage.package_of_name st.installed nv.name |> fun inst_nv ->
@@ -745,6 +759,7 @@ let info st ~fields ~raw ~where ?normalise ?(show_empty=false)
     Pinning_target;
     Source_hash;
     Field "url.src";
+    Field "url.swhid";
     Field "url.checksum";
     Field "homepage";
     Field "doc";
--- a/src/core/opamHash.ml
+++ b/src/core/opamHash.ml
@@ -46,14 +46,7 @@ let kind_of_string s = match String.lowercase_ascii s with
   | _ -> invalid_arg "OpamHash.kind_of_string"
 
 let is_hex_str len s =
-  String.length s = len &&
-  try
-    String.iter (function
-        | '0'..'9' | 'A'..'F' | 'a'..'f' -> ()
-        | _ -> raise Exit)
-      s;
-    true
-  with Exit -> false
+  String.length s = len && OpamStd.String.is_hex s
 
 let len = function
   | `MD5 -> 32
--- a/src/core/opamSWHID.ml
+++ b/src/core/opamSWHID.ml
@@ -1,0 +1,113 @@
+(**************************************************************************)
+(*                                                                        *)
+(*    Copyright 2022 OCamlPro                                             *)
+(*                                                                        *)
+(*  All rights reserved. This file is distributed under the terms of the  *)
+(*  GNU Lesser General Public License version 2.1, with the special       *)
+(*  exception on linking described in the file LICENSE.                   *)
+(*                                                                        *)
+(**************************************************************************)
+
+module SWHO = Swhid_core.Object
+module SWH_ID = SWHO.Core_identifier
+module SWH_Compute = Swhid_core.Compute
+
+type t = SWH_ID.t
+
+let compare = SWH_ID.compare
+let equal a b = compare a b = 0
+
+let of_string_opt s =
+  match SWH_ID.of_string s with
+  | Ok i -> Some i
+  | Error _ -> None
+let of_string s =
+  match of_string_opt s with
+  | None -> invalid_arg "OpamSWHSWH_ID.of_string"
+  | Some i -> i
+let to_string = SWH_ID.to_string
+
+let to_json s = `String (to_string s)
+let of_json = function
+  | `String s -> of_string_opt s
+  | _ -> None
+
+let hash s = SWHO.Hash.to_string (SWH_ID.get_hash s)
+
+module O = struct
+  type nonrec t = t
+  let to_string = to_string
+  let to_json = to_json
+  let of_json = of_json
+  let compare = SWH_ID.compare
+end
+
+module Set = OpamStd.Set.Make(O)
+module Map = OpamStd.Map.Make(O)
+
+
+(** Url handling *)
+
+let prefix = "swhid.opam.ocaml.org/"
+
+let is_valid url =
+  let open OpamUrl in
+  url.backend = `http
+  && (String.equal url.transport "http" || String.equal url.transport "https")
+  && OpamStd.String.starts_with ~prefix url.path
+
+let of_url url =
+  try
+    Some (of_string (OpamStd.String.remove_prefix ~prefix url.OpamUrl.path))
+  with Invalid_argument _ -> None
+
+let to_url swh =
+  let path = Printf.sprintf "%s%s" prefix (to_string swh) in
+  OpamUrl.{ transport = "https"; backend = `http ; hash = None; path }
+
+(** Identifier computing *)
+
+module SHA1 = struct
+  let digest_string_to_hex = OpamSHA.sha1_string
+end
+module OS = struct
+
+  let contents dir =
+    try Some (OpamSystem.ls dir)
+    with _ -> None
+
+  let typ name =
+    try
+      Some (if Sys.is_directory name then SWH_Compute.Dir else SWH_Compute.File)
+    with _ -> None
+
+  let read_file name =
+    try Some (OpamSystem.read name)
+    with OpamSystem.File_not_found _ -> None
+
+    (*
+      - [0o120000] if [f] is a symlink
+      - [0o040000] if [f] is a directory
+      - [0o100755] if [f] is an executable file
+      - [0o100644] if [f] is a regular file *)
+  let permissions name =
+    if not (Sys.file_exists name) then raise Not_found else
+    let { Unix.st_kind; _ } = Unix.lstat name in
+    match st_kind with
+    | Unix.S_DIR -> Some 0o040000
+    | Unix.S_LNK -> Some 0o120000
+    | Unix.S_REG ->
+      Some (if OpamSystem.is_exec name then 0o100755 else 0o100644)
+    | Unix.S_CHR | Unix.S_BLK | Unix.S_FIFO | Unix.S_SOCK -> None
+
+  let base name = OpamFilename.(Base.to_string (basename (of_string name)))
+
+end
+
+module Compute = SWH_Compute.Make(SHA1)(OS)
+
+let compute dir =
+  match Compute.directory_identifier_deep
+          (OpamFilename.Dir.to_string dir) with
+  | Error _ ->  None
+  | Ok identifier -> Some (SWHO.Hash.to_string (SWHO.get_hash identifier))
--- a/src/repository/opamDownload.ml
+++ b/src/repository/opamDownload.ml
@@ -193,3 +193,28 @@ let download ?quiet ?validate ~overwrite ?compress ?checksum url dstdir =
   in
   download_as ?quiet ?validate ~overwrite ?compress ?checksum url dst @@|
   fun () -> dst
+
+
+(** Stdout output retrieval and post requests management *)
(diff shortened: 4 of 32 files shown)


src/core/opamHash.ml
```diff
@@ -148,3 +148,50 @@ end
...
+  let of_json = function
+    | `String s -> of_string_opt s
+    | _ -> None
+
+  (** XXX Keep that? *)
+  let compare = String.compare
+  let equal = String.equal
+
+  module O = struct
+    type _t = t
```
zapashcanon, line 186: `type nonrec t = t` ?

src/core/opamUrl.ml
```diff
@@ -308,3 +308,21 @@ module Op = struct
...
+module SWHID = struct
+
+  let prefix = "swhid.opam.ocaml.org/"
+  let is_valid u =
+    u.backend = `http
+    && (String.equal u.transport "http" || String.equal u.transport "https")
+    && OpamStd.String.starts_with ~prefix u.path
+  let of_url u =
+    try
+      Some (OpamHash.SWHID.of_string (OpamStd.String.remove_prefix ~prefix u.path))
```
zapashcanon, line 321: IIUC this doesn't support having directly `swh:...` as an URL ? I think we should support both *now* (even if in the beginning we use only `http{s}://swhid.opam.ocaml.org/...` in the official opam repository) to ease the transition to `swh:...` only - which is registered at IANA etc.

rjbou: Tried with simply `swh:...`, it is resolved to a ssh url. There will be another url kind to handle, and besides that, older opams will print the ssh url.

src/core/opamSWHID.ml
```diff
@@ -0,0 +1,160 @@
...
+  try
+    Some (of_string (OpamStd.String.remove_prefix ~prefix url.OpamUrl.path))
+  with Invalid_argument _ -> None
+
+let to_url swh =
+  let path = Printf.sprintf "%s%s" prefix (to_string swh) in
+  OpamUrl.{ transport = "https"; backend = `http ; hash = None; path }
+
+
+(** SWHID retrieval functions *)
```
zapashcanon: You can remove everything in this file starting from this line. :)

rjbou:
Updated!

tests/reftests/swhid.unix.test
```diff
@@ -0,0 +1,145 @@
...
+  checksum: "md5=039a77d9ffb9b22ab845df30436390c1"
+}
+### opam switch create fallback --empty
+### opam install snappy-ok
+The following actions will be performed:
+=== install 1 package
+  - install snappy-ok 1
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> retrieved snappy-ok.1  (https://ygrek.org/p/release/ocaml-snappy/ocaml-snappy-0.1.2.tar.gz)
```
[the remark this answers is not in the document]
rjbou: No other solution, i'm afraid, we need to retrieve the archive from swh. The tests don't pass the CI in any way, as we depend on distant opam repo in some other tests, and then need network, it is not a new issue.

src/client/opamListCommand.ml
```diff
@@ -540,8 +540,24 @@ let detail_printer ?prettify ?normalise ?(sort=false) st nv =
...
+           | None -> raise Not_found
+         else
+         let opam =
+           if f = "url.mirrors" then
+             OpamFile.(OPAM.with_url_opt
+                         (OPAM.url opam >>| URL.with_swhid_opt None)
+                         opam)
+           else opam
+         in
+         List.assoc f (OpamFile.OPAM.to_list opam)
```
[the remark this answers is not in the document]
rjbou: Agree, factorisation habit, but sometimes in less clear :)
No better solution than proposed

kit-ty-kate (approved):

src/repository/opamDownload.ml
```diff
@@ -193,3 +193,169 @@ let download ?quiet ?validate ~overwrite ?compress ?checksum url dstdir =
...
+      "obj_id": "4453cfbdab1a996658cd1a815711664ee7742380"
+     }
+  *)
+
+  let get_output ?(post=false) url =
+    get_output ~post url @@| function
+    | Some out ->
+      Some (String.concat "" out)
+    | None ->
+      OpamConsole.error "Software Heritage fallback need %s or %s installed"
```
dra27, line 259:
```suggestion
      OpamConsole.error "Software Heritage fallback needs %s or %s installed"
```

src/repository/opamDownload.ml
```diff
@@ -193,3 +193,169 @@ let download ?quiet ?validate ~overwrite ?compress ?checksum url dstdir =
...
+  let fallback_err fmt = Printf.sprintf ("SWH fallback: "^^fmt)
+
+  let get_url ?(timeout=6) swhid =
+    let attempts = timeout in
+    let hash = OpamSWHID.hash swhid in
+    let rec aux timeout =
+      if timeout <= 0 then
+        Done (Not_available
+                (Some (fallback_err "timeout"),
+                 fallback_err "%d attempts tried, aborting" attempts))
```
dra27, line 288:
```suggestion
                 fallback_err "%d attempts tried; aborting" attempts))
```

src/repository/opamDownload.ml
```diff
@@ -193,3 +193,169 @@ let download ?quiet ?validate ~overwrite ?compress ?checksum url dstdir =
...
+  (* for the moment only used in sources, not extra sources or files *)
+  let archive_fallback ?timeout urlf dirnames =
+    match OpamFile.URL.swhid urlf with
+    | None -> Done (Result None)
+    | Some swhid ->
+      if check_post_tool () then
+        (* Add a global modifier and/or command for default answering *)
+        if OpamConsole.confirm ~default:false
+            "Source %s is not available. Do you want to try to retrieve it \
+             from Software Heritage cache? It may take few minutes."
```
dra27, line 309: Possibly include a link to https://www.softwareheritage.org for information about what it is?

src/repository/opamDownload.ml
```diff
@@ -193,3 +193,169 @@ let download ?quiet ?validate ~overwrite ?compress ?checksum url dstdir =
...
+                     Result (Some "SWH fallback"))
+                  else
+                    Not_available (
+                      Some (fallback_err "archive not valid"),
+                      fallback_err
+                        "archive corrupted, opam file swhid %S vs computed %S"
+                        hash computed)))
+        else
+          Done (Not_available
+                  (Some (fallback_err "no retrieval"),
```
dra27, line 354: Possibly "skip" rather than "no"?

dra27 (commented):
Some suggestions on messages

src/repository/opamDownload.ml
```diff
@@ -193,3 +193,169 @@ let download ?quiet ?validate ~overwrite ?compress ?checksum url dstdir =
...
+let get_output ~post ?(args=[]) url =
+  let cmd_args =
+    download_args ~url ~out:"-" ~retry:OpamRepositoryConfig.(!r.retries)
+      ~compress:false ()
+    @ args
+  in
+  let cmd_args =
+    if post then
+      match cmd_args with
+      | ("wget"|"curl" as cmd)::args -> Some (cmd :: ["-X"; "POST"] @ args)
```
AltGr, line 215: It's probably time we do the http requests through an OCaml library!
(unrelated to the nice work here, I know :) )

src/repository/opamDownload.ml
```diff
@@ -193,3 +193,169 @@ let download ?quiet ?validate ~overwrite ?compress ?checksum url dstdir =
...
+      Some (match status with
+          | "done" -> `Done (OpamUrl.of_string fetch_url)
+          | "pending" -> `Pending
+          | "new" -> `New
+          | "failed" -> `Failed
+          | _ -> `Unknown)
+
+  let fallback_err fmt = Printf.sprintf ("SWH fallback: "^^fmt)
+
+  let get_url ?(timeout=6) swhid =
```
AltGr, line 281: Nitpicking I know, but this is not technically a timeout but a `max_tries`

AltGr (changes requested):
Very nice work!

I have one concern: IIUC, the fallback download handling is triggered as soon as the initial download fails. Isn't this a problem for scenarios where the network is down ?
A solution might be to ping `softwareheritage.org` first ?

src/repository/opamDownload.ml
rjbou: agree! It's quite hackish, working but hackish

kit-ty-kate:
> A solution might be to ping `softwareheritage.org` first ?

a real ping isn’t a good solution as some (weird but real) networks disallow ICMP requests

src/repository/opamDownload.ml
rjbou: yes, this is in the todolist with other libs (tar/git/etc.)

rjbou:
In fact, we can simply retrieve the default page and drop the result.

src/repository/opamDownload.ml
```diff
@@ -225,6 +225,7 @@ module SWHID = struct
 
   let log fmt = OpamConsole.log "CURL(SWHID)" fmt
 
+  let main_url = OpamUrl.of_string "http://softwareheritage.com"
```
kit-ty-kate, line 228:
```
This domain name has been registered with Gandi.net
```

rjbou: wrong url indedd

kit-ty-kate: what about just sending a request such as `curl -X POST https://archive.softwareheritage.org/api/1`. It returns no output nor failure?

rjbou: In that way, we can put the check directly on fallback mechanism, not in `OpamUpdate`. :+1:

kit-ty-kate: even better, they have an api endpoint just for that: https://docs.softwareheritage.org/devel/apidoc/swh.web.api.views.ping.html?highlight=ping#get--api-1-ping-

kit-ty-kate (approved):

tests/reftests/swhid.unix.test
```diff
@@ -0,0 +1,139 @@
...
+### opam install snappy-swhid-ko
+The following actions will be performed:
+=== install 1 package
+  - install snappy-swhid-ko 2
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+Source https://fa.ke/url.tar.gz is not available. Do you want to try to retrieve it from Software Heritage cache (https://www.softwareheritage.org)? It may take few minutes. [y/n] y
+[ERROR] Failed to get sources of snappy-swhid-ko.2: SWH fallback: Unknown swhid
+
+OpamSolution.Fetch_fail("SWH fallback: Unknown swhid")
```
AltGr, line 111: Extra exn print ?

AltGr (approved):

tests/reftests/swhid.unix.test
rjbou: Already existing issue, see l.39

avsm:
For my own education, is there any documentation about:
- what Software Heritage IDs are
- how they are added to an opam package
- why they are different from a conventional content-addressed store (e.g. something like IPFS/hash fetching)
- can the fallback be switched off, and/or does this affect purely offline use of opam with local files?

Seems like a nice feature, but I'm missing a little background.

dra27:
@avsm - apologies, this got discussed at a dev meeting in September, but only put into notes from the meeting, and not commented back (which is my fault...)!

> what Software Heritage IDs are
The SWHID (see also [their documentation](https://docs.softwareheritage.org/devel/swh-model/persistent-identifiers.html)) is just a(nother) content hash.

> how they are added to an opam package

This PR (and opam 2.2) is focussed on being able to consume the content hashes - for adding them, the intention is to have existing package submission tools (`opam-publish`, `dune-release`, etc.) generate the SWHID along with other hashes from the release tarball. opam-repository _may_ then choose to lint based on these. Part of the Software Heritage project work for programming language ecosystems is that they monitor opam-repository and ensure that the archives referenced are actually in their archive (it is also possible - and desirable - that the package submission would itself submit the archive to SWH, but the point is that the SWHID being in an opam file is neither gated nor requires its actually being submitted to the service).

Adding the facility to use the service in opam 2.2 obviously doesn't mandate deploying it for all packages on opam-repository (or in a future version) - just as adding the additional sha hash functions earlier in opam 1.x didn't.

> why they are different from a conventional content-addressed store (e.g. something like IPFS/hash fetching)

They're related, indeed largely equivalent. The Software Heritage project, though, is focussed on persistence, where IPFS gives availability. I don't have an answer as to why they end up using different content hashes - although there is a [project in progress](https://www.softwareheritage.org/2022/02/10/building-bridge-to-the-software-heritage-archive/) to provide the Software Heritage archive over IPFS.

> can the fallback be switched off, and/or does this affect purely offline use of opam with local files?

Yes - in fact, unless I've misread, it is not automatically enabled (it's an interactive prompt and would be off in batch/non-interactive mode, @rjbou?).

opam being able to retrieve archives from Software Heritage is a partial solution to an upstream server being down (opam.ocaml.org, GitHub, etc.) but that's not the primary point. Primarily, this provides an external solution and ongoing solution for things like https://github.com/ocaml/opam-source-archives and should mean that we never end up scrabbling to find archives in old server images etc. when gforge services get shut down or personal web servers get taken offline, etc.

rjbou:
> can the fallback be switched off, and/or does this affect purely offline use of opam with local files?

The fallback is enabled by default, it can be disabled by `opam option swh_fallback=false`. It is used in last resort when archive is unreachable via 1. url defined in opam package 2. in opam repository cache & its mirrors. From there, a prompt asks if you want to try download via SWH (see [in test](https://github.com/rjbou/opam/blob/swhid2/tests/reftests/swhid.unix.test#L88-L90)). Before calling the fallback/prompting, we ensure that network is up & SWH api is up.

avsm:
Many thanks for the updates. One quick design question though, after reading the SWHID docs linked: it looks like they [define a URL schema](https://docs.softwareheritage.org/devel/swh-model/persistent-identifiers.html#uri-scheme), so why don't we just interpret this schema in `url` fields (just as we do for https+git for example)?

zapashcanon:
We wanted to use it but we couldn't, mainly for compatibility with old clients. See [this comment](https://github.com/ocaml/opam/pull/4859#discussion_r778476033).

rdicosmo:
Happy to see this moving forward!
As a side note, a working group is assembling to start the normalization process of SWHIDs, and it would be great if some of the contributors to this work in opam could join. See more information here https://www.swhid.org/

dra27 (approved):

PR #4861, opened 2021-10-07
Fix reverting additions to PATH-like variables

dra27:
For environment variables which can be split (e.g. `PATH` on `:`), `OpamEnv.unzip_to` attempts to find the point in the variable at which a value was added. This is fine if a _single_ value was added, but fails if the addition was multiple values (for example, if a setenv instruction added two directories to `PATH` in one `+=`).

This is fixed by first splitting the value being searched according to the same rule as the environmen variable and ensuring they all match in order.

- [x] Finalise test case
- [x] Double-check that it definitely fixes the Windows issue (I've rebased and edited it on Linux)

### DIFF
--- a/tests/reftests/env.test
+++ b/tests/reftests/env.test
@@ -18,8 +18,10 @@ Switch invariant: ["nv"]
 Done.
 ### opam env | grep "NV_VARS" | '[:;]' -> '-' | '[:;]' -> '-' | '[:;]' -> '-'
 NV_VARS='${BASEDIR}/OPAM/setenv/doc/nv-${OPAMTMP}/OPAM/setenv/share/nv'- export NV_VARS-
-### opam exec -- opam env --revert | grep "NV_VARS"
+### opam exec -- opam env --revert | grep "NV_VARS" | '[:;]' -> '-' | '[:;]' -> '-' | '[:;]' -> '-'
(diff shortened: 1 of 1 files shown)


dra27:
This opam file provides the basis for a test:

```
opam-version: "2.0"
license: "ISC"
synopsis: "Test package for setenv field"
description: "Test case for setenv field"
maintainer: "platform@lists.ocaml.org"
setenv: [
  [CAML_LD_LIBRARY_PATH += "%{_:doc}%:%{_:share}%"]
]
homepage: "https://opam.ocaml.org"
bug-reports: "https://github.com/ocaml/opam/issues"
authors: ["David Allsopp"]
```
This package needs to be installed on its own (or the `ocaml` package masks the problem) with `opam switch create --repos=setenv-test-repo setenv-test --packages=setenv`. With current opam `opam env --revert` will not include `CAML_LD_LIBRARY_PATH` (because opam fails to detect the need to revert the environment update at all) and with this PR it will include `export CAML_LD_LIBRARY_PATH=''`.

@rjbou - I wasn't sure where best to integrate this as a reftest... perhaps we can share some of the infrastructure used for the root-versions test in order to set-up a repository with just this package, or is there already an easier way in another test?

PR #4863, opened 2021-10-09
Document custom licenses

### LINKED ISSUE #4862: Syntax for avoiding spdx warning 62 ?
Unless I missed it*, it seems there is no syntax to specify a license identifier for licenses that do not fall under the SPDX ones. 

Not everything out there fits into SPDX identifiers; if only non-libre software. Here's an example:

```
/Users/dbuenzli/sync/repos/odig/opam: Warnings.
  warning 62: License doesn't adhere to the SPDX standard, see https://spdx.org/licenses/:
 "ParaType-Free-Font-License", "DejaVu-fonts"
```

For example `ParaType-Free-Font-License` is what is used by debian in the corresponding [font package](https://metadata.ftp-master.debian.org/changelogs//main/f/fonts-paratype/fonts-paratype_20181108-3_copyright). `DejaVu-fonts` is my invention, it's a complicated set of licenses (I suspect the debian package specification for that one is underspecified); it just means go check it out if you need to. 

If there is no established convention, maybe a `not-spdx:` prefix ? (Though I highly dislike the `spdx` jargon). 

*: I have to admit I have difficulties to navigate the `spdx` standard and website which it seems has been taken over by enterprisy bureaucrats.

kit-ty-kate:
Closes #4862

### DIFF
--- a/doc/pages/Manual.md
+++ b/doc/pages/Manual.md
@@ -802,8 +802,8 @@ files.
   a list of strings listing the original authors of the software.
 
 - <a id="opamfield-license">`license: [ <string> ... ]`</a>:
-  The SPDX ID of the license(s) under which the source software is available
-  (see http://spdx.org/licenses/).
+  The SPDX expression of the license(s) under which the source software is available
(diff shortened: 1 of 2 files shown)


dbuenzli:
The link https://spdx.github.io/spdx-spec/appendix-IV-SPDX-license-expressions/ is a 404

dbuenzli:
I think this would be the place to mention a semantics if you get a list of expressions. I would suggest the expression are simply `AND`ed, at least that's what I had in mind when I wrote lists in my `license:` fields.

PR #4877, opened 2021-10-28
Make yes the default answer to opam init's setup in interactive mode

kit-ty-kate:
Beginner hitting the issue in: https://discuss.ocaml.org/t/beginner-issue-ocamlc-not-found-in-tree-or-in-path/8701/3

I'm not sure why the default answer to this critical experience of opam is "no" by default.
Changing it to "yes" by default would be more inviting for beginners and would not change anything for other people (given this would only change the default in the interactive mode)

I've also made it loop on itself if the answer could not be parsed so the behaviour is less surprising (e.g. people writing "yes" in their own language but no error message is shown by opam)

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -22,6 +22,8 @@ users)
 
 ## Init
   * Run the sandbox check in the temporary directory [#4787 @dra27 - fix #4783]
+  * Make "yes" the default answer to opam init's setup in interactive mode [#4877 @kit-ty-kate]
+  * Make the interactive mode of opam init's setup ask the user again when the answer isn't recognized instead of using the default value [#4877 @kit-ty-kate]
 
 ## Config report
   *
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
> On a matter of hygiene, I think the old default should be preserved for CLI < 2.2

mmh, I’m not sure a runtime interactive behaviour should be covered by OPAMCLI.

dra27:
The idea with `--cli` is that it would be used in instructions (`README`s, blog posts, etc.), and while I agree it's (very) minor here, the principle of CLI versioning is that `opam init --cli=2.1` would behave the same in 2.2+, even if it's only behaving differently if the user sits on "Return" 🙂

AltGr:
I skipped this when doing #5057 but the two should clearly be linked!

dra27:
#5057 will be updated to include this!

PR #4880, opened 2021-10-28
Make the SAT preemptive check optional, by-solver

AltGr:
Closes #4658, see the discussion there.

It'd probably be worth it to check the Dose3 check times though; they
may be due to the increased complexity of the repo, but also maybe to
the new dose3 version ?

Also I didn't re-check the response times of the different solvers in
case of complex, unsat cases : the worrying case here was mccs stalling
for tenths of seconds, even sometimes timing out, while dose3 (which we
would call anyway) would detect the conflict in a fraction of a second.


kit-ty-kate:
Thanks!

Issue #4311, opened 2020-08-12
[opam 2.1~alpha3] opam list --installable is really slow


kit-ty-kate:
With `opam 2.1~alpha3`:
```
opam@5ab5892e7927:~$ time opam-2.1 list --installable --depends-on ppx_tools
# Packages matching: installable & depends-on(ppx_tools)
[...]
real	3m53.765s
user	3m52.204s
sys	0m0.683s
```
With `opam 2.0.7`:
```
opam@5ab5892e7927:~$ time opam list --installable --depends-on ppx_tools
# Packages matching: installable & depends-on(ppx_tools)
[...]
real	0m20.094s
user	0m19.603s
sys	0m0.206s
```
You can use `docker run --rm -it ocurrent/opam:debian-10-ocaml-4.10` to reproduce this.

rjbou:
The bottleneck is in [universe check](https://github.com/ocaml/opam/blob/master/src/solver/opamSolver.ml#L477). @AltGr, an idea?

kit-ty-kate:
Just a follow-up, here is the same command when the switch invariant is not the default one when upgrading from 2.0:
```
$ opam-2.1 switch invariant
["ocaml-base-compiler" {= "4.11.1"}]
$ time opam-2.1 list --installable --depends-on ppx_tools
[...]
real	8m24.987s
user	8m24.435s
sys	0m0.453s
```
(it seems to be getting worse? :worried:)

emillon:
I bisected this and this has been introduced in #3894 (51e35215d43724bc7257a0d2be6bf619ed32194a is good, 8ca4d4ce93cfe5381d109ea82b09180552f864d7 is bad, couldn't test the 6 commits in between).

Armael:
FWIW I'm hitting the same issue; an `opam list --installable` on a 4.09.0 switch took 5 minutes where it used to take only a few seconds.

kit-ty-kate:
Another example, using opam-repository at `fb801efd13b2e6370271608f2bfd528f2eea72bb`, the following command:
```
opam list -s --color=never --depends-on minios-xen.0.7 --coinstallable-with minios-xen.0.7 --installable --all-versions --recursive --depopts
```
takes:
- `1m27s` with opam 2.0.7
- `19m43s` with opam 2.1.0~beta4

avsm:
[...]

PR #4882, opened 2021-11-02
Optimisation for partial installability queries

AltGr:
Ref. #4311


kit-ty-kate (approved):
Looks good to me! It’s even faster than before. Thank you so much!

Issue #4885, opened 2021-11-03
os-distribution


diremy:
```
$ opam --version
2.1.0
```
opam has a global configuration variable `os-distribution`, but this variable seems to be ignored at least in most (if not all) cases, instead trying to guess the its value dynamically by checking which one of the two commands `brew` or `port` (in this order) is executable.

Here is my global config: 
```
$ opam var | grep os
os                macos             # Inferred from system
os-distribution   macports          # Set through 'opam config set-global'
os-family         bsd               # Set through 'opam config set-global'
os-version        11.6              # Inferred from system
```
However, opam seems to bypass this configuration:
```
$ opam config report | grep system
# system               arch=x86_64 os=macos os-distribution=homebrew os-version=11.6
```
The reason is that (for some reason) I have both `port` and `brew` binaries installed. 
Indeed, removing the executable rights to the `brew` command
```
$ chmod -x /usr/local/bin/brew 
```
fixes the problem:
```
$ opam config report | grep system
# system               arch=x86_64 os=macos os-distribution=macports os-version=11.6
```
But I consider this to be a bug since the global `os-distribution` variable seems to be always overriden.  

As a result,  `opam` 
  - always wants to install external dependencies with `brew`, which I must refuse.
  - fails to install most `conf-*` packages even when these are correctly installed with macports.

My only solution is thus to make `brew` non-executable.

rjbou:
One "issue" here is that opam let you think that you changed the internal `os-distribution` while you just set up a new value for `os-distribution` variable, it have only an effect on variables handling, see [this test](https://github.com/ocaml/opam/blob/master/tests/reftests/show.test#L311) for an example.
We should permit overriding or at least a way to set a priority instead of default ones ([brew is tested first, then macports](https://github.com/ocaml/opam/blob/master/src/state/opamSysPoll.ml#L94)).

diremy:
> One "issue" here is that opam let you think that you changed the internal `os-distribution` while you just set up a new value for `os-distribution` variable, it have only an effect on variables handling.

I don't understand: `os-distribution` is not a system variable. It is only an opam variable.  When I say that `os-distribution` is `macports`, opam could maybe check that this is meaningful, i.e. check the existence of `port`, but I don't understand the reason not to take my global setting into account.  

(BTW, I don't understand the use of the global variable os-distribution if it is always overridden by opam at runtime.)

> see [this test](https://github.com/ocaml/opam/blob/master/tests/reftests/show.test#L311) for an example. 

[...]

PR #4886, opened 2021-11-03
[PoC] Make the os variables overridable using opam var

kit-ty-kate:
PoC for #4885 

This is currently only a PoC because the code is horrible. However I couldn’t find the way in a reasonable amount of time to make it compile without cyclic module dependencies (between OpamGlobalState and OpamSysPoll).
Please feel free to push into this branch if you can think of a better way.


kit-ty-kate:
Superseded by https://github.com/ocaml/opam/pull/4892

PR #4892, opened 2021-11-05
Resolve variables before polling system

### LINKED ISSUE #4885: os-distribution

rjbou:
fix #4885

### DIFF
--- a/src/client/opamAction.ml
+++ b/src/client/opamAction.ml
@@ -602,7 +602,10 @@ let make_command st opam ?dir ?text_command (cmd, args) =
     let open OpamStd.Option.Op in
     String.concat " | " [
       OpamVersion.(to_string current);
-      (OpamSysPoll.os () +! "unknown") ^"/"^ (OpamSysPoll.arch () +! "unknown");
+      (let env = st.switch_global.global_variables in
+       Printf.sprintf "%s/%s"
+         (OpamSysPoll.os env +! "unknown")
+         (OpamSysPoll.arch env +! "unknown"));
       (OpamStd.List.concat_map " " OpamPackage.to_string
          OpamPackage.Set.(elements @@
                           inter st.compiler_packages st.installed_roots));
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -1292,13 +1292,9 @@ let config cli =
              OpamFilename.prettify
                (fst (self_upgrade_exe (OpamStateConfig.(!r.root_dir))))
            else "no");
-        print "system" "arch=%s os=%s os-distribution=%s os-version=%s"
-          OpamStd.Option.Op.(OpamSysPoll.arch () +! "unknown")
-          OpamStd.Option.Op.(OpamSysPoll.os () +! "unknown")
-          OpamStd.Option.Op.(OpamSysPoll.os_distribution () +! "unknown")
-          OpamStd.Option.Op.(OpamSysPoll.os_version () +! "unknown");
         try
           OpamGlobalState.with_ `Lock_none @@ fun gt ->
+          print "system" "%s" (OpamSysPoll.to_string gt.global_variables);
           let module Solver = (val OpamSolverConfig.(Lazy.force !r.solver)) in
           print "solver" "%s"
             (OpamCudfSolver.get_name (module Solver));
--- a/src/state/opamSwitchState.ml
+++ b/src/state/opamSwitchState.ml
@@ -178,7 +178,8 @@ module Installed_cache = OpamCached.Make(struct
     let name = "installed"
   end)
 
-let depexts_status_of_packages_raw ~depexts global_config switch_config packages =
+let depexts_status_of_packages_raw
+    ~depexts ?env global_config switch_config packages =
   let open OpamSysPkg.Set.Op in
   let syspkg_set, syspkg_map =
     OpamPackage.Set.fold (fun nv (set, map) ->
@@ -195,7 +196,7 @@ let depexts_status_of_packages_raw ~depexts global_config switch_config packages
   in
   let syspkg_set = syspkg_set -- bypass in
   let ret =
-    match OpamSysInteract.packages_status syspkg_set with
+    match OpamSysInteract.packages_status ?env syspkg_set with
     | avail, not_found ->
       let avail, not_found =
         if OpamStateConfig.(!r.no_depexts) then
@@ -516,6 +517,7 @@ let load lock_kind gt rt switch =
       lazy OpamPackage.Map.empty
     else lazy (
       depexts_status_of_packages_raw gt.config switch_config
+        ~env:gt.global_variables
         (Lazy.force available_packages)
         ~depexts:(fun package ->
             let env =
@@ -771,7 +773,7 @@ let depexts st nv =
 
 let depexts_status_of_packages st set =
   depexts_status_of_packages_raw st.switch_global.config st.switch_config set
-    ~depexts:(depexts st)
+    ~env:st.switch_global.global_variables ~depexts:(depexts st)
 
 let depexts_unavailable st nv =
   depexts_unavailable_raw (Lazy.force st.sys_packages) nv
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -97,38 +97,35 @@ type families =
   | Suse
 
 (* System status *)
-let family =
-  let family = lazy (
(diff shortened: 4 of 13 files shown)


src/state/opamSysInteract.ml
```diff
@@ -97,41 +97,37 @@ type families =
   | Suse
 
 (* System status *)
-let family =
-  let family = lazy (
-    match OpamSysPoll.os_family () with
-    | None ->
+let family ?env () =
```
kit-ty-kate, line 100:
```suggestion
let family ~env () =
```
I’d rather have the new parameter as non-optional to avoid easy tricky bugs when used

src/state/opamSwitchState.ml
```diff
@@ -738,7 +740,7 @@ let depexts st nv =
 
 let depexts_status_of_packages st set =
   depexts_status_of_packages_raw st.switch_global.config st.switch_config set
-    ~depexts:(depexts st)
+   ~env:st.switch_global.global_variables ~depexts:(depexts st)
```
kit-ty-kate, line 743:
```suggestion
    ~env:st.switch_global.global_variables ~depexts:(depexts st)
```
just a small nitpick (missing space)

src/state/opamSysInteract.ml
kit-ty-kate: (The env parameter is also present at several other places and the code currently doesn’t compile with just this suggestion)

src/client/opamCommands.ml
```diff
@@ -1307,90 +1302,94 @@ let config cli =
...
-          let process nv =
-            try
-              let conf = OpamSwitchState.package_config state nv.name in
-              let bindings =
-                let f (name, value) =
-                  (OpamVariable.Full.create nv.name name,
-                   OpamVariable.string_of_variable_contents value)
+          match OpamStateConfig.get_switch_opt () with
+          | None -> print "current-switch" "%s" "non set"
+          | Some switch ->
```
rjbou, line 1307: it's the only addition, rest is indentation

src/state/opamSysPoll.ml
```diff
@@ -143,13 +151,16 @@ let os_family_lazy = lazy (
...
 )
-let os_family () = Lazy.force os_family_lazy
+let os_family = syspoll "os-family" os_family_lazy
 
 let variables =
   List.map
     (fun (n, v) ->
        OpamVariable.of_string n,
-       lazy (Lazy.force v >>| fun v -> OpamTypes.S v))
+       lazy (try Lazy.force v >>| fun v -> OpamTypes.S v
```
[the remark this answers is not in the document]
rjbou: spliited in #5199

src/client/opamCommands.ml
rjbou: split in #5198

src/client/opamAction.ml
```diff
@@ -602,7 +602,10 @@ let make_command st opam ?dir ?text_command (cmd, args) =
     let open OpamStd.Option.Op in
     String.concat " | " [
       OpamVersion.(to_string current);
-      (OpamSysPoll.os () +! "unknown") ^"/"^ (OpamSysPoll.arch () +! "unknown");
+      (let env =st.switch_global.global_variables in
+      Printf.sprintf "%s/%s"
+      (OpamSysPoll.os ~env () +! "unknown")
+      (OpamSysPoll.arch ~env () +! "unknown"));
```
kit-ty-kate, line 608: Reformatting slightly
```suggestion
      (let env = st.switch_global.global_variables in
       Printf.sprintf "%s/%s"
       (OpamSysPoll.os ~env () +! "unknown")
       (OpamSysPoll.arch ~env () +! "unknown"));
```

kit-ty-kate (approved):
Apart from the minor reformat suggestion, LGTM.

src/state/opamSysPoll.ml
```diff
@@ -22,6 +22,14 @@ let command_output c =
 
 let norm s = if s = "" then None else Some (String.lowercase_ascii s)
 
+let syspoll var laz ?(env=OpamVariable.Map.empty) () =
```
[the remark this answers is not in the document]
rjbou: Indeed, it is more clear

rjbou:
We need the `~env` to specify the environment to look in for variables.

src/state/opamSysPoll.ml
```diff
@@ -10,6 +10,10 @@
 
 open OpamStd.Option.Op
 
+let make_lazy f =
+  let v = Lazy.from_fun f in
+  fun () -> Lazy.force v
+
```
dra27, line 16: Sorry - looking at how it's then gone below, I this could be dropped and just use `Lazy.from_fun` below...

src/state/opamSysPoll.ml
```diff
@@ -142,29 +142,42 @@ let os_family_lazy = lazy (
...
   | _ -> os_distribution ()
-)
-let os_family () = Lazy.force os_family_lazy
+let os_family = make_lazy poll_os_family
 
 let variables =
   List.map
     (fun (n, v) ->
        OpamVariable.of_string n,
        lazy (Lazy.force v >>| fun v -> OpamTypes.S v))
```
dra27, line 151: ...with...
```suggestion
       Lazy.map (Option.map OpamTypes.S))
```
...and...

src/state/opamSysPoll.ml
```diff
@@ -142,29 +142,42 @@ let os_family_lazy = lazy (
...
-      "arch", arch_lazy;
-      "os", os_lazy;
-      "os-distribution", os_distribution_lazy;
-      "os-version", os_version_lazy;
-      "os-family", os_family_lazy;
+      "arch", lazy (arch ());
+      "os", lazy (os ());
+      "os-distribution", lazy (os_distribution ());
+      "os-version", lazy (os_version ());
+      "os-family", lazy (os_family ());
```
dra27, line 157:
```suggestion
      "arch", arch;
      "os", os;
      "os-distribution", os_distribution;
      "os-version", os_version;
      "os-family", os_family;
```

src/state/opamSysPoll.ml
```diff
@@ -142,29 +142,42 @@ let os_family_lazy = lazy (
...
     ]
 
-let cores_lazy = lazy (OpamSystem.cpu_count ())
-let cores () = Lazy.force cores_lazy
+let cores = make_lazy OpamSystem.cpu_count
```
dra27, line 160: OK, maybe `make_lazy` is useful still exactly here :)

src/state/opamSysPoll.ml
```diff
@@ -142,29 +142,42 @@ let os_family_lazy = lazy (
...
+
+(* Exported functions *)
+let resolve_or_poll var poll env =
+  match OpamVariable.Full.read_from_env (OpamVariable.Full.of_string var) with
+  | Some (S c) -> Some c
+  | _ ->
+    match OpamVariable.Map.find_opt (OpamVariable.of_string var) env with
+    | Some (lazy (Some (OpamTypes.S c)), _) -> Some c
+    | _ -> poll ()
```
dra27, line 169:
```suggestion
    | _ -> Lazy.force poll
```

dra27 (commented):
A possible further disentangling of laziness!

PR #4893, opened 2021-11-05
Factorise fetching of shared archive source packages

rjbou:
Regroup same (shared) archive source packages for fetching. They are regrouped in a single action `'Fetch` node, that now takes several packages. Download functions are also updated to handle fetching an archive once and copy to sources directories.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -30,6 +30,9 @@ users)
 ## Config report
   *
 
+## Actions
+  *  Add a `'Fetch` action with several packages: one node to download once and prepare source once for packages that share same archive [#4893 @rjbou - fix #3741]
+
 ## Install
   * Make the status of pinned packages more explicit during installation [#4987 @kit-ty-kate - fix #4925]
   * Better recognize depexts on Gentoo, NetBSD, OpenBSD [#5065 @mndrix]
@@ -272,6 +275,8 @@ users)
   * Add `OpamSwitchCommand.previous_switch` [#4910 @kit-ty-kate]
 ## opam-repository
   * `OpamRepositoryConfig`: add in config record `repo_tarring` field and as an argument to config functions, and a new constructor `REPOSITORYTARRING` in `E` environment module and its access function [#5015 @rjbou]
+  * New download functions for shared source, old ones kept [#4893 @rjbou]
+
 ## opam-state
 ## opam-solver
   * `OpamCudf`: Change type of `conflict_case.Conflict_cycle` (`string list list` to `Cudf.package action list list`) and `cycle_conflict`, `string_of_explanations`, `conflict_explanations_raw` types accordingly [#4039 @gasche]
@@ -281,11 +286,14 @@ users)
   * `OpamSolver.load_cudf_universe`: change staging of `add_invariant` [#5024 @AltGr]
   * `OpamSolver.coinstallable_subset`: add `add_invariant` optional argument [#5024 @AltGr]
   * `OpamSolver.installable`: use `installable_subset` that uses `coinstallable_subset` [#5024 @kit_ty_kate]
+  * `OpamSolver.explicit`: when adding fetch nodes, add shared source ones. Change of `sources_needed` argument type [#4893 @rjbou]
 ## opam-format
   * `OpamStd.ABSTRACT`: add `compare` and `equal`, that added those functions to `OpamSysPkg` and `OpamVariable` [#4918 @rjbou]
   * Add OpamPackage.Version.default returning the version number used when no version is given for a package [#4949 @kit-ty-kate]
   * Add `OpamPath.Switch.man_dirs` [#4915 @rjbou]
   * `OpamFile.Config`: order list of installed switches according their last use, update `with_switch` accordingly, and add `previous_switch` [#4910 @AltGr]
+  * Change ``Fetch` action to take several packages, in order to handle shared fetching of packages [#4893 @rjbou]
+
 ## opam-core
   * OpamSystem: avoid calling Unix.environment at top level [#4789 @hannesm]
   * `OpamStd.ABSTRACT`: add `compare` and `equal`, that added those functions to `OpamFilename`, `OpamHash`, `OpamStd`, `OpamStd`, `OpamUrl`, and `OpamVersion` [#4918 @rjbou]
--- a/src/client/opamAction.ml
+++ b/src/client/opamAction.ml
@@ -237,33 +237,55 @@ let preprocess_dot_install st nv build_dir =
   in
   files, really_process_dot_install, config
 
-let download_package st nv =
-  log "download_package: %a" (slog OpamPackage.to_string) nv;
+let download_shared_source st url nvs =
+  let labelise pkg_str = OpamStd.List.concat_map ", " pkg_str nvs in
+  log "download_package: %a%a"
+    (slog (fun _ -> labelise OpamPackage.to_string)) ()
+    (slog (fun url -> match url, nvs with
+         | None, _ | _, [_] -> ""
+         | Some url, _ -> " " ^ OpamUrl.to_string (OpamFile.URL.url url))) url;
   if OpamStateConfig.(!r.dryrun) || OpamClientConfig.(!r.fake)
-  then Done None
+  then Done None else
+  let nvs =
+    (* filter out version-pinned packages since we already have their source *)
+    List.filter (fun nv ->
+        let dir = OpamSwitchState.source_dir st nv in
+        not (OpamPackage.Set.mem nv st.pinned &&
+             OpamFilename.exists_dir dir &&
+             OpamStd.Option.Op.(
+               OpamPinned.find_opam_file_in_source nv.name dir >>|
+               OpamFile.OPAM.safe_read >>=
+               OpamFile.OPAM.version_opt) = Some nv.version))
+      nvs
+  in
+  if nvs = [] then Done None
   else
-  let dir = OpamSwitchState.source_dir st nv in
-  if OpamPackage.Set.mem nv st.pinned &&
-     OpamFilename.exists_dir dir &&
-     OpamStd.Option.Op.(
-       OpamPinned.find_opam_file_in_source nv.name dir >>|
-       OpamFile.OPAM.safe_read >>=
-       OpamFile.OPAM.version_opt)
-     = Some nv.version
-  then Done None
-  else
-  let print_action msg =
-    OpamConsole.msg "%s retrieved %s.%s  (%s)\n"
+  let print_action =
+    OpamConsole.msg "%s retrieved %s  (%s)\n"
       (if not (OpamConsole.utf8 ()) then "->"
        else OpamActionGraph.
               (action_color (`Fetch []) (action_strings (`Fetch []))))
-      (OpamConsole.colorise `bold (OpamPackage.name_to_string nv))
-      (OpamPackage.version_to_string nv)
-      msg;
   in
-  OpamUpdate.cleanup_source st
-    (OpamPackage.Map.find_opt nv st.installed_opams)
-    (OpamSwitchState.opam st nv);
+  let print_single_actions msgs =
+    List.iter (fun (nv, msg) ->
+        print_action
+          (OpamConsole.colorise `bold (OpamPackage.name_to_string nv)
+           ^ "." ^ (OpamPackage.version_to_string nv))
+          msg)
+      msgs
+  in
+  let print_full_action msg =
+    print_action
+      (labelise @@ fun nv ->
+       (OpamConsole.colorise `bold (OpamPackage.name_to_string nv))
+       ^ "." ^ (OpamPackage.version_to_string nv))
+      msg
+  in
+  List.iter (fun nv ->
+      OpamUpdate.cleanup_source st
+        (OpamPackage.Map.find_opt nv st.installed_opams)
+        (OpamSwitchState.opam st nv))
+    nvs;
   OpamProcess.Job.catch (fun e ->
       let na =
         match e with
@@ -272,30 +294,48 @@ let download_package st nv =
       in
       Done (Some na))
   @@ fun () ->
-  OpamUpdate.download_package_source st nv dir @@| function
+  OpamUpdate.download_shared_package_source st url nvs @@| function
   | Some (Not_available (s, l)), _ ->
-    let msg = match s with None -> l | Some s -> s in
-    OpamConsole.error "Failed to get sources of %s: %s"
-      (OpamPackage.to_string nv) msg;
+    let msg = OpamStd.Option.default l s in
+    OpamConsole.error "Failed to get sources of %s%s: %s"
+      (labelise OpamPackage.to_string)
+      (match url, nvs with
+       | None, _ | _, [_] -> ""
+       | Some url, _ ->
+         Printf.sprintf " (%s)" (OpamUrl.to_string (OpamFile.URL.url url)))
+      msg;
     Some (s, l)
-  | _, ((name, Not_available (s, l)) :: _) ->
+  | _, ((nv, name, Not_available (s, l)) :: _) ->
     let msg = match s with None -> l | Some s -> s in
     OpamConsole.error "Failed to get extra source \"%s\" of %s: %s"
       name (OpamPackage.to_string nv) msg;
     Some (s, l)
   | Some (Result msg), _ ->
-    print_action msg; None
+    print_full_action msg; None
   | Some (Up_to_date msg), _ ->
-    print_action msg; None
+    print_full_action msg; None
   | None, [] -> None
   | None, (e :: es as extras) ->
-    if List.for_all (function _, Up_to_date _ -> true | _ -> false) extras then
-      print_action "cached"
-    else begin match e, es with
-      | (_, Result msg), [] -> print_action msg
-      | _, _ -> print_action (Printf.sprintf "%d extra sources" (List.length extras))
-    end;
+    if List.for_all (function _, _, Up_to_date _ -> true | _ -> false) extras then
+      print_full_action "cached"
+    else
+      (match e, es with
+       | (_, _, Result msg), [] -> print_full_action msg
+       | _, _ ->
+         print_single_actions
+           (List.map (fun (nv, _, _) ->
+                nv,
+                (Printf.sprintf "%d extra sources"
+                   (List.length
+                      (List.filter (fun (nv',_,_) ->
+                           OpamPackage.compare nv nv' = 0)
+                          extras))))
+               extras));
     None
+
+let download_package st nv =
+  download_shared_source st
+    (OpamFile.OPAM.url (OpamSwitchState.opam st nv)) [nv]
 
 (* Prepare the package build:
    * apply the patches
--- a/src/client/opamAction.ml
+++ b/src/client/opamAction.ml
@@ -256,7 +256,7 @@ let download_package st nv =
     OpamConsole.msg "%s retrieved %s.%s  (%s)\n"
       (if not (OpamConsole.utf8 ()) then "->"
        else OpamActionGraph.
-              (action_color (`Fetch ()) (action_strings (`Fetch ()))))
+              (action_color (`Fetch []) (action_strings (`Fetch []))))
       (OpamConsole.colorise `bold (OpamPackage.name_to_string nv))
       (OpamPackage.version_to_string nv)
       msg;
--- a/src/client/opamAction.mli
+++ b/src/client/opamAction.mli
@@ -14,14 +14,21 @@
 open OpamTypes
 open OpamStateTypes
 
-(** [download t pkg] downloads the source of the package [pkg] into its locally
-    cached source dir. Returns [Some (short_errmsg option, long_errmsg)] on error,
-    [None] on success. See {!OpamTypes.Not_available}.
+(** [download_package t pkg] downloads the source of the package [pkg] into its
+    locally cached source dir. Returns [Some (short_errmsg option,
+    long_errmsg)] on error, [None] on success. See {!OpamTypes.Not_available}.
 
     This doesn't update dev packages that already have a locally cached
     source. *)
 val download_package:
   rw switch_state -> package -> (string option * string) option OpamProcess.job
+
+(** [download_same_source_package t url packages]
+    As [download_package], download upstream shared source [url] between
+    [packages]. *)
+val download_shared_source:
+  rw switch_state -> OpamFile.URL.t option -> package list ->
+  (string option * string) option OpamProcess.job
 
 (** [prepare_package_source t pkg dir] updates the given source [dir] with the
     extra downloads, overlays and patches from the package's metadata
--- a/src/client/opamAction.mli
+++ b/src/client/opamAction.mli
@@ -51,7 +51,7 @@ val build_package:
   exn option OpamProcess.job
 
 (** [install_package t pkg] installs an already built package. Returns
-    [None] on success, [Some exn] on error. Do not update OPAM's
+    [None] on success, [Some exn] on error. Do not update opam's
     metadata. See {!build_package} to build the package. *)
 val install_package:
   rw switch_state -> ?test:bool -> ?doc:bool -> ?build_dir:dirname -> package ->
--- a/src/client/opamSolution.ml
+++ b/src/client/opamSolution.ml
@@ -592,14 +592,29 @@ let parallel_apply t
       | `Change _ | `Reinstall _ -> assert false
     else
     match action with
-    | `Fetch [nv] ->
-      log "Fetching sources for %s" (OpamPackage.to_string nv);
-      (OpamAction.download_package t nv @@+ function
-        | None ->
-          store_time (); Done (`Successful (installed, removed))
-        | Some (_short_error, long_error) ->
-          Done (`Exception (Fetch_fail long_error)))
-    | `Fetch _ -> assert false
+    | `Fetch nvs ->
+      log "Fetching sources for %s"
+        (OpamStd.Format.pretty_list (List.map OpamPackage.to_string nvs));
+      ((match nvs with
+          | [nv] ->
+            OpamAction.download_package t nv
+          | _ ->
+            let url =
+              match OpamStd.List.filter_map (OpamSwitchState.url t) nvs with
+              | u::_ as urls ->
+                let checksums =
+                  List.map OpamFile.URL.checksum urls
+                  |> List.flatten
+                  |> List.sort_uniq OpamHash.compare
+                in
+                Some (OpamFile.URL.with_checksum checksums u)
+              | [] -> None
+            in
+            OpamAction.download_shared_source t url nvs) @@+ function
+       | None ->
+         store_time (); Done (`Successful (installed, removed))
+       | Some (_short_error, long_error) ->
+         Done (`Exception (Fetch_fail long_error)))
 
     | `Build nv ->
       if assume_built && OpamPackage.Set.mem nv requested then
--- a/src/client/opamSolution.ml
+++ b/src/client/opamSolution.ml
@@ -459,10 +459,41 @@ let parallel_apply t
   let action_graph = (* Add build and fetch actions *)
     let noop_remove nv =
       OpamAction.noop_remove_package t nv in
-    PackageActionGraph.explicit
-      ~noop_remove
-      ~sources_needed:(fun p -> if OpamPackage.Set.mem p sources_needed then [p] else [])
-      action_graph
+    let sources_needed =
+      (* used to factorise fetches: creates [`Fetch] nodes with several
+         packages that download once the archive and copies its content in
+         source directories. Handles only http and rsync backend with at least
+         one checksum. *)
+      let shared_source =
+        OpamPackage.Set.fold (fun nv url_nvs ->
+            match OpamSwitchState.url t nv with
+            | Some url ->
+              let checksums = OpamFile.URL.checksum url in
+              let url = OpamFile.URL.url url in
+              (match url.OpamUrl.backend, checksums with
+               | (`http | `rsync ) , _::_
+                 when OpamFilename.is_archive
+                     (OpamFilename.of_string url.OpamUrl.path) ->
+                 OpamUrl.Map.update url
+                   (OpamPackage.Set.add nv)
+                   OpamPackage.Set.empty url_nvs
+               | _ -> url_nvs)
+            | None -> url_nvs)
+          sources_needed OpamUrl.Map.empty
+        |> OpamUrl.Map.values
+        |> List.filter (fun s -> not @@ OpamPackage.Set.is_singleton s)
+      in
+      log "Regroup shared source packages: %s"
+        (OpamStd.List.to_string OpamPackage.Set.to_string shared_source);
+      fun p ->
+        try
+          List.find (OpamPackage.Set.mem p) shared_source
+          |> OpamPackage.Set.elements
+        with Not_found ->
+        try [ OpamPackage.Set.find (OpamPackage.equal p) sources_needed ]
+        with Not_found -> []
+    in
+    PackageActionGraph.explicit ~noop_remove ~sources_needed action_graph
   in
   let action_graph =
     if download_only then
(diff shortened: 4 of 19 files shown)


tests/reftests/shared-fetch.test
```diff
@@ -0,0 +1,117 @@
...
+Now run 'opam upgrade' to apply any package updates.
+### opam install no-extra extra non-shared | unordered
+The following actions will be performed:
+  - recompile no-extra   3
+  - recompile extra      3
+  - recompile non-shared 4
+===== 3 to recompile =====
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> retrieved extra.3+no-extra.3  (file://${BASEDIR}/shared.tgz)
```
kit-ty-kate, line 109:
```suggestion
-> retrieved extra.3 & no-extra.3  (file://${BASEDIR}/shared.tgz)
```
I think it would look more consistent if the packages are separated by a space at least, to not confuse it with the package `extra` at version `3+no-extra.3`. Also it looks a bit prettier in my opinion.

rjbou: It will break the "rule" of one single word after action name. But I don't know if it used somewhere...

kit-ty-kate: If it’s really an issue, could there not be two kinds of name "internal" and "pretty-printed" ?

tests/reftests/shared-fetch.test
```diff
@@ -0,0 +1,117 @@
...
+for p in $@; do
+  file="REPO/packages/${p%.*}/$p/opam"
+  echo "url {" >> $file
+  echo "src: \"${BASEDIR}/$arch.tgz\"" >> $file
+  MD5=$(openssl md5 $arch.tgz | cut -d' ' -f2)
+  echo "checksum: \"md5=$MD5\"" >> $file
+  echo "}" >> $file
+done
+### sh mkurl.sh shared no-extra.3 extra.3
+### sh mkurl.sh non-shared non-shared.4
```
kit-ty-kate, line 35: It looks like this test needs to be duplicated for it to pass on (non-cygwin) Windows
```
diff --git a/tests/reftests/shared-fetch.test b/tests/reftests/shared-fetch.out
index faf44ff3..2bcf991a 100644
--- a/tests/reftests/shared-fetch.test
+++ b/tests/reftests/shared-fetch.out
@@ -38,22 +38,17 @@ done
 
 [default] synchronised from file://${BASEDIR}/REPO
-Now run 'opam upgrade' to apply any package updates.
+[ERROR] At ${OPAMTMP}/default/packages/extra/extra.3/opam:8:9-8:9::
+        illegal escape sequence
+[ERROR] Strict mode: aborting
+[ERROR] Could not update repository "default": OpamStd.OpamSys.Exit(30)
+# Return code 40 #
 ### opam switch create shared --empty
 ### opam install no-extra extra non-shared | unordered
-The following actions will be performed:
-  - install extra      3
-  - install non-shared 4
-  - install no-extra   3
-===== 3 to install =====
-
--> retrieved extra.3+no-extra.3  (file://${BASEDIR}/shared.tgz)
--> retrieved non-shared.4  (file://${BASEDIR}/non-shared.tgz)
--> installed extra.3
--> installed no-extra.3
--> installed non-shared.4
-Done.
+[ERROR] No package named no-extra found.
+[ERROR] No package named extra found.
+[ERROR] No package named non-shared found.
+# Return code 5 #
```

rjbou: It seems to come from `${BASEDIR}`, cf  [this run](https://ci.appveyor.com/project/rjbou/opam/builds/41459574/job/emefj30g2gcif7eo#L2673)
```
### cat REPO/packages/extra/extra.3/opam
+opam-version: "2.0"
+build: [
+  [ "ocaml" "shared.ml" ]
+  [ "ocaml" "extra_uniq.ml" ]
+]
+#patches: ["shared.patch"]
+url {
+src: "${BASEDIR}/shared.tgz"
+checksum: "md5=a13945d953fc7ab20f4f0b80f00211cb"
+}

File: master_changes.md [hunk 3]
```diff
@@ -27,6 +27,9 @@ users)
 ## Config report
   *
 
+
+## Actions
+  *  Add a `'Fetches' action to share fetches: one node to download once and prepare source once for packages that share same archive [#XXX @rjbou - fix #XXX]
```
kit-ty-kate, line 32:
```suggestion
## Actions
  *  Add a `'Fetches' action to share fetches: one node to download once and prepare source once for packages that share same archive [#4893 @rjbou - fix #3741]
```

master_changes.md
```diff
@@ -198,6 +201,8 @@ users)
 ## opam-client
   * `OpamStd.ABSTRACT`: add `compare` and `equal`, that added those functions to `OpamCLIVersion` [#4918 @rjbou]
 ## opam-repository
+  * New download functions, old ones kept [#XXX rjbou]
+  * Add `OpamRepository.is_cached` [#XXX]
```
kit-ty-kate, line 205:
```suggestion
  * New download functions, old ones kept [#4893 @rjbou]
  * Add `OpamRepository.is_cached` [#4893 @rjbou]
```

src/client/opamAction.ml
```diff
@@ -237,33 +237,55 @@ let preprocess_dot_install st nv build_dir =
   in
   files, really_process_dot_install, config
 
-let download_package st nv =
-  log "download_package: %a" (slog OpamPackage.to_string) nv;
+let download_same_source_packages st url nvs =
+  let labelise pkg_str = OpamStd.List.concat_map "+" pkg_str nvs in
```
kit-ty-kate, line 241:
```suggestion
  let labelise pkg_str = OpamStd.List.concat_map " & " pkg_str nvs in
```

src/repository/opamRepository.ml
```diff
@@ -238,20 +243,66 @@ let pull_from_mirrors label ?working_dir ?subpath cache_dir destdir checksums ur
...
+            in
+            let long =
+              Printf.sprintf "Failed to copy source of:\n%s"
+                (OpamStd.Format.itemize (fun (nv, msg) ->
+                     Printf.sprintf "%s: %s" nv msg)
+                    failing)
+            in
+            Done (Not_available (Some simple, long)))
   in
+  let label = OpamStd.List.concat_map "+" fst dirnames in
```
kit-ty-kate, line 305:
```suggestion
  let label = OpamStd.List.concat_map " & " fst dirnames in
```

src/solver/opamActionGraph.ml
```diff
@@ -103,6 +120,9 @@ module MakeAction (P: GenericPackage) : ACTION with type package = P.t
   let to_string a = match a with
     | `Remove p | `Install p | `Reinstall p | `Build p | `Fetch p ->
       Printf.sprintf "%s %s" (action_strings a) (P.to_string p)
+    | `Fetches (p, pl) ->
+      Printf.sprintf "%s %s" (action_strings a)
+        (OpamStd.List.concat_map "+" P.to_string (p::pl))
```
kit-ty-kate, line 125:
```suggestion
        (OpamStd.List.concat_map " & " P.to_string (p::pl))
```

kit-ty-kate (approved):
Looks good apart from a minor remark (UI change)

src/client/opamAction.ml
```diff
@@ -237,33 +237,55 @@ let preprocess_dot_install st nv build_dir =
...
-     OpamFilename.exists_dir dir &&
-     OpamStd.Option.Op.(
-       OpamPinned.find_opam_file_in_source nv.name dir >>|
-       OpamFile.OPAM.safe_read >>=
-       OpamFile.OPAM.version_opt)
-     = Some nv.version
-  then Done None
+  then Done None else
+  let nvs =
+    (* we pinned remove package that already has their source *)
```
rjbou, line 250:
```suggestion
    (* filter out version-pinned packages since we already have their source *)
```

src/solver/opamActionGraph.ml
```diff
@@ -52,10 +53,14 @@ let symbol_of_action =
   | `Build _ ->
       utf8_symbol Symbols.greek_small_letter_lambda
                   ~alternates:[Symbols.six_pointed_black_star] "B"
-  | `Fetch _ ->
+  | `Fetch _  ->
```
rjbou, line 56:
```suggestion
  | `Fetch _ ->
```

src/client/opamSolution.ml
```diff
@@ -474,6 +478,79 @@ let parallel_apply t
...
+               in
+               PackageActionGraph.remove_edge g v1 v2;
+               PackageActionGraph.add_edge g fetches v2;
+               v1 :: vertexes
+             | None -> vertexes)
+          | _ -> vertexes)
+        action_graph []
+    in
+    List.iter (PackageActionGraph.remove_vertex g) removed_vertexes;
+    g
```
AltGr, line 554: Unnecessarily complex? I'd rather iter on`packages` directly. And removing the connected edges is automatic when you remove a vertex.
```suggestion
    List.iter (fun nvs ->
      let v =
        let n, nvs =
          match OpamPackage.Set.elements set with
          | n::(_::_ as nvs) -> n, nvs
          | _ -> assert false
        in
        `Fetches (n, nvs)
      in
      OpamPackage.Set.iter (fun p ->
          let v1 = `Fetch p in
          PackageActionGraph.iter_succ (PackageActionGraph.add_edge g v) g v1;
          PackageActionGraph.remove_vertex g v1)
        nvs)
      packages;
    g
```

src/client/opamSolution.ml
```diff
@@ -474,6 +478,79 @@ let parallel_apply t
...
+                       OpamPackage.Set.empty hashes)
+                   hashes checksums;
+               | _ -> acc)
+            | None -> acc)
+          sources_needed (OpamUrl.Map.empty, OpamHash.Map.empty)
+      in
+      OpamUrl.Map.fold (fun _url hashes packages ->
+          (* should we check urls ? *)
+          let _, choosen =
+            (* should we keep other checksums or drop them ? *)
```
AltGr, line 513: The general rule is that caching/indexing may behave differently, but all of them are verified when checking. We're not allowed to drop!
This raises a question when there are multiple sources with different hashes: here we will now have to merge all hashes, so from a set of packages with the same source, adding a wrong hash to one of them could "contaminate" the others and prevent installation for all of them. It seems alright though, I don't see how we could factorise without doing it like this.

src/client/opamSolution.ml
```diff
@@ -474,6 +478,79 @@ let parallel_apply t
...
+                   OpamHash.Set.empty urls,
+                 List.fold_left (fun hashes h ->
+                     OpamHash.Map.update h (OpamPackage.Set.add nv)
+                       OpamPackage.Set.empty hashes)
+                   hashes checksums;
+               | _ -> acc)
+            | None -> acc)
+          sources_needed (OpamUrl.Map.empty, OpamHash.Map.empty)
+      in
+      OpamUrl.Map.fold (fun _url hashes packages ->
```
AltGr, line 510: Folding by url is important here: otherwise there could be an attack using a collision on a weak hash (md5) on a different package/url. It would be detected later on since we check all hashes upon download, but would still deny installation…

src/repository/opamRepository.mli
```diff
@@ -36,6 +36,14 @@ val pull_tree:
   ?subpath:string -> dirname -> OpamHash.t list -> url list ->
   string download OpamProcess.job
 
+val pulls_tree:
```
AltGr, line 39: add doc / more explicit name ? It's not clear what that `s` means.

AltGr (changes requested):
A few small changes required here and there. Most important, ensure that all supplied checksums, aggregated from merged `url` entries, are verified upon download.

src/client/opamAction.ml
```diff
@@ -237,33 +237,55 @@ let preprocess_dot_install st nv build_dir =
   in
   files, really_process_dot_install, config
 
-let download_package st nv =
-  log "download_package: %a" (slog OpamPackage.to_string) nv;
+let download_same_source_packages st url nvs =
+  let labelise pkg_str = OpamStd.List.concat_map " & " pkg_str nvs in
```
kit-ty-kate, line 241:
```suggestion
  let labelise pkg_str = OpamStd.List.concat_map ", " pkg_str nvs in
```

src/repository/opamRepository.ml
```diff
@@ -245,20 +250,66 @@ let pull_from_mirrors label ?working_dir ?subpath cache_dir destdir checksums ur
...
+            in
+            let long =
+              Printf.sprintf "Failed to copy source of:\n%s"
+                (OpamStd.Format.itemize (fun (nv, msg) ->
+                     Printf.sprintf "%s: %s" nv msg)
+                    failing)
+            in
+            Done (Not_available (Some simple, long)))
   in
+  let label = OpamStd.List.concat_map " & " fst dirnames in
```
kit-ty-kate, line 312:
```suggestion
  let label = OpamStd.List.concat_map ", " fst dirnames in
```

src/solver/opamActionGraph.ml
```diff
@@ -103,6 +120,9 @@ module MakeAction (P: GenericPackage) : ACTION with type package = P.t
   let to_string a = match a with
     | `Remove p | `Install p | `Reinstall p | `Build p | `Fetch p ->
       Printf.sprintf "%s %s" (action_strings a) (P.to_string p)
+    | `Fetches (p, pl) ->
+      Printf.sprintf "%s %s" (action_strings a)
+        (OpamStd.List.concat_map " & " P.to_string (p::pl))
```
kit-ty-kate, line 125:
```suggestion
        (OpamStd.List.concat_map ", " P.to_string (p::pl))
```

tests/reftests/shared-fetch.test
```diff
@@ -0,0 +1,117 @@
...
+### opam switch create shared --empty
+### opam install no-extra extra non-shared | unordered
+The following actions will be performed:
+  - install extra      3
+  - install non-shared 4
+  - install no-extra   3
+===== 3 to install =====
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> retrieved extra.3 & no-extra.3  (file://${BASEDIR}/shared.tgz)
```
kit-ty-kate, line 51:
```suggestion
-> retrieved extra.3, no-extra.3  (file://${BASEDIR}/shared.tgz)
```

tests/reftests/shared-fetch.test
```diff
@@ -0,0 +1,117 @@
...
+Now run 'opam upgrade' to apply any package updates.
+### opam install no-extra extra non-shared | unordered
+The following actions will be performed:
+  - recompile no-extra   3
+  - recompile extra      3
+  - recompile non-shared 4
+===== 3 to recompile =====
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> retrieved extra.3 & no-extra.3  (file://${BASEDIR}/shared.tgz)
```
kit-ty-kate, line 109:
```suggestion
-> retrieved extra.3, no-extra.3  (file://${BASEDIR}/shared.tgz)
```

kit-ty-kate (commented):
Suggestion from @AltGr

src/client/opamSolution.ml
```diff
@@ -223,13 +226,14 @@ let display_error (n, error) =
...
-  | `Build nv          -> f "compiling" nv
-  | `Fetch nv          -> f "fetching sources for" nv
+  | `Change (`Up, _, nv)   -> f "upgrading to" [nv]
+  | `Change (`Down, _, nv) -> f "downgrading to" [nv]
+  | `Install nv        -> f "installing" [nv]
+  | `Reinstall nv      -> f "recompiling" [nv]
+  | `Remove nv         -> f "removing" [nv]
+  | `Build nv          -> f "compiling" [nv]
+  | `Fetch nv          -> f "fetching sources for" [nv]
+  | `Fetches (nv,nvs)    -> f "multi fetching sources for" (nv::nvs)
```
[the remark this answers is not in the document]
rjbou: If it is the same message, there is no difference (and user notification), that what failed is the fetching of this unique source for all theses packages. But maybe we don't need it?

src/client/opamSolution.ml
```diff
@@ -474,6 +478,79 @@ let parallel_apply t
...
+            match OpamSwitchState.url t nv with
+            | Some url ->
+              let checksums = OpamFile.URL.checksum url in
+              let url = OpamFile.URL.url url in
+              (match url.OpamUrl.backend, checksums with
+               | (`http | `rsync ) , _::_
+                 when OpamFilename.is_archive
+                     (OpamFilename.of_string url.OpamUrl.path) ->
+                 OpamUrl.Map.update url
+                   (fun hset -> OpamHash.Set.(union hset (of_list checksums)))
```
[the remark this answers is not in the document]
rjbou: héhé

src/state/opamUpdate.ml
```diff
@@ -520,37 +520,57 @@ let cleanup_source st old_opam_opt new_opam =
...
-  let opam = OpamSwitchState.opam st nv in
+let download_package_source_t st url nvs =
   let cache_dir = OpamRepositoryPath.download_cache st.switch_global.root in
-  let cache_urls = active_caches st nv in
-
+  let cache_urls =
+    List.map fst nvs
+    |> List.map (active_caches st)
+    |> List.flatten
+    |> List.sort_uniq compare
```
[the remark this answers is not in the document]
rjbou: indeed... Sort came from uniq

src/format/opamTypes.mli
```diff
@@ -238,6 +238,7 @@ type 'a concrete_action = [
   | 'a atomic_action
   | `Build of 'a
   | `Fetch of 'a
+  | `Fetches of 'a * 'a list
```
kit-ty-kate, line 241:
```suggestion
  | `Fetch of 'a * 'a list
```
Any particular reason you kept the singular `Fetch` action?

src/client/opamSolution.ml
rjbou: In fact there was something wrong: only the url of the first/chosen package is kept, so if another package define another checksums, they are not checked.
We need to check all checksums for each package: after downloading the archive, before copying its content in source directory, check package defined checksums, and fail otherwise.

AltGr: yes definitely 2. ; it's mitigated by the fact that it's only true for the package you are trying to install in the current run.

src/format/opamTypes.mli
AltGr: Hm yes could be a good idea; depends how inconvenient it is in the cases where you used to match on the `Fetch` action and would have to assert that the list is empty. Are there even any such cases ?

src/client/opamSolution.ml
rjbou: outdated

src/client/opamSolution.ml
```diff
@@ -474,6 +478,79 @@ let parallel_apply t
...
+      OpamUrl.Map.fold (fun _url hashes packages ->
+          (* should we check urls ? *)
+          let _, choosen =
+            (* should we keep other checksums or drop them ? *)
+            OpamHash.Set.fold (fun hash ((num_nv, _nvs) as acc) ->
+                match OpamHash.Map.find_opt hash hash_nvs with
+                | Some nvset ->
+                  let number = OpamPackage.Set.cardinal nvset in
+                  if number > num_nv &&
+                     OpamRepository.is_cached cache_dir [hash] = [] then
```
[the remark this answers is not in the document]
rjbou: outdated

src/repository/opamRepository.mli
rjbou: it need another name, if you have ideas...

src/client/opamAction.mli
```diff
@@ -14,15 +14,22 @@
...
 
     This doesn't update dev packages that already have a locally cached
     source. *)
 val download_package:
   rw switch_state -> package -> (string option * string) option OpamProcess.job
 
+(** [download_same_source_package t url packages]
+    As [download_package], download upstream shared source [url] between
+    [packages]. *)
+val download_same_source_packages:
```
rjbou, line 29: another name ?

src/state/opamUpdate.mli
```diff
@@ -82,6 +82,14 @@ val download_package_source:
   'a switch_state -> package -> dirname ->
   (string download option * (string * string download) list) OpamProcess.job
 
+(** As [download_package_source] but for several packages sharing the same
+    source. If [url] is None, do nothing. Downloads and synchronise upstream
+    source in their respective source directories. *)
+val download_same_packages_source:
```
rjbou, line 88: another name?

rjbou:
Updated! @AltGr

tests/reftests/shared-fetch.test
```diff
@@ -0,0 +1,240 @@
...
+          - md5=a-given-hash (MISMATCH)
+          - md5=a-given-hash (match)
+          - sha512=a-given-hash (match)
+
+[ERROR] extra.3, intra.3, out.3: Checksum mismatch for file://${BASEDIR}/shared.tgz:
+          expected md5=a-given-hash
+          got      md5=a-given-hash
+[ERROR] Failed to get sources of extra.3, intra.3, out.3 (file://${BASEDIR}/shared.tgz): Checksum mismatch
+
+OpamSolution.Fetch_fail("Checksum mismatch")
```
kit-ty-kate, line 232: Can’t we catch this and fail gracefully instead? Also I think the three errors above are a bit redundant

tests/reftests/shared-fetch.test
```diff
@@ -0,0 +1,240 @@
...
+
+
+<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
++- The following actions failed
+| - fetch extra 3 3 3
```
kit-ty-kate, line 237: why is there `3 3 3` here? This seems a bit odd

rjbou: good catch!

tests/reftests/shared-fetch.test
rjbou: for that we need to fully rework error handling system...

src/format/opamTypes.mli
```diff
@@ -237,7 +237,7 @@ type 'a inst_action = [
 type 'a concrete_action = [
   | 'a atomic_action
   | `Build of 'a
-  | `Fetch of 'a
+  | `Fetch of 'a * 'a list
```
rjbou, line 240: in fact we can even have a `'Fetch of 'a list`. It will just lead to some empty list checks, instead of always merge `'Fetch (p,pl) -> let pl = p::pl in...`

kit-ty-kate: that's a good idea, i think it would be easier to use this way

rjbou: updated

src/repository/opamRepository.ml
```diff
@@ -245,20 +245,66 @@ let pull_from_mirrors label ?working_dir ?subpath cache_dir destdir checksums ur
...
+      fun archive msg ->
+        OpamFilename.with_tmp_dir_job @@ fun tmpdir ->
+        let copies () =
+          OpamParallel.map ~jobs:3
+            ~command:(fun (label, local_dirname) ->
+                let text = OpamProcess.make_command_text label label in
+                OpamProcess.Job.with_text text @@
+                (try
+                   OpamFilename.cleandir local_dirname;
+                   OpamFilename.copy_dir ~src:tmpdir ~dst:local_dirname;
```
AltGr, line 278: I am wondering if, in terms of performance, it may not actually be better to extract n times rather than extract once and copy… In any case extracting directly to one of the `dirnames` rather than a tmpdir could remove a step, and enable to discard the special case for a single target dir ?

src/repository/opamRepository.mli
AltGr: `pull_multiple_trees` ? `pull_tree_multiple` ? what's most bothering me is that where it is placed, the `s` doesn't seem like a plural.

src/solver/opamActionGraph.ml
```diff
@@ -82,8 +82,18 @@ module MakeAction (P: GenericPackage) : ACTION with type package = P.t
     | `Install p, `Install q
     | `Reinstall p, `Reinstall q
     | `Build p, `Build q
-    | `Fetch p, `Fetch q
+    | `Fetch [p], `Fetch [q]
       -> P.compare p q
```
AltGr, line 86: This case can be removed ?

src/solver/opamActionGraph.ml
```diff
@@ -82,8 +82,18 @@ module MakeAction (P: GenericPackage) : ACTION with type package = P.t
...
+    | `Fetch pl, `Fetch ql ->
+      let rec aux comp pl ql =
+        if comp <> 0 then comp else
+        match pl, ql with
+        | [], [] -> 0
+        | [], _::_ -> -1
+        | _::_, [] -> 1
+        | p::pl, q::ql -> aux (P.compare p q) pl ql
+      in
+      aux 0 pl ql
```
AltGr, line 96: Hopefully the case shouldn't happen, but to be complete we should probably compare up to package ordering ? E.g something like 
```suggestion
      aux 0 (List.sort P.compare pl) (List.sort P.compare ql)
```
?
Well or use sets everywhere and their compare function directly, but that would probably prove extremely inconvenient everywhere else, especially for matching.

src/solver/opamActionGraph.ml
```diff
@@ -111,22 +124,34 @@ module MakeAction (P: GenericPackage) : ACTION with type package = P.t
...
+          pkg_to_string p
+        | `Fetch (p::pl) ->
+          let append _ = "," in
+          let rec aux str p = function
+            | [] -> pkg_to_string p::str
+            | p'::pl -> aux (pkg_to_string ~append p::str) p' pl
+          in
+          aux [] p pl
+          |> List.rev
+          |> List.flatten
```
AltGr, line 147: I don't think it matters since this function is used for printing the actions summary, which doesn't include fetches; nevertheless, this should remain a single "table cell" (with a `String.concat " "`) rather than multiple table cells. Especially since we will be appending more columns to the right (for action reasons), the returned `list list` is expected to be a matrix, i.e. all rows of the same length

src/state/opamUpdate.mli
AltGr: hm, maybe something with `shared` as suggested above, `download_shared_package_source` ?

tests/reftests/shared-fetch.test
AltGr: I agree with @kit-ty-kate here, but the situation is almost the same before this PR ([thanks to tests added before code :+1:](https://github.com/ocaml/opam/pull/4893/commits/ca040c65b126b9132e0c13fe48488d9964800e6b#diff-4ff8f50ddeb5f06fc5b5c42cf768bad7a2082bf230668ae0f2e8735081ae6ec9R236-R246))

Probably worth opening an issue. I could live with the redundant message — it's not something that you should see often after all — but the `OpamSolution.Fetch_fail` really is ugly.
IIRC a long time ago we tried to get rid of some of the redundant messages, but it's very difficult to ensure that you won't end up in cases where the message you removed was actually the only one appearing and as a result the error is not printed at all :shrug:

AltGr (approved):
Still have a few remarks, but good to go!

src/repository/opamRepository.ml
rjbou: I kept the same old behavior of : extract in a temp directory then copy to dirname. See discussion in #5018

src/solver/opamActionGraph.ml
rjbou: it can yes... i kept it to not define/cal la function in case of single package, which is most of the cases. But maybe non sens  optim

src/client/opamAction.mli
rjbou:
```suggestion
val download_shared_source:
```

src/client/opamSolution.ml
```diff
@@ -834,28 +889,31 @@ let parallel_apply t
       (* Cleanup build/install actions when one of them failed, it's verbose and
          doesn't add information *)
       let successful =
+        let suceed p = not @@ List.mem (`Install p) failed in
```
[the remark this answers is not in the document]
rjbou:
```suggestion
        let was_successful p = not @@ List.mem (`Install p) failed in
```

src/solver/opamActionGraph.ml
rjbou: noted!

src/state/opamUpdate.ml
```diff
@@ -479,29 +479,37 @@ let pinned_packages st ?(working_dir=OpamPackage.Name.Set.empty) names =
...
-      OpamStd.List.filter_map (fun rel ->
-          if OpamStd.String.contains ~sub:"://" rel
-          then
-            let r = OpamUrl.parse_opt ~handle_suffix:false rel in
-            if r = None then
-              OpamConsole.warning "Invalid cache url %s, skipping" rel;
-            r
-          else Some OpamUrl.Op.(root_url / rel))
-        (OpamFile.Repo.dl_cache repo_def)
+    List.fold_left (fun (repos, caches as acc) nv ->
```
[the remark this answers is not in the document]
rjbou: i do it according packages order, which is from the set...

src/state/opamUpdate.ml
```diff
@@ -525,37 +533,52 @@ let cleanup_source st old_opam_opt new_opam =
     OpamFilename.rmdir
       (OpamSwitchState.source_dir st (OpamFile.OPAM.package new_opam))
 
-let download_package_source st nv dirname =
-  let opam = OpamSwitchState.opam st nv in
+let download_package_source_t st url nvs =
```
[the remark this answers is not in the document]
rjbou:
```suggestion
let download_package_source_t st url nv_dirs =
```

src/state/opamUpdate.mli
rjbou:
```suggestion
val download_shared_package_source:
```

src/repository/opamRepository.ml
AltGr: ha! Indeed; so if we want to optimise we'll need to both extract many times and optimise extraction to avoid the copy. Does Kate's PR already do this ? an option could be to have the tmp dir besides the target dir (to ensure they're on the same fs) then move instead of copy…

kit-ty-kate (approved):

src/repository/opamRepository.mli
AltGr: `pulls` is still not a plural :laughing:

src/solver/opamActionGraph.ml
```diff
@@ -111,19 +123,26 @@ module MakeAction (P: GenericPackage) : ACTION with type package = P.t
...
         :: name_of_action a
-        :: OpamConsole.colorise `bold
-          (P.name_to_string (OpamTypesBase.action_contents a))
         :: match a with
-        | `Remove p | `Install p | `Reinstall p | `Build p | `Fetch p ->
-          (P.version_to_string p ^ append p) :: []
+        | `Remove p | `Install p | `Reinstall p | `Build p | `Fetch [p] ->
+          pkg_to_string p
+        | `Fetch pl ->
+          (List.map P.to_string pl |> String.concat ", ")
```
AltGr, line 140:
```suggestion
          (OpamStd.List.concat_map ", " P.to_string pl)
```

src/repository/opamRepository.mli
rjbou: oh, i forgot about this one. Maybe `pull_shared_tree`? but it is not exactly that...

src/solver/opamActionGraph.mli
```diff
@@ -43,12 +43,16 @@ module type SIG = sig
       The argument [noop_remove] is a function that should return `true`
       for package where the `remove` action is known not to modify the
       filesystem (such as `conf-*` package).
-      The argument [sources_needed] is a function that should return `true`
+      The argument [sources_needed p] is a function that should return `true`
       for packages that require fetching sources (packages that do not
```
AltGr, line 47:
```suggestion
```

kit-ty-kate:
Thanks a lot!

PR #4899, opened 2021-11-08
repo: don't display global message when `this-switch' is given

### LINKED ISSUE #4889: opam repo add: silence warning when --this-switch is explicitly passed
Hi,

When using `opam repo add NAME URL`, the following message gets printed:

> Repository NAME has been added to the selections of switch SWITCH only.
> Run `opam repository add NAME --all-switches|--set-default' to use it in all existing switches, or in newly created switches, respectively.

I would expect that explicitly passing `--this-switch` (equivalent to the default behavior) would silence this warning, but that is not the case. There does not seem to be a way to silence it.

What do you think about printing the warning only if this behavior is selected by default?

For context, the mirage 4 client is generating this in build scripts.

rjbou:
fix #4889

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -2067,7 +2067,7 @@ let repository cli =
       mk_vflag_all ~cli ~section:scope_section [
         cli_original, `No_selection, ["dont-select"],
           "Don't update any selections";
-        cli_original, `Current_switch, ["this-switch"],
+        cli_original, `This_switch, ["this-switch"],
           "Act on the selections for the current switch (this is the default)";
         cli_original, `Default, ["set-default"],
           "Act on the default repository selection that is used for newly \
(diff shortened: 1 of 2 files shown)


src/client/opamCommands.ml
```diff
@@ -2270,8 +2270,9 @@ let repository cli =
       let global = List.mem `Default scope in
       let switches =
         if scope = [] ||
-           List.exists (function `Current_switch | `Switch _ -> true
-                               | _ -> false)
+           List.exists (function
+               | `This_switch | `Current_switch | `Switch _ -> true
+               | _ -> false)
```
kit-ty-kate, line 2275: Side note: we might want to get rid of those dangerous `| _ -> ...` patterns at some point in the future (i.e. enable `-w 4` (fragile pattern matching))

kit-ty-kate (approved):
Tested successfully. Thanks!

PR #4903, opened 2021-11-10
More accurate error message on not found (package) variable with `opam var`

### LINKED ISSUE #4889: opam repo add: silence warning when --this-switch is explicitly passed

rjbou:
See test file for outputs
fix #4889


src/client/opamConfigCommand.ml
```diff
@@ -375,6 +375,10 @@ let switch_doc switch =
   Printf.sprintf "switch %s"
     (OpamConsole.colorise `bold (OpamSwitch.to_string switch))
 
+let no_self () =
```
rjbou, line 378:
```suggestion
let no_self_variable_error () =
```

kit-ty-kate (approved):

dra27 (commented):
I pushed a "suggestion" as cb51e895a402bd71a42d735807a32424628590c5 as it would have spammed the review with comments to do it with "review suggestions" - if you don't like it, please drop it at the next rebase!

I think it's a nice usability/consistency improvement! (_edit_: "it's" = "the PR overall", not the commit I added! 🙂)

PR #4904, opened 2021-11-10
Warn when setting a variable that shadow an option

### LINKED ISSUE #4730: Global variables should not shadow options
`opam var --global jobs=42` sets the global variable `jobs` which creates a strange state for opam in that the `jobs` variable is now `42` but opam will only schedule up to the _previous_ value of `jobs`.

The correct course of action here is `opam option jobs=42` - `opam var` should not permit shadowing the options.

rjbou:
ask user what they want to do
fix #4730

### DIFF
--- a/src/client/opamConfigCommand.ml
+++ b/src/client/opamConfigCommand.ml
@@ -936,6 +936,19 @@ type ('var,'config) var_confset =
   }
 
 let set_var var value conf =
+  let svar = OpamVariable.Full.to_string var in
+  let global_option, switch_option =
+    let exists f = List.exists (fun (name, _) -> String.equal svar name) f in
+    (exists OpamFile.Config.fields),
+    (exists OpamFile.Switch_config.fields)
+  in
+  if global_option || switch_option then
+    (let scope = if global_option then "global" else "switch" in
+     OpamConsole.warning
+       "You are setting a variable that has the same name as a %s option.\n\
+        Did you mean to use `opam option` instead? You can revert this variable \
+        using: 'opam var %s= --%s'\n"
+       scope svar scope);
   let conf = conf (OpamVariable.Full.variable var) in
   let global_vars = conf.stv_vars in
   let rest = List.filter (fun v -> not (conf.stv_find v)) global_vars in
--- a/tests/reftests/var-option.test
+++ b/tests/reftests/var-option.test
@@ -17,12 +17,8 @@ Removed variable group in switch var-option
 ### opam option sys-pkg-manager-cmd= --global | grep -v configuration
 ### opam var arch=x86_64 --global
 Added '[arch "x86_64" "Set through 'opam var'"]' to field global-variables in global configuration
-### opam var jobs=7 --global
-Added '[jobs "7" "Set through 'opam var'"]' to field global-variables in global configuration
 ### opam var make=make --global
 Added '[make "make" "Set through 'opam var'"]' to field global-variables in global configuration
-### opam var opam-version=68.79 --global
-Added '[opam-version "68.79" "Set through 'opam var'"]' to field global-variables in global configuration
 ### opam var os=linux --global
 Added '[os "linux" "Set through 'opam var'"]' to field global-variables in global configuration
 ### opam var os-distribution=lorem --global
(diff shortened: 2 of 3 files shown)


kit-ty-kate (approved):

tests/reftests/var-option.test
```diff
@@ -428,8 +426,8 @@ Removed variable dolore in global configuration
 ### opam option eval-variables
 []
 ### opam option global-variables
-[[foo "global-foo" "Set through 'opam var'"] [bin "global-bin" "Set through 'opam var'"] [os-version "dolor" "Set through 'opam var'"] [os-family "ipsum" "Set through 'opam var'"] [os-distribution "lorem" "Set through 'opam var'"] [os "linux" "Set through 'opam var'"] [opam-version "68.79" "Set through 'opam var'"] [make "make" "Set through 'opam var'"] [jobs "7" "Set through 'opam var'"] [arch "x86_64" "Set through 'opam var'"] [sys-ocaml-version "4.08.0" "Set through 'opam var'"]]
-### # Check uniable operations
+[[foo "global-foo" "Set through 'opam var'"] [bin "global-bin" "Set through 'opam var'"] [os-version "dolor" "Set through 'opam var'"] [os-family "ipsum" "Set through 'opam var'"] [os-distribution "lorem" "Set through 'opam var'"] [os "linux" "Set through 'opam var'"] [make "make" "Set through 'opam var'"] [arch "x86_64" "Set through 'opam var'"] [sys-ocaml-version "4.08.0" "Set through 'opam var'"]]
+### # Check unable operations
```
dra27, line 430: Exceedingly minor, but "unavailable" or "disallowed"?

src/client/opamConfigCommand.ml
```diff
@@ -779,6 +779,20 @@ type ('var,'config) var_confset =
...
+    List.exists (fun (name, _) -> String.equal svar name) OpamFile.Config.fields
+  in
+  let switch_option =
+    List.exists (fun (name, _) -> String.equal svar name)
+      OpamFile.Switch_config.fields
+  in
+  if global_option || switch_option then
+    (OpamConsole.msg
+       "You are setting a variable that have the same name than a %s option.\n\
+        Maybe you want to use 'opam option'\n"
```
dra27, line 792:
```suggestion
       "You are setting a variable that has the same name as a %s option.\n\
        Maybe you want to use 'opam option' instead\n"
```

src/client/opamConfigCommand.ml
```diff
@@ -779,6 +779,20 @@ type ('var,'config) var_confset =
...
 let set_var svar value conf =
+  let global_option =
+    List.exists (fun (name, _) -> String.equal svar name) OpamFile.Config.fields
+  in
+  let switch_option =
+    List.exists (fun (name, _) -> String.equal svar name)
+      OpamFile.Switch_config.fields
+  in
+  if global_option || switch_option then
+    (OpamConsole.msg
```
dra27, line 790: Slightly concerned about compatibility - although it's almost guaranteed to be the case that setting, say, `jobs` with `opam var` was a mistake, there could be situations where this is correct. Can the question be guarded either for when we're in interactive mode or for when we're in the default or 2.2 CLI?

dra27 (commented):
Wondering if this should be a warning, rather than a prompt?

tests/reftests/var-option.test
rjbou:
```suggestion
### # Check disallowed operations
```

src/client/opamConfigCommand.ml
dra27: Fresh thought, coming back to this - how about displaying a warning and advertising the command to revert to it (`opam var set FOO=`) but then setting it without any prompt. It's almost certainly a mistake on the user's part, so we tell them how to fix it, but we still do as we were told (if under protest!)

rjbou:
Updated!

src/client/opamConfigCommand.ml
```diff
@@ -936,6 +936,21 @@ type ('var,'config) var_confset =
...
+  let switch_option =
+    List.exists (fun (name, _) -> String.equal svar name)
+      OpamFile.Switch_config.fields
+  in
+  if global_option || switch_option then
+    (let scope = if global_option then "global" else "switch" in
+     OpamConsole.warning
+       "You are setting a variable that has the same name as a %s option.\n\
+        Did you mean to use `opam option` instead? You can revert this variable's \
+        setting with: 'opam var %s= --%s'\n"
```
kit-ty-kate, line 952:
```suggestion
        Did you mean to use `opam option` instead? You can revert this variable \
        using: 'opam var %s= --%s'\n"
```

kit-ty-kate (approved):

PR #4906, opened 2021-11-11
rewrite bwrap.sh to be posix compliant

saagraa:
This replaces bwrap.sh with a posix compliant version.

readlink got replaced with `cd` and `pwd -P`.

I swapped the bind arguments for the dune cache. In the original, the host physical path was being passed as the sandbox path, now the physical path is the input, and the possibly symlinked path is the output.

### DIFF
--- a/src/state/shellscripts/bwrap.sh
+++ b/src/state/shellscripts/bwrap.sh
@@ -1,9 +1,4 @@
-#!/usr/bin/env bash
-
-set -ue
-
-# Used to make /* match all files (even those beginning with a '.')
-shopt -s dotglob
+#!/bin/sh -e
 
 if ! command -v bwrap >/dev/null; then
     echo "The 'bwrap' command was not found. Install 'bubblewrap' on your system, or" >&2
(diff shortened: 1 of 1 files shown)


dra27 (commented):
The `ARGS` array is important because the final call to `bwrap` is an exec. That behaviour needs to be preserved before we worry about the `readlink` parts.

I'm not certain that the choices are anything other than a `sh` invocation (at which point we have an escaping hell that we got out of using bash arrays) or recursive calls to the script, which eliminates the use of functions to give horrors like this:

```sh
#!/bin/sh
case "$1" in
  finished)
    shift
    exec ./show "$@";;
  building)
    shift
    if [ "$1" = 'second extra argument' ] ; then
      exec "$0" 'building' 'first extra argument' "$@"
    else
      exec "$0" finished "$@"
    fi;;
  *)
    exec "$0" 'building' 'second argument' "$@";;
esac
```

This eventually calls `./show` with two extra prepended arguments!

src/state/shellscripts/bwrap.sh
```diff
@@ -127,4 +125,4 @@ esac
 
 # Note: we assume $1 can be trusted, see https://github.com/projectatomic/bubblewrap/issues/259
 # As of now we are compatible up to 0.1.8, '--' can be added here when we require >= 0.3.0
-exec bwrap "${ARGS[@]}" "$@"
+exec bwrap "$ARGS" "$@"
```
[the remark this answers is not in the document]
saagraa: Would reusing the `$@` array with `set` work?

```
# at beginning, pop the operation
OP=$1; shift

# in mount function
set -- "$B" "$2" "$3" "$@"

# at end
exec "bwrap" "$@"
```

src/state/shellscripts/bwrap.sh
```diff
@@ -1,9 +1,4 @@
-#!/usr/bin/env bash
-
-set -ue
```
kit-ty-kate, line 3: Could you leave `set -ue` as is? Moving it to the shebang line does not set them if the script is called using `sh` directly (e.g. `sh ./bwrap.sh`) and thus has a different behaviour.

src/state/shellscripts/bwrap.sh
```diff
@@ -16,43 +11,50 @@ fi
...
-    if [ -d "$dir" ]; then
-      case "$1" in
-          ro) B="--ro-bind";;
-          rw) B="--bind";;
-          sym) B="--symlink";;
-      esac
-      ARGS=("${ARGS[@]}" "$B" "$2" "$3")
-    fi
+  if [ -d "$2" ]; then
+          case $2 in
```
kit-ty-kate, line 24: could you keep the original indentation even if it is not consistent to ease the review process?

src/state/shellscripts/bwrap.sh
```diff
@@ -16,43 +11,50 @@ fi
...
+                  rw) B="--bind";;
+                  sym) B="--symlink";;
+          esac
+          ARGS="$ARGS $B $2 $3"
+  fi
 }
 
+#$1 is permission, rest are directories mapped 1-1 from host to sandbox
 add_mounts() {
-    local flag="$1"; shift
```
kit-ty-kate, line 38: `local` is a valid POSIX sh construct. Could you keep it?

src/state/shellscripts/bwrap.sh
```diff
@@ -72,28 +74,24 @@ done
...
-        if [[ $f =~ $ccache_dir_regex ]]; then
-          ccache_dir=${BASH_REMATCH[1]}
-          break
-        fi
-      done
+      ccache_dir=$(ccache -p | awk -F '=' '/cache_dir/ { print $2; exit }')
       CCACHE_DIR=${CCACHE_DIR-$HOME/.ccache}
       ccache_dir=${ccache_dir-$CCACHE_DIR}
-      add_mounts rw "$ccache_dir"
+      add_mounts "rw" "$ccache_dir"
```
kit-ty-kate, line 80:
```suggestion
      add_mounts rw "$ccache_dir"
```
Could you keep changes to the minimal?

kit-ty-kate (commented):
Thanks, this contribution is appreciated. However aside from David’s comments on behaviours not matching up, I have a few requests to make it easier to review.

src/state/shellscripts/bwrap.sh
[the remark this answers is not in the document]
dra27: Yes - that at least at removes the horrid calls to `exec`! It'll either need functions to be avoided or all function calls need to be passed "$@" as part of the call.

dra27:
It's possible to test the sandbox "in situ" by backing up `~/.opam/opam-init/hooks/sandbox.sh` and putting the new `bwrap.sh` in its place.

src/state/shellscripts/bwrap.sh
saagraa: `local` isn't posix yet. (https://github.com/emersion/mrsh/issues/183). That said, its probably supported enough that keeping it wouldn't hurt.

src/state/shellscripts/bwrap.sh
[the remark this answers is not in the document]
dra27: Actually, of course it does eliminate the use of functions completely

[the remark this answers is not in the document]
saagraa: Yeah, I'm imagining horrible things with loops and switches

In pseudo code:
```
function_name="add_mount"
while (function_name) {
  switch function_name:
  case "add_mount": ...
  case "add_mounts": ...
  case "return": ...
}
```
Dont worry, I wont do this. Probably.


I'll explore the various options in this thread a bit and rework this PR with what works best.

PR #4908, opened 2021-11-12
Use --no-depexts in CLI 2.0 mode

dra27:
As we continue the puzzle for allowing CI workflows to work properly with CLI versioning, we hit another slight tweak. In opam-depext 1.2 we focused on ensuring that `opam depext` worked in a compatible way. However, there is a slightly subtle interaction with CLI versioning and built-in depext:

```
OPAMCLI=2.0 opam install conf-npm
```
requires, on Ubuntu, the `npm` package but the package may succeed if the user has installed npm manually (or via another mechanism than apt). opam's built-in depext is now a nuisance, as it forces interaction, cf. [this ocaml-cohttp workflow](https://github.com/dra27/ocaml-cohttp/runs/4190685413?check_suite_focus=true#step:7:211):

```
===== ∗ 163 =====

    npm

Let opam run your package manager to install the required system packages?
(answer 'n' for other options) [Y/n] n

    apt-get install npm

You can retry with '--assume-depexts' to skip this check, or run 'opam option depext=false' to permanently disable handling of system packages altogether.
```

In CLI 2.0 mode, `--assume-depexts` of course isn't available, neither - correctly - is `--confirm-level=unsafe-yes`. This PR changes the default mode of depext to be `--assume-depext` in CLI 2.0 mode. We could select `--no-depexts` as the default, but I think the resulting messages, e.g. from [this adapted workflow](https://github.com/dra27/ocaml-cohttp/runs/4190938818?check_suite_focus=true#step:7:208) provides better output if the depexts are _actually_ missing:

```
===== ∗ 163 =====

    npm

```

so I think it's worth the cost of the extra depext check, as a failure later on (if `conf-npm` didn't install) would have a visible remedy further up.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -16,6 +16,7 @@ users)
 ## Global CLI
   * Fix typo in error message for opam var [#4786 @kit-ty-kate - fix #4785]
   * Add cli 2.2 handling [#4853 @rjbou]
+  * --assume-depext is the default in CLI 2.0 mode [#4908 @dra27]
 
 ## Plugins
   *
--- a/master_changes.md
+++ b/master_changes.md
@@ -16,7 +16,7 @@ users)
 ## Global CLI
   * Fix typo in error message for opam var [#4786 @kit-ty-kate - fix #4785]
   * Add cli 2.2 handling [#4853 @rjbou]
-  * --assume-depext is the default in CLI 2.0 mode [#4908 @dra27]
+  * --no-depexts is the default in CLI 2.0 mode [#4908 @dra27]
 
 ## Plugins
   *
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -600,7 +600,7 @@ let create_build_options
     assume_depexts; no_depexts;
   }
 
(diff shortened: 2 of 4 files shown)


kit-ty-kate:
I think `--no-depext` would be more accurate if we’re going for a closer emulation of opam 2.0’s behaviour. no-depext would prevent unavailable packages from being uninstallable even if installed manually

dra27:
That is a very good point (and apologies if you also made it this morning!)

kit-ty-kate:
CI failed with
```
  File "src/client/opamArg.ml", line 627, characters 17-34:
  627 |     ?no_depexts:(flag b.no_depexts || OpamCLIVersion.Op.(cli @= cli2_0))
                         ^^^^^^^^^^^^^^^^^
  Error: This expression has type bool option
         but an expression was expected of type bool
```

dra27:
Oops - that'll teach me to push changes quickly before dashing out, thanks for the fixup @rjbou!

kit-ty-kate (approved):

PR #4909, opened 2021-11-12
Make the 0install solver non-optional

kit-ty-kate:
Making the 0install solver a truely builtin solver would make considerably easier to have a script for people to test their package’s lower-bound dependency constraints, among other things.

Such inclusion would’ve been painful before due to OCaml < 4.08 not being supported by 0install-solver but this restriction was lifted in https://github.com/ocaml/opam/pull/4775


dra27:
Can we keep `--with-0install-solver` (experience from OCaml reveals it's a nuisance to remove the switches) - so `--with-0install-solver` is a no-op and `--without-0install-solver` is an error. Partial example at https://github.com/ocaml/ocaml/blob/91995dee123670c3f249fdebf55836f25e22c90b/configure.ac#L311-L313, but it's slightly messier when it's the `--without` option which wants to be an error. I'll look it up!

The other thing was the message about a built-in solver - IIRC the 0install solver is still not at _exact_ feature parity with MCCS so can we still have a different warning if MCCS is not available? Obviously, the warning can note that the `0install` solver is there and available.

kit-ty-kate:
> Can we keep `--with-0install-solver` (experience from OCaml reveals it's a nuisance to remove the switches) - so `--with-0install-solver` is a no-op and `--without-0install-solver` is an error. Partial example at https://github.com/ocaml/ocaml/blob/91995dee123670c3f249fdebf55836f25e22c90b/configure.ac#L311-L313, but it's slightly messier when it's the `--without` option which wants to be an error. I'll look it up!
> 
> The other thing was the message about a built-in solver - IIRC the 0install solver is still not at _exact_ feature parity with MCCS so can we still have a different warning if MCCS is not available? Obviously, the warning can note that the `0install` solver is there and available.

Done

dra27 (approved):
Good to go when AppVeyor catches up!

AltGr:
I am a little late, but this feels a bit inconsistent with the fact that the actual default built-in solver — mccs — has always been, and remains, optional. Why not just change the `opam` file to mark it as a strong dependency, without removing the feature from the build-system ?

Of course the question is totally different if we use the 0install lib for other purposes, outside of the specific solver implementation. like was discussed for example for conflicts extraction. But this doesn't seem to be the case at the moment ?

dra27:
It seems OK to me inasmuch as it's pure OCaml (unlike mccs and z3) and we've ensured that we're aligned with compiler versions, so it reduces the complexity of the build system?

dra27:
I think there may be benefits to opam-solver library users (niche, I know) being able to rely on the 0install solver being available for 2.2+?

AltGr:
Fair enough... I was just surprised that opam didn't compile anymore on my laptop without installing new deps ;)
So I thought "oh are we using 0install directly already then?" and found out it's just a build-system change.
Since the first point mentioned here is for it to be a «truely builtin solver» I had to say it's a weird argument.

But indeed you have a point that anyone able to compile opam must have no problem compiling 0install, and it doesn't add much weight, so there is no problem with it... it's just I don't really get the motivation at this point :)

Issue #4688, opened 2021-05-30
Feature request: support `opam switch -`


CraigFe:
In interacting with Opam switches, I quite often find myself wanting to switch to a particular switch in order to perform a specific thing, then switch back to the switch I was on before:

```bash
$ opam switch plugin-host
$ ...                     # rebuild an opam plugin
$ opam switch irmin+4.10  # resume normal operation
```

The Git CLI has an idiom for this w.r.t branches, allowing `git switch -`:

```bash
$ git switch plugin-host
$ ...
$ git switch -             # go back to branch before `plugin-host`
```

(In fact, Git has a more general `@{-N}` syntax for referencing historic branch states, and `-` is a synonym of `@{-1}`.)

It might be nice if the `opam` CLI could support a similar syntax, although this would likely be a breaking change as it's currently possible to name an Opam switch "`-`".

AltGr:
Indeed that could be convenient, it just means we would need to store a list of previously selected switches in the config.

For now, and for simple commands I'd just use `--switch plugin-host` for the intermediate commands :)

PR #4910, opened 2021-11-12
Add support for "opam switch -" (go to previous switch)

kit-ty-kate:
This implements https://github.com/ocaml/opam/issues/4688 with the same behaviour as `git switch -`

cc @CraigFe

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -2448,7 +2448,14 @@ let switch cli =
           shell_eval_invocation shell
             (opam_env_invocation ~switch:"SWITCH" ~set_opamswitch:true ())
             |> Manpage.escape));
-  ] @ mk_subdoc ~cli ~defaults:["","list";"SWITCH","set"] commands
+  ] @
+    mk_subdoc ~cli ~defaults:["","list";"SWITCH","set"]
+      ~extra_defaults:[
+        cli_from cli2_2, "-",
+        "Switches back to the previous non-local switch (similar to $(b,git switch -)). \
+         To set the current switch to a switch named $(i,-) use $(b,--cli=2.1) \
+         or lower, or use $(b,opam switch set -)"
+      ] commands
     @ [
       `S Manpage.s_examples;
       `Pre "    opam switch create 4.08.0";
@@ -2792,6 +2799,10 @@ let switch cli =
       `Ok ()
     | Some `current, [] ->
       OpamSwitchCommand.show ();
+      `Ok ()
+    | Some `default "-", [] when OpamCLIVersion.Op.(cli @>= cli2_2) ->
+      OpamGlobalState.with_ `Lock_write @@ fun gt ->
+      OpamSwitchCommand.switch_previous `Lock_none gt;
       `Ok ()
     | Some `set, [switch]
     | Some `default switch, [] ->
--- a/src/client/opamSwitchCommand.ml
+++ b/src/client/opamSwitchCommand.ml
@@ -390,6 +390,19 @@ let switch lock gt switch =
     (OpamSwitch.to_string switch) (OpamSwitch.to_string switch)
     (OpamStd.Format.itemize OpamSwitch.to_string installed_switches)
 
+let switch_previous lock gt =
+  match OpamFile.Config.previous_switch gt.config with
+  | Some switch_name ->
+    OpamConsole.msg
+      (if OpamStateConfig.(!r.dryrun) || OpamClientConfig.(!r.show)
+       then "Would select opam switch %s.\n"
+       else "Selecting opam switch %s.\n")
+      (OpamConsole.colorise `yellow (OpamSwitch.to_string switch_name));
+    switch lock gt switch_name
+  | None ->
+    OpamConsole.error_and_exit `Not_found
+      "No previously used switch could be found"
+
 let import_t ?ask importfile t =
   log "import switch";
 
--- a/src/format/opamFile.ml
+++ b/src/format/opamFile.ml
@@ -1265,6 +1265,14 @@ module ConfigSyntax = struct
   let repositories t = t.repositories
   let installed_switches t = t.installed_switches
   let switch t = t.switch
+  let previous_switch t =
+    let not_current = match t.switch with
+      | Some switch -> fun x -> not (OpamSwitch.equal switch x)
+      | None -> fun _ -> true
+    in
+    OpamStd.List.find_opt (fun switch ->
+        not_current switch && not (OpamSwitch.is_external switch))
+      t.installed_switches
   let jobs t = t.jobs
   let dl_tool t = t.dl_tool
   let dl_jobs t = t.dl_jobs
@@ -1295,8 +1303,16 @@ module ConfigSyntax = struct
   let with_repositories repositories t = { t with repositories }
   let with_installed_switches installed_switches t =
     { t with installed_switches }
-  let with_switch_opt switch t = { t with switch }
-  let with_switch switch t = { t with switch = Some switch }
+  let with_switch switch t =
+    let installed_switches =
+      switch :: List.filter (fun x -> not (OpamSwitch.equal switch x))
+        t.installed_switches
+    in
+    { t with switch = Some switch; installed_switches }
+  let with_switch_opt switch_opt t =
+    match switch_opt with
+    | Some switch -> with_switch switch t
+    | None -> { t with switch = None }
   let with_jobs jobs t = { t with jobs = Some jobs}
   let with_jobs_opt jobs t = { t with jobs }
   let with_dl_tool dl_tool t = { t with dl_tool = Some dl_tool }
--- a/tests/reftests/switch-set.test
+++ b/tests/reftests/switch-set.test
@@ -1,0 +1,77 @@
+N0REP0
+### # Test the "go to previous switch" feature
+### opam switch create foo --empty
+### opam switch - --show
+[ERROR] No previously used switch could be found
+# Return code 5 #
+### opam switch create bar --empty
+### opam switch - --show
+Would select opam switch foo.
+### opam switch show
+bar
+### opam switch -
+Selecting opam switch foo.
+### opam switch - --show
+Would select opam switch bar.
+### opam switch show
+foo
+### opam switch -
+Selecting opam switch bar.
+### opam switch - --show
+Would select opam switch foo.
+### opam switch show
+bar
+### opam switch bar
+### opam switch - --show
+Would select opam switch foo.
+### opam switch -
+Selecting opam switch foo.
+### opam switch - --show
+Would select opam switch bar.
+### opam switch show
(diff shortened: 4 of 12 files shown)


src/format/opamFile.ml
```diff
@@ -1297,6 +1299,7 @@ module ConfigSyntax = struct
     { t with installed_switches }
   let with_switch_opt switch t = { t with switch }
   let with_switch switch t = { t with switch = Some switch }
+  let with_previous_switch previous_switch t = { t with previous_switch }
```
rjbou, line 1302: According the convention, `with_x` takes `'x` value, while `with_x_opt` takes `'x option` value (like `with_switch`).

src/format/opamFile.ml
```diff
@@ -1361,6 +1365,9 @@ module ConfigSyntax = struct
       if t.switch = None then with_switch sw t
       else Pp.bad_format "Multiple switch specifications"
     in
+    let with_previous_switch sw t =
+      with_previous_switch (Some sw) t
+    in
```
rjbou, line 1370: There is no specific behavior added, better raise it up, with `with_previous_switch`.

src/client/opamSwitchCommand.ml
```diff
@@ -312,6 +312,11 @@ let create
   let update_config = update_config && not (OpamSwitch.is_external switch) in
   let comp_dir = OpamPath.Switch.root gt.root switch in
   let simulate = OpamStateConfig.(!r.dryrun) || OpamClientConfig.(!r.show) in
+  (* TODO: We might want to make some more tests (quotes?, ...) *)
+  let switch_str = OpamSwitch.to_string switch in
+  if String.equal switch_str "-" then
+    OpamConsole.error_and_exit `Bad_arguments
+      "The name of the switch is not valid";
```
rjbou, line 319: This check need to be added in `OpamSwitch.check` function, that is called at each `OpamSwitch.of_string` (that is called by cli argument parsing, among other places.)

tests/reftests/switch-set.test
```diff
@@ -0,0 +1,37 @@
...
+[default] synchronised from file://${BASEDIR}/REPO
+Now run 'opam upgrade' to apply any package updates.
+### opam switch create foo test
+
+<><> Installing new switch packages <><><><><><><><><><><><><><><><><><><><><><>
+Switch invariant: ["test"]
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> installed test.1
+Done.
```
rjbou, line 17: To simplify the test, you can create empty switches, no need of a package.

rjbou (changes requested):
Thanks!
Some modifications suggested, and some doc missing (manpage, website).
Also, should we add an format upgrade function to convert already present switches named `-`, in case they exist? Or we can assume that they don't exist (it wasn't forbidden, so it can exist).

kit-ty-kate:
All done. The switch name `-` now does not interfere with `opam switch -`. It is still possible to create such a switch and to set it (using either `opam switch set -` or `opam switch --cli=2.1 -`) so no issues converting.

I’m still trying to figure out the manpage/doc situation but other than that this should be ready

src/client/opamSwitchCommand.ml
```diff
@@ -392,6 +392,13 @@ let switch lock gt switch =
     (OpamSwitch.to_string switch) (OpamSwitch.to_string switch)
     (OpamStd.Format.itemize OpamSwitch.to_string installed_switches)
 
+let switch_previous lock gt =
+  match OpamFile.Config.previous_switch gt.config with
+  | Some switch_name -> switch lock gt switch_name
```
rjbou, line 397: should we check here previous switch existence, in order to display a more specific error message than `switch` one?

rjbou:
On non avoiding `-` as a switch name, we loose correspondence between `opam switch x` and `opam switch set x`. It's not an issue itself, but we need it to be clear on switch doc (and later on release doc). Old behavior is well kept with the cli check.

src/client/opamSwitchCommand.ml
kit-ty-kate: I don’t think it’s necessary, the default error message from opam switch is perfectly fine i think

kit-ty-kate:
> On non avoiding `-` as a switch name, we loose correspondence between `opam switch x` and `opam switch set x`. It's not an issue itself, but we need it to be clear on switch doc (and later on release doc). Old behavior is well kept with the cli check.

is the new manpage entry clear enough?

rjbou:
It is clear, i wrote the comment before you last push :)

rjbou (approved):

src/client/opamCommands.ml
```diff
@@ -2438,7 +2438,14 @@ let switch cli =
           shell_eval_invocation shell
             (opam_env_invocation ~switch:"SWITCH" ~set_opamswitch:true ())
             |> Manpage.escape));
-  ] @ mk_subdoc ~cli ~defaults:["","list";"SWITCH","set"] commands
+  ] @
+    mk_subdoc ~cli ~defaults:["","list";"SWITCH","set"]
+      ~extra_defaults:[
+        cli_from cli2_2, "-",
+        "Switches back to the previous switch (similar to $(b,git switch -)). \
```
kit-ty-kate, line 2445:
```suggestion
        "Switches back to the previous non-local switch (similar to $(b,git switch -)). \
```

src/client/opamSwitchCommand.ml
kit-ty-kate:
```suggestion
  | Some switch_name ->
    if not (OpamGlobalConfig.switch_exists gt switch_name) then
      OpamConsole.error_and_exit `Not_found "The previous switch no longer exists";
    switch lock gt switch_name
```
Something like that?

src/client/opamSwitchCommand.ml
```diff
@@ -392,6 +392,16 @@ let switch lock gt switch =
     (OpamSwitch.to_string switch) (OpamSwitch.to_string switch)
     (OpamStd.Format.itemize OpamSwitch.to_string installed_switches)
 
+let switch_previous lock gt =
+  match OpamFile.Config.previous_switch gt.config with
+  | Some switch_name ->
+    if not (OpamGlobalState.switch_exists gt switch_name) then
+      OpamConsole.error_and_exit `Not_found "The previous switch no longer exists";
```
rjbou, line 399:
```suggestion
      OpamConsole.error_and_exit `Not_found "The previous switch %s no longer exists"
        (OpamSwitch.to_string switch);
```

rjbou (approved):

rjbou:
We need to check for opam root, but otherwise, lgtm!

rjbou:
I need to rework the opam root test first, to make it no opamroot bump specific

rjbou:
Reworked in #4913 & #4926 for the bump

AltGr:
> On non avoiding - as a switch name, we loose correspondence between opam switch x and opam switch set x. It's not an issue itself, but we need it to be clear on switch doc (and later on release doc). Old behavior is well kept with the cli check.

that correspondance was already not there (which is why `set` still exists), you could have switches called `create` or `set`.

AltGr:
LGTM; I can suggest an idea to avoid having to bump the format though: order the `installed-switches:` field by order of most recently used first. This would be fully backwards-compatible and consistent, and could even maybe find other uses.

AltGr:
Couldn't resist and tried an implementation :)
There are differences though, I cut the patch in 2 to make the test difference obvious in ac19c215

src/format/opamFile.ml
```diff
@@ -1266,7 +1265,14 @@ module ConfigSyntax = struct
   let repositories t = t.repositories
   let installed_switches t = t.installed_switches
   let switch t = t.switch
-  let previous_switch t = t.previous_switch
+  let previous_switch t =
+    let not_current = match t.switch with
+      | Some s -> (<>) s
+      | None -> fun _ -> true
```
kit-ty-kate, line 1271:
```suggestion
      | Some switch -> fun x -> not (String.equal switch x)
      | None -> fun _ -> true
```

src/format/opamFile.ml
```diff
@@ -1297,10 +1303,15 @@ module ConfigSyntax = struct
   let with_repositories repositories t = { t with repositories }
   let with_installed_switches installed_switches t =
     { t with installed_switches }
-  let with_switch_opt switch t = { t with switch }
-  let with_switch switch t = { t with switch = Some switch }
-  let with_previous_switch_opt previous_switch t = { t with previous_switch }
-  let with_previous_switch previous_switch t = { t with previous_switch = Some previous_switch }
+  let with_switch switch t =
+    let installed_switches =
+      switch :: List.filter ((<>) switch) t.installed_switches
```
kit-ty-kate, line 1308:
```suggestion
      switch :: List.filter (fun x -> not (String.equal switch x)) t.installed_switches
```

tests/reftests/switch-set.test
```diff
@@ -70,10 +70,8 @@ Switch ${BASEDIR} and all its packages will be wiped. Are you sure? [Y/n] y
 ### opam switch remove foo --yes
 Switch foo and all its packages will be wiped. Are you sure? [Y/n] y
 ### opam switch -
-[ERROR] The previous switch 'foo' no longer exists
-# Return code 5 #
+Selecting opam switch -.
 ### opam switch - --show
-[ERROR] The previous switch 'foo' no longer exists
-# Return code 5 #
+Would select opam switch bar.
```
kit-ty-kate, line 75: I’m not a big fan of the new behaviour but given it’s only going to be used very rarely, I’m fine with it.

src/format/opamFile.ml
```diff
@@ -1265,6 +1265,14 @@ module ConfigSyntax = struct
   let repositories t = t.repositories
   let installed_switches t = t.installed_switches
   let switch t = t.switch
+  let previous_switch t =
+    let not_current = match t.switch with
+      | Some switch -> fun x -> not (String.equal switch x)
```
kit-ty-kate, line 1270:
```suggestion
      | Some switch -> fun x -> not (OpamSwitch.equal switch x)
```

src/format/opamFile.ml
```diff
@@ -1295,8 +1303,15 @@ module ConfigSyntax = struct
   let with_repositories repositories t = { t with repositories }
   let with_installed_switches installed_switches t =
     { t with installed_switches }
-  let with_switch_opt switch t = { t with switch }
-  let with_switch switch t = { t with switch = Some switch }
+  let with_switch switch t =
+    let installed_switches =
+      switch :: List.filter (fun x -> not (String.equal switch x)) t.installed_switches
```
kit-ty-kate, line 1308:
```suggestion
      switch :: List.filter (fun x -> not (OpamSwitch.equal switch x)) t.installed_switches
```

rjbou (approved):

PR #4914, opened 2021-11-16
Fix deps-only with depopts

### LINKED ISSUE #4911: opam install --deps-only tries to remove packages when deps already present
# Problem

opam is removing packages against expectation.

Consider my switch, where I have my full dependency set installed. The switch could be considered "complete" or the superset of all packages that will ever need to be installed.

Given that the switch is full/complete, it does not seem appropriate that a command of `opam install --deps-only <pkgs>` attempts to **strictly uninstall** packages not in the `<pkgs>` arg.

Example:

```
$ opam install --deps-only js_of_ocaml js_of_ocaml-lwt js_of_ocaml-compiler js_of_ocaml-ocamlbuild js_of_ocaml-toplevel js_of_ocaml-ppx js_of_ocaml-ppx_deriving_json js_of_ocaml-tyxml
  ⊘ remove cohttp-lwt-unix   4.0.0
  ⊘ remove # snip snip snip, 13 other removals, no additions
  ⊘ remove logs              0.7.0
===== ⊘ 15 =====
```

Why remove `cohttp-lwt-unix`, or `logs`? It appears that `--deps-only` is behaving more like `--exclusively-set-the-switch-to-these-deps-only` (not a real flag :smile:). This is not what the manpage says `--deps-only` does, however. Per the man page,

```
       --deps-only
           Install all its dependencies, but don't actually install the
           package.
```

As one can observe, there is nothing to suggest removal of sibling packages from the switch. `install --deps-only` appears to have an undeclared side-effect. When we remove `--deps-only` from the same command:

```
$ opam install js_of_ocaml js_of_ocaml-lwt js_of_ocaml-compiler js_of_ocaml-ocamlbuild js_of_ocaml-toplevel js_of_ocaml-ppx js_of_ocaml-ppx_deriving_json js_of_ocaml-tyxml
  ∗ install   js_of_ocaml-compiler          3.11.0
  # snip snip, no removals
  ↻ recompile cohttp-lwt-unix               4.0.0  [uses logs]
===== ∗ 8   ↻ 15 =====
```

one can observe that no such removal occurs, which is consistent with expectations for an install command.

It is worthy of mention that installs **can occur** when the resolver needs to change versions of a package for overall switch compatibility. That is not what is occurring here--there are removals _only_, not recompiles with renegotiated package versions.

Thanks for the time, and consideration!

rjbou:
Packages to reinstall/recompile were removed. Fix #4911 
There is no more removal of depopts, but remains some unneeded recompiling (trigered by the main package).

### DIFF
--- a/src/solver/opamSolver.ml
+++ b/src/solver/opamSolver.ml
@@ -772,9 +772,9 @@ let dump_universe universe oc =
 
 let filter_solution filter t =
   let t = OpamCudf.ActionGraph.copy t in
-  let rec rm ?(not_reinstall=fun _ -> true) iter_deps v =
-    if OpamCudf.ActionGraph.mem_vertex t v && not_reinstall v then
-      (iter_deps (rm ~not_reinstall iter_deps) t v;
+  let rec rm ?(is_reinstall=fun _ -> false) iter_deps v =
+    if OpamCudf.ActionGraph.mem_vertex t v && not (is_reinstall v) then
+      (iter_deps (rm ~is_reinstall iter_deps) t v;
        OpamCudf.ActionGraph.remove_vertex t v)
   in
   OpamCudf.ActionGraph.iter_vertex (function
@@ -783,16 +783,16 @@ let filter_solution filter t =
       | (`Install nv | `Change (_,_,nv)) as a
         when not (filter (OpamCudf.cudf2opam nv)) ->
         (* If a predecessor is a remove, it is a reinstall fragmented *)
-        let not_reinstall v =
+        let is_reinstall v =
           match v with
           | `Install nv ->
-            List.for_all (function
-                | `Remove nv' -> OpamCudf.Package.compare nv nv' <> 0
-                | _ -> true)
+            List.exists (function
+                | `Remove nv' -> OpamCudf.Package.equal nv nv'
+                | _ -> false)
               (OpamCudf.ActionGraph.pred t v)
-          | _ -> true
+          | _ -> false
         in
-        rm OpamCudf.ActionGraph.iter_succ ~not_reinstall a
+        rm OpamCudf.ActionGraph.iter_succ ~is_reinstall a
       | _ -> ())
     t;
   t
--- a/src/solver/opamSolver.ml
+++ b/src/solver/opamSolver.ml
@@ -772,18 +772,27 @@ let dump_universe universe oc =
 
 let filter_solution filter t =
   let t = OpamCudf.ActionGraph.copy t in
-  let rec rm iter_deps v =
-    if OpamCudf.ActionGraph.mem_vertex t v then (
-      iter_deps (rm iter_deps) t v;
-      OpamCudf.ActionGraph.remove_vertex t v
-    ) in
-  OpamCudf.ActionGraph.iter_vertex
-    (function
+  let rec rm ?(not_reinstall=fun _ -> true) iter_deps v =
+    if OpamCudf.ActionGraph.mem_vertex t v && not_reinstall v then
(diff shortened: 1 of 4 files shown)


src/solver/opamSolver.ml
```diff
@@ -772,19 +772,27 @@ let dump_universe universe oc =
...
   OpamCudf.ActionGraph.iter_vertex
-    (function
-      | `Remove nv as a when not (filter (OpamCudf.cudf2opam nv)) ->
-        rm OpamCudf.ActionGraph.iter_pred a
-      | (`Install nv | `Change (_,_,nv)) as a
-        when not (filter (OpamCudf.cudf2opam nv)) ->
-        rm OpamCudf.ActionGraph.iter_succ a
-      | _ -> ())
+    (fun a ->
+       match a with
```
kit-ty-kate, line 782:
```suggestion
    (function
```
`a` is not used anywhere (it’s all redefined by the `as a` below) and this would avoid the change of indentation

src/solver/opamSolver.ml
```diff
@@ -772,19 +772,27 @@ let dump_universe universe oc =
...
+       | `Remove nv as a when not (filter (OpamCudf.cudf2opam nv)) ->
+         rm OpamCudf.ActionGraph.iter_pred a
+       | (`Install nv | `Change (_,_,nv)) as a
+         when not (filter (OpamCudf.cudf2opam nv)) ->
+         (* If a predecessor is a remove, it is a reinstall fragmented *)
+         let check v =
+           List.exists (function
+               | `Remove nv' -> nv = nv'
```
kit-ty-kate, line 790: In my opinion polymorphic equality should be avoided whenever possible to avoid nasty bugs in the future. Could you use the appropriate `equal` function (or set the type, e.g. ``(nv : [`A | `B | `C]) = nv’``)

src/solver/opamSolver.ml
```diff
@@ -772,19 +772,27 @@ let dump_universe universe oc =
 
 let filter_solution filter t =
   let t = OpamCudf.ActionGraph.copy t in
-  let rec rm iter_deps v =
-    if OpamCudf.ActionGraph.mem_vertex t v then (
-      iter_deps (rm iter_deps) t v;
+  let rec rm ?(check=fun _ -> true) iter_deps v =
```
kit-ty-kate, line 775: it is not clear from the implementation what the new `check` parameter checks for. Could this be renamed something like `is_not_already_set_for_removal`? (if i understand the intent correctly)

kit-ty-kate (approved):
Looks good to me on the surface. Let’s wait for @AltGr for the in-depth review but that should be good to go IMO.

src/solver/opamSolver.ml
```diff
@@ -772,18 +772,27 @@ let dump_universe universe oc =
...
 let filter_solution filter t =
   let t = OpamCudf.ActionGraph.copy t in
-  let rec rm iter_deps v =
-    if OpamCudf.ActionGraph.mem_vertex t v then (
-      iter_deps (rm iter_deps) t v;
-      OpamCudf.ActionGraph.remove_vertex t v
-    ) in
-  OpamCudf.ActionGraph.iter_vertex
-    (function
+  let rec rm ?(not_reinstall=fun _ -> true) iter_deps v =
```
cdaringe, line 775: an aside only, it may be nice to use the positive case convention. `?(reinstall=fun _ -> false)`, that way readers don't need to process the double negative to comprehend the positive result.

kit-ty-kate: good point. Done. Thanks!

PR #4917, opened 2021-11-17
Add clean `--untracked`

### LINKED ISSUE #4831: opam clean --untracked
We currently do not have a way to detect and clean (interactively) a switch containing external/modified files.
It would be nice to have this option around so that when we get warnings such as:
```
[NOTE] While removing ocaml-base-compiler.4.12.0: not removing non-empty directories:
  - share
```
we could do something about it and know what was added or modified in this directory.

Voiced in https://github.com/ocaml/opam/issues/4781

rjbou:
fix #4831
See test file for behaviour

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -3674,11 +3674,110 @@ let clean cli =
     mk_flag ~cli cli_original ["a"; "all-switches"]
       "Run the switch cleanup commands in all switches. Implies $(b,--switch-cleanup)"
   in
+  let untracked =
+    mk_flag ~cli (cli_from cli2_2) ["untracked"]
+      "Clean untracked file and directories"
+  in
+  let remove_untracked ~dry_run st  =
+    let root = st.switch_global.root in
+    let sw = st.switch in
+    let open OpamFilename in
+    (* installed files *)
+    let files, dirs =
+      OpamPackage.Set.fold (fun nv (files, dirs) ->
+          let changes =
+            OpamFile.Changes.safe_read
+              (OpamPath.Switch.changes root sw (OpamPackage.name nv))
+          in
+          List.fold_left (fun (files, dirs) relpath ->
+              let path =
+                Op.(OpamPath.Switch.root root sw // relpath)
+              in
+              if exists path then Set.add path files, dirs else
+              let pathd =
+                Op.(OpamPath.Switch.root root sw / relpath)
+              in
+              if exists_dir pathd then files, Dir.Set.add pathd dirs
+              else files, dirs)
+            (files, dirs) (OpamStd.String.Map.keys changes))
+        st.installed (Set.empty, Dir.Set.empty)
+    in
+    let dirs_of_files =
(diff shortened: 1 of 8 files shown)


src/client/opamCommands.ml
```diff
@@ -3663,7 +3667,7 @@ let clean cli =
     in
     let switches =
       if all_switches then OpamGlobalState.switches gt
-      else if switch then match OpamStateConfig.get_switch_opt () with
+      else if switch || untracked then match OpamStateConfig.get_switch_opt () with
```
kit-ty-kate, line 3670: One could call `opam clean --untracked --all-switches`
```suggestion
      else if switch then match OpamStateConfig.get_switch_opt () with
```

tests/reftests/clean.test
```diff
@@ -0,0 +1,91 @@
...
+  - ${BASEDIR}/OPAM/clean/lib/things/second
+Remove directory? [Y/n] y
+Cleaning up switch clean
+### opam install things
+The following actions will be performed:
+  - install things 1
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> installed things.1
+Done.
```
kit-ty-kate, line 91: Could you add something like:
```
### <modify an installed file>
### opam clean --untracked
<see if it’s actually removed>
### opam clean --untracked
<see if nothing is actually done>
```

src/client/opamCommands.ml
rjbou: It is already handled by the line above
```
       if all_switches then OpamGlobalState.switches gt 
```

kit-ty-kate: Ah my bad

tests/reftests/clean.test
rjbou: with an installed or not installed package ?

kit-ty-kate: why not both, just in case

rjbou: ok

AltGr:
Nice! I have a few comments:
- the code would probably deserve to be in its own function
- the interactive mode is likely to spam a *lot* of questions... maybe list them and ask once ? Otherwise require `--dry-run`/`--yes` a bit like `git clean -n/f`, maybe ?
- do you make sure to minimise the tree, e.g. if there is a whole subtree that is expected to be removed, ask just once for its root (or if you want to list everything, make sure to list the contents first, then the dir — if empty — with an `rmdir`)
- `OpamConsole.msg "Cleaning up switch %s\n" (OpamSwitch.to_string sw);` should probably be above the code

rjbou:
Updated!
```
### opam clean --untracked
Cleaning up switch clean
Remaining directories and files:
  - ${BASEDIR}/OPAM/clean/bin/first
  - ${BASEDIR}/OPAM/clean/lib/things/second
  - ${BASEDIR}/OPAM/clean/lib/things

Remove them? [Y/n] y

kit-ty-kate (approved):

PR #4924, opened 2021-11-19
[CI]: Upgrade autoconf from 2.69 to 2.71

kit-ty-kate:
Ubuntu has patches on top of 2.69 so it’s impossible to regenerate the configure file as it is in CI on something other than a Debian based system.


configure.ac
```diff
@@ -1,6 +1,6 @@
 dnl The line below must be formatted AC_INIT(opam,VERSION) with no extra spaces
```
dra27, line 1:
```suggestion
AC_PREREQ([2.71])
dnl The line below must be formatted AC_INIT(opam,VERSION) with no extra spaces
```

dra27 (commented):
I'd add the `AC_PREREQ` just to force an error for users with an older version (it's a >= test, so it won't cause autoconf 2.72 to fail, etc.)

kit-ty-kate:
Not urgently needed since https://github.com/ocaml/opam/pull/5090. Closing for now.

PR #4926, opened 2021-11-22
Upgrade root version to 2.2~alpha

rjbou:
* [x] queued on #5305

### DIFF
--- a/src/state/opamFormatUpgrade.ml
+++ b/src/state/opamFormatUpgrade.ml
@@ -1102,6 +1102,10 @@ let from_2_0_to_v2_1 ~on_the_fly _ conf =
    | None -> info_jobs_changed ~prev_jobs:1);
   OpamFile.Config.with_jobs_opt None conf, gtc_none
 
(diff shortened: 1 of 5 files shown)


src/state/opamFormatUpgrade.ml
```diff
@@ -1093,14 +1094,28 @@ let from_2_0_to_v2_1 _ conf =
...
+     upgrades
+*)
 
 let latest_version = OpamFile.Config.root_version
 
 let latest_hard_upgrade = (* to *) v2_0_beta5
 
-(* intermediates roots that need an hard upgrade *)
-let intermediate_roots = [
+(* intermediates roots that needs an hard upgrade if upgraded from them *)
```
dra27, line 1117:
```suggestion
(* intermediates roots that need a hard upgrade when upgrading from them *)
```

dra27 (approved):
Minor tweak of a comment - the two commits specific to the PR LGTM

PR #4927, opened 2021-11-24
bootstrap: update ocaml version

rjbou:
To integrate mingw fix


dra27:
(put in 2.1.2 only because we haven't actually tagged it yet)

kit-ty-kate (approved):

kit-ty-kate:
I think we should backport it to 2.0 as well. Currently it’s not possible to compile the 2.0 branch on the latest macOS with mccs due to the use of C++20 and the installation by ocaml < 4.12 of the VERSION file

kit-ty-kate:
Merging in master. I’m opening backport PRs right away.

PR #4931, opened 2021-11-24
On version pinning don't read archive opam file

### LINKED ISSUE #4608: Remove the need to read the opam file inside the archive on a version pin
In https://github.com/ocaml/opam/pull/4580 @rjbou fixed a bug in the case we're reading the opam file inside the archive but the opam file is invalid. It seems that we should not look into the archive at all. We should review that at some point.

rjbou:
fix #4608

### DIFF
--- a/src/client/opamAction.ml
+++ b/src/client/opamAction.ml
@@ -245,11 +245,12 @@ let download_package st nv =
   let dir = OpamSwitchState.source_dir st nv in
   if OpamPackage.Set.mem nv st.pinned &&
      OpamFilename.exists_dir dir &&
-     OpamStd.Option.Op.(
-       OpamPinned.find_opam_file_in_source nv.name dir >>|
-       OpamFile.OPAM.safe_read >>=
-       OpamFile.OPAM.version_opt)
-     = Some nv.version
+     (OpamSwitchState.is_version_pinned st (OpamPackage.name nv) ||
+      OpamStd.Option.Op.(
+        OpamPinned.find_opam_file_in_source nv.name dir >>|
+        OpamFile.OPAM.safe_read >>=
+        OpamFile.OPAM.version_opt)
+      = Some nv.version)
   then Done None
   else
   let print_action msg =
--- a/master_changes.md
+++ b/master_changes.md
@@ -40,7 +40,7 @@ users)
   * [BUG] Fix `set-invariant: default repos were loaded instead of switch repos [#4866 @rjbou]
 
 ## Pin
-  *
+  * Don't read opam file in source for version pin [#4931 @rjbou - #4608]
 
 ## List
   * Some optimisations to 'opam list --installable' queries combined with other filters [@altgr]


src/client/opamAction.ml
```diff
@@ -245,11 +245,12 @@ let download_package st nv =
   let dir = OpamSwitchState.source_dir st nv in
   if OpamPackage.Set.mem nv st.pinned &&
      OpamFilename.exists_dir dir &&
-     OpamStd.Option.Op.(
-       OpamPinned.find_opam_file_in_source nv.name dir >>|
-       OpamFile.OPAM.safe_read >>=
-       OpamFile.OPAM.version_opt)
-     = Some nv.version
+     (OpamSwitchState.is_version_pinned st (OpamPackage.name nv) ||
```
kit-ty-kate, line 248: Does this also return `true` when doing `opam pin add pkg.version <repo>`? In that case, there is no need to look into the archive/repo either

kit-ty-kate (commented):
I’m not sure what the entire if statement is trying to do in the first place.

"do not do anything (equivalent to `--no-action`) if e.g. doing something like `opam pin add pkg.version <archive/repo>` and the version number is the same as the one in the archive/repo" ? That seems a bit suspicious to me

kit-ty-kate:
I had a go at trying to debug this suspicious `if` and i have a repro case:
```
$ cat test/dune.opam
opam-version: "2.0"
version: "2.0.1"
build: ["sh" "-c" "ls && false"]
url {
  src: "https://github.com/ocaml/dune/archive/main.tar.gz"
}
$ opam pin add dune.2.0.1 test/
[...]

#=== ERROR while compiling dune.2.0.1 =========================================#
# context     2.1.0 | linux/x86_64 | ocaml-variants.4.13.0+trunk | pinned(file:///home/opam/test)
# path        ~/.opam/4.13/.opam-switch/build/dune.2.0.1
# command     /bin/sh -c ls && false
# exit-code   1
# env-file    ~/.opam/log/dune-236-a0af0d.env
# output-file ~/.opam/log/dune-236-a0af0d.out
### output ###
# dune.opam
```

So I really think this whole `if` statement should be removed as a whole.

kit-ty-kate (approved):

rjbou:
It is not suspicious, that `if` statement is present for a reason: optimisation. If the package is version pinned, there is no need to retrieve it again / synchronise it. Maybe we can say/decide that this optimisation is not efficient, but that's another question.
`opam pin add pkg.version <non-archive-url>` is different than a version pin. In the first case, it is a path/git/etc pin with given version, while a version pin is "pin pkg with repo opam file at that version", so sources are taken from url archive (or repo, on official opam repository we take only archives url but they can be vcs ones), and is not prone to change/update unless the opam repository changes.
So if you `opam pin add pkg.version url`, it pins to path/vcs url, it takes the opam file from the url, and it needs to synchronise from it each time. The `dune.2.0.1` pinning/install behavior is the wanted one.

We need to check `opam pin pkg.version archive-url` behavior, and see what we want for that.

We also need to add some pinning and downloading/syncing tests, to check and be sure that we don't sync more that we need.

> "do not do anything (equivalent to `--no-action`) if e.g. doing something like `opam pin add pkg.version <archive/repo>` and the version number is the same as the one in the archive/repo" 

From where is this citation? Maybe we need to clarify that doc.

PR #4933, opened 2021-11-25
repo add/set-url: add `--no-action`

### LINKED ISSUE #4617: opam repo set-url - add --no-action
At present `opam repo set-url` both changes the URL for the repo and issues an update. It would be useful (cf. https://github.com/ocurrent/docker-base-images/issues/99#issue-842144658) to be able to have `git remote set-url`-like behaviour where the URL is updated but not yet fetched (although the default of actually updating is much more desirable than git's...)

rjbou:
to not initialise the repo directly. It will be fetched on next update. See test file

Fixes https://github.com/ocaml/opam/issues/4617

TODO
* [ ] rebase
* [ ] queued in #6327

### DIFF
--- a/src/state/opamFormatUpgrade.ml
+++ b/src/state/opamFormatUpgrade.ml
@@ -764,7 +764,7 @@ let from_1_3_dev7_to_2_0_alpha root ~on_the_fly:_ conf =
   in
   OpamFile.Repos_config.write (OpamPath.repos_config root)
     (OpamRepositoryName.Map.of_list
-       (List.map (fun (_, r, u) -> r, Some (u,None)) prio_repositories));
+       (List.map (fun (_, r, u) -> r, Some (u,true,None)) prio_repositories));
   let prio_repositories =
     List.stable_sort (fun (prio1, _, _) (prio2, _, _) -> prio2 - prio1)
       prio_repositories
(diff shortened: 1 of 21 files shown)


rjbou:
TODO: add non initialised information to have a more specific error message

dra27:
Note from an dev meeting (a few weeks ago 🙈): could possibly add a flag internally to the repo state so that we can identify the first `opam update` since the url was set so that the warning could be better (to differentiate, say, the network being down and not warning about all the URLs) i.e. allow opam to know the difference between "this URL has never worked / never been checked" and "this URL has started failing since last time"

dra27:
Cross-referencing #4617, which this will close!

src/state/opamFormatUpgrade.ml
```diff
@@ -1066,22 +1066,33 @@ let downgrade_2_1_switches root conf =
...
 
-let from_2_1_alpha2_to_v2_1_rc root conf =
-  downgrade_2_1_switches root conf
+let v2_2_alpha = OpamVersion.of_string "2.2~alpha"
 
-let from_2_1_rc_to_v2_1 _ conf = conf
+let from_2_1_to_2_2_alpha_repo root _conf =
+  let f = OpamPath.repos_config root in
+  OpamStd.Option.iter (OpamFile.Repos_config.write f)
+    (OpamFile.Repos_config.read_opt f)
```
dra27, line 1084: Shouldn't this be iterated over to set all of the initialised fields to true, given that they must have been initialised before? (cf. the v1.3+ upgrade)?

rjbou:
Discussion in dev meeting: This needs an hue rebase, and it leads to question of updating the repos-config file format. We need to cary more information, other PR add fields too.
see #6327

PR #4939, opened 2021-11-30
opam admin cache: only fill missing by default

AltGr:
and add a `--check-all` option for the previous behaviour


kit-ty-kate:
Looks good. Thanks!

PR #4940, opened 2021-12-01
Windows: remove double printing on commands and their output

rjbou:
A specific behavior is set for non windows devices to handle verbose printing of commands and their output. This makes windows workflow go into the fallback one, and have the double printing.

### DIFF
--- a/src/core/opamProcess.ml
+++ b/src/core/opamProcess.ml
@@ -639,8 +639,8 @@ let exit_status p return =
   if isset_verbose_f () then
     stop_verbose_f ()
   else if p.p_verbose then
-    (verbose_print_cmd p;
-     List.iter verbose_print_out stdout;
+    (List.iter verbose_print_out stdout;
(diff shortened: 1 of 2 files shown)


src/core/opamProcess.ml
```diff
@@ -639,8 +639,8 @@ let exit_status p return =
   if isset_verbose_f () then
     stop_verbose_f ()
   else if p.p_verbose then
-    (verbose_print_cmd p;
```
rjbou, line 642: It is already printed on `set_verbose_process` call.

src/core/opamProcess.ml
```diff
@@ -639,8 +639,8 @@ let exit_status p return =
...
-     List.iter verbose_print_out stdout;
+    (List.iter verbose_print_out stdout;
+     if p.p_stdout <> p.p_stderr then
```
rjbou, line 643: By default `stderr` is set to `stdout`.

PR #4947, opened 2021-12-06
Make the release script setup-less using QEMU, Docker and Rosetta 2

kit-ty-kate:
One button press to release. So far the QEMU setup is by far the slowest bit (~3/4 hours) but at least it does not require any supervision or setup.

PR on top of #4946

### DIFF
--- a/release/readme.md
+++ b/release/readme.md
@@ -18,7 +18,7 @@
 * Create a new file at ~/.github/jar/infra with: {"scopes":[],"token":"ghp_XXX","app":{"name":"infra","url":"https://developer.github.com/v3/oauth_authorizations/"},"url":"https://api.github.com/authorizations/YYY","id":"YYY","note":"infra"}
   * Replace XXX by the token
   * Replace YYY by its ID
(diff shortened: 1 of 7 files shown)


release/readme.md
```diff
@@ -5,16 +5,31 @@
...
+  "scopes":[],
+  "token":"ghp_XXX",
+  "app":{"name":"infra","url":"https://developer.github.com/v3/oauth_authorizations/"},
+  "url":"https://api.github.com/authorizations/YYY",
+  "id":"YYY",
+  "note":"infra"
+}
+```
+  * Replace XXX by the token
+  * Replace YYY by its ID (you can find it in the url of the token page)
```
kit-ty-kate, line 30: I think we should probably replace this section with just "upload everything by hand", it’s faster and I still had issues yesterday with git-upload-release somehow

AltGr: Yeah in practice I have the same experience with unreliable `git-upload-release`, and uploading by hand is OK

AltGr:
This is hard to review, but I have no objections to merging if you've double-checked the generated binaries (that they work, and are properly linked) :+1:

PR #4949, opened 2021-12-07
Switch the default version when undefined from ~dev to dev

kit-ty-kate:
Makes it easier to mix pin packages and packages from the repository without having to set the version again when encountering an issue.

When you take the development version of something you expect more often to be ahead of the latest release rather than the opposite.

Issue encountered by @bikallem

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -41,7 +41,7 @@ users)
   * [BUG] Fix `set-invariant: default repos were loaded instead of switch repos [#4866 @rjbou]
 
 ## Pin
-  *
+  * Switch the default version when undefined from ~dev to dev [#4949 @kit-ty-kate]
 
 ## List
   * Some optimisations to 'opam list --installable' queries combined with other filters [@altgr]
@@ -202,6 +202,7 @@ users)
 ## opam-solver
 ## opam-format
   * `OpamStd.ABSTRACT`: add `compare` and `equal`, that added those functions to `OpamSysPkg` and `OpamVariable` [#4918 @rjbou]
+  * Add OpamPackage.Version.default returning the version number used when no version is given for a package [#4949 @kit-ty-kate]
 ## opam-core
   * OpamSystem: avoid calling Unix.environment at top level [#4789 @hannesm]
   * `OpamStd.ABSTRACT`: add `compare` and `equal`, that added those functions to `OpamFilename`, `OpamHash`, `OpamStd`, `OpamStd`, `OpamUrl`, and `OpamVersion` [#4918 @rjbou]
--- a/src/client/opamPinCommand.ml
+++ b/src/client/opamPinCommand.ml
@@ -364,7 +364,7 @@ exception Nothing_to_do
 
 let default_version st name =
   try OpamPackage.version (OpamSwitchState.get_package st name)
-  with Not_found -> OpamPackage.Version.of_string "~dev"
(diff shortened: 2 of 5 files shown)


src/client/opamPinCommand.mli
```diff
@@ -86,7 +86,3 @@ val parse_pins: string list -> (name * version option * url * string option) lis
     [url]), and returns it *)
 val read_opam_file_for_pinning:
   ?quiet:bool -> name -> OpamFile.OPAM.t OpamFile.t -> url -> OpamFile.OPAM.t option
-
-(** The default version for pinning a package: depends on the state, what is
-    installed and available, and defaults to [~dev]. *)
-val default_version: 'a switch_state -> name -> version
```
rjbou, line 92: For consistency, better put it on `OpamPinned`

kit-ty-kate: done

kit-ty-kate: Actually I can’t put in in `OpamPinned`:
```
  Error: dependency cycle between modules in _build/default/src/state:
     OpamUpdate
  -> OpamPinned
  -> OpamSwitchState
  -> OpamPinned
```

rjbou: hmm...

bikallem:
Not sure if the issue belongs to this PR, but here is another issue:
I did `opam pin . --with-version 0.6.1` on root dir of repo/branch https://github.com/bikallem/resto/tree/no-lwt-4.12.0%2Bdomains. However, it seems opam pin doesn't respect the dependencies of opam packages. That is `resto` should be pinned/installed first and then `resto-lwt` and so on and so forth. However, this doesn't seem to be happening, I get the following error on my first try with the `opam pin . --with-version 0.6.1' command,
```
#=== ERROR while compiling resto-lwt.0.6.1 ====================================#
# context     2.1.1 | linux/x86_64 | ocaml-variants.4.12.0+domains | pinned(git+file:///home/bikal/ocamllabs/resto#no-lwt-4.12.0+domains#2b60417c)
# path        ~/.opam/4.12.0+domains/.opam-switch/build/resto-lwt.0.6.1
# command     ~/.opam/opam-init/hooks/sandbox.sh build dune build -p resto-lwt -j 47
# exit-code   1
# env-file    ~/.opam/log/resto-lwt-133343-63c2fb.env
# output-file ~/.opam/log/resto-lwt-133343-63c2fb.out
### output ###
# File "src/dune", line 11, characters 12-17:
# 11 |  (libraries resto lwt lwt.unix cohttp-lwt cohttp-lwt-unix conduit-lwt-unix))
#                  ^^^^^
# Error: Library "resto" not found.
# Hint: try:
#   dune external-lib-deps --missing --no-config --root . --ignore-promoted-rules --default-target @install --always-show-command-line --promote-install-files --release --only-packages resto-lwt -p resto-lwt --profile release -j 47 @install
```
I need to execute `opam pin . --with-version 0.6.1` several times for the packages to be pinned correctly. Each time the error is because dependent packages are not pinned first. Could this be addressed too in this PR or is this a separate issue altogether?

kit-ty-kate:
> That is resto should be pinned/installed first and then resto-lwt and so on and so forth.

@bikallem I’m not seeing the dependency to `resto` in your `resto-lwt` package anywhere

PR #4960, opened 2021-12-12
opam lint: Fix E67, check checksum only if vcs backend

kit-ty-kate:
Reported in https://discuss.ocaml.org/t/opamlint-error-67-checksum-specified-with-a-non-archive-url/8998

The problem with lint-check 67 (Checksum specified with a non archive url) is that it is checked using `OpamSystem.is_archive`, which expects as parameter: a file on the disk. However currently, only the base name from the url is given (which does not exist on the disk).

This is why the `zip` check fails: the file is not present on the filesystem and thus cannot be checked and returns `false`.
It only works for `tar.*` files because `OpamSystem.Tar.is_archive` also checks the extension if the file does not exists on the system.

Should render https://github.com/ocaml/opam/pull/4834 obsolete

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -100,6 +100,7 @@ users)
   * E57 (capital on synopsis) not trigger W47 (empty descr) [#5070 @rjbou]
   * [BUG] Fix linting packages from repository with tarred repositories, the file in temporary repository was no more avaiable when lint is done [#5068 @rjbou]
   * Update repository package filename display [#5068 @rjbou]
+  * E67: check checksums only for vcs urls [#4960 @rjbou]
 
 ## Repository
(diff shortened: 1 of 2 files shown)


dra27:
One to verify from ocaml-variants.3.09.1+metaocaml. The URL has a query portion: does the filename check for the archive type trigger here? From opam-repo-ci, it gets `error 67: Checksum specified with a non archive url: "https://github.com/metaocaml/metaocaml-archive/blob/master/MetaOCaml_309_alpha_030.tar.gz?raw=true - md5=28fca7ec424a1fd9fcdcf89fa2929f2a"`

rjbou:
Updated

dra27 (approved):
LGTM (I checked that the strange metaocaml archive URL works correctly too)

PR #4961, opened 2021-12-13
OpamSystem.real_path: Remove the double chdir trick on OCaml >= 4.13.0

### LINKED ISSUE #4793: TODO: Get rid of the double Unix.chdir hack from the libraries
While debugging https://github.com/ocurrent/opam-repo-ci/pull/120, @AltGr brought to my attention that some of the opam library functions (mainly `OpamProcess.create` and `OpamSystem.real_path` used in `OpamFilename.Dir.of_string`) use `Unix.chdir` twice (once to go in the directory and the second to go back to the previous directory) and are thus not thread-safe.

For `OpamProcess.create` it should be easy to fix: simply use `Unix.fork` + `Unix.execve` manually instead of `Unix.create_process_env`, but `OpamSystem.real_path` seems more tedious to reimplement.

* OCaml 4.13 added `Unix.realpath` to do just that (https://github.com/ocaml/ocaml/pull/10047, thanks @dbuenzli ❤️) so maybe it would be "good enough" to use this function when OCaml 4.13 is detected and the legacy (2012!) code otherwise.
* Another solution would be to port the C binding code from OCaml 4.13 in a similar manner to https://github.com/ocaml/opam/pull/4274

kit-ty-kate:
Partially fixes #4793

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,10 +41,10 @@ env:
   OPAM12CACHE: ~/.cache/opam1.2/cache
   # These should be identical to the values in appveyor.yml and the other workflows
   OPAM_REPO: https://github.com/ocaml/opam-repository.git
-  OPAM_TEST_REPO_SHA: ec07d6d5f363d1c5b9ed3b927d0df1d58dad333c
-  OPAM_REPO_SHA: ec07d6d5f363d1c5b9ed3b927d0df1d58dad333c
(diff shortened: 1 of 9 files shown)


kit-ty-kate:
Note for reviewers: I wasn’t able to move it to `OpamCompat` due to the `let p = OpamCoreConfig.(!r.log_dir) in` in `OpamSystem.getchdir` (cyclic dependency)

dra27:
Assuming it now passes, my two commits should go on 2.1 as well (it speeds up the build)

PR #4963, opened 2021-12-14
Add lock test

### LINKED ISSUE #4313: `opam install --locked` silently ignores the --locked option
The behavior of `install --locked` is really confusing, especially because there's no indication that anything is wrong.

The thing I would expect to work:

```bash
# Pin package(s) in the current repo but don't install deps yet
opam pin add -y -n .
# Install depexts
opam depext mypackage -y
# Now install deps
opam install . -y --locked
```

In this case, the `--locked` argument does absolutely nothing and the deps are installed only using mypackage.opam.

After reading `opam install --help`, I realized that `--locked` only works for with `opam pin`:

```bash
# Pin package(s) in the current repo but don't install deps yet
opam pin add -y -n .
# Install depexts
opam depext mypackage -y
# Now install deps in a really convoluted way
opam pin remove -y mypackage
opam pin add -y --locked .
```

There are two things I'd propose changing:

1. Please make `opam install --locked` use the lock file
2. Until (1) is done, make `opam install --locked` return an error message instead of silently ignoring the argument

rjbou:
~fix # 4313 # 4936~
Split into #5079  #5080 this one add only a test


tests/reftests/run.ml
```diff
@@ -571,6 +571,7 @@ let run_test ?(vars=[]) ~opam t =
                 (filter @ common_filters dir) str
             in
             print_string str
+            with Sys_error _ -> print_string (file ^ ": filenot found\n")
```
kit-ty-kate, line 574:
```suggestion
            with Sys_error _ -> print_string (file ^ ": file not found\n")
```

src/state/opamUpdate.ml
```diff
@@ -255,7 +255,8 @@ let pinned_package st ?version ?(working_dir=false) name =
     (* Do the update *)
     fetch_dev_package urlf srcdir ~working_dir ?subpath nv @@+ fun result ->
     let new_source_opam =
-      OpamPinned.find_opam_file_in_source name srcdir_find >>= fun f ->
+      let locked = OpamStateConfig.(!r.locked) <> None in
+      OpamPinned.find_opam_file_in_source ~locked name srcdir_find >>= fun f ->
```
rjbou, line 259: do we want that ? it's more coherent, but the only way to give a lock option to `opam update` is using the environment variable

master_changes.md
```diff
@@ -73,8 +76,14 @@ users)
...
 
+## Upgrade
+  * Handle lock files when upgrading pinned packages [#4963 @rjbou]
+
+## Update
+  * Handle lock files when environment variable is given [#4963 @rjbou]
+
 ## Opamfile
-  *
+  * Fix substring errors in preserved_format [#4941 @rjbou - fix #4936]
```
rjbou, line 86:
```suggestion
  * 
```

PR #4964, opened 2021-12-15
System: homogeneise is_archive tar & zip

rjbou:
If file exists check magic number, otherwise check extension


rjbou:
discussion from #4960

kit-ty-kate (approved):

kit-ty-kate:
Thanks!

Issue #4779, opened 2021-07-31
opam repository set-url --strict does not fail on errors


kit-ty-kate:
```
$ opam repository set-url --strict default "git://github.com/GgnDpSngh/opam-repository.git#ef9a4b1448581b3de004c85ce9409eaacee2bdd0"
[default] Initialised
[ERROR] In /home/opam/.opam/repo/default/packages/elina/elina.1.1/opam:
        This file is for package 'elina.1.1' but has mismatching fields 'version:1.3.
[ERROR] Strict mode: aborting
[ERROR] Could not update repository "default": OpamStd.OpamSys.Exit(30)
$ echo $?
0
```
This should return a non-zero exit code and fail, but it succeeds.

PR #4968, opened 2021-12-16
repository: on set-url fetch failure (sync or file error), revert url change an fetch old one

### LINKED ISSUE #4780: opam 2.1.0 has different behaviours on macOS on packages with mismatching version field
```
$ opam repository set-url --strict default "git://github.com/GgnDpSngh/opam-repository.git#ef9a4b1448581b3de004c85ce9409eaacee2bdd0"
[default] Initialised
[ERROR] In /home/opam/.opam/repo/default/packages/elina/elina.1.1/opam:
        This file is for package 'elina.1.1' but has mismatching fields 'version:1.3.
[ERROR] Strict mode: aborting
[ERROR] Could not update repository "default": OpamStd.OpamSys.Exit(30)
```
This succeeds (see #4779), but now on Linux:
```
$ opam pin add -k version -yn elina.1.1 1.1
elina is now pinned to version 1.1
```
succeeds, however on macOS the package does not even exist anymore (it existed before, see https://opam.ocaml.org/packages/elina/):
```
$ opam show elina
[ERROR] No package matching elina found
```

rjbou:
fix #4780 #4779

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -144,12 +144,12 @@ users)
   * When several checksums are specified, instead of adding in the cache only the archive by first checksum, name by best one and link others to this archive [#4696 rjbou]
   * Update opam repository man doc regarding removal of the last repository in a switch [#4435 - fixes #4381]
   * Don't display global message when `this-switch` is given [#4899 @rjbou - fix #4889]
-  * Set the priority of user-set archive-mirrors higher than the repositories'.
-    This allows opam-repository to use the default opam.ocaml.org cache and be more resilient to changed/force-pushed or unavailable archives. [#4830 @kit-ty-kate - fixes #4411]
+  * Set the priority of user-set archive-mirrors higher than the repositories'.  This allows opam-repository to use the default opam.ocaml.org cache and be more resilient to changed/force-pushed or unavailable archives. [#4830 @kit-ty-kate - fixes #4411]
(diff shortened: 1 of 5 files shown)


kit-ty-kate:
Thanks!

It feels a bit weird to have fetch the old one again. Is it not possible to have an atomic operation setting the repository config at the end instead of before fetching it?

rjbou:
There is two things: the repo config and repo content. The repository config is overwritten because of called function (minimal modifications makes it easier to overwrite it with the old one). And repository content, on new repository fetch/loading, the old repo is no more present (overwritten), so to keep a consistent state, better fetch the old one.
One solution would be to change the behavior to fetch in a temporary directory, check that there is no failure and then fetch in the internal directory.

kit-ty-kate (approved):
> One solution would be to change the behavior to fetch in a temporary directory, check that there is no failure and then fetch in the internal directory.

ah ok I didn’t know fetching a repository was a non-relocable action. I thought would have been easy to fetch in temporary directory, check the repository then do an atomic move to its final directory on success (or remove on failures).

But ok then. LGTM

dra27:
From dev meeting this morning: there's a slightly different slant on the same idea, which should work! Rather than atomic update at the end - which we can't do easily at the moment - we could back-up whatever is there beforehand (Git repo hash, tarball, directory) and if the set-url fails, then rename that backup back/reset to the SHA.

rjbou:
Updated accordingly, and added a test to check that it is rollbacked to its ineer state (and not remote one)

rjbou:
done :white_check_mark:

kit-ty-kate (approved):

PR #4969, opened 2021-12-16
Completely refactor the way orphan packages are handled (and add/port tests)

AltGr:
Orphan packages are packages which are installed within a switch, but no longer available. Without special handling, they become invisible to the solver, which has quite a few drawbacks:
- bad error messages
- forced removals upon unrelated actions
- removal instead of up/downgrade if another version is still available
- etc.

For these reasons, they need special care. The current approach is complex code in OpamClient that analyses them, their relationships with the user request, and lies about the universe so they can be kept when unrelated, or if they don't matter.

That code was tricky, complex, computing (reverse) dependency cones, intersections thereof and whatnot, costly — requiring a few preliminary conversions of the opam state to a solver universe, and, I found out, still incomplete and incorrect around the edges.

This PR implements a much simpler solution, that supplies the whole universe to the solver with a minimum of lies (orphans are included, but with their unavailability encoded as a fake dependency). Then, there is post-processing on the action graph after the solution is computed for removing sub-graphs that are disjoint from the requested actions, to avoid unrelated actions from being triggered.

The additional tests show a good behaviour in the considered cases (including some where the old system was incorrect).

IMPORTANT: major change compared to the previous behaviour
- you are no longer allowed to perform actions on dependents of orphan packages; this is more consistent and in accordance with the similar policy on reinstalls that was adopted some time ago.

RELATED THINGS TO CONSIDER:
- to ease the case above, consider adding an `opam pin --current <pkg>` that pins the package to its current state. Allows manual bypassing of an upstream removed packages, and would also offer a more consistent workaround for skipping pending reinstalls
- `opam upgrade --fixup` has been simplified because it used to consider the specific status of orphan packages. Maybe consider re-implementing it with something like `opam upgrade --criteria=fixup --best-effort --update-invariant` ?


kit-ty-kate (approved):
The code looks fine, the tests are clear and show that it works. LGTM

AltGr:
Note: this is now based on top of #4973, which includes most of the tests, and should be rebased once that is merged.

AltGr:
2 opam-rt tests are failing:
- reinstall test has been ported to reftests and can be discarded (there are some adjustments because of the no longer allowed reinstalls)
- I am still working on the `pin-advanced` test, and should have a fix shortly (on either side!)

AltGr:
Ok, now all should be ok:
- removed ported tests and a tiny tweak on https://github.com/ocaml-opam/opam-rt/pull/76
- turns out my last fix on this PR still had one blind spot, but that should now be OK.

AltGr:
Ok, only remaining failures should be fixed by the PR on opam-rt; apparently the trick that used the same-named branch of opam-rt for tests is no longer here, so we should probably merge #4973, then https://github.com/ocaml-opam/opam-rt/pull/76, then re-run CI for this PR.

src/state/opamSwitchState.mli
```diff
@@ -253,8 +253,10 @@ val unavailable_reason:
   'a switch_state -> ?default:string -> name * OpamFormula.version_formula ->
   string
 
-(** Returns whether or not the package can be upgraded to a version tagged with avoid-version *)
-val can_upgrade_to_avoid_version : OpamPackage.Name.t -> 'a switch_state -> bool
+(** Returns [true] when the package has the [avoid-version] flag and there isn't
+    already a version with that flag installed (which disables the
+    constraint) *)
+val avoid_version : 'a switch_state -> package -> bool
```
kit-ty-kate, line 259: Since this PR also changes the code for the handling of the `avoid-version` flag, i just noticed there wasn’t any tests for this flag. Could you add some if you have some time?

kit-ty-kate (commented):
LGTM aside from a small request

src/state/opamSwitchState.mli
AltGr: I moved some code around and replaced a Set.fold+exists by a filter for efficiency, but that should be functionnally equivalent; indeed there is no fundamental reason for this to be in this PR, I just stumbled upon it while I was writing the PR.

You're right about tests — though I'd prefer not to stall this and add them in another PR.

Issue #4112, opened 2020-03-16
opam init failed (parallel build?)


morrocoya:
I am unable to finish opam installation on mac OS (4 core, I5).
`opam init` fails while building ocaml-base-compiler, apparently in relation with parallel building. `opam init --jobs=1` gives similar outcome.

```
#=== ERROR while compiling ocaml-base-compiler.4.10.0 =========================#
# context     2.0.6 | macos/x86_64 |  | https://opam.ocaml.org#f16912c3
# path        ~/.opam/default/.opam-switch/build/ocaml-base-compiler.4.10.0
# command     ~/.opam/opam-init/hooks/sandbox.sh build ./configure --prefix=/Users/aponte/.opam/default -C CC=cc ASPP=cc -c
# exit-code   1
# env-file    /var/folders/_z/gr_lx9n153vfqj45h36l25hw0000gp/T/opam-aponte-20461/ocaml-base-compiler-20461-736596.env
# output-file /var/folders/_z/gr_lx9n153vfqj45h36l25hw0000gp/T/opam-aponte-20461/ocaml-base-compiler-20461-736596.out
### output ###
# [...]
# checking sys/shm.h usability... yes
# checking sys/shm.h presence... yes
# checking for sys/shm.h... yes
# checking for dirent.h... yes
# checking for sys/select.h... yes
# checking for off_t... yes
# checking size of int... 0
# checking size of long... 0
# checking size of long *... 0
# checking size of short... 0
# checking size of long long... 0
# configure: error: Neither 32 nor 64 bits architecture.



<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫 
┌─ The following actions failed
│ λ build ocaml-base-compiler 4.10.0
└─ 
<><> ocaml-base-compiler.4.10.0 troubleshooting <><><><><><><><><><><><><><>  🐫 
=> A failure in the middle of the build may be caused by build parallelism
      (enabled by default).
      Please file a bug report at https://github.com/ocaml/ocaml/issues
```

edit rjbou: formatting

rjbou:
It is more an ocaml configure error of your system detection
```
# checking size of int... 0
# checking size of long... 0
# checking size of long *... 0
# checking size of short... 0
# checking size of long long... 0
# configure: error: Neither 32 nor 64 bits architecture.
```

rjbou:
Any news on this one ?

morrocoya:
[...]

PR #4971, opened 2021-12-16
Print user-friendly error messages when a command is not found, before calling the sandbox

kit-ty-kate:
bwrap for instance will exit with the confusing "bwrap: execvp <cmd>: No such file or directory"

Example of such confused user can be found here: https://github.com/ocaml/opam/issues/4112#issuecomment-996208463

### DIFF
--- a/src/state/shellscripts/bwrap.sh
+++ b/src/state/shellscripts/bwrap.sh
@@ -125,6 +125,11 @@ case "$COMMAND" in
         exit 2
 esac
 
+if ! command -v "$1" >/dev/null; then
+    echo "[ERROR] Command not found: $1" >&2
+    exit 10
+fi
+
 # Note: we assume $1 can be trusted, see https://github.com/projectatomic/bubblewrap/issues/259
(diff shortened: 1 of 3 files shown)


src/state/shellscripts/bwrap.sh
```diff
@@ -125,6 +125,11 @@ case "$COMMAND" in
         exit 2
 esac
 
+if ! command -v "$1" >/dev/null; then
+    echo "[ERROR] Command not found: $1" >&2
```
rjbou, line 129: Wondering, `echo -e` is not portable?

kit-ty-kate: We can leave that to another PR. It’s the way echo is used everywhere else in the sandbox scripts

kit-ty-kate: `-e` is not POSIX complient: https://www.unix.com/man-page/posix/1posix/echo/

rjbou:
Thanks!

PR #4981, opened 2022-01-04
Add some debug log to OpamCudf.extract_explanations to help debug #4373

kit-ty-kate:
The following dockerfile (derived from https://github.com/ocaml/opam/issues/4373#issuecomment-992725743):
```
FROM ocaml/opam:debian-11-ocaml-4.12@sha256:40813dc1391636b4a428afbf0e678d2c81b80d9eade8a5214d8ee271fae155f3
RUN git clone https://github.com/kit-ty-kate/opam.git -b extract_explanations-debug
WORKDIR opam
RUN opam exec -- ./configure --with-vendored-deps
RUN opam exec -- make
RUN sudo ln -f _build/default/src/client/opamMain.exe /usr/bin/opam
RUN opam init --reinit -ni
RUN opam remote set-url default git+https://github.com/ocaml/opam-repository.git#a5d7cdc0c91452b0aef4fa71c331ee5237f6dddd
RUN opam pin add --debug --debug-level=3 --with-version 0.6.1 git+https://github.com/bikallem/resto.git#e725f56dd2d5ca0148272d754a0f29ffe76f94e8
```
now reports:
```
00:10.228  CLIENT                          conflict!
[ERROR] Package conflict!
00:10.228  CUDF                            Conflict reporting
00:10.228  CUDF                            Reasons: [
  Dependency (dose-dummy-request.???, [resto-cohttp-server], [resto-cohttp-server.0.6.1]);
  Dependency (resto-cohttp-server.0.6.1, [cohttp-lwt-unix>=4.0.0], [cohttp-lwt-unix.4.0.0]);
  Dependency (resto-cohttp-server.0.6.1, [conduit-lwt-unix>=5.0.0], [conduit-lwt-unix.5.0.0]);
  Conflict (conduit-lwt-unix.1.0.3, conduit-lwt-unix.5.0.0, conduit-lwt-unix);
  Conflict (conduit-lwt-unix.1.4.0, conduit-lwt-unix.5.0.0, conduit-lwt-unix);
  Conflict (conduit-lwt-unix.4.0.1, conduit-lwt-unix.5.0.0, conduit-lwt-unix);
  Conflict (conduit-lwt-unix.1.5.0, conduit-lwt-unix.5.0.0, conduit-lwt-unix);
  Conflict (conduit-lwt-unix.1.2.0, conduit-lwt-unix.5.0.0, conduit-lwt-unix);
[...]

PR #4982, opened 2022-01-05
Fix empty conflict explanations

### LINKED ISSUE #4373: Unhelpful conflict messages
### *** Please add your unhelpful conflict message to this [discussion](https://github.com/ocaml/opam/discussions/5130) ***
---

While trying to install `expect.0.0.6` on OCaml 4.12:
```
[ERROR] Package conflict!
No solution found, exiting
    - (invariant) -> ocaml-variants = 4.12.0+trunk -> ocaml >= 4.12.0
```
I would have expected:
```
```
as the latest version of `batteries` indicates.

kit-ty-kate:
Fixes #4373

### DIFF
--- a/src/solver/opamCudf.ml
+++ b/src/solver/opamCudf.ml
@@ -854,7 +854,10 @@ let extract_explanations packages cudfnv2opam reasons : explanation list =
               Map.update pkg (CS.union chain) CS.empty acc
             in
             let ds = get deps p in
-            let dsc = ds %% all_conflicting in
+            let dsc = match Hashtbl.length missing with
+              | 0 -> ds (* Hack to fix https://github.com/ocaml/opam/issues/4373. We should try to do better at some point *)
+              | _ -> ds %% all_conflicting
+            in
             if not (Set.is_empty dsc) then
               dsc ++ seen1, Set.fold append_to_chains (dsc -- seen1) new_chains
             else
--- a/master_changes.md
+++ b/master_changes.md
@@ -160,6 +160,7 @@ users)
 ## Client
   * Check whether the repository might need updating more often [#4935 @kit-ty-kate]
   * ✘ It is no longer possible to process actions on packages that depend on a package that was removed upstream [#4969 @altgr]
+  * Fix (at least some of the) empty conflict explanations [#4982 @kit-ty-kate]
 
 ## Internal
   * Add license and lowerbounds to opam files [#4714 @kit-ty-kate]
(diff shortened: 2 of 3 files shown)


src/solver/opamCudf.ml
```diff
@@ -848,7 +902,7 @@ let extract_explanations packages cudfnv2opam reasons : explanation list =
               Map.update pkg (CS.union chain) CS.empty acc
             in
             let ds = get deps p in
-            let dsc = ds %% all_conflicting in
+            let dsc = ds in
```
kit-ty-kate, line 905: it looks like this line alone also fixes the issue.. I’ll have a look

src/solver/opamCudf.ml
```diff
@@ -969,6 +1026,9 @@ let extract_explanations packages cudfnv2opam reasons : explanation list =
         | _ -> false)
   in
   match explanations with
+  | [] ->
+    OpamConsole.error_and_exit `Internal_error
+      "Internal error while computing conflict explanations: sorry, please report."
```
AltGr, line 1031: In general a conflict is the end of story for opam commands... still, it feels weird to hard error out from a function which sole purpose is to print an informative message. Does returning the "internal error" code really matter here, or couldn't we print the error message but just return `[]` ?

src/solver/opamCudf.ml
```diff
@@ -854,7 +908,10 @@ let extract_explanations packages cudfnv2opam reasons : explanation list =
               Map.update pkg (CS.union chain) CS.empty acc
             in
             let ds = get deps p in
-            let dsc = ds %% all_conflicting in
+            let dsc = match Hashtbl.length missing with
+              | 0 -> ds
+              | _ -> ds %% all_conflicting
+            in
```
AltGr, line 914: This feels a bit fragile; maybe we could at least check that there is no "missing" below `p` in the dep tree rather than in general ?

This said, it already avoids the very problematic issue of no messages at all, and thank you *a lot* for finding a workaround.

src/solver/opamCudf.ml
kit-ty-kate: Note: this is part of https://github.com/ocaml/opam/pull/4981.

Mmh, we could make it into `OpamConsole.error` instead but this would make opam print two different user direction:
```
[ERROR] Package conflict!
[ERROR] Internal error while computing conflict explanations: sorry, please report.
Sorry, no solution found: there seems to be a problem with your request.
```
I’d rather push users to report instead of making them try to fiddle with the request and maybe give up reporting.

AltGr:
This may not be perfect, but already works around most problematic cases in practice, so I think it should be merged (once rebased). Thanks Kate!

PR #4983, opened 2022-01-05
Add tests for #4373, and the avoid-version flag

AltGr:


tests/reftests/conflict-resto.test
```diff
@@ -0,0 +1,416 @@
+a5d7cdc0c914
```
kit-ty-kate, line 1: It might be simpler and less costly to run to just have the simple generic case described in https://github.com/ocaml/opam/issues/4373#issuecomment-1006121785 instead

AltGr: Indeed, and that would be interesting as well; I'd say "real-world" cases have value too, and the cost of extracting the opam root is only for the first run though

PR #4986, opened 2022-01-07
[2.1 backport] Speed up and fix AppVeyor runs

kit-ty-kate:
This fixes issues with the Cygwin run when the bootstrap ocaml compiler does not match the compiler being tested in appveyor.
It also brings faster CI runs.

Backports to 2.1 as mentioned in https://github.com/ocaml/opam/pull/4961#issuecomment-1007509058

### DIFF
--- a/appveyor_test.sh
+++ b/appveyor_test.sh
@@ -1,5 +1,7 @@
 #!/bin/bash
 
-opam init -y -a --compiler=ocaml.4.12.0 git+https://github.com/ocaml/opam-repository#$OPAM_TEST_REPO_SHA
+COLD_OCAML="$(sed -ne 's/URL_ocaml = .*ocaml-.*ocaml-\(.*\)\.tar\.gz/ocaml.\1/p' src_ext/Makefile)"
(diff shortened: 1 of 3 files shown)


appveyor_test.sh
```diff
@@ -2,6 +2,6 @@
 
 COLD_OCAML="$(sed -ne 's/URL_ocaml = .*ocaml-.*ocaml-\(.*\)\.tar\.gz/ocaml.\1/p' src_ext/Makefile)"
 # This is supposed to be able to select ocaml-system using the bootstrap compiler
-opam init -y -a --compiler=$COLD_OCAML git+$OPAM_REPO#$OPAM_TEST_REPO_SHA
+opam init -y -a --compiler=$COLD_OCAML git+https://github.com/ocaml/opam-repository#$OPAM_TEST_REPO_SHA
```
dra27, line 5: This line was correct - we need this line from master in appveyor.yml:

https://github.com/ocaml/opam/blob/a2c0e226b9d47fd972dacd2d9387f2f3420a8412/appveyor.yml#L15

Issue #4925, opened 2021-11-22
Warn about actions on pinned packages


ulugbekna:
Currently, if a user has a package pinned, they can reinstall a certain version of the package, but they'll still have the pinned version. Say, I pin ocamlformat master, then reinstall 0.19.0 successfully as indicated by opam install, but I'll still have the master version of ocamlformat when I run it. Does it make sense to warn that the package being (re)installed is actually pinned and the reinstallation, for example, won't do anything? (it probably still changes versions for the solver to see)

rjbou:
On its display opam already say you that it is synchronising pinned packages, even if there is not changes:

```
<><> Synchronising pinned packages ><><><><><><><><><><><><><><><><><><><><><><>
[name.version] synchronised (no changes)

The following actions will be performed:

```

dra27:
Notes from dev meeting: we should change the display of pinned versioned so that instead of just being `version*` it is `version (pinned)`. Separately, we should also consider using a suffix on the version by default (although this may potentially break workflows, so it's worth considering the two changes separately)

PR #4987, opened 2022-01-07
Make the status of pinned packages more explicit

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/4925

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -28,7 +28,7 @@ users)
   *
 
 ## Install
-  *
+  * Make the status of pinned packages more explicit during installation [#4987 @kit-ty-kate - fix #4925]
 
 ## Remove
   *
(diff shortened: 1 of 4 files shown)


rjbou:
Just a thought: can we use the pin character 🖈 (U+1F588) if available, otherwise `pinned` or another one character more visible and intelligible than `*`? see [OpamConsole.utf8_symbol](https://github.com/ocaml/opam/blob/master/src/core/opamConsole.ml#L120)

kit-ty-kate:
> Just a thought: can we use the pin character 🖈 (U+1F588) if available, otherwise `pinned` or another one character more visible and intelligible than `*`? see [OpamConsole.utf8_symbol](https://github.com/ocaml/opam/blob/master/src/core/opamConsole.ml#L120)

That can always be done later but I think this is a bad idea anyway. A pushpin emoji such as 📌 (U+1F4CC) is still confusing and implicit for new comers who do not know about the pin system and have just copy/pasted command lines around (or simply used `opam install .`)

rjbou:
In the same way, they won't understand neither the `(pinned)` message, no ? But they will more easily look for it as it is a word.

Issue #4566, opened 2021-02-25
[RFC] New 'revision' field and variable


kit-ty-kate:
## Current state of things

In https://github.com/ocaml/opam-repository/issues/10531, users are rightfully pointing that any changes made in opam-repository can break things.

The only workaround for users that we have is "try to find the git commit that broke things and `opam repository set-url` it", which isn't ideal.

The only workaround for opam-repository maintainers that we have is "try not to break things and if a new package is modified in a significant way, try to create a new `<pkg>.<ver>-1` package with the changes", which is both not ideal and very annoying because:
- we have to draw the line of when to do a revision which is a fine line (doing a new revision package for every single modifications is not an option, we'd be nearing a trillion packages at this point..)
- it keeps the old unmodified versions around and polluting the repository with version of packages that should not be used by default
- adding a new revision package is not possible from the github interface and thus annoying to do for day to day maintainance
- the new revision will not be used by packages which uses `= "1.0.1"` or `<= "1.0.1"` type of constraints. We try to invite people to avoid them but some still go through (e.g. some tezos packages) or are advised (e.g. `{= version}`)

## The proposal

To fix all this issue I'm proposing that next time we break the opam file compatibility (most likely opam 3.0), a new `revision` field and variable would be added.
What would they do?

The new `revision` field would:
- Be a natural number
- Be set to `0` if not set
- Tell what is the current (major) revision of the package
- Would make any request to a `<pkg>.<ver>-<revision>` package an alias to `<pkg>.<ver>` (at the revision `<revision>`)
  + Other separators than `-` could be used to avoid breaking compatibility. e.g. `:`

This field would work in tandem with its variable counterpart.
The new `revision` variable would:
- Be available in every filter
- Be set to the current `revision` variable if not given
- Be set to any of the previous revisions if requested explicitly by `opam install <pkg>.<ver>-<revision>`

## Example

```
opam-version: "2.0"
revision: "3"
maintainer: "pietro.abate@inria.fr"
authors: [...]
homepage: "http://www.mancoosi.org/software/"
bug-reports: "https://gitlab.com/irill/dose3/-/issues"
license: "LGPL-3.0-or-later with OCaml-LGPL-linking-exception"
dev-repo: "git+https://gitlab.com/irill/dose3.git"
build: [
  ["./configure"]
  [make "printconf"]
  [make] {revision < "2"}
  [make "libs" "apps"] {revision >= "2"}
]
install: [make "installlib"]
remove: [
[...]

PR #4996, opened 2022-01-14
Use simplified version comparison function

### LINKED ISSUE #4272: Why is going from version ~x to version ~x~y an upgrade?
During upgrade I saw these updates:
```
  - upgrade   coq-compcert-64        3.7~coq-platform to 3.7~coq-platform~open-source
  - upgrade   coq-compcert           3.7~coq-platform to 3.7~coq-platform~open-source
```
I wonder why going from `3.7~coq-platform` to `3.7~coq-platform~open-source` is considered an upgrade. Shouldn't according to the documentation `3.7~coq-platform~open-source` be smaller in the version ordering than `3.7~coq-platform` since `~` is smaller than any character, including end of line?

See: (https://opam.ocaml.org/doc/Manual.html#Package-Formulas)

It looks like `~` is considered as smaller than end of line only at the beginning of a word, but not in the middle of a word.

AltGr:
As specified in the docs, and not as handled in Debian: we used to run the 
Debian comparison from Dose, which handled epoch prefixes and revision
suffixes in an undocumented way (undocumented in opam, not blaming Dose
here!).

Closes #4272

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -144,7 +144,8 @@ users)
   * Actually allow multiple state caches to co-exist [#4934 @dra27 - fix #4554 properly this time]
 
 ## Opam file format
-  *
+  * ✘ Version comparison is now strictly as specified in the docs, and doesn't
+    handle Debian-like "epoch" prefixes or "revision" suffixes [@AltGr - fix #4272]
 
 ## Solver
   * [BUG] Remove z3 debug output [#4723 @rjbou - fix #4717] [2.1.0~rc2 #4720]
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
I thought we had agreed in #4566 to keep it to handle revisions

AltGr:
Huh, I am sorry if I skipped that part of a discussion... but I don't see a link between the explicit package revisions discussed in #4566 and the general version-comparison function (on strings) changed here. It may indeed be possible to implement #4566 using debian-like revision syntax, but I don't think it would be a good idea to squeeze that info into the version string instead of using a separate record field.

rjbou (approved):

kit-ty-kate (changes requested):
I don’t think we should change this now without discussing it extensively. This is really not urgent and this break more things than it tries to solve.

PR #5002, opened 2022-01-18
Avoid trying to be too clever when upgrading switches from opam 2.0

### LINKED ISSUE #4501: [opam.2.1.0~beta4] Migration from 2.0 removes all constraints in the invariant
```
$ opam-2.0 switch
#   switch  compiler                    description
->  4.11    ocaml-base-compiler.4.11.1  4.11
$ opam-2.1 switch invariant
["ocaml-base-compiler"]
```
This is a problem. This should results in `["ocaml-base-compiler" {= "4.11.1"}]` instead.

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/4501
See https://github.com/ocurrent/opam-repo-ci/pull/148
See https://github.com/ocurrent/docker-base-images/pull/150

I don’t see any good reason for trying to guess what the user meant instead of letting the user choose what they want afterward.

Transforming a 2.0 switch to `"<compiler-package>"` (without constraint) is dangerous (as shown above)
Transforming a 2.0 switch to `"<compiler-package>" {>= "<current-version>"}` is equally dangerous as it breaks switches named after its current compiler if the compiler is the latest. We can’t assume people with the latest compiler always want the latest

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -65,6 +65,7 @@ users)
   * [BUG] Fix `set-invariant: default repos were loaded instead of switch repos [#4866 @rjbou]
   * Add support for `opam switch -` (go to previous non-local switch) [#4910 @kit-ty-kate - fix 4866]
   * On loading, check for executable external files if they are in `PATH`, and warn if not the case [#4932 @rjbou - fix #4923]
+  * Avoid trying to be too clever when upgrading switches from opam 2.0 [#5002 @kit-ty-kate - fix #4501]
 
 ## Pin
   * Switch the default version when undefined from ~dev to dev [#4949 @kit-ty-kate]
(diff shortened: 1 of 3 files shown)


AltGr:
I don't really have an opinion on removing the magic... the cases pointed out in the above are very specific to a convoluted CI setup, and the code might still be helpful for normal users (the "remove the constraint" part was intended for system compilers, where I think it results in a better UX) ; then again, magic is bad in general and being conservative as proposed ensures that the behaviour remains similar after the opam version bump.

In the end my main worry is that it may be late to change this: we may start having the issue of both choices , depending on the migration path followed by the user...

dra27:
As noted in the end of the description in #5176, if we got to do the upgrade process, we'd probably opt to skip have the `>=` invariant trick, but it seems better to have 2.0->2.1 and 2.0->2.2 do the same upgrade, so we'll leave the main bug as fixed by #5176

PR #5005, opened 2022-01-18
Big speedup on reverse dependency calculation

AltGr:
We need to revert a huge simply-linked structure; the big cost was allocating 
the intermediate data, so instead of creating the graph or going through sets, 
we use the integer identifiers already defined by Cudf and store that in a 
simple table.

### DIFF
--- a/src/solver/opamSolver.ml
+++ b/src/solver/opamSolver.ml
@@ -91,7 +91,7 @@ let cudf_versions_map universe packages =
 
 let name_to_cudf name =
   let name_s = OpamPackage.Name.to_string name in
-  if String.for_all (function
+  if OpamStd.String.for_all (function
       | 'a'..'z' | 'A'..'Z' | '0'..'9' | '@' | '/' | '+' | '(' | ')' | '.' | '-'
         -> true
       | _ -> false)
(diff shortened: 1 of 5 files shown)


kit-ty-kate (approved):
Indeed! I can confirm an almost 2x speedup on `--depends-on` + `--recursive`.
e.g. `opam list --depends-on dune --recursive -s --all-versions` take
* 9.4s before this PR
* 5.1s with this PR

And no difference in the result :tada:

No noticable difference could be obversed with non-recursive queries

src/solver/opamSolver.ml
```diff
@@ -90,7 +90,14 @@ let cudf_versions_map universe packages =
     pmap OpamPackage.Map.empty
 
 let name_to_cudf name =
-  Dose_common.CudfAdd.encode (OpamPackage.Name.to_string name)
+  let name_s = OpamPackage.Name.to_string name in
+  if String.for_all (function
```
kit-ty-kate, line 94: String.for_all is only available since OCaml 4.13. We probably need to introduce `OpamStd.String.for_all`

rjbou:
All these `oppam list` PRs highlighted that there is no opam list test... Feel free to add one, or i can add it later

kit-ty-kate:
Done in https://github.com/ocaml/opam/pull/5006

AltGr:
Test failure doesn't seem related to the PR :confused:

PR #5006, opened 2022-01-18
Add some simple tests for the "opam list" command

kit-ty-kate:


AltGr:
Thanks!
But, hum, 46000 lines of test may be a tad overkill ? :grin: Would e.g. adding a pattern e.g. `a*` to the command decrease the value of the test ?

kit-ty-kate:
Fair enough, I replaced `dune` with `lacaml` (on which packages are depending on some specific versions and sometimes also optionally, which makes it a good test case). Only 231 lines this time!

PR #5007, opened 2022-01-19
Fix opamroot-versions.test glitch

AltGr:
I had to use `[)]` to escape rather than `\\)`, because backslashes interfere 
with the windows-path rewrite rules...

### DIFF
--- a/tests/reftests/opamroot-versions.test
+++ b/tests/reftests/opamroot-versions.test
@@ -232,126 +232,126 @@ unsupported or missing file format version; should be 2.0 or older
 ### :III:1:a: Bad config file
 ### cp config.4.8.err $OPAMROOT/config
 ### # ro global state
-### opam switch | "(${OPAMROOTVERSION})" -> "current"
+### opam switch | "[(]${OPAMROOTVERSION}[)]" -> "current"
 FORMAT                          File errors in ${BASEDIR}/OPAM/config, ignored fields: At ${BASEDIR}/OPAM/config:3:0-3:8::
 Invalid field neant
 GSTATE                          LOAD-GLOBAL-STATE @ ${BASEDIR}/OPAM
 FORMAT                          File errors in ${BASEDIR}/OPAM/config, ignored fields: At ${BASEDIR}/OPAM/config:3:0-3:8::
 Invalid field neant
-GSTATE                          root version (4.8) is greater than running binary's (current); load with best-effort (read-only)
+GSTATE                          root version (4.8) is greater than running binary's current; load with best-effort (read-only)
 #  switch  compiler  description
 ### :III:1:b: Good config file
 ### cp config.4.8 $OPAMROOT/config
 ### # ro global state
-### opam switch | "(${OPAMROOTVERSION})" -> "current"
+### opam switch | "[(]${OPAMROOTVERSION}[)]" -> "current"
 GSTATE                          LOAD-GLOBAL-STATE @ ${BASEDIR}/OPAM
-GSTATE                          root version (4.8) is greater than running binary's (current); load with best-effort (read-only)
+GSTATE                          root version (4.8) is greater than running binary's current; load with best-effort (read-only)
 #  switch  compiler  description
 ### :III:2:a: Bad repo config file :
 ### cp repos-config.err $OPAMROOT/repo/repos-config
 ### # ro global state, ro repo state
-### opam list --no-switch | "(${OPAMROOTVERSION})" -> "current"
+### opam list --no-switch | "[(]${OPAMROOTVERSION}[)]" -> "current"
 GSTATE                          LOAD-GLOBAL-STATE @ ${BASEDIR}/OPAM
-GSTATE                          root version (4.8) is greater than running binary's (current); load with best-effort (read-only)
+GSTATE                          root version (4.8) is greater than running binary's current; load with best-effort (read-only)
 RSTATE                          LOAD-REPOSITORY-STATE @ ${BASEDIR}/OPAM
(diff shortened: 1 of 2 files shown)


tests/reftests/opamroot-versions.test
```diff
@@ -836,11 +836,11 @@ The following actions will be performed:
...
 -> installed i-am-package.2
 Done.
-### opam option jobs=4 | "(${OPAMROOTVERSION})" -> "current"
+### opam option jobs=4 | "[(]${OPAMROOTVERSION}[)]" -> "current"
 GSTATE                          LOAD-GLOBAL-STATE @ ${BASEDIR}/OPAM
 [WARNING] Removing global switch 'this-internal-error' as it no longer exists
-FMT_UPG                         Light config upgrade, from 2.0 to current
-This version of opam requires an update to the layout of ${BASEDIR}/OPAM from version 2.0 to version current, which can't be reverted.
+FMT_UPG                         Light config upgrade, from 2.0 to 2.1
+This version of opam requires an update to the layout of ${BASEDIR}/OPAM from version 2.0 to version 2.1, which can't be reverted.
```
kit-ty-kate, line 843: matching against `[(]2.1[)]` will not work when `2.1` is not in a parenthesis.

kit-ty-kate:
Can’t we simply have a special rules that says "if we have a variable inside the regexp part, then take it as a static string and not as a regexp" ?

AltGr:
yes, that would be better, but more tedious to implement

tests/reftests/opamroot-versions.test
AltGr: of course not, it's the point

AltGr: all the replacements in the test file were already made to match more than just the version number (I guess because of similar issues ?), in some case it matches also the surrounding quotes, in some others the next token. Here it was the parens, but they were ignored since put in a regexp unescaped.

kit-ty-kate: I don’t understand your point. Why would we want an string-replace that doesn’t replace?

AltGr: oh indeed, in the hurry was too eager in my rewrite. The patch is fine for parts that were problematic, e.g. [here](https://github.com/ocaml/opam/pull/5007/files#diff-02dcea38573b45faeca1b186a45f675f54b60c70337fdbf8f15babbd75b3c1c3L253-R255) where we *do* want to match the parens, and that allows us to discriminate. But indeed down the line it misses some rewrites.

AltGr: (I just assumed too quickly that the parens in the regexp where expected as literals originally, as they don't make much sense otherwise)

tests/reftests/opamroot-versions.test
```diff
@@ -938,12 +938,12 @@ opam-version: "2.0"
 roots: ["i-am-package.2" "i-am-sys-compiler.2"]
 ### :V:1:e: reinit from 2.0
 ### ocaml generate.ml 2.0
-### opam init --reinit --bypass-checks --no-setup | "(${OPAMROOTVERSION})" -> "current"
+### opam init --reinit --bypass-checks --no-setup | "[ (]${OPAMROOTVERSION}($|[ ,)])" -> " current"
```
kit-ty-kate, line 941:
```suggestion
### opam init --reinit --bypass-checks --no-setup
```
Why not simply remove theme all together? We can simply update the tests when we bump the opam-root-version

rjbou: the idea of https://github.com/ocaml/opam/commit/c41b8daa2a5842addbd57c96aa2b93367bc591da in #4913 is to not have to bump the test at each opam-root-version bump

rjbou: well, to make the update minimal, [here](https://github.com/ocaml/opam/pull/4926/commits/40314fde489897d0cfccfd36f0ac722b769d35ad) is a example of update for bump

kit-ty-kate: i think it would be fine. It’s just one `dune runtest --auto-promote` away so I don’t see any downsides removing these. I only see downsides keeping them in

rjbou: The new output is not just taken as is, it need to be checked, updated, etc. ; and by reduced the diffs, it's simpler.
I'm missing the downsides of keeping them ? the regexp issue is fixed by #5009

kit-ty-kate: Keeping it you have to maintain them (e.g. what this PR is trying to do) and it makes it harder for us to read the reftests (e.g. you can’t tell if "current" comes from the sed or from whatever in opam) Also sometimes you might want something to not change for whatever reason and making it not change makes it harder to check that.

rjbou: For the name of the replacement, we can use another word that is not used and not meant to be used in opam, like `current-opamroot-version-for-test` that is human readable and should appear in opam messages (or if we are all in a hurry at that moment, but there will be more serious concerns than this test in that case :D).
For maintenance, reftests evolve, reftests engine evolves too. Something was merged with a buggy behaviour that nobody saw. My thought on keeping it is to try to have both world: fixed bug & enhanced reftest (ideally stable).
I agree with the no change that hide issues, It went into rabbit hole when debugging rec pin on windows with reftest because I didn't take into account path rewriting on reftest output. Maybe mistakenly, but here I assume that hiding is not so bug prone as it is consistent with the defined opam root version (the variable is set using opam output), and if something is broken in this test it's not just the version that will be changed, other lines will be highlighted in the diff.

AltGr: We're all overthinking this :D
Not certain it's worth it, but now that it's done let's merge, we keep our hands free to simplify in case of problems in the future.

PR #5015, opened 2022-01-21
Improve performance by disabling repo tarring

### LINKED ISSUE #4586: IO bottleneck: repository decompression
When calling `opam install <pkg>` one of the first processes opam calls are:
```
/usr/bin/tar xfz /home/opam/.opam/repo/default.tar.gz -C /tmp/opam-5953-44547
....
rm -rf /tmp/opam-5953-44547/default
```
These two commands seem to be the main IO bottleneck in CI where `/tmp` is not tmpfs for technical reason (cached containerization). I feel like we might see a good performance improvement by decompressing the repository in OCaml directly probably in some kind of stream mode if the thing that needs the repository information can be used this way.

Partially related to the later discussion in https://github.com/ocaml/opam/issues/3050

rjbou:
~that need a new name, ftm it is hideously named `OPAMNOREPOTARRING`~
/cc @AltGr 
fix #4586 
edit:
By default disabled,  environment variable `OPAMREPOSITORYTARRING` permit to enable it.

### DIFF
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -209,6 +209,12 @@ let environment_variables =
        'download-command' value from the main config file.";
       "NOCHECKSUMS", cli_original, (fun v -> NOCHECKSUMS (env_bool v)),
       "enables option --no-checksums when available.";
+      "REPOSITORYTARRING", cli_from cli2_2,
+      (fun b -> REPOSITORYTARRING (env_bool b)),
+      "internally store the repositories as tar.gz files. This can be much \
+       faster on filesystems that don't cope well with scanning large trees \
+       but have good caching in /tmp. However this is slower in the \
+       general case.";
       "REQUIRECHECKSUMS", cli_original,
       (fun v -> REQUIRECHECKSUMS (env_bool v)),
       "Enables option `--require-checksums' when available \
--- a/tests/reftests/init.test
+++ b/tests/reftests/init.test
@@ -1,6 +1,6 @@
 009e00fa
 ### OPAMYES=1
-### tar xzf ${OPAMROOT}/repo/default.tar.gz
+### cp -R ${OPAMROOT}/repo/default .
 ### rm -rf ${OPAMROOT}
 ### <opamrc>
 eval-variables: [ sys-ocaml-version ["false"] "no system compiler" ]
--- a/tests/reftests/repository.test
+++ b/tests/reftests/repository.test
@@ -1,0 +1,56 @@
+N0REP0
+### <pkg:foo.1>
+opam-version: "2.0"
+build: ["test" "-f" "bar"]
+### <pkg:foo.1:bar>
+some content
+### opam switch create tarring --empty
+### : Internal repository storage as archive or plain directory :
+### opam update -vv | grep '^\+' | '.*(/|\\)diff(\.exe)?"? "' -> 'diff "'
+diff "-ruaN" "default" "default.new" (CWD=${BASEDIR}/OPAM/repo)
+### ls $OPAMROOT/repo | grep -v "cache"
+default
+lock
+repos-config
+### opam install foo -vv | grep '^\+' | '.*(/|\\|")test(\.exe)?"? "' -> 'test "'
+test "-f" "bar" (CWD=${BASEDIR}/OPAM/tarring/.opam-switch/build/foo.1)
+### opam remove foo
+The following actions will be performed:
+  - remove foo 1
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> removed   foo.1
+Done.
+### OPAMREPOSITORYTARRING=1
+### opam repository add tarred ./REPO --set-default
+[tarred] Initialised
+### ls $OPAMROOT/repo | grep -v "cache"
+default
+lock
+repos-config
+tarred.tar.gz
+### opam repository remove default --all
+### opam update -vv | grep '^\+' | '.*tar(\.exe)?"? "xfz" ".*(/|\\)OPAM(/|\\)repo(/|\\)tarred\.tar\.gz" "-C" ".*"' -> 'tar xfz OPAM/repo/tarred.tar.gz -C TMPDIR' | '.*tar(\.exe)?"? "cfz" ".*(/|\\)OPAM(/|\\)repo(/|\\)tarred\.tar\.gz\.tmp" "-C" ".*" "tarred"' -> 'tar cfz OPAM/repo/tarred.tar.gz.tmp -C TMPDIR tarred' | '.*(/|\\)diff(\.exe)?"? "' -> 'diff "'
+tar xfz OPAM/repo/tarred.tar.gz -C TMPDIR
+diff "-ruaN" "tarred" "tarred.new" (CWD=${OPAMTMP})
+tar cfz OPAM/repo/tarred.tar.gz.tmp -C TMPDIR tarred
+### opam install foo -vv | grep '^\+' | '.*(/|\\|")test(\.exe)?"? "' -> 'test "' | '.*tar(\.exe)?"? "xfz" ".*(/|\\)OPAM(/|\\)repo(/|\\)tarred\.tar\.gz" "-C" ".*"' -> 'tar xfz OPAM/repo/tarred.tar.gz -C TMPDIR'
+tar xfz OPAM/repo/tarred.tar.gz -C TMPDIR
+test "-f" "bar" (CWD=${BASEDIR}/OPAM/tarring/.opam-switch/build/foo.1)
+### ls $OPAMROOT/repo | grep -v "cache"
+lock
+repos-config
+tarred.tar.gz
+### OPAMREPOSITORYTARRING=0
+### opam update -vv | grep '^\+' | '.*tar(\.exe)?"? "xfz" ".*(/|\\)OPAM(/|\\)repo(/|\\)tarred\.tar\.gz" "-C" ".*"' -> 'tar xfz OPAM/repo/tarred.tar.gz -C TMPDIR' | '.*(/|\\)diff(\.exe)?"? "' -> 'diff "'
+tar xfz OPAM/repo/tarred.tar.gz -C TMPDIR
+diff "-ruaN" "tarred" "tarred.new" (CWD=${OPAMTMP})
+tar xfz OPAM/repo/tarred.tar.gz -C TMPDIR
+### ls $OPAMROOT/repo | grep -v "cache"
+lock
+repos-config
+tarred
+### ls $OPAMROOT/repo/tarred
+packages
+repo
+### diff -ru ./REPO $OPAMROOT/repo/tarred
--- a/master_changes.md
+++ b/master_changes.md
@@ -18,6 +18,7 @@ users)
   * Add cli 2.2 handling [#4853 @rjbou]
   * --no-depexts is the default in CLI 2.0 mode [#4908 @dra27]
   * [BUG] Fix behaviour on closed stdout/stderr [#4901 @altgr - fix #4216]
+  * Add `OPAMREPOSITORYTARRING` environment variable to enable repository tarring optimisation, it is disabled by default because it is an optimisation only on some os/configurations [#5015 @rjbou]
 
 ## Plugins
   *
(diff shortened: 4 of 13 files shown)


kit-ty-kate:
I just tried https://github.com/ocaml/opam/issues/4586#issuecomment-1018730454 on this PR and I get the exact same behaviour: `/usr/bin/tar xfz /home/opam/.opam/repo/default.tar.gz -C /tmp/opam-5953-44547` is still called somewhere else

kit-ty-kate:
Actually, the tar call appears a bit after the first archives retrivals if that helps.

kit-ty-kate (changes requested):
As mentioned in https://github.com/ocaml/opam/issues/4586#issuecomment-1018897393 I think https://github.com/ocaml/opam/pull/3752 should be scraped entirely instead. An opt-in variable doesn’t make much sense in my opinion.

dra27:
From dev meeting: not arrived a name yet, but for 2.2 let's consider flipping this around the other way and disabling the optimisation by default and having the variable to opt in. Some benchmarking comparing the two on NVMe/SSD (i.e. "modern") systems before committing this would be handy.

rjbou:
Updated!

src/client/opamArg.ml
```diff
@@ -209,6 +209,10 @@ let environment_variables =
        'download-command' value from the main config file.";
       "NOCHECKSUMS", cli_original, (fun v -> NOCHECKSUMS (env_bool v)),
       "enables option --no-checksums when available.";
+      "REPOSITORYTARRING", cli_from cli2_2,
+      (fun b -> REPOSITORYTARRING (env_bool b)),
+      "if set, disable repository tarring in temporary directory \
+       optimisation.";
```
kit-ty-kate, line 215:
```suggestion
      "if set, enable repository tarring in temporary directory \
       optimisation.";
```

tests/reftests/init.test
```diff
@@ -1,6 +1,6 @@
 009e00fa
 ### OPAMYES=1
-### tar xzf ${OPAMROOT}/repo/default.tar.gz
+### cp -R ${OPAMROOT}/repo/default .
```
rjbou, line 3: Should we untie this test from opam repo?  @AltGr

kit-ty-kate (changes requested):
So fast!!! ❤️ 

I did some timings and here is what i got locally on my macOS system:
For `opam update default` (which is a local git repository):
* `master` (5311f9b9e) takes 17 seconds
* This PR takes 14 secondes
* This PR (+ some manual tasks, see below) takes 6 seconds!!

In the third case I had to do some manual clean up tasks to get it to properly work and I think this is a bug in the PR:
```
$ ls ~/.opam/repo
alpha.tar.gz
default.tar.gz
lock
repos-config
state-205E9585.cache
$ opam update default
$ ls ~/.opam/repo
alpha.tar.gz
default.tar.gz
lock
repos-config
state-205E9585.cache
```
Here we can see that with this PR, opam does not remove the `.tar.gz` and simply use them as before (with some consistant improvements in time somehow). However as soon as I removed the `.tar.gz` files there, I got a really big x2 speedup and no `tar.gz` were reintroduced.

There must be something missing in this PR in its current state for opam to clean up the tar.gz if they already exist.

tests/reftests/repository.test
```diff
@@ -0,0 +1,38 @@
...
+repos-config
+tarred.tar.gz
+### opam repository rem default --all
+### opam update -vv | grep '^\+'
++ /usr/bin/tar "xfz" "${BASEDIR}/OPAM/repo/tarred.tar.gz" "-C" "${OPAMTMP}"
++ /usr/bin/diff "-ruaN" "tarred" "tarred.new" (CWD=${OPAMTMP})
++ /usr/bin/tar "cfz" "${BASEDIR}/OPAM/repo/tarred.tar.gz.tmp" "-C" "${OPAMTMP}" "tarred"
+### opam install foo -vv | grep '^\+'
++ /usr/bin/tar "xfz" "${BASEDIR}/OPAM/repo/tarred.tar.gz" "-C" "${OPAMTMP}"
++ ${BASEDIR}/OPAM/opam-init/hooks/sandbox.sh "build" "test" "-f" "bar" (CWD=${BASEDIR}/OPAM/optim/.opam-switch/build/foo.1)
```
kit-ty-kate, line 38: We can test for the behaviour mentioned in my review above by adding this test:
```suggestion
+ ${BASEDIR}/OPAM/opam-init/hooks/sandbox.sh "build" "test" "-f" "bar" (CWD=${BASEDIR}/OPAM/optim/.opam-switch/build/foo.1)
### OPAMREPOSITORYTARRING=0
### opam update -vv | grep '^\+'
+ /usr/bin/diff "-ruaN" "default" "default.new" (CWD=${BASEDIR}/OPAM/repo)
### ls $OPAMROOT/repo | grep -v "cache"
default
lock
repos-config
```

kit-ty-kate:
The test also fails on Windows:
```
File "tests/reftests/repository.test", line 1, characters 0-0:
diff --git a/tests/reftests/repository.test b/tests/reftests/repository.out
old mode 100755
new mode 100644
index 1eac20e2..050b1233
--- a/tests/reftests/repository.test
+++ b/tests/reftests/repository.out
@@ -12,7 +12,7 @@ default
 lock
 repos-config
 ### opam install foo -vv | grep '^\+'
-+ ${BASEDIR}/OPAM/opam-init/hooks/sandbox.sh "build" "test" "-f" "bar" (CWD=${BASEDIR}/OPAM/optim/.opam-switch/build/foo.1)
++ /usr/bin/test "-f" "bar" (CWD=${BASEDIR}/OPAM/optim/.opam-switch/build/foo.1)
 ### opam remove foo
   - remove foo 1
@@ -35,4 +35,4 @@ tarred.tar.gz
 + /usr/bin/tar "cfz" "${BASEDIR}/OPAM/repo/tarred.tar.gz.tmp" "-C" "${OPAMTMP}" "tarred"
 ### opam install foo -vv | grep '^\+'
 + /usr/bin/tar "xfz" "${BASEDIR}/OPAM/repo/tarred.tar.gz" "-C" "${OPAMTMP}"
-+ ${BASEDIR}/OPAM/opam-init/hooks/sandbox.sh "build" "test" "-f" "bar" (CWD=${BASEDIR}/OPAM/optim/.opam-switch/build/foo.1)
++ /usr/bin/test "-f" "bar" (CWD=${BASEDIR}/OPAM/optim/.opam-switch/build/foo.1)
         git (internal) (exit 1)
(cd _build/default && /usr/bin/git --no-pager diff --no-index --color=always -u tests/reftests/repository.test tests/reftests/repository.out)
```

kit-ty-kate:
Rebased for CI to pass

src/state/opamRepositoryState.ml
```diff
@@ -171,27 +171,32 @@ let load lock_kind gt =
...
+              OpamFilename.remove tar;
+              OpamConsole.error_and_exit `Aborted
+                "%s.\nRun `opam update --repositories %s` to fix the issue"
+                s (OpamRepositoryName.to_string name))
+          in
+          Hashtbl.add repos_tmp name tmp)
+      repositories
+  else
+    OpamRepositoryName.Map.iter (fun _name repo ->
+        OpamFilename.remove (OpamRepositoryPath.tar gt.root repo.repo_name))
```
kit-ty-kate, line 198: I’m a bit surprised to see the remove being in `lock` instead of `OpamUpdate`.
I would expect only `opam update` to actually do anything.

I haven’t tried locally yet but unless I’ve missed something, I would expect such code to make opam crash for users calling `opam install <pkg>` just after having upgraded from opam 2.1 because they would have to get the content of the repository but it wouldn’t be available anywhere anymore.

But it’s just a hunch from reading the code, I’ll try locally later.

kit-ty-kate: Yep it’s making it fail:
```
opam@feac57c1cc94:~/opam$ opam-2.1 install ocamlfind
  - install ocamlfind 1.9.2

-> retrieved ocamlfind.1.9.2  (http://download.camlcity.org/download/findlib-1.9.2.tar.gz)
-> installed ocamlfind.1.9.2
Done.
opam@feac57c1cc94:~/opam$ ls ~/.opam/repo
beta  default  lock  repos-config  state-0287C1DB.cache
opam@feac57c1cc94:~/opam$ opam-2.1 update
This development version of opam requires an update to the layout of /home/opam/.opam from version 2.0 to version 2.1, which can't be reverted.

Continue? [Y/n] y
Format upgrade done.

[beta] no changes from git+https://github.com/ocaml/ocaml-beta-repository
[default] synchronised from file:///home/opam/opam-repository

<><> Synchronising development packages <><><><><><><><><><><><><><><><><><><><>
opam@feac57c1cc94:~/opam$ ls ~/.opam/repo
beta.tar.gz  default.tar.gz  lock  repos-config  state-0287C1DB.cache
opam@feac57c1cc94:~/opam$ opam-2.1 remove ocamlfind
  - remove ocamlfind 1.9.2

-> removed   ocamlfind.1.9.2
Done.
opam@feac57c1cc94:~/opam$ ls ~/.opam/repo
beta.tar.gz  default.tar.gz  lock  repos-config  state-0287C1DB.cache
opam@feac57c1cc94:~/opam$ ./opam install ocamlfind
[ERROR] No package named ocamlfind found.
opam@feac57c1cc94:~/opam$ ls ~/.opam/repo
lock  repos-config  state-0287C1DB.cache  state-030998F6.cache
```
I believe it’s also making the opam-rt testsuite fail

kit-ty-kate:
I had a shot at fixing it: https://github.com/ocaml/opam/pull/5015/commits/1907fee216f6d0cfdca58d0fbb7436270774e3cc I pushed it in your branch, feel free to remove it if you prefer another solution.
Here is the dockerfile I’ve used to show the new fixed behaviour:
```
FROM ocaml/opam:debian-11-ocaml-4.14
RUN git clone https://github.com/kit-ty-kate/opam -b disable-repo-optim
WORKDIR opam
RUN opam-2.1 update
RUN rm -rf ~/.opam
RUN opam-2.1 init --bare
RUN ls ~/.opam/repo
RUN ./opam switch create test --empty
RUN ./opam install conf-pkg-config
RUN ls ~/.opam/repo
RUN ./opam update
RUN ls ~/.opam/repo
```
as well as its corresonding steps 11 and 13:
```
Step 11/13 : RUN ls ~/.opam/repo
default.tar.gz
lock
repos-config
state-0287C1DB.cache
state-0F87227E.cache
Step 12/13 : RUN ./opam update

[default] no changes from https://opam.ocaml.org
Step 13/13 : RUN ls ~/.opam/repo
default
lock
repos-config
state-0F87227E.cache
```

src/state/opamRepositoryState.ml
rjbou: Indeed, It is the wrong place

src/client/opamArg.ml
```diff
@@ -209,6 +209,10 @@ let environment_variables =
        'download-command' value from the main config file.";
       "NOCHECKSUMS", cli_original, (fun v -> NOCHECKSUMS (env_bool v)),
       "enables option --no-checksums when available.";
+      "REPOSITORYTARRING", cli_from cli2_2,
+      (fun b -> REPOSITORYTARRING (env_bool b)),
+      "if set, enable repository tarring in temporary directory \
+       optimisation.";
```
AltGr, line 215:
```suggestion
      "internally store the repositories as tar.gz files. This can be much \
       faster on filesystems that don't cope well with scanning large trees \
       but have good caching in /tmp.";
```

kit-ty-kate (approved):
To me it looks good to merge as-is (once CI is green of course)

I did some more tests in docker. Using this dockerfile:
```
FROM ocaml/opam:debian-11-ocaml-4.14
RUN git clone https://github.com/rjbou/opam -b disable-repo-optim
WORKDIR opam
RUN opam-2.1 update
RUN sudo apt install -yy time
RUN time opam-2.1 update
RUN ./opam update
RUN time ./opam update
```
* With opam-2.1, `opam update` takes: 1m16s
* With this PR, `opam update` takes: 45s

I’ve also tested on bare metal on a machine with `/tmp` mounted as `tmpfs` (15G) and I’m also getting a x2 performance improvement:
* With opam 2.1, `opam update` takes: 16s
* With this PR, `opam update` takes: 8s

:tada:

src/client/opamArg.ml
kit-ty-kate: maybe adding "but is slower in the general case" at the end would be helpful too i think

dra27:
Fly-by comment: Cygwin's `tar` only deals in Cygwin paths, so even on Windows, these paths are translated to Unix versions, which is even more horrible than the present contortions we have to do with `\` and `/`. The temporary directory appears here _correctly_ in Unix format (i.e. as `/tmp` instead of `C:\cygwin\tmp\`) and is not going to be caught by the normal substitutions easily. Can the test instead do something to get rid of that path completely from the output?

kit-ty-kate:
All green! Thanks a lot @dra27!

src/client/opamArg.ml
AltGr: If you wish! `but have good caching in /tmp` let suppose that if it's not the case you won't get a speedup but not that it'll be a big slowdown indeed :)

rjbou:
Thanks all!

PR #5016, opened 2022-01-21
Add development tools recommendation in opam file (variable) and cli

### LINKED ISSUE #4959: Allow to define dependencies as "development only", regardless how the package is built
At the moment, there is the option of using `{dev}` variable to define "development dependencies" in a package X. However, this doesn't work when the applications or packages that depend on X are not building it from a release archive.

As a more concrete example, as an author of a package, I would like to define `ocamlformat` as a development dependency of my package for those that want to contribute and develop features on it, but not impose `ocamlformat` as a dependency of those users that depend on my package, _regardless how they build it_ (from a release archive or otherwise).

Maybe there could be a new option `--recommended-dev` to support this use case, or a `with-dev` variable, similar to `with-test` and `with-doc`.

rjbou:
~At the moment it's not finished, it needs a proper look at the modification (it's just seddish adds)~
fix #4959
**@rjbou edit:**
All clear! Renamed to `with-tools` : available in cli `opam install pkg --with-tools` and in opam files as a variable `depends: "foo" {with-tools}`. Have the same behaviour than `with-test` and `with-doc`.

### DIFF
--- a/doc/pages/Manual.md
+++ b/doc/pages/Manual.md
@@ -409,9 +409,9 @@ three scopes:
 
 3. Some fields define their own local variables, like `success` in the field
    [`post-messages`](#opamfield-post-messages). Other examples of this include
-   the [`with-test`](#pkgvar-with-test) and [`with-doc`](#pkgvar-with-doc)
-   variables, available in the `depends:`, `depopts:`, `build:`, `install:` and
-   `remove:` fields.
+   the [`with-test`](#pkgvar-with-test), [`with-doc`](#pkgvar-with-doc) and
+   [`with-tools`](#pkgvar-with-tools) variables, available in the `depends:`,
+   `depopts:`, `build:`, `install:` and `remove:` fields.
 
     Within package definition files, the variables `name` and `version`, as
     shortcuts to `_:name` and `_:version`, corresponding to the package being
@@ -520,6 +520,8 @@ Additionally, the following are limited to some package fields (`depends:`,
 - <a id="pkgvar-with-test">`with-test`</a>: only true if tests have been
   enabled for this specific package
 - <a id="pkgvar-with-doc">`with-doc`</a>: similarly for documentation
+- <a id="pkgvar-with-tools">`with-tools`</a>: similarly for developer tools
+
 
 The following are only available in the `depends:` and `depopts:` fields, and
 are used as dependency flags (they don't have a defined `true` or `false` value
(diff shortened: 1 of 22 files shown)


src/client/opamLockCommand.ml
```diff
@@ -252,7 +254,7 @@ let lock_opam ?(only_direct=false) st opam =
   let all_depopts =
     OpamFormula.packages st.packages
       (OpamFilter.filter_deps
-         ~build:true ~test:true ~doc:true ~dev:true ~default:true ~post:false
+         ~build:true ~test:true ~doc:true ~wdev:true ~default:true ~post:false
```
kit-ty-kate, line 257: why did `opam lock` had `dev` to `true` in the first place? This feels a bit weird

doc/pages/Manual.md
```diff
@@ -412,6 +412,7 @@ three scopes:
    the [`with-test`](#pkgvar-with-test) and [`with-doc`](#pkgvar-with-doc)
    variables, available in the `depends:`, `depopts:`, `build:`, `install:` and
    `remove:` fields.
+   // TODO add here, do we keep it for packages filter or all ?
```
kit-ty-kate, line 415: what do you mean by that?

kit-ty-kate (commented):
Looks good! Thanks!! ❤️

rjbou:
No, it's not yet good, i set it in a commit, but not in the PR message itself, it's definitively a wip, i just made a sed  ftm.

src/client/opamLockCommand.ml
rjbou, line 257:
```suggestion
         ~build:true ~test:true ~doc:true ~wdev:true ~dev:true ~default:true ~post:false
```

rjbou: i may say to have all dependencies, and filter afterwards?

doc/pages/Manual.md
rjbou: do we want to keep that variable only for `depends:` & co, or we allow it also for command fields like `install`.
It is meant to be a developer setup variable... setup is just packages or commands too ?

AltGr:
How does this interact with the previous behaviour ? E.g. are `dev` deps still turned on automatically when the package doesn't have a fixed checksum ? If so do we have a `--without-dev` to counter it ? The manpage should be clearer about this.

rjbou:
It doesn't replace `dev`, and ye i agree the doc need to be quite clear about this new option/variable. `with-dev` is meant for development environment, and do not interact at all with `dev`/`--dev` variables/option. see referenced issue for more complete  view of the needs.

dra27:
From discussions last December, we proposed `--with-tools`

dbuenzli:
I guess I'm a bit late to the party but I think `--with-tools` is a pretty confusing name. Packages do install tools or not (I have many packages that optionally install command line tools if cmdliner is installed). I can already see people trying to install them using this command line switch.  Maybe  `--with-dev-tools` ? 

(Personally I'm not entirely convinced by the actual need for the initial request, I'm not sure what's wrong with tagging these dependencies with `dev`, it looks like micro management of dependencies)

dbuenzli:
In fact I think `for-dev` would be a much better and less confusing name.

PR #5017, opened 2022-01-21
Move cold compiler on master to 4.13.1

dra27:
- Move the cold compiler to 4.13 in order to have a version which doesn't suffer from the `SIGSTKSZ` problem with glibc 2.34
- Bump the src_ext dependencies to latest versions while here
- In 4.13, the FlexDLL sources needed to be available _before_ `configure` is run, so move the extraction further up
- The Cygwin32 build is _really_ struggling to build a compiler and run the testsuite at once - the final commit tweaks the build processes so that instead of running `make world` followed by `make opt.opt` we do the fractionally faster `make world.opt` (if the secondary compiler were being built, it does `make world` followed by `make opt`).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -254,10 +254,10 @@ endif
 
 .PHONY: compiler cold
 compiler:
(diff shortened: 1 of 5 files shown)


dra27:
This one's passed on AppVeyor on [my fork](https://ci.appveyor.com/project/dra27/opam/builds/42294406)

dra27:
Merging now to unblock the [base image builder](https://images.ci.ocaml.org)!

PR #5022, opened 2022-01-22
Add some unix-only cram tests to test for #4216 before fixing it

kit-ty-kate:
Will help to merge #4901

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -221,6 +221,8 @@ users)
   * Make all the tests work on macOS/arm64 [#5019 @kit-ty-kate]
 
   * Add clean test for untracked option [#4915 @rjbou]
+## Cram tests
+  * Add some unix-only cram tests to test for https://github.com/ocaml/opam/issues/4216 before fixing it [#5022 @kit-ty-kate]
 
 ## Github Actions
   * Add solver backends compile test [#4723 @rjbou] [2.1.0~rc2 #4720]
--- a/master_changes.md
+++ b/master_changes.md
@@ -135,6 +135,7 @@ users)
   * Update bootstrap ocaml to 4.12.1 to integrate mingw fix [#4927 @rjbou]
   * Update bootstrap to use `-j` for Unix (Windows already does) [#4988 @dra27]
   * Update cold compiler to 4.13 [#5017 @dra27]
+  * Change minimum required Dune to 2.7 [#5022 @kit-ty-kate]
 
 ## Infrastructure
   * Fix caching of Cygwin compiler on AppVeyor [#4988 @dra27]
--- a/master_changes.md
+++ b/master_changes.md
@@ -223,6 +223,7 @@ users)
 
   * Add clean test for untracked option [#4915 @rjbou]
 ## Cram tests
+### Cram tests
   * Add some unix-only cram tests to test for https://github.com/ocaml/opam/issues/4216 before fixing it [#5022 @kit-ty-kate]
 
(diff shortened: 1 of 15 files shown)


rjbou:
is there a reason to not use reftest engine (`run.ml`) ?

kit-ty-kate:
I needed support for actual POSIX shell syntax (which `run.ml` does not allow) and the bultin dune cram tests are much simpler to use too. Since we don’t need windows support for those specific tests, it’s much better to simply use the bultin cram tests.

master_changes.md
```diff
@@ -212,6 +213,8 @@ users)
   * Escape regexps characters in string replacements primitives [#5009 @kit-ty-kate]
   * Automatically update default repo when adding a package file [#5004 @AltGr]
 
+## Cram tests
+  * Add some unix-only cram tests to test for https://github.com/ocaml/opam/issues/4216 before fixing it [#5022 @kit-ty-kate]
```
rjbou, line 217: Only if a re run / update pr is done

```suggestion
### Cram tests
  * Add some unix-only cram tests to test for https://github.com/ocaml/opam/issues/4216 before fixing it [#5022 @kit-ty-kate]
```

tests/cram/unix/list.t
```diff
@@ -0,0 +1,8 @@
+Test https://github.com/ocaml/opam/issues/4216
+  $ export OPAMROOT=./opamroot
+  $ opam init --bare -n git+https://github.com/ocaml/opam-repository.git#0ec9bbf73a91b8e90b15bb11735859c9552a9888 > /dev/null
+  $ OCAMLRUNPARAM=b opam list --depends-on dune -s --all-versions | head -n1
+  0install.2.14
+  Fatal error: exception Sys_error("Broken pipe")
+  Raised by primitive operation at OpamCliMain.main_catch_all in file "src/client/opamCliMain.ml", line 357, characters 4-16
+  Called from Dune__exe__OpamMain in file "src/client/opamMain.ml", line 12, characters 9-28
```
rjbou, line 8: locally test passes with these modifications

```suggestion
  $ OPAMSWITCH= OCAMLRUNPARAM=b opam list --depends-on dune -s --all-versions | head -n1
  0install.2.14
  Fatal error: exception Sys_error("Broken pipe")
  Raised by primitive operation at Stdlib in file "stdlib.ml" (inlined), line 337, characters 0-54
  Called from OpamCliMain.main_catch_all in file "src/client/opamCliMain.ml", line 357, characters 4-16
  Called from OpamCliMain.main in file "src/client/opamCliMain.ml" (inlined), line 476, characters 2-8
  Called from Dune__exe__OpamMain in file "src/client/opamMain.ml", line 12, characters 9-28
```

AltGr:
We're still not rid of opam-rt. **please** don't multiply the test frameworks unless there really is no other way. Think of the added maintainance cost: one more place to look every time for testing, different rules to run or upgrade the tests, etc.

kit-ty-kate:
If we really want to demultiply the test frameworks we can do it at a later date. I’m responsible for it, I’m not going anywhere. Right now, this is moving us forward.

AltGr:
I for one prefer to keep things clean in the first place rather than let others clean up after me; this could have least have been discussed.
Is https://github.com/ocaml/opam/pull/5031/commits/cd4815e3a3630ee7d73ce64c9003efc7b7d5c8e7 so overly complex to you ?

Issue #4135, opened 2020-04-08
opam list --installable doesn't take the currently installed packages into account


kit-ty-kate:
If one of the package currently installed is not available anymore (I suspect mainly the compiler packages), then `opam list --installable` does not work anymore.

Here is a test-case using docker:
```Dockerfile
FROM ocurrent/opam:debian-10-opam
RUN opam init --disable-sandboxing -ay --bare
RUN opam repository set-url --set-default default git://github.com/ocaml/opam-repository.git
RUN opam switch create -y 4.09.1
RUN opam list -s --all-versions --installable --depopts --depends-on ocp-index.1.2
# That worked..

# Now let's set the repository to the commit just before ocaml 4.09.1 gets added
RUN opam repository set-url --set-default default git://github.com/ocaml/opam-repository.git#a322837f3918a7b1a220a3c47c1ca3353d1d6a01
RUN opam list -s --all-versions --installable --depopts --depends-on ocp-index.1.2
# That one doesn't work (empty output) :(
```
The first call to `opam list --installable` returns:
```
ocaml-top.1.1.5
ocp-browser.1.0.0
ocp-browser.1.2
ocp-index-top.0.4.1
ocp-index-top.0.4.3
ocp-index-top.0.5.0
user-setup.0.1
user-setup.0.2
user-setup.0.3
user-setup.0.4
user-setup.0.5
user-setup.0.6
user-setup.0.7
```
but the second does not return anything, and even `opam list -s --all-versions --installable`, if we would try that, does not return anything at all.

Would it be possible to add the currently installed packages to the global state to be consistent with `opam install`?

rjbou:
Then `opam list --installable` would contains all installed packages... Seems logic, from switch installability point of view.

AltGr:
There is special handling internally for what we call "orphan packages": packages which are installed but no longer available ; the policy is to try to keep them as long as they can be kept unchanged, but they can't be recompiled, for example.

The code is in `OpamClient` (and quite complex, involving computation of full dependency and reverse-dependency cones); it would make sense to try to do this processing too for installability computation, but that might be pretty tricky. It would not be reasonable to attempt to duplicate it, in any case.

AltGr:
The above no longer applies; however, the issue remains. It makes sense from a logics point of view: if your current state is already invalid, whatever the package you add it is still invalid, _ergo_ the package is considered not installable...

[...]

PR #5024, opened 2022-01-25
Add pre-processing to coinstallability checks

AltGr:
Addresses some remaining costly cases in #4311

The patch includes a small reorganisation of `OpamSolver`, but the general
idea is to fix the performance regression compared to 2.0:

- with the introduction of solver invariants, the pre-processing that trimmed
 packages conflicting with the base in `OpamState` was removed
- it was replaced by something much more general (and reliable) at the
 `OpamCudf` level
- but only for calls to the external solver, until now

NOTE: this enforces the invariant even for `opam install
--coinstallable-with`, which is consistent with 2.0 but had changed in 2.1.
Without it we can't really expect reasonable performance in general anyway.

### DIFF
--- a/tests/reftests/list.test
+++ b/tests/reftests/list.test
@@ -231,9 +231,47 @@ rankers.2.0.1
 rankers.2.0.7
 ### OPAMVAR_arch=x86_64
 ### opam list --depends-on lacaml -s --all-versions --coinstallable-with phylogenetics.x
-Fatal error:
-Not_found
-# Return code 99 #
+gpr.1.2.1
+gpr.1.3.0
+gpr.1.3.1
+gpr.1.4.0
+gpr.1.4.1
+gpr.1.5.0
+libsvm.0.8.3
+libsvm.0.9.0
+libsvm.0.9.1
+libsvm.0.9.2
+libsvm.0.9.3
+libsvm.0.9.4
+libsvm.0.10.0
+oml.0.0.1
+oml.0.0.2
+oml.0.0.3
+oml.0.0.5
+oml.0.0.6
+owl.0.2.0
+owl.0.2.1
+phylogenetics.0.0.0
+pilat.1.1
+pilat.1.2
+pilat.1.3
+slap.0.0.0
+slap.0.1.0
+slap.0.2.0
+slap.0.2.1
+slap.0.2.2
+slap.0.2.3
+slap.1.0.0
+slap.1.0.1
+slap.2.0.0
+slap.2.0.1
+slap.2.0.2
+slap.3.0.0
+slap.3.0.1
+slap.4.0.0
+slap.4.0.1
+slap.4.1.0
+slap.4.1.1
 ### opam list --depends-on lacaml -s --all-versions --installable gpr phylogenetics
 gpr.1.2.1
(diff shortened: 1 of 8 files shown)


AltGr:
Test case:
```
opam list -s --depends-on minios-xen.0.7 --coinstallable-with minios-xen.0.7 --rec
```
from 5+ minutes on master to 12s here

AltGr:
You can start reviewing, but hold on a little more I found a bug. Shouldn't be anything big.

AltGr:
Fixed!

AltGr:
Squeezed in a one-line fix for #4135

tests/reftests/list.test
```diff
@@ -230,6 +230,47 @@ rankers.1.0.0
 rankers.2.0.1
 rankers.2.0.7
 ### opam list --depends-on lacaml -s --all-versions --coinstallable-with phylogenetics.x
```
AltGr, line 232: coinstallability with a non-existent version used to return nothing. Now it returns what would be compatible with that version, which seems more useful

AltGr:
> Squeezed in a one-line fix for #4135

Actually dropped that, it wasn't needed ; I will comment further on #4135

See the actual diff on `opam list` outputs [here](https://github.com/ocaml/opam/pull/5024/files/97f3607c9a993f1844132434775db85dc5008f32..2b356231405b9adbe9d9844b19a8c89acaf19519#diff-9aaf5ec8b2ac5631c72b620e445583522adf2f17a1a0977308225cf8bd3bd0f2)

gasche:
@Armael and myself hit performances issues in `OpamSolver.installable` when re-running Marracheck after rebasing on the opam master (last time was a while ago, we were working with something 2.0-ish). We call it to get the set of installable packages in a switch with just a compiler installed. The call would previously take seconds (we don't know how much exactly, but it was fast enough that we didn't even bother to log it) and it now takes around seven minutes on my machine.

We warmly welcome progress on this :-)

Armael:
FWIW, in marracheck we run `OpamSolver.installable` on a universe corresponding to a freshly installed switch (containing only the base compiler).
On opam/master this takes 14 minutes (!) on my machine; with this PR it does not seem to be much faster and takes 13 minutes.

kit-ty-kate:
@Armael what is your switch invariant in your use-case?

Armael:
I have `invariant: [ "ocaml-base-compiler" {= "4.13.1"} ]` (in `.opam/theswitch/.opam-switch/switch-config`)

rjbou:
note from dev meeting: installable is not directly impacted by this PR (coinstallable is) . Better merge and see how it's going for this one.

AltGr:
Indeed, this reimplements `OpamSolver.installable_subset` but doesn't change `OpamSolver.installable`; I assume the latter could be implemented from the former, but I didn't change it right away to go incrementally (they are bound to different Dose functions underneath, but these should be semantically equivalent).

So it's normal that you don't get a speedup yet; as a benchmark you could try `installable_subset` with all packages as argument, where the switch invariant will be used for trimming and you should benefit from the optimisation already.

kit-ty-kate (approved):

PR #5027, opened 2022-01-25
some `opam var` and `opam option` fixes

### LINKED ISSUE #5025: opam var XXX fails when no switch is selected
When no switch is selected, for example because I don't have a global switch and I'm not in a folder with a local switch, the command `opam var` and all the sub commands are returning an error code 50.

```
~$ opam init --bare
~$ cd ~/project-with-local-switch
~/project-with-local-switch$ opam sw create . 4.12.1 --no-install
~$ cd
~$ opam var; echo $?

arch              x86_64                  # Inferred from system
exe                                       # Suffix needed for executable filenames (Windows)
jobs              255                     # The number of parallel jobs set up in opam configuration
make              make                    # The 'make' command to use
opam-version      2.1.2                   # The currently running opam version
os                linux                   # Inferred from system
os-distribution   debian                  # Inferred from system
os-family         debian                  # Inferred from system
os-version        10                      # Inferred from system
root              /home/louis.roche/.opam # The current opam root directory
switch                                    # The identifier of the current switch
sys-ocaml-arch                            # Target architecture of the OCaml compiler present on your system
sys-ocaml-cc                              # Host C Compiler type of the OCaml compiler present on your system
sys-ocaml-libc                            # Host C Runtime Library type of the OCaml compiler present on your system
sys-ocaml-version                         # OCaml version present on your system independently of opam, if any

<><> Configuration variables from the current switch ><><><><><><><><><><><><><>
[ERROR] No switch is currently set. Please use 'opam switch' to set or install a switch
50
~$ opam var root; echo $?
[ERROR] No switch is currently set. Please use 'opam switch' to set or install a switch
50
~$ cd ~/project-with-local-switch
~/project-with-local-switch$ opam var root; echo $?
/home/louis.roche/.opam
0
```

That's with opam version 2.1.2. With 2.0.8 the behavior was different from what I could reproduce and `opam config var root` wouldn't return an error.

https://github.com/ocamllabs/vscode-ocaml-platform/issues/701

rjbou:
* Don't error when displaying if switch is not set fix #5025
  * Try to set a variable with option `--switch <sw>` fails instead of writing a wrong `switch-config` file
  * When a field is defined in switch and global scope, try to determine the scope also by checking switch selection
 
the second point need to be backported to 2.1.3


kit-ty-kate (approved):
Looks good!

As a side note, it might be a good idea to remove `OpamStateConfig.get_switch` in the future. It looks like its design brings issues such as this fairly easily (If i understand what the issue was correctly)

rjbou:
Not really, `get_switch` must be used carefully, but it helps to homogeneise the error & its message. The `get_switch_opt` function is present, and can be used. But i agree, we need to take a tour and check its use in opam, and replace needed ones.

AltGr (approved):
Good to merge.
Remark: the `~no_switch` flag is defined but never used.

rjbou:
Certainly a kept old implementation, it is not needed.
Just to keep code somewhere in case one day it is needed:
```ocaml
let with_switch ~display gt lock_kind ?no_switch st k =
  match st with
  | Some st -> k st.switch st.switch_config
  | None ->
    match OpamStateConfig.get_switch_opt () with
    | None ->
      (match no_switch with
       | None ->
         OpamConsole.error_and_exit `Configuration_error "No switch selected"
       | Some no_switch ->
         no_switch (OpamSwitch.of_string "no-switch")
           OpamFile.Switch_config.empty)
```

PR #5030, opened 2022-01-26
Fix make doc

kit-ty-kate:
dune 2.7 changed the default access to the `_build` to be read-only.


avsm:
this should fix the opam.ocaml.org build, thanks @kit-ty-kate! cc @dbuenzli @mseri who reported.

PR #5031, opened 2022-01-26
Stop test framework proliferation

AltGr:

PR #5033, opened 2022-01-28
Revert the addition of the unix-only cram test (#5022)

AltGr:
In favour of a built-in 'reftest' feature implemented in #5031"

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -1,4 +1,2 @@
-(lang dune 2.7)
+(lang dune 2.0)
 (name opam)
-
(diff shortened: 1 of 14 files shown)


AltGr:
As suggested by @dra27, splitted this from #5022 (which should then be rebased).

Note that this isn't a revert of the merge commit directly, because that also includes a much-wanted fix in 44f7e7dc

kit-ty-kate (approved):

rjbou (approved):

PR #5038, opened 2022-02-03
OpamParallel, MakeGraph(_).to_json: fix incorrect use of List.assoc

Armael:
The use of `List.assoc` is incorrect: it relies on the polymorphic equality on an unknown type of vertices (a parameter of the functor), which thus might not be comparable with (=).

Spotted by @art-w , cc @gasche

### DIFF
--- a/src/core/opamParallel.ml
+++ b/src/core/opamParallel.ml
@@ -383,8 +383,12 @@ module MakeGraph (X: VERTEX) = struct
       let vertex_to_json (i, v) = (string_of_int i, X.to_json v) in
       `O (List.map vertex_to_json vertex_map) in
     let edges =
-      let vertex_inv_map = List.map (fun (i, v) -> (v, i)) vertex_map in
-      let index v = List.assoc v vertex_inv_map in
+      let module VertexMap = Map.Make(Vertex) in
+      let vertex_inv_map =
+        List.fold_left (fun m (i, v) -> VertexMap.add v i m)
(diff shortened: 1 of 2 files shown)


src/core/opamParallel.ml
```diff
@@ -383,8 +383,12 @@ module MakeGraph (X: VERTEX) = struct
       let vertex_to_json (i, v) = (string_of_int i, X.to_json v) in
       `O (List.map vertex_to_json vertex_map) in
     let edges =
-      let vertex_inv_map = List.map (fun (i, v) -> (v, i)) vertex_map in
-      let index v = List.assoc v vertex_inv_map in
+      let module VertexMap = Map.Make(Vertex) in
+      let vertex_inv_map =
+        List.fold_left (fun m (i, v) -> VertexMap.add v i m)
```
kit-ty-kate, line 388: @AltGr can `vertex_map` contain duplicated `v` ? If so, I believe the correct fix is to use `List.fold_right` to get the correct map out of it.

Armael: I don't think that's possible: `vertex_map` is defined above as the contents of a set of vertices; this guarantees that there are no duplicates.

kit-ty-kate (approved):
*sigh, something something https://github.com/ocaml/ocaml/pull/9928*



Anyway... ^^" Thanks a lot! :heart: Looks good to me.

kit-ty-kate:
Thanks a lot!

Issue #4690, opened 2021-05-31
Opam `setenv` for path lists is broken on Windows/Cygwin


MisterDA:
Opam doesn't handle well `setenv` in Windows / Cygwin, especially with file lists.
To demonstrate, I'm using the opam images for Docker on Windows. The image ships OCaml for Windows from fdopen, and opam 2.0.8.

Dockerfile

``` Dockerfile
FROM  ocaml/opam:windows-mingw-ocaml-4.13
COPY bugs.cmd .
RUN .\bugs.cmd
```

bugs.cmd

``` cmd
ocaml-env exec -- echo %PKG_CONFIG_PATH%
@rem nothing, ok

ocaml-env exec -- printenv PKG_CONFIG_PATH
@rem nothing, ok

ocaml-env exec -- opam depext -yi conf-pkg-config

ocaml-env exec -- echo %PKG_CONFIG_PATH%
@rem wrong: nothing's exported

ocaml-env exec -- printenv PKG_CONFIG_PATH
@rem good C:/opam/.opam/4.13/lib/pkgconfig

ocaml-env exec -- C:\cygwin64\bin\bash.exe --login -c "printenv PKG_CONFIG_PATH"
@rem C:/opam/.opam/4.13/lib/pkgconfig:C:/opam/.opam/4.13/lib/pkgconfig
@rem wrong: duplicated, uses : as path separator, should use cygwin-style paths

C:\cygwin64\bin\bash.exe --login -c "ocaml-env exec -- printenv PKG_CONFIG_PATH"
@rem C:/opam/.opam/4.13/lib/pkgconfig;C:/opam/.opam/4.13/lib/pkgconfig:
@rem wrong: duplicated, ends with a :, should use cygwin-style paths
```

For reference, the `ENTRYPOINT` of the image is `["ocaml-env" "exec" "--64" "--"]` and its `CMD` is `cmd.exe`.

I have to work around this issue with lines such as

``` sh
ocaml-env exec -- opam source mirage-crypto.0.10.4
cd mirage-crypto.0.10.4
ocaml-env exec -- opam install -y --deps-only --with-test ./mirage-crypto.opam
ocaml-env exec -- env PKG_CONFIG_PATH=/cygdrive/c/opam/.opam/4.13/lib/pkgconfig dune build -p mirage-crypto @install
[...]

PR #5039, opened 2022-02-03
Revert environment variables instead of setting them to an empty string 

rjbou:


src/client/opamConfigCommand.ml
```diff
@@ -103,6 +109,9 @@ let print_sexp_env env =
 let rec print_fish_env env =
   let set_arr_cmd ?(modf=fun x -> x) k v =
     let v = modf @@ OpamStd.String.split v ':' in
+    if v = [] then
+      OpamConsole.msg "set -ge %s;\n" k
+    else
```
rjbou, line 114: need to check for fish if it's ok to unset manpath

rjbou: in fact, it seems that it fixes it

rjbou:
In fact it was broken in windows & fish since a while

dra27:
Related to #4690

src/client/opamConfigCommand.ml
```diff
@@ -103,6 +109,9 @@ let print_sexp_env env =
 let rec print_fish_env env =
   let set_arr_cmd ?(modf=fun x -> x) k v =
     let v = modf @@ OpamStd.String.split v ':' in
```
dra27, line 111: Academic, but this should hopefully fix the tests:
```suggestion
    let v = modf @@ OpamStd.String.split v OpamStd.Sys.path_sep in
```

PR #5040, opened 2022-02-03
[documentation] document the options of OpamSolver.dependencies

gasche:
This is a documentation-only PR: @Armael and myself were looking at `OpamSolver.dependencis`, and we did not know what the labelled argument were referring to. (We are still not fully sure that our documentation is correct.)

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -246,6 +246,7 @@ users)
   * Document custom licenses [#4863 @kit-ty-kate - fix #4862]
   * Add OpenBSD & FreeBSD in the precompiled binaries list [#5001 @mndrix]
   * install.md: fix brew instructions, spelling [#4421 @johnwhitington]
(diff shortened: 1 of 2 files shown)


kit-ty-kate (approved):

AltGr:
> `[post]: include the dependencies of the post-install actions`

it's actually like `build`: "include post dependencies" (ie dependencies flagged `{post}`)

AltGr:
Otherwise all good, I meant to add :sweat_smile:

gasche:
Re "post": actually we started with just "include post dependencies", and we realized we didn't know what those were, so this is our best guess. I will make the proposed change. (In fact "post dependencies" are documented in the manual!)

gasche:
@AltGr I pushed a new version, let me know what you think.

kit-ty-kate:
Thanks!

PR #5049, opened 2022-02-08
Fix opam-core for OCaml 5.00

kit-ty-kate:
The standalone `bigarray` library was removed in OCaml 5.00 and thus makes packages using opam-core with e.g.
```
#=== ERROR while compiling topkg-care.1.0.5 ===================================#
# context              2.2.0~alpha~dev | linux/x86_64 | ocaml-variants.5.00.0+trunk | file:///home/opam/opam-repository
# path                 ~/.opam/5.00/.opam-switch/build/topkg-care.1.0.5
# command              ~/.opam/opam-init/hooks/sandbox.sh build ocaml pkg/pkg.ml build --pkg-name topkg-care --dev-pkg false
# exit-code            1
# env-file             ~/.opam/log/topkg-care-10-c10615.env
# output-file          ~/.opam/log/topkg-care-10-c10615.out
### output ###
# ocamlfind ocamldep -package 'opam-format cmdliner fmt.cli logs.cli bos.setup' -package 'webbrowser webbrowser.cli' -modules src-care/topkg_care.ml > src-care/topkg_care.ml.depends
# + ocamlfind ocamldep -package 'opam-format cmdliner fmt.cli logs.cli bos.setup' -package 'webbrowser webbrowser.cli' -modules src-care/topkg_care.ml > src-care/topkg_care.ml.depends
# ocamlfind: Package `bigarray' not found - required by `opam-core'
# Command exited with code 2.
# pkg.ml: [ERROR] cmd ['ocamlbuild' '-use-ocamlfind' '-classic-display' '-j' '4' '-tag' 'debug'
#      '-build-dir' '_build' 'CHANGES.md' 'LICENSE.md' 'README.md'
#      'topkg-care.opam' 'src-care/topkg_care.a' 'src-care/topkg_care.cmxs'
#      'src-care/topkg_care.cmxa' 'src-care/topkg_care.cma'
#      'src-care/topkg_care_text.cmx' 'src-care/topkg_care_pkg.cmx'
#      'src-care/topkg_care_opam.cmx' 'src-care/topkg_care_ocamlfind.cmx'
#      'src-care/topkg_care_ocamlbuild.cmx' 'src-care/topkg_care_ipc.cmx'
#      'src-care/topkg_care_delegate.cmx' 'src-care/topkg_care_archive.cmx'
#      'src-care/topkg_care.cmx' 'src-care/topkg_care.cmi'
#      'src-care/topkg_care.mli' 'src-bin/topkg_bin.native'
#      'src-bin/toy_github_delegate.native' 'test/unsupportive-delegate'
#      'test/echo-delegate']: exited with 10
```
The solution implemented here is to depend on `bigarray-compat` instead of `bigarray` to keep compatibility with OCaml < 4.07

I’ve also added @dra27’s script to generate the `configure` file and make it actually portable (the version available in CI is heavily patched by ubuntu/debian and thus makes it impossible to get the same output on other platforms)


kit-ty-kate:
I have no idea what’s wrong with the cache of specifically the "Builds, tests & co / test" runs in GHA. @rjbou any idea? A missing key maybe?

dra27:
The key for "opam bootstrap cache" needs to include *.opam in the `hashFiles` bit

kit-ty-kate:
Indeed that fixed it. Thanks a lot! (I used `src_ext/Makefile` and `src_ext/Makefile.source`, similar to the other archive cache to make it easier and a bit more correct).

I’ll squash the last two commits and it should be green tomorrow

dra27:
It’s easier, but it’s not strictly more correct - the opam bootstrap cache does an `opam install . --deps-only` which is independent of the src_ext `Makefile`s… but given that a _new_ dependency should always be being added to both, the worst that does is rebuild a bit more often than needed (e.g. when bumping the src_ext deps)

kit-ty-kate:
Ah, I thought I had read the commands but it might have been different ones, my bad.

.github/workflows/main.yml
```diff
@@ -208,7 +208,7 @@ jobs:
         path: |
           ${{ env.OPAMBSROOT }}/**
           ~/.cache/opam-local/bin/**
-        key: opam-${{ runner.os }}-${{ env.OPAMBSVERSION }}-${{ matrix.ocamlv }}-${{ env.OPAM_REPO_SHA }}-${{ hashFiles ('.github/scripts/main/opam-bs-cache.sh', '.github/scripts/main/preamble.sh') }}
+        key: opam-${{ runner.os }}-${{ env.OPAMBSVERSION }}-${{ matrix.ocamlv }}-${{ env.OPAM_REPO_SHA }}-${{ hashFiles ('.github/scripts/main/opam-bs-cache.sh', '.github/scripts/main/preamble.sh', 'src_ext/Makefile.sources', 'src_ext/Makefile') }}
```
rjbou, line 211:
```suggestion
        key: opam-${{ runner.os }}-${{ env.OPAMBSVERSION }}-${{ matrix.ocamlv }}-${{ env.OPAM_REPO_SHA }}-${{ hashFiles ('.github/scripts/main/opam-bs-cache.sh', '*.opam') }}
```
or something like that

configure.ac
```diff
@@ -326,7 +326,7 @@ echo
 
 # Dependencies
 AC_CHECK_OCAML_PKG([unix])
-AC_CHECK_OCAML_PKG([bigarray])
+AC_CHECK_OCAML_PKG([bigarray-compat])
```
AltGr, line 329: Shouldn't this allow both ?
```suggestion
AC_CHECK_OCAML_PKG([bigarray bigarray-compat])
```

kit-ty-kate: no, bigarray-compat only requires bigarray in OCaml < 4.07

kit-ty-kate: And `bigarray` does not exists at all in OCaml >= 5.00

PR #5053, opened 2022-02-09
Reimplement the depext UI with a menu

### LINKED ISSUE #5026: depext: opam tries to be intelligent but fails; permit to choose the system package manager
How to bypass the failing check of host dependencies in opam?

```
 >which emacs
/usr/bin/emacs
>opam install conf-emacs
  ∗ install conf-emacs 1

    emacs

    apt-get install emacs

You may now install the packages on your system.
When you are done: check again and continue? [Y/n] y

The following remain to be installed: emacs
When you are done: check again and continue? [Y/n] y

The following remain to be installed: emacs
When you are done: check again and continue? [Y/n]
```

AltGr:
Closes #5026
Additionally, also shows the hint on C-c :)

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -22,6 +22,7 @@ users)
   * Refresh the actions list output, now sorted by action/package rather than dependency [#5045 @kit-ty-kate @AltGr - fix #5041]
   * Put back the actions summary as part of confirmation question [#5045 @AltGr]
   * Error report display: print action name [#5045 @AltGr]
+  * Refactored depext-related questions, with a flat menu instead of nested y/n questions [#5053 @AltGr - fix #5026]
 
 ## Plugins
   *
--- a/src/client/opamPinCommand.ml
+++ b/src/client/opamPinCommand.ml
@@ -150,11 +150,10 @@ let copy_files st opam =
 (* Returns the new opam file, without writing it to disk *)
 let edit_raw name temp_file =
   let rec edit () =
-    if OpamStd.Sys.tty_in then
-      (OpamConsole.msg "Press enter to start \"%s\" (this can be customised by \
-                        setting EDITOR or OPAMEDITOR)... "
-         OpamClientConfig.(!r.editor);
-       ignore (read_line ()));
+    OpamConsole.pause
+      "Press enter to start \"%s\" (this can be customised by setting EDITOR \
+       or OPAMEDITOR)..."
+      OpamClientConfig.(!r.editor);
     let edited_ok =
       try
         Sys.command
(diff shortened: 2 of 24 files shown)


AltGr:
Here is what it looks like:
image

AltGr:
Pushing a small style update. Here are a few screenshots to show the interaction:

#### Trying to do it manually, then ignoring:
image

#### With depext-run-installs=false (different menu)
image

The code is a little spaghetti, but there is probably no better way. I have the before/after flow charts (on paper!) if you are interested ^^

dra27:
For the manual menu, I'd suggest:

---

`apt-get install libzmq3-dev`

Would you like opam to:
1. Check again, as the package is now installed
2. Attempt installation anyway, and permanently register that this external dependency is present, but not detectable
3. Abort the installation

AltGr:
> I wonder if numbered options and a full sentence for each works more clearly:

Probably so, although in my display the brackets + table alignment didn't help with that, so we could have fiddled with that as well.
One of the ideas behind the `y/n/i/q` was to have y/n be consistent with existing questions, but I don't know if there is a point in that (keeping the same behaviour for "default" on <enter> being "y"; although `n` of most y/n questions would be closer to "abort" in this case.
The big plus of numbers is that it's rather obvious, and easy to read with much less clutter.

AltGr:
image

dbuenzli:
I find the `>` prompt slightly strange (maybe because that's my shell prompt) and a bit inconsistent with the usual way opam asks for stuff. 

Just saying, I can live with that.

AltGr:
> Just saying, I can live with that.

I already had doubts about it, your input is welcome.

AltGr:
menu
:grin:

AltGr:
I'd be happy to see how it fares on MacOS shells; but I am actually doing the minimum in terms of escape codes magic so it shouldn't be too much trouble. And at worst, there will be a little garbage on top of the menu when moving around, but it will still be readable; or the arrow keys just won't work, which is an ok fallback.

kit-ty-kate:
On macOS the arrows work fine but some thing that throws me a bit off is that it seems that the prompt is waiting for only one character.

For example is you go and input `2` (only one character, no clicking enter) it directly does the action instead of waiting for the user to tap the enter key.

In the same kind of behaviour, if you type anything but one of the character that the prompt is waiting for, nothing happens and your character does not appear. This feels kind of stressful, even claustrophobic a little, it feels like opam just broke and froze your computer (^C still works but still)

kit-ty-kate:
> In the same kind of behaviour, if you type anything but one of the character that the prompt is waiting for, nothing happens and your character does not appear. This feels kind of stressful, even claustrophobic a little, it feels like opam just broke and froze your computer (^C still works but still)

the same freezing behaviour happens on a simple installation prompt. The `y` (default) character is already there but no character appears if you type anything.

I think we should keep the old `[Y/n] <no character>` type prompt as well

AltGr:
Rebased; windows test broken due to last cygwin/git update (unrelated to this PR)

PR #5056, opened 2022-02-11
Fix 'opam init' and 'opam init --reinit' when the `jobs` variable has been set in the opamrc or the current config

rjbou:


master_changes.md
```diff
@@ -25,6 +25,7 @@ users)
 
 ## Init
   * Run the sandbox check in the temporary directory [#4787 @dra27 - fix #4783]
+  * [BUG] Fix config update with init config [#5056 @rjbou]
```
kit-ty-kate, line 28:
```suggestion
  * [BUG] Fix `opam init` and `opam init --reinit` when the `jobs` variable has been set in the opamrc or the current config. [#5056 @rjbou]
```
(if i understand correctly)
The current PR title/changelog entry is confusing and vague, and I had to look at the code to understand what the bug fix was about. It would be nice if the title were to deliver that information directly.

PR #5055, opened 2022-02-10
Minor tweak to GitHub Actions

dra27:
This PR overhauls our CI system to migrate the Windows testing from AppVeyor to GitHub Actions. With the caches primed, the macOS test job appears to be the slow step with the entire run taking 40-45 minutes.

The change-set is very big! Although GitHub has increased the cache limits since I originally started prototyping this last summer, it's still the case that adding the Windows compilers and Cygwin puts tremendous pressure on our cache quotas. This PR necessarily includes an overhaul of non-Windows parts as a result. I've also included - for better or worse (probably the latter) - a rudimentary generator for the YAML workflows. In its defence, it found some bugs. The original version of this PR had macOS, Windows and Linux as separate workflow files, making the case for a generator blindingly obvious. The main thing it's trying to do is ensure that the caches are used consistently (keys, paths, etc.) and also that the jobs which get fanned out over multiple platforms are identical. It does also allow platforms and versions to be added relatively quickly if they're needed for some reason. Critically, it's entirely possible - and, indeed, how I did it - to hack `main.yml`, commit those changes and only _then_ update `ci.ml` to ensure the changes made have been done consistently.

I haven't turned off AppVeyor in this PR yet, as strictly speaking we're not yet doing _exactly_ the same test matrix, but that can follow very soon after.

Headline changes:
- actions/cache contains some critical flaws (all of which are known to GitHub, most of which have been contributed as fixes and all of which have not been merged!). In particular, reliably sharing caches between Unix and Windows requires forcing those caches to be saved using gzip, not zstd. Especially on Windows, but noticeably on Unix as well, it's desirable in the early stages to be able to test for the presence of a cache, but not download it - i.e. where you just want to know if the stage is required. As a result, I've forked actions/cache and added two trivial parameters to it `check-only` and `force-gzip`. It's a couple of commits at https://github.com/dra27/cache/commits/opam and is trivial to rebase when actions/cache is updated (it seems to be updated quite infrequently). I suggest that we move the commits to a fork in ocaml-opam. Again, it _significantly_ reduces run times when the caches are actually primed. Looking at the (cached) run of this PR in [GitHub Actions](https://github.com/dra27/opam/actions/runs/1822506039), you can see the Analyse and Cygwin job taking < 10s where they can take over a minute if they actually extract the files.
[...]

PR #5067, opened 2022-02-24
GHA: add gentoo depext test

rjbou:
& constraint github actions path changes trigger only for files of the given job

### DIFF
--- a/.github/scripts/depexts/generate-actions.sh
+++ b/.github/scripts/depexts/generate-actions.sh
@@ -2,7 +2,7 @@
 
 set -eu
 
-#for target in alpine debian archlinux centos opensuse fedora oraclelinux ubuntu; do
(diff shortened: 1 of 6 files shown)


kit-ty-kate:
Given this PR is not critical and smaller in scope and thus easier to rebase than https://github.com/ocaml/opam/pull/5055 I would suggest merging it after https://github.com/ocaml/opam/pull/5055 instead of the opposit

rjbou:
I don't mind which one first. It is just needed for the next change on depext system (not a common change indeed).
Anyway, they conflict on main & script workflows (adjustment of trigger paths), the first merged need to adjust/port modification into the second.

rjbou:
todo: add mac brew & macport

kit-ty-kate (approved):

Issue #3943, opened 2019-07-27
opam search --depends-on/require-by can quite misleading


dbuenzli:
`opam search` makes it easy to make wrong inferences. For example:
```
> opam search --depends-on bos | grep odoc  
odoc           1.4.1       OCaml documentation generator
```
`odoc` 1.4.1 no longer depends on `bos`, but earlier versions did. 

Also since since I'm moving to at least 4.03 and/or 4.05 I'm trying to get rid of my `result` package dependencies. But I'm often surprised by the unioning behaviour of `--required-by`: 

```
> opam search --required-by topkg | grep result 
result     1.4         Compatibility Result module
> opam search --required-by topkg.1.0.1 | grep result 
> 
```

It would be nice to have an option, say `--latest` that 

1. on `--depends-on PKG` only show packages whose latest version depends on PKG.
2. on `--required-by PKG` only show packages that are required by the 
the latest version of `PKG` (I know `PKG.X.X.X` does that but then I have to remember the latest version number of my packages and I do not, but `opam` does).

jeremiedimino:
I'd like something like this as well. Most of the t ime, I'm only interested in the latest version of packages.

avsm:
To clarify this, could you describe the usecases you're putting this search to?

We need to define "latest version" slightly more precisely, since it also includes package versions that are eliminated by the ocaml compiler version, architecture and os constraints. So in this case, do you want to _eliminate_ reverse dependencies that are not at their latest versions irrespective of installability, or to find the latest set of fresh reverse dependencies that are installable in the current switch.

For a CI revdeps usecase, filtering through the results of the existing command and then testing local installability of the resulting package set is what works. It would be useful to understand what other uses this feature might need.

dbuenzli:
> So in this case, do you want to _eliminate_ reverse dependencies that are not at their latest versions irrespective of installability, or to find the latest set of fresh reverse dependencies that are installable in the current switch.

Both. The `--installable` flag already exists so:
```
--latest               # as per latest version described in the repo
--latest --installable # as per latest version installable in the switch
```

jeremiedimino:
Myself, I'm interested in properties that hold at the tip of the project. I'm assuming that the highest version is the one that is the closest to the tip of the project.

[...]

PR #5075, opened 2022-02-28
Do not list the package itself when running opam list --depends-on/--required-by

### LINKED ISSUE #4446: opam list --recursive --depends-on <pkg> should not include the package itself

kit-ty-kate:
Superseeds https://github.com/ocaml/opam/pull/4804
Fixes #4446

In my opinion `OpamStd.Set.fixpoint` should not return the initial set to start with, but this behaviour seems to be used extensively in the codebase and it seems too much work to entengle every uses.
It would be nice to change this behaviour at some point in the future though, this might bit us again.


master_changes.md
```diff
@@ -56,6 +56,7 @@ users)
   * Improve performance of some opam list combination (e.g. --available --installable) [#4999 @kit-ty-kate]
   * Improve performance of opam list --conflicts-with when combined with other filters [#4999 @kit-ty-kate]
   * Fix coinstallability filter corner case [#5024 @AltGr]
+  * Do not list the package itself when running `opam list --depends-on`/`--required-by` [#???? @kit-ty-kate - fix #4446]
```
kit-ty-kate, line 59:
```suggestion
  * Do not list the package itself when running `opam list --depends-on`/`--required-by` [#5075 @kit-ty-kate - fix #4446]
```

src/core/opamStd.ml
```diff
@@ -255,15 +255,19 @@ module Set = struct
       then failwith (Printf.sprintf "duplicate entry %s" (O.to_string elt))
       else add elt t
 
-    let fixpoint f =
+    let fixpoint ?(including_init=true) f curset =
```
kit-ty-kate, line 258:
```suggestion
    (* TODO: Make ~including_init:false the default *)
    let fixpoint ?(including_init=true) f curset =
```

dbuenzli:
Maybe worth looking at #3943 along the way.

kit-ty-kate:
> Maybe worth looking at #3943 along the way.

Mmh, this is a very distinct issue as far as I can see.

dbuenzli:
Sure but I just mentioned it since you are in the area.

PR #5079, opened 2022-03-03
Fix lock behaviour on pinning

### LINKED ISSUE #4313: `opam install --locked` silently ignores the --locked option

rjbou:
* Fix pinned status display in action list
* Fix repin of locked pins when no changes is detected
fix #4313

### DIFF
--- a/src/client/opamAuxCommands.ml
+++ b/src/client/opamAuxCommands.ml
@@ -347,16 +347,11 @@ let simulate_local_pinnings ?quiet ?(for_view=false) st to_pin =
   in
   let local_packages = OpamPackage.keys local_opams in
   let pinned =
-    if for_view then
-      (* For `opam show`, to display local files instead of the stored on, we
-         need to have on the pinned set only the new simulated pinned ones instead
-         of really pinned ones. *)
-      let open OpamPackage.Set.Op in
-      st.pinned
-      -- OpamPackage.packages_of_names st.pinned
-        (OpamPackage.names_of_packages local_packages)
-      ++ local_packages
-    else st.pinned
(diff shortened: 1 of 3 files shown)


src/client/opamAuxCommands.ml
```diff
@@ -304,11 +309,13 @@ let autopin_aux st ?quiet ?(for_view=false) ?recurse ?subpath atom_or_local_list
           &&
           (* For `opam show`, we need to check does the opam file changed to
              perform a simulated pin if so *)
-          (not for_view ||
+          ((not for_view && not locked) ||
```
AltGr, line 312: I wonder why `not_locked` needs to be checked here, it should already have affected the contents of either opam file (and I would expect the situation to be symmetric w.r.t. the flag ?)

PR #5090, opened 2022-03-06
configure: Use gmake instead of make on Unix systems (fixes BSDs)

kit-ty-kate:
Without this, currently `./configure` fails on *BSD because opam’s makefiles expect GNU Make.

I also took a few commits from https://github.com/ocaml/opam/pull/5049 to make it easy to modify on macOS


rjbou (approved):
lgtm! We need to port these changes to #5055

PR #5093, opened 2022-03-09
Restore the shebang in configure

dra27:
Slightly too much of `tools/autogen` copied over from ocaml/ocaml in #5090. OCaml's configure script is prefixed with another script, so the bit which removes the `--runstate` stuff also deletes the shebang, as it's already there.


kit-ty-kate:
Thanks!

PR #5094, opened 2022-03-11
Patch mccs for Cygwin32 in lib-ext

dra27:
Split off from #5055 so that CI there should pass.


dra27:
Just for the proverbial record, the Cygwin32 build in AppVeyor uses Cygwin's `flexlink` which passes this linker flag as a [non-upstreamed patch](https://cygwin.com/git-cygwin-packages/?p=git/cygwin-packages/flexdll.git;a=blob;f=0.34-cygwin.patch;h=1e50a0cfb808a0f89e2757ea250b24d2b4255c3a;hb=HEAD#l46).

The proposed GHA pipeline's building OCaml from source and bootstrapping flexlink, which is what's the causing the failure there.

kit-ty-kate (approved):

kit-ty-kate:
Thanks!

PR #5098, opened 2022-03-15
Add `jsonm` and `uutf` dependencies

### LINKED ISSUE #5085: Add json library
For the moment there is just an internal [json encoder & basic printer(https://github.com/ocaml/opam/blob/master/src/core/opamJson.mli).
Software heritage fallback need to parse json (ftm it is done via regexp), and in order to have a better support of json output (mainly tests), it is better to have a printer. Instead of rewrite another json library, proposal was done to integrate `jsonm`.

rjbou:
They are helpful for json handling: for testing (and updating) json output and for SWHID request parsing (ftm it is done with regexp) #4859

fix #5085


configure
```diff
@@ -1,9 +1,10 @@
 #! /bin/sh
 # Guess values for system-dependent variables and create Makefiles.
-# Generated by GNU Autoconf 2.69 for opam 2.2.0~alpha~dev.
+# Generated by GNU Autoconf 2.71 for opam 2.2.0~alpha~dev.
```
kit-ty-kate, line 3: you need to use autoconf 2.69. E.g. on macOS:
```
./shell/autogen /opt/homebrew/opt/autoconf@2.69/bin/autoconf
```

master_changes.md
```diff
@@ -150,6 +150,7 @@ users)
   * Bring the autogen script from ocaml/ocaml to be compatible with non-ubuntu-patched autoconf [#5090 @kit-ty-kate #5093 @dra27]
   * configure: Use gmake instead of make on Unix systems (fixes BSDs) [#5090 @kit-ty-kate]
   * Patch AltGr/ocaml-mccs#36 in the src_ext build to fix Cygwin32 [#5094 @dra27]
+  * Add `jsonm` (and `uutf`) dependency [#5098 @rjbou]
```
rjbou, line 153:
```suggestion
  * Add `jsonm` (and `uutf`) dependency [#5098 @rjbou - fix #5085]
```

src_ext/dune-uutf-src
```diff
@@ -0,0 +1,6 @@
+(library
+  (name uutf)
+  (public_name uutf)
+  (flags :standard -w -3)
+  (libraries cmdliner)
```
dra27, line 5: This pulls in a dependency on `cmdliner` for `opam-core` which breaks the 4.03+ requirement. It's OK just to drop it - `cmdliner` is used for the `utftrip` program, which isn't being built

opam-core.opam
```diff
@@ -33,5 +33,6 @@ depends: [
   "re" {>= "1.9.0"}
   "dune" {>= "2.0.0"}
   "cppo" {build & >= "1.1.0"}
+  "jsonm"
```
dra27, line 36: Academic, but shouldn't be in a PR where jsonm starts getting used, or has that already happened?

src/core/dune
```diff
@@ -2,7 +2,7 @@
   (name        opam_core)
   (public_name opam-core)
   (synopsis    "OCaml Package Manager core internal stdlib")
-  (libraries   re ocamlgraph unix bigarray
+  (libraries   re ocamlgraph unix bigarray jsonm
```
dra27, line 5: As above

dra27 (commented):
The patch for jsonm can only affect the lib-pkg build, so it would be in `jsonm.pkg`, rather than `jsonm.common`. However, it would seem easier just to add a dummy uchar package in the lib-pkg build - it's just a `META` file?

opam-core.opam
rjbou, line 36:
```suggestion
```

src_ext/dune-uutf-src
rjbou, line 5:
```suggestion
```

dra27 (approved):
LGTM!

kit-ty-kate (approved):

PR #5104, opened 2022-03-23
Silence w70

dra27:


src/core/dune
```diff
@@ -19,6 +19,9 @@
                   ((action (run %{bin:cppo} -D "VERSION %{read-lines:version}" %{input-file})) opamVersion)))
   (wrapped     false))
 
+(rule
+  (copy opamStubsTypes.ml opamStubsTypes.mli))
+
```
kit-ty-kate, line 24:
```suggestion
                  ((action (run %{bin:cppo} -D "VERSION %{read-lines:version}" %{input-file})) opamVersion)))
  (modules_without_implementation opamStubsTypes)
  (wrapped     false))

```

src/solver/dune
```diff
@@ -14,3 +14,6 @@
                (:include ../ocaml-flags-configure.sexp)
                (:include ../ocaml-context-flags.sexp)))
   (wrapped     false))
+
+(rule
+  (copy opamCudfSolverSig.ml opamCudfSolverSig.mli))
```
kit-ty-kate, line 19:
```suggestion
               (:include ../ocaml-context-flags.sexp)))
  (modules_without_implementation opamCudfSolverSig)
  (wrapped     false))
```

src/tools/dune
```diff
@@ -21,6 +21,8 @@
 
 (rule (with-stdout-to opam_admin_topstart.ml (echo "include Opam_admin_top\n\nlet _ = Topmain.main ()")))
 
+(rule (write-file opam_installer.mli ""))
```
kit-ty-kate, line 24: I think this could be an empty file instead. I think the less custom rules, the better

src/core/dune
dra27: I don't think that works - this is a module without an interface?

src/solver/dune
dra27: Similarly

src/tools/dune
dra27: I confess I have the opposite opinion - indeed, Dune 3.0 seems to have that, too (given that it's auto-generating these empty files in some contexts - I think including this one)

src/core/dune
kit-ty-kate: ah right, then we just need to rename the files at the same time `.ml --> .mli`

src/tools/dune
kit-ty-kate: fair enough then

src/core/dune
dra27: No - that's why I didn't... that changes the library and I didn't have the courage of my convictions to do that just to silence a warning 🙂

dra27: (cf. the breakages in opam-file-format prior to 2.1.0!)

rjbou:
Thanks!

Issue #4647, opened 2021-04-27
feature request: do not rebuild packages when only unimportant constraint has been changed


kit-ty-kate:
Currently opam uses `OpamFile.OPAM.effectively_equal` to detect whether or not a package should be rebuilt.
However in most cases, rebuilding will not change anything but the installed opam file in `$OPAMROOT/<switch>/.opam-switch/packages`.

My proposal is that if only `depends`, `depopts`, `conflicts`, `conflict-class`, `available` or `remove` changes then:
* if none of the related dependencies change then we can ignore this change and only change the opam file for internal use.

I'm not sure exactly how to implement this, my guess is that it will cost an extra run of the solver, but maybe there is a simpler way to do this by just changing how opam requests rebuilds.

kit-ty-kate:
`depexts` can also be added to this list, though it is implemented slightly different from the others as I understand it.

dra27:
Having just discussed this some more, there's a chance that this is really simple! At this point when the switch state is loaded:
https://github.com/ocaml/opam/blob/a92a7ed108b8edafb9ff3de0c98accf7c94e3c4a/src/state/opamSwitchState.ml#L366-L373

If we replace `OpamFile.OPAM.effectively_equal` with an alternate version which also ignores the fields above, then what should happen is that the package would no longer be marked for _compulsory_ reinstall. The solver still runs at `opam upgrade` regardless, so if the dependency formula change actually affects the switch then everything happens as normal. At present, because the package is marked for reinstall, we see the "upstream changes" message and opam reinstalls it anyway.

hannesm:
My 2 cents: if the src changes, or some patches have been added/modified/removed, a rebuild should be done.

kit-ty-kate:
> My 2 cents: if the src changes, or some patches have been added/modified/removed, a rebuild should be done.

of course. See the proposal above

dra27:
> My 2 cents: if the src changes, or some patches have been added/modified/removed, a rebuild should be done.

Good grief, yes! The checks already in opam skip rebuilding if, say, you change the maintainer field of the opam package.

[...]

PR #5118, opened 2022-04-14
Rebuild packages only when what is going to be installed can change

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/4647

The tests are nowhere near complete but it’s a start and it shows that https://github.com/ocaml/opam/issues/4647#issuecomment-1098914526 discussed earlier is the only thing needed to get it working amazingly. Thanks a lot!

In this PR I’ve also fixed a bug where adding or removing the `plugin` flag does not reinstall a package, which shouldn’t be the case as currently plugin creating and delation are only done during the installation and removal actions respectively.

Regarding the code, I went the `?(modulo_state=false)` approach instead of duplicating the function because it made thing clearer regarding what is the difference between `effectively_part` and `effectively_part_modulo_state`

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -173,6 +173,8 @@ users)
   * Repository state: stop scanning directory once opam file is found [#4847 @rgrinberg]
   * Fix reverting environment additions to PATH-like variables when several dirs added at once [#4861 @dra27]
   * Actually allow multiple state caches to co-exist [#4934 @dra27 - fix #4554 properly this time]
+  * Don’t rebuild packages when updating dependencies or availablity, unless the current state needs to be changed [#5118 @kit-ty-kate - fix #4647]
+  * Rebuild packages when removing or adding the "plugin" flag [#5118 @kit-ty-kate]
 
 ## Opam file format
   *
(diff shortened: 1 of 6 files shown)


tests/reftests/reinstall.test
```diff
@@ -0,0 +1,87 @@
...
+depends: ["b" {= version}]
+### opam upgrade --show
+Already up-to-date.
+Nothing to do.
+### :::::: Check that a rebuild occurs if the build field changes
+### <pkg:a.1>
+opam-version: "2.0"
+build: ["opam" "var" "switch"]
+depends: ["b" {= version}]
+### opam upgrade -y --verbose | '^\+ .+ "build"' -> '+'
```
kit-ty-kate, line 49:
```suggestion
### opam upgrade -y --verbose | '^\+ .+ "build"' -> '+' | '^\+ .+opam\.exe ' -> '+ "opam" '
```

kit-ty-kate:
All green and ready to go

master_changes.md
```diff
@@ -327,6 +329,7 @@ users)
   * `OpamFile.Config`: order list of installed switches according their last use, update `with_switch` accordingly, and add `previous_switch` [#4910 @AltGr]
   * Change ``Fetch` action to take several packages, in order to handle shared fetching of packages [#4893 @rjbou]
   * `OpamFile.OPAM.to_string_with_preserved_format`: handle substring errors [#4941 @rjbou - fix #4936]
+  * [API] `OpamFile.OPAM.effective_part` and `OpamFile.OPAM.effectively_equal` now take an optional `?modulo_state:bool` parameter, that if `true`, eliminates the fields relying on the state of the switch (depends, available, …). This is `false` by default.
```
rjbou, line 332:
```suggestion
  * `OpamFile.OPAM.effective_part` and `OpamFile.OPAM.effectively_equal` now take an optional `?modulo_state:bool` parameter, that if `true`, eliminates the fields relying on the state of the switch (depends, available, …). This is `false` by default. [#5118 @kit-ty-kate]
```

rjbou (approved):

Issue #5058, opened 2022-02-15
TODO: Have a look at the patches from Debian


kit-ty-kate:
https://salsa.debian.org/ocaml-team/opam/-/tree/master/debian/patches

AltGr:
https://salsa.debian.org/ocaml-team/opam/-/blob/master/debian/patches/0004-Use-HOME-env-variable-instead-of.patch should really not change anything since the functions called (e.g. `OpamFilename.of_string`) detect the `~` and rewrite it using `HOME` in the exact same way the patch does :thinking: 

I am guessing the "fix" might come from a lintian check that frowns upons strings like `~/.cache` in code.

kit-ty-kate:
cc @glondu

glondu:
@mehdid did the patch... I don't have more explanation than what's written in the patch :-/ I will try to remove it in the next upload to see what happens.

mehdid:
IIRC, it had to do with reproducible builds. Using ~ doesn't ensure the path is the same after each build. So I set HOME in https://salsa.debian.org/ocaml-team/opam/-/blob/master/debian/rules and use it in the code to ensure it remains always the same.

kit-ty-kate:
@mehdid but as @AltGr explained this is already exactly what `OpamFilename.Dir.of_string` does. What is then the difference?

kit-ty-kate:
I had a look at the patches again and noticed a new patch ported over from Fedora (https://src.fedoraproject.org/rpms/opam/c/b2a3f3e63c552c7fd7060da41c556f77a907d394?branch=rawhide) @jamesjer what is the reason for this patch?

jamesjer:
When we introduced dune 3.11 into Fedora, opam started failing to build like this:
```
Running[1]: (cd _build/default/src/client && /usr/bin/sh -c 'git describe --exact HEAD || echo [dev]') > _build/default/src/client/git-describe 2> /dev/null
Running[2]: (cd _build/default/src/client && /usr/bin/sh -c 'git rev-parse --quiet --verify HEAD || echo .') > _build/default/src/client/git-sha 2> /dev/null
sed -f process.sed opam.install > processed-opam.install
sed: can't read opam.install: No such file or directory
```
The patch avoids the build failure.  I don't know why this didn't happen with dune 3.10 and earlier.

kit-ty-kate:
I don't believe the patch does the right thing but in any case this failure with dune 3.11 was fixed in 2.1.6 so you should be able to remove this patch

[...]

PR #5129, opened 2022-04-28
Add a 'test' target

kit-ty-kate:
Comment from the original author:
```
This patch is useful only for Debian: It helps dh_auto_test to detect
that there is a testsuite and run it accordingly. dh_auto_test checks
'test' and 'check' targets, but not 'tests'.
```
Partly f.i.x.e.s https://github.com/ocaml/opam/issues/5058


kit-ty-kate:
I’m guessing it must be networked in a non-dockerized sandbox (it wouldn’t work with docker due to bubblewrap)

PR #5152, opened 2022-05-25
`Unix.realpath`: move compatibility layer into `OpamCompat`

rjbou:


src/core/opamCompat.ml
```diff
@@ -44,13 +44,27 @@ end
...
-  let map_file = Bigarray.Genarray.map_file
+  #if OCAML_VERSION < (4, 6, 0)
+    let map_file = Bigarray.Genarray.map_file
+  #endif
+
+  let realpath s =
+    let getchdir s =
+      let p =
+        try Sys.getcwd ()
+        with Sys_error _ -> Filename.get_temp_dir_name ()
```
rjbou, line 61: log dir was used, it is replaced by temp dir

PR #5155, opened 2022-05-27
Console menu: don't print special characters when output is not tty

rjbou:


dra27 (approved):
There's some other things which confuse me about this function, but I expect that'll become clearer when I implement the Windows side of it... and the fix looks correct, regardless!

kit-ty-kate:
Thanks!

PR #5156, opened 2022-05-27
depext menu: don't loop eternally in scripts when automatical install is off

rjbou:
In depext menu for manual install : 
```
Would you like opam to:
> 1. Check again, as the package is now installed
  2. Attempt installation anyway, and permanently register that this external dependency is present, but not detectable
  3. Abort the installation
```
Default is to check again is the package has been installed. So when launched in script (or test), it loops forever.
Added in `OpamConsole.menu` an [noninteractive] optional argument, that is the default if output is not a tty.


dra27 (approved):
LGTM

kit-ty-kate:
Thanks!

PR #5157, opened 2022-05-27
depext: improve check again routine

rjbou:
* Take into account not found packages instead of only available ones
* Specify printed message
* No more print system package manager helper if missing packages are only not found ones.


kit-ty-kate (approved):

rjbou:
as #5257 was reverted in favor to #5295, reconsidering it

AltGr (approved):

PR #5159, opened 2022-05-31
Docs: Update package versions for Ubuntu in distributions list

javiljoen:
I'm not the package maintainer, but I've added an entry for the current LTS version of Ubuntu and removed the obsolete ones.

Please let me know if I should squash these changes into a single commit or something.


rjbou:
Thanks for the update!

rjbou:
Thanks!

PR #5160, opened 2022-06-01
Add `--with-tools` test, resolve for post messages

### LINKED ISSUE #4959: Allow to define dependencies as "development only", regardless how the package is built

rjbou:
fix #4959
follow up of #5016

### DIFF
--- a/tests/reftests/with-tools.test
+++ b/tests/reftests/with-tools.test
@@ -1,0 +1,57 @@
+N0REP0
+### <pkg:a.1>
+opam-version: "2.0"
+depends: [ "b" "tool" {with-tools} ]
+build: [ [ "touch" "one" ] [ "touch" "two" ] {with-tools} ]
+install: [ [ "cp" "one" "%{lib}%" ] [ "cp" "two" "%{lib}%" ] {with-tools} ]
+post-messages:
+[ "Thanks for installing the packages"
+  "and its tool dependencies too" {with-tools} ]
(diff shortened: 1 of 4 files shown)


tests/reftests/with-tools.test
```diff
@@ -0,0 +1,52 @@
...
+  - install   tool 1 [required by a]
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> removed   a.1
+-> installed tool.1
+-> installed a.1
+Done.
+
+<><> a.1 installed successfully <><><><><><><><><><><><><><><><><><><><><><><><>
+=> Thanks for installing the packages
```
rjbou, line 48: No postmessage

rjbou: fixed

kit-ty-kate (approved):

PR #5161, opened 2022-06-04
Require autoconf 2.71 for configure

dra27:
There's now a GHA runner for Ubuntu 22.04 (in beta).

The noise in `ci.ml` and `lib.ml` can be ignored - the `platform` type gets converted to a partionable GADT so that it's possible to have `Specific (Linux, "22.04")` for a specific runner and still have `Linux` everywhere else.

The important change is the minor tweak in `main.yml` to use the correct runner and the removal of some of the `sed` trickery in `shell/autogen`


kit-ty-kate:
It might probably be worth adding this line to the `configure.ac` too: https://github.com/ocaml/opam/pull/4924/files#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810R1

But otherwise LGTM

kit-ty-kate:
Thanks a lot!

PR #5162, opened 2022-06-11
Make OpamConfigCommand.global_allowed_fields fully lazy

LasseBlaauwbroek:
When it is not lazy, it causes other lazy values to be forced, notably
`OpamStd.Env.list`. This is unexpected, and a bit of a problem for people using
Opam as a library because they may want to set additional environment variables
before callin Opam.

Additionally, I'd expect this to be faster.


kit-ty-kate (approved):
Looks good to me but I’ll leave the merge decision to someone with more experience with this specific part of the code

LasseBlaauwbroek:
I have an OCaml program that uses Opam as a library to install a bunch of Coq packages and performs some benchmarks. Some of these Coq packages have Pip dependencies that are not explicitly listed in the Opam package. So I'm installing a separate Python virtualenv for them and adding the virtualenv to `PATH`. (Yes, I know it is a hack :-( )

PR #5164, opened 2022-06-17
chore: Set permissions for GitHub actions

naveensrinivasan:
Restrict the GitHub token permissions only to the required ones; this way, even if the attackers will succeed in compromising your workflow, they won’t be able to do much.

- Included permissions for the action. https://github.com/ossf/scorecard/blob/main/docs/checks.md#token-permissions

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#permissions

https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

[Keeping your GitHub Actions and workflows secure Part 1: Preventing pwn requests](https://securitylab.github.com/research/github-actions-preventing-pwn-requests/)

Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/workflows/changelog_check.yml
+++ b/.github/workflows/changelog_check.yml
@@ -4,6 +4,9 @@ on:
   pull_request:
     branches: master
 
+permissions:
+  contents: read
+
 jobs:
   diff:
     runs-on: ubuntu-latest
--- a/.github/workflows/depexts.yml
+++ b/.github/workflows/depexts.yml
@@ -21,6 +21,8 @@ env:
 
 jobs:
   opam-cache:
+    permissions:
+      contents: none
     runs-on: ubuntu-latest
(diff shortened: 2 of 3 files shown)


dra27:
Thanks for looking at this, @naveensrinivasan! It caused me to review our settings and discover that the default permissions for the workflows were more open than they needed to be, which I've changed.

However, having set them to be read-only, I'm not sure that setting this in the workflow files is necessary - in particular, I'm confused as to why `main.yml` hasn't had any changes and why `deoexts.yml` only sets permissions on one of its jobs.

dra27:
Closing this, as it's definitely incomplete and I'm not sure it's necessary - but thanks for triggering the check of our repo settings!

Issue #5000, opened 2022-01-16
[META issue] Backport release: 2.1.3


kit-ty-kate:
## Fix PRs
* #4882
*  #4999
*  #4934
*  #4998
*  #4978
*  #4971
*  #5017
*  #4982
*  #5030
*  #5056
* #5097
* #5030 
* #5027
* #5117
* #5176
* #5047
* #4879
* #5223
* #5233 

# Reftests updates
* #4965
* #4913
* #4966 
* #4979 
* #4974 
* #5009 
* #5004 
* #5024 (very partial)
* #4963
* #5162
* #5131

AltGr:
woohoo issue #5000 ! :tada:

rjbou:
released! :tada:

PR #5174, opened 2022-07-11
OpamConfigCommand: avoid capturing the Unix.environment at top level

hannesm:
This is a continuation of #4111 #4789, this time in OpamConfigCommand, which
captures (via a call to OpamInitDefault.init_config ()) the Unix environment
at startup. This is crucial if opam-client is used as a library that modifies
the environment before execution of opam functionality.

This PR is for the 2.1 branch but should apply cleanly on the main branch as well. Are there plans to continue 2.1, or will that version be abandoned in favour of 2.2?

### DIFF
--- a/src/client/opamConfigCommand.ml
+++ b/src/client/opamConfigCommand.ml
@@ -602,8 +602,6 @@ let global_allowed_fields, global_allowed_sections =
   let allowed_fields =
     let open OpamStd.Option.Op in
     let open OpamFile in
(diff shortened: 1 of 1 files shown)


hannesm:
NB I `git grep` for OpamInitDefaults.init_config` to ensure nothing else calls it at top level.

rjbou:
This is fixed by #5162 on master. I add it to list of PRs to backport in #5000

PR #5176, opened 2022-07-11
_Correct_ trying to be too clever when upgrading switches from opam 2.0

dra27:
This is an alternate approach from #5002 for fixing #4501.

When upgrading a switch from opam 2.0, opam needs to convert a list of base packages to a switch invariant. The process used for that applies three heuristics for package `foo.version`:
- If there's only one version of `foo` available, then the invariant `"foo"` is used (i.e. with no version)
- If there's more than one version, but `version` is the latest, then the invariant `"foo" {>= "version"}` is used (lower-bound with upgrades)
- Otherwise the precise `"foo" {= "version"}` is used (exactly equivalent)

All this is attempting to reconstruct whether the user originally ran `opam switch create my-switch foo` or `opam switch create my-switch foo.version`. There are two intentions:
1. Automatic _upgrade_ to the latest version of a compiler (guessed from the installation of the latest version of the compiler)
2. Automatic _update_ of the system compiler, which is largely broken in opam 2.0

If we had the chance to turn back the clock, I think intention 1 is a mistake - it's not stable over time (e.g. until 14 June 2022, `ocaml-base-compiler.4.14.0` would have inferred `"ocaml-base-compiler" {>= "4.14.0"}` but now that there's an alpha release of OCaml 5, it would instead infer `"ocaml-base-compiler" {= "4.14.0"}`) and it's very difficult to be sure that's what the user actually meant, anyway. Intention 2, on the other hand, is a big usability improvement and was a headline reason for adding switch invariants in the first place - system compiler upgrades are basically a mess in opam 2.0, but as long as you end up with `"ocaml-system"` as the switch invariant, they work as expected in 2.1+.

The bug seen on the base images in #4501 I think is simply an accident of #3894 when invariants were added. The set of available package versions was _already present_ as part of depext updating, and the trim of non-pinned versions makes sense there. The fix I propose here, therefore, is to fix the original code only: when computing the invariant, don't filter out the other versions of a pinned package.

I've tested this in Docker for both the original bug and a system compiler upgrade.

It's tempting to eliminate the `>=` heuristic completely, but on the basis that we haven't (I think) had any bug reports about it, I think it's probably better to have 2.1 and 2.2 both do the same (slightly odd) thing, than have the 2.0-2.1 upgrade differ from the 2.0-2.2 upgrade. In the future, I think this heuristic would be much better as a hint - e.g. have `opam switch create system ocaml-system.4.14.0` emit a _warning_ that you probably meant `"ocaml-system"` as the invariant, but that's for another day.

### DIFF
--- a/src/core/opamCompat.ml
+++ b/src/core/opamCompat.ml
@@ -130,3 +130,15 @@ module Stdlib = Pervasives
 #else
 module Stdlib = Stdlib
 #endif
+
+module Lazy =
+#if OCAML_VERSION >= (4, 13, 0)
+  Lazy
+#else
+struct
+  include Lazy
+
+  let map f x =
+    lazy (f (force x))
+end
+#endif
--- a/.github/workflows/ci.ml
+++ b/.github/workflows/ci.ml
@@ -517,7 +517,7 @@ let main oc : unit =
     ("CYGWIN_EPOCH", "2");
   ] in
   let keys = [
-    ("archives", "archives-${{ hashFiles('src_ext/Makefile.sources', 'src_ext/Makefile', '.github/scripts/common/preamble.sh', '.github/scripts/main/preamble.sh', '.github/scripts/main/archives-cache.sh') }}-${{ env.OPAM_REPO_SHA }}");
+    ("archives", "archives-1-${{ hashFiles('src_ext/Makefile.sources', 'src_ext/Makefile', '.github/scripts/common/preamble.sh', '.github/scripts/main/preamble.sh', '.github/scripts/main/archives-cache.sh') }}-${{ env.OPAM_REPO_SHA }}");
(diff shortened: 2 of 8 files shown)


kit-ty-kate (approved):

src/core/opamCompat.ml
```diff
@@ -130,3 +130,14 @@ module Stdlib = Pervasives
 #else
 module Stdlib = Stdlib
 #endif
+
+#if OCAML_VERSION >= (4, 13, 0)
+module Lazy = Lazy
+#else
+struct
```
kit-ty-kate, line 137:
```suggestion
module Lazy = struct
```

dra27: Oops - fixed (slightly differently, w.r.t. to the other modules)

rjbou (approved):

PR #5181, opened 2022-07-13
switch import: fix installed packages reinstall when import file changed

rjbou:
The check for reinstall of pinned packages is done at switch loading. Switch import change overlays afterwards, so a second check need to be done to mark changed one to reinstall before launching resolution.


kit-ty-kate (approved):
Apart from the small POSIX-erie, LGTM

kit-ty-kate:
Thanks

PR #5198, opened 2022-07-26
config report: don't fail if no switch is set 

rjbou:


kit-ty-kate:
LGTM. I can see the tests are in the upcoming https://github.com/ocaml/opam/pull/4892. Thanks!

PR #5199, opened 2022-07-26
config report: catch windows system polling failure

rjbou:
To display full config report

### DIFF
--- a/src/state/opamSysPoll.ml
+++ b/src/state/opamSysPoll.ml
@@ -149,7 +149,10 @@ let variables =
   List.map
     (fun (n, v) ->
        OpamVariable.of_string n,
-       lazy (Lazy.force v >>| fun v -> OpamTypes.S v))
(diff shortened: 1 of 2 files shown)


src/state/opamSysPoll.ml
```diff
@@ -149,7 +149,10 @@ let variables =
   List.map
     (fun (n, v) ->
        OpamVariable.of_string n,
-       lazy (Lazy.force v >>| fun v -> OpamTypes.S v))
+       lazy (try Lazy.force v >>| fun v -> OpamTypes.S v
+             (* We need to catch the assert failure on
+                [OpamSys.with_process_in] for windows *)
```
kit-ty-kate, line 154: where does this assert come from?

rjbou: https://github.com/ocaml/opam/blob/35c22705f2d7b7121df0cae7ca3662fb3a3e6874/src/core/opamStd.ml#L768-L770

rjbou:
Wrong fix

PR #5203, opened 2022-07-26
Eliminate the separate constructors for powershell

dra27:
Record whether the shell was detected as powershell or pwsh, although this information isn't actually used at the moment. Decreases the chance of accidentally only matching on one constructor.

@jonahbeckford - I'm not sufficiently familiar with PowerShell: are there situations where we might need to care whether it's pwsh (PowerShell 6.0) or classic PowerShell Desktop/Core (5.x and earlier)? I'm wondering whether we could risk dropping the `--shell=powershell` option completely?

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -254,6 +254,7 @@ users)
   * Make SHA computation faster by using ocaml-sha [#5042 @kit-ty-kate]
   * Make OpamConfigCommand.global_allowed_fields fully lazy [#5162 @LasseBlaauwbroek]
   * Overhaul Windows C stubs and update for Unicode [#5190 @dra27]
+  * Unify constructors for powershell hosts [#5203 @dra27]
 
 ## Internal: Windows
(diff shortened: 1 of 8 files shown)


jonahbeckford:
There is no `pwsh` for the bundled PowerShell 3.0 and 5.0 in Windows; only `powershell` is present.

dra27:
> And even if it did, could just probe the PATH to find `pwsh.exe` or `powershell.exe` (whichever was found first).

I’ve got some follow-up code for the parent process detection where it gets the full path of the parent process instead of just the name (I wanted this to be able to reliably ask PowerShell where profile files should go) so we should be able to have both the shell and the exact path to it. I’m hoping this may also help with disambiguating running under Cygwin’s PATH and being able to avoid WSL’s `bash` in the Windows system32, too

> +1 LTGM

kit-ty-kate:
Thanks!

PR #5205, opened 2022-07-26
'opam env' for directories with spaces

### LINKED ISSUE #5204: `opam env --switch=SWITCHDIR` is not quoted in PowerShell
I'll send PR.

After creating a local Opam switch in a Windows PowerShell 5 directory with a space, like:

```powershell
mkdir "Person 1"
cd "Person 1"
opam dkml init
```

the switch suggestion shows:

```
# Run (& opam env --switch=C:\Person 1) -split '\r?\n' | ForEach-Object { Invoke-Expression $_ } to update the current
shell environment
```

That needs to be:

```powershell
(& opam env --switch="C:\Person 1") -split '\r?\n' | ...
```

jonahbeckford:
- [x] Please update `master_changes.md` file with your changes.

Fixes #5204 .

It was not just a Windows problem, although only macOS of the *nix variants commonly uses spaces in their directory names.

### DIFF
--- a/src/state/opamEnv.ml
+++ b/src/state/opamEnv.ml
@@ -376,17 +376,56 @@ let shell_eval_invocation shell cmd =
   | _ ->
     Printf.sprintf "eval $(%s)" cmd
 
+(** Returns if the file path needs to be quoted by any supported {!shell}.
+
+    This function does not concern itself with how the file path should be
+    quoted.
+
+    This function treats variable expansions ($) and array expansions for
+    PowerShell (@) and history expansions (!) as needing quotes.
+
+    All other characters come from the following references:
+
+    Bash (metacharacter)
+      https://www.gnu.org/software/bash/manual/html_node/Definitions.html
+      SPACE TAB | & ; ( ) < >
+
+    PowerShell
+      https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_special_characters?view=powershell-5.1
+      https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_quoting_rules?view=powershell-5.1
+      SPACE `
+
+    Command Prompt
+      https://ss64.com/nt/syntax-esc.html
+      SPACE TAB & \ < > ^ | % = ( )
+*)
+let filepath_needs_quote path =
+  let f = function
+    | '$' | '@' | '!'
+    | ' ' | '\t' | '|' | '&' | ';' | '(' | ')' | '<' | '>'
+    | '`'
+    | '\\' | '^' | '%' -> true
+    | _ -> false
+  in
+  OpamCompat.String.exists f path
+
 (** Returns "opam env" invocation string together with optional root and switch
     overrides *)
 let opam_env_invocation ?root ?switch ?(set_opamswitch=false) shell =
-  let quoted_arg arg = match shell with
+  let shell_arg argname pathval =
+    let quoted = match shell with
     | SH_win_cmd | SH_win_powershell | SH_pwsh ->
-      Printf.sprintf " --%s=\"%s\"" arg
+      Printf.sprintf " \"--%s=%s\"" argname
     | SH_sh | SH_bash | SH_zsh | SH_csh | SH_fish ->
-      Printf.sprintf " --%s='%s'" arg
+      Printf.sprintf " '--%s=%s'" argname
+    in
+    if filepath_needs_quote pathval then
+      quoted pathval
+    else
+      Printf.sprintf " --%s=%s" argname pathval
   in
-  let root = OpamStd.Option.map_default (quoted_arg "root") "" root in
-  let switch = OpamStd.Option.map_default (quoted_arg "switch") "" switch in
+  let root = OpamStd.Option.map_default (shell_arg "root") "" root in
+  let switch = OpamStd.Option.map_default (shell_arg "switch") "" switch in
   let setswitch = if set_opamswitch then " --set-switch" else "" in
   Printf.sprintf "opam env%s%s%s" root switch setswitch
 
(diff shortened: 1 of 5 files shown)


kit-ty-kate (approved):
LGTM

jonahbeckford:
Err... DOS Command Prompt needs double-quotes. Sending a commit once I've built locally and tested it again.

src/state/opamEnv.ml
```diff
@@ -378,9 +378,13 @@ let shell_eval_invocation shell cmd =
 
 (** Returns "opam env" invocation string together with optional root and switch
     overrides *)
-let opam_env_invocation ?root ?switch ?(set_opamswitch=false) () =
-  let root = OpamStd.Option.map_default (Printf.sprintf " --root=%s") "" root in
-  let switch = OpamStd.Option.map_default (Printf.sprintf " --switch=%s") "" switch in
+let opam_env_invocation ?root ?switch ?(set_opamswitch=false) shell =
+  let quoted_arg arg = match shell with
+    | SH_win_powershell | SH_pwsh -> Printf.sprintf " --%s=\"%s\"" arg
+    | _ -> Printf.sprintf " --%s='%s'" arg
```
kit-ty-kate, line 384: Could you list them explicitly to avoid feature issues?
```suggestion
    | SH_win_powershell | SH_pwsh -> Printf.sprintf " --%s=\"%s\"" arg
    | All | The | Others -> Printf.sprintf " --%s='%s'" arg
```

jonahbeckford: Yes. Will do.

jonahbeckford:
All fixed.

dra27:
Would it be OK - as with the `set` commands etc in `opam env` to do this only when it's necessary? i.e. to display `--switch=foo` rather than `--switch='foo'`).

More importantly, I'd like to double-check the handling if Windows opam is called by another executable for `--switch="foo"` ... on the Unix shell that gets passed as the string `--switch=foo` in argv, but on Windows it's actually two parameters... I've got a feeling that's cmd-specific and I can't remember what cmdliner does if it actually receives the literal string `--switch="foo"` (the solution otherwise is to advice `"--switch=foo"`)

dbuenzli:
> on the Unix shell that gets passed as the string `--switch=foo` in argv,

But that must be your shell removing the quotes. 

> I can't remember what cmdliner does if it actually receives the literal string `--switch="foo"` (the solution otherwise is to advice `"--switch=foo"`)

That depends on your argument converter. But if you use `Arg.string` you get `"foo"` of course. Everything after the `=` is the option's argument.

rjbou:
updated with a test

jonahbeckford:
The latest commit:
1. Places the quotes around the entire argument:
  * `for /f "tokens=*" %i in ('opam env "--switch=C:\Person 1"') do @%i` in Command Prompt
  * `eval $(opam env '--switch=C:\Person 1')` in Bash
  * `(& opam env "--switch=C:\Person 1") -split '\r?\n' | ForEach-Object { Invoke-Expression $_ }` in PowerShell
2. Uses quotes only when quotes are needed

tests/reftests/env.test
```diff
@@ -53,3 +53,23 @@ OPAM_SWITCH_PREFIX=${BASEDIR}/OPAM/conffile
...
+[defaut] Initialised
+### opam switch create "./$SW" nv --root "$RT"
+
+<><> Installing new switch packages <><><><><><><><><><><><><><><><><><><><><><>
+Switch invariant: ["nv"]
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> installed nv.1
+Done.
+# Run eval $(opam env --root='${BASEDIR}/root 2' --switch='${BASEDIR}/switch w spaces') to update the current shell environment
```
kit-ty-kate, line 73:
```suggestion
# Run eval $(opam env '--root=${BASEDIR}/root 2' '--switch=${BASEDIR}/switch w spaces') to update the current shell environment
```

tests/reftests/env.test
```diff
@@ -53,3 +53,23 @@ OPAM_SWITCH_PREFIX=${BASEDIR}/OPAM/conffile
...
+### opam env --sw "./$SW" | grep OPAM_SWITCH_PREFIX | ';' -> ':'
```
kit-ty-kate, line 74:
```suggestion
### opam env --sw "./$SW" | grep OPAM_SWITCH_PREFIX | ';' -> ':' | '\\_opam\'' -> '/_opam\''
```

tests/reftests/env.test
```diff
@@ -53,3 +53,23 @@ OPAM_SWITCH_PREFIX=${BASEDIR}/OPAM/conffile
...
+### opam switch create "./$SW" nv --root "$RT"
+
+<><> Installing new switch packages <><><><><><><><><><><><><><><><><><><><><><>
+Switch invariant: ["nv"]
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> installed nv.1
+Done.
+# Run eval $(opam env '--root=${BASEDIR}/root 2' '--switch=${BASEDIR}/switch w spaces') to update the current shell environment
+### opam env --sw "./$SW" | grep OPAM_SWITCH_PREFIX | ';' -> ':' | '\\_opam\'' -> '/_opam\''
```
kit-ty-kate, line 74:
```suggestion
### opam env --sw "./$SW" | grep OPAM_SWITCH_PREFIX | ';' -> ':' | '\\_opam' -> '/_opam'
```

tests/reftests/env.test
rjbou: This should be handled by test engine, see https://github.com/ocaml/opam/blob/35899ba5c8d3dfa856076383ad5ae4ab64c66f4e/tests/reftests/env.test#L15-L16

rjbou: Anyway, this last test is not required for this PR (no extra escaping done in it)

kit-ty-kate: ah, then I don’t know how to fix this properly

kit-ty-kate:
Thanks a lot!

Issue #4867, opened 2021-10-15
Invariant/compiler messages needs update


rjbou:
On an switch creation or empty switch there is no *package is not compiler message* while in a switch with packages it is:
```
$ opam switch create foo ocaml

<><> Installing new switch packages <><><><><><><><><><><><><><><><><><><><><><>
Switch invariant: ["ocaml"]
^C

[WARNING] Switch foo left partially installed
$ opam switch set-invariant ocaml-base-compiler --sw foo
The switch invariant was set to ocaml-base-compiler
$ opam switch set-invariant ocaml --sw foo
The switch invariant was set to ocaml

$ opam sw set-invariant ocaml --sw bar
[WARNING] Packages ocaml don't have the 'compiler' flag set.
The switch invariant was set to ocaml
```

And not found package error message refers to a compiler package
```
$ opam switch set-invariant ocamll
[ERROR] No compiler matching `ocamll' found, use `opam switch list-available' to see what is available, or use `--packages' to select packages explicitly.
```

rjbou:
Fixed by #5208

PR #5208, opened 2022-07-29
Compiler packages into invariant packages

### LINKED ISSUE #5201: Importing just exported switch workflow broken
From #2211

@dbuenzli wrote:

>Now that compilers became packages this no longer works:
>```
>  * No agreement on the version of ocaml-base-compiler:
>   - (invariant) → ocaml-base-compiler >= 4.14.0
>    - ocaml = 4.12.0 → ocaml-base-compiler < 4.12.1~
>    You can temporarily relax the switch invariant with `--update-invariant'
>No solution found, exiting
>```
>I certainly do not want to relax the switch invariant since this will result in having in 4.12 in my 4.14 switch.

Underlying issue is that `compiler packages` are not well stored with switch invariant setup

related to #4572

### LINKED ISSUE #5202: Switch listing misdisplay
Since 2.1, it's invariant package(s) that is shown on compiler column, regardless if it is a compiler or not.

related #4572

rjbou:
Since invariant is defined and updated, compiler_packages is no more
consistent the way it is. To keep field name, it is still named
`compiler-packages`, but its content is now invariant formula installed
packages and their recursive dependencies. It implies also that switch
state need to be update at each time an install/removal is done, or
invariant changed.

* Remove `universe.u_base` field
* Compute invariant packages instead of compiler packages
* Change PEF `base` field into `invariant-pkg`
* opam list `--invariant` instead of `--base`
  * change column name to invariant
  * display invariant packages instead of compiler ones
* switch: fill empty synopsis with invariant instead of compiler packages

fix #5202 
fix #5201 

* [x] add function doc
* [x] remove commented code

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -83,9 +83,12 @@ users)
   * Add support for `opam switch -` (go to previous non-local switch) [#4910 @kit-ty-kate - fix 4866]
   * On loading, check for executable external files if they are in `PATH`, and warn if not the case [#4932 @rjbou - fix #4923]
   * When inferring a 2.1+ switch invariant from 2.0 base packages, don't filter out pinned packages as that causes very wide invariants for pinned compiler packages [#5176 @dra27 - fix #4501]
-  * Really install invariant formula if not installed in switch [#5188 @rjbou]
+  * When setting invariant, really install invariant formula if not installed in switch [#5188 @rjbou]
+  * When setting invariant, update switch state to compute invariant packages [#5208 @rjbou]
   * On import, check that installed pinned packages changed, reinstall if so [#5181 @rjbou - fix #5173]
   * [BUG] Enforce extra-source to have a checksum when using "opam switch export --freeze" [#5418 @kit-ty-kate]
+  * Update compiler / base packages handling: always updated, the field contains installed packages resolving invariant formula [#5208 @rjbou]
+  * Fill empty switch synopsis with invariant formula instead of compiler package name [#5208 @rjbou]
 
 ## Config
   * Reset the "jobs" config variable when upgrading from opam 2.0 [#5284 @kit-ty-kate]
@@ -115,6 +118,8 @@ users)
   * Improve performance of opam list --conflicts-with when combined with other filters [#4999 @kit-ty-kate]
   * Fix coinstallability filter corner case [#5024 @AltGr]
   * Improve performance for recursive `--required-by` and `depends-on` [#5337 @rjbou]
+  * Colorise as unavailable (magenta) packages that are specified in the invariant formula and that do not verify it (previous was non installed compiler package) [#5208 @rjbou]
+  * ✘ Change `--base` into `--invariant`, column name and the content is invariant formula installed dependencies [#5208 @rjbou]
 
 ## Show
   * Add `depexts` to default printer [#4898 @rjbou]
@@ -270,6 +275,7 @@ users)
   * shell/bootstrap-ocaml.sh: do not fail if curl/wget is missing [#5223 @kit-ty-kate]
   * Add `swhid_core` dependency [#4859 @rjbou]
   * Remove unused vendored dependency: result [#5465 @kit-ty-kate]
+  * `opam-state` depends on `opam-solver` [#5208 @rjbou]
 
 ## Infrastructure
   * Fix caching of Cygwin compiler on AppVeyor [#4988 @dra27]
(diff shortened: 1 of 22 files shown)


opam-state.opam
```diff
@@ -28,6 +28,7 @@ build: [
 depends: [
   "ocaml" {>= "4.08.0"}
   "opam-repository" {= version}
+  "opam-solver" {= version}
```
kit-ty-kate, line 31: That's unfortunate. Is there no way of moving things around to avoid this? I feel like this will cause problem with many plugins that just want opam-state and not all the dependencies and C requirements pulled by opam-solver

rjbou: At current state, we need to compute dependency cone at state stage.

src/state/opamSwitchAction.ml
```diff
@@ -174,18 +174,19 @@ let update_switch_state ?installed ?installed_roots ?reinstall ?pinned st =
   let installed = installed +! st.installed in
   let reinstall0 = Lazy.force st.reinstall in
   let reinstall = (reinstall +! reinstall0) %% installed in
-  let compiler_packages =
-    OpamPackage.Set.filter (OpamFormula.verifies st.switch_invariant) installed
-  in
```
kit-ty-kate, line 179: Can't we just keep this `compiler_packages` as-is with just the roots instead of the whole tree?

tests/reftests/resolve-variables.test
```diff
@@ -14,6 +14,7 @@ The following actions will be performed:
 
 <><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
 -> installed foo.1
+[WARNING] Undefined filter variable build in dependencies of foo.1
```
kit-ty-kate, line 17: Is that expected?

master_changes.md
```diff
@@ -35,6 +35,7 @@ users)
   * ◈ Add `tree` subcommand to display a dependency tree of currently installed packages [#5171 @cannorin - fix #3775]
   * ◈ Add `why` subcommand to examine how the versions of currently installed packages get constrained (alias to `tree --rev-deps`) [#5171 @cannorin - fix #3775]
   * Make the plugin lookup faster when mistyping a subcommand [#5297 @kit-ty-kate]
+  * Use menu for init setup [#5057 @AltGr #5217 @dra27]
```
rjbou, line 38:
```suggestion
```

src/client/opamSwitchCommand.ml
```diff
@@ -262,21 +266,18 @@ let install_compiler
   else if OpamPackage.Set.is_empty has_comp_flag then
     OpamConsole.note
       "Packages %s don't have the 'compiler' flag set (nor any of their \
-       direct dependencies).\n\
+       dependencies).\n\
```
AltGr, line 269: I am not sure this message still makes sense if you allow any depth of dependencies: after all basically all packages will have a depend on a compiler at some depth, so it's unlikely to ever appear. The original intent was to avoid a warning for "aliases over compilers" packages like `ocaml`.

src/state/opamSwitchState.ml
```diff
@@ -1438,3 +1432,29 @@ let reverse_dependencies st ~build ~post =
...
+
+
+(* invariant computation *)
+
+let invariant_root_packages st =
+  OpamPackage.Set.filter (OpamFormula.verifies st.switch_invariant) st.installed
+
+let compute_invariant_packages st =
+  let pkgs = invariant_root_packages st in
+  dependencies ~build:false ~post:false ~depopts:false ~installed:true
```
AltGr, line 1444: Shouldn't this rather be `~build:true ~post:true` ? It seems at least the post dependencies could legitimately be consideded part of the base.

AltGr (approved):

tests/reftests/resolve-variables.test
rjbou: yes, because we recompute and re-resolve at new switch state recording

src/client/opamSwitchCommand.ml
rjbou: Removed it then.

src/state/opamSwitchState.ml
rjbou: true

rjbou:
Removed `OpamSwitchState.compute_compiler_packages`. As we no more check if "invariant contains" a compiler packages, it is no more used.

dra27:
@rjbou - the extra commit I've pushed allows this one to close #4867 as well I think, if you're happy with it?

rjbou:
lgtm!

PR #5210, opened 2022-08-01
Fix some of the unhelpful conflict messages by merging formulas that include each other

kit-ty-kate:
I spent today trying to fix one of the issue mentionned in the first comment of https://github.com/ocaml/opam/issues/4373 reproduced (somewhat) by:
```
RUN git -C ~/opam-repository reset --hard 143dd2a2f59f5befbf3cb90bb2667f911737fbf8
RUN opam update
RUN opam pin ocaml-base-compiler --current
RUN opam pin ocaml-options-vanilla --current
RUN opam pin ocaml --current
RUN opam pin ocaml-config --current
RUN opam install expect.0.0.6
```
which gives:
```
Step 9/9 : RUN opam install expect.0.0.6
 ---> Running in 3845d2fd821b
[ERROR] Package conflict!
No solution found, exiting
  * Missing dependency:
    not available because the package is pinned to version 4.12.1
  * Missing dependency:
    - expect >= 0.0.6 -> batteries -> ocaml < 4.01.0
    not available because the package is pinned to version 4.12.1
  * Missing dependency:
    - expect >= 0.0.6 -> batteries -> ocaml < 4.02.0
    not available because the package is pinned to version 4.12.1
  * Missing dependency:
    - expect >= 0.0.6 -> batteries -> ocaml < 4.03.0
    not available because the package is pinned to version 4.12.1
  * Missing dependency:
    - expect >= 0.0.6 -> batteries -> ocaml < 4.04.0
    not available because the package is pinned to version 4.12.1
  * Missing dependency:
    - expect >= 0.0.6 -> batteries -> ocaml < 4.05.0
    not available because the package is pinned to version 4.12.1
  * Missing dependency:
    - expect >= 0.0.6 -> batteries -> ocaml < 4.06.0
    not available because the package is pinned to version 4.12.1
  * Missing dependency:
    - expect >= 0.0.6 -> batteries -> ocaml < 4.07.0
    not available because the package is pinned to version 4.12.1
  * Missing dependency:
    - expect >= 0.0.6 -> batteries -> ocaml < 4.08.0
    not available because the package is pinned to version 4.12.1
  * Missing dependency:
    - expect >= 0.0.6 -> batteries -> ocaml < 4.10.0
    not available because the package is pinned to version 4.12.1
  * Missing dependency:
    - expect >= 0.0.6 -> batteries -> ocaml < 4.11.0
    not available because the package is pinned to version 4.12.1
  * Missing dependency:
    not available because the package is pinned to version 4.12.1

The command '/bin/sh -c opam install expect.0.0.6' returned a non-zero code: 20
```
for some reason this is a bit different than the comment mentionned in #4373 but I suppose this is caused by the the multiple pins --current. In any case this sort of issue happens all the time on macOS/arm64 as shown by https://github.com/ocaml/opam/issues/4373#issuecomment-885215378

The goal of the change is to merge similar `Missing` cases. I know @AltGr mentioned that the code already filters something but I couldn’t find that part of the code and figured it would be fine to duplicate this, even if just for debugging purpose.

With this change `opam install expect.0.0.6` shows the expected:
```
[ERROR] Package conflict!
  * Missing dependency:
    not available because the package is pinned to version 4.12.1

No solution found, exiting
```


tests/reftests/conflict-4373.test
```diff
@@ -1,10 +0,0 @@
-c1d23f0e
-### OPAMVAR_enable_ocaml_beta_repository=1 opam switch create --fake 4.12.0
-[ERROR] No compiler matching `4.12.0' found, use `opam switch list-available' to see what is available, or use `--packages' to select packages explicitly.
-# Return code 5 #
-### opam install --show expect.0.0.6
-[ERROR] No switch is currently set. Please use 'opam switch' to set or install a switch
-# Return code 50 #
-### opam install --show ppx_expect.v0.14.0
-[ERROR] No switch is currently set. Please use 'opam switch' to set or install a switch
-# Return code 50 #
```
kit-ty-kate, line 10: Note for developers: this test was included in https://github.com/ocaml/opam/pull/4504 for the same purpose but has never tested anything

tests/reftests/unhelpful-conflicts-002.test
```diff
@@ -28,40 +28,7 @@ ocaml-config is now pinned locally (version 1)
...
-    not available because the package is pinned to version 4.12.0
-  * Missing dependency:
-    - expect >= 0.0.6 -> batteries -> ocaml < 4.10.0
-    not available because the package is pinned to version 4.12.0
-  * Missing dependency:
-    - expect >= 0.0.6 -> batteries -> ocaml < 4.11.0
-    not available because the package is pinned to version 4.12.0
-  * Missing dependency:
-    - expect >= 0.0.6 -> batteries -> ocaml < 4.12.0
+    - ocaml < 4.12.0
```
rjbou, line 31: For these conflict message tests, it can be helpful to know what is the "intended output". I'm not sure if it is always simple to know in advance.
Here if we know that it is `ocaml < 4.12.0` the conflict that we want to highlight (i suppose that from the pin above), a comment may help to know that the fix is good (and not a simplification that outputs the wrong message).
This case is simple thought to determine thought.

src/solver/opamCudf.ml
```diff
@@ -1024,6 +1022,7 @@ let extract_explanations packages cudfnv2opam reasons : explanation list =
                  not (List.exists (function `Conflict (_,_,has_invariant) -> has_invariant | _ -> false) explanations)
             in
             let msg = `Conflict (msg1, msg2, msg3) in
+            (* TODO: remove this List.mem *)
```
rjbou, line 1025:
```suggestion
            (* TODO: remove this List.mem, cf [explanations] below *)

rjbou (approved):
Thanks!

kit-ty-kate:
The review highlighted a small issue i would like to fix and understand first so I'm closing this one in favor of https://github.com/ocaml/opam/pull/6106

Issue #4992, opened 2022-01-13
`Required setup` message contains incorrect path to opam root when using `OPAMROOT`


Khady:
If I set the opam root to be non default, with the OPAMROOT env variable, during `opam init` the `required setup` section still points to `~/.opam` instead of the content of OPAMROOT

```
$ export OPAMROOT="/tmp/opam-root"
$ opam init --bare
[ ... ]
<><> Required setup - please read <><><><><><><><><><><><><><><><><><><><><><><>

  In normal operation, opam only alters files within ~/.opam.
[ ... ]
```

opam 2.1.0

AltGr:
Suggestion:
- if $OPAMROOT is not defined:
```
In normal operation, opam only alters files within ~/.opam.
```
as before
- if it is defined:
```
In normal operation, opam only alters files within $OPAMROOT.
```
(verbatim, don't expand the variable)

dra27:
Only problem with that is that it doesn't cover `--root`. It's reasonable to assume that this only happens to a user who knows about roots (because they have set `OPAMROOT` or added `--root`) so it could reasonably be: "opam only alters files within your opam root (~/.opam)" and display the actual value if it's not the default. Referring to the opam root in the message loosely covers the case where you then unset `OPAMROOT` later.

dra27:
(putting in 2.2 as the Windows shell stuff touches this message anyway, and it's not a big change!)

PR #5212, opened 2022-08-02
Create opam root in `%LOCALAPPDATA%\opam` on Windows

dra27:
If `.opam` is found in either `%HOME%` or `%USERPROFILE%` then use that, but roots are otherwise created in `%LOCALAPPDATA%\opam`.

This strategy is forwards-compatible with any switch to XDG (which would definitely introduce the situation where _existing_ roots would be read from locations other than where a _new_ root would be init'd). The location for Windows was discussed in #3766.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -37,6 +37,7 @@ users)
 ## Init
   * Run the sandbox check in the temporary directory [#4787 @dra27 - fix #4783]
   * [BUG] Fix `opam init` and `opam init --reinit` when the `jobs` variable has been set in the opamrc or the current config. [#5056 @rjbou]
+  * Use `.opam` from `%HOME%` or `%USERPROFILE%` on Windows, only if found; otherwise use `%LOCALAPPDATA%\opam` as root. [#5212 @dra27]
 
 ## Config report
   * [BUG] Don't fail is no switch is set [#5198 @rjbou]
--- a/master_changes.md
+++ b/master_changes.md
@@ -38,6 +38,7 @@ users)
   * Run the sandbox check in the temporary directory [#4787 @dra27 - fix #4783]
   * [BUG] Fix `opam init` and `opam init --reinit` when the `jobs` variable has been set in the opamrc or the current config. [#5056 @rjbou]
   * Use `.opam` from `%HOME%` or `%USERPROFILE%` on Windows, only if found; otherwise use `%LOCALAPPDATA%\opam` as root. [#5212 @dra27]
+  * Display actual location of OPAMROOT in `opam init` if `--root` or `OPAMROOT` have been set [#5212 @dra27 - fix #4992]
 
 ## Config report
   * [BUG] Don't fail is no switch is set [#5198 @rjbou]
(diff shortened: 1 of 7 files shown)


master_changes.md
```diff
@@ -36,6 +36,7 @@ users)
 ## Init
   * Run the sandbox check in the temporary directory [#4787 @dra27 - fix #4783]
   * [BUG] Fix `opam init` and `opam init --reinit` when the `jobs` variable has been set in the opamrc or the current config. [#5056 @rjbou]
+  * Use `.opam` from `%HOME%` or `%USERPROFILE%` on Windows, only if found; otherwise use `%LOCALAPPDATA%\opam` as root. [#5212 @dra27]
```
jonahbeckford, line 39: Small nit:

jonahbeckford:
```suggestion
  * Use `.opam` from `%HOME%` or `opam` from `%USERPROFILE%` on Windows, only if found; otherwise use `%LOCALAPPDATA%\opam` as root. [#5212 @dra27]
```

jonahbeckford (changes requested):

jonahbeckford:
This is great; ordinary opam.exe will be able to read OPAMROOT from Diskuv's opam.exe.

master_changes.md
dra27: What I've put is correct inasmuch as it's what it's doing - `%USERPROFILE%` is being used as the fallback if `%HOME%` isn't set - it's only in `%LOCALAPPDATA%` that I'm using `opam` rather than `.opam`?

dra27:
The use of `%HOME%` also makes it compatible with opam-repository-mingw

master_changes.md
jonahbeckford: Oops. I mixed up the two. Cancel the suggestion!

jonahbeckford:
LGTM

dra27:
Looks like a tweak is needed to the runners. `opam init` is also incorrectly displaying `~\.opam` as the install location.

AltGr (approved):

dra27:
Thanks for the review, @AltGr! I've hopefully fixed the CI - beforehand, `OpamStd.Sys.home` was falling to the default of `.opam` in the CWD which was a dreadful default for Windows, but actually correct for the runners (that's fixed by setting `OPAMROOT`)

There's an extra commit to be checked, though. I've altered the code so that you now get the correct display on Windows:

```
  In normal operation, opam only alters files within ~\AppData\Local\opam.
```

but while there it was worth fixing #4992 as well. If the root is _not_ the default, then you get this slightly different message:

```
  In normal operation, opam only alters files within your opam root
    (~\AppData\Local\opam by default; currently C:\Devel\opam-2\root).
```

We could possibly be more explicit about _why_ the root has been overridden (there are three possibilities: `--root`, `OPAMROOT` or a legacy location - i.e. somewhere a root is _read_ from but not written like `%HOME%\.opam` on Windows) but I expect that it's fine as it is.

kit-ty-kate (approved):

rjbou (approved):

kit-ty-kate:
Thanks a lot!

PR #5217, opened 2022-08-03
Fixes to opam init menu

dra27:
Three little bugs, all to do with changing the shell using the new `opam init` menu:

- On bash, the option crashes opam with `Not_found` because the list of shells used is the ones for init scripts, not the complete of shells - I've introduced a list containing the full list of shells for this purpose.
- When selecting a new shell, the default was forced to "Yes". That's a slight issue when the Windows shell comes because "Yes" isn't always an option but it also doesn't really make sense - the default should stay the same
- Having selected a new shell, `opam init` then wrote the `.profile` (or whatever) using the syntax of the old shell!


rjbou (approved):
lgtm!

PR #5218, opened 2022-08-03
GHA: Fix the archlinux script

kit-ty-kate:
archlinux does not upgrade dependencies by default without the 'u' argument and leads to issues where some packages can get installed with several versions of glibc


rjbou (approved):

rjbou:
thanks!

PR #5219, opened 2022-08-03
Update subpath display

rjbou:
Some examples:

```diff
-root-a_p is now pinned to file://${BASEDIR}/pinnes (a) (version dev)
+root-a_p is now subpath-pinned to directory /a in file://${BASEDIR}/pinnes (version dev)

-root-a-i_g is now pinned to git+file://${BASEDIR}/pinnes#master (a/i) (version dev)
+root-a_p is now subpath-pinned to directory /a in file://${BASEDIR}/pinnes (version dev)

-root-a_p.dev    rsync  file://${BASEDIR}/pinnes (a)
-root-b_g.dev    git    git+file://${BASEDIR}/pinnes#master (b)
+root-a_p.dev    rsync  directory /a in file://${BASEDIR}/pinnes
+root-b_g.dev    git    directory /b in git+file://${BASEDIR}/pinnes#master

--> retrieved root-a-i_g.dev  (git+file://${BASEDIR}/pinnes#master(a/i))
+-> retrieved root-a-i_g.dev  (directory /a/i in git+file://${BASEDIR}/pinnes#master)

-[root-a-i_g.l1] synchronised (git+file://${BASEDIR}/pinnes#master(a/i))
-[root-a-k_p.l1] synchronised (file://${BASEDIR}/pinnes(a/k))
+[root-a-i_g.l1] synchronised (directory /a/i in git+file://${BASEDIR}/pinnes#master)
+[root-a-k_p.l1] synchronised (directory /a/k in file://${BASEDIR}/pinnes)

```

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -78,6 +78,7 @@ users)
   * ◈ New option `opam pin --current` to fix a package in its current state (avoiding pending reinstallations or removals from the repository) [#4973 @AltGr - fix #4970]
   * [BUG] Fix using `--working-dir` with non pinned packages: it was not downloading sources as they were remove from package that need sources [#5082 @rjbou - fix #5060]
   * [NEW] Reactivate subpath and recursive pinning `--recursive` and `--subpath` [#4876 @rjbou]
+    * Change display from `git+https://url#hash (subpath)` to `/subpath in git+https://url#hash` [#5219 @rjbou]
   * scan: show subpaths [#4876 @rjbou]
   * [BUG] Fix windows path for subpath, by introducing their own type in `OpamFilename` [#4876 @rjbou]
   * [BUG] Fix repin of locked pins when there is no change in lock file [#5079 @rjbou - fix #4313]
@@ -482,3 +483,5 @@ users)
   * `OpamCompat`: add `Lazy` module and `Lazy.map` function [#5176 @dra27]
   * `OpamConsole.menu`: add `noninteractive` option to choose a different default when output is not a tty [#5156 @rjbou]
   * `OpamStd.Sys`: add `all_shells` list of all supported shells [#5217 @dra27]
+  * `OpamUrl`: add `to_string_w_subpath` to display subpath inside urls (before hash) [#5219 @rjbou]
+  * `OpamFilename.SubPath`: remove `pretty_string` in favor to `OpamUrl.to_string_w_subpath` [#5219 @rjbou]
--- a/src/client/opamAction.ml
+++ b/src/client/opamAction.ml
@@ -619,10 +619,8 @@ let make_command st opam ?dir ?text_command (cmd, args) =
               nv.name
           in
           let rev = OpamProcess.Job.run (OpamRepository.revision src u) in
-          Printf.sprintf "pinned(%s%s%s)"
-            (OpamUrl.to_string u)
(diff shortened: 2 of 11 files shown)


dra27:
Could we even avoid the square brackets in the display?

rjbou:
We need to differentiate the main url and the subpath. We can use another separator.

dra27:
I was thinking we could kinda avoid it, but that would be inconsistent with `opam pin add` (i.e. unless we could make `opam pin add git+https://github.com/dra27/package/subdir` automatically infer the sub-directory, then I agree we need some kind of separator).

The problem is that the notation is opam specific, but perhaps we can therefore make it more explicit that there's a subpath pin involved. Possibility:

```diff
-root-a_p is now pinned to file://${BASEDIR}/pinnes (a) (version dev)
+root-a_p is now subpath-pinned to directory /a in file://${BASEDIR}/pinnes (version dev)

-root-a-i_g is now pinned to git+file://${BASEDIR}/pinnes#master (a/i) (version dev)
+root-a-i_g is now subpath-pinned to directory /a/i in git+file://${BASEDIR}/pinnes#master (version dev)

-root-a_p.dev    rsync  file://${BASEDIR}/pinnes (a)
-root-b_g.dev    git    git+file://${BASEDIR}/pinnes#master (b)
+root-a_p.dev    rsync  /a in file://${BASEDIR}/pinnes
+root-b_g.dev    git    /b in git+file://${BASEDIR}/pinnes#master


--> retrieved root-a-i_g.dev  (git+file://${BASEDIR}/pinnes#master(a/i))
+-> retrieved root-a-i_g.dev  (/a/i in git+file://${BASEDIR}/pinnes#master)

-[root-a-i_g.l1] synchronised (git+file://${BASEDIR}/pinnes#master(a/i))
-[root-a-k_p.l1] synchronised (file://${BASEDIR}/pinnes(a/k))
+[root-a-i_g.l1] synchronised (/a/i in git+file://${BASEDIR}/pinnes#master)
+[root-a-k_p.l1] synchronised (/a/k in file://${BASEDIR}/pinnes)
```
??

rjbou:
I like the idea, but i'm wondering if these two examples are explicit for user. I give it a quick try:
```
[root-a_p.l2] synchronised (/a in file://${BASEDIR}/pinnes)
```
-> do we understand that `a` is a subpath?
```
+Ok, root-b_g is no longer subpath-pinned to /b in git+file://${BASEDIR}/pinnes#master (version l2)
```
-> it is not clear if it is the subpath-pinning that is removed or the pinning

dra27:
Good point - perhaps:

```
[root-a_p.l2] synchronised (directory /a from file://${BASEDIR}/pinnes)
```

When un-pinning, perhaps there's no need for status at all - the fundamental change is that we've un-pinned that repo, so it could be the same message for both sub-path and normal?

rjbou:
> When un-pinning, perhaps there's no need for status at all - the fundamental change is that we've un-pinned that repo, so it could be the same message for both sub-path and normal?

It is possible, it will complicate API as in different function, it's  `OpamPinCommand.string_of_pinned" that is called for printing.

rjbou:
It's a local function :+1:

kit-ty-kate (approved):
I like that `subpath-pinned to directory /<path> in ...` idea. LGTM

Issue #5224, opened 2022-08-05
Solver timeout with opam reinstall but not opam install


kit-ty-kate:
For a single (non-installed) package, `opam reinstall` will lead to solver timeout (8 minutes) whereas `opam install` will work just fine (13 seconds)

Here is a reproducible example:
```
git clone --recursive "https://github.com/ocaml/opam-repository.git" && cd "opam-repository" && git fetch origin "refs/pull/21924/head" && git reset --hard a94864fd
git fetch origin master
git merge 3235916a162a59d7c82dac3fe24214975d48f1aa
cat > ../Dockerfile <<'END-OF-DOCKERFILE'
FROM ocaml/opam:ubuntu-22.04-ocaml-4.14@sha256:e36f7e8cbfe7643ddc7a271dad0cbfb3007891b04da868f2fdb15909fda709bb
USER 1000:1000
WORKDIR /home/opam
RUN sudo ln -f /usr/bin/opam-dev /usr/bin/opam
RUN opam init --reinit -ni
ENV OPAMDOWNLOADJOBS="1"
ENV OPAMERRLOGLEN="0"
ENV OPAMSOLVERTIMEOUT="500"
ENV OPAMPRECISETRACKING="1"
RUN rm -rf opam-repository/
COPY --chown=1000:1000 . opam-repository/
RUN opam repository set-url --strict default opam-repository/
RUN opam pin add -k version -yn tezos-client-005-PsBabyM1.14.0 14.0
RUN opam update --depexts
RUN time opam reinstall --show-action tezos-client-005-PsBabyM1.14.0
RUN time opam install --show-action tezos-client-005-PsBabyM1.14.0
END-OF-DOCKERFILE
docker build -f ../Dockerfile .
```

kit-ty-kate:
Looking at the cudf requests, the only differences were:
```
--- install-1.cudf	2022-08-05 15:28:34.000000000 +0100
+++ reinstall-1.cudf	2022-08-05 15:28:35.000000000 +0100
@@ -1,5 +1,5 @@
 # Solver: builtin-mccs+glpk
-# Criteria: -removed,-count[avoid-version,changed],-count[version-lag,request],-count[version-lag,changed],-count[missing-depexts,changed],-changed
+# Criteria: -changed,-count[avoid-version:,true],-count[version-lag:,false],-count[missing-depexts:,true]
 preamble:
[...]

Issue #5227, opened 2022-08-05
opam reinstall does not show the reason for installing packages


kit-ty-kate:
For example, opam install does:
```
$ opam install --show dose3.6.1
The following actions would be performed:
=== install 21 packages
  - install base-bytes        base   [required by base64]
  - install base64            3.5.0  [required by dose3]
  - install camlbz2           0.7.0  [required by dose3]
  - install camlzip           1.11   [required by dose3]
  - install conf-aclocal      2      [required by camlbz2]
  - install conf-autoconf     0.1    [required by camlbz2]
  - install conf-automake     1      [required by conf-aclocal]
  - install conf-libbz2       1      [required by camlbz2]
  - install conf-pkg-config   2      [required by conf-zlib]
  - install conf-which        1      [required by conf-autoconf]
  - install conf-zlib         1      [required by camlzip]
  - install csexp             1.5.1  [required by dune-configurator]
  - install cudf              0.10   [required by dose3]
  - install dose3             6.1
  - install dune-configurator 3.4.1  [required by parmap]
  - install ocamlfind         1.9.5  [required by camlbz2, camlzip]
  - install ocamlgraph        2.0.0  [required by dose3]
  - install parmap            1.2.4  [required by dose3]
  - install re                1.10.4 [required by dose3]
  - install seq               base   [required by re]
  - install stdlib-shims      0.3.0  [required by dose3]
```
whereas opam reinstall does:
```
$ opam reinstall --show dose3.6.1
dose3.6.1 is not installed. Install it? [y/n] y
The following actions would be performed:
=== install 18 packages
  - install base-bytes      base
  - install base64          3.5.0
  - install camlbz2         0.7.0
  - install camlzip         1.11
  - install conf-aclocal    1.0.0
  - install conf-autoconf   0.1
  - install conf-libbz2     1
  - install conf-pkg-config 2
  - install conf-which      1
  - install conf-zlib       1
  - install cudf            0.10
  - install dose3           6.1
  - install ocamlbuild      0.14.1
  - install ocamlfind       1.9.5
  - install ocamlgraph      2.0.0
  - install parmap          1.0-rc11
  - install re              1.7.1
  - install stdlib-shims    0.3.0
  ```
In the second case, the reasons are missing

PR #5228, opened 2022-08-05
Use the default criteria during reinstall/upgrade when requesting at least one non-installed package

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/5224 as well as discrepancies in the result given by the solver between `opam install <pkg>` and `opam reinstall <pkg>` when `<pkg>` is not installed (see the differences in the solver output in the description of https://github.com/ocaml/opam/issues/5227)

---------
~TODO:~
- [x] Some tests should be added to prevent regressions

### DIFF
--- a/tests/reftests/reinstall.test
+++ b/tests/reftests/reinstall.test
@@ -86,7 +86,6 @@ The following actions would be performed:
 === install 1 package
   - install   c 1
 ### :::::: Check that reinstalling a non-installed package is equivalent to installing it
-### :::::: TODO: This is currently broken
 ### <pkg:d.1>
 opam-version: "2.0"
 ### opam reinstall d --cudf=reinstall --show --yes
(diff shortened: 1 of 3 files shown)


AltGr (approved):
Makes sense to me!

tests/reftests/reinstall.test
```diff
@@ -85,3 +85,88 @@ The following actions would be performed:
...
+ version: 1
+ conflicts: d
+-opam-query: 1
+ opam-name: d
+ opam-version: 1
+ 
+# Return code 1 #
+### :::::: Check again with one install and one non-installed
+### :::::: TODO: This is currently broken
+### opam reinstall d a --cudf=reinstall --show --yes
```
rjbou, line 113:
```suggestion
### <pkg:a>
opam-version: "2.0"
### opam reinstall d a --cudf=reinstall --show --yes
```
to avoid having `b` and `c` installs and their extra output

tests/reftests/reinstall.test
```diff
@@ -85,3 +85,88 @@ The following actions would be performed:
...
+### opam reinstall d --cudf=reinstall --show --yes
+d is not installed. Install it? [y/n] y
+The following actions would be performed:
+=== install 1 package
+  - install d 1
+### opam install d --cudf=install --show --yes
+The following actions would be performed:
+=== install 1 package
+  - install d 1
+### diff -u install-1.cudf reinstall-1.cudf
```
rjbou, line 100:
```suggestion
### diff -U 0 install-1.cudf reinstall-1.cudf
```
if portable ?

kit-ty-kate: I used `diff` without argument instead. That should be the most portable

tests/reftests/reinstall.test
kit-ty-kate: I think it's fine to have `b` and `c` around. It exhibit more behaviours and still tests the code-path i need here.

rjbou (approved):
thanks!

PR #5229, opened 2022-08-05
Show the reason for installing packages when using opam reinstall

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/5227

---------
~TODO:~
- [x] Some tests should be added to prevent regressions


AltGr (approved):
Note that this might also change behaviour with e.g. `with-test` flags in the package files; but this is a fix in this case as well!

kit-ty-kate:
Queued on top of #5228

rjbou (approved):
thanks!

PR #5230, opened 2022-08-05
Backport #5223 in 2.1.3

kit-ty-kate:
cc @hannesm


hannesm:
thanks for this and #5223 -- a tiny nitpick is that the script requires curl/wget only in some circumstances (so maybe the warning should be `[WARNING] This script may need curl or wget`.

PR #5236, opened 2022-08-09
Fix lots of issues with autopin, deps-only, dev deps, depexts...

AltGr:


AltGr:
Grmbl I think I found yet another issue : if the package is already pinned, but to a different target, and running with `--deps-only`, it seems the correct opam file is not used (at least I had new dependencies not getting picked up).

It's pretty complex though: we should not update the pinning with the new URL, yet we need to recover the opam file for the current target.

Do we need at all to allow `opam install ./foo2/foo.opam --deps-only` when `foo` is already pinned to `./foo1` ?

(it's not *that* far-fetched if you consider usage of `git worktree` without local switches for each of them)

kit-ty-kate:
> Do we need at all to allow `opam install ./foo2/foo.opam --deps-only` when `foo` is already pinned to `./foo1` ?

I think it would make sense

src/state/opamSysInteract.ml
```diff
@@ -669,6 +673,7 @@ let install_packages_commands_t ?(env=OpamVariable.Map.empty) sys_packages =
   | Netbsd -> ["pkgin", yes ["-y"] ("install" :: packages)], None
   | Openbsd -> ["pkg_add", yes ~no:["-i"] ["-I"] packages], None
   | Suse -> ["zypper", yes ["--non-interactive"] ("install"::packages)], None
+  | Test_environment -> ["echo", packages], None
```
rjbou, line 676: `sudo_run_command` need also to be updated

tests/reftests/autopin.test
```diff
@@ -0,0 +1,368 @@
...
+  "opam-core" {= version}
+  "opam-file-format" {>= "2.1.0"}
+  "re" {>= "1.9.0"}
+  "dune" {>= "2.0.0"}
+]
+depexts: ["some-depext"]
+### OPAMYES=1 OPAMFAKE=1 OPAMNODEPEXTS=0
+### opam var --global os-family=test-environment
+Added '[os-family "test-environment" "Set through 'opam var'"]' to field global-variables in global configuration
+### PKGDIRNAME=foo
```
rjbou, line 29: why externalise `foo`?

src/client/opamSolution.ml
```diff
@@ -422,7 +422,7 @@ let parallel_apply t
          (let spkgs = OpamSysPkg.Set.Op.(bypass -- !bypass_ref) in
           OpamConsole.note
             "Requirement for system package%s %s overridden in this switch. Use \
-             `opam option depext-bypass-=%s' to revert."
+             `opam option 'depext-bypass-=%S'' to revert."
```
kit-ty-kate, line 425: Are the single quotes really needed if we use `%S`?
```suggestion
             `opam option depext-bypass-=%S to revert."
```

AltGr: Yes, otherwise the shell removes the double quotes:
```shell-session
$ echo foo="bar"
foo=bar
```

kit-ty-kate: does this option really takes quotes? That's a bit surprising

AltGr: yes, it's based on the opam syntax I guess; I agree it's cumbersome — but all I did here was fix the message so that it printed an actually working command !

tests/reftests/autopin.test
rjbou: for windows, no more neded

tests/reftests/autopin.test
```diff
@@ -0,0 +1,394 @@
...
+  "opam-core" {= version}
+  "opam-file-format" {>= "2.1.0"}
+  "re" {>= "1.9.0"}
+  "dune" {>= "2.0.0"}
+]
+depexts: ["some-depext"]
+### OPAMYES=1 OPAMFAKE=1 OPAMNODEPEXTS=0
+### opam var --global os-family=dummy-success
+Added '[os-family "dummy-success" "Set through 'opam var'"]' to field global-variables in global configuration
+### opam switch create ./foo | unordered
```
AltGr, line 29: Where dir `--deps-only` go ?

rjbou: good question...

AltGr:
Just update this:
- now based on #5295 → should be rebased once (if) merged
- re-added the changelog patch
- the norepo test is great!

@rjbou I :heavy_check_mark:  and give you back the control on this. Thanks!

AltGr:
Note: https://github.com/ocaml/opam/pull/5236#issuecomment-1214361502 is still valid & we shouldn't forget to open a separate issue if still present once this is merged.

Issue #5239, opened 2022-08-11
opam depext support for macos


MSoegtropIMC:
I wanted to switch from opam 2.0.X to opam 2.1.2, but the integrated depext still doesn't seem to support macOS:

```
opam install conf-pkg-config
[NOTE] External dependency handling not supported for OS family 'macos'.
       You can disable this check using 'opam option --global depext=false'
The following actions will be performed:
  ∗ install conf-pkg-config 2

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫 
[ERROR] The compilation of conf-pkg-config.2 failed at "pkg-config --help".

#=== ERROR while compiling conf-pkg-config.2 ==================================#
# context     2.1.2 | macos/x86_64 | ocaml-option-flambda.1 ocaml-variants.4.13.1+options | https://opam.ocaml.org#73531078
# path        ~/.opam/__coq-platform.2022.04.1~8.16+rc1~2022.09~preview1/.opam-switch/build/conf-pkg-config.2
# command     ~/.opam/opam-init/hooks/sandbox.sh build pkg-config --help
# exit-code   71
# env-file    ~/.opam/log/conf-pkg-config-16221-c99f60.env
# output-file ~/.opam/log/conf-pkg-config-16221-c99f60.out
### output ###
# sandbox-exec: execvp() of 'pkg-config' failed: No such file or directory



<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫 
┌─ The following actions failed
│ λ build conf-pkg-config 2
└─ 
╶─ No changes have been performed
```

I wonder what the recommendation of the opam team for multi platform projects is. Should I stick with opam 2.0.X for the time being? Am I missing something?

```
# opam config report
# opam-version         2.1.2 
# self-upgrade         no
# system               arch=x86_64 os=macos os-distribution=macos os-version=11.6
# solver               builtin-mccs+glpk
# install-criteria     -removed,-count[avoid-version,changed],-count[version-lag,request],-count[version-lag,changed],-count[missing-depexts,changed],-changed
# upgrade-criteria     -removed,-count[avoid-version,changed],-count[version-lag,solution],-count[missing-depexts,changed],-new
# jobs                 1
# repositories         4 (http), 3 (local) (default repo at 73531078)
# pinned               0
# current-switch       __coq-platform.2022.04.1~8.16+rc1~2022.09~preview1
[...]

PR #5240, opened 2022-08-11
Improve the error message when neither MacPorts or Homebrew could be detected on macOS

kit-ty-kate:
Suggestion by @MSoegtropIMC in https://github.com/ocaml/opam/issues/5239


dra27 (approved):
LGTM!

PR #5251, opened 2022-08-17
Fix some empty conflict messages by filtering out conflicts of a package against itself early

kit-ty-kate:
The "self-constraints" are used for CUDF but it is an invariant behaviour for opam (several versions of the same package cannot be installed in the same switch)

Somehow doing this early fixes **at least** https://github.com/ocaml/opam/discussions/5130#discussioncomment-3398215

TODO:
* [x] Test more of the test cases
* [x] Add them to the testsuite

Fixes part of #4373


kit-ty-kate:
Requires #5253 to be merged first.

kit-ty-kate:
Should not be needed anymore after https://github.com/ocaml/opam/pull/5263

PR #5260, opened 2022-08-23
Silence env hooks errors

Firobe:
In some cases (such as when using [`opam-nix`](https://github.com/tweag/opam-nix)+`direnv`), we are in a shell with the env hook enabled but `opam` has been replaced by a dummy binary.

In this situation, `eval $(opam env)` will fail, which is expected. However, it will output annoying errors each time the hook is loaded. I think errors in the hook should always be silenced.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -365,6 +365,7 @@ users)
   * dash: recognize dash as a POSIX shell for opam env [#4816 @jonahbeckford]
   * pwsh,powershell: use $env: for opam env [#4816 @jonahbeckford]
   * command prompt: use SET for opam env [#4816 @jonahbeckford]
+  * silence env hooks errors [#5260 @Firobe]
 
 ## Doc
   * Standardise `macOS` use [#4782 @kit-ty-kate]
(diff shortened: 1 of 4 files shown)


rjbou (approved):

kit-ty-kate (commented):
I feel like this would silence potentially dangerous situation (e.g. maybe there is bug in `opam env` that would end up outputting random characters)

Is there no way for you to instead make the dummy binary output nothing on stdout instead?

Firobe:
> Is there no way for you to instead make the dummy binary output nothing on stdout instead?

The binary is out of my control in this case, but if silencing the errors is not a solution, then while not optimal I could try to wrap it inside another in my control, although that would be very hacky in my case.

I'll let you make the call :)

dra27:
We discussed this further today - the decision's not to go with this change. On the one hand, if something does go wrong with stock `opam`, the user is necessarily spammed with error messages on the console until the issue is resolved. On the other hand, if something has gone wrong, the non-advanced user may be left with no clue as to _why_ the environment isn't being set correctly.

It's especially strange in this case where the only use-case given so far for ignoring errors is to allow for a _shim_ of the `opam` command to be ignored (i.e. we're altering a script in `opam` to work around a problem caused by just _part_ of `opam` being altered by something else).

PR #4975, opened 2021-12-23
Add 'install --formula', rewrite '--deps-only'

AltGr:

PR #5261, opened 2022-08-26
Fix --best-effort regression introduced in #4975

LasseBlaauwbroek:
In PR #4796 I fixed a problem with `--best-effort`. Subsequently #4975 seems to
have broken this again. Not sure if this is the correct fix. There may also be
additional places where this is going wrong...

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -1116,7 +1116,7 @@ let install_t t ?ask ?(ignore_conflicts=false) ?(depext_only=false)
   let requested =
     OpamPackage.Name.Set.of_list (List.rev_map fst (atoms @ deps_atoms))
   in
-  let packages = OpamPackage.packages_of_names t.packages requested in
+  let packages = OpamFormula.packages_of_atoms t.packages (atoms @ deps_atoms) in
   let solution =
     let reinstall = if assume_built then Some pkg_reinstall else None in
     OpamSolution.resolve t Install
--- a/tests/reftests/best-effort.test
+++ b/tests/reftests/best-effort.test
@@ -10,6 +10,4 @@ opam-version: "2.0"
 opam-version: "2.0"
 depends: ["unavailable"]
 ### opam install --show --best-effort test.2
-The following actions would be performed:
-=== install 1 package
-  - install test 1
+Nothing to do.
--- a/tests/reftests/best-effort.test
+++ b/tests/reftests/best-effort.test
@@ -1,0 +1,15 @@
+N0REP0
+### opam switch create best-effort --empty
+### : Check that --best-effort works as intended when requesting a specific version of a package
+### <pkg:unavailable.1>
+opam-version: "2.0"
+available: false
+### <pkg:test.1>
+opam-version: "2.0"
+### <pkg:test.2>
+opam-version: "2.0"
+depends: ["unavailable"]
+### opam install --show --best-effort test.2
+The following actions would be performed:
+=== install 1 package
+  - install test 1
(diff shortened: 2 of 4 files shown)


src/client/opamClient.ml
```diff
@@ -1116,7 +1116,7 @@ let install_t t ?ask ?(ignore_conflicts=false) ?(depext_only=false)
   let requested =
     OpamPackage.Name.Set.of_list (List.rev_map fst (atoms @ deps_atoms))
   in
-  let packages = OpamPackage.packages_of_names t.packages requested in
+  let packages = OpamFormula.packages_of_atoms t.packages (atoms @ deps_atoms) in
```
kit-ty-kate, line 1119: Could you show an example of the bug? I'm a bit surprise this change is not equivalent

LasseBlaauwbroek: The example is the same as in #4796. The difference between `packages_of_names` and `packages_of_atoms` is that the former does not take into account the version constraints of the atoms. Hence, it will allow any version of the requested package to be installed instead of a specifically requested version.

kit-ty-kate: ah i see.

kit-ty-kate (approved):

rjbou (approved):
Certainly rebasing #4975 reintroduced it, my mistkae. Thanks for re-fixing it!
If we can add a test it would be better to avoid that.

LasseBlaauwbroek:
(I'm not familiar with the testing code so if someone else wants to do that I'd be grateful)

kit-ty-kate:
> (I'm not familiar with the testing code so if someone else wants to do that I'd be grateful)

Could you provide a rough test-case (something that fails in your case) so we can try add it to the reftests on our side?

LasseBlaauwbroek:
The test from #4796 `opam install coq-graph-theory.0.8 --best-effort` I guess. Also, if it fits within the testing framework, you could count how many instances of `opam-query: 1` occur in the cudf file.

tests/reftests/best-effort.test
```diff
@@ -0,0 +1,13 @@
...
+### <pkg:unavailable.1>
+opam-version: "2.0"
+available: false
+### <pkg:test.1>
+opam-version: "2.0"
+### <pkg:test.2>
+opam-version: "2.0"
+depends: ["unavailable"]
+### opam install --show --best-effort test.2
+Nothing to do.
```
kit-ty-kate, line 13: So I've added a simple test but the output of `opam install` doesn't seem right here.
It should have the same output as without `--best-effort`:
```
diff --git a/tests/reftests/best-effort.test b/tests/reftests/best-effort.test
index f358d5496..e30179eb8 100644
--- a/tests/reftests/best-effort.test
+++ b/tests/reftests/best-effort.test
@@ -9,5 +9,11 @@ opam-version: "2.0"
 ### <pkg:test.2>
 opam-version: "2.0"
 depends: ["unavailable"]
-### opam install --show --best-effort test.2
-Nothing to do.
+### opam install --show test.2
+[ERROR] Package conflict!
+  * Missing dependency:
+    - unavailable
+    unmet availability conditions: 'false'
+
+No solution found, exiting
+# Return code 20 #
```

LasseBlaauwbroek: Why? You are trying to install a package that cannot be installed. Hence `--best-effort` just skips it and does nothing, while the normal command throws an error.

kit-ty-kate: Mmh, ok fair enough, I've never really used `--best-effort` ^^"

LasseBlaauwbroek: The simplest test that I can think of is to have `my-package.2` that is unavailable and `my-package.1` that is available. Then, the output of `opam install my-package.2` and `opam install my-package.2 --best-effort` should be the same after this PR, but different before this PR.

kit-ty-kate:
Thanks a lot!

PR #5171, opened 2022-07-01
[WIP] Add `opam tree` subcommand

cannorin:
Closes #3775 .

This PR adds a subcommand `opam tree` which displays a dependency tree as a Unicode art or ASCII art.

It can also display a reverse-dependency tree (through `opam tree --rev-deps` option or `opam why` alias), which can be useful to examine how dependency versions get constrained.

image

## TODO
- [ ] ask for feedbacks
- [ ] complete doc and man
- [ ] add tests
- [ ] update `master_changes.md` file

PR #5266, opened 2022-08-30
Fix a bug Windows CI appearing when doc/man/opam-topics.inc changes

kit-ty-kate:
This issue showed up in https://github.com/ocaml/opam/pull/5171:
```
 error: Your local changes to the following files would be overwritten by checkout:
	doc/man/opam-topics.inc
Please commit your changes or stash them before you switch branches.
Aborting
+ git diff
warning: in the working copy of 'doc/man/opam-topics.inc', CRLF will be replaced by LF the next time Git touches it
```


kit-ty-kate:
Fixed by #5281 instead

PR #5284, opened 2022-09-14
Reset the "jobs" config variable when upgrading from opam 2.0

kit-ty-kate:
See https://github.com/ocurrent/ocaml-dockerfile/pull/92

### DIFF
--- a/src/state/opamFormatUpgrade.ml
+++ b/src/state/opamFormatUpgrade.ml
@@ -1074,7 +1074,26 @@ let from_2_1_alpha2_to_v2_1_rc root conf =
 
 let from_2_1_rc_to_v2_1 _ conf = conf
 
-let from_2_0_to_v2_1 _ conf = conf
+let from_2_0_to_v2_1 _ conf =
+  (* In opam < 2.1 "jobs" was set during initialisation
+     This creates problems when upgrading from opam 2.0 as it
+     sets the job count for good even if the CPU is replaced.
+     See https://github.com/ocurrent/ocaml-dockerfile/pull/92 *)
+  let info_jobs_changed ~prev_jobs =
+    OpamConsole.note
+      "The 'jobs' option was reset, its value was %d and its new value \
+      will vary according to the current number of cores on your machine. \
+       If it really was intended, you can set it again using:\n    \
+       opam option jobs=%d --global" prev_jobs prev_jobs;
+  in
+  (* We print a note in case the previous value of 'jobs' does not
+     match the default in opam 2.0 as we can't determine if the value
+     was modified or if the hardware changed. *)
+  (match OpamFile.Config.jobs conf with
+   | Some prev_jobs when prev_jobs = max 1 (OpamSysPoll.cores () - 1) -> ()
+   | Some prev_jobs -> info_jobs_changed ~prev_jobs
+   | None -> info_jobs_changed ~prev_jobs:1);
+  OpamFile.Config.with_jobs_opt None conf
 
 let latest_version = OpamFile.Config.root_version
 
--- a/master_changes.md
+++ b/master_changes.md
@@ -83,6 +83,9 @@ users)
   * Really install invariant formula if not installed in switch [#5188 @rjbou]
   * On import, check that installed pinned packages changed, reinstall if so [#5181 @rjbou - fix #5173]
 
+## Config
+  * Reset the "jobs" config variable when upgrading from opam 2.0 [#5284 @kit-ty-kate]
+
 ## Pin
   * Switch the default version when undefined from ~dev to dev [#4949 @kit-ty-kate]
   * ◈ New option `opam pin --current` to fix a package in its current state (avoiding pending reinstallations or removals from the repository) [#4973 @AltGr - fix #4970]
--- a/tests/reftests/opamroot-versions.test
+++ b/tests/reftests/opamroot-versions.test
@@ -681,11 +681,15 @@ No configuration file found, using built-in defaults.
 ### opam option jobs | "${OPAMROOTVERSION}$" -> "current"
 GSTATE                          LOAD-GLOBAL-STATE @ ${BASEDIR}/OPAM
 FMT_UPG                         On-the-fly config upgrade, from 2.0 to current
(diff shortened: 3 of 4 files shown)


src/state/opamFormatUpgrade.ml
```diff
@@ -1074,7 +1074,15 @@ let from_2_1_alpha2_to_v2_1_rc root conf =
 
 let from_2_1_rc_to_v2_1 _ conf = conf
 
-let from_2_0_to_v2_1 _ conf = conf
+let from_2_0_to_v2_1 _ conf =
+  match OpamFile.Config.jobs conf with
+  | Some n when n = OpamSystem.cpu_count () - 1 ->
```
rjbou, line 1079:
```suggestion
  | Some n when n = OpamSysPoll.cores ()) - 1 ->
```

rjbou:
Agree on the idea, but what if user set its jobs variable to that value? Should we add a warning to inform that we are removing that value?

src/state/opamFormatUpgrade.ml
rjbou, line 1079: Why this condition? If cpu changed meanwhile, it can no more be the good value

kit-ty-kate: Ah.. I was tired ^^"

kit-ty-kate:
> Agree on the idea, but what if user set its jobs variable to that value? Should we add a warning to inform that we are removing that value?

Indeed that's a good idea. I've done that and fixed the rest of the issues. It should be good to be reviewed once again.

src/state/opamFormatUpgrade.ml
```diff
@@ -1074,7 +1074,17 @@ let from_2_1_alpha2_to_v2_1_rc root conf =
...
+  (* In opam < 2.1 "jobs" was set during initialisation
+     This creates problems when upgrading from opam 2.0 as it
+     sets the job count for good even if the CPU is replaced.
+     See https://github.com/ocurrent/ocaml-dockerfile/pull/92 *)
+  (* We print a warning in case the previous value of 'jobs' does not
+     match the default in opam 2.0 as we can't determine if the value
+     was modified or if the hardware changed. *)
+  if OpamFile.Config.jobs conf = OpamSysPoll.cores () - 1 then
+    OpamConsole.warning "The 'jobs' option was reset";
+  OpamFile.Config.with_jobs_opt None conf
```
rjbou, line 1087:
```suggestion
  match  OpamFile.Config.jobs conf with
  | Some j ->
    OpamConsole.warning
      "The 'jobs' option was reset, its value value was %d. To reset it, use\n    \
        `opam option jobs=%d --global`" j j;
    OpamFile.Config.with_jobs_opt None conf
   | None -> conf
```

Something like ?

kit-ty-kate:
```suggestion
  if OpamFile.Config.jobs conf = OpamSysPoll.cores () - 1 then
    OpamConsole.warning
      "The 'jobs' option was reset, its value value was %d. \
       If it really was intended, you can set it again using:\n    \
       `opam option jobs=%d --global`" j j;
  OpamFile.Config.with_jobs_opt None conf
```

src/state/opamFormatUpgrade.ml
```diff
@@ -1074,7 +1074,20 @@ let from_2_1_alpha2_to_v2_1_rc root conf =
...
+     See https://github.com/ocurrent/ocaml-dockerfile/pull/92 *)
+  (* We print a warning in case the previous value of 'jobs' does not
+     match the default in opam 2.0 as we can't determine if the value
+     was modified or if the hardware changed. *)
+  if OpamFile.Config.jobs conf = Some (OpamSysPoll.cores () - 1) then
+    OpamConsole.note
+      "The 'jobs' option was reset, its value was %d. \
+       If it really was intended, you can set it again using:\n    \
+       opam option jobs=%d --global" j j;
+  OpamFile.Config.with_jobs_opt None conf
```
rjbou, line 1090:
```suggestion
  (match OpamFile.Config.jobs conf with
  | Some j when j = OpamSysPoll.cores () - 1 ->
    (* We print a warning in case the previous value of 'jobs' does not
       match the default in opam 2.0 as we can't determine if the value
       was modified or if the hardware changed. *)
    OpamConsole.note
      "The 'jobs' option was reset, its value was %d. \
       If it really was intended, you can set it again using:\n    \
       opam option jobs=%d --global" j j
   | _ -> ());
  OpamFile.Config.with_jobs_opt None conf
```

kit-ty-kate:
```suggestion
  let info_jobs_changed ~prev_jobs =
    OpamConsole.note
      "The 'jobs' option was reset, its value was %d. \
       If it really was intended, you can set it again using:\n    \
       opam option jobs=%d --global" prev_jobs prev_jobs;
  in
  (* We print a note in case the previous value of 'jobs' does not
     match the default in opam 2.0 as we can't determine if the value
     was modified or if the hardware changed. *)
  (match OpamFile.Config.jobs conf with
   | Some prev_jobs when prev_jobs = max 1 (OpamSysPoll.cores () - 1) -> ()
   | Some prev_jobs -> info_jobs_changed ~prev_jobs
   | None -> info_jobs_changed ~prev_jobs:1);
  OpamFile.Config.with_jobs_opt None conf
```

src/state/opamFormatUpgrade.ml
```diff
@@ -1074,7 +1074,25 @@ let from_2_1_alpha2_to_v2_1_rc root conf =
...
 
-let from_2_0_to_v2_1 _ conf = conf
+let from_2_0_to_v2_1 _ conf =
+  (* In opam < 2.1 "jobs" was set during initialisation
+     This creates problems when upgrading from opam 2.0 as it
+     sets the job count for good even if the CPU is replaced.
+     See https://github.com/ocurrent/ocaml-dockerfile/pull/92 *)
+  let info_jobs_changed ~prev_jobs =
+    OpamConsole.note
+      "The 'jobs' option was reset, its value was %d. \
```
kit-ty-kate, line 1084:
```suggestion
      "The 'jobs' option was reset, its value was %d and its new value \
      will vary according to the current number of cores on your machine. \
```

AltGr (approved):

PR #5285, opened 2022-09-15
Reftest: add path replacement command command

rjbou:
Command are resolved in opam, which leads to some full path printing. Add `sed-cmd` command to help this rewrite.

```diff
-### opam install ongoing -v | '.*(/|\\|")test(\.exe)?"? "' -> 'test "' 
+### opam install ongoing -v | sed-cmd test
```

### DIFF
--- a/tests/reftests/run.ml
+++ b/tests/reftests/run.ml
@@ -36,11 +36,13 @@
        (`foo'bar'` is not translated to `foobar`)
      * Variable expansion in arguments (`$FOO` or `${FOO}`). Undefined variables
        are left as-is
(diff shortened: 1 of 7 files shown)


tests/reftests/run.ml
```diff
@@ -413,6 +415,20 @@ module Parse = struct
             get_rewr (unordered, (posix_re re, Grep) :: acc) r
           | "|" :: "unordered" :: r ->
             get_rewr (true, acc) r
+          | "|" :: "sed-cmd" :: cmd :: r ->
```
rjbou, line 418: not sure about the name, any ideas ?

PR #5286, opened 2022-09-15
Reftests: handle cygwin paths

rjbou:
Some windows paths were not prefixed with cygwin path, and then not catch by path regexp.
Also, for some tests we get `PATH` from environment that is in cygwin style while `Sys.getcwd` gives the windows style one. Add in `BASEDIR` regexp both style.

### DIFF
--- a/tests/reftests/repository.test
+++ b/tests/reftests/repository.test
@@ -24,17 +24,17 @@ opam-version: "2.0"
 build: ["test" "-f" "baz"]
 ### <REPO/packages/foo/foo.2/files/baz>
 some content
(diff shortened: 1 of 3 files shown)


tests/reftests/repository.test
```diff
@@ -61,13 +61,13 @@ opam-version: "2.0"
 build: ["test" "-f" "baz"]
 ### <REPO/packages/foo/foo.4/files/baz>
 some content
-### opam update -vv | grep '^\+' | '.*tar(\.exe)?"? "xfz" ".*(/|\\)OPAM(/|\\)repo(/|\\)tarred\.tar\.gz" "-C" ".*"' -> 'tar xfz OPAM/repo/tarred.tar.gz -C TMPDIR' | '.*tar(\.exe)?"? "cfz" ".*(/|\\)OPAM(/|\\)repo(/|\\)tarred\.tar\.gz\.tmp" "-C" ".*" "tarred"' -> 'tar cfz OPAM/repo/tarred.tar.gz.tmp -C TMPDIR tarred' | sed-cmd diff | '.*(/|\\)patch(\.exe)?"?.*"' -> 'patch'
-tar xfz OPAM/repo/tarred.tar.gz -C TMPDIR
+### opam update -vv | grep '^\+' | sed-cmd tar | sed-cmd diff | sed-cmd patch | "patch-.*" -> "patchfile"
```
kit-ty-kate, line 64:
```suggestion
### opam update -vv | grep '^\+' | sed-cmd tar | sed-cmd diff | sed-cmd patch | 'patch-.*"' -> 'patchfile"'
```

dra27 (commented):
I like the effect on the tests! Some possible tweaks to exactly what we do with the paths

rjbou:
superseed by #5723

PR #5295, opened 2022-09-22
Fix behaviour of menu and depexts in non-interactive environments

AltGr:
The rationale is now to assume the "no" case in the menus whenever there is no input to read from.

### DIFF
--- a/src/client/opamSolution.ml
+++ b/src/client/opamSolution.ml
@@ -1160,9 +1160,8 @@ let install_depexts ?(force_depext=false) ?(confirm=true) t packages =
     else if OpamFile.Config.depext_run_installs t.switch_global.config then
       if confirm then menu t sys_packages else auto_install t sys_packages
     else
-      manual_install ~noninteractive:`Ignore t sys_packages
+      manual_install t sys_packages
   and menu t sys_packages =
-    (* Called only if run install is true *)
     let answer =
       let pkgman =
         OpamConsole.colorise `yellow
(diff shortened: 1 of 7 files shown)


src/client/opamSolution.ml
```diff
@@ -1228,17 +1228,18 @@ let install_depexts ?(force_depext=false) ?(confirm=true) t packages =
       map_sysmap (fun _ -> OpamSysPkg.Set.empty) t
     with Failure msg ->
       OpamConsole.error "%s" msg;
-      manual_install ~noninteractive:`Quit t sys_packages
+      check_again t sys_packages
   and check_again t sys_packages =
-    let needed, _notfound = OpamSysInteract.packages_status sys_packages in
+    let needed, _notfound = OpamSysInteract.packages_status ~env sys_packages in
```
AltGr, line 1233: This one was well hidden, but at least one test caught it. Maybe the parameter shouldn't have been optional :grimacing:

src/client/opamSolution.ml
```diff
@@ -1228,17 +1228,18 @@ let install_depexts ?(force_depext=false) ?(confirm=true) t packages =
...
     let installed = OpamSysPkg.Set.diff sys_packages needed in
     let t, sys_packages =
       map_sysmap (fun sysp -> OpamSysPkg.Set.diff sysp installed) t, needed
     in
     if OpamSysPkg.Set.is_empty sys_packages then t else
       (OpamConsole.error "These packages are still missing: %s\n"
          (syspkgs_to_string sys_packages);
-       entry_point t sys_packages)
+       if OpamStd.Sys.tty_in then entry_point t sys_packages
+       else give_up ())
```
AltGr, line 1242: Always avoid looping if we don't read from a tty

src/client/opamSolution.ml
kit-ty-kate: Totally agree! Optional parameters should be avoided as much as possible IMO.

AltGr:
As mentionned in private, this is more restrictive when running non-interactively but better reflects what I had in mind when adding the menu. I am open to discussion if it has some unwanted effects though, esp. since different workarounds have been proposed in the meantime.

rjbou (approved):
thanks!

rjbou:
The new mechanism is good for me! clearer and more local

PR #5303, opened 2022-09-28
Support --json option in opam tree

### LINKED ISSUE #5298: `opam tree` JSON output for Opam Dependency Submission
I think we all know that https://github.com/ocaml/opam/issues/3775 and https://github.com/ocaml/opam/pull/5171 eventually merged and it works pretty well. But the tree is output for humans, not for machines. I would like to propose a subsequent PR implementation that outputs JSON for ODS. The structure here doesn't have to be for ODS, I think it would be better to have a more general structure. Does anyone have any thoughts?

cannorin:
Closes #5298.

This PR adds support for `--json` option in `opam tree` command.

Here is an example output: https://gist.github.com/cannorin/ae87f55695c38356a9b880e0418a7a12

Currently, the output JSON has the type `Forest` in the following TS definition:

```typescript
interface Package {
  name: string;
  version: string;
}

interface DepsTree extends Package {
  dependencies: DepsNode[];
}

interface DepsNode extends DepsTree {
  satisfies: string | null;
  is_dup: boolean;
}

interface RevdepsTree extends Package {
  requirements: RevdepsNode[];
}

interface RevdepsNode extends RevdepsTree {
  demands: string | null;
  is_dup: boolean;
}

type Forest = DepsTree[] | RevdepsTree[]
```

We discuss the structure of the output JSON in #5298.

TODO:
- [x] Add tests
- [x] ~Update doc and man~ not needed
- [x] Update `master_changes.md`

### DIFF
--- a/tests/reftests/json.unix.test
+++ b/tests/reftests/json.unix.test
@@ -730,3 +730,576 @@ Done.
     }
   ]
 }
+### : tree :
+### opam switch create tree --empty
+### opam tree --json=out.json
+[ERROR] No package is installed
+# Return code 5 #
+### json-cat out.json
+{
+  "opam-version": "currentv",
+  "command-line": [
+    "${OPAM}",
+    "tree",
+    "--json=out.json"
+  ]
+}
+### <pkg:a.1>
+opam-version: "2.0"
+### <pkg:b.1>
+opam-version: "2.0"
+depends: "a"
+### <pkg:c.1>
+opam-version: "2.0"
+depends: "b"
+### <pkg:d.1>
+opam-version: "2.0"
+### <pkg:c-build.1>
+opam-version: "2.0"
+### <pkg:c.1>
+opam-version: "2.0"
+depends: [ "b" "a" "c-build" {build} ]
+### <pkg:e.1>
+opam-version: "2.0"
+depends: [ "a" "d" {<"2" & with-test}  "b" {post} ]
+### <pkg:f.1>
+opam-version: "2.0"
+depends: [ "b" {<"2"} ]
+### <pkg:g.1>
+opam-version: "2.0"
+depends: [ "b" {>="1" & <"3"} ]
+### <pkg:h.1>
+opam-version: "2.0"
+depends: [ "f" "g" ]
+### <pkg:i.1>
+opam-version: "2.0"
+depends: [
+  "b" {
+    <"3"
+    & (!with-test | <  "2")
+    & (!with-doc  | >= "1")
+  }
+]
+### <pkg:j.1>
+opam-version: "2.0"
+depends: [ "a" "k" {with-test} ]
+### <pkg:k.1>
+opam-version: "2.0"
+### opam tree a b c d e f g h i j --json=out.json
+The following actions are simulated:
+=== install 11 packages
+  - install a       1
+  - install b       1
+  - install c       1
+  - install c-build 1 [required by c]
+  - install d       1
+  - install e       1
+  - install f       1
+  - install g       1
+  - install h       1
+  - install i       1
+  - install j       1
+
+a.1
+
+b.1
+'-- a.1 [*]
+
+c.1
+|-- a.1 [*]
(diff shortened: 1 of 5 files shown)


tests/reftests/json.unix.test
```diff
@@ -730,3 +730,436 @@ Done.
...
+[ERROR] No package is installed
+# Return code 5 #
+### json-cat out.json
+{
+  "opam-version": "currentv",
+  "command-line": [
+    "${OPAM}",
+    "tree",
+    "--json=out.json"
+  ]
```
rjbou, line 745: Should we add here that there was no output (the error ?)

rjbou (commented):
Thanks for the PR!
I've moved the test to json testfile, and added some commands to test (with error, with simulated)

tests/reftests/json.unix.test
cannorin: I suppose having `null` or having no result (= `undefined`) is enough? The current implementation does not add a result, which means it would be `undefined` if read from JS.

tests/reftests/json.unix.test
```diff
@@ -730,3 +730,436 @@ Done.
...
+    "d",
+    "e",
+    "f",
+    "g",
+    "h",
+    "i",
+    "j",
+    "--json=out.json"
+  ],
+  "tree": [
```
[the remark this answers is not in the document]
cannorin: I'm not very sure, as there is no other command which outputs the solution to JSON with `OpamSolver.solution_to_json` as far as I can find.

rjbou:
thoughts from dev meeting: we should add switch name into the json, no need to add error / simulated state

cannorin:
@rjbou I've added the switch name into the json, and now it has the following definition (of interface `Output`):

```typescript
interface Package {
  name: string;
  version: string;
}

interface DepsTree extends Package {
  dependencies: DepsNode[];
}

interface DepsNode extends DepsTree {
  satisfies: string | null;
  is_dup: boolean;
}

interface RevdepsTree extends Package {
  requirements: RevdepsNode[];
}

interface RevdepsNode extends RevdepsTree {
  demands: string | null;
  is_dup: boolean;
}

interface Output {
  switch?: string; // the name of the switch. will be absent if `--no-switch` is specified.
  result: DepsTree[] | RevdepsTree[];
}
```

If we have more things to add, we can just put them as new members of the `Output` interface. cc: @smorimoto

cannorin:
Oops I broke some other tests in `json.unix.test` and I'm fixing it now.

kit-ty-kate:
@smorimoto Could you confirm on whether or not this is what you wanted?

smorimoto:
@kit-ty-kate Yes! That looks OK! Let's go!

kit-ty-kate (approved):

smorimoto:
When will this be deployed?

cannorin:
`is_duplicate` may be better.

I didn't come up with better terms than demands/satisfies...

tests/reftests/json.unix.test
```diff
@@ -730,3 +730,438 @@ Done.
...
+          "name": "a",
+          "version": "1",
+          "satisfies": null,
+          "is_dup": true,
+          "dependencies": []
+        }
+      ]
+    }
+  ]
+}
```
rjbou, line 1167: add rev deps example

dra27:
`is_duplicate` and `satisfies` seem good - might be clearer with the example to figure whether something other `demands`/`requirements` wanted (and this is all just fixing names - the fundamental layout looks fine)

tests/reftests/json.unix.test
rjbou: It is a reminder for me :)

smorimoto:
Updates?

rjbou:
queued on #5405

rjbou:
Thanks!

smorimoto:
Thank you for everyone!

PR #5308, opened 2022-10-07
Add opam pin remove --all

kit-ty-kate:
I often find the need to clean all pins from the current switch i am working on, be it because the pin packages have all been released or other reasons, and listing all the packages or repos one by one can be extremely tedious.

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -3184,6 +3184,10 @@ let pin ?(unpin_only=false) cli =
        changes, and may also be used to keep a package that was removed \
        upstream."
   in
+  let all =
+    mk_flag ~cli (cli_from cli2_2) ["all"]
+      "When unpinning, removes all pins in the given switch."
+  in
   let guess_names kind ?locked ~recurse ?subpath url k =
     let found, cleanup =
       match OpamUrl.local_dir url with
@@ -3300,7 +3304,7 @@ let pin ?(unpin_only=false) cli =
   let pin
       global_options build_options
       kind edit no_act dev_repo print_short recurse subpath normalise
-      with_version current
+      with_version current all
       command params () =
     apply_global_options cli global_options;
     apply_build_options cli build_options;
@@ -3317,6 +3321,10 @@ let pin ?(unpin_only=false) cli =
         `list
       | Some `scan, [url] ->
         `scan url
+      | Some `remove, [] when all ->
+        `remove_all
+      | _, _ when all ->
+        `incorrect
       | Some `remove, (_::_ as arg) ->
         `remove arg
       | Some `edit, [nv]  ->
@@ -3418,6 +3426,14 @@ let pin ?(unpin_only=false) cli =
       else
         (OpamSwitchState.drop @@ OpamClient.PIN.unpin st ~action to_unpin;
          `Ok ())
+    | `remove_all ->
+      OpamGlobalState.with_ `Lock_none @@ fun gt ->
+      OpamSwitchState.with_ `Lock_write gt @@ fun st ->
+      let to_unpin =
+        OpamPackage.Set.to_list_map OpamPackage.name (OpamPinned.packages st)
+      in
+      OpamSwitchState.drop @@ OpamClient.PIN.unpin st ~action to_unpin;
+      `Ok ()
     | `edit nv  ->
       (match (fst package) nv with
        | `Ok (name, version) ->
@@ -3515,7 +3531,7 @@ let pin ?(unpin_only=false) cli =
           $global_options cli $build_options cli
           $kind $edit $no_act $dev_repo $print_short_flag cli cli_original
           $recurse cli $subpath cli
-          $normalise $with_version $current
+          $normalise $with_version $current $all
           $command $params)
 
 (* SOURCE *)
--- a/src/core/opamStd.mli
+++ b/src/core/opamStd.mli
@@ -29,6 +29,7 @@ module type SET = sig
   val choose_opt: t -> elt option
 
   val of_list: elt list -> t
+  val to_list_map: (elt -> 'b) -> t -> 'b list
   val to_string: t -> string
   val to_json: t OpamJson.encoder
   val of_json: t OpamJson.decoder
--- a/tests/reftests/pin.test
+++ b/tests/reftests/pin.test
@@ -176,13 +176,61 @@ echo "another-inexistant" "inexistant"
 -> installed bar.dev
 Done.
 ### : Test opam pin remove <pkg>.<version>
-### opam pin
-bar.dev    rsync             file://${BASEDIR}/bar
-foo.1      local definition
-nip.dev    rsync             file://${BASEDIR}/nip
-qux.dev    rsync             file://${BASEDIR}/qux
 ### opam pin remove --no-action bar.dev
 Ok, bar is no longer pinned to file://${BASEDIR}/bar (version dev)
 ### opam pin remove --no-action nip.wrong-version
 [ERROR] nip is pinned but not to version wrong-version. Skipping.
 # Return code 2 #
+### : Test opam pin remove --all
+### OPAMNODEPEXTS=1
+### opam pin add --no-action bar.dev ./bar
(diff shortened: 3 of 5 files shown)


avsm:
Yes please!  I also have the case that it's difficult to undo an `opam install`'s pins

```
opam install .
# proceeds to pin all the local packages
# how do i undo the previous command? opam remove doesn't unpin them.
```

kit-ty-kate:
Could you open a new issue to track that problem?

src/client/opamCommands.ml
```diff
@@ -3374,6 +3380,12 @@ let pin ?(unpin_only=false) cli =
       else
         (OpamSwitchState.drop @@ OpamClient.PIN.unpin st ~action to_unpin;
          `Ok ())
+    | `remove_all ->
+      OpamGlobalState.with_ `Lock_none @@ fun gt ->
+      OpamSwitchState.with_ `Lock_write gt @@ fun st ->
+      let to_unpin = OpamPackage.Set.to_list_map OpamPackage.name (OpamPinned.packages st) in
+      OpamSwitchState.drop (OpamClient.PIN.unpin st ~action to_unpin);
```
rjbou, line 3387:
```suggestion
      OpamSwitchState.drop @@ OpamClient.PIN.unpin st ~action to_unpin;
```
Just to keep the same style

src/client/opamCommands.ml
```diff
@@ -3280,6 +3284,8 @@ let pin ?(unpin_only=false) cli =
         `list
       | Some `scan, [url] ->
         `scan url
+      | Some `remove, _ when all ->
+        `remove_all
```
rjbou, line 3288: We should error when remove is given with an argument and `--all`.

src/client/opamCommands.ml
```diff
@@ -3374,6 +3380,12 @@ let pin ?(unpin_only=false) cli =
       else
         (OpamSwitchState.drop @@ OpamClient.PIN.unpin st ~action to_unpin;
          `Ok ())
+    | `remove_all ->
+      OpamGlobalState.with_ `Lock_none @@ fun gt ->
+      OpamSwitchState.with_ `Lock_write gt @@ fun st ->
+      let to_unpin = OpamPackage.Set.to_list_map OpamPackage.name (OpamPinned.packages st) in
```
rjbou, line 3386:
```suggestion
      let to_unpin =
        OpamPackage.Set.to_list_map OpamPackage.name (OpamPinned.packages st)
      in
```

tests/reftests/pin.test
```diff
@@ -175,3 +175,28 @@ echo "another-inexistant" "inexistant"
...
+  - remove nip dev
+  - remove qux dev
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> removed   bar.dev
+-> removed   foo.1
+-> removed   nip.dev
+-> removed   qux.dev
+Done.
+### opam pin
```
rjbou, line 202: we should add also a test with `--no-action --all`

tests/reftests/pin.test
```diff
@@ -175,3 +175,28 @@ echo "another-inexistant" "inexistant"
 -> installed qux.dev
 -> installed bar.dev
 Done.
+### opam pin
+bar.dev    rsync             file://${BASEDIR}/bar
+foo.1      local definition
+nip.dev    rsync             file://${BASEDIR}/nip
+qux.dev    rsync             file://${BASEDIR}/qux
+### opam pin remove -a
```
rjbou, line 183:
```suggestion
### opam pin remove --all
```
better always explicit argument in tests

src/core/opamStd.mli
```diff
@@ -29,6 +29,7 @@ module type SET = sig
   val choose_opt: t -> elt option
 
   val of_list: elt list -> t
+  val to_list_map: (elt -> 'b) -> t -> 'b list
```
rjbou, line 32: Can these changes be in a separate commit?

kit-ty-kate:
All done

src/client/opamCommands.ml
```diff
@@ -3147,6 +3147,10 @@ let pin ?(unpin_only=false) cli =
        changes, and may also be used to keep a package that was removed \
        upstream."
   in
+  let all =
+    mk_flag ~cli (cli_from cli2_2) ["a"; "all"]
+      "When unpinning, removes all pin in the given switch."
```
dra27, line 3152:
```suggestion
      "When unpinning, removes all pins in the given switch."
```

dra27 (commented):
Tiny tweak to the message.

More importantly, `-a` has a lot of different meanings across opam already, so this doesn't technically make things that much worse, however `opam remove -a` removes automatically-installed packages but `opam pin remove -a` would now mean remove _all_ pins. Might it be worth either using a different letter (`opam admin` already has `-A` for a `--all` option) or just not having a short form at all for this one?

src/client/opamCommands.ml
```diff
@@ -3280,6 +3284,11 @@ let pin ?(unpin_only=false) cli =
         `list
       | Some `scan, [url] ->
         `scan url
+      | Some `remove, params when all ->
+        if params = [] then
+          `remove_all
+        else
+          `incorrect
```
rjbou, line 3291:
```suggestion
      | Some `remove, [] when all ->
        `incorrect
      | Some `remove, params when all ->
        `remove_all
```

avsm:
I'm not convinced we need this PR because:

1) `opam pin remove -a` has a different meaning from `opam remove -a` (as @dra27 notes)
2) `opam pin remove $(opam pin -s)` accomplishes what this PR does -- we could document this.
3) `opam pin install` and `opam pin remove` are not currently symmetric, and there's no easy way to do this.

I'd recommend not adding more CLI options to opam when there are existing unix-style solutions that work, and focussing on fixing the really rough edges like 3) above which is hard to do (as you cannot easily probe the set of locally pinned packages without some wildcard magic)

tests/reftests/pin.test
```diff
@@ -176,13 +176,66 @@ echo "another-inexistant" "inexistant"
...
 foo.1      local definition
 nip.dev    rsync             file://${BASEDIR}/nip
 qux.dev    rsync             file://${BASEDIR}/qux
-### opam pin remove --no-action bar.dev
+### OPAMNODEPEXTS=1
+### opam pin
+bar.dev    rsync             file://${BASEDIR}/bar
+foo.1      local definition
+nip.dev    rsync             file://${BASEDIR}/nip
+qux.dev    rsync             file://${BASEDIR}/qux
```
rjbou, line 198:
```suggestion
```
the opam pin is doubled

rjbou (approved):

src/client/opamCommands.ml
```diff
@@ -3184,6 +3184,10 @@ let pin ?(unpin_only=false) cli =
        changes, and may also be used to keep a package that was removed \
        upstream."
   in
+  let all =
+    mk_flag ~cli (cli_from cli2_2) ["all"]
+      "When unpinning, removes all pins in the given switch."
```
rjbou, line 3189: Should we add a word about this command being a no-op on other pin commands ?

kit-ty-kate: I chose to make the use of `--all` outside of `remove` without any parameter an error instead. Does that work for you?

tests/reftests/pin.test
kit-ty-kate: Done

src/client/opamCommands.ml
rjbou: good for me!

tests/reftests/pin.test
```diff
@@ -176,13 +176,66 @@ echo "another-inexistant" "inexistant"
 -> installed bar.dev
 Done.
 ### : Test opam pin remove <pkg>.<version>
+### opam pin remove --no-action bar.dev
```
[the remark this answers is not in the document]
rjbou: yes, i had in mind to put it later indeed. Perfect!

rjbou (approved):

PR #5309, opened 2022-10-10
Replace fgrep with grep -F

MisterDA:
> The egrep and fgrep commands, which have been deprecated since release 2.5.3 (2007), now warn that they are obsolescent and should be replaced by grep -E and grep -F.

https://lists.gnu.org/archive/html/info-gnu/2022-09/msg00001.html

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -294,6 +294,7 @@ users)
   * Unify constructors for powershell hosts [#5203 @dra27]
   * Fix lazy compilation of regular expression in OpamFormula.atom_of_string [#5211 @dra27]
   * [BUG] Display correct exception backtrace on uncaught exception on Windows [#5216 @dra27]
+  * Use grep -F instead of fgrep, as the latter is deprecated [#5309 @MisterDA]
 
 ## Internal: Windows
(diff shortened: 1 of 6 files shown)


rjbou (approved):

master_changes.md
```diff
@@ -294,6 +294,7 @@ users)
   * Unify constructors for powershell hosts [#5203 @dra27]
   * Fix lazy compilation of regular expression in OpamFormula.atom_of_string [#5211 @dra27]
   * [BUG] Display correct exception backtrace on uncaught exception on Windows [#5216 @dra27]
+  * fgrep is obsolescent; using grep -F [#???? @MisterDA]
```
kit-ty-kate, line 297:
```suggestion
  * Use grep -F instead of fgrep as the later is deprecated [#5309 @MisterDA]
```

dra27: s/later/latter/ 🙂

dra27:
LGTM, thanks, and should be applied to 2.0 and 2.1. RIP `fgrep` and `egrep` ☠

rjbou:
thanks!

MisterDA:
BTW I checked and there were no uses of `egrep`.

PR #5311, opened 2022-10-12
doc: update switch manpage

### LINKED ISSUE #5307: Man-pages: make consequences of `eval $(opam env --switch=SWITCH --set-switch)` clearer
Hi. I've been using `eval $(opam env --switch=SWITCH --set-switch)` for a while to change the switch since that is what it says in the man-pages. Is that still recommended, also when using local switches? Might it be a good idea to point out in the man-pages that it overrides the `OPAMSWITCH` variable? I've talked about that with @kit-ty-kate.

rjbou:
fix #5307 : mention shell hooks prior to eval opam env

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -2549,12 +2549,14 @@ let switch cli =
          $(b,--no-install) is specified."
         OpamArg.dir_sep OpamSwitch.external_dirname);
     `P (Printf.sprintf
-         "$(b,opam switch set) sets the default switch globally, but it is also \
-         possible to select a switch in a given shell session, using the \
-         environment. For that, use $(i,%s)."
-        OpamEnv.(
-          shell_eval_invocation shell
-            (opam_env_invocation ~switch:"SWITCH" ~set_opamswitch:true shell)
+          "$(b,opam switch set) sets the default switch globally. The shell \
+           hook, when enabled, synchronises the current shell session with \
+           this switch, unless the current directory is a local switch, when \
+           that local switch is used instead. You can always use $(i,%s) to \
+           specify the switch explicitly, which overrides the shell hook."
+          OpamEnv.(
+            shell_eval_invocation shell
+              (opam_env_invocation ~switch:"SWITCH" ~set_opamswitch:true shell)
             |> Manpage.escape));
   ] @
     mk_subdoc ~cli ~defaults:["","list";"SWITCH","set"]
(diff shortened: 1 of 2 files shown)


src/client/opamCommands.ml
```diff
@@ -2549,12 +2549,13 @@ let switch cli =
...
-         "$(b,opam switch set) sets the default switch globally, but it is also \
-         possible to select a switch in a given shell session, using the \
-         environment. For that, use $(i,%s)."
-        OpamEnv.(
-          shell_eval_invocation shell
-            (opam_env_invocation ~switch:"SWITCH" ~set_opamswitch:true shell)
+          "$(b,opam switch set) sets the default switch globally. If shell \
+           hook is enabled, current shell session is always synchronised with \
+           selected switch (or automatically selected local switch). \
+           Otherwise, you can use $(i,%s)."
```
kit-ty-kate, line 2555: Something like that to be more explicit (not sure how to turn the phrase)
```suggestion
           Otherwise, you can use $(i,%s) to force the current switch to be the one set (overrides local switches)."
```

src/client/opamCommands.ml
```diff
@@ -2549,12 +2549,13 @@ let switch cli =
...
-        OpamEnv.(
-          shell_eval_invocation shell
-            (opam_env_invocation ~switch:"SWITCH" ~set_opamswitch:true shell)
+          "$(b,opam switch set) sets the default switch globally. If shell \
+           hook is enabled, current shell session is always synchronised with \
+           selected switch (or automatically selected local switch). \
+           Otherwise, you can use $(i,%s) to force the current switch to be the one set (overrides local switches)."
+          OpamEnv.(
+            shell_eval_invocation shell
+              (opam_env_invocation ~switch:"SWITCH" ~set_opamswitch:true shell)
```
dra27, line 2558: Possibly? Not sure!

```suggestion
          "$(b,opam switch set) sets the default switch globally. The shell hook, when \
           enabled, synchronises the current shell session with this switch, unless \
           the current directory is a local switch, when that local switch is used instead. \
           You can always use $(i,%s) to specify the switch explicitly, which overrides the shell hook."
          OpamEnv.(
            shell_eval_invocation shell
              (opam_env_invocation ~switch:"SWITCH" ~set_opamswitch:true shell)
```

dra27 (commented):
Possible improvement (not sure!)

PR #5321, opened 2022-10-18
Make archive-mirrors modifiable via opam config

### LINKED ISSUE #5316: question / feature request: an opam command to set `archive-mirrors` field
Of course, I can manually edit `~/.opam/config`, but it would be great (avoiding typos, etc.) if I could use `opam config set jobs 10` (or `opam config set archive-mirrors http://my-local-mirror`) and also `unset` these again.

As far as I can tell, the `opam config set` and `opam config set-global` are only about variables, not about the data stored in the config file.

hannesm:
fixes #5316

### DIFF
--- a/src/client/opamConfigCommand.ml
+++ b/src/client/opamConfigCommand.ml
@@ -673,6 +673,19 @@ let global_allowed_fields, global_allowed_sections =
         "jobs", Atomic,
         Config.with_jobs_opt
           (InitConfig.jobs in_config ++ Config.jobs Config.empty);
+        "archive-mirrors", Modifiable (
+          (fun nc c ->
+             Config.with_dl_cache (Config.dl_cache nc @ Config.dl_cache c) c),
+          (fun nc c ->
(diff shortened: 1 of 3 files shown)


src/client/opamConfigCommand.ml
```diff
@@ -673,6 +673,9 @@ let global_allowed_fields, global_allowed_sections =
         "jobs", Atomic,
         Config.with_jobs_opt
           (InitConfig.jobs in_config ++ Config.jobs Config.empty);
+        "archive-mirrors", Atomic,
```
rjbou, line 676: This field is a list of urls, it should be `Modifiable` (appending urls) not `Atomic` (replacement)

rjbou (changes requested):

rjbou (approved):
Updated to make it an extendable field + add some test

hannesm:
thanks :)

kit-ty-kate:
Thanks!

PR #5325, opened 2022-10-21
Allow opam pin remove to take a package as argument

kit-ty-kate:
Reasoning:

`opam pin list` returns something of the form:
```
<pkg>.<version>    <kind>  <target>
```
However users often simply expect that selecting the first column (`<pkg>.<version>`), done very simply by double-clicking on the name on the terminal, to unpin the package would work. However doing this will end up in a rather surprising failure:
```
[ERROR] No package pinned to this target found, or invalid package name/url: <pkg>.<version>
```

This PR removes this surprising restriction

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -3036,8 +3036,8 @@ let pin ?(unpin_only=false) cli =
      For source pinnings, the package version may be specified by using the \
      format $(i,NAME).$(i,VERSION) for $(i,PACKAGE), in the source opam file, \
      or with $(b,edit).";
-    cli_original, "remove", `remove, ["NAMES...|TARGET"],
-    "Unpins packages $(i,NAMES), restoring their definition from the \
+    cli_original, "remove", `remove, ["PACKAGES...|TARGET"],
+    "Unpins packages $(i,PACKAGES), restoring their definition from the \
      repository, if any. With a $(i,TARGET), unpins everything that is \
      currently pinned to that target.";
     cli_original, "edit", `edit, ["NAME"],
@@ -3363,8 +3363,15 @@ let pin ?(unpin_only=false) cli =
             match as_url with
             | Some ((_::_) as url) -> err, url @ acc
             | _->
-              match (fst package_name) arg with
-              | `Ok name -> err, name::acc
+              match (fst package) arg with
+              | `Ok (name, None) -> err, name::acc
+              | `Ok (name, Some version) ->
+                (match OpamPinned.version_opt st name with
+                 | Some v when not (OpamPackage.Version.equal v version) ->
+                   OpamConsole.error "%s is pinned but not to version %s. Skipping."
+                     (OpamPackage.Name.to_string name) (OpamPackage.Version.to_string version);
+                   true, acc
+                 | Some _ | None -> err, name::acc)
               | `Error _ ->
                 OpamConsole.error
                   "No package pinned to this target found, or invalid package \
--- a/src/state/opamPinned.mli
+++ b/src/state/opamPinned.mli
@@ -18,6 +18,10 @@ open OpamStateTypes
     @raise Not_found when appropriate *)
 val version: 'a switch_state -> name -> version
 
+(** If the package is pinned, returns its version.
+    Otherwise returns [None]. *)
+val version_opt: 'a switch_state -> name -> version option
+
 (** Returns the package with the pinned-to version from a pinned package name.
     @raise Not_found when appropriate *)
 val package: 'a switch_state -> name -> package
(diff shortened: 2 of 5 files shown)


src/client/opamPinCommand.ml
```diff
@@ -703,16 +703,26 @@ let unpin_one st nv =
...
-      match OpamPinned.package_opt st name with
+      let nv = match OpamPinned.package_opt st name, version with
+        | Some nv, Some v when OpamPackage.Version.equal (OpamPackage.version nv) v -> Some nv
+        | Some _, Some v ->
+          OpamConsole.error_and_exit `Bad_arguments "%s is pinned but not to version %s."
+            (OpamPackage.Name.to_string name) (OpamPackage.Version.to_string v)
+        | Some nv, None -> Some nv
+        | None, _ ->
+          OpamConsole.note "%s is not pinned." (OpamPackage.Name.to_string name);
+          None
```
rjbou, line 723: Is there a reason to not have this check in `OpamCommands.pin` directly ?

kit-ty-kate: I don't know, it was there before. I didn't change this part of the code

kit-ty-kate: I move the note back to the original position to make it clearer.

rjbou: Maybe i selected too much lines :) I wasn't referring to that note, but to the version check (and its consequence to update the api).

kit-ty-kate: ah i see. Mmh, I'll have a look to see if it's easier this way.

rjbou: Plus it will permit to filter packages (names) to unpin, removing ones with wrong version instead of failing if only have a wrong version.

kit-ty-kate: Yep. I don't know why i did it this way in the first place but your suggestion is indeed much simpler and better. This is now done

src/client/opamCommands.ml
```diff
@@ -3363,8 +3363,17 @@ let pin ?(unpin_only=false) cli =
             match as_url with
             | Some ((_::_) as url) -> err, url @ acc
             | _->
-              match (fst package_name) arg with
-              | `Ok name -> err, name::acc
+              match (fst package) arg with
+              | `Ok (name, None) -> err, name::acc
+              | `Ok (name, Some version) ->
+                (match OpamPinned.package_opt st name with
```
kit-ty-kate, line 3369: Actually this can be simplified with `OpamPinned.version`. Gimme a sec.

kit-ty-kate: Done.

rjbou (approved):

src/client/opamCommands.ml
```diff
@@ -3363,8 +3363,15 @@ let pin ?(unpin_only=false) cli =
...
-              | `Ok name -> err, name::acc
+              match (fst package) arg with
+              | `Ok (name, None) -> err, name::acc
+              | `Ok (name, Some version) ->
+                (match OpamPinned.version_opt st name with
+                 | Some v when not (OpamPackage.Version.equal v version) ->
+                   OpamConsole.error "%s is pinned but not to version %s."
+                     (OpamPackage.Name.to_string name) (OpamPackage.Version.to_string version);
+                   true, acc
+                 | Some _ | None -> err, name::acc)
```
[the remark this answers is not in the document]
kit-ty-kate: yes, otherwise it would simply ignore the argument without saying anything.

src/client/opamCommands.ml
```diff
@@ -3363,8 +3363,15 @@ let pin ?(unpin_only=false) cli =
...
             | Some ((_::_) as url) -> err, url @ acc
             | _->
-              match (fst package_name) arg with
-              | `Ok name -> err, name::acc
+              match (fst package) arg with
+              | `Ok (name, None) -> err, name::acc
+              | `Ok (name, Some version) ->
+                (match OpamPinned.version_opt st name with
+                 | Some v when not (OpamPackage.Version.equal v version) ->
+                   OpamConsole.error "%s is pinned but not to version %s."
```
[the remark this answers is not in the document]
kit-ty-kate: Done. If you need more to be done in the manpage feel free to add a suggestion there

src/client/opamCommands.ml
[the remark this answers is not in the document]
rjbou: or says that the package is not pinned so skipping it?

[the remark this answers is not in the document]
kit-ty-kate: It's already done later in the underlying function:
```
opam pin remove test
[NOTE] test is not pinned.
```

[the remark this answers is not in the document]
rjbou: fair enough

src/state/opamPinned.mli
```diff
@@ -18,6 +18,10 @@ open OpamStateTypes
     @raise Not_found when appropriate *)
 val version: 'a switch_state -> name -> version
 
+(** Returns the version the package is pinned to, if pinned.
+    Returns [None] otherwise. *)
```
rjbou, line 22:
```suggestion
(** If pinned, returns the version the package is pinned to, [None] otherwise. *)
```

something like that ?

PR #5327, opened 2022-10-25
reftests: updates on init test

rjbou:
Update it to be no repo + add erroneous case where invariant got removed from install actions

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -119,6 +119,7 @@ users)
   * [BUG]: head -c is not posix compliant. Use cut -b instead. [#5989 @madroach]
   * Add bad cudf package name encoding (dose3 lib) [#6055 @rjbou]
   * Add test for filter operators in opam file [#5642 @rjbou]
+  * Update init test to make it no repo [#5327 @rjbou]
 
 ### Engine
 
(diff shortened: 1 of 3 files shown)


dra27:
This is likely to be a pain, so feel free to object! I like the switch to N0REP0 because it's faster, but might it be worth using the same package names and versions? i.e. instead of comp.2 call it ocaml.4.02.3 and comp.3 be ocaml.4.05.0, etc.?

I'm being slow, but I don't understand the second commit - is the point that we're ensuring that without the conflict-class that we _do_ incorrectly install syscomp.2, or is it that there a bug which needs fixing?

rjbou:
> I'm being slow, but I don't understand the second commit - is the point that we're ensuring that without the conflict-class that we _do_ incorrectly install syscomp.2, or is it that there a bug which needs fixing?

My mistake, it was me wrongly defining the packages. I wanted to highlight a wrong behaviour of installing a package that doesn't comply invariant, but `comp` dependencies (post especially) wasn't well defined.

rjbou:
Updated with ocaml-* package names

kit-ty-kate (approved):

PR #5329, opened 2022-10-27
Reftest: `opam list` add an usecase with faulty dependencies computation

rjbou:


tests/reftests/list.test
```diff
@@ -355,3 +355,61 @@ The following actions would be performed:
...
+<><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
+[norepo] synchronised from file://${BASEDIR}/REPO
+Now run 'opam upgrade' to apply any package updates.
+### opam list --required-by top --recursive
+# Packages matching: rec-required-by(top) & (installed | available)
+# Name       # Installed # Synopsis
+choice       --
+other-choice --
+top          --
+unrelated    --
```
kit-ty-kate, line 415: `shouldnt` should be part of the list here right? (but not previously obviously)

rjbou: yes, forgot to add it to the repo :)

Issue #5124, opened 2022-04-21
depext does not support nixOS


RyanGibb:
When installing packages with opam:
```
[NOTE] External dependency handling not supported for OS family 'nixos'.
       You can disable this check using 'opam option --global depext=false'
```

And system dependencies need to be manually installed.

It would be nice if opam could support nix by interacting with `nix-env` or creating a local `default.nix/shell.nix`

kit-ty-kate:
I don’t know much about Nix but as I understand it, it wouldn’t be really useful as the path to C libraries would change all the time and opam/ocaml could store some of them.

We could change the note to encourage nixos users to use https://github.com/timbertson/opam2nix instead.

But maybe my assumptions are very wrong, as I said I don’t know Nix much.
cc @timbertson @rgrinberg in case you have some inputs as Nix users.

RyanGibb:
The paths to the libraries would change, but using `nix-shell` in the directory with `default.nix`/`shell.nix` sets global flags that point to these locations.

To illustrate this with an example, [ocaml-dns](https://github.com/mirage/ocaml-dns) required the `gmp` package. Installing it globally results in it (as expected) not being found as it's in `/nix/store/<hash>-gmp_<version>`.

Creating the file `shell.nix` with contents:
```
with import <nixpkgs> {};
mkShell {
  nativeBuildInputs = [
    ocaml
    opam
    dune_2
    pkg-config
    gcc
    bintools-unwrapped
    gmp
  ];
}
```

And calling `nix-shell` in the working directory sets appropriate flags like
```
% echo $NIX_CFLAGS_COMPILE
[...]

Issue #4440, opened 2020-11-19
depexts: Find the optimal scheme for conf-* packages for maxium coverage and non-overlaping


kit-ty-kate:
# Current state of things in opam-repository

Currently depexts in opam-repository (mostly inside `conf-*` packages) typically look like this:
```
depexts: [
  ["libzstd-dev"] {os-family = "debian"}
  ["libzstd-devel"] {os-distribution = "centos"}
  ["libzstd-devel"] {os-distribution = "rhel"}
  ["libzstd-devel"] {os-distribution = "fedora"}
  ["libzstd-devel"] {os-family = "suse"}
  ["zstd-dev"] {os-distribution = "alpine"}
  ["zstd"] {os-distribution = "homebrew" & os = "macos"}
  ["zstd"] {os = "freebsd"}
]
```
This is problematic if we want to support more than just some specific distributions but also their "derived distributions" (e.g. Manjaro for Archlinux, PopOS for Ubuntu, …)

I've tried using `os-family` instead of `os-distribution` for all of them, as it is what is used in opam internally anyway to detect and launch the commands but for selecting package some issues arise from that approach, mainly:
1) Some of the main distributions are derived from each others. For instance debian and ubuntu
```
  ["libmysqlclient-dev"] {os-distribution = "ubuntu"}
  ["default-libmysqlclient-dev"] {os-distribution = "debian"}
```
In this example requested packages have a different name. However if we were to use `os-family` here it would result in the right behaviour on `ubuntu-derived` distribution, but not on the main ubuntu as the `os-family` variable would pick up `debian`.
The same thing happens with `centos` being derived from `rhel` but the set of packages can vary wildly (especially given the sub-repositories structure – e.g. EPEL, PowerTools, … – in CentOS)
2) To complicate things even further, some distributions might have different names. From https://github.com/ocaml/opam/blob/3ee922d3e621bd0ee7427f33196bd977009131ef/src/state/opamSysInteract.ml for instance, Archlinux can be called `arch` or `archlinux`, Oraclelinux can be called `ol` or `oraclelinux`, OpenSUSE can be called `suse` or `opensuse` (I remember the change of the last one in OpenSUSE 15.x or whatever, for sure)

These things can be encoded using disjunction such as:
```
  ["libmysqlclient-dev"] {os-distribution = "ubuntu" | os-family = "ubuntu"}
  ["default-libmysqlclient-dev"] {os-distribution != "ubuntu" & os-family = "debian"}
```
```
[...]

PR #5332, opened 2022-10-28
Add support for nixos depexts

gridbugs:


src/state/opamSysInteract.ml
```diff
@@ -129,6 +130,7 @@ let family ~env () =
       failwith
         "External dependency handling for macOS requires either \
          MacPorts or Homebrew - neither could be found"
+    | "nixos" -> Nixos
```
avsm, line 133: This does lead to the question of what opam will/should do if running on macOS with nix-darwin (as I do).  I have both Homebrew and nix-darwin installed -- it would be nice to be able to explicitly configure which depext backend to use rather than have it inferred by opam.

RyanGibb: Regarding what it will do, I believe `OpamSysPoll.os_family env` will return [`homebrew`/`macports`](https://github.com/ocaml/opam/blob/3434460674f4df689dc982971ba9c2e96d349902/src/state/opamSysPoll.ml#L94-L97), so the behavior will be unchanged by this PR.

Regarding what it should do, I don't think there's much to be gained from installing depexts on macOS through `nix-env`. My opinion is `nix-env` isn't the 'right' way to use Nix, but is a useful function to provide an interface similar to traditional package managers. As such, this PR is useful in that it provides the path of least resistance to getting a project building on a NixOS system with existing OCaml tooling, as opposed to using one of the various projects for building OCaml projects with Nix.

Regarding explicitly configuring which backend to use, opam is already using homebrew by default and macports if homebrew is unavailable. Using `nix-env` could be useful in providing a third option (with a very minor change), i.e.:
```
else if OpamSystem.resolve_command "nix-env" <> None then Some "nix"
```
Perhaps explicitly configuring which depext backend to use could be another PR.

RyanGibb:
Hi Stephen:-)

It looks like this addresses https://github.com/ocaml/opam/issues/5124.

While I had an idea to generate `shell.nix` files from opam, I think using `nix-env` is more aligned with opam's existing approach, so this seems like a great addition to me.

The only thing is I prefer not to use `nix-env`, instead configuring per-project environments or installing packages globally declaratively. Does opam prompt the user before installing packages by default?

src/state/opamSysInteract.ml
```diff
@@ -734,6 +839,7 @@ let update ?(env=OpamVariable.Map.empty) () =
     | Gentoo -> Some (`AsAdmin "emerge", ["--sync"])
     | Homebrew -> Some (`AsUser "brew", ["update"])
     | Macports -> Some (`AsAdmin "port", ["sync"])
+    | Nixos -> Some (`AsUser "nix-channel", ["--update"])
```
RyanGibb, line 842: This won't do anything if the NixOS system is (like mine) using flakes to provide a instance Nixpkgs instances instead of channels. However, we definitely shouldn't call `nix flake lock --update-input nixpkgs` from opam so I don't have a better suggestion.

A more general point: does this have opam invoking a command to update the entire system? I notice `pacman` is invoked with `-Sy`, which is [not recommended](https://wiki.archlinux.org/title/System_maintenance#Avoid_certain_pacman_commands) as refreshing wityhout updating results in partial upgrades. I think this should be `-Syu`. Unless there's a good reason not too I'll create a new issue for this.

src/state/opamSysInteract.ml
avsm: Thanks Ryan, that all sounds very sensible to me.

src/state/opamSysInteract.ml
kit-ty-kate: This is mostly use for CI systems that really benefit from a non-platform way of updating the system package repository.

> A more general point: does this have opam invoking a command to update the entire system? I notice `pacman` is invoked with `-Sy`, which is [not recommended](https://wiki.archlinux.org/title/System_maintenance#Avoid_certain_pacman_commands) as refreshing wityhout updating results in partial upgrades. I think this should be `-Syu`. Unless there's a good reason not too I'll create a new issue for this.

From opam's point of view this is safe to do because installing a package is done using `pacman -Su`.

gridbugs: > This won't do anything if the NixOS system is (like mine) using flakes to provide a instance Nixpkgs instances instead of channels.

I feel like I'm always learning new ways to use nix.

gridbugs:
> The only thing is I prefer not to use `nix-env`, instead configuring per-project environments or installing packages globally declaratively. Does opam prompt the user before installing packages by default?

Opam doesn't prompt the user before installing packages. I can see how a nix user might be bothered by it invoking `nix-env` by surprise, more so than users of other operating systems. Would it make sense to add a prompt before installing any packages on nixos?

RyanGibb:
> > The only thing is I prefer not to use `nix-env`, instead configuring per-project environments or installing packages globally declaratively. Does opam prompt the user before installing packages by default?
> 
> Opam doesn't prompt the user before installing packages. I can see how a nix user might be bothered by it invoking `nix-env` by surprise, more so than users of other operating systems. Would it make sense to add a prompt before installing any packages on nixos?

I think it would, if there's a clean way to do that without effecting other platforms?

I think I see some logic for prompting the user in https://github.com/gridbugs/opam/blob/3434460674f4df689dc982971ba9c2e96d349902/src/client/opamSolution.ml#L1146-L1278 but I'm not too familiar with this codebase.

src/state/opamSysInteract.ml
```diff
@@ -596,6 +598,108 @@ let packages_status ?(env=OpamVariable.Map.empty) packages =
...
+        |  _ -> None
+      in
+      let get_all_available_package_names () =
+        (* Returns a list containing the names of all available packages. We
+           use this to determine which specified packages are available in the
+           repo. The command for querying whether a single package is available
+           is slow, and takes the same amount of time as listing all the
+           packages (~30s). We need to check the availability of multiple
+           packages, so it's faster to list all packages once and then search
+           the output for the specified packages.
```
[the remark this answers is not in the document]
RyanGibb: Actually, I've just found the `--attr/-A` option of `nix-env` which allows us to install packages by their attr name, a list of which can be obtained from `nix eval nixpkgs\#pkgs --apply builtins.attrNames`. This is also beneficial as the attr name is guaranteed to be globally unique, and some derivations might not have a pname.

An aside: there are `114` instances `nixos` `depext` dependencies in `opam-repository`. Will this also pick up non-`nixos` depexts?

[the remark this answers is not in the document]
RyanGibb: Another option, perhaps the best solution, is just to call `nix search nixpkgs\#<attr-name>` for every package, and install them with `nix-env -iA <attr-name>`. Calling `nix search` for every package is not expensive due to nix expression caching.

Having said all of this, the `nix` command interfaces are unstable, and flakes are still experimental, and adding two nix backends (for flakes and non-flakes) adds additional complexity, so perhaps it's best to leave this as-is for now.

RyanGibb:
> The specific command that caused the crash was `opam monorepo depext -y ...` in the `depext-lockfile` target of the makefile generated by mirage.

If you use `mirage configure -t <target> --no-depext` it will omit the depext logic from the makefile.

> I think our choices are:
> 
>     * keep the behaviour prior to this PR and rely on nixos users to disallow depexts via the global opam config

As @Leonidas-from-XIV says in https://github.com/tarides/opam-monorepo/pull/344 I think we want to minimize reliance on global mutable state. The `mirage` `--no-depext` option might be sufficient for your immediate problem?

>     * have opam invoke `nix-env` to install packages, possibly with a prompt to confirm, but which is overriden by passing `-y`
>     * have opam check what packages need to be installed, and then exit with a message instructing the user to install those packages by one of the many methods available to nix users
> 
> The third option appeals to me because it would let me continue managing packages declaratively, but opam would still be telling me what packages it expects to be available. But it appears that `nix-env --query` doesn't know about packages that were installed by other means (e.g. if I do `nix-shell -p cmake` and then `nix-env --query cmake` it doesn't get picked up). @RyanGibb is there a way to check whether a package is available in the current shell?

Short of just trying to invoke a command (which through conversations with David Allsop I think this is how opam currently checks installation status?) I don't know a good way to do this.

However, a slight twist on the third option: we could also generate a `shell.nix` so all the user has to do is call `nix-shell` (or the flake equivalents `flake.nix` with a `devShell` output and call `nix develop`). I.e., https://github.com/ocaml/opam/issues/5124. This would remove the need to probe package availability. We would just add all depexts to this file, if they're already installed they won't be installed again.

RyanGibb:
Speaking of mirage, I'd also encourage you to try [github.com/ryanGibb/hillingar](https://github.com/ryanGibb/hillingar). It invokes the opam solver, so should 'just work'. An example is here [RyanGibb/mirage-hello-hillingar](https://github.com/RyanGibb/mirage-hello-hillingar).

The only thing is it doesn't pick up depexts in the monorepo [yet](https://github.com/RyanGibb/hillingar/issues/2) so they need to be manually added to the flake.

RyanGibb:
Also, @sternenseemann makes a good point [here[(https://discourse.nixos.org/t/query-all-pnames-in-nixpkgs-with-flakes/22879) that while binary depexts may work with `nix-env`, libraries won't be found by the linker. I think `nix-shell`/`nix develop` will set appropriate flags to remedy this.

rjbou:
Hej everyone! Finally catching up this PR. Thanks all for these highly needed inputs (there is no nix user in opam team). Here's some answer / inputs on the discussion.

> This does lead to the question of what opam will/should do if running on macOS with nix-darwin (as I do). I have both Homebrew and nix-darwin installed -- it would be nice to be able to explicitly configure which depext backend to use rather than have it inferred by opam.

After discussing it in dev meeting, we end up with the conclusion that we should keep that PR for `nixos` distribution only. Nix can be installed and used in several platform, and to do it the good way, we have to implement first `os-package-repository` variable (see #4440) to be able to cleanly and clearly specify the wanted package manager to use and its repository.

> Opam doesn't prompt the user before installing packages

Opam always prompt before calling system package manager, we don't want to install something in user's computer without validation. Even with `--yes` the prompt if shown, for automatisation (CI scripts), there is option  `--confirm-level=unsafe-yes` that runs in a non interactive mode (opam and underlying package manager). You can see an example of prompt in this [test](https://github.com/ocaml/opam/blob/master/tests/reftests/pin.test#L161-L167) (for test, system package manager command is `echo`)

> The command for querying whether a single package is available is slow, and takes the same amount of time as listing all the packages (~30s)

Speed is crucial here, for several commands this installed/availability computation is done. it can't take more then 1-2s at most.

> Having said all of this, the nix command interfaces are unstable, and flakes are still experimental, and adding two nix backends (for flakes and non-flakes) adds additional complexity, so perhaps it's best to keep using nix-env.

> binary depexts may work with nix-env, libraries won't be found by the linker. I think nix-shell/nix develop will set appropriate flags to remedy this.

Better not rely on an experimental feature as no everyone is on opam master branch (nor last release) ; and it's better to have more coverage on packages detection. Which one should be used then?

kit-ty-kate:
> Opam always prompt before calling system package manager, we don't want to install something in user's computer without validation. Even with `--yes` the prompt if shown, for automatisation (CI scripts), there is option `--confirm-level=unsafe-yes` that runs in a non interactive mode (opam and underlying package manager). You can see an example of prompt in this [test](https://github.com/ocaml/opam/blob/master/tests/reftests/pin.test#L161-L167) (for test, system package manager command is `echo`)

While this is true, it is to be noted that we do not show the command that opam is going to run. I think we'd want that here (and that would be an improvement in my opinion)

src/state/opamSysInteract.ml
```diff
@@ -596,6 +598,108 @@ let packages_status ?(env=OpamVariable.Map.empty) packages =
...
+        let args = ["--query"; pname] in
+        match run_command nix_env args with
+        | (0, _) -> true
+        | (1, _) -> false
+        | (code, output) ->
+          Printf.ksprintf failwith
+            "failed with unexpected exit code %d running command:\n    %s\n\nOutput:\n%s"
+            code (String.concat " " (nix_env::args)) (String.concat "\n" output)
+      in
+      let all_pnames_of_nix_package_json : OpamJson.t -> string list option = function
```
rjbou, line 615: Wouldn't be faster to directly parse output (list filtering / regexp) than work on parsed json output ? see `Arch` for example.

src/state/opamSysInteract.ml
```diff
@@ -596,6 +598,108 @@ let packages_status ?(env=OpamVariable.Map.empty) packages =
...
+          OpamSysPkg.Set.elements not_installed
+          |> List.map (fun pkg -> OpamSysPkg.to_string pkg, pkg)
+          |> OpamStd.String.Map.of_list
+        in
+        let available_to_install =
+          get_all_available_package_names ()
+          |> List.filter_map (fun pname -> OpamStd.String.Map.find_opt pname not_installed_by_str)
+          |> OpamSysPkg.Set.of_list
+        in
+        (available_to_install, OpamSysPkg.Set.diff not_installed available_to_install)
```
rjbou, line 702: Internal function [compute_sets](https://github.com/ocaml/opam/blob/master/src/state/opamSysInteract.ml#L152) simplifies this computation.

src/state/opamSysInteract.ml
```diff
@@ -596,6 +598,108 @@ let packages_status ?(env=OpamVariable.Map.empty) packages =
...
+        (* Returns a list containing the names of all available packages. We
+           use this to determine which specified packages are available in the
+           repo. The command for querying whether a single package is available
+           is slow, and takes the same amount of time as listing all the
+           packages (~30s). We need to check the availability of multiple
+           packages, so it's faster to list all packages once and then search
+           the output for the specified packages.
+        *)
+        let args = ["--query"; "--available"; "--json"] in
+        let (code, output_lines) = run_command nix_env args in
```
rjbou, line 668: Is there a reason to not use `run_query_command`?

src/state/opamSysInteract.ml
```diff
@@ -678,6 +782,7 @@ let install_packages_commands_t ?(env=OpamVariable.Map.empty) sys_packages =
     [`AsAdmin "port", yes ["-N"] ("install"::packages)],
     None
   | Netbsd -> [`AsAdmin "pkgin", yes ["-y"] ("install" :: packages)], None
+  | Nixos -> [`AsUser "nix-env", "--install" :: packages], None
```
rjbou, line 785: Does `nix-env` have a non-interactive option?

rjbou:
> > Opam always prompt before calling system package manager, we don't want to install something in user's computer without validation. Even with --yes the prompt if shown, for automatisation (CI scripts), there is option --confirm-level=unsafe-yes that runs in a non interactive mode (opam and underlying package manager). You can see an example of prompt in this [test](https://github.com/ocaml/opam/blob/master/tests/reftests/pin.test#L161-L167) (for test, system package manager command is echo)

> While this is true, it is to be noted that we do not show the command that opam is going to run. I think we'd want that here (and that would be an improvement in my opinion)

True, but the issue here is to acknowledge user that `nix-env` will be used (cf. https://github.com/ocaml/opam/pull/5332#issuecomment-1296471936). Packages are listed just above the menu. If you want to see the command before running it, it's not intuitive, but there is a way, see https://github.com/ocaml/opam/pull/5053#issuecomment-1036354826 (we can discuss depext menu UI in another issue).

RyanGibb:
> After discussing it in dev meeting, we end up with the conclusion that we should keep that PR for nixos distribution only. Nix can be installed and used in several platform, and to do it the good way, we have to implement first os-package-repository variable (see https://github.com/ocaml/opam/issues/4440) to be able to cleanly and clearly specify the wanted package manager to use and its repository.

Makes sense! I do think this nix shell solution (as opposed to the `nix-env` solution) could benefit other platforms (e.g. MacOS) by providing a more reproducible way to provide depexts, but it sounds like this is blocked by some opam limitations.

> Opam always prompt before calling system package manager, we don't want to install something in user's computer without validation. Even with --yes the prompt if shown, for automatisation (CI scripts), there is option --confirm-level=unsafe-yes that runs in a non interactive mode (opam and underlying package manager). You can see an example of prompt in this [test](https://github.com/ocaml/opam/blob/master/tests/reftests/pin.test#L161-L167) (for test, system package manager command is echo)

That's great to hear.

> > The command for querying whether a single package is available is slow, and takes the same amount of time as listing all the packages (~30s)

> Speed is crucial here, for several commands this installed/availability computation is done. it can't take more then 1-2s at most.

This make flake nix expression caching all the more attractive...

> > Having said all of this, the nix command interfaces are unstable, and flakes are still experimental, and adding two nix backends (for flakes and non-flakes) adds additional complexity, so perhaps it's best to keep using nix-env.

> > binary depexts may work with nix-env, libraries won't be found by the linker. I think nix-shell/nix develop will set appropriate flags to remedy this.

> Better not rely on an experimental feature as no everyone is on opam master branch (nor last release) ; and it's better to have more coverage on packages detection. Which one should be used then?

I think we should take an opinionated stance and use flakes given the performance benefits (through expression caching). While they are technically still experimental, they are widely adopted in the Nix ecosystem. We can avoid using the unstable command line APIs if we aren't using the `nix-env` solution. I would be interested to hear any arguments against this, though.

RyanGibb:
> > Opam always prompt before calling system package manager, we don't want to install something in user's computer without validation. Even with `--yes` the prompt if shown, for automatisation (CI scripts), there is option `--confirm-level=unsafe-yes` that runs in a non interactive mode (opam and underlying package manager). You can see an example of prompt in this [test](https://github.com/ocaml/opam/blob/master/tests/reftests/pin.test#L161-L167) (for test, system package manager command is `echo`)
> 
> While this is true, it is to be noted that we do not show the command that opam is going to run. I think we'd want that here (and that would be an improvement in my opinion)

To clarify, what I think @gridbugs is pondering, and what I'm proposing, is creating a file that describes the depexts to be used. See here: https://nixos.wiki/wiki/Development_environment_with_nix-shell. There's many idiomatic arguments for this but the key reason I see for doing it is that library depexts installed with `nix-env` won't be found by the linker.

The only difficulty with this is it requires quite a different model of interacting with the system than traditional package managers, and I don't know much refactoring would be required to implement it.

RyanGibb:
Also, looking at this from a birds eye view, what we're trying to do here is to get existing OCaml tooling working on a NixOS system. However, as we're not building inside a Nix derivation, we don't get the benefits of reproducible builds (in the system state sense - not strict binary compatibility sense). It may be worth checking out one of the number of projects for building OCaml/Opam projects on Nix, notably [tweag/opam-nix](https://github.com/tweag/opam-nix) (which uses the opam solver and should 'just work'), or using [nix-ocaml/nix-overlays](https://github.com/nix-ocaml/nix-overlays) (if you want to be hardcore about it and package your own dependencies in Nix).

Profpatsch:
Any progress on this?

RyanGibb:
> There's many idiomatic arguments for this but the key reason I see for doing it is that library depexts installed with `nix-env` won't be found by the linker.

To expand on this, a real world example I always have issues with is the GNU Multiple Precision Arithmetic Library (`gmp`). Installing it any mechanism (`nix-env` or flakes) doesn't set [`NIX_CFLAGS_COMPILE`](https://nixos.org/manual/nixpkgs/stable/#cc-wrapper), which is only done in a derivation.

A workaround is to use:
`CFLAGS="-isystem ``nix eval --inputs-from /etc/nixos --raw nixpkgs#gmp``" opam reinstall conf-gmp`

kit-ty-kate:
Superseded by https://github.com/ocaml/opam/pull/5982 which was just merged and is going to be part of opam 2.4 (whose first alpha is in the process of being released). Thanks for the proposal and all the discussions

PR #5335, opened 2022-10-31
Improve non UTF8 output

haesbaert:
~~Currently when UTF8 is disabled via OPAMUTF8, `to_aligned_strings` assumes `explicit` and also ignores the ascii symbol representation, example:~~
```
  - recompile carton-git               0.4.4   [uses logs]
  - recompile conduit-lwt-unix         5.1.0   [uses logs]
```
~~While the UTF8 will not ignore explicit and will output:~~
```
  <swirl symbol> carton-git               0.4.4   [uses logs]
  <swirl symbol> conduit-lwt-unix         5.1.0   [uses logs]
```
~~This diff fixes it to output:~~
```
  R carton-git               0.4.4   [uses logs]
  R conduit-lwt-unix         5.1.0   [uses logs]
```

~~Also use a '*' instead of '--' when it's meant~~

EDITED, see https://github.com/ocaml/opam/pull/5335#issuecomment-1297066062


kit-ty-kate:
Based on the screenshots in https://github.com/ocaml/opam/pull/5045 I'm pretty sure this was not a bug but intended code.
Aside from that I personally find the change rather unsettling and would not personally endorse it.
```
-  - upgrade opam-core        2.0.7 to 2.1.0~beta3 (pinned) [required by opam-format]
+  U opam-core        2.0.7 to 2.1.0~beta3 (pinned) [required by opam-format]
 === install 2 packages
-  - install opam-file-format 2.0.0                         [required by opam-format]
-  - install opam-format      2.1.0~beta3 (pinned)
+  * opam-file-format 2.0.0                         [required by opam-format]
+  * opam-format      2.1.0~beta3 (pinned)
```

Past this, could you explain the issue you are trying to fix in your use-case?

dbuenzli:
(Sorry for the slight OT)

@haesbaert Do you use non UTF-8 output for technical reasons (i.e. non UTF-8 aware environment) or because of preference ? 

I'm curious because I generally no longer provide non UTF-8 outputs when I write cli tools. But I don't really have data to backup that choice.

haesbaert:
> (Sorry for the slight OT)
> 
> @haesbaert Do you use non UTF-8 output for technical reasons (i.e. non UTF-8 aware environment) or because of preference ?
> 
> I'm curious because I generally no longer provide non UTF-8 outputs when I write cli tools. But I don't really have data to backup that choice.

Mostly I've been using Terminus (https://terminus-font.sourceforge.net/) for over a decade now and I kinda love it, but it misses some UTF8 glyphs that opam uses. I've been trying to use other fonts, Monoid has been the runner up, but I gave up and went back to Terminus this morning.

kit-ty-kate:
Could you update the tests? (you can do this automatically using `dune build --profile=release --auto-promote @reftest`)
Could you also share a screenshot of the `Error report` section that also uses the same piece of code? Here is what it looks like with UTF-8: https://github.com/ocaml/opam/pull/5045#issuecomment-1032702813

haesbaert:
> Out of curiosity, why do you need that utf8 & non utf8 outputs have the same number of columns?

I apologize, I didn't want to imply we need to have the same number of columns, but I'd argue that it's a question of elegance (the glyphs from an output shouldn't affect the structure of the output), and (there are probably better ways to do it but...):
```
opam remove --color=never --dry-run --no foobar| tail -n +2 |awk '{print $3}'
```

kit-ty-kate:
> Ofc, I'll work on it, was just trying to feel the waters before :)

The water on my side so far is "meh, why bother for 2 characters. Plus it looks less pretty and less organized without colours (which we use in CI)" but i can be convinced otherwise by being shown pretty screenshots and see the effect of the change in the updated tests.

kit-ty-kate:
Closing for now as we find the change unsatisfying. We also make no guarantees on the result of things like:
```
opam remove --color=never --dry-run --no foobar| tail -n +2 |awk '{print $3}'
```
as they are UI elements and subjects to change. Please use `--json` instead for cases where you need to parse that kind of output

Issue #5334, opened 2022-10-31
opam admin add-constraint fails with Not_found


kit-ty-kate:
Using opam master (5fb5c15b2d7d5f813d1363a307ceb9ed58e49ba9), calling `CAMLRUNPARAM=b opam admin add-constraint "ocaml<5.0"` on commit https://github.com/ocaml/opam-repository/commit/6b61f8e16d6ef37ad7c6611977c9dc02755d0280 will raise the following uncaught exception:
```
Fatal error:
Not_found
Backtrace:
  Raised at Stdlib__List.assoc in file "list.ml", line 191, characters 10-25
  Called from OpamFormula.simplify_ineq_formula.(fun) in file "src/format/opamFormula.ml" (inlined), line 558, characters 13-27
  Called from OpamFormula.simplify_ineq_formula.(fun) in file "src/format/opamFormula.ml", line 563, characters 50-62
  Called from OpamFormula.map in file "src/format/opamFormula.ml", line 141, characters 25-34
  Called from OpamFormula.simplify_ineq_formula in file "src/format/opamFormula.ml", line 563, characters 20-66
  Called from OpamFormula.simplify_version_formula in file "src/format/opamFormula.ml" (inlined), line 600, characters 2-53
  Called from OpamFilter.simplify_extended_version_formula in file "src/format/opamFilter.ml", line 627, characters 35-75
  Called from OpamAdminCommand.add_constraint_command.cmd.add_cstr in file "src/client/opamAdminCommand.ml", line 1022, characters 12-58
  Called from OpamAdminCommand.add_constraint_command.cmd.(fun) in file "src/client/opamAdminCommand.ml", line 1043, characters 42-81
  Called from OpamFormula.map in file "src/format/opamFormula.ml", line 141, characters 25-34
  Called from OpamFormula.map in file "src/format/opamFormula.ml", line 141, characters 25-34
  Called from OpamFormula.map in file "src/format/opamFormula.ml", line 141, characters 25-34
  Called from OpamAdminCommand.add_constraint_command.cmd.(fun) in file "src/client/opamAdminCommand.ml", line 1041, characters 10-201
  Called from Stdlib__Map.Make.iter in file "map.ml", line 297, characters 20-25
  Called from Stdlib__Map.Make.iter in file "map.ml", line 297, characters 10-18
  Called from Stdlib__Map.Make.iter in file "map.ml", line 297, characters 10-18
[...]

PR #5336, opened 2022-10-31
Fix opam admin add-constraint failing with Not_found in some situations

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/5334
old man yells at polymorphic compare meme

*one more for the books https://github.com/ocaml/ocaml/pull/9928* :disappointed:

### DIFF
--- a/tests/reftests/dune.inc
+++ b/tests/reftests/dune.inc
@@ -1,5 +1,22 @@
 
 (rule
+ (alias reftest-admin-add-constraint)
+ (action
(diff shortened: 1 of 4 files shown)


tests/reftests/dune.inc
```diff
@@ -1,4 +1,21 @@
 
+(rule
+ (alias reftest-add-constraint)
```
rjbou, line 3:
```suggestion
 (alias reftest-admin-add-constraint)
```

PR #5337, opened 2022-10-31
Move dependencies computation from `opam-solver` to `opam-state`

rjbou:
`OpamSolver.dependencies` and `OpamSolver.reverse_dependencies` are for the moment only used in `opam-client`. The main reason to move them is that #5208 use one of them in `opam-state` (`opamSwitchState`), making it depend on `opam-solver` and all of its dependencies. 
Besides, with the rewrite, `dependencies` is more efficient: for a big request (`opam list --required-by patdiff,memtrace_viewer,tezos,bap,mirage --recursive`, resulting to ~800 unique packages), from 1.95s to 0.67s.

Remaining:
* [x] Add some log
* [x] Comment opam-solver ones?
* [x] Add doc
* [x] Add changes

### DIFF
--- a/src/state/opamSwitchState.ml
+++ b/src/state/opamSwitchState.ml
@@ -1305,3 +1305,136 @@ let update_repositories gt update_fun switch =
   OpamFile.Switch_config.write
     (OpamPath.Switch.switch_config gt.root switch)
     conf
+
+
+(* dependencies computation *)
+
+let dependencies_filter_to_formula_t ~build ~post st nv =
+  let env v =
+    if List.mem v OpamPackageVar.predefined_depends_variables then
+      match OpamVariable.Full.to_string v with
+      | "build" -> Some (B build)
+      | "post" -> Some (B post)
+      | _ -> None
+    else
+      OpamPackageVar.resolve_switch ~package:nv st v
+  in
+  OpamFilter.filter_formula ~default:true env
+
+let dependencies_t base_deps_compute deps_compute
+    ~depopts ~installed ?(unavailable=false) universe packages =
+  if OpamPackage.Set.is_empty packages then OpamPackage.Set.empty else
+  let base =
+    packages ++
+    if installed then  universe.u_installed
+    else if unavailable then universe.u_packages
+    else universe.u_available
+  in
+  log ~level:3 "dependencies packages=%a"
+    (slog OpamPackage.Set.to_string) packages;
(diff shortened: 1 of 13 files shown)


kit-ty-kate:
> Comment opam-solver ones?

I'd be in favour of removing it. It'll always be in git if we want it back

src/state/opamSwitchState.ml
```diff
@@ -1305,3 +1305,169 @@ let update_repositories gt update_fun switch =
...
+           "ERR: Undefined filter variable %s in dependencies of %s")
+        (OpamVariable.Full.to_string v) (OpamPackage.to_string nv);
+    r
+  in
+  OpamFilter.filter_formula ~default:true env formula
+
+let dependencies ~depopts ~build ~post ~installed
+    ?(test=OpamStateConfig.(!r.build_test))
+    ?(doc=OpamStateConfig.(!r.build_doc))
+    ?(dev_setup=OpamStateConfig.(!r.dev_setup))
```
rjbou, line 1341: Do we need all these , `universe` is already computed using them. Here, they are used to specify dependencies formula

src/state/opamSwitchState.ml
```diff
@@ -1305,3 +1305,160 @@ let update_repositories gt update_fun switch =
...
+          packages ++ acc)
+      nvs OpamPackage.Set.empty
+  in
+  let rec aux all deps =
+    let new_deps = get_deps deps in
+    if OpamPackage.Set.is_empty new_deps then all
+    else aux (all ++ new_deps) (new_deps -- all)
+  in
+  let result = aux packages packages in
+  log "Dependencies (%.3f) result=%a" (timer ())
```
rjbou, line 1371: Do we keep that at level 1? With #5208, it will be displayed at each switch state update

kit-ty-kate: Yes i think that's fine. It's debug anyway so if we find it's too much data we can always bump its level up later.

kit-ty-kate (approved):

rjbou:
Some more data on the new implem on ~8 000 runs: l each opam package name, two runs

| | average | min | max |
| --  | -- | -- | -- |
Direct dependencies |  6.73 |  2.36 |  19.95 |
Reverse dependencies |  1.12 |  0.95 | 1.37 |

image
image

kit-ty-kate:
Awesome work. Thanks a lot!

PR #5338, opened 2022-11-03
Fix cudf weak dependencies computation

rjbou:
On cudf strong and weak dependencies computation, some weak dependencies were wrongly kept, originated from #4627
fix test added in #5329


kit-ty-kate (approved):

kit-ty-kate:
Thanks a lot!

PR #5340, opened 2022-11-06
Update IRC links

purplearmadillo77:
Fix documentation to mention IRC channels in liberachat instead of freenode.


rjbou:
Thanks!

PR #5342, opened 2022-11-08
git, hg: Use full revision

reynir:
Before, the revision was at most the first 8 bytes in the revision. This can be problematic for reproducibility since the revision may then be ambiguous. I don't know much about mercurial and how their revisions look like.

CC @hannesm

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -199,6 +199,7 @@ users)
   * Use 4.08's unnamed functor arguments to silence warning 67 [#4775 @dra27]
   * git: disable colored output [#4884 @rjbou]
   * Check if a source is up to date with subpath [#4876 @rjbou]
+  * git, hg: Use the full SHA1 revision instead of just the 8 first characters [#5342 @reynir]
 
 ## Build
   * Bump src_exts and fix build compat with Dune 2.9.0 [#4752 @dra27]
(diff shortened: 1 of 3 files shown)


hannesm:
I completely agree with this PR. I do not understand the reason for using only the first 8 characters.

For reference, the code was introduced nearly 10 years ago and never changed https://github.com/ocaml/opam/commit/fb8fece2b0b5603accb523f664e98d5c174ac154

For reference, `dune` uses 7 characters (since 3.0.0 https://github.com/ocaml/dune/commit/6b69c61040f9c298166fe3c2e4e29109e5b1834e, also https://github.com/ocaml/dune/issues/4855).

I understand the desire to have a stable output (i.e. a fixed length), but then I also don't quite understand why the conditional may ever hit the case `<= 8` characters.

hannesm:
Please let me note that git increased at some point from using 6 characters to 7 (in git describe), and with more sophisticated attacks on SHA1, I don't see any value in truncating the output.

kit-ty-kate (approved):

rjbou:
Agree on the idea, it needs some testing for mercurial part

hannesm:
I'm wondering what the status of this PR is and what can be done to get it merged? Since three opam developers are supporting this PR, I propose to merge it rather sooner than being delayed for unclear reasons ("it needs some testing for mercurial part" -- so who's using mercurial and opam these days? If this is a showstopper, maybe revert the changes to opamHg.ml?)

rjbou (approved):

kit-ty-kate:
I've just tested locally using mercurial and it works fine (the default hash size for mercurial is 12)

Thanks a lot!

PR #5351, opened 2022-11-14
Fix cleaning up switch on error on Windows

dra27:
If switch creation fails and the user elects to clean-up the switch, there is at present an error message displayed when the switch is erased. This is caused by the write lock on the switch still being held. The fix is to drop the switch lock prior to deletion - the global lock is still held.


rjbou (approved):
Maybe worth adding a comment on why this is done this way (switch lock released before cleaning). otherwise lgtm

dra27:
TODO note - it's possible that this can be done while taking the write lock, but using `O_SHARE_DELETE` to permit the file to be removed, even though it's still open (@dra27 will check).

dra27:
The approach in #5435 is better than this one (since the lock is then held right up to the deletion)

PR #5352, opened 2022-11-15
Expand variable interpolations in `build-env`

dra27:
Variable interpolations (`%{lib}%`, etc.) are expanded in `setenv` but weren't expanded in `build-env`.

I'd be willing to regard this as having been a bug, and don't expect that we'll find any instances of `build-env` being used where `%{foo}%` was supposed to be treated as a literal string (with this PR, such strange usage should now be written `%%{foo}%%`).

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -166,6 +166,7 @@ users)
   * Add `with-tools` variable for recommended tools [#5016 @rjbou]
   * Add `x-locked` extension fields for overlay internal use, it stores if the files originate from a locked file, if so its extension [#5080 @rjbou]
   * Set `depext-bypass` parsing with depth 1, no needed brakcet if single package [#5154 @rjbou]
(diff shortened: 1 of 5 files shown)


kit-ty-kate:
What would be the backward compatibility story with this change? Could we maybe detect that the package contains `available: opam-version >= "2.2"` when `build-env` contains variables?

dra27:
Hah, I was just preparing a long justification for why breaking backwards compatibility might be OK... but then I discovered that this is actually fixing a 1.2.2 -> 2.0 **regression**! `build-env` in OPAM 1.2.2 not only expanded variables, but the feature is also in use (and presently broken, therefore) in opam-repository - see sibylfs-lem.0.4.0!

rjbou (approved):
LGTM, missing api changelog & test

PR #5356, opened 2022-11-16
Fix case-preserving environment updates on Windows

dra27:
There's existing code (of mine) which tries to deal with the fact that environment variable names are case insensitive on Windows. In `OpamStd.Env`, this is dealt with largely correctly in `get`, but the handling in `OpamEnv` is not. When dealing with environment _updates_, it's necessary to transform the names of any updates - i.e. `PATH += "foo"` needs to be transformed to `Path += "foo"` to use the same case as the base environment.

The problem is manifesting itself principally with `PATH`, which by default is actually `Path` on Windows. When reverting the updates in `OpamEnv.get_pure` (for example, to evaluate `eval_variables`), the switch's `bin` directory is correctly ignored when resolving the command, but when the command is executed, it re-emerges, because of a duplicate instruction in the env block for both `Path` (with the switch directory, as that comes from the calling environment) and `PATH` (as determined in `OpamEnv.get_pure`, which reverts the updates previously made).

In particular, with this branch, `opam config list` stops reporting the _switch's_ compiler settings in `sys-ocaml-arch`, `sys-ocaml-cc`, and `sys-ocaml-libc`, but I still need to mull some more whether this is strictly the correct approach (and the environment update code is already nightmarishly complicated even on Unix...)

### DIFF
--- a/src/core/opamStd.mli
+++ b/src/core/opamStd.mli
@@ -413,11 +413,21 @@ module Env : sig
 
   (** {3 Environment variable handling} *)
 
+  (** Environment variable names *)
+  module Name : sig
+    include ABSTRACT with type t = private string
+
+    val equal_string: t -> string -> bool
+
+  end
+
   val get: string -> string
 
   val getopt: string -> string option
 
-  val list: unit -> (string * string) list
+  val getopt_full: Name.t -> Name.t * string option
+
+  val list: unit -> (Name.t * string) list
 end
 
 (** {2 System query and exit handling} *)
--- a/src/core/opamStd.mli
+++ b/src/core/opamStd.mli
@@ -386,6 +386,9 @@ end
 (** {2 Manipulation and query of environment variables} *)
 
 module Env : sig
+
+  (** {3 Generic functions} *)
+
   (** Remove from a c-separated list of string the ones with the given prefix *)
   val reset_value: prefix:string -> char -> string -> string list
 
@@ -395,12 +398,6 @@ module Env : sig
       other elements with the same [prefix] they are kept in the second list.
   *)
   val cut_value: prefix:string -> char -> string -> string list * string list
-
-  val get: string -> string
-
-  val getopt: string -> string option
-
-  val list: unit -> (string * string) list
 
   (** Utility function for shell single-quoted strings. In most shells,
       backslash escapes are not allowed and a single quote needs to be replaced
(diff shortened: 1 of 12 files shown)


rjbou:
queued on #5374

src/core/opamStd.mli
```diff
@@ -413,6 +410,30 @@ module Env : sig
 
   (** Utility function for PowerShell strings. *)
   val escape_powershell: string -> string
+
+  (** {3 Environment variable handling} *)
+
+  (** Environment variable names *)
```
rjbou, line 416: worth add some documentation to explain the purpose of the module, ie the need for windows to have specific comparison

src/core/opamStd.mli
```diff
@@ -413,6 +410,30 @@ module Env : sig
...
+  module Envname : sig
```
rjbou, line 417: `s/Envname/Name`

src/core/opamStd.mli
```diff
@@ -413,6 +410,30 @@ module Env : sig
...
+    type t = private string
+
+    val of_string: string -> t
+
+    val compare: t -> t -> int
+    val equal: t -> t -> bool
+    val equal_string: t -> string -> bool
+
+    module Set : SET with type elt = t
+    module Map : MAP with type key = t
```
rjbou, line 427: Use module type  `OpamStd.ABSTRACT` ?
```suggestion
  module Name: OpamSTd.ABSTRACT with type t = private string
```

src/state/opamEnv.ml
```diff
@@ -177,16 +198,18 @@ let expand (updates: env_update list) : env =
...
-      List.filter (fun (k,_,_) -> let k = String.uppercase_ascii k in List.for_all (fun (u,_,_,_) -> u <> k) updates) env
+      (* Preserve the case of updates which are already in env *)
+      map_update_names (OpamStd.Env.Envname.Set.of_list (List.map (fun (k, _, _) -> k) env)) updates
     else
-      List.filter (fun (k,_,_) -> List.for_all (fun (u,_,_,_) -> u <> k) updates)
-        env
+      updates
+  in
+  let update_keys =
+    List.fold_left (fun m (k,_,_,_) -> OpamStd.Env.Envname.(Set.add (of_string k) m)) OpamStd.Env.Envname.Set.empty updates
```
rjbou, line 209:
```suggestion
    List.fold_left (fun m (k,_,_,_) -> 
      OpamStd.Env.Envname.(Set.add (of_string k) m)) 
        OpamStd.Env.Envname.Set.empty updates
```

rjbou (approved):
lgtm! just some cosmetic changes.

rjbou:
added a test

src/core/opamStd.mli
dra27: Done!

PR #5364, opened 2022-11-22
Fix trailing space on header messages

rjbou:


dra27 (approved):
It feels necessary to cite https://dilbert.com/strip/1996-08-20 in approving this! 🙂

PR #5372, opened 2022-11-29
Check for available packages with homebrew

gridbugs:
Signed-off-by: Stephen Sherratt <stephen@sherra.tt>

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -67,6 +67,7 @@ users)
     * ◈ Rename --with-tools` to `--with-dev-setup` [#5214 @rjbou - fix #4959]
   * Use the default criteria during reinstall/upgrade when requesting at least one non-installed package [#5228 @kit-ty-kate]
   * Show the reason for installing packages when using opam reinstall [#5229 @kit-ty-kate]
+  * Check for available packages with homebrew [#5372 @gridbugs]
 
 ## Remove
   *
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -502,7 +502,28 @@ let packages_status ?(env=OpamVariable.Map.empty) packages =
       |> List.map OpamSysPkg.of_string
       |> OpamSysPkg.Set.of_list
     in
-    compute_sets sys_installed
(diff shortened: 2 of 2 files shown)


kit-ty-kate:
Does it take aliases into account? The reason why https://github.com/ocaml/opam/pull/4548 didn't get merged is because i couldn't find any ways of detecting aliases properly and quickly enough.

gridbugs:
> Thanks for the PR! Do you have average time for the new command?

It takes about 3 seconds for me. It uses a single invocation of `brew search '/^(pkg1|pkg2|...)$/'

gridbugs:
> Does it take aliases into account? The reason why #4548 didn't get merged is because i couldn't find any ways of detecting aliases properly and quickly enough.

Yep it handles aliases. I tested with sqlite3 which doesn't show up in the output of `brew formulae`.
```
$ brew search --formula '/^sqlite3$/'
==> Formulae
sqlite3
```
and
```
$ dune exec ./src/client/opamMain.exe -- install --depext-only --dry-run /tmp/a.opam 2> /dev/null

    neovim sqlite3 tmux

<><> Handling external dependencies <><><><><><><><><><><><><><><><><><><><>  🐫

$ cat /tmp/a.opam
opam-version: "2.0"
maintainer: "foo"
description: "bar"
depends: [
]
depexts: [
  "sqlite3"
  "tmux"
  "neovim"
]
```

kit-ty-kate:
> It takes about 3 seconds for me. It uses a single invocation of `brew search --formula '/^(pkg1|pkg2|...)$/'

3 seconds sounds like quite a lot. We're usually aiming for something < 1 second in the worst case.
This command is ran during many opam commands such as `opam show` which would be a dealbreaker for many users if it was that slow.

kit-ty-kate:
Now that https://github.com/ocaml/opam/pull/6489 has been merged in 2.6.0, this PR could be considered. I've added it to the 2.7 milestone and we'll see how it goes. @gridbugs feel free to rebase it if you have sometime, otherwise we'll do it at some point

PR #5374, opened 2022-12-01
Add and use `List` function with comparison functions

rjbou:
To avoid using polymorphic ones.
This PR is the first of more cleaning in code about polymorphic comparison usage

### DIFF
--- a/src/format/opamFile.ml
+++ b/src/format/opamFile.ml
@@ -918,7 +918,7 @@ module Syntax = struct
             | Variable (name, v) ->
               let name = name.pelem in
               (try
-                 let ppa = List.assoc name fields in
+                 let ppa = OpamStd.List.assoc String.equal name fields in
                  match snd (Pp.print ppa t) with
                  | None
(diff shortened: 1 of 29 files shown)


src/format/opamFile.ml
```diff
@@ -953,12 +953,15 @@ module Syntax = struct
...
-                 let ppa = List.assoc section_kind sections in
+                 let ppa =
+                   OpamStd.List.assoc String.equal section_kind sections
+                 in
                  let print_sec ppa t =
                    match snd (Pp.print ppa t) with
                    | None -> None
                    | Some v ->
-                     try Some (List.assoc section_name v) with Not_found -> None
+                     OpamStd.List.assoc_opt (Option.equal String.equal)
```
rjbou, line 963:
```suggestion
                     OpamStd.List.assoc_opt (OpamStd.Option.equal String.equal)
```

dra27 (approved):
Monomorphic joy! LGTM

PR #5380, opened 2022-12-06
Reftests: factorise some repo hashes

rjbou:


kit-ty-kate:
Have you checked that those two tests still have the same behaviours if this patch was applied before their respective fixes?

rjbou:
`conflict-4373` has no core modifications, and `list` just ec6f65e005dc, and changing repo hash doesn't change the output.

rjbou:
#5414 used in another test a removed hash in this PR. Need to be reviewed

kit-ty-kate:
I’m not sure it’s worth potentially breaking carefully crafted tests just to save 3 secondes in total test time

rjbou:
What is broken ?

kit-ty-kate:
I don’t know, I saying potentially. When writing tests there are many factors that appears when choosing a repository hash, many of which might not appear in the results before or after the commit that added the test and its fix.

To me it’s not worth the hassle.

rjbou:
In the current state of the PR, I checked that it is the same before and after fix, when there is an after

kit-ty-kate:
> many of which might not appear in the results before or after the commit

as i said, many invariants do not appear in the results before or after

rjbou:
oups, i misread. Then what are the other things that you check when choosing an hash, except that it permits you to have the "good" test output ?

Issue #5373, opened 2022-12-01
opam init cannot find tools


kolmodin:
I'm trying to run `opam init` for the very first time.

```
$ opam init
No configuration file found, using built-in defaults.
Checking for available remotes: none.
  - you won't be able to use rsync and local repositories unless you install the rsync command on your system.
  - you won't be able to use git repositories unless you install the git command on your system.
  - you won't be able to use mercurial repositories unless you install the hg command on your system.
  - you won't be able to use darcs repositories unless you install the darcs command on your system.

[WARNING] Recommended dependencies -- most packages rely on these:
  - make
  - cc
[ERROR] Missing dependencies -- the following commands are required for opam to operate:
  - curl or wget: A download tool is required, check env variables OPAMCURL or OPAMFETCH
  - diff
  - patch
  - tar
  - unzip
  - getconf
  - bwrap: Sandboxing tool bwrap was not found. You should install 'bubblewrap'. See https://opam.ocaml.org/doc/FAQ.html#Why-does-opam-require-bwrap.
```

However, I do have most (or all) of those tools.
Using `strace` I see that `opam` finds the corrcect path, but seems to reject it and continue the search.

```
$ strace opam init
...
statx(AT_FDCWD, "/usr/local/sbin/curl", AT_STATX_SYNC_AS_STAT, STATX_BASIC_STATS, 0xffefaee0) = -1 ENOENT (No such file or directory)
statx(AT_FDCWD, "/usr/local/bin/curl", AT_STATX_SYNC_AS_STAT, STATX_BASIC_STATS, 0xffefaee0) = -1 ENOENT (No such file or directory)
statx(AT_FDCWD, "/usr/sbin/curl", AT_STATX_SYNC_AS_STAT, STATX_BASIC_STATS, 0xffefaee0) = -1 ENOENT (No such file or directory)
statx(AT_FDCWD, "/usr/bin/curl", AT_STATX_SYNC_AS_STAT, STATX_BASIC_STATS, {stx_mask=STATX_BASIC_STATS|STATX_MNT_ID, stx_attributes=0, stx_mode=S_IFREG|0755, stx_size=268504, ...}) = 0
statx(AT_FDCWD, "/sbin/curl", AT_STATX_SYNC_AS_STAT, STATX_BASIC_STATS, 0xffefaee0) = -1 ENOENT (No such file or directory)
[...]

PR #5381, opened 2022-12-06
Fix opam unable to find executables on systems where users belong to more than 32 groups when opam is built using musl libc

kit-ty-kate:
Details of the issue in musl described in https://www.openwall.com/lists/musl/2021/07/03/1
Fixes https://github.com/ocaml/opam/issues/5373


dra27:
I think there's a couple of bits in `configure.ac` from https://github.com/ocaml/opam/pull/4265/commits/5377b60d1df9f6dd2c42b5a386630fcc745271b4#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810 which also need reverting.

We should double-check that `Unix.access` doesn't regress either https://github.com/ocaml/opam/pull/4265/commits/d6faa8d45a3b06270863fb1c8d0848d7b45c6f36 or https://github.com/ocaml/opam/pull/4265/commits/6f756efc48da31e917a4a36f4c557b3e155d076b (this is probably just a case of checking the Open Group POSIX docs... I'll be _amazed_ if it regresses the first and quite surprised if it regresses the second!). It would be nice to check that the original bug being fixed in #4265, but that may be too tricky to set-up again.

dra27:
Annoying, `Unix.access` checks `ruid` and `rgid` rather than `euid` and `egid` (which is what `exec` will do). Ideally we therefore want to use `faccessat` with `AT_EACCESS` instead, which should be exactly equivalent to the present check, but without needing the libacl hackery for Cygwin (and, in fairness, any other Unix system which chooses to grant permissions with ACLs instead). However, that would cause `opam` to require C stubs for all builds, and I'm not so sure about that :shrug:

dra27:
Discussed today: we decided we're not worried about requiring stubs in opam-core for all platforms, so wrapping [`faccessat`](https://pubs.opengroup.org/onlinepubs/9699919799.2008edition/functions/faccessat.html) should be fine, which we can then use instead of `Unix.access`. Happy for either of us to get to this - either way, we get rid of the libacl bindings!

master_changes.md
```diff
@@ -157,6 +157,7 @@ users)
   * Reduce allocations in OpamVersionCompare [#6144 @talex5]
   * Speedup OpamVersionCompare by 25% by removing the unused handling of epoch [#5518 @kit-ty-kate]
   * Fix error in `OpamSystem.transform_patch` - patches were only applied when debugging [#6182 @dra27 regression since #3449]
+  * Remove --with-acl option (reverts a Cygwin fix in #4265) [#5381 @kit-ty-kate - fix #5373]
```
rjbou, line 160:
```suggestion
  * Remove `--with-acl` option from configure and its related C stubs(reverts a Cygwin fix in #4265) [#5381 @kit-ty-kate - fix #5373]
```

rjbou: + update of API changes 
```
  * `OpamACL`: remove module
```

src/stubs/common/opamCommonStubs.mli
```diff
@@ -1 +0,0 @@
-external check_executable : string -> unit = "opam_check_executable"
```
rjbou: These files were added in previous commit, should we squash this commit with 4e9b31230e7db1973407c35b772f02d947dc98d3 ?

src/core/opamStubsTypes.ml
```diff
@@ -114,7 +114,7 @@ type win32_version_info = {
     (** Non-fixed string table. First field is a pair of Language and Codepage ID. *)
 }
 
-external check_executable : string -> unit = "opam_check_executable"
+external is_executable : string -> bool = "opam_is_executable"
```
rjbou, line 117: `check_executable` was added in this PR, is it possible to have a single commit that adds it as `is_executable` ?

src/core/opamCommonStubs.c
```diff
@@ -1,15 +1,10 @@
 /**************************************************************************/
 /*                                                                        */
-/*                                 OCaml                                  */
+/*    Copyright 2024 Kate Deplaix                                         */
```
rjbou, line 3: Same, it should be squashed

rjbou (commented):
Some format remarks : commits to squash, etc.
Also, commit messages are not very explicit, an update with more information could help.

master_changes.md
```diff
@@ -272,4 +272,5 @@ users)
   * `OpamHash`: export `compare_kind` [#5561 @rjbou]
   * `OpamFilename`: add `might_escape` to check if a path is escapable, ie contains `<sep>..<sep>` [#5561 @rjbou]
   * Add `OpamStd.Sys.getconf` [#5950 @kit-ty-kate]
+  * Fix opam unable to find executables on systems where users belong to more than 32 groups when opam is built using musl libc [#5381 @kit-ty-kate - fix #5373]
```
rjbou, line 277:
```suggestion
  * `OpamStd.Sys.resolve_command`: Fix opam unable to find executables on systems where users belong to more than 32 groups when opam is built using musl libc [#5381 @kit-ty-kate - fix #5373]

rjbou (commented):
Does `doc/index.html` need to be updated by the last commit (or others) ?

kit-ty-kate:
> Does `doc/index.html` need to be updated by the last commit

done. Feel free to squash if you're fine with the two last fixup commits

rjbou (approved):
Thanks!

PR #5383, opened 2022-12-09
Workaround incorrect `<limits.h>` in musl in the release scripts

dra27:
Alternate for #5373 - musl has a pre-kernel 2.6.4 value of `32` for `NGROUPS_MAX`. This PR simply patches the `limits.h` to correct that (in the Docker builds only!) so that OCaml is compiled with the 2.6.4+ value of 65536.

In passing, a slightly inelegant (but tested!) hack to add `-lsha_stubs` only when needed (so that `make -C release TAG=2.1.3 x86_64-linux` works with master)


kit-ty-kate (approved):

kit-ty-kate:
Thanks!

PR #5385, opened 2022-12-12
Add `admin` reftest

rjbou:
* [x] Queued on #5386


tests/reftests/admin.test
```diff
@@ -0,0 +1,536 @@
...
+### : lint :
+### opam admin lint
+
[KIn base-comp.base:
+           warning 47: Synopsis should start with a capital and not end with a dot
+
[KIn ocaml.1.2:
+           warning 56: It is discouraged for non-compiler packages to use 'setenv:'
+
[KIn risus.1:
```
rjbou, line 392: todo: check the output
```suggestion

In base-comp.base:
           warning 47: Synopsis should start with a capital and not end with a dot

In ocaml.1.2:
           warning 56: It is discouraged for non-compiler packages to use 'setenv:'

In risus.1:
```

tests/reftests/dune.inc
```diff
@@ -1,20 +1,20 @@
 
 (rule
- (alias reftest-admin-add-constraint)
```
kit-ty-kate, line 3: I think we can just keep `admin.test` and `admin-add-constraint` separate tests.
To me it makes more sense as `add-constraint` is the most used admin subcommand (plus it makes the change more readable)

PR #5388, opened 2022-12-14
Add dependencies only switch import

### LINKED ISSUE #5200: feature request: opam switch import --deps-only
Hello,

I've a feature request -- similar to `opam install --deps-only`, can we have a `opam switch import --deps-only` -- which only imports the dependencies, but not the root set, of a switch export file? I tried myself at such a patch, but lacked API knowledge to figure out the path.

Best,

Hannes

rjbou:
Flag `--deps-only` for `opam import` that install only non root & compiler packages.
Fixes #5200

* [x] queued on #5208

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -151,6 +151,7 @@ users)
 
 # API updates
 ## opam-client
+  * `OpamSwitchCommand.import`: add optional `?deps_only` argument to install only dependencies of root packages [#5388 @rjbou]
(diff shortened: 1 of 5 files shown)


kit-ty-kate (approved):
i've split and improved the test in a separate commit. lgtm

kit-ty-kate:
Thanks!

PR #5390, opened 2022-12-14
Post 2.1.4 release fixes

kit-ty-kate:
Just a collection of minor fixes to the release script to make it actually work


release/release.sh
```diff
@@ -1,6 +1,9 @@
 #!/usr/bin/env bash
 set -uex
 
+unset $(env | cut -d= -f1 | grep -v '^HOME$' | grep -v '^PWD$' | grep -v '^USER$' | grep -v '^SHELL$' | grep -v '^TERM$')
```
dra27, line 4:
```suggestion
unset $(env | cut -d= -f1 | grep -Fx HOME PWD USER SHELL TERM)
```
but actually `unset` doesn't require the variables to be set at all, so isn't:
```suggestion
unset HOME PWD USER SHELL TERM
```
OK?

dra27 (approved):
Then `unset` can be optionally simplified, but otherwise LGTM!

release/release.sh
dra27: Oh `-v`! Sorry... it is the `grep`, form but `unset $(env | cut -d= -f1 | grep -Fvx HOME PWD USER SHELL TERM)`!

Issue #3782, opened 2019-03-13
GNU patch < 2.7.6 doesn't apply executable bit correctly


dra27:
Found while investigating build failures ocaml/opam-repository#12674

Versions of patch prior to 2.7.6 (includes Debian 9) don't set the executable bit on new files when reading a git patch (see http://git.savannah.gnu.org/cgit/patch.git/commit/?id=592e1f9163d6261359aa87c2c99d411c0dacf6f3)

This causes new executable files to lose the `+x` bit when copied from the Git clone to `.opam/repo/...`

2.7.6. is the very latest release of GNU patch - I expect the solution to this will be to post-process and check the permissions.

AltGr:
One more reason to reimplement our own patch handling :)

dra27:
Indeed!

rjbou:
cf. https://github.com/hannesm/patch

dra27:
Cool - it would need extending to parse Git headers... to be honest, it would probably want extending line-by-line against GNU patch 🤦‍♂️

hannesm:
the `patch` is atm mainly code extracted from conex (where, due to dependencies, and being able to cope with non-unix systems, I don't call out to `patch`). for that, I don't have the need for "parse git headers" or similar features (which does _not_ mean that that should not be added, just that I won't spend much time on it). the scope is also only to handle unified diffs (atm even ignoring merge conflicts) -- if the diff does not apply cleanly, this should result in an error state (but iirc atm its accepted).

@dra27 I don't understand what you mean by "it would probably want extending line-by-line against GNU patch"

dra27:
@hannesm - I mean that patch is so well-specified that it’ll probably need someone to go through its code and reverse-engineer the precise behaviours which want adding

kit-ty-kate:
I got bitten by this again in https://github.com/ocaml/opam-repository/pull/18064
I'll make the opam-repo-ci linter fail if this case is encountered (https://github.com/ocurrent/opam-repo-ci/issues/74) but this is still quite annoying

Issue #3639, opened 2018-10-26
Could not update local repository on mac


feihong:
Problem is identical to that described in #3433, but on Mac OS X instead of OpenBSD. My fix was to run `brew install gpatch`, after which `opam update` worked without issue. Perhaps you could update the [Mac section of the install docs](https://github.com/ocaml/opam/blob/master/doc/pages/Install.md#osx)?

```
$ opam config report
# opam config report
# opam-version      2.0.0
# self-upgrade      no
# system            arch=x86_64 os=macos os-distribution=homebrew os-version=10.13.6
# solver            builtin-mccs+glpk
# install-criteria  -removed,-count[version-lag,request],-count[version-lag,changed],-changed
# upgrade-criteria  -removed,-count[version-lag,solution],-new
# jobs              3
# repositories      1 (http), 2 (version-controlled) (default repo at 3e5ac5b0)
# pinned            1 (git)
# current-switch    /Users/fhsu/work/cool-project
```

rjbou:
opam uses internally `gpatch` [only is the os is an {Open,Free}BSD](https://github.com/ocaml/opam/blob/master/src/core/opamSystem.ml#L1265). Does Mac OS X aliases `patch` to `gpatch` once installed ?

rjbou:
Doc updated, thanks for the report!

fpottier:
I would like to note that I just observed a similar issue under MacOS 10.15.7, where `patch` is `gpatch 2.5.8`, and the issue was resolved by upgrading to `gpatch 2.7.6`.

dra27:
macOS is rapidly becoming the new Windows for shipping with tediously ancient versions of tools (similar recurrent pain in OCaml with GNU make on macOS). Thanks for the data point - I'm hoping that `opam update` in 2.2 will not involve patch files, but the patch pre-processor may need to be battle-tested on some even older versions of patch!

kit-ty-kate:
I believe this was fixed by https://github.com/ocaml/opam/pull/5893 back in opam 2.1.6

Issue #3482, opened 2018-07-24
NetBSD installation issues


maroneze:
I just installed opam on a NetBSD 7.1.2 (via `pkg_add ocaml-opam`), and it seemed to work fine, however I cannot do `opam init` since I have no solvers, and I cannot find NetBSD packages for `aspcud`, `packup` nor `mccs`.

I wonder if I missed something?

I found a mention to NetBSD in [this issue](https://github.com/ocsigen/lwt/issues/177), but the CUDF Solver Farm link mentioned by rneswold is broken.

In the [Specifying solver dependencies](http://opam.ocaml.org/doc/Specifying_Solver_Preferences.html) page in the opam website, the link to aspcud is also broken.

I managed to get packup and compile it (after installing gmp, configuring CPATH and LIBRARY_PATH to include the `/usr/pkg/include` and `/usr/pkg/lib` respectively), then I installed it, tried `opam init` again, and this time I had another error:

```
=-=- Fetching repository information
[ERROR] Could not update repository "default": "/usr/pkg/bin/curl
 --write-out %{http_code}\\n --retry 3 --retry-delay 2 --user-agent opam/2.0.0~beta5
 -L -o /tmp/opam-3783-c61e7b/index.tar.gz.part https://opam.ocaml.org/index.tar.gz"
 exited with code 60
[ERROR] Initial download of repository failed
```

The error was due to the fact that I did not have the [NetBSD CA certificates](https://www.cambus.net/installing-ca-certificates-on-netbsd/), however it was not obvious from the error message.

After doing all of that, I finally had a (mostly) working opam installation, with OCaml 4.06.0. Still, when trying to install several packages, I have opam freezing systematically while applying patches to package `num`. With `OPAMDEBUG=1`, I get:

```
00:15.530 PARALLEL  Next task in job 980392853: /usr/bin/patch
 -p1 -i /root/.opam/default/.opam-switch/build/num.1.0/findlib-install.patch
Processing 6/24: [num: patch]
```

My virtual machine indicates no disk nor network accesses, and there is no CPU usage. It really seems frozen. If I Ctrl+C and try again, the same thing happens. 

I tried installing another package which also has a patch, such as `camlzip`, but it worked as expected.

At this point, I have no idea if it's an issue with opam 2, with the `num` package, with NetBSD, my installation, or something else. I'd like a suggestion on what to try next.

rjbou:
It seems that you had a succession of tiny errors & incompatibility.

[...]

PR #5400, opened 2022-12-22
Replace the dependency on GNU patch by a strict dependency on git

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/3433
Fixes https://github.com/ocaml/opam/issues/3782
Fixes https://github.com/ocaml/opam/issues/3639 for good
Helps https://github.com/ocaml/opam/issues/3482

While looking at https://github.com/ocaml/opam/issues/3639 happening for many macOS users, I realized that `git apply` basically ships with its own version of GNU patch which works correctly.

The original issue is that macOS is a POSIX-complient OS. However POSIX-complient patch seems to be a nightmare: you can't tell it to delete files. It'll just make it empty instead. There are non-posix options (`-E`) that deletes empty files anyway but those have another problem where it can't distinguish between you deleting the file and emptying the file... This issue also appears on BSDs.

The solution here is to make git a required dependency. Most users will have it anyway and many features are missing without it so it's not too much of a stretch to make it mandatory.

### DIFF
--- a/doc/pages/Manual.md
+++ b/doc/pages/Manual.md
@@ -888,10 +888,11 @@ files.
 - <a id="opamfield-patches">`patches: [ <string> { <filter> } ... ]`</a>: a list
   of files relative to the project source root (often added through the `files/`
   metadata subdirectory). The listed patch files will be applied sequentially to
-  the source as with the `patch` command, stripping one level of leading
-  directories (`-p1`) -- which is what version control systems generally use .
-  Variable interpolation is available, so you can specify `patches: [ "file" ]`
-  to have the patch processed from `file.in`.
+  the source using the `patch` command (before opam 2.2) or `git apply`
+  (since opam 2.2), stripping one level of leading directories (`-p1`) -- which
+  is what version control systems generally use. Variable interpolation is
+  available, so you can specify `patches: [ "file" ]` to have the patch
+  processed from `file.in`.
 
     Patches may be applied conditionally by adding _filters_.
 
@@ -1531,8 +1532,9 @@ them modified with [`opam option --global`](man/opam-option.html).
     - `quorum`: integer, the currently defined quorum
     - `repo`: directory containing the already-validated state of the repository
       (empty for an initial validation)
-    - `patch`: for incremental validation, filename of a patch applying to
-      `repo` (with `patch -p1`) and that needs verification
+    - `patch`: for incremental validation, filename of a patch applying to `repo`
+      (using `patch -p1` before opam 2.2, and using `git apply -p1` since opam 2.2)
+      and that needs verification
     - `dir`: for initial validation, the directory to verify
     - `incremental`: `false` if doing an initial validation based on `dir`,
       `true` for an incremental validation based on `repo` and `patch`.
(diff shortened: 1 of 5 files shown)


dra27:
I have a recollection that this wasn't a silver bullet on the Windows side _but_ I think that may have been before the patch rewriter was added. Regardless, this is well-worth revising, thanks! Even if we end up having to keep the patch pre-processing, at least there's only one patch program then to be considering...

MisterDA:
This is great, I've had problems with GNU patch, which lead me to replace this code
```sh
cd ~/opam-repository \
&& (git cat-file -e $OPAM_HASH || git fetch origin master) \
&& git reset -q --hard $OPAM_HASH \
&& git --no-pager log --no-decorate -n1 --oneline \
&& opam update -u
```

with

```sh
cd /home/opam/opam-repository \
&& (git cat-file -e $OPAM_HASH || git fetch origin opam2) \
&& git reset -q --hard $OPAM_HASH \
&& git --no-pager log --no-decorate -n1 --oneline \
&& rsync -ar --update --exclude='.git' ./ /cygdrive/c/opam/.opam/repo/default \
&& ocaml-env exec --64 -- opam update -u
```

master_changes.md
```diff
@@ -23,6 +23,7 @@ users)
   * Disable ACL in Cygwin internal install to avoid permission mismatch errors [#5796 @kit-ty-kate - fix #5781]
   * Add `sys-pkg-manager-cmd` as an accepted field in opamrc files [#5847 @rjbou - fix #5844]
   * Fix `git-location` handling in init config file [#5848 @rjbou - fix #5845]
+  * Replace the dependency on GNU patch by a strict dependency on git [#5400 @kit-ty-kate - fix #3433 #3782 #3639]
```
rjbou, line 26: + change in the API section, proposal
```
  * `OpamSystem.patch`: no more call external tool `patch` but uses `git -c core.autocrlf=false apply --unsafe-paths -p1 <patch>` [#5400 @kit-ty-kate - fix #3433 #3782 #3639]
```

rjbou (commented):
It needs also an update of the documentation, see [`patches:`](https://opam.ocaml.org/doc/Manual.html#opamfield-patches) section.
On the code/change itself, lgtm!

kit-ty-kate:
done

doc/pages/Manual.md
```diff
@@ -888,10 +888,11 @@ files.
 - <a id="opamfield-patches">`patches: [ <string> { <filter> } ... ]`</a>: a list
   of files relative to the project source root (often added through the `files/`
   metadata subdirectory). The listed patch files will be applied sequentially to
-  the source as with the `patch` command, stripping one level of leading
-  directories (`-p1`) -- which is what version control systems generally use .
-  Variable interpolation is available, so you can specify `patches: [ "file" ]`
-  to have the patch processed from `file.in`.
+  the source using the `patch` command (before opam 2.2) or `git apply`
```
rjbou, line 891: There is also a change needed line 1535
```
    - `patch`: for incremental validation, filename of a patch applying to
      `repo` (with `patch -p1`) and that needs verification
```

rjbou:
Reverted in #5891, see comment there for explanation

PR #5402, opened 2022-12-22
Fix the reftests on OCaml 5.0

kit-ty-kate:
With this PR, the reftests are all green on my local machine.

### DIFF
--- a/src/client/opamAdminRepoUpgrade.ml
+++ b/src/client/opamAdminRepoUpgrade.ml
@@ -362,7 +362,7 @@ let do_upgrade repo_root =
       O.with_substs [OpamFilename.Base.of_string conf_script_name] |>
       O.with_build [
         List.map (fun s -> CString s, None)
-          [ "ocaml"; "unix.cma"; conf_script_name ],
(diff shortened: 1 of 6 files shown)


src/client/opamAdminRepoUpgrade.ml
```diff
@@ -362,7 +364,7 @@ let do_upgrade repo_root =
       O.with_substs [OpamFilename.Base.of_string conf_script_name] |>
       O.with_build [
         List.map (fun s -> CString s, None)
-          [ "ocaml"; "unix.cma"; conf_script_name ],
+          [ "ocaml"; conf_script_name ],
```
dra27, line 367: This really shouldn't matter any more, as this only affects the 1.2 upgrade, but is there an issue I haven't spotted with not doing the change above and instead having:
```suggestion
          [ "ocaml"; "-I"; "+unix"; "unix.cma"; conf_script_name ],
```
(apart from the fact that my proposed version is ugly!)

Being ultra-paranoid, the file generated before is compilable, where with the current change it becomes a toplevel-only script?

PR #5403, opened 2022-12-26
Remove unused variable `siz`

purplearmadillo77:
When compiling opam (most recent git) with GCC 11.3.0 on Cygwin64, I get the following error when developer mode is enabled (and NOT during normal compilation since `-Werror` is only activated in developer mode):
```
File "src/stubs/libacl/dune", line 12, characters 19-26:
12 |     (names         opamACL)
                        ^^^^^^^
opamACL.c: In function ‘OPAM_get_acl_executable_info’:
opamACL.c:44:12: error: unused variable ‘siz’ [-Werror=unused-variable]
   44 |     size_t siz;
      |            ^~~
cc1: all warnings being treated as errors
make: *** [Makefile:144: build-opam] Error 1
```
It doesn’t seem like the `siz` variable is used at all in this file, so I’ve removed it.

Additionally, I wasn't sure which category this would fall under in the changes file - I've currently put it under "Build" since the above was a building problem.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -240,6 +240,7 @@ users)
   * Update the bootstrap compiler to 4.14.0 [#5250 @kit-ty-kate]
   * Upgrade the vendored dune to 3.5.0 to fix make cold in an OCaml 5.0 env [#5355 @kit-ty-kate]
   * Upgrade vendored deps to support lib-ext in OCaml 5.0 [#5355 @kit-ty-kate / @dra27]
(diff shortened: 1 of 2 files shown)


dra27 (approved):
Thanks! Section for master_changes is just fine - would you be able to re-push with the PR number?

kit-ty-kate:
> Thanks! Section for master_changes is just fine - would you be able to re-push with the PR number?

I've done it for them. It's easier and faster this way.

kit-ty-kate:
Thanks

PR #5404, opened 2022-12-29
Fix: distinguish MSYS2 from Cygwin

jonahbeckford:
The original code looked like it had a bad merge at some point in its history (duplicates the `if` condition):

```ocaml
            if (OpamString.ends_with ~suffix:"cygwin1.dll" tx ||
                OpamString.ends_with ~suffix:"msys-2.0.dll" tx) then
            (* ... *)
            else if OpamString.ends_with ~suffix:"msys-2.0.dll" tx then
```

Without this PR, MSYS2 uses `cp -PRr` instead of the intended `rsync`; `rsync` is a reasonable fallback for the lack of symlinks in MSYS2, and it copies binaries correctly (ex. if both `dune` and `dune.exe` are in the same directory, `cp` will fail).

The PR does not affect Cygwin which by default uses symlinks.

This PR is the real fix to https://github.com/ocaml/dune/blob/fa3aa2891ac74828a81f73894fe247f69c530c81/ci/build-test.sh#L82-L94 and similar failures.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -563,7 +563,7 @@ users)
   * `OpamFilter`: add `?custom` argument in `to_string` to tweak the output [#5171 @cannorin]
 
 ## opam-core
-  * `OpamStd.Sys`: fix `get_windows_executable_variant` to distinguish MSYS2 from Cygwin, esp. for rsync rather than symlinking [@jonahbeckford]
+  * `OpamStd.Sys`: fix `get_windows_executable_variant` to distinguish MSYS2 from Cygwin, esp. for rsync rather than symlinking [#5404 @jonahbeckford]
   * OpamSystem: avoid calling Unix.environment at top level [#4789 @hannesm]
   * `OpamStd.ABSTRACT`: add `compare` and `equal`, that added those functions to `OpamFilename`, `OpamHash`, `OpamStd`, `OpamStd`, `OpamUrl`, and `OpamVersion` [#4918 @rjbou]
   * `OpamHash`: add `sort` from strongest to weakest kind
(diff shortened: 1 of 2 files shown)


master_changes.md
```diff
@@ -559,6 +559,7 @@ users)
   * `OpamFilter`: add `?custom` argument in `to_string` to tweak the output [#5171 @cannorin]
 
 ## opam-core
+  * `OpamStd.Sys`: fix `get_windows_executable_variant` to distinguish MSYS2 from Cygwin, esp. for rsync rather than symlinking [@jonahbeckford]
```
jonahbeckford, line 562:
```suggestion
  * `OpamStd.Sys`: fix `get_windows_executable_variant` to distinguish MSYS2 from Cygwin, esp. for rsync rather than symlinking [#5404 @jonahbeckford]
```

src/core/opamStd.ml
```diff
@@ -1137,11 +1137,8 @@ module OpamSys = struct
         let rec f a =
           match input_line c with
           | x ->
-            (* Treat MSYS2's variant of `cygwin1.dll` called `msys-2.0.dll` equivalently.
-               Confer https://www.msys2.org/wiki/How-does-MSYS2-differ-from-Cygwin/ *)
```
dra27, line 1141: Good catch - perhaps this comment (the reference to MSYS2's wiki is useful) could be down below in the `is_cygwin_variant` function?

dra27 (approved):
Thanks! Possible new location for the deleted comment 🙂

src/core/opamStd.ml
jonahbeckford: Yep!

dra27 (approved):
LGTM, thanks!

rjbou (approved):

Issue #5399, opened 2022-12-22
"opam list --recursive --coinstallable-with <pkg>.<version>" returns all versions of <pkg>


kit-ty-kate:
opam-repository commit: https://github.com/ocaml/opam-repository/commit/2f19abe6398a316822fcdaa2aa628ef4183ed33f
opam version: https://github.com/ocaml/opam/commit/f93526d411e6f7ef0b4c9d5190ff6d98cf624d28
```
$ opam switch create tmp ocaml-base-compiler.4.08.1
$ opam list -s --color=never --depends-on 'dune.3.6.1' --coinstallable-with 'dune.3.6.1' --installable --all-versions --recursive --depopts | grep "^dune\."
dune.2.0.0
dune.2.0.1
dune.2.1.1
dune.2.1.2
dune.2.1.3
dune.2.2.0
dune.2.3.0
dune.2.3.1
dune.2.4.0
dune.2.5.0
dune.2.5.1
dune.2.6.0
dune.2.6.1
dune.2.6.2
dune.2.7.0
dune.2.7.1
dune.2.8.0
dune.2.8.1
dune.2.8.2
dune.2.8.4
dune.2.8.5
dune.2.9.0
dune.2.9.1
dune.2.9.2
dune.2.9.3
dune.3.0.2
dune.3.0.3
dune.3.1.0
dune.3.1.1
dune.3.2.0
dune.3.3.0
dune.3.3.1
dune.3.4.0
dune.3.4.1
dune.3.5.0
dune.3.6.0
dune.3.6.1
```
`dune.3.6.1` is here due to #4446 but `--coinstallable-with` should get rid of the other versions – and usually does – but does not for `dune` specifically, for some reason.

kit-ty-kate:
The bug seems to be triggered only on packages that contain a `conflicts` field:
```
$ opam list --coinstallable-with tls.0.15.3 tls.0.15.4 --all-versions -A -s
tls.0.15.4
$ opam list --coinstallable-with ocamlfind.1.9.3 ocamlfind.1.9.5 --all-versions -A -s
```

PR #5414, opened 2023-01-11
Fix "opam list -s --coinstallable-with pkg.1 pkg.2" listing pkg.2 as coinstallable with pkg.1

kit-ty-kate:
Fix https://github.com/ocaml/opam/issues/5399


tests/reftests/list-coinstallable.test
```diff
@@ -0,0 +1,7449 @@
+697eab693df71fb665a546e779348a9a823dea2c
```
kit-ty-kate, line 1: One weird thing here is that I've tried to use other hashes but they weren't able to show the same behaviour :/ @AltGr do you have any idea why?

kit-ty-kate: nevermind i found a way. It must've been something that made `opam-0install.0.4.2` unavailable with https://github.com/ocaml/opam-repository/commit/7371c1d9 (this behaviour of `opam list --coinstallable-with` is quite surprising and annoying to work with)

dra27 (approved):
LGTM!

PR #5418, opened 2023-01-13
Enforce extra-source to have a checksum when using "opam switch export --freeze"

kit-ty-kate:
Freeze-exporting packages that have extra-sources without any checksum shouldn't be allowed. This PR fixes this bug.

Side note for reviewers: you can see the diff better using github's "hide whitespace" option


hannesm:
looks great to me.

rjbou (approved):

rjbou:
lgtm, & thanks for the withespace tip!

Issue #185, opened 2012-09-21
opam switch should update ~/.ocamlinit?


avsm:
Setting up `~/.ocamlinit` is the last major pain in having a fully functional `utop` top-level with all the right libraries available.  For instance, mine is currently:

```
#use "topfind"
#camlp4o
#thread
#require "core.top";;
#require "async";;
#require "re";;
#require "uri";;
#require "cohttp.lwt";;
```

which is pretty non-obvious to a beginner.   Would it be possible for OPAM to manage the `.ocamlinit` file, either explicitly (with markers like OASIS has to allow custom changes) or hints to the user?

I imagine this would require packages to also declare their top-level exports, or perhaps this can be automatically inferred via ocamlfind predicates somehow.

This is related to #45, except that I'm now convinced that `utop` is a better solution than the vanilla ocaml toplevel (and doesn't require the shell wrapper).  Also eventually: http://caml.inria.fr/mantis/view.php?id=5699

PR #5430, opened 2023-01-30
Remove ancient unused ocamlinit script

kit-ty-kate:
I'm not seeing it being used anywhere so it might be good to remove it to avoid confusion.


dra27:
History-wise, it appears to have been added in https://github.com/ocaml/opam/commit/b8e5e445cf518438d4a593e53336ee0044e7c5d4 as a preliminary solution to #185 prior to https://github.com/ocaml/opam/commit/a70325769d4632cac2f372cf0ec6dabccf01da36 which was removed in #2611!

The full background is very nicely summarised in https://github.com/ocaml/ocaml/pull/1841#issuecomment-397732003

dra27 (approved):

PR #5435, opened 2023-02-06
Always use O_SHARE_DELETE with Unix.openfile

dra27:
Follows-up https://github.com/ocaml/opam/pull/5351#issuecomment-1322077295 with a better approach. The fds used for locking and for the output of running commands are now opened with `O_SHARE_DELETE`. This has no effect on Unix, but it makes the semantics of these files more POSIX-like on Windows. In particular, it means that a switch can be deleted while the lock is still held. This is both simpler than #5351 and also philosophically more accurate - it means that nothing can attempt to take a write lock until after the directory has been completely unlinked.

It's possible that there are one or two other places - in particular, the file used when monitoring an external solver would be another candidate, but these are less critical. In particular, these two mean that cleaning up switches on error no longer displays a fault in opam and it also means that a "crashed" background process doesn't prevent a switch from being deleted.


rjbou (approved):
lgtm! thanks!

PR #5220, opened 2022-08-03
`opam init` for Windows shells

dra27:
Sat on top of #5217, and also still adding commits.

PR #5438, opened 2023-02-07
Stop using `mv` and switch to `robocopy` on Windows

dra27:
Lots to clean-up in this, obviously. Overlaps with #4823 and #3217.

This PR is enough in #5220 to run `opam init` without Cygwin pre-installed.

### DIFF
--- a/src/core/opamFilename.ml
+++ b/src/core/opamFilename.ml
@@ -109,8 +109,10 @@ let exec dirname ?env ?name ?metadata ?keep_going cmds =
     (fun () -> OpamSystem.commands ?env ?name ?metadata ?keep_going cmds)
 
 let move_dir ~src ~dst =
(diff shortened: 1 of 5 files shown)


src/core/opamFilename.ml
```diff
@@ -109,8 +109,10 @@ let exec dirname ?env ?name ?metadata ?keep_going cmds =
     (fun () -> OpamSystem.commands ?env ?name ?metadata ?keep_going cmds)
 
 let move_dir ~src ~dst =
-  OpamSystem.command ~verbose:(OpamSystem.verbose_for_base_commands ())
-    [ "mv"; Dir.to_string src; Dir.to_string dst ]
+  (* XXX Windows - destination must be on the same drive or this fails *)
```
kit-ty-kate, line 112: Isn’t that the same behaviour as on Unix? `Sys.rename` (morrally) uses `Unix.rename` afaik

PR #5441, opened 2023-02-10
Restore oldcwd if process creation fails

dra27:
Spotted when package commands fail during switch creation - the cwd is left in the build directory for the package which on Windows then prevents the switch from being cleaned up.


src/core/opamProcess.ml
```diff
@@ -450,6 +450,8 @@ let create ?info_file ?env_file ?(allow_stdin=not Sys.win32) ?stdout_file ?stder
       close_stdin  ();
       close_stdout ();
       close_stderr ();
+      begin try Unix.chdir oldcwd
+      with Unix.Unix_error _ -> () end;
```
kit-ty-kate, line 454: Looks obviously correct but couldn’t we merge both uses of `Unix.chdir` using `Fun.protect` starting below the first `Unix.chdir` at the top of the function? This way if another reraise or uncaught exception occures it will be handled automatically

dra27: Yeah, I should having avoided being lazy (the laziness is having to add `Fun.protect` to `OpamCompat` <sigh>)

kit-ty-kate (approved):

master_changes.md
```diff
@@ -618,3 +619,4 @@ users)
   * `OpamStd.Option`: add `equal` function [#5374 @rjbou]
   * `OpamStd.Compare`: add module to flag polymorphic comparison functions in opam codebase [#5374 @kit-ty-kate @rjbou]
   * `OpamStd.Env.`: introduce OpamStd.Env.Name to abstract environment variable names [#5356 @dra27]
+  * `OpamCompat`: Add `Fun.protect` [#5441 @kit-ty-kate]
```
rjbou, line 622: :heart:

rjbou (approved):

rjbou:
thanks!

master_changes.md
AltGr: Hum we already had `OpamStd.Fun.finally` which is the same function (way before there was a `Fun` module in the stdlib :/)
It'd be best to ensure we consistently use the same function everywhere.

PR #5069, opened 2022-02-25
lint W47: clarify empty description message

rjbou:
Specify that descr section is empty instead of description
fix #4989 
```diff
-warning 47: Synopsis (or description first line) should start with a capital and not end with a dot
+warning 47: Synopsis (or 'descr' section first line) should start with a capital and not end with a dot

PR #5442, opened 2023-02-13
opam lint: Enforce synopsis being always present

kit-ty-kate:
Built on top of https://github.com/ocaml/opam/pull/5069

Since https://github.com/ocaml/opam/pull/5070, opam doesn't enforce `synopsis` to be present, only the presence of either synopsis or description is required.

This breaks the behaviour from opam 2.1 and older which always required synopsis to be present and non-empty.

It is desirable to keep this check around to ensure `opam list` shows something helpful (i.e. anything but blank space)

### DIFF
--- a/src/state/opamFileTools.ml
+++ b/src/state/opamFileTools.ml
@@ -654,8 +654,8 @@ let t_lint ?check_extra_files ?(check_upstream=false) ?(all=false) t =
       "It is discouraged for non-compiler packages to use 'setenv:'"
       (t.env <> [] && not (has_flag Pkgflag_Compiler t));
     cond 57 `Error
-      "Synopsis and description must not be both empty"
-      (t.descr = None || t.descr = Some OpamFile.Descr.empty);
+      "Synopsis must not be empty"
+      (match t.descr with None -> true | Some d -> String.equal (OpamFile.Descr.synopsis d) "");
     (let vars = all_variables ~exclude_post:false ~command:[] t in
      let exists svar =
        List.exists (fun v -> v = OpamVariable.Full.of_string svar) vars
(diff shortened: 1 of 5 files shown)


src/state/opamFileTools.ml
```diff
@@ -654,8 +654,8 @@ let t_lint ?check_extra_files ?(check_upstream=false) ?(all=false) t =
       "It is discouraged for non-compiler packages to use 'setenv:'"
       (t.env <> [] && not (has_flag Pkgflag_Compiler t));
     cond 57 `Error
-      "Synopsis and description must not be both empty"
-      (t.descr = None || t.descr = Some OpamFile.Descr.empty);
+      "Synopsis must not be empty"
+      (match t.descr with None -> true | Some d -> OpamFile.Descr.synopsis d = "");
```
rjbou, line 658:
```suggestion
      (match t.descr with None -> true | Some d -> String.equal OpamFile.Descr.synopsis d "");
```

tests/reftests/lint.test
```diff
@@ -401,6 +401,19 @@ ${BASEDIR}/lint.opam: Warnings.
...
+maintainer: "maint@tain.er"
+license: "ISC"
+dev-repo: "hg+https://to@li.nt"
+bug-reports: "https://nobug"
+### opam lint ./lint.opam
+${BASEDIR}/lint.opam: Errors.
+             error 57: Synopsis must not be empty
+# Return code 1 #
+### <lint.opam>
+opam-version: "2.0"
```
rjbou, line 416: This check and the one below need to go on `E57` section line 624.

rjbou: also, i forgot to change W47 comment line with the new description (line 388), can you update it and E57 one too (line 624) ?

rjbou (changes requested):

rjbou (approved):

PR #5449, opened 2023-02-21
Reftests: Clone opam-repository once and use it locally instead of downloading individual commits

kit-ty-kate:
~This is faster on slower hardware but 20 seconds slower on fast hardware (2.50s on M1Pro vs. 2.30s before)~
~The size of artifacts in `_build` also jump from 5.4GB to 7.1GB sadly.~

~If anyone has an idea how to reduce the size I'd gladly take it.~

See next comment for the updated summary

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -120,6 +120,7 @@ users)
 ### Engine
  * Fix gcc < 14.3 bug on mingw i686 [#6624 @kit-ty-kate]
   * Fix support for removing local link directories [#6450 @kit-ty-kate]
+  * Clone opam-repository once and use it locally instead of downloading individual commits [#5449 @kit-ty-kate]
 
 ## Github Actions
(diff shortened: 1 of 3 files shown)


kit-ty-kate:
Thanks to @dra27 the version using `git worktree` now produces a smaller tests directory than before this PR and is only 10 to 15s slower than the original on fast hardware (still faster on slower ones)

dra27 (approved):
Seems win-win to me, thank you! Happy for this to be merged, too, @rjbou?

kit-ty-kate:
To make the CI even faster we could even take the bare clone of opam-repository that is already present in the cache and move it to `tests/reftests/opam-repository.git`

kit-ty-kate:
@arozovyk does this speed up `make test` for you?

Issue #5450, opened 2023-02-22
opam does not compile packages to ARM arch on macOS


tiuno:
```
# opam config report
# opam-version         2.1.4 
# self-upgrade         no
# system               arch=x86_64 os=macos os-distribution=homebrew os-version=12.5
# solver               builtin-mccs+glpk
# install-criteria     -removed,-count[avoid-version,changed],-count[version-lag,request],-count[version-lag,changed],-count[missing-depexts,changed],-changed
# upgrade-criteria     -removed,-count[avoid-version,changed],-count[version-lag,solution],-count[missing-depexts,changed],-new
# jobs                 7
# repositories         1 (http), 1 (version-controlled) (default repo at 15da45a2)
# pinned               0
# current-switch       default
# ocaml:native         true
# ocaml:native-tools   true
# ocaml:native-dynlink true
# ocaml:stubsdir       /Users/colin/.opam/default/lib/ocaml/stublibs:/Users/colin/.opam/default/lib/ocaml
# ocaml:preinstalled   false
# ocaml:compiler       4.14.1
```

On macOS Monterey with a M1 arm64 machine, `opam` does not detect the host architecture properly.

Even if set manually, opam will not compile packages for the correct architecture.

To recreate this behaviour, install opam following the site instructions, configure opam  architecture manually to arm64, create a switch and install a package.
The problem is then observed by checking the kind of file produced using macOS `file`.

```
arch -arm64 /usr/bin/bash -c "sh <(curl -fsSL https://raw.githubusercontent.com/ocaml/opam/master/shell/install.sh)"

opam init --bare
opam var --global arch=arm64
opam switch create arm 4.14.0
opam switch set arm
eval $(opam env)
opam install dune
file $(command -v ocamlc) $(command -v dune)
```

The last two commands will should produce similar outputs like the following :  `.opam/arm/bin/ocamlc: Mach-O 64-bit executable x86_64` instead of the expected `Mach-O 64-bit executable arm64`.

Is this the expected behaviour? This yields incompatibility issues with third party libraries.

kit-ty-kate:
Your default terminal seems to be running in x86_64 mode, so any programs run by opam (even if opam if an arm64 executable) will be run in x86_64 mode.
If you really want to keep the x86_64 mode by default in your terminal you will need to run every opam call through `arch -arm64`

tiuno:
[...]

PR #5451, opened 2023-02-22
Enforce the macOS sandbox script to use /bin/bash instead of /usr/bin/env bash for a more consistent experience

kit-ty-kate:
For most users this is strictly equivalent, however in some cases another (possibly broken) bash could be installed and used instead.

Fixes https://github.com/ocaml/opam/issues/5450

### DIFF
--- a/src/state/shellscripts/sandbox_exec.sh
+++ b/src/state/shellscripts/sandbox_exec.sh
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
 set -ue
 
 POL='(version 1)(allow default)(deny network*)(deny file-write*)'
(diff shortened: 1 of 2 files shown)


dra27:
This PR changes the Linux script at the moment - it's `sandbox_exec.sh` for macOS?

src/state/shellscripts/sandbox_exec.sh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env bash
+#!/bin/bash
```
dra27, line 1: On the basis that the shell-check/change-all-the-shebangs-style PRs in future might regress this:
```suggestion
#!/bin/bash
# This script is only used on macOS, where /bin/bash is both guaranteed to exist and
# and to be for the native architecture, which is why /usr/bin/env bash is not used.
# See https://github.com/ocaml/opam/issues/5450
```

dra27 (approved):
Suggestion for an extra comment just to avoid the risk of this being regressed by any shell-checking!

rjbou (approved):

PR #5452, opened 2023-02-24
Implement accessors for `OpamProcess.command` type

Leonidas-from-XIV:
This is useful when wanting to implement other runners for `OpamProcess.Job.Op`, e.g. based on Dune's Fibers, Lwt or Async as it allows to run the process using their process spawning abstractions.

### DIFF
--- a/src/core/opamProcess.mli
+++ b/src/core/opamProcess.mli
@@ -12,7 +12,18 @@
 (** Process and job handling, with logs, termination status, etc. *)
 
 (** The type of shell commands *)
-type command
+type command = private {
+  cmd: string;
+  args: string list;
+  cmd_text: string option;
+  cmd_dir: string option;
+  cmd_env: string array option;
+  cmd_stdin: bool option;
+  cmd_stdout: string option;
+  cmd_verbose: bool option;
+  cmd_name: string option;
+  cmd_metadata: (string * string) list option;
+}
 
 (** Builds a shell command for later execution.
     @param env         environment for the command
--- a/master_changes.md
+++ b/master_changes.md
@@ -678,3 +678,4 @@ users)
   * `OpamURL`: add module `SWHID` that contains helpers from and to internal swhid url [#4859 @rjbou]
   * `OpamSystem.read_command_output`: add an optional parameter to unmerge stdout and stderr [#4859 @rjbou]
   * `OpamSWHID`: add module to handle swhid [#4859 @rjbou]
+  * `OpamProcess`: expose the `command` type as a private type [#5452 @Leonidas-from-XIV]


kit-ty-kate (approved):

src/core/opamProcess.ml
```diff
@@ -212,6 +212,12 @@ let default_verbose () = OpamCoreConfig.(!r.verbose_level) >= 2
 let is_verbose_command c =
   OpamStd.Option.default (default_verbose ()) c.cmd_verbose
 
+let command_cmd c = c.cmd
+let command_args c = c.args
+let command_dir c = c.cmd_dir
+let command_env c = c.cmd_env
+let command_allow_stdin c = c.cmd_stdin
```
rjbou, line 219: Please use same naming than existing
```suggestion
let cmd_of_command c = c.cmd
let args_of_command c = c.args
let dir_of_command c = c.cmd_dir
let env_of_command c = c.cmd_env
let allow_stdin_of_command c = c.cmd_stdin
```

src/core/opamProcess.mli
```diff
@@ -42,6 +42,12 @@ val string_of_command: command -> string
 val text_of_command: command -> string option
 val is_verbose_command: command -> bool
 
+val command_cmd : command -> string
+val command_args : command -> string list
+val command_dir : command -> string
+val command_env : command -> string array
+val command_allow_stdin : command -> bool option
```
rjbou, line 49:
```suggestion
val command_cmd : command -> string
val command_args : command -> string list
val command_dir : command -> string option
val command_env : command -> string array option
val command_allow_stdin : command -> bool option
```

rjbou (changes requested):
Ok on the idea, comme comments/fix

src/core/opamProcess.ml
kit-ty-kate: I'm not sure about using this naming convention for those accessors. The `*_of_*` functions in OCaml usually imply some kind of processing (e.g. conversion from one type to the other)

Accessors are more straightforward. Furthermore I think it isn't clear for users what those do without reading the implementation, especially for `cmd_of_command` it isn't clear to me from the name what the difference between `cmd_of_command` and `string_of_command` is. `command_cmd` looks more like the typical conversion for accessors to me.

rjbou (approved):

src/core/opamProcess.mli
Leonidas-from-XIV: Thanks! I think this was due me vendoring opam and dune apparently just believed the interfaces without building the implementations. Will be careful about this in the future!

src/core/opamProcess.ml
Leonidas-from-XIV: I would generally agree, which is why originally I had them using the more usual accessor "convention". In the end I am ok with whatever OPAM maintainers prefer.

rjbou: Good question. The idea behind my proposal is to keep it homogeneous with existing ones.  We'll end up with `text_of_command` and `command_args`, both direct accessors. 
`string_of_` is another naming convention, in opam it is always used to have the full string of a type (with Module.to_string)

kit-ty-kate: The only existing one is `text_of_command`, the rest are not accessors. With that in mind i think it's fine to have two naming conventions.

Leonidas-from-XIV: I've pushed a separate commit renaming them, it can be removed if desired before merging.

Leonidas-from-XIV:
@kit-ty-kate I tried it and that works as well for me. I didn't expose it originally because exposing a type can be painful later if the type is to be changed.

kit-ty-kate:
Thanks!

PR #5455, opened 2023-02-28
Depext filter enhancement: resolve package variables

rjbou:
`depexts:` filter can now handle package variables. They are stored and resolved on the go & gradually. On install, packages appearing in solution are considered as installed for resolving. See `depexts.test` for examples.

* [x] queued on #5453 
* [ ] add test on unavailability

### DIFF
--- a/tests/reftests/depexts.test
+++ b/tests/reftests/depexts.test
@@ -33,3 +33,199 @@ false
 [NOTE] You can retry with '--assume-depexts' to skip this check, or run 'opam option depext=false' to permanently disable handling of system packages.
 
 # Return code 10 #
+### : filter with installed :
+### opam var --global os-family=dummy-success
+Added '[os-family "dummy-success" "Set through 'opam var'"]' to field global-variables in global configuration
+### <pkg:conf-b.1>
+opam-version: "2.0"
+depends: "conf-os"
+### <pkg:conf-os.1>
+opam-version: "2.0"
+depends: [ "conf-os1" | "conf-os2" ]
+depexts:[
+  [ "os1-depext" ] { conf-os1:installed }
+  [ "os2-depext" ] { conf-os2:installed }
+]
+### <pkg:conf-os2.1>
+opam-version: "2.0"
+conflicts: "conf-os1"
+### <pkg:conf-os1.1>
+opam-version: "2.0"
+conflicts: "conf-os2"
+### opam install conf-os | sed-cmd echo
+The following actions will be performed:
+=== install 2 packages
+  - install conf-os  1
+  - install conf-os1 1 [required by conf-os]
+
+The following system packages will first need to be installed:
+    os1-depext
+
+<><> Handling external dependencies <><><><><><><><><><><><><><><><><><><><><><>
+
+opam believes some required external dependencies are missing. opam can:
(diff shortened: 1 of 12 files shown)


tests/reftests/depexts.test
```diff
@@ -33,3 +33,199 @@ false
...
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> removed   conf-os1.1
+Done.
+### OPAMVAR_os_family=dummy-success::0 opam install conf-os --show
+The following actions would be performed:
+=== install 2 packages
+  - install conf-os  1
+  - install conf-os1 1 [required by conf-os]
+[WARNING] These additional system packages are required, but not available on your system: os1-depext
```
kit-ty-kate, line 109: I'm not sure to understand this test result. Why is this just a warning if the system package isn't available?

tests/reftests/depexts.test
```diff
@@ -33,3 +33,199 @@ false
...
+### OPAMVAR_os_family=dummy-success::0 opam install conf-os
+The following actions will be performed:
+=== install 2 packages
+  - install conf-os  1
+  - install conf-os1 1 [required by conf-os]
+[WARNING] These additional system packages are required, but not available on your system: os1-depext
+[ERROR] Some packages have missing dependencies after resolution:
```
kit-ty-kate, line 116: Same sort of question than above here: why are we displaying the error so late?

tests/reftests/depexts.test
```diff
@@ -33,3 +33,199 @@ false
...
+           - conf-os.1: os1-depext
```
kit-ty-kate, line 117: Shouldn't also say something about `os2-depext` ?

tests/reftests/depexts.test
rjbou: It should be a simple message indeed, or an error, not a warning.

tests/reftests/depexts.test
rjbou: We check depext availability when we try to install once we have the solution. We resolve last filters (here `conf-os1:installed`) with an environment that contains packages to install as installed.

tests/reftests/depexts.test
rjbou: In fact, here it's  with `--show`, it errors at install step.

tests/reftests/depexts.test
rjbou: As we resolve with the solution, `conf-os1:installed` resolves to true while `conf-os2` resolves to false, that's why we display only `os1-depext`.
We can recheck availability at this stage, and print a message for `os2-depext`.

kit-ty-kate:
I'm rather unconvinced by the solution and I'm still confused as to why this is necessary for Windows support.
Here is the summary of the problems i can see after thinking about it some more:
* Proper support requires recursive solve (do one solve, check if it installs unavailable depexts, do another solve without this package, …)
* The current way of doing (one solve, check if it's actually possible quite late in the process) feels quite jenky and doesn't really help users understand issues
* It makes it quite hard to understand what a package is doing and unpredictable

Could describe what the problem it's trying to solve is?

rjbou:
Closing as not needed nor wanted behaviour.

PR #5457, opened 2023-02-28
windows: opam root redirection when path contains spaces

rjbou:
On Windows, opam may install a local cygwin, but cygwin doesn't support having a space in its path. We need then to redirect all opam root into a path with no space. Redirection is done at first init, and afterwards it is followed for each opam call.
So opam root directory can contain either:
*  an unique file `redirected-opamroot` that contains the new path
* or the already known opam root.

* [x] update changes
* [x] update message with relevant root dir
* [x] set default path, or give user the choice
* [x] add an error if `--root`/`OPAMROOT` path contains space
* [x] add readme.txt in C:\opamroot

related to #5220

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -1645,6 +1645,114 @@ let reinit ?(init_config=OpamInitDefaults.init_config()) ~interactive
   in
   OpamRepositoryState.drop rt
 
+let has_space s = OpamStd.String.contains_char s ' '
+
+let default_redirect_root = OpamFilename.Dir.of_string "C:\\opamroot"
+
+let setup_redirection target =
+  let {contents = {OpamStateConfig.original_root_dir = root; _}} =
+    OpamStateConfig.r
+  in
+  let target =
+    match target with
+    | Some target -> target
+    | None ->
+      OpamFilename.mkdir default_redirect_root;
+      let readme = OpamFilename.Op.(default_redirect_root // "ReadMe.txt") in
+      if not (OpamFilename.exists readme) then
+        OpamFilename.write readme
+          "This directory is used to contain redirected opam roots.\n\n\
+           The contents may be shared with other users on this system.";
+      OpamSystem.mk_unique_dir ~dir:(OpamFilename.Dir.to_string default_redirect_root) ()
+  in
+  let root_dir = OpamFilename.Dir.of_string target in
+  OpamFilename.write (OpamPath.redirected root) target;
+  OpamStateConfig.update ~root_dir ();
+  root_dir
+
+let get_redirected_root () =
+  let {contents = {OpamStateConfig.original_root_dir = root; root_from; _}} =
+    OpamStateConfig.r
+  in
+  let r = OpamConsole.colorise `bold (OpamFilename.Dir.to_string root) in
+  let collision =
+    let collision = OpamConsole.utf8_symbol OpamConsole.Symbols.collision "" in
+    if collision = "" then
+      ""
+    else
+      " " ^ collision
+  in
+  let options = [
+    `Redirect, Printf.sprintf
+      "Redirect files to a directory in %s"
+      (OpamConsole.colorise `bold (OpamFilename.Dir.to_string default_redirect_root));
+    `Ask, "Redirect files to an alternate directory";
+    `Endure, Printf.sprintf
+      "Do not redirect anything and stick with %s%s" r collision;
+    `Quit, "Abort initialisation"
+  ] in
+  let default, explanation =
+    match root_from with
+    | `Command_line ->
+      (* The user has been explicit with --root; nemo salvet modo... *)
+      `Endure,
+      "You have specified a root directory for opam containing a space."
+    | `Env ->
+      (* The user has perhaps carelessly set an environment variable *)
+      `Redirect,
+      "Your OPAMROOT environment variable contains a space."
+    | `Default ->
+      (* The user has fallen victim to the defaults of Windows Setup and has a
+         space in their user name *)
+      `Redirect,
+      Printf.sprintf
+        "By default, opam would store its data in:\n\
+           %s\n\
+         however, this directory contains a space." r
+  in
+  let rec ask () =
+    let check r =
+      if Filename.is_relative r then begin
+        OpamConsole.msg
+          "That path is relative!\n\
+           Please enter an absolute path without spaces.\n";
+        ask ()
+      end else if has_space r then begin
+        OpamConsole.msg
+          "That path contains contains a space!\n\
+           Please enter an absolute path without spaces.\n";
+        ask ()
+      end else
+        Some (Some r)
+    in
+    OpamStd.Option.replace check (OpamConsole.read "Root directory for opam: ")
+  in
+  let rec menu () =
+    match OpamConsole.menu "Where should opam store files?" ~default ~options
+            ~no:default with
+    | `Redirect ->
+      Some None
+    | `Endure ->
+      None
+    | `Ask ->
+      let r = ask () in
+      if r = None then
+        menu ()
+      else
+        r
+    | `Quit ->
+      OpamStd.Sys.exit_because `Aborted
+  in
+  OpamConsole.header_msg "opam root file store";
+  OpamConsole.msg
+    "\n\
+     %s\n\
+     \n\
+     Many parts of the OCaml ecosystem do not presently work correctly\n\
+     when installed to directories containing spaces. You have been warned!%s\n\
+     \n" explanation collision;
+  Option.map setup_redirection (menu ())
+
 let init
     ~init_config ~interactive
     ?repo ?(bypass_checks=false)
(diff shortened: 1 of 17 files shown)


kit-ty-kate:
note from dev meeting: for the directory we could create a new directory in `C:\`, it should be possible to create one without admin priviledges

src/client/opamClient.ml
```diff
@@ -1143,7 +1143,27 @@ let init
     shell =
   log "INIT %a"
     (slog @@ OpamStd.Option.to_string OpamRepositoryBackend.to_string) repo;
-  let root = OpamStateConfig.(!r.root_dir) in
+  let root =
+    let root = OpamStateConfig.(!r.root_dir) in
+    if Sys.win32 &&
+       OpamStd.String.contains_char (OpamFilename.Dir.to_string root) ' ' then
+      (let default = Filename.concat "C:" "opamroot" in
```
rjbou, line 1150:
```suggestion
      (let default = "C:\\opamroot" in
```

tests/reftests/env.test
```diff
@@ -439,60 +439,6 @@ The following actions will be performed:
...
-### opam switch create "./$SW" nv --root "$RT"
-
-<><> Installing new switch packages <><><><><><><><><><><><><><><><><><><><><><>
-Switch invariant: ["nv"]
-
-<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
--> installed nv.1
-Done.
-# Run eval $(opam env '--root=${BASEDIR}/root 2' '--switch=${BASEDIR}/switch w spaces') to update the current shell environment
-### opam env --root "$RT" --switch "./$SW" | grep "NV_VARS" | ';' -> ':'
```
kit-ty-kate, line 460: why is this test not needed anymore? What was it testing to being with?

kit-ty-kate (commented):
General question: why do we redirect the entire opam root instead of just installing Cygwin in `C:\opam-cygwin` or similar? What are the advantage/inconvenient of both solutions?

src/client/opamClientConfig.ml
```diff
@@ -210,6 +210,13 @@ let opam_init ?root_dir ?strict ?solver =
 
   (* (i) get root dir *)
   let root = OpamStateConfig.opamroot ?root_dir () in
+  if Sys.win32
+     (* if default, redirection will be handled by opam init, or should have
+        been handled *)
+     && (root_dir <> None || OpamStateConfig.E.root () <> None)
+     && OpamStd.String.contains_char (OpamFilename.Dir.to_string root) ' ' then
+    OpamConsole.error "You opam root directory contains a space, this may lead \
+                       to several malfunction... bzzz.... nooo💥";
```
rjbou, line 219: This is a opening for proposal :)

tests/reftests/env.test
kit-ty-kate: I'm not seeing that command in either of those files

src/client/opamClientConfig.ml
kit-ty-kate: that looks fine by me

tests/reftests/env.test
rjbou: ah,the command itself, true.

rjbou: I think that originally it was here to show that the switch was well created?

src/client/opamClient.ml
```diff
@@ -1143,7 +1143,27 @@ let init
...
+    if Sys.win32 &&
+       OpamStd.String.contains_char (OpamFilename.Dir.to_string root) ' ' then
+      (let default = "C:\\opamroot" in
+       let new_root_f =
+         if OpamConsole.confirm ~default:false
+             "Your opam root path '%s' contains a space, we'll redirect to \
+              '%s'.\nDo you want to choose and enter another spaceless folder"
+             (OpamFilename.Dir.to_string root) default then
+           match OpamConsole.read "Opam root: " with
+           | Some r -> r
```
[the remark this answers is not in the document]
rjbou: updated

tests/reftests/env.test
rjbou: updated

src/client/opamClient.ml
```diff
@@ -1163,6 +1193,12 @@ let init
           (OpamFilename.Dir.to_string root);
         if not (OpamConsole.confirm "Proceed?") then
           OpamStd.Sys.exit_because `Aborted);
+      (* Add the readme file in C:\opamroot as redirected *)
+      (if Sys.win32 && redirected then
```
kit-ty-kate, line 1197:
```suggestion
      (if redirected then
```

kit-ty-kate:
> General question: why do we redirect the entire opam root instead of just installing Cygwin in `C:\opam-cygwin` or similar? What are the advantage/inconvenient of both solutions?

my question wasn't answered. Looks fine otherwise on a glance, i'll test it tomorrow

rjbou:
I'm aware that your question wasn't answered. I hadn't time to answer yesterday, just to update the PR (not even to test the new commits).
The main point, if I remember well, is to keep everything under the [opam root](https://opam.ocaml.org/doc/FAQ.html#What-changes-does-opam-do-to-my-filesystem). Besides, users are used to delete opam root to restart from scratch. If internal cygwin is a separated directory, it can lead to partial install left in that case.
I agree that on purely technical part, it'll be easier (and less changes) if we store only the Cygwin install under `C:`.
@dra27 Do you remember another point we discussed at the time ?

src/client/opamClientConfig.ml
dra27: I like it - although the 💥 should be guarded with a UTF-8 trick (like the camel!)

kit-ty-kate: done

tests/reftests/env.test
dra27: This is safe - the main thing is that the primary tests in #5935 (further up the file) are still there. The noise in these other parts suggests that the tests should possibly be split up a bit further (but's _definitely_ a post 2.2.0 thing!)

tests/reftests/init.win32.test
```diff
@@ -0,0 +1,28 @@
...
+### opam switch create --empty test
+### opam switch --debug --debug-level=-1
+CLI                             Parsing CLI version 2.2
+GSTATE                          LOAD-GLOBAL-STATE @ C:\opamroot
+SWITCH                          list
+#   switch  compiler  description
+->  test              test
+### echo $OPAMROOT
+roots/with path
+### rm -rf 'C:\opamroot'
```
dra27, line 28: This definitely can't stay - it blows away the user's opam root!

kit-ty-kate: but then the tests are not reproducible on Windows at all

kit-ty-kate:
The tests have been split off in #6006

src/client/opamClient.ml
```diff
@@ -1645,6 +1645,114 @@ let reinit ?(init_config=OpamInitDefaults.init_config()) ~interactive
...
+    match target with
+    | Some target -> target
+    | None ->
+      OpamFilename.mkdir default_redirect_root;
+      let readme = OpamFilename.Op.(default_redirect_root // "ReadMe.txt") in
+      if not (OpamFilename.exists readme) then
+        OpamFilename.write readme
+          "This directory is used to contain redirected opam roots.\n\n\
+           The contents may be shared with other users on this system.";
+      OpamSystem.mk_unique_dir ~dir:(OpamFilename.Dir.to_string default_redirect_root) ()
```
kit-ty-kate, line 1666:
```suggestion
let default_redirect_root = lazy (OpamSystem.mk_unique_dir ~dir:"C:\\opamroot" ())

let setup_redirection target =
  let {contents = {OpamStateConfig.original_root_dir = root; _}} =
    OpamStateConfig.r
  in
  let target =
    match target with
    | Some target -> target
    | None ->
      OpamFilename.mkdir (Lazy.force default_redirect_root);
      let readme = OpamFilename.Op.(Lazy.force default_redirect_root // "ReadMe.txt") in
      if not (OpamFilename.exists readme) then
        OpamFilename.write readme
          "This directory is used to contain redirected opam roots.\n\n\
           The contents may be shared with other users on this system.";
```

src/client/opamClient.ml
```diff
@@ -1645,6 +1645,114 @@ let reinit ?(init_config=OpamInitDefaults.init_config()) ~interactive
...
+    let collision = OpamConsole.utf8_symbol OpamConsole.Symbols.collision "" in
+    if collision = "" then
+      ""
+    else
+      " " ^ collision
+  in
+  let options = [
+    `Redirect, Printf.sprintf
+      "Redirect files to a directory in %s"
+      (OpamConsole.colorise `bold (OpamFilename.Dir.to_string default_redirect_root));
```
kit-ty-kate, line 1688:
```suggestion
      (OpamConsole.colorise `bold (OpamFilename.Dir.to_string (Lazy.force default_redirect_root)));
```

src/client/opamClient.ml
dra27: The idea is to `C:\opamroot\ReadMe.txt` explaining _all_ of the redirected files, not to have one in each directory

src/client/opamClient.ml
dra27: The message isn't correct with this change - it was intentional that it displays "Redirect files to a directory in C:\opamroot" - the user can get the actual subdirectory with `opam var root` afterwards, but that has a "scary" random number in it which I was purposefully not displaying by default!

tests/reftests/init.win32.test
rjbou: It's ok for CI, on tester computer, we can add a note about it in the test and not remove something outside the test directories

kit-ty-kate (approved):

rjbou:
:tada:

rjbou:
Thanks all!

PR #5465, opened 2023-03-07
Remove unused vendored dependency: result

kit-ty-kate:
This doesn't seem to be used anywhere. `topkg` stopped depending on it in 1.0.4 and `base64` did it in 1.0.1


dra27 (approved):
LGTM!

PR #5466, opened 2023-03-07
Bump the minimum requirement to build any of the opam libraries to OCaml >= 4.08

kit-ty-kate:
Depends on the outcome of https://discuss.ocaml.org/t/does-anyone-use-the-opam-libraries-with-ocaml-4-08/11607


dra27 (approved):
Let's do it!

PR #5467, opened 2023-03-08
Bump vendored dependency: jsonm 1.0.2

kit-ty-kate:
No more patches!


dra27 (approved):
Very pleasing, and long may it last! 🙂

PR #5471, opened 2023-03-09
lock: fix pin-depends with `with-` variables

### LINKED ISSUE #5428: opam lock does not include pin-depends for `with-test` dependencies
Here is a minimized version of my project's opam file:
```
opam-version: "2.0"
maintainer: "foo@bar.baz"
depends: [
  "bisect_ppx" {with-test}
  "dune" {>= "3.4"}
  "ocaml" {>= "5.0.0"}
  "ocaml-protoc-plugin"
]
synopsis: "repro for pinned test deps"
pin-depends: [
  [ "bisect_ppx.dev" "git+https://github.com/anmonteiro/bisect_ppx#cc442a08e3a2e0e18deb48f3a696076ac0986728" ]
  [ "ocaml-protoc-plugin.dev" "git+https://github.com/dialohq/ocaml-protoc-plugin#b814b305520563fff58388682cb360660cc29c47" ]
]
```

When I install the dependencies (`opam install . --with-test`) and execute `opam lock .`, here is my `opam.locked` file:
```
opam-version: "2.0"
maintainer: "foo@bar.baz"
depends: [
  "bisect_ppx" {= "dev" & with-test}
  "dune" {= "3.6.1"}
  "ocaml" {= "5.0.0"}
  "ocaml-protoc-plugin" {= "dev"}
]
synopsis: "repro for pinned test deps"
pin-depends: [
  [ "ocaml-protoc-plugin.dev" "git+https://github.com/dialohq/ocaml-protoc-plugin#b814b305520563fff58388682cb360660cc29c47" ]
]
```

As you can see, the pin-depends of the `with-test` dependency doesn't get added, which later creates issues when doing:
```
opam install . --locked --with-test
```

Tested with opam 2.1.4 and opam master.

rjbou:
fix #5428

### DIFF
--- a/tests/reftests/lock.test
+++ b/tests/reftests/lock.test
@@ -1315,6 +1315,6 @@ license: "MIT"
 maintainer: "maint@tain.er"
 name: "tolock"
 opam-version: "2.0"
-pin-depends: ["a-simple-dep.dev" "git+https://github.com/dbuenzli/cmdliner#f239981642a"]
+pin-depends: [["a-doc-dep.dev" "git+https://github.com/dbuenzli/cmdliner#e8f8890fe48"] ["a-simple-dep.dev" "git+https://github.com/dbuenzli/cmdliner#f239981642a"] ["a-test-dep.dev" "git+https://github.com/dbuenzli/cmdliner#1de361182ab"]]
 synopsis: "A word"
 version: "dev"
--- a/tests/reftests/lock.test
+++ b/tests/reftests/lock.test
@@ -1235,3 +1235,86 @@ available: opam-version >= "2.1"
 url {
 }
 x-locked: "locked"
+### : Pin depends :
+### rm -r tolock
+### opam switch create pin-depends --empty
+### <pin:tolock/tolock.opam>
+opam-version: "2.0"
+depends: [
+  "foo"
+  "a-simple-dep"
+  "a-test-dep" {with-test}
+  "a-doc-dep" {with-doc}
+]
+build: [ "test" "-f" "content" ]
+pin-depends: [
+  [ "a-simple-dep.dev" "git+https://github.com/dbuenzli/cmdliner#f239981642a" ]
+  [ "a-test-dep.dev"   "git+https://github.com/dbuenzli/cmdliner#1de361182ab" ]
+  [ "a-doc-dep.dev"    "git+https://github.com/dbuenzli/cmdliner#e8f8890fe48" ]
+]
+### <tolock/content>
+it's here
+### <pin:opam-file>
+opam-version: "2.0"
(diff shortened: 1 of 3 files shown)


tests/reftests/lock.test
```diff
@@ -1239,3 +1239,86 @@ available: opam-version >= "2.1"
...
+-> retrieved a-test-dep.dev  (no changes)
+-> installed a-test-dep.dev
+-> installed foo.2
+-> retrieved tolock.dev  (file://${BASEDIR}/tolock)
+-> installed tolock.dev
+Done.
+### opam lock tolock
+Generated lock files for:
+  - tolock.dev: ${BASEDIR}/tolock.opam.locked
+### opam-cat tolock.opam.locked
```
kit-ty-kate, line 1310: cat would be way more readable
```suggestion
### cat tolock.opam.locked
```

kit-ty-kate (commented):
lgtm apart from this small suggestion

tests/reftests/lock.test
```diff
@@ -1239,3 +1239,86 @@ available: opam-version >= "2.1"
...
+  [ "a-simple-dep.dev" "git+https://github.com/dbuenzli/cmdliner#f239981642a" ]
+  [ "a-test-dep.dev"   "git+https://github.com/dbuenzli/cmdliner#1de361182ab" ]
+  [ "a-doc-dep.dev"    "git+https://github.com/dbuenzli/cmdliner#e8f8890fe48" ]
+]
+### <tolock/content>
+it's here
+### <pin:opam>
+opam-version: "2.0"
+version: "dev"
+### OPAMEDITOR="cp -f ${BASEDIR}/opam"
```
kit-ty-kate, line 1264: To fix the tests:
```suggestion
### <pin:opam-file>
opam-version: "2.0"
version: "dev"
### OPAMEDITOR="cp -f ${BASEDIR}/opam-file"
```

tests/reftests/lock.test
rjbou: `opam-cat` is used to avoid having the test change in case the format/padding on an opam file printer changes, the order of fields, etc. It's a reproducible way to print opam file in reftest when we care only about the content.

tests/reftests/lock.test
kit-ty-kate: See CI state of the previous commit. The logs disappeared but the tests would fail because `./opam` would clash with `./OPAM` on case-insensitive file systems (macOS and Windows in github action)

rjbou: oh, ok. Yup, too old PR to have the logs. Plus I hit it in another PR.

kit-ty-kate (approved):

Issue #5440, opened 2023-02-09
Opam config nuked after running out of disk space


sim642:
This is a rather obscure scenario, but I happened to run out of disk space in the middle of some `opam upgrade`. Some arbitrary package build then failed with `Sys_error("No space left on device")`.

I didn't expect this to be much of an issue, but after freeing up disk space and rerunning opam, nothing works because `~/.opam/config` is now completely empty (which I suppose it shouldn't be). I guess opam tried to rewrite the file but couldn't due to disk space and was thus left in such broken state.

```
# opam config report
# opam-version         2.1.2 
# self-upgrade         no
# system               arch=x86_64 os=linux os-distribution=ubuntu os-version=22.04
# read-state           In /home/simmo/.opam/config:
unsupported or missing file format version; should be 2.0 or older
```

Not sure if much could be done about such corner case. Maybe writing the new file before removing the old one and renaming?

sim642:
`opam init --reinit` nor `opam init --bare` don't seem to fix anything either.

kit-ty-kate:
to avoid this to happen for other people we should enforce writes to be atomic on our end. For config files it shoudn’t be too hard, I’ll look into it.

dra27:
Note that there's a commit already for atomic writes in https://github.com/ocaml/opam/pull/5417/commits/bb0d675dba3c6c128470d7fc41160837bf757a78 - that could possibly be extracted early?

rjbou:
Closing as duplicate of #2715.

Issue #4157, opened 2020-04-24
Opam configuration corrupted on ENOSPC


g2p:
If `opam update --upgrade` encounters errors due to lack of free space, the next `opam upgrade` will attempt to remove all packages.

Looking at the `--debug` output, this is because `~/.opam/repo/repos-config` got corrupted and replaced by an empty file; all packages are considered orphaned after this.

This debug message should probably be upgraded to a user-visible warning (or error, if actively used):

```
00:00.251  STATE                  Ignoring globally-selected repository "default", no configured repository by this name found
```

Quick fix for anyone hitting this:

```
opam repository add default --set-default https://opam.ocaml.org
``` 

```
# opam config report
# opam-version      2.0.7 
# self-upgrade      no
# system            arch=x86_64 os=linux os-distribution=ubuntu os-version=20.04
# solver            builtin-mccs+glpk
# install-criteria  -removed,-count[version-lag,request],-count[version-lag,changed],-changed
# upgrade-criteria  -removed,-count[version-lag,solution],-new
# jobs              7
# repositories      
# pinned            0
# current-switch    4.10.0+fp+flambda
```

rjbou:
Thanks for the report. Opam tries to keep a consistent state when a command or an action fail, but when there is no more space, it's hard to write to good configuration.
I'll close this issue, as duplicated of #2715, we can continue the discussion there.

PR #5473, opened 2023-03-10
Write config files atomically

### LINKED ISSUE #2715: opam behavior when disk is full
On 2.0, when doing `opam switch remove foo` while disk is full:

- opam fails with an exception
- the switch is not removed
- `~/.opam/config` is replaced by an empty file

I understand that handling a full disk is complicated, but I would like if it at least didn't made my opam install unusable. :(

rjbou:
To avoid their unfortunate deletion, eg when no more disk space available
fix #2715 #4157 #5440
* [ ] queued on #5417

### DIFF
--- a/src/format/opamFile.ml
+++ b/src/format/opamFile.ml
@@ -449,16 +449,6 @@ module LineFile (X: LineFileArg) = struct
   include MakeIO(IO)
 end
 
-module AtomicLineFile (X: LineFileArg) = struct
-  module IO = struct
-    include LineFileOps(X)
-    let atomic = true
-  end
-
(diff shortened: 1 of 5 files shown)


src/format/opamFile.ml
```diff
@@ -1020,6 +1061,7 @@ end
 module type SyntaxFileArg = sig
   val internal: string
   val format_version: OpamVersion.t
+  val atomic : bool
```
rjbou, line 1064: split in 2 commits ?

rjbou:
Superseed by #5489

Issue #5448, opened 2023-02-21
space savings in opam switches


edwintorok:
>   If your issue concerns a package not building, please report to
>   https://github.com/ocaml/opam-repository/issues or to the package maintainer
>   unless you are confident it is an issue in the opam tool itself.

```
opam config report
# opam config report
# opam-version         2.1.3
# self-upgrade         no
# system               arch=x86_64 os=linux os-distribution=fedora os-version=37
# solver               builtin-mccs+glpk
# install-criteria     -removed,-count[avoid-version,changed],-count[version-lag,request],-count[version-lag,changed],-count[missing-depexts,changed],-changed
# upgrade-criteria     -removed,-count[avoid-version,changed],-count[version-lag,solution],-count[missing-depexts,changed],-new
# jobs                 8
# repositories         1 (http), 3 (version-controlled) (default repo at ffef6370)
# pinned               4 (git)
# current-switch       trunk
[WARNING] Opam packages conf-libcurl.2 and conf-ncurses.1 depend on the following system packages that are no longer installed: libcurl-devel ncurses-devel
# ocaml:native         true
# ocaml:native-tools   true
# ocaml:native-dynlink true
# ocaml:stubsdir       /home/edvint/.opam/trunk/lib/ocaml/stublibs:/home/edvint/.opam/trunk/lib/ocaml
# ocaml:preinstalled   false
# ocaml:compiler       4.13.1```

There is some discussion on space savings by compressing files produced by the compiler.
However there might be also some low-hanging fruit in terms of space savings in .opam itself which seems to contain a lot of redundant or unnecessary files.

```
du -csh ~/.opam/trunk/.opam-switch/sources/
487M	/home/edvint/.opam/trunk/.opam-switch/sources/
487M	total
```

I'm not sure what the purpose of that is (merlin/LSP jump to source, would it work?), but source files are also (sometimes?) available and installed alongside libraries in '~/.opam/trunk/lib'.

The raw sources like that are sometimes very big because they also contain unit test source code, and unit test source input data. And especially for monorepos that contain multiple packages these sources are duplicated N times... (well in my case I have a monorepo with 56 .opam files, so the sources are duplicated up to 56x times, and each source contains ~30MB of test input data too).

*if* the files here serve an essential purpose, I 'd recommend:
* keeping only relevant source files (e.g. based on file extension, or based on what files are referenced by .cmt* files/etc. to account for preprocessing and renaming files via `copy#`)
[...]

PR #5474, opened 2023-03-10
test: add sources directory cleaning test

rjbou:
related #5448

### DIFF
--- a/tests/reftests/clean.test
+++ b/tests/reftests/clean.test
@@ -1,4 +1,5 @@
 N0REP0
+### : untracked option :
 ### <pkg:things.1>
 opam-version: "2.0"
 build : [ "touch" "first" "second" ]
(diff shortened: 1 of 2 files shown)


tests/reftests/clean.test
```diff
@@ -108,3 +109,160 @@ Remaining files:
...
+# Return code 1 #
+### opam remove waste
+The following actions will be performed:
+=== remove 1 package
+  - remove waste 1
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> removed   waste.1
+Done.
+### test -d OPAM/clean/.opam-switch/sources/waste
```
kit-ty-kate, line 153:
```suggestion
### ls OPAM/clean/.opam-switch/sources
```

tests/reftests/clean.test
```diff
@@ -109,3 +110,175 @@ Remaining files:
...
+The following actions will be performed:
+=== install 1 package
+  - install waste 1
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> retrieved waste.1  (file://${BASEDIR}/waste.tgz)
+-> installed waste.1
+Done.
+### ls OPAM/clean/.opam-switch/sources/
+waste.1
```
dra27, line 158: This one's part of the original bug, right? After opam install, there's no need for it to be there?

tests/reftests/clean.test
```diff
@@ -109,3 +110,175 @@ Remaining files:
...
+The following actions will be performed:
+=== recompile 1 package
+  - recompile waste 1
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> retrieved waste.1  (cached)
+-> removed   waste.1
+-> installed waste.1
+Done.
+### ls OPAM/clean/.opam-switch/sources
```
dra27, line 187: Correct the difference here is that `opam install waste` has left the sources directory populated, but `opam reinstall waste` has removed it?

tests/reftests/clean.test
```diff
@@ -109,3 +110,175 @@ Remaining files:
...
+### opam remove waste
+The following actions will be performed:
+=== remove 1 package
+  - remove waste 1
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> removed   waste.1
+Done.
+### opam install ./waste
```
dra27, line 196:
```suggestion
### ls OPAM/clean/.opam-switch/sources
### opam install ./waste
```
to be sure that the `opam remove` didn't re-extract, or unneccessary? It seems worth terminating each test case with the explicit ls?

tests/reftests/clean.test
```diff
@@ -109,3 +110,175 @@ Remaining files:
...
+=== install 1 package
+  - install waste 1 (pinned)
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> retrieved waste.1  (file://${BASEDIR}/waste)
+-> installed waste.1
+Done.
+### ls OPAM/clean/.opam-switch/sources
+waste
+### opam unpin waste
```
dra27, line 241: It feels like there should be more `ls` checks after these tests, confirming that exactly when the `waste` directory should still be there?

dra27 (commented):
Apparently for me it's much easier to suggest to _other_ people to add comments to their tests, than it is to remember to put comments in _my_ tests in the first place.

I think we should definitely tag the cases which are errors (I think at least the remaining waste.1 after `opam install` is an error??) and we probably can't have too many `ls` calls in this test!

tests/reftests/clean.test
rjbou: yes

kit-ty-kate (approved):

kit-ty-kate:
Thanks!

PR #5475, opened 2023-03-10
Failing switch reinstall no more drop switch layout

### LINKED ISSUE #5347: Answering "no" to "opam switch reinstall" will delete all the files in your switch directory
Let's take this simple Dockerfile as a base:
```
FROM ocaml/opam:debian-11-ocaml-4.14
RUN ls .opam/4.14/lib
```
the last command shows that the `lib` directory inside the current switch is not empty and does contain ocaml and so on.
Now call these commands one at a time inside the resulting docker container and answer `n` to the question `Proceed with 8 installations? [y/n]`:
```
unset OPAMCONFIRMLEVEL
unset OPAMYES
unset OPAMPRECISETRACKING
unset OPAMERRLOGLEN
opam switch reinstall
```
Next, look into `.opam/4.14/lib/` the directory is empty (as well as all the other directories in the switch)

Tested with opam master (https://github.com/ocaml/opam/commit/2c8f6c1461e921892f1570f0ed7824185463e96c)

rjbou:
It rolls back to previous state when opam reinstall fails
fix #5347
It is done by cleaning the switch, and saving a copy in a temporary directory. Open to discussion if there is better ways to recover.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -129,6 +129,7 @@ users)
 ## Reftests
 ### Tests
   * Add switch removal test: failure on removal linked switch [#6276 @btjorge]
+  * Add switch reinstall test with failures [#5475 @rjbou]
 
 ### Engine
 
--- a/master_changes.md
+++ b/master_changes.md
@@ -39,6 +39,7 @@ users)
 
 ## Switch
   * [BUG] Fix `opam switch remove <dir>` failure when it is a linked switch [#6276 @btjorge - fix #6275]
+  * [BUG] Failing switch reinstall partially delete switch layout [#5475 @rjbou - fix #5347]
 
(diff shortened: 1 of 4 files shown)


kit-ty-kate (changes requested):
I'm not sure i like this fix. I think copying a whole switch around is asking for users to just Ctrl-C because it takes too long on some systems and give up on the command. More issues can also appear if the users get annoyed and type Ctrl-C several times at the wrong time (the switch can still be removed, same as before)

I'm not sure why the `switch reinstall` subcommand uses `switch import` internally. This is the reason why the fix takes this approach. However I think we should instead do the equivalent of `opam reinstall st.installed`

PR #5476, opened 2023-03-10
Support atomic environment updates in OpamFile

dra27:
Extracted from #5417, which now just deals with the fully revertible environment updates part.

- [x] Writing the environment file in `opam env` is still taking the switch write lock, which isn't necessary
  - we keep taking write lock as something is written in switch
- [x] Add hook specific read-only option: it is like read-only, but permit environment files writing.
  - We want to keep hooks as readonly. The only downside is that when a local switch is moved, the hook takes some times (switch loading) to recompute environment ; prompt will take that time until an opam commands in the terminal with write lock writes the new environment
    - We can still add the hook-read-only (read-only except for env) in the future, when it makes sense to move a switch
- [x] Test for moving a switch (verify `PATH` updates)

### DIFF
--- a/src/format/opamFile.ml
+++ b/src/format/opamFile.ml
@@ -76,6 +76,7 @@ end
 
 module type IO_Arg = sig
   val internal : string
+  val atomic : bool
   type t
   val empty : t
   val of_channel : 'a typed_file -> in_channel  -> t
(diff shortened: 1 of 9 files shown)


rjbou (approved):
lgtm!
I updated api changes.

dra27:
Test added - up to you whether we want to deal with the lock before or after merge, @rjbou?

rjbou:
From dev meeting: add hook specific option to permit read-only except for environment file ; it resolves switch lock issue in the same time as there is no clash possible on environment files name (temporary files)

src/format/opamFile.ml
```diff
@@ -76,6 +76,7 @@ end
 
 module type IO_Arg = sig
   val internal : string
+  val atomic : bool
```
kit-ty-kate, line 79: Is there a downside to always having all file writes as atomic?

rjbou: Atomic takes few more steps: first write on tmp, then rename on final location, instead of directly writing on final location.
related #5473

kit-ty-kate: Yes but that's fine. I don't think i'm able to see any usecase where this isn't better

kit-ty-kate:
Thanks

PR #5480, opened 2023-03-13
config report: add invariant and compiler packages

### LINKED ISSUE #5478: Include invariant packages in `opam config report` and error messages
Following on from a [dev meeting discussion](https://github.com/ocaml/opam/wiki/2023-Developer-Meetings#backlog), thanks #5208 we should reliably be able to include the invariant packages in the output of `opam config report` and also with error messages. This has the benefit of including the compiler version and configuration for user reports.

Ideally, we'd filter out `{post}` dependencies - i.e. including `ocaml-base-compiler.5.0.0` or `ocaml-variants.5.0.0+options` + `ocaml-options-bytecode-only` but excluding `base-unix`, `base-threads`, etc., simply for brevity.

rjbou:
See test for output
fix #5478


src/client/opamCommands.ml
```diff
@@ -1456,6 +1456,14 @@ let config cli =
               );
             print "current-switch" "%s"
               (OpamSwitch.to_string state.switch);
+            print "invariant" "%s"
+              (OpamFormula.to_string state.switch_invariant);
+            print "invariant compiler" "%s"
```
kit-ty-kate, line 1461:
```suggestion
            print "invariant-compiler" "%s"
```
To follow convention

kit-ty-kate (approved):

PR #5481, opened 2023-03-14
Rebuild the reftest roots if root version changes

dra27:
Switching between branches prior to the 2.2 bump, the roots for reftests in _build end up either too new or requiring upgrade. This little tweak makes them depend on `OpamFile.Config.root_version`


rjbou (approved):
Thanks!

Issue #5487, opened 2023-03-21
Read-only access to the opam 2.2 state fails when opam-state 2.1.* is used


kit-ty-kate:
```
#=== ERROR while compiling opam-dune-lint.0.2 =================================#
# context              2.2.0~alpha~dev | linux/x86_64 | ocaml-base-compiler.4.14.1 | file:///home/opam/opam-repository
# path                 ~/.opam/4.14/.opam-switch/build/opam-dune-lint.0.2
# command              ~/.opam/opam-init/hooks/sandbox.sh build dune build -p opam-dune-lint -j 255 @install @runtest
# exit-code            1
# env-file             ~/.opam/log/opam-dune-lint-7-c1162b.env
# output-file          ~/.opam/log/opam-dune-lint-7-c1162b.out
### output ###
# File "tests/test_vendoring.t", line 1, characters 0-0:
# File "tests/test_empty_dune.t", line 1, characters 0-0:
# File "tests/test_dune.t", line 1, characters 0-0:
#          git (internal) (exit 1)
# (cd _build/.sandbox/3887dc4e8b1d66b4fd37e214ba236a87/default && /usr/bin/git --no-pager diff --no-index --color=always -u ../../../default/tests/test_dune.t tests/test_dune.t.corrected)
# diff --git a/../../../default/tests/test_dune.t b/tests/test_dune.t.corrected
# index 35227f0..c5e6ec7 100644
# --- a/../../../default/tests/test_dune.t
# +++ b/tests/test_dune.t.corrected
# @@ -33,23 +33,20 @@ Replace all version numbers with "1.0" to get predictable output.
#  Check that the missing libraries are detected:
#  
#    $ opam-dune-lint </dev/null
# -  test.opam: changes needed:
# -    "fmt" {>= "1.0"}                         [from /]
# -    "bos" {with-test & >= "1.0"}             [from /]
# -    "opam-state" {with-test & >= "1.0"}      [from /]
# -  Note: version numbers are just suggestions based on the currently installed version.
# -  Run with -f to apply changes in non-interactive mode.
# -  [1]
# +  [ERROR] Refusing write access to /home/opam/.opam, which is more recent than this version of opam (2.2~alpha > 2.1), aborting.
# +  opam-dune-lint: internal error, uncaught exception:
# +                  OpamStd.OpamSys.Exit(15)
# +                  
# +  [125]
```
[...]

PR #5488, opened 2023-03-21
[API] Make the ?lock_kind parameters non-optional to avoid breaking the library users after they upgrade their opam root

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/5487

The current code shows issues with our internal uses of such functions, opening as a draft for now and I will point out where the issues are.

### DIFF
--- a/src/client/opamListCommand.ml
+++ b/src/client/opamListCommand.ml
@@ -389,7 +389,9 @@ let apply_selector ~base st = function
        in
        let selections =
          if switch = st.switch then OpamSwitchState.selections st
-         else OpamSwitchState.load_selections st.switch_global switch
+         else
+           OpamSwitchState.load_selections ~lock_kind:`Lock_none
+             st.switch_global switch
        in
        List.fold_left (fun acc f ->
            let name =
--- a/src/state/opamSwitchState.ml
+++ b/src/state/opamSwitchState.ml
@@ -18,11 +18,11 @@ let slog = OpamConsole.slog
 
 open OpamStateTypes
 
-let load_selections ?lock_kind gt switch =
-  OpamStateConfig.Switch.safe_read_selections ?lock_kind gt switch
+let load_selections ~lock_kind gt switch =
+  OpamStateConfig.Switch.safe_read_selections ~lock_kind gt switch
 
-let load_switch_config ?lock_kind gt switch =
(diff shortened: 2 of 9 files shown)


src/state/opamSwitchState.ml
```diff
@@ -256,7 +256,7 @@ let load lock_kind gt rt switch =
     OpamFilename.flock lock_kind (OpamPath.Switch.lock gt.root switch)
   in
   let switch_config = load_switch_config ~lock_kind gt switch in
-  if OpamStateConfig.is_newer_than_self gt then
+  if OpamStateConfig.is_newer_than_self ~lock_kind:`Lock_write gt then
```
kit-ty-kate, line 259:
```suggestion
  if OpamStateConfig.is_newer_than_self ~lock_kind gt then
```

kit-ty-kate (commented):
Shouldn't it be the same lock_kind as the one given as argument?

src/client/opamListCommand.ml
```diff
@@ -351,7 +351,7 @@ let apply_selector ~base st = function
        in
        let selections =
          if switch = st.switch then OpamSwitchState.selections st
-         else OpamSwitchState.load_selections st.switch_global switch
+         else OpamSwitchState.load_selections ~lock_kind:`Lock_write st.switch_global switch
```
kit-ty-kate, line 354: Should this one have `Lock_read` here instead?

rjbou, line 354:
```suggestion
         else OpamSwitchState.load_selections ~lock_kind:`Lock_none st.switch_global switch
```

src/client/opamListCommand.ml
```diff
@@ -389,7 +389,9 @@ let apply_selector ~base st = function
        in
        let selections =
          if switch = st.switch then OpamSwitchState.selections st
-         else OpamSwitchState.load_selections st.switch_global switch
+         else
+           OpamSwitchState.load_selections ~lock_kind:`Lock_none
```
rjbou, line 393:
```suggestion
           (* keep synchronised with OpamCommands.list switch loading lock *)
           OpamSwitchState.load_selections ~lock_kind:`Lock_none
```

rjbou (approved):

PR #5492, opened 2023-03-24
Fix parsing of OPAMFETCH (support quotes / proper POSIX shell syntax)

### LINKED ISSUE #5490: Tokenize OPAMFETCH value properly, not splitting by whitespaces
```
# opam config report
# opam-version         2.2.0~alpha~dev (7f4558d65bbfbf4524e413ab7d06d8ac5f4484d3)
# self-upgrade         no
# system               arch=x86_64 os=macos os-distribution=homebrew os-version=13.2.1
# jobs                 7
# repositories         1 (http), 1 (version-controlled) (default repo at 5811dd22)
# pinned               0
# current-switch       5.0
# invariant            ocaml-base-compiler >= 5.0 & ocaml-base-compiler < 5.1
# compiler-packages    ocaml-base-compiler.5.0.0, ocaml-options-vanilla.1
# ocaml:native         true
# ocaml:native-tools   true
# ocaml:native-dynlink true
# ocaml:stubsdir       /Users/mukka/.opam/5.0/lib/ocaml/stublibs:/Users/mukka/.opam/5.0/lib/ocaml
# ocaml:preinstalled   false
# ocaml:compiler       5.0.0
```

Because OPAM splits OPAMFETCH value by spaces, the use-case below does not work as expected.

```sh
# The package is in private Github repo, so we supply access token in Authorize header.
$ OPAMFETCH='wget --header "Authorization: token >>>redacted<<<" %{url}% -O %{out}%' opam install coreutil

... snip ...
+ /usr/bin/wget "--header" "\"Authorization:" "token" ">>>redacted<<<\"" "-O" "/home/scjung/.opam/4.12.1/.opam-switch/sources/coreutil.1.10/v1.10.tar.gz.part" "%{url}"
- --2023-03-24 10:31:27--  http://token/
- Resolving token (token)... failed: Name or service not known.
- wget: unable to resolve host address ‘token’
```

As you may see in the log, ```"Authorization: token ..."``` argument is parsed into seperate words.

OPAM should tokenize OPAMFETCH value in the way shell tokenizes[1, 2], so that a quoted argument is parsed as a single string rather than separate words.

[1] https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html
[2] https://github.com/dbuenzli/bos/blob/master/src/bos_cmd.ml#L46-L112

kit-ty-kate:
Fixes #5490


scjung:
It works! Thanks for the fix 👍 

```
Processing  1/3: [coreutil.1.70: http]
+ /usr/local/bin/wget "--header" "Authorization: token >>>redacted<<" "https://github.com/>>redacted<</coreutil/archive/refs/tags/v1.70.tar.gz" "-O" "/Users/mukka/.opam/5.0/.opam-switch/sources/coreutil.1.70/v1.70.tar.gz.part"
- --2023-03-27 09:18:43--  https://github.com/>>redacted<</coreutil/archive/refs/tags/v1.70.tar.gz
- Resolving github.com (github.com)... 20.200.245.247
- Connecting to github.com (github.com)|20.200.245.247|:443... connected.
- HTTP request sent, awaiting response... 302 Found
```

src/core/opamStd.mli
```diff
@@ -260,6 +260,12 @@ module List : sig
   val fold_left_map: ('s -> 'a -> ('s * 'b)) -> 's -> 'a list -> 's * 'b list
 end
 
+module Char : sig
+
+  val is_whitespace : char -> bool
```
dbuenzli, line 265: What about rather following what is being proposed [upstream](https://github.com/ocaml/ocaml/pull/13695). That way, in a few years, you can likely simply delete the code. In other words: 

```ocaml
module Char : sig 
  include Stdlib.Char
  module Ascii : sig 
    val is_white : char -> bool
  end
end

Comment (kit-ty-kate) [reply src/core/opamStd.mli:265]: I knew about the PR but I didn't use the same interface at the time for several reasons:
* the two definitions are different (e.g. `\v` and `\f` are not handled by the opam definition but are with the stdlib one)
* it wasn't merged at the time and we have a dedicated module for compatibility functions (https://github.com/ocaml/opam/blob/master/src/core/opamCompat.ml) so it would've felt a bit weird to have a yet-to-be-part-of-the-stdlib function in it
* OpamStd isn't meant to be a mirror of the stdlib in terms of naming, OpamCompat is

Now that it has been merged i can think about it again although i personally feel like it would be better to wait for the release of 5.4 to change the function. In the meantime i'm happy to add a TODO comment to remind ourselves to change it when 5.4 is released.

File: src/core/opamCmd.mli [hunk 2]
```diff
@@ -0,0 +1,13 @@
...
+(*                                                                        *)
+(*    Copyright 2025 Kate Deplaix                                         *)
+(*                                                                        *)
+(*  All rights reserved. This file is distributed under the terms of the  *)
+(*  GNU Lesser General Public License version 2.1, with the special       *)
+(*  exception on linking described in the file LICENSE.                   *)
+(*                                                                        *)
+(**************************************************************************)
+
+val of_string : string -> string list
```
rjbou, line 11: The new module should be in its own commit.

src/core/opamCmd.ml
```diff
@@ -0,0 +1,187 @@
...
+(*    Copyright 2025 Kate Deplaix                                         *)
+(*    Copyright 2015 The bos programmers                                  *)
+(*                                                                        *)
+(*  All rights reserved. This file is distributed under the terms of the  *)
+(*  GNU Lesser General Public License version 2.1, with the special       *)
+(*  exception on linking described in the file LICENSE.                   *)
+(*                                                                        *)
+(**************************************************************************)
+
+(* NOTE: Inspired from @dbuenzli's astring library *)
```
rjbou, line 12: Maybe it worth adding an url ?

src/core/opamCmd.mli
rjbou, line 11: Is there a reason to add a new module instead of adding it in `OpamStd`?

src/core/opamShellCommand.mli
```diff
@@ -0,0 +1,13 @@
...
+(*                                                                        *)
+(*    Copyright 2025 Kate Deplaix                                         *)
+(*                                                                        *)
+(*  All rights reserved. This file is distributed under the terms of the  *)
+(*  GNU Lesser General Public License version 2.1, with the special       *)
+(*  exception on linking described in the file LICENSE.                   *)
+(*                                                                        *)
+(**************************************************************************)
+
+val of_string : string -> string list
```
kit-ty-kate: I prefer having both usage and definition in the same commit

kit-ty-kate: In my opinion `OpamStd` should only be an overlay over the standard library, a `Cmd` module doesn't fit that. Also the definition is big enough to warrant its own module, and in the future i'd also like to add more functions to this module like `to_string`.

src/core/opamCompat.ml
```diff
@@ -8,6 +8,19 @@
...
+  [@@@warning "-32"]
+
+  module Ascii = struct
+
+    (** NOTE: OCaml >= 5.4.0 *)
+    let is_white = function ' ' | '\t' .. '\r'  -> true | _ -> false
+
+  end
+
+  include Stdlib.Char
```
kit-ty-kate: This doesn't compile with OCaml >= 5.4

PR #5494, opened 2023-03-24
src_ext: Remove unused/untested Makefile targets

kit-ty-kate:
I don't think anyone use that, and we're not testing it either.
Also it complicates vendoring of dependencies so I feel like any simplification in there is always welcome.


dra27:
Slightly in passing - I'm completely for removing this, but preferably after 2.2 (i.e. after there's a released Windows opam which does away with the use case for it). It is useful (and used, certainly by me!) for developing opam on Windows, because the build is a lot faster than the vendored route when working on opam itself. It's also in use in the [base image builder](https://images.ci.ocaml.org) at the moment. It is also tested - see https://github.com/dra27/opam/commit/ee6e3a5d60d37aeb1406a161874fedb41af35bd6 which demonstrates the breakage this commit presently causes in the Cold job (the Cold job in the PR is passing because #4776 means that the build is quietly downloading the dependencies which haven't been installed).

dra27 (approved):
RIP, with fond memories of the workshop in Princeton where I was testing most of this 🙂

PR #5496, opened 2023-03-29
Add `package` stanza on all rules that depend on opamMain.exe.exe

Leonidas-from-XIV:
The compilation fails if

```sh
$ dune build --only-packages opam-core,opam-format,opam-repository
```

As it tries to build the binary, but can't due to the opam package not being compiled. This adds the `package` stanza to the associated rules to avoid the rules being triggered when the `opam` package is not selected.


rjbou:
Thanks!

PR #5498, opened 2023-03-31
Replace usage of CPPO by dune features and small helper

Leonidas-from-XIV:
To ease vendoring reducing the amount of dependencies is helpful, as such this PR removes the usage of CPPO in favor of generating some files via dune and implementing a very simple CPPO standin that only does conditional compilation dispatching on the OCaml version (which is the only feature required).

This doesn't yet remove the CPPO dependency, I'll remove the code that if it is decided that this PR is a decent approach to take. Looking forward to your comments.


src/core/gen/gen.ml
```diff
@@ -0,0 +1,67 @@
...
+    match String.trim line with
+    | "#else" -> loop (not output) ic
+    | "#endif" -> loop true ic
+    | _otherwise -> (
+      match Re.exec_opt if_ocaml_version line with
+      | None -> (
+        (* not a condition, it's source code *)
+        (match output with
+        | true -> print_endline line
+        | false -> ());
```
kit-ty-kate, line 46:
```suggestion
        if output then
          print_endline line;
```

src/core/gen/gen.ml
```diff
@@ -0,0 +1,67 @@
...
+  group (rep1 digit); char '.'; group (rep1 digit); char '.'; group (rep1 digit)
+  ])
+  |> Re.compile
+
+let current_version =
+  Scanf.sscanf Sys.ocaml_version "%u.%u.%u" (fun a b c -> (a, b, c))
+
+let greater_or_equal (major', minor', patch') =
+  let major, minor, patch = current_version in
+  match major > major', major = major with
```
kit-ty-kate, line 21: The prime was forgotten
```suggestion
  match major > major', Int.equal major major' with
```

src/core/gen/gen.ml
```diff
@@ -0,0 +1,67 @@
...
+  | true, _ -> true
+  | false, false -> false
+  | false, true -> (
+      match minor > minor', minor = minor' with
+      | true, _ -> true
+      | false, false -> false
+      | false, true -> (
+        match patch > patch', patch = patch' with
```
kit-ty-kate, line 29:
```suggestion
        match patch > patch', Int.equal patch patch' with
```

src/core/gen/gen.ml
```diff
@@ -0,0 +1,67 @@
...
+        | true, _ -> true
+        | false, false -> false
+        | false, true -> true))
```
kit-ty-kate, line 32: You can simplify this using polymorphic comparison on tuples (safe as long as you set the type somewhere):
```suggestion
let greater_or_equal v = (v : (int * int * int)) >= current_version
```

src/core/gen/gen.ml
```diff
@@ -0,0 +1,67 @@
...
+
+let rec loop output ic =
+  match input_line ic with
+  | line -> (
+    match String.trim line with
+    | "#else" -> loop (not output) ic
+    | "#endif" -> loop true ic
+    | _otherwise -> (
+      match Re.exec_opt if_ocaml_version line with
+      | None -> (
```
kit-ty-kate, line 42:
```suggestion
      | None ->
```

src/core/gen/gen.ml
```diff
@@ -0,0 +1,67 @@
...
+        (* not a condition, it's source code *)
+        (match output with
+        | true -> print_endline line
+        | false -> ());
+        loop output ic
+        )
```
kit-ty-kate, line 48:
```suggestion
```

src/core/gen/gen.ml
```diff
@@ -0,0 +1,67 @@
...
+      | Some groups -> (
+        let group = Re.Group.get_opt groups in
+        match group 1, group 2, group 3 with
+        | Some major, Some minor, Some patch ->
+          let major = int_of_string major in
+          let minor = int_of_string minor in
+          let patch = int_of_string patch in
+          let required = major, minor, patch in
+          let output = greater_or_equal required in
```
kit-ty-kate, line 57:
```suggestion
          let output = greater_or_equal (major, minor, patch) in
```

kit-ty-kate (commented):
Also cppo wasn't removed from:
* `opam-core.opam`
* `configure.ac`
* `src_ext/*`
* `.gitignore`

Looks good overall as a concept though.

Leonidas-from-XIV:
As I was removing CPPO from the files you mentioned, I realized that Extlib depends on CPPO, but looking through the source code, I can't really find any usage of Extlib - on the contrary, I can only find conflicts with older versions of Extlib.

It looks like the Extlib dependency was removed from the code but the auxillary files still reference it. Shall I remove that too and if so, in this PR or another?

Leonidas-from-XIV:
Ok, as I was digging deeper to remove extlib, I realized it is required for CUDF and dose3. These could probably be removed but it is a larger undertaking.

So I would suggest the following for now: remove CPPO usage in OPAM for now, but vendor it for the dependencies. That way it can be removed finally once these dependencies remove Extlib or Extlib removes CPPO.

Leonidas-from-XIV:
True; that's a good point, I've removed the dependency. Squashed all commits and rebased.

src/core/dune
```diff
@@ -14,15 +14,28 @@
...
                (:include ../ocaml-flags-configure.sexp)
                (:include ../ocaml-context-flags.sexp)))
-  (preprocess  (per_module
-                  ((action (run %{bin:cppo} -V OCAML:%{ocaml_version} %{input-file})) opamCompat opamSystem)
-                  ((action (run %{bin:cppo} %{read-lines:developer} %{input-file})) opamCoreConfig)
-                  ((action (run %{bin:cppo} -D "VERSION %{read-lines:version}" %{input-file})) opamVersion)))
   (wrapped     false))
 
+(rule
+  (with-stdin-from opamCompat.pp.ml
```
hannesm, line 20: since `pp` is conventionally used for ppx AFAIU, why did you choose `.pp.ml` here? I'd suggest to use for example `opamCompat.ml.in` instead.

src/core/gen/gen.ml
```diff
@@ -0,0 +1,51 @@
...
+        let group = Re.Group.get_opt groups in
+        match group 1, group 2, group 3 with
+        | Some major, Some minor, Some patch ->
+          let major = int_of_string major in
+          let minor = int_of_string minor in
+          let patch = int_of_string patch in
+          let output = greater_or_equal (major, minor, patch) in
+          loop output ic
+        | _ ->
+          loop output ic
```
hannesm, line 44: your "error handling" seems to "ignore any errors", which does contradict rules of least surprise. of course for the current state of opamCompat.ml this is fine, but I wonder how hard and complex it now becomes to extend that OCaml modiule, given this ad-hoc script?

hannesm: sorry for the noise, please ignore my comment.

src/core/gen/gen.ml
```diff
@@ -0,0 +1,75 @@
...
+
+let should_output = List.hd
+
+let pop = List.tl
+
+let if_statement s =
+  if String.length s >= 3 then
+    let if' = String.sub s 0 3 in
+    String.equal if' "#if"
+  else false
```
kit-ty-kate, line 33:
```suggestion
  String.length s >= 3 &&
  String.equal (String.sub s 0 3) "#if"
```

kit-ty-kate (commented):
Suggestion to make it more compact and fix indentation

rjbou (approved):

kit-ty-kate:
Thanks!

PR #5499, opened 2023-03-31
depexts: reword message

AltGr:
I just noticed that "attempt installation anyway" could be misleading because of
the ambiguity between "system installation" and "opam installation".

"Continue anyway" is much more obvious, esp. if you read quickly.


kit-ty-kate:
Thanks

PR #5500, opened 2023-04-03
explicitly connect /dev/tty. Allows running via sh

sabine:
It should be super simple and quick to install opam on various Linux distros and macOS. Some of the targeted OSes might not have the bash command needed for the invocation
```
bash -c "sh <(curl -fsSL https://raw.githubusercontent.com/ocaml/opam/master/shell/install.sh)"
```

Instead, `install.sh` should be invokable via

```
curl -fsSL https://raw.githubusercontent.com/ocaml/opam/master/shell/install.sh | sh
```

However, a problem with that is that, when piping the downloaded script into `sh`, `/dev/tty` is not connected, so we cannot read from stdin.

This patch remedies this issue by changing `install.sh` to explicitly `read` from `/dev/tty`. It also factors out the prompting into a function `prompt`.

Note: if any of the other commands called by `install.sh` require user interaction via stdin, they should also be explicitly connected to `/dev/tty`. I have **not** checked whether they do!

### DIFF
--- a/shell/install.sh
+++ b/shell/install.sh
@@ -364,15 +364,25 @@ usage() {
     echo "    --restore   VERSION    Restore a backed up opam binary and root"
     echo "    --version   VERSION    Install this specific version instead of $VERSION"
     echo "    --download-only        Download binary in current directory and check its sha512"
+    echo "    --tty                  Allow interactive questions to be asked even if the script is being piped to (e.g. curl https://../install.sh | sh)"
     echo
     echo "The default is to backup if the current version of opam is 1.*, or when"
     echo "using '--fresh' or '--dev'"
+}
+
+read_tty() {
+  if [ "$TTY" = 1 ]; then
+    read -r "$1" < /dev/tty
+  else
+    read -r "$1"
+  fi
 }
 
 RESTORE=
 NOBACKUP=
 FRESH=
 DOWNLOAD_ONLY=
+TTY=
 
 while [ $# -gt 0 ]; do
     case "$1" in
@@ -398,6 +408,8 @@ while [ $# -gt 0 ]; do
             FRESH=1;;
         --download-only)
             DOWNLOAD_ONLY=1;;
+        --tty)
+            TTY=1;;
         --help|-h)
             usage; exit 0;;
         *)
@@ -552,7 +564,7 @@ if [ -n "$RESTORE" ]; then
     fi
     if [ "$NOBACKUP" = 1 ]; then
         printf "## This will clear %s and %s. Continue ? [Y/n] " "$OPAM" "$OPAMROOT"
-        read -r R
+        read_tty R
         case "$R" in
             ""|"y"|"Y"|"yes")
                 xsudo rm -f "$OPAM"
@@ -584,7 +596,7 @@ fi
 
 while true; do
     printf "## Where should it be installed ? [%s] " "$DEFAULT_BINDIR"
-    read -r BINDIR
+    read_tty BINDIR
     if [ -z "$BINDIR" ]; then BINDIR="$DEFAULT_BINDIR"; fi
 
     if [ -d "$BINDIR" ]; then break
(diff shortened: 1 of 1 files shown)


shell/install.sh
```diff
@@ -358,8 +366,8 @@ if [ -n "$RESTORE" ]; then
         exit 1
     fi
     if [ "$NOBACKUP" = 1 ]; then
-        printf "## This will clear $OPAM and $OPAMROOT. Continue ? [Y/n] "
-        read R
+        prompt "## This will clear $OPAM and $OPAMROOT. Continue ? [Y/n] "
+        R=$?
```
kit-ty-kate, line 370: unless I misread something I believe `$?` is always a number
```suggestion
```

sabine: You're correct. This one was an oversight from before I understood that :see_no_evil:

kit-ty-kate:
Do we know if anyone is using the install script in an automated way (e.g. `(echo /usr/bin/ ; yes) | bash -c "sh <(curl -fsSL https://raw.githubusercontent.com/ocaml/opam/master/shell/install.sh)"`)? If some do this change will break them as it would make it very hard to pipe any kind of input to this script.

sabine:
This is a good question. There is the possibility to extend this patch to conditionally connect `/dev/tty` based on a new flag, so that breaking automated scripts could be fixed by passing the new flag to this script. At least, that's what the rustup install script does. 🤔

sabine:
I found a way to conditionally connect `/dev/tty` in a backwards-compatible way by using an environment variable that must be set in order to explicitly connect to `/dev/tty`.

Then, the script can be invoked as:
```
curl -fsSL https://raw.githubusercontent.com/ocaml/opam/master/shell/install.sh | set TTY=1 sh
```
This is backwards-compatible for automated scripts, as explicitly connecting to `/dev/tty` is opt-in instead of opt-out.

shell/install.sh
```diff
@@ -186,6 +186,22 @@ usage() {
     echo "using '--fresh' or '--dev'"
 }
 
+# Conditionally connect /dev/tty when
+# the environment variable TTY is set to 1
+if ${TTY+false}; then
```
kit-ty-kate, line 191: isn't this construct `bash` specific? Also don't forget to put quotes around variables, otherwise people can inject random code using the environment variable

shell/install.sh
```diff
@@ -186,6 +186,22 @@ usage() {
...
+  TTY=0
+fi
+R=
+prompt() {
+  # expects to be called as prompt "Do you want to do this? [Y/n] "
+  printf "$1"
+  if [ $TTY = 1 ]; then
```
kit-ty-kate, line 198:
```suggestion
  if [ "$TTY" = 1 ]; then
```

kit-ty-kate:
> Then, the script can be invoked as:
> 
> ```
> curl -fsSL https://raw.githubusercontent.com/ocaml/opam/master/shell/install.sh | set TTY=1 sh
> ```

`set` does not do anything but I believe what you want is `env` instead (which is POSIX-complient)

shell/install.sh
sabine: The `${TTY+false}` syntax passes https://www.shellcheck.net/ and is advertised as a POSIX shell feature [here](https://www.grymoire.com/Unix/Sh.html#uh-33). But now that I read the explanation, I am puzzled why it would work at all. :thinking:

sabine: I changed the conditional to `TTY="${TTY-0}"` - that is shorter and makes more sense.

kit-ty-kate: ah right nevermind sorry, i thought it was something bash added. It is indeed POSIX-complient and present as an example in the man page for POSIX sh(1): https://www.unix.com/man-page/posix/1p/sh/

kit-ty-kate:
This PR is queued on https://github.com/ocaml/opam/pull/6313. There will be a couple of things to do to make it correct afterwards too, i'll handle it.
I'll also change the `TTY` env variable to a `--tty` argument to make it even more backward compatible and more fitting to the rest of the script. It has also the advantage to guide users on how to pass extra parameters when using the `| sh` method. E.g.:
```
curl -fsSL https://opam.ocaml.org/install.sh | sh -s -- --tty --version 2.2.0
```
and without any extra parameters:
```
curl -fsSL https://opam.ocaml.org/install.sh | sh -s -- --tty
```

rjbou:
Should the documentation be updated too? To advice to use `sh` piping instead of `bash`.

kit-ty-kate:
> Should the documentation be updated too? To advice to use `sh` piping instead of `bash`.

I think it should be done separately to make sure there isn't any issue. But in a second phase yes

rjbou (approved):

kit-ty-kate:
Thanks!

PR #5501, opened 2023-04-03
Run opam init for a fresh install

sabine:
This PR extends #5500 by adding a call to `opam init` in the case of a fresh install.

Before, only in the case of a non-fresh install, `opam init --reinit -ni` was called.


sabine:
I'm not 100% sure this is the right way to go about it. An alternative would be to prompt `Run opam init now? [Y/n]`.

The intent behind this is to make installing `opam` and setting it up more obvious/faster for newcomers by railroading them through the obvious steps.

kit-ty-kate:
> An alternative would be to prompt `Run opam init now? [Y/n]`.

i'd much prefer this solution. I'll do that once #5500 is merged

Issue #5502, opened 2023-04-04
Up to 4x performance regression compared to c8d7c19


kit-ty-kate:
https://github.com/ocaml/opam/pull/5208 made it so that the dependency cone of the switch invariant is recomputed everytime the switch state changes.

The issues is that everytime a package is installed or removed, the switch state is changed so the invariant is recomputed.

This causes a sizeable performance regression for `opam remove` and `opam upgrade` (and less sizeable but still there for `opam install`). For example:
* with c8d7c19 `opam remove -a patdiff` takes 8 seconds on my machine
* after the merge of https://github.com/ocaml/opam/pull/5208 it now takes 30 seconds
* I've tried locally to remove the call to `OpamSwitchState.universe` from `OpamSwitchState.dependencies` but it still takes twice the time as before (15 seconds) due to `OpamSwitchState.dependencies` alone (taking itself between 0.1 and 0.2 seconds per call)

Also, I'm not sure how related this is to the performance issue but I've also noticed that the last dynamic line during `opam install/remove` (e.g. `Processing 80/209: [dune: ocaml 9]`) kept blinking and wasn't visible with master (https://github.com/ocaml/opam/commit/0a174e1f8ebcfb9fc885c1ab823eca4b38c0d56c) but was always there with c8d7c19.
If this isn't related to the current problem we should create a separate issue to track this.

PR #5503, opened 2023-04-04
Fix performance regression in opam install/remove/upgrade/reinstall

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/5502 (including the dynamic line blinking described at the end)

The current state of the PR is quite rough but it works fine, the performance is back to pre c8d7c19 level and the testsuite passes.


src/state/opamSwitchState.ml
```diff
@@ -1316,26 +1318,30 @@ let dependencies_filter_to_formula_t ~build ~post st nv =
...
+    else if unavailable then st.packages
+    else Lazy.force st.available_packages
   in
   log ~level:3 "dependencies packages=%a"
     (slog OpamPackage.Set.to_string) packages;
   let timer = OpamConsole.timer () in
   let base_depends =
     let filter = base_deps_compute base in
+    let get_deps = get_deps st ~force_dev_deps:false ~test:false ~doc:false ~dev_setup:false ~requested_allpkgs:packages in
+    let opams = OpamPackage.Set.fold (fun pkg opams -> OpamPackage.Map.add pkg (OpamPackage.Map.find pkg st.opams) opams) base OpamPackage.Map.empty in
```
kit-ty-kate, line 1336: is there not a function to do that somewhere already? It feels possibly like a very common pattern in opam's code but I wasn't able to find an example

tests/reftests/resolve-variables.test
```diff
@@ -14,7 +14,6 @@ The following actions will be performed:
 
 <><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
 -> installed foo.1
-[WARNING] Undefined filter variable build in dependencies of foo.1
```
kit-ty-kate, line 17: TODO: these output should stay

rjbou (approved):

tests/reftests/resolve-variables.test
rjbou: Nop, it's ok

AltGr (approved):
Nice, LGTM!

PR #5507, opened 2023-04-12
Fix linting on opam-crowbar.opam

kit-ty-kate:


rjbou (approved):
Updated synopsis. Thanks

Issue #5504, opened 2023-04-05
"resolution of the request timed out" even with OPAMSOLVERTIMEOUT=1200


MdeLv:
I have an old ocsigen.2.19.3 configuration that I don't want&need to upgrade.

After the problem described at https://github.com/ocaml/opam/issues/5485, it was destroyed but with some tweak, I could save it with a `opam switch export` and  built it with `opam switch import`.

I've the following error message with the package `ocsigen-ppx-rpc` requires by `ocsigen-start`:
```
$ opam install ocsigen-ppx-rpc
[ERROR] Sorry, resolution of the request timed out.                                                       
        Try to specify a more precise request, use a different solver, or increase the allowed time by    
        setting OPAMSOLVERTIMEOUT to a bigger value (currently, it is set to 60.0 seconds).               
'opam install ocsigen-ppx-rpc --verbose' failed.                                                          
```
Using `OPAMSOLVERTIMEOUT=1200` or `OPAMSOLVERTIMEOUT=1200.0` gives nothing more.

Using flag `--verbose` gives **NO MORE** details.

For unknown reason, the existing and working `ocsigen-start.2.19.3` webapp was expecting `ocsigen-ppx-rpc.1.0` for compilation while this package is **not** a dependency (it is an ocsigen-start dependecy from 4.6.0). See https://opam.ocaml.org/packages/ocsigen-ppx-rpc/ .

I suspect that it may have been caused by opam tool and opam repository that are not enough strict/demanding regarding declared opam files. ITO, one can declare a wrong opam file.
See [TODO : find previous discussion about that.]

btw, what is the recommended command/way to produce the **real** dependencies of a package?
[ocamldep](https://v2.ocaml.org/manual/depend.html) gives only references to external compilation units.
 
Possibly related issues:
https://github.com/ocaml/opam/issues/3447
https://github.com/ocaml/opam/issues/4203
https://github.com/ocaml/opam/issues/3447
https://github.com/ocaml/opam/issues/685

kit-ty-kate:
This is a duplicate of https://github.com/ocaml/opam/issues/5353.

I've opened https://github.com/ocaml/opam/pull/5510. With this, setting `OPAMSOLVERTOLERANCE=0.0003` or above will fix your issue at hand.

MdeLv:
Thanks for this useful feedback.

Side question:
[...]

Issue #5353, opened 2022-11-15
Infinit loop in the mccs solver


kit-ty-kate:
* Using opam master (https://github.com/ocaml/opam/commit/061d7305ce1f555c6ead0c81c5d8874504426b39):
* Using opam-repository commit https://github.com/ocaml/opam-repository/commit/03cdcda5b3a74772bd5f92ff9bcfb1b1310ceaf3
```
$ opam switch create 4.08 ocaml-base-compiler.4.08.1
$ OPAMSOLVERTIMEOUT=6666666 opam install ocsigen-toolkit.2.10.1
```
Even after 2 hours, the default solver (builtin-mccs+glpk) seems stuck. Here are the latest debug log from mccs (using `--debug-level=2`):
```
Can reduce graph.
Initial size: 2157 packages (8 installed, 2149 uninstalled), 269 virtual packages
Final size: 2150 packages (8 installed, 2142 uninstalled), 226 virtual packages
Constructing initial basis...
Number of 0-1 knapsack inequalities = 7126
Constructing conflict graph...
Conflict graph has 1370 + 453 = 1823 vertices
```
but nothing after. @AltGr do you know how to debug this?

Issue #4794, opened 2021-08-11
Timeout trying to create a switch with slack-backup (an old package)


Armael:
@kit-ty-kate prompted me to report this here.

Trying to create a fresh switch with the package slack-backup currently results in a solver timeout. At the time of writing, the HEAD of opam-repository is 7d8c924.

The exact command I'm running is:
```
OPAMSOLVERTIMEOUT=120 opam switch create --packages slack-backup,ocaml-base-compiler.4.02.3 slack-backup
```

kit-ty-kate:
I'll add that this was tested with opam 2.1.0 and I can also reproduce it on a different machine.

rjbou:
Using `OPAMSOLVERTOLERANCE` there is no timeout 🙌

Issue #3447, opened 2018-07-11
Internal opam solver failure


lpw25:
According to `opam ci` there are opam solver failures on two versions of the prof_spacetime package:

```
$ opam ci log prof_spacetime.0.1.0 --compiler=4.05

====> prof_spacetime.0.1.0 4.05 Debian 9 (Stretch) amd64 (internal opam solver failure) (opam-repository 88fd9cd8):
<no logs available>

====> prof_spacetime.0.1.0 Dockerfile for 4.05 Debian 9 (Stretch) amd64:
# Autogenerated by OCaml-Dockerfile scripts
FROM ocaml/opam2:debian-9-ocaml-4.05
RUN opam switch 4.05
ENV OPAMYES 1
WORKDIR /home/opam/opam-repository
RUN git pull origin master
RUN git checkout 88fd9cd811d22c98d2bd573d053cd5818bdae488
RUN opam update
RUN opam install -y depext
RUN opam depext -iy jbuilder ocamlfind
RUN opam depext -uiv prof_spacetime.0.1.0
```

```
$ opam ci log prof_spacetime.0.1.0 --compiler=4.04

====> prof_spacetime.0.1.0 4.04 Debian 9 (Stretch) amd64 (internal opam solver failure) (opam-repository 88fd9cd8):
<no logs available>

====> prof_spacetime.0.1.0 Dockerfile for 4.04 Debian 9 (Stretch) amd64:
# Autogenerated by OCaml-Dockerfile scripts
FROM ocaml/opam2:debian-9-ocaml-4.04
RUN opam switch 4.04
ENV OPAMYES 1
WORKDIR /home/opam/opam-repository
RUN git pull origin master
RUN git checkout 88fd9cd811d22c98d2bd573d053cd5818bdae488
RUN opam update
RUN opam install -y depext
RUN opam depext -iy jbuilder ocamlfind
RUN opam depext -uiv prof_spacetime.0.1.0
```

I was told it would be useful to report this fact here.

avsm:
Thanks; I suspect this is due to the large number of cohttp packages in the stable opam-repo

rjbou:
Update:
```
$ opam ci logs prof_spacetime 
prof_spacetime: multiple build failures found with different configuration parameters.
Please refine the command to select exactly one of the following:
  opam-ci logs prof_spacetime.0.1.0 --compiler=4.04 --arch=amd64 --distro=debian-9
[...]

PR #5510, opened 2023-04-14
Add the OPAMSOLVERTOLERANCE environment variable to allow users to fix solver timeouts for good

kit-ty-kate:
Partially fixes https://github.com/ocaml/opam/issues/5504
Partially fixes https://github.com/ocaml/opam/issues/3447
Partially fixes https://github.com/ocaml/opam/issues/5353
Partially fixes https://github.com/ocaml/opam/issues/4794
Fixes https://github.com/ocaml/opam/issues/3230
Requires https://github.com/AltGr/ocaml-mccs/pull/40

Optimizing solvers can be an annoying breed, sometimes a problem is just too complex for solver A but easy for solver B, and on some other problem solver B might take too long where solver A takes no time at all.

To partly solve the numerous timeout issues that have been plagued opam, I introduce this new environment variable that allows `builtin-mccs+glpk` (default solver) to not wait around for too long when it found a solution but the solution is not the optimum solution.

The downside of having unoptimized solutions is that opam might install out-of-date packages so it is to be used parsimoniously.

~TODO: add tests~ too complicated to do reliably

### DIFF
--- a/doc/pages/Tricks.md
+++ b/doc/pages/Tricks.md
@@ -150,6 +150,7 @@ The following sequence of commands tries to install as much packages as possible
 opam update
 opam switch create . ocaml-base-compiler.$(VERSION)
 export OPAMSOLVERTIMEOUT=3600
+export OPAMSOLVERTOLERANCE=1.0
 opam list --available -s | xargs opam install --best-effort --yes
 # Be patient
 ```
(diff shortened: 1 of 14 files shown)


AltGr (approved):

rjbou (approved):

rjbou:
:raised_hands:

PR #5511, opened 2023-04-17
Stop the configure script from downloading and vendoring dependencies by default

kit-ty-kate:
Currently if dune is not detected, or a dependency is not detected by the configure script (said configure script can well be outdated, e.g. https://github.com/ocaml/opam/pull/5497), the vendoring mode is enabled by default without asking the user.

I think this is not a good default and if it were to be released now i would probably weigh against merging it in this state in opam-repository as the sandbox does not allow network access and the the general behaviour can become non-deterministic very easily. Other system package managers maintainers would probably also get upset by this behaviour.

I thus propose to simply switch the behaviour to "vendor only when `--with-vendored-deps` is given"

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -402,23 +402,15 @@ AS_IF([test "x${enable_checks}" != "xno" && {
        test "x$OCAML_PKG_swhid_core" = "xno" ||
        test "x$CPPO" = "x" ||
        test "x$OCAML_PKG_mccs$MCCS_ENABLED" = "xnotrue";}],[
-  AS_IF([test "x${with_vendored_deps}" = "xno"],[
-    AC_MSG_ERROR([Dependencies missing and --without-vendored-deps given; override with --disable-checks])
-  ],[
-    AS_IF([test "x${with_vendored_deps}" != "xyes"],[
-      echo "============================================================================"
-      echo "Some dependencies are missing. Vendored mode has been automatically enabled."
-      echo "The sources for opam's dependencies will be extracted in src_ext and built."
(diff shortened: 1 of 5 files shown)


configure.ac
```diff
@@ -402,23 +402,15 @@ AS_IF([test "x${enable_checks}" != "xno" && {
...
-  ],[
-    AS_IF([test "x${with_vendored_deps}" != "xyes"],[
-      echo "============================================================================"
-      echo "Some dependencies are missing. Vendored mode has been automatically enabled."
-      echo "The sources for opam's dependencies will be extracted in src_ext and built."
-      echo "============================================================================"
-    ])
-    AC_SUBST(VENDORED,"true")
+  AS_IF([test "x${with_vendored_deps}" != "xyes"],[
+    AC_MSG_ERROR([Dependencies missing. Use --with-vendored-deps or --disable-checks])
```
rjbou, line 406: I'd highlight more this part, as previous opposite message, surrounded with `====[...]`

rjbou (approved):

configure.ac
kit-ty-kate: I don't think this is needed. The previous message were surrounded by `===` because it was a simple notice and does not end the configure script.

`AC_MSG_ERROR` makes it very clear it is an error and makes the configure script fail. The output will look like this:
```
$ ./configure
[...]
checking whether g++ really is a C++ compiler... no

configure: error: Dependencies missing. Use --with-vendored-deps or --disable-checks.
```

PR #5512, opened 2023-04-17
doc: `?` evaluates to `true` if defined

emillon:
The existing sentence seems to indicate that it evaluates to `true` when it contains undefined variables.

### DIFF
--- a/doc/pages/Manual.md
+++ b/doc/pages/Manual.md
@@ -571,8 +571,9 @@ The following are allowed in filters:
 - Idents
 - Parentheses
 - Logical operators (binary AND `&`, binary OR `|`, prefix, unary NOT `!`)
-- The unary operator `?` for detecting whether an expression contains undefined
-  variables
+- The unary operator `?` for detecting whether a value is defined. `?expr`
+  evaluates to `true` if `expr` evaluates to a defined value, and `false` if it
+  evaluates to an undefined value.
 - Binary relational operators (`=`, `!=`, `<`, `<=`, `>`, `>=`)
 
 The comparisons are done using [Version Order](#version-ordering), including for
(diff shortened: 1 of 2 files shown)


doc/pages/Manual.md
```diff
@@ -571,8 +571,8 @@ The following are allowed in filters:
 - Idents
 - Parentheses
 - Logical operators (binary AND `&`, binary OR `|`, prefix, unary NOT `!`)
-- The unary operator `?` for detecting whether an expression contains undefined
-  variables
+- The unary operator `?` for detecting whether an expression evaluates to a
+  defined value
```
kit-ty-kate, line 575:
```suggestion
  defined value. If the expression `expr` is an undefined value then `?expr` will be evaluated to `false`. If it does not then it is evaluated to `true`.
```
We could also simplify and linking to the example 2 paragraphs below

emillon: putting it in full sounds good to me, but I think this should mention evaluation: `?expr` is valid for all types `expr` and it will actually evaluate `expr`. How about something like:

> - The unary operator `?` for detecting whether a value is defined. `?expr` evaluates to `false` if `expr` evaluates to an undefined value, and `true` otherwise.

It's in a weird order, but this avoids talking about "defined values"

> - The unary operator `?` for detecting whether a value is defined. `?expr` evaluates to `true` if `expr` evaluates to a defined value, and `false` if it evaluates to an undefined value.

kit-ty-kate: I like the second proposal ^_^

emillon: I updated to that

kit-ty-kate (approved):

kit-ty-kate:
Thanks!

Issue #5479, opened 2023-03-12
opam is utterly slow on basic errors 


dbuenzli:
A basic command line typo takes more than 3s to respond. Not very pleasant.

```
> opam --version 
2.1.4
> time opam sitch
opam: unknown command `sitch', did you mean `switch' ?
Usage: opam COMMAND ...
Try `opam --help' for more information.
opam sitch  3.66s user 0.82s system 93% cpu 4.810 total
```

avsm:
Looks like it's the 2.1.x depexts probing that's causing the slowdown.

```
% OPAMDEBUG=1 opam sitch
00:00.024  GSTATE                 LOAD-GLOBAL-STATE @ /Users/avsm/.opam
00:00.025  RSTATE                 LOAD-REPOSITORY-STATE @ /Users/avsm/.opam
00:00.069  CACHE(repository)      Loaded /Users/avsm/.opam/repo/state-323C573F.cache in 0.044s
00:00.069  RSTATE                 Cache found
00:00.069  STATE                  LOAD-SWITCH-STATE @ 5.0b2
00:00.070  CACHE(installed)       Loaded /Users/avsm/.opam/5.0b2/.opam-switch/packages/cache in 0.000s
00:00.282  STATE                  Switch state loaded in 0.212s
00:00.286  SYSTEM                 mkdir /var/folders/pv/2n59f7453qq5fn5c_vxkb0pm0000gn/T/opam-avsm-59418
00:00.298  GSTATE                 Failed to evaluate global variable sys-ocaml-version: "ocamlc": command not found.
00:00.357  XSYS                   Adding to env { LC_ALL=C }
00:01.061  STATE                  depexts loaded in 0.703s
```

and with depexts disabled

```
% OPAMDEBUG=1 OPAMNODEPEXTS=1 opam sitch
00:00.022  GSTATE                 LOAD-GLOBAL-STATE @ /Users/avsm/.opam
00:00.023  RSTATE                 LOAD-REPOSITORY-STATE @ /Users/avsm/.opam
00:00.069  CACHE(repository)      Loaded /Users/avsm/.opam/repo/state-323C573F.cache in 0.045s
00:00.069  RSTATE                 Cache found
00:00.069  STATE                  LOAD-SWITCH-STATE @ 5.0b2
00:00.069  CACHE(installed)       Loaded /Users/avsm/.opam/5.0b2/.opam-switch/packages/cache in 0.000s
[...]

PR #5518, opened 2023-04-21
Speedup OpamVersionCompare by 15%

kit-ty-kate:
~Doing so does, for example, speeds-up typical `opam show <pkg>` by 50%~

I also tried a completely new implementation of `OpamPackage.Version.compare` that would have sped-up the comparison function by 30% instead but doing so made `OpamPackage.Version.of_string` slower, as i was actually parsing the string first and pre-processed it so the comparison function is faster, but it wasn't worth it in the end.

~F.i.x.e.s. https://github.com/ocaml/opam/issues/5479 as it speeds-up switch loading time from 0.22s to 0.044s on my M1Pro.~
Related to https://github.com/ocaml/opam/issues/4245 (cc @emillon)

Requires #6078

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -122,6 +122,7 @@ users)
   * Stop using polymorphic comparison when comparing `OpamTypes.switch_selections` [#6102 @kit-ty-kate]
   * Remove the meta opam packages opam and opam-admin [#6115 @kit-ty-kate]
   * Reduce allocations in OpamVersionCompare [#6144 @talex5]
+  * Speedup OpamVersionCompare by 25% by removing the unused handling of epoch [#5518 @kit-ty-kate]
 
 ## Internal: Windows
 
--- a/src/core/opamVersionCompare.ml
+++ b/src/core/opamVersionCompare.ml
@@ -15,7 +15,6 @@
 let is_digit = function
   | '0'..'9' -> true
   | _ -> false
-;;
 
 (* [skip_while_from i f w m] yields the index of the leftmost character
  * in the string [s], starting from [i], and ending at [m], that does
@@ -23,7 +22,6 @@ let is_digit = function
 let rec skip_while_from i f w m =
   if i = m then i
   else if f w.[i] then skip_while_from (i + 1) f w m else i
-;;
 
 (* splits a version into (epoch,rest), without the separating ':'. The
  * epoch is delimited by the leftmost occurrence of ':' in x, and is ""
@@ -36,7 +34,6 @@ let extract_epoch x =
     in (epoch,rest)
   with
     | Not_found -> ("",x)
-;;
 
 (* splits a version into (prefix,revision). The revision starts on the
  * right-most occurrence of '-', or is empty in case the version does
@@ -49,7 +46,6 @@ let extract_revision x =
     (before,after)
   with
     | Not_found -> (x,"")
-;;
 
 (* character comparison uses a modified character ordering: '~' first,
    then letters, then anything else *)
@@ -64,12 +60,11 @@ let compare_chars c1 c2 = match c1 with
   | _ -> (match c2 with
       | '~'|'a'..'z'|'A'..'Z' -> 1
       | _ -> Char.compare c1 c2)
-;;
 
 (* return the first index of x, starting from xi, of a nun-null
  * character in x.  or (length x) in case x contains only 0's starting
  * from xi on.  *)
-let skip_zeros x xi xl = skip_while_from xi (fun c -> c = '0') x xl;;
+let skip_zeros x xi xl = skip_while_from xi (fun c -> c = '0') x xl
 
 (* compare versions chunks, that is parts of version strings that are
  * epoch, upstream version, or revisision. Alternates string comparison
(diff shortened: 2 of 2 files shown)


src/core/opamVersionCompare.ml
```diff
@@ -15,43 +15,27 @@
...
-    if i = m then i
+    if Int.equal i m then i
     else if f w.[i] then loop (i + 1) else i
   in loop i
-;;
-
-(* splits a version into (epoch,rest), without the separating ':'. The
- * epoch is delimited by the leftmost occurrence of ':' in x, and is ""
- * in case there is no ':' in x.  *)
-let extract_epoch x =
```
kit-ty-kate, line 33: Note for reviewers: This was removed because `OpamPackage.Version.of_string` disallows `:`

kit-ty-kate:
Crap, my testing method for `opam show` and the misspelled command wasn't good.
What I was actually testing was the performance difference between OCaml 4.14.1 and 5.0.0 (how is that giving us 50% performance for free!?!)
Thanks to #5525 we're actually able to detect performance changes reliably so at least that's the good news.

The speedup for OpamVersionCompare is still present but it's not as significant for real-life commands sadly :(

kit-ty-kate:
TODO: Use `In_channel.input_all` when available

kit-ty-kate:
The latest benchmark shows a 18% speedup, in line with the local tests. I believe this is ready to go.

kit-ty-kate:
After https://github.com/ocaml/opam/pull/6144 the PR now improves performance by 28%

src/core/opamVersionCompare.ml
talex5: That's good to know - getting rid of the remaining allocs should be a lot easier if we don't need to support epochs!

master_changes.md
```diff
@@ -122,6 +122,7 @@ users)
   * Stop using polymorphic comparison when comparing `OpamTypes.switch_selections` [#6102 @kit-ty-kate]
   * Remove the meta opam packages opam and opam-admin [#6115 @kit-ty-kate]
   * Reduce allocations in OpamVersionCompare [#6144 @talex5]
+  * Speedup OpamVersionCompare by 15% [#5518 @kit-ty-kate]
```
rjbou, line 125:
```suggestion
  * Speedup OpamVersionCompare by 15%: simplify comparison code [#5518 @kit-ty-kate]
```

src/core/opamVersionCompare.ml
```diff
@@ -15,28 +15,13 @@
...
-;;
 
 (* [skip_while_from i f w m] yields the index of the leftmost character
  * in the string [s], starting from [i], and ending at [m], that does
  * not satisfy the predicate [f], or [length w] if no such index exists.  *)
 let rec skip_while_from i f w m =
-  if i = m then i
+  if Int.equal i m then i
   else if f w.[i] then skip_while_from (i + 1) f w m else i
-;;
```
rjbou, line 26: The cosmetic change should be in a separate commit

src/core/opamVersionCompare.ml
```diff
@@ -15,28 +15,13 @@
...
   | '0'..'9' -> true
   | _ -> false
-;;
 
 (* [skip_while_from i f w m] yields the index of the leftmost character
  * in the string [s], starting from [i], and ending at [m], that does
  * not satisfy the predicate [f], or [length w] if no such index exists.  *)
 let rec skip_while_from i f w m =
-  if i = m then i
+  if Int.equal i m then i
```
rjbou, line 23: Does the removal of polymorphic compare speeds up the comparison ?

src/core/opamVersionCompare.ml
```diff
@@ -80,8 +63,8 @@ let compare_chunks x y =
   and yl = String.length y
   in
   let rec loop_lexical xi yi =
-    assert (xi <= xl && yi <= yl);
-    match (xi=xl,yi=yl) with (* which of x and y is exhausted? *)
+    assert (Int.compare xi xl <= 0 && Int.compare yi yl <= 0);
```
rjbou, line 66: For readability, i'm wondering if it is better to have the annotation form:
 `(xi <= (xl:int) && yi <= (yl:int))` vs `(Int.compare xi xl <= 0 && Int.compare yi yl <= 0)`?

rjbou (commented):
On the idea and the code itself, lgtm!
I'm wondering if we should update the header as now we are changing that piece of code. Previous version is still the same code that the one in [dose](https://gitlab.com/irill/dose3/-/blob/master/src/versioning/debian.ml?ref_type=heads) (modulo formatting). 
Also, can you update the main comment to highlight the change you made, the origin of the speedup.

master_changes.md
kit-ty-kate: done

src/core/opamVersionCompare.ml
kit-ty-kate: it does a little bit but i've done it mostly to clean things up

src/core/opamVersionCompare.ml
kit-ty-kate: done

src/core/opamVersionCompare.ml
kit-ty-kate: done

src/core/opamVersionCompare.ml
rjbou: In that case, it should be in its proper commit, if not part of the speedup

kit-ty-kate: it's already the case in this PR

rjbou: As you already did!

rjbou (approved):
Thanks!!

PR #5522, opened 2023-04-24
configure: Ensure a complementary (32bit on 64bit platforms and 64bit on 32bit platforms) C compiler is installed on Windows

kit-ty-kate:
~`src/stubs/win32/cc64` was never written before because the C compiler was always the wrong one (x86_32 when only the x86_64 is available and x86_64 when only the x86_32 one is available)~

~I'm not sure how it ever worked before. Maybe the typical setup always has both x86_64 and x86_32 so it somehow worked out.~

A complementary C compiler (32bit on 64bit platforms and 64bit on 32bit platforms) is required to build opam-putenv on Windows, but the check for those compilers only output a warning instead of an error. Said warning made it seem like it way an optional binary, however this is not the case and is always required on Windows AFAIK.

cc @MisterDA

### DIFF
--- a/configure
+++ b/configure
@@ -5315,8 +5315,7 @@ fi
   if  test "x${CC64}" = "xno"
 then :
 
-    { printf "%s\n" "$as_me:${as_lineno-$LINENO}: WARNING: Complementary C compiler not found - opam-putenv will not be built" >&5
-printf "%s\n" "$as_me: WARNING: Complementary C compiler not found - opam-putenv will not be built" >&2;}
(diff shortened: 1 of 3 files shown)


kit-ty-kate:
Reading https://github.com/ocaml/opam/commit/316f7ac0a43fde4e0546236009e712b98cdf0b61 (the commit that introduced this line and the cc64 concept), I'm even more confused now.

rjbou:
Having a warning makes more difficult to debug, see #5601

rjbou (approved):

Issue #5529, opened 2023-05-01
`opam upgrade` shows an empty list of packages no longer available


jchavarri:
Sometimes I ran into this error when running `opam upgrade`:

```
[WARNING] Upgrade is not possible because of conflicts or packages that are no longer available:

You may run "opam upgrade --fixup" to let opam fix the current state.
```

Running the command with `fixup` flag fixes it, so that part is fine. But as there is not much one can do as a user because the list of conflicts is empty, maybe opam could internally rerun the command using `fixup` automatically, when detecting such situations?

kit-ty-kate:
Does https://github.com/ocaml/opam/pull/5530 work for you?

How to test the branch:
```
make cold
sudo install ./opam <path to replace your local opam if you want>
```

jchavarri:
Thanks for the fix! I don't have a reproducible case, but as soon as I run into the problem again, I will try the suggested PR.

PR #5530, opened 2023-05-01
Do not show a message on upgrade if everything is up-to-date

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/5529

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -171,7 +171,7 @@ let upgrade_t
       OpamConsole.errmsg "%s\n%s\n"
         (OpamStd.Format.itemize (fun x -> x) cycles)
         "You may try upgrading packages individually to work around this."
-    end else begin
+    end else if reasons <> [] then begin
       OpamConsole.warning
         "Upgrade is not possible because of conflicts or packages that \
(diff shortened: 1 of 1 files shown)


src/client/opamClient.ml
```diff
@@ -171,7 +171,7 @@ let upgrade_t
       OpamConsole.errmsg "%s\n%s\n"
         (OpamStd.Format.itemize (fun x -> x) cycles)
         "You may try upgrading packages individually to work around this."
-    end else begin
+    end else if reasons <> [] then begin
```
jchavarri, line 174: I am not sure I understand this fix. If the code reaches this point, it means there are conflicts, but for some reason, `reasons` and `cycles` are both empty 🤔 

But if the error messages below are not shown, won't it be worse than the current situation? My concern is that opam will finish without upgrading, but the user has no indication of what has happened.

kit-ty-kate: ah you're right I did it too fast without really looking. Mmh, maybe that was fixed in master with https://github.com/ocaml/opam/pull/5263. Could you try simply master instead (whenever you have a reproducible case)?

PR #5535, opened 2023-05-03
The conflicts field's filter does not support package variables: update lint

kit-ty-kate:
See https://github.com/ocaml/opam-repository/pull/23736


rjbou (approved):
Thanks!

PR #5537, opened 2023-05-05
opam pin add: Allow specifying multiple packages

reynir:
When pinning packages you often want to pin multiple packages from one source but not all. This PR allows `opam pin add` specifying multiple packages instead of at most one package name.


kit-ty-kate (approved):

kit-ty-kate:
I just had the need for this today, however i needed to use `--current` which is not supported by this PR i believe. I think it would be nice to also include `opam pin add --current <pkg1> <pkg2>...` to this PR

PR #5444, opened 2023-02-14
2.1.5 backport commits

kit-ty-kate:
In this PR
*  #5352
*  #5350
*  #5315 
*  #5342
*  #5538

See https://github.com/ocaml/opam/issues/5409

PR #5538, opened 2023-05-08
Fix opam installing packages without checking their checksum when the local cache is corrupted in some cases

kit-ty-kate:
Reported by @roburio
Pre-checked by @rjbou 
Ready for integration in https://github.com/ocaml/opam/pull/5444


rjbou (approved):
Thanks!

PR #5543, opened 2023-05-10
Better cygwin support in core 

rjbou:
* At `opam_init` and global state loading, update internal configuration to retrieve cygwin binary path and store it in configuration (`OpamCoreConfig`)
* When defined, add cygwin binary path to build environment

### TODO
* [x] queued on #5542

PR #5544, opened 2023-05-10
init: detect local cygwin installation

rjbou:
At init && reinit, add detection and setup of a pre-existent cygwin installation.

It contains commit of #5543 (see https://github.com/ocaml/opam/pull/5543#issuecomment-1589723915)

#### TODO
* [x] queued on #5543
* [x] queued on #5542

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -635,6 +635,181 @@ let init_checks ?(hard_fail_exn=true) init_config =
   if hard_fail && hard_fail_exn then OpamStd.Sys.exit_because `Configuration_error
   else not (soft_fail || hard_fail)
 
+let windows_checks config =
+  let vars = OpamFile.Config.global_variables config in
+  let env =
+    List.map (fun (v, c, s) -> v, (lazy (Some c), s)) vars
+    |> OpamVariable.Map.of_list
+  in
+  let success cygcheck =
+    let config =
+      let os_distribution = OpamVariable.of_string "os-distribution" in
+      let update vars =
+        OpamFile.Config.with_global_variables
+          ((os_distribution, S "cygwin", "Set by opam init")::vars)
+          config
+      in
+      match OpamStd.List.pick (fun (v,_,_) ->
+          OpamVariable.equal v os_distribution)
+          vars with
+      | Some (_, S "cygwin", _), _ -> config
+      | None, vars ->  update vars
+      | Some (_, vc, _), vars ->
+        OpamConsole.warning
+          "'os-distribution' already set to another value %s"
+          (OpamVariable.string_of_variable_contents vc);
+        if OpamConsole.confirm ~default:false "Override?" then
+          (OpamConsole.msg
+             "You can revert this setting using \
+              'opam var --global os-distribution=%s'"
+             (OpamVariable.string_of_variable_contents vc);
+           update vars)
+        else
+          OpamStd.Sys.exit_because `Aborted
+    in
+    OpamFile.Config.with_sys_pkg_manager_cmd
+      (OpamStd.String.Map.add "cygwin" cygcheck
+         (OpamFile.Config.sys_pkg_manager_cmd config))
+      config
+  in
+  let get_cygwin = function
+    | Some cygcheck
+      when OpamFilename.exists cygcheck
+        && OpamStd.Sys.is_cygwin_cygcheck
+             ~cygbin:(Some OpamFilename.(Dir.to_string (dirname cygcheck))) ->
+      (* Should display the name using the converted path of /, not the bin dir *)
+      OpamConsole.note "Using Cygwin installation at %s for depexts"
+        (OpamFilename.(Dir.to_string (dirname cygcheck)));
+      success cygcheck
+    | Some _ | None ->
+      let rec menu () =
+        let enter_paths () =
+          let prompt_setup () =
+            let options = [
+              `download, "Let opam downloads it";
+              `manual, "Manually enter its location on disk";
+              `abort, "Abort initialisation";
+            ]
+            in
+            OpamConsole.menu
+              "Opam needs Cygwin setup executable 'setup-x86_64.exe'"
+              ~default:`download ~no:`download ~options
+          in
+          let rec enter_setup () =
+            match prompt_setup () with
+            | `abort -> OpamStd.Sys.exit_because `Aborted
+            | `download -> None
+            | `manual ->
+              match OpamConsole.read "Enter path of Cygwin setup executable:" with
+              | None -> None
+              | Some setup ->
+                let setup = OpamFilename.of_string setup in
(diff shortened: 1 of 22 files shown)


src/client/opamClient.ml
```diff
@@ -635,6 +635,150 @@ let init_checks ?(hard_fail_exn=true) init_config =
...
+      (* Should display the name using the converted path of /, not the bin dir *)
+      OpamConsole.note "Using Cygwin installation at %s for depexts"
+        (OpamFilename.(Dir.to_string (dirname cygcheck)));
+      success cygcheck
+    | Some _ | None ->
+      let rec menu () =
+        let enter_paths () =
+          let prompt_setup () =
+            let options = [
+              `manual, "Specify Cygwin setup path";
```
AltGr, line 689: Enter the path were the
```suggestion
              `manual, "Manually enter a path were `cygcheck.exe` can be found";
```

I am not sure it's cygcheck.exe we are after here but we should point to something explicit the user can find, even if they don't know much about Cygwin. Or maybe "Manually enter the prefix of an existing Cygwin installation (e.g. C:\\cygwin)" with the most usual default path cygwin installs at ?

src/client/opamClient.ml
```diff
@@ -635,6 +635,150 @@ let init_checks ?(hard_fail_exn=true) init_config =
...
+              `download,
+              "Let opam downloads Cygwin setup and stores it in \
+               Cygwin installation path";
```
AltGr, line 692:
```suggestion
              "Automatically download Cygwin and set it up at ???";
```
I am not sure what `Cygwin installation path` means in this context.

src/client/opamClient.ml
```diff
@@ -635,6 +635,150 @@ let init_checks ?(hard_fail_exn=true) init_config =
...
+            ]
+            in
+            OpamConsole.menu "Opam needs Cygwin setup executable" ~no:`download ~options
+          in
+          let rec enter_path () =
+            match OpamConsole.read "Enter the path to the Cygwin installation:" with
+            | None -> None
+            | Some "" ->
+              OpamConsole.msg "Empty path\n";
+              enter_path ()
```
AltGr, line 703:
```suggestion
```
`OpamConsole.read` will return `None` and never `Some ""` on empty input, so this case seems superfluous. But since it's different from the previous one I don't know what to make of this :grimacing: 

Also be careful on looping on `None` since that's what non-interactive mode returns (but here it's guarded by the menu so should be OK)

src/client/opamClient.ml
```diff
@@ -635,6 +635,150 @@ let init_checks ?(hard_fail_exn=true) init_config =
...
+            | Some entry -> enter_setup entry
+          and enter_setup entry =
+            match prompt_setup () with
+            | `abort -> OpamStd.Sys.exit_because `Aborted
+            | `download -> Some (entry, None)
+            | `manual ->
+              match OpamConsole.read "Enter path of Cygwin setup:" with
```
AltGr, line 710:
```suggestion
              match OpamConsole.read "Enter the path to the Cygwin setup executable (e.g. `setup-x86_64.exe`)" with
```

src/client/opamClient.ml
```diff
@@ -635,6 +635,150 @@ let init_checks ?(hard_fail_exn=true) init_config =
...
+        match enter_paths () with
+        | Some config -> config
+        | None -> menu ()
+      in
+      OpamConsole.header_msg "Unix support infrastructure";
+      OpamConsole.msg
+        "\n\
+         opam and the OCaml ecosystem in general require various Unix tools \
+         in order to operate correctly. At present, this requires \
+         an already present Cygwin installation to provide these tools.\n\n";
```
AltGr, line 739:
```suggestion
         opam and the OCaml ecosystem in general depend on various Unix tools \
         in order to operate correctly. At present, this requires \
         a pre-existing Cygwin installation.\n\n";
```

src/state/opamSysInteract.ml
```diff
@@ -237,6 +239,65 @@ module Cygwin = struct
   let cygroot config = get_opt (cygroot_opt config)
   let cygsetup config = get_opt (cygsetup_opt config)
 
+  let download_setupexe dst =
+    let open OpamProcess.Job.Op in
+    OpamFilename.with_tmp_dir_job @@ fun dir ->
+    OpamDownload.download ~overwrite:true url_setupexe dir @@| fun filename ->
+    (* XXX check signature *)
```
AltGr, line 246: In the meantime, as a cheap workaround, we could just display the sha512 and ask for confirmation.
Apparently cygwin postes those at https://cygwin.com/sha512.sum

AltGr (approved):
The interaction and detection mechanisms look good, but I think we need to be extra clear with what exactly we are asking the user (in particular paths, etc.) and in what context.

It may (in time ;) ) be the first interaction with OCaml of ppl not very familiar with the command-line so we need to be extra careful.

src/client/opamClient.ml
rjbou: *Manually enter the prefix of an existing Cygwin installation (e.g. C:\cygwin)* is good indeed. It needs cygwin prefix, or `/bin` location of cygwin, so it's good with that proposal.

src/client/opamClient.ml
rjbou: It's Cygwin prefix, e.g `C:\cygwin64`. Opam look up to setup.exe in cygwin prefix. So we ask for its location to copy it, or propose to download it to store it in the cygwin prefix.

src/state/opamSysInteract.ml
```diff
@@ -237,6 +239,65 @@ module Cygwin = struct
...
+    OpamFilename.(mkdir (dirname dst));
```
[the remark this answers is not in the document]
rjbou: true

src/state/opamSysInteract.ml
rjbou:
```
 let download_setupexe dst =
    let open OpamProcess.Job.Op in
    OpamFilename.with_tmp_dir_job @@ fun dir ->
    OpamDownload.download ~overwrite:true url_setupexe_sha512 dir @@| fun file ->
    let checksum =
      let content = OpamFilename.read file in
      let re =
        Re.(compile @@ seq [
            group @@ repn
              (alt [ digit ; rg 'A' 'F'; rg 'a' 'f' ]) 128 (Some 128);
            rep space;
            str "setup-x86_64.exe"
          ])
      in
      try Some (OpamHash.sha512 Re.(Group.get (exec re content) 1))
      with Not_found -> None
    in
    OpamDownload.download_as ~overwrite:true ?checksum url_setupexe dst

rjbou:
Some examples

Menu with a pre-existent cygwin and defaults

default setup

Menu with specified cygwin & setup location

full specified

Menu with no pre-existent cygwin installation

no cygwin at default

PR #5545, opened 2023-05-10
init: install cygwin internally

rjbou:
At init, add an item in the cygwin menu to propose to have a cygwin install internal to opam. It is handled by opam itself (prefix stored in `<opamroot>/.cygwin/root`), when depext are requested, there is no confirmation requested.
It also adds some flags to `init`, to have a cywgin configuration non interactively: `--no-cygwin-setup`|`--cygwin-internal-install`|`--cygwin-local-install` and `--cygwin-location <path>`.
These are experimental as they can change name from the finale release.

#### TODO
* [x] queued on #5544
* [x] queued on #5543
* [x] queued on #5542

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -717,13 +717,13 @@ let windows_checks config =
                 let options = [
                   `manual,
                   "Manually enter prefix of an existing Cygwin installation \
-                   (e.g. C:\\cygwin64)";
+                   (e.g. D:\\cygwin64)";
                   `default,
                   (Printf.sprintf "Use default Cygwin installation at %s"
                      OpamSysInteract.Cygwin.default_cygroot);
                   `abort, "Abort initialisation";
                 ] in
-                OpamConsole.menu "Opam needs pre-existent Cygwin installation"
+                OpamConsole.menu "Cygwin location"
                   ~default:`default ~no:`default ~options
               in
               (match prompt_cygroot () with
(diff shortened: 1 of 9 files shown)


src/client/opamClient.ml
```diff
@@ -635,6 +635,168 @@ let init_checks ?(hard_fail_exn=true) init_config =
...
+            in
+            match cygcheck with
+            | Ok cygcheck -> Some (success cygcheck)
+            | Error msg -> OpamConsole.error "%s" msg; None
+        in
+        let prompt () =
+          let options = [
+            `Internal, "Install and maintain an internal Cygwin installation";
+            `Specify, "Enter the location of a Cygwin installation and setup path";
+            `Abort, "Abort initialisation";
```
AltGr, line 734:
```suggestion
            `Internal, "Automatically create an internal Cygwin installation that will be managed by opam";
            `Specify, "Enter the location of an existing Cygwin installation";
            `Abort, "Abort initialisation";
```

src/client/opamClient.ml
```diff
@@ -635,6 +635,168 @@ let init_checks ?(hard_fail_exn=true) init_config =
...
+          ] in
+          OpamConsole.menu "How should opam acquire Cygwin?" ~no:`Internal ~options
```
AltGr, line 736:
```suggestion
          OpamConsole.menu "How should opam handle Cygwin?" ~no:`Internal ~options
```

src/client/opamInitDefaults.ml
```diff
@@ -144,6 +144,16 @@ let required_tools ~sandboxing () =
     ["sandbox-exec"], None, Some macos_filter;
   ] else []
 
+let required_packages_for_cygwin =
+  [
+    "diffutils";
+    "git"; (* XXX hg & mercurial ? *)
```
AltGr, line 150: Those are so little used nowadays that we shouldn't bother with them by default

src/client/opamCommands.ml
```diff
@@ -303,11 +303,27 @@ let init cli =
...
+      "Use preexistent Cygwin install";
+      cli_from cli2_2, `no, ["no-cygwin-setup"],
+      "Don't setup Cygwin";
+    ]
+  in
+  let cygwin_location =
+    mk_opt ~cli (cli_from cli2_2) ["cygwin-location"] "DIR"
+      "Specify Cygwin root location"
+      Arg.(some dirname) None
+  in
```
rjbou, line 320: Other proposal: `--cygwin [no|internal|default|<location>]`

AltGr: Hm I like the proposal (but it's ok as is)

src/client/opamClient.ml
```diff
@@ -781,8 +800,38 @@ let windows_checks config =
...
+                with
+                | Ok cygcheck -> cygcheck
+                | Error msg ->
+                  OpamConsole.error_and_exit `Not_found
+                    "Error while checking default Cygwin install (%s): %s"
+                    (OpamSysInteract.Cygwin.default_cygroot) msg)
+             | `location dir ->
+               (match
+                  OpamSysInteract.Cygwin.check_install
+                  (OpamFilename.Dir.to_string dir)
```
kit-ty-kate, line 824:
```suggestion
                    (OpamFilename.Dir.to_string dir)
```

kit-ty-kate (approved):
aside from the minor indentation change, lgtm (on a code standpoint only, i haven't tested on an actual windows machine)

src/client/opamCommands.ml
rjbou: now, we do

PR #5546, opened 2023-05-10
depexts & cygwin: internal mechanism for available packages

rjbou:
*to be updated*

#### TODO
* [ ] queued on #5545
* [ ] queued on #5544
* [ ] queued on #5543
* [ ] queued on #5542


rjbou:
Withdrawed PR. The code can be relevant if one day we want to have.

PR #5549, opened 2023-05-12
Allow passing in a custom runner to `OpamStd.Sys`

Leonidas-from-XIV:
This adds a way to create a custom `OpamStd.Sys` using a different runner (in my case Dune's `Process` which uses Dune's `Fiber.t`), which allows us to better control the processes OPAM launches.

### DIFF
--- a/src/core/opamStd.ml
+++ b/src/core/opamStd.ml
@@ -917,8 +917,7 @@ module UnitRunner : Runner = struct
     let stdio = Unix.open_process_full cmd argv in
     let (stdout, _, _) = stdio in
     Fun.protect
-      (fun () ->
-        return (input_line stdout))
+      (fun () -> return (input_line stdout))
       ~finally:(fun () ->
         let _ : Unix.process_status = Unix.close_process_full stdio in
         ())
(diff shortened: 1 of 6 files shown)


src/core/opamStd.ml
```diff
@@ -879,21 +878,67 @@ module OpamSys = struct
...
+
+  let return v = (fun () -> v)
+
+  let map v f = return (f (v ()))
+
+  let with_process_in cmd args =
     if Sys.win32 then
       assert false;
-    let path = split_path_variable (Env.get "PATH") in
+    let path = Path.split_variable (Env.get "PATH") in
```
rgrinberg, line 912: It would be great if callers could do the `which` lookups itself and patch absolute paths. Dune has its own mechanism for looking up binaries in PATH which respects the watch mode.

Leonidas-from-XIV: I've kept it as-is to not change the current OPAM behavior and API. Dune would use a different runner implementation where we can look up binaries the Dune way.

rgrinberg: Sounds good. Can the API be tweaked not to require the shell? I notice that currently:

1. The command and arguments are concatenated. Would it be possible to take a `prog:string -> argv:string list` instead?
2. Instead of relying on shell redirection, can we set stderr/stdout/stdin by providing fd's?

PR #5552, opened 2023-05-16
Return the retrieved checksums on failure

Leonidas-from-XIV:
Before this change it was not possible to get the actual checksum from the API, because OPAM would download the file, compare checksums, emit an error message and delete the file with the wrong checksum.

So to be able to get the actual checksum, one needed to disable the validation and reimplement the validation on top of the unvalidated API call. This PR returns the invalid checksums.


Leonidas-from-XIV:
@hannesm If can be useful if you know you're expecting the checksum to change. E.g. you have a checksum specified, then deliberately change the tarball, then whoever is retrieving it using this API can tell you "hey, the checksum changed, is this deliberate? If so, I can actually also update the expected checksum to match".

A bit like the promotion mechanism in Dune, you can of course edit the files manually, but if the change is expected, why not let the computer do it.

PR #5555, opened 2023-05-23
Apply `autoupdate` to silence `autogen` warnings

MisterDA:


rjbou (commented):
I've added an autopudate check in hygiene job.

What is the origin/point of the second commit `OPAMCLIVersion.of_string` ?

MisterDA:
> I've added an autopudate check in hygiene job.

Is this why you reverted the first commit? To check that `autogen` would fail in the CI?

> What is the origin/point of the second commit `OPAMCLIVersion.of_string` ?

Well, I had to debug why opam was failing to parse/extract its version from the configure file. It was nice to print the actual string to see what happened, a bit like "No such file or directory" is enhanced if you actually know which file or directory has not been found.

rjbou:
>> I've added an autopudate check in hygiene job.

> Is this why you reverted the first commit? To check that `autogen` would fail in the CI?

Oups, it was to test locally. I'll run some tests on my repo to confirm that the error is triggered.

>> What is the origin/point of the second commit `OPAMCLIVersion.of_string` ?

> Well, I had to debug why opam was failing to parse/extract its version from the configure file. It was nice to print the actual string to see what happened, a bit like "No such file or directory" is enhanced if you actually know which file or directory has not been found.

Oh i see, for that, it's better to open another PR, to keep them single purpose.

MisterDA:
> Oh i see, for that, it's better to open another PR, to keep them single purpose.

Done!

MisterDA:
Rebased, I think this is good to go.

MisterDA:
Rebased the PR on the current master.

rjbou (approved):
Github action tested & validated, good to go. Thanks!

PR #5557, opened 2023-05-23
Fix openssl missing message

MisterDA:
With the backticks, the shell would try to execute openssl which is already missing.


rjbou (approved):
thanks!

PR #5561, opened 2023-05-26
lint: add errors about duplicated extra-files and checksums & when extra-file paths contains '..'

rjbou:
New lints:
* E70 *Field 'extra-files' contains duplicated files*
* E71 *Field 'url.checksum' contains duplicated checksums*
* E72 *Field 'extra-sources.checksum' contains duplicated checksums*
* E73 *Field 'extra-files' contains path with '..'* (it can be removed by #5562)

Update:
* W37 *Missing field 'dev-repo'*: update test
* 

More testcases:
* E59 *url doesn't contain a checksum*
* E65 *URLs must be absolute*

/cc @hannesm @reynir 

* [x] queued on #5560
* [x] add changelog

PR #5560, opened 2023-05-26
Reftests: add tests to check url handling behaviours

rjbou:
Three similar tests, one per each field: url, extra-sources & extra-files.
Each test check presence, checksum validation, and cache populating.
test introduced in #5538 is integrated in url test file.

/cc @hannesm @reynir

queued
* [x] #5639 
* [x] #5640

PR #5562, opened 2023-05-26
opamfile: parse error on escapable paths

rjbou:
Escapable path, containing `..` now raise a parse error.

Removes lint 71 introduced in #5561

/cc @hannesm @reynir 

* [x] queued on #5560
* [x] queued on #5561

### DIFF
--- a/src/core/opamFilename.ml
+++ b/src/core/opamFilename.ml
@@ -9,6 +9,16 @@
 (*                                                                        *)
 (**************************************************************************)
 
+let is_escapable =
+  let re =
+    Re.(compile @@ seq [
+        alt [ char '/'; char '\\'];
+        str "..";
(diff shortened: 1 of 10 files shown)


AltGr (commented):
A parse error on files is actually a pretty good idea, and a likely security improvement.

But enforcing the invariant at the `OpamFilename` level seems pretty strict to me, I am afraid that it could break a few things. Couldn't that affect, for example, commands like `opam pin ../foo/foo.opam` ? Or is the path always resolved and made absolute soon enough ?

hannesm:
I looked at the last commit, and am convinced that it avoids `..` being in a path. From a security point of view, I do appreciate this PR. But of course @AltGr question is valid - maybe the result is too strict?

PR #5563, opened 2023-05-26
source: add `--no-checksums` & `--require-checksums`  flags

rjbou:
/cc @hannesm @reynir 

* [ ] queued on #5560
* [ ] queued on #5561


src/client/opamCommands.ml
```diff
@@ -3616,8 +3616,27 @@ let source cli =
...
+  let no_checksums =
+    mk_flag ~cli (cli_from cli2_2) ["no-checksums"]
+      "Do not verify the checksum of downloaded source.\
+       This is equivalent to setting $(b,\\$OPAMNOCHECKSUMS) to \"true\"."
+  in
+  let req_checksums =
+    mk_flag ~cli (cli_from cli2_2) ["require-checksums"]
+      "Enforce checksum verification befor downloading sources.\
+       This is equivalent to setting $(b,\\$OPAMREQUIRECHECKSUMS) to \"true\"."
+  in
```
AltGr, line 3628: might be worth sharing the flags definitions with the ones used in `OpamArg.build_options` ? (maybe that was too tedious to be worth it, in that case it's alright)

AltGr (approved):

hannesm:
Appreciated, and this looks great to me. As @AltGr mentioned, it may be useful to share some code. Thanks a lot.

src/client/opamCommands.ml
rjbou: It is possible to share it easily with `build_options`, (cf. d69eb84ee125643fce1a1dd0f344c2c5d539f5ad).
I've tried to move them from build options to global options, but we loose the ability to notify that some new commands now permit and use `--*-checksum` options (no possibility to describe cli validity fine grained). Cf ef3b1e2a149404163eec26d28e7d7b9247fc9cb0

kit-ty-kate (commented):
I don't understand this feature request. What does this bring in terms of change for users?

I understand `--require-checksums`, although i feel like this would have more usability as a global `opam option`.
But `--no-checksums` i really don't understand. If the goal is to allow packages without a checksum, then why not simply add a `--allow-no-checksums` argument instead of ignoring the checksum altogether?

I'm also wondering what how `opam source` is special compared to the other commands that might fetch packages.

Also, looking at the code, the two new arguments do not do anything aside from setting their own internal config value.

rjbou:
As discussed in dev meeting, the idea is to harmonise option for `opam source`, make `--no-checksum` and `--require-checksum` option available.
related #6311 & #6312

src/client/opamCommands.ml
```diff
@@ -3710,8 +3710,19 @@ let source cli =
       "Choose package without consideration for \
        the current (or any other) switch (installed or pinned packages, etc.)"
   in
-  let source global_options atom dev_repo pin no_switch dir () =
+  let no_checksums = no_checksums cli (cli_from cli2_2) in
+  let req_checksums = require_checksums cli (cli_from cli2_2) in
```
kit-ty-kate, line 3714:
```suggestion
  let no_checksums = no_checksums cli (cli_from cli2_4) in
  let req_checksums = require_checksums cli (cli_from cli2_4) in
```

kit-ty-kate (approved):

PR #5566, opened 2023-06-01
Print version when failing to parse it

MisterDA:
Extracted from #5555: 

>>> What is the origin/point of the second commit `OpamCLIVersion.of_string`?
>>
>> Well, I had to debug why opam was failing to parse/extract its version from the configure file. It was nice to print the actual string to see what happened, a bit like "No such file or directory" is enhanced if you actually know which file or directory has not been found.
>
> Oh i see, for that, it's better to open another PR, to keep them single purpose.

https://github.com/ocaml/opam/pull/5555#issuecomment-1568758308


rjbou:
thanks

PR #5577, opened 2023-06-15
bench: set debian version to bullseye

rjbou:


punchagan:
We (the `current-bench` maintainers) were going to send this in a while, but thanks for doing this!  We did a temporary fix on our end to deal with builds failing on `debian:bookworm` by forcing use of `debian:bullseye`, while we looking into the reason for this failure and fix it.

PR #5587, opened 2023-06-27
Extract arch and os from machine triplet when available, add machine triplet variable.

toots:
This PR is the `opam` counter part of the problems discussed in https://github.com/ocaml/ocaml/pull/12326.

> In some situations, the runtime environment used to build the compiler is not the same as the kernel architecture. I have seen two cases of this lately:
> * Raspberry pi version 4 ships by default with a 64 bit kernel and 32 bit runtime. Please don't ask me why.
> * Docker can run images from an architecture that is not the kernel's architecture, for instance 32 bit arm runtime with a 64 bit kernel. This is a much more likely use-case.
> * 
> In situations like these, the configure script picks up the architecture from the kernel. 

The `opam` binary is also impacted by this issue as it queries the architecture using `uname` which returns the host processor architecture, not the runtime architecture. Here's an example from inside a docker debian/armhf image:
```shell
$ opam var

arch              arm64            # Inferred from system
exe                                # Suffix needed for executable filenames (Windows)
jobs              3                # The number of parallel jobs set up in opam configuration
make              make             # The 'make' command to use
opam-version      2.1.5            # The currently running opam version
os                linux            # Inferred from system
os-distribution   debian           # Inferred from system
os-family         debian           # Inferred from system
os-version        12               # Inferred from system
root              /home/opam/.opam # The current opam root directory
switch            4.14.1           # The identifier of the current switch
sys-ocaml-arch    arm              # Target architecture of the OCaml compiler present on your system
sys-ocaml-cc      cc               # Host C Compiler type of the OCaml compiler present on your system
sys-ocaml-libc    libc             # Host C Runtime Library type of the OCaml compiler present on your system
sys-ocaml-version 4.13.1           # OCaml version present on your system independently of opam, if any
```

With these changes, the returned architecture is as we expect:
```shell
$ dune exec ./src/client/opamMain.exe var

arch              arm                 # Inferred from system
exe                                   # Suffix needed for executable filenames (Windows)
jobs              3                   # The number of parallel jobs set up in opam configuration
machine-triplet   arm-linux-gnueabihf # Inferred from system
make              make                # The 'make' command to use
opam-version      2.2.0~alpha~dev     # The currently running opam version
os                linux               # Inferred from system
os-distribution   debian              # Inferred from system
os-family         debian              # Inferred from system
os-version        12                  # Inferred from system
root              /home/opam/.opam    # The current opam root directory
switch            4.14.1              # The identifier of the current switch
sys-ocaml-arch    arm                 # Target architecture of the OCaml compiler present on your system
sys-ocaml-cc      cc                  # Host C Compiler type of the OCaml compiler present on your system
sys-ocaml-libc    libc                # Host C Runtime Library type of the OCaml compiler present on your system
sys-ocaml-version 4.13.1              # OCaml version present on your system independently of opam, if any
```

The `opam` binary, however, it making all sort of assumptions about the system it runs on, `os`, `distribution`, package manager etc. It, thus make sense that it should also be more accurate with the runtime's architecture.

With these changes, it should be possible to have `opam` invoke the OCaml's `configure` script with the `--build` parameter that matches the machine triplet.

This solution is more desirable than the proposed changes in the OCaml compiler `configure` script in that the compiler has less assumptions to make about the system and leaves the user in charge of dealing with corner-cases.


src/state/opamSysPoll.ml
```diff
@@ -171,6 +175,7 @@ let variables =
       "os-distribution", os_distribution;
       "os-version", os_version;
       "os-family", os_family;
+      "machine-triplet", lazy (OpamStd.Sys.machine_triplet ());
```
gasche, line 178: The house style seems to suggest separate auxiliary definitions

```ocaml
let poll_machine_triplet () = OpamStd.Sys.machine_triplet ()
let machine_triplet = Lazy.from_fun poll_machine_triplet
```

Issue #5597, opened 2023-07-08
opam init using the wrong wget


runeimp:
>   If your issue concerns a package not building, please report to
>   https://github.com/ocaml/opam-repository/issues or to the package maintainer
>   unless you are confident it is an issue in the opam tool itself.

```
$ opam config report
# opam config report
# opam-version         2.1.5 
# self-upgrade         no
# system               arch=x86_64 os=macos os-distribution=homebrew os-version=13.3
[ERROR] No switch is currently set. Please use 'opam switch' to set or install a switch
# read-state           OpamStd.OpamSys.Exit(50)
```

When I tried installing `opam` via Homebrew the initial installation seemed to go fine but when I tried to init...

```sh
$ opam init
No configuration file found, using built-in defaults.
Checking for available remotes: rsync and local, git, mercurial.
  - you won't be able to use darcs repositories unless you install the darcs command on your system.


<><> Fetching repository information ><><><><><><><><><><><><><><><><><><><>  🐫 
Fatal error:
/usr/local/bin/opam: "create_process" failed on /usr/local/opt/wget: Permission denied
```

I have both wget and curl installed. Why is opam trying to use `/usr/local/opt/wget`, which is just a directory, instead of `/usr/local/bin/wget` which is in the PATH? Also if it is actually doing a check why doesn't it check if the path is an executable binary instead of just for existence?

I renamed the `.opam` directory for a "clean start" and tried `OPAMFETCH=curl opam init` on the command line. I also tried setting environment variables in my Bash config, restarting my terminal (KiTTY), checking my environment variables before testing, and I still get the same error.

### 1st config

```
export OPAMFETCH=curl
```

### 2nd Config

```
export OPAMFETCH=/usr/local/bin/wget
```

### 3rd Config

```
export OPAMFETCH=curl
export OPAMCURL=/usr/bin/curl
```

Each time I get the same error trying to execute `/usr/local/opt/wget`.

rjbou:
Can you share the content of your path ? If opam looks at `/usr/local/opt`, it should be because it is in your path.
There is an issue that there is no check on directory vs file before doing permission check.

runeimp:
Yes, `/usr/local/opt` is in the PATH before `/usr/local/bin`. To be honest I should probably remove it from my PATH as it no longer seems to be a repository for binaries as it had been in the past for Homebrew. None-the-less, shouldn't the OPAM... environment variables take priority? What is the point of their existence if the system just checks the path and ignores them?

[...]

PR #5606, opened 2023-07-19
resolve command: don't select directories

### LINKED ISSUE #5585: Permission Denied error
I've ran into this issue on fresh installs in WSL, Ubuntu Hyper-V and now  Endeavour OS (arch). WSL and Hyper-V I never managed to get working. Things were working fine in Endeavour OS, but after rebooting I'm getting this error again. In Ubuntu the error was the same except that path was `/var/lib/git` Any help would be greatly appreciated. I followed the instructions here for all of the installs. https://ocaml.org/docs/up-and-running

report
```
# opam config report
# opam-version         2.1.4 
# self-upgrade         no
# system               arch=x86_64 os=linux os-distribution=endeavouros os-version=rolling
# jobs                 11
# repositories         1 (http) (default repo at 7b30d632)
# pinned               0
# current-switch       /home/daoko/projects/ocaml/dream
# read-state           /usr/bin/opam: "create_process" failed on /var/lib/pacman: Permission denied

```
```
opam install ocaml-lsp-server 
Fatal error:
/usr/bin/opam: "create_process" failed on /var/lib/pacman: Permission denied
```

rjbou:
This PR contains 2 different fixes, that can't be split. It fixes #5585, #5650 and partially #5597, as it fixes command resolve that is it happen to be a directory. I also add a test for that, and to have the test working, reftest engine needs a fix. It was keeping duplicated environment variables in environment given as argument to process creation, and it was not expanding `PATH` on variable export.
It also add syntax to declare env vars paths additions `PATH+=path/to` or `PATH=+path/to`, to make it portable between unix & windows (`:` versus `;`).

Todo:
* [x] update changes

### DIFF
--- a/src/core/opamSystem.ml
+++ b/src/core/opamSystem.ml
@@ -441,11 +441,11 @@ let t_resolve_command =
     else fun f ->
       try
         let open Unix in
-        let uid = geteuid () in
+        let {st_uid; st_gid; st_perm; st_kind; _} = stat f in
+        if st_kind <> Unix.S_REG then false else
         let groups = OpamStd.IntSet.of_list (getegid () :: Array.to_list (getgroups ())) in
-        let {st_uid; st_gid; st_perm; _} = stat f in
         let mask =
-          if uid = st_uid then
+          if geteuid () = st_uid then
             0o100
           else if OpamStd.IntSet.mem st_gid groups then
             0o010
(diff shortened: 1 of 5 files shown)


kit-ty-kate (approved):

src/core/opamSystem.ml
```diff
@@ -488,9 +488,12 @@ let t_resolve_command =
...
       else name
     in
-    let possibles = OpamStd.List.filter_map (fun path ->
-        let candidate = Filename.concat path name in
-        if Sys.file_exists candidate then Some candidate else None) path
+    let possibles =
+      OpamStd.List.filter_map (fun path ->
+          let candidate = Filename.concat path name in
+          if Sys.file_exists candidate && not (Sys.is_directory candidate) then
+            Some candidate else None)
```
AltGr, line 495: This is slightly redundant, as `Sys.is_directory` will fail if the file doesn't exist
```suggestion
          match Sys.is_directory candidate with
          | false -> Some candidate
          | true -> None
          | exception (Sys_error _) -> None)
```

AltGr (approved):
Good catch !

src/core/opamSystem.ml
kit-ty-kate: A more compressed version can be used instead:
```suggestion
          match Sys.is_directory candidate with
          | false -> Some candidate
          | true | exception (Sys_error _) -> None)
```

rjbou:
Removed new test on windows (unix only) until ocaml/dune#8651 is resolved.

PR #5607, opened 2023-07-19
Fix `OPAMFETCH`/`OPAMCURL` handling

rjbou:
`OPAMCURL` is ignored if `OPAMFETCH` is set. This is changed to handle `OPAMCURL` specification when `OPAMFETCH` is set to `curl` (simple command, or new arguments). If `OPAMFETCH` is set to another tool, OPAMCURL is ignored.

Todo:
* [x] update changes
* [x] queud on #5654

### DIFF
--- a/tests/reftests/run.ml
+++ b/tests/reftests/run.ml
@@ -468,8 +468,9 @@ module Parse = struct
               *)
               opt_quoted @@ [
                 alpha; char ':';
-                rep1 @@ seq [ char '\\'; rep1 @@ diff any (with_quote_set "\\") ];
-                char '\\';
+                rep1 @@ seq [ char '\\'; opt @@ char '\\';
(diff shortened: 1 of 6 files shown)


AltGr (approved):
Seems reasonable to me!

tests/reftests/run.ml
```diff
@@ -422,17 +422,29 @@ module Parse = struct
...
             let re =
-              seq [
-                Re.str "+ ";
-                rep any;
-                alt [ set "/\\\"" ];
-                Re.str cmd;
-                opt @@ Re.str ".exe";
-                opt @@ char '"';
-                Re.str " "
+              alt [
```
rjbou, line 425: #5657 has a better resolution of `sed-cmd` than this one.

kit-ty-kate:
Thanks!

PR #5612, opened 2023-07-24
Various cleanups related to OCaml 5.0

kit-ty-kate:
This PR brings a couple of cleanups noticed when using OCaml >= 5.0:
* In OCaml 5.0 the `bigarray` library was removed. Since opam now requires at least OCaml 4.08 and bigarray has been part of the standard library since 4.08 we can now remove the dependencies on it
* Since OCaml 5.0 `Printf.kprintf` now shows a deprecation warning when used. `Printf.ksprintf` should be used instead

### DIFF
--- a/Makefile.config.in
+++ b/Makefile.config.in
@@ -4,7 +4,7 @@ mandir = @mandir@
 version = @PACKAGE_VERSION@
 FETCH = @fetch@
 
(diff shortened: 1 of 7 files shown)


rjbou (approved):
Thanks!

rjbou:
Benchmark is green, but it is not updated on github. Merging! Thanks!

PR #5616, opened 2023-07-25
Update release scripts

rjbou:
* Update sign script to sign only binaries, and sign archive 
* Makefile: use vendoredps deps instead of make lib-ext
* Bump OpenBSD to 7.3
* Update release script to run it on mac or linux
  * I updated it according my setup
  * It is not yet tested with current script.

### DIFF
--- a/release/Makefile
+++ b/release/Makefile
@@ -108,8 +108,7 @@ host: $(OUTDIR)/opam-full-$(VERSION).tar.gz build/$(HOST).env
 	    MAKE=$(MAKE) \
 	    $(EXPORTS_$(HOST_OS)); \
 	  cd build/opam-full-$(VERSION) && \
-	  ./configure && \
-	  { $(MAKE) lib-ext || ./configure --with-vendored-deps; } && \
+	  ./configure --with-vendored-deps && \
 	  echo "$(call LINKING,$(HOST_OS))" >src/client/linking.sexp && \
 	  $(MAKE) opam; \
 	)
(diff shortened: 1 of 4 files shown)


release/Makefile
```diff
@@ -108,8 +108,7 @@ host: $(OUTDIR)/opam-full-$(VERSION).tar.gz build/$(HOST).env
 	    MAKE=$(MAKE) \
 	    $(EXPORTS_$(HOST_OS)); \
 	  cd build/opam-full-$(VERSION) && \
-	  ./configure && \
-	  { $(MAKE) lib-ext || ./configure --with-vendored-deps; } && \
+	  ./configure --with-vendored-deps && \
```
kit-ty-kate, line 111: This is still required to make releases of the 2.1 branch

kit-ty-kate (changes requested):
What is the purpose of running it on Linux if we need to run it on mac anyway to build the mac binaries?

release/Makefile
rjbou: There is also another thing that is 2.1 related, sha dependency. Should we have conditional commands according 2.1/2.2 ?

kit-ty-kate: What is sha dependency?

rjbou: This [one](https://github.com/ocaml/opam/blob/master/release/Makefile#L58)

rjbou:
> What is the purpose of running it on Linux if we need to run it on mac anyway to build the mac binaries?

I don't have access to a graphical mac session :) I build everything in my deb, and the mac binaries via ssh on a mac.

rjbou:
it's on he device requirement
> * Launch docker using the Docker GUI macOS app

kit-ty-kate:
I wrote it that way only because I'm doing that personally and I've never had to do in on a mac without GUI but it should be possible

Issue #5491, opened 2023-03-24
Invariant syntax in opam config report doesn't match the expected syntax by --invariant


kit-ty-kate:
`opam config report` since #5480 returns the invariant for the current switch:
```
# invariant            ocaml-base-compiler >= 5.0 & ocaml-base-compiler < 5.1
```
However this syntax isn't accepted by other parts of opam dealing with invariants (e.g. `--invariant`)
It should be something like:
```
# invariant            "ocaml-base-compiler" >= "5.0" & "ocaml-base-compiler" < "5.1"
```
instead

rjbou:
It is dsisplayed in a readable way, it is not purpose of `opam config report` to be given as input of other commands.

kit-ty-kate:
I disagree. I think having something consistent is more human readable than having 3 different incompatible syntaxes.

PR #5619, opened 2023-07-26
Make opam config report and opam switch set-invariant return the actual invariant syntax expected by --invariant

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/5491


dra27 (approved):
I like it, and I'll ask @rjbou to look away while I hit merge...

rjbou (approved):
thanks!

PR #5624, opened 2023-08-03
Ensure all archives are retrieved in `make cache-archives`

dra27:
Small fix for 2.1 only - the full tarball doesn't include the 0install solver. master/2.2 isn't affected by this.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -61,7 +61,7 @@ New option/command/subcommand are prefixed with ◈.
   *
 
 ## Infrastructure
-  *
+  * Ensure the full archive includes the 0install sources [#5624 @dra27]
 
 ## Admin
   *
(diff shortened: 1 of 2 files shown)


dra27:
Repository SHA for the various updated, so GHA at least passes. AppVeyor is failing from the good ol' timeouts from the 2.1 reftests. No idea what to do about the ocaml-benchmarks failure!

rjbou (approved):
thanks!

rjbou:
there is no benchmarks on 2.1 branch, and from what i know it is not possible to select on which jobs/branch launch the benchmarks.

kit-ty-kate (approved):
I've tested using the release script. The new archive contains as expected: ocaml, flexdll, mccs, 0install and opam-0install-cudf

PR #5629, opened 2023-08-09
`OpamFilter` file substitutions with source & destination

Leonidas-from-XIV:
Adds a new function `OpamFilter.expand_interpolations_in_file_full` that takes both input path and also output path instead of inferring it.

This feature is upstreamed from the dune fork of OPAM.

### DIFF
--- a/src/format/opamFilter.mli
+++ b/src/format/opamFilter.mli
@@ -129,6 +129,8 @@ val ident_string: ?default:string -> env -> fident -> string
 (** Like [ident_value], but casts the result to a bool *)
 val ident_bool: ?default:bool -> env -> fident -> bool
 
+val expand_interpolations_in_file_full: env -> src:filename -> dst:filename -> unit
+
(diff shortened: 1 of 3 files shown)


src/format/opamFilter.mli
```diff
@@ -129,6 +129,8 @@ val ident_string: ?default:string -> env -> fident -> string
 (** Like [ident_value], but casts the result to a bool *)
 val ident_bool: ?default:bool -> env -> fident -> bool
 
+val expand_interpolations_in_file_full: env -> src:filename -> dst:filename -> unit
```
kit-ty-kate, line 132:
```suggestion
val expand_interpolations_in_file_full: env -> src:filename -> dst:filename -> unit
(** Same as [expand_interpolations_in_file] but allows to set the source [src] and
    destination [dst] files independently instead of implying [src] = [dst].in *)
```

kit-ty-kate (approved):

kit-ty-kate:
Thanks!

PR #5633, opened 2023-08-15
Expose string_interp_regex

gridbugs:
This function is needed by dune to parse variable interpolations from opam fields.


kit-ty-kate (approved):

kit-ty-kate:
Thanks!

Issue #5462, opened 2023-03-05
ensure /tmp is writeable from the sandbox


avsm:
https://github.com/ocaml/opam/pull/4589 modified TMPDIR handling such that `/tmp` is no longer accessible from within the sandbox, preferring to enforce the use of `$TMPDIR` instead.

The FHS does [require that /tmp exists and is writeable](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/ch03s18.html), but does not mandate how large it can be.  Some programs, such as gappa (see https://github.com/ocaml/opam-repository/pull/23433) create domain sockets in /tmp without using TMPDIR, and are broken by the sandbox change in #4589.

I think we should go back to `/tmp` being writeable, to be more FHS-friendly.  The default behaviour can still be to use `/tmp-opam` as the scripts do right now, but with an addition `/tmp` mounted in.

PR #5634, opened 2023-08-15
Sandboxes: Make /tmp writable to restore POSIX compliancy

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/5462

This PR still fixes the issue described in https://github.com/ocaml/opam/pull/4589 while restoring POSIX compliency as described in https://github.com/ocaml/opam/issues/5462 by making `/tmp` writable and propagate its content regardless of the previous value of `TMPDIR`.

The value of `TMPDIR` does not change and still provides a cleared, fast and unshared directory on Linux (the macOS sandbox is not expressive enough to be able to do that but that behaviour does not change either)


avsm:
Looks right to me, and I agree that making /tmp writeable according to POSIX is the right move for the sandbox.

AltGr (commented):
In normal use this seems OK ;
the bit I still have uncertainty about is that it seems to me that this can relax the effectfulness of the sandbox *in the case where your opamroot is in /tmp*

What could happen (please correct if that's not the case!) is that `/tmp` is rw, `/tmp/opamroot` is rw, and only `/tmp/opamroot/myswitch` is rebound ro. As a consequence, the sandboxed process would be able to alter the opam root (outside of the switch).

I am not sure this is a problem for a few reasons:
- an opamroot in /tmp is not expected for production use anyway I guess
- the sandbox is disabled anyway in tests ?

still, if one was to use an opamroot in /tmp to check that the package correctly followed the sandboxing rules, and the package build rules would run say `opam switch create`, the test wouldn't catch it (or so I presume)

I don't know if this is a real problem, but it should at least be exposed and discussed

kit-ty-kate:
> What could happen (please correct if that's not the case!) is that `/tmp` is rw, `/tmp/opamroot` is rw, and only `/tmp/opamroot/myswitch` is rebound ro. As a consequence, the sandboxed process would be able to alter the opam root (outside of the switch).

fair point. https://github.com/ocaml/opam/pull/5634/commits/f69b92af09f127d469043bf29a04ce81409f5b28 takes care of this problem. Whether the opam root is defined through `OPAMROOT` or `--root` the `OPAMROOT` variable will always be present when it is in its non-default place.

AltGr (approved):

rjbou (approved):

Issue #5600, opened 2023-07-13
`make cold` on Windows: flexlink error


rjbou:
_Originally posted by @XantreGodlike in https://github.com/ocaml/opam/discussions/5166#discussioncomment-6378760_

I can't build it. I followed instruction but has this error:
```log
$ make cold
env MAKE=make BOOTSTRAP_EXTRA_OPTS= BOOTSTRAP_TARGETS=world.opt BOOTSTRAP_ROOT=.. BOOTSTRAP_DIR=bootstrap ./shell/bootstrap-ocaml.sh
configure: Configuring OCaml version 4.14.1
checking build system type... x86_64-pc-cygwin
checking host system type... x86_64-pc-cygwin
checking target system type... x86_64-pc-cygwin
checking for ld... ld
checking how to print strings... printf
checking for gcc... x86_64-w64-mingw32-gcc.exe
checking whether the C compiler works... yes
checking for C compiler default output file name... a.exe
checking for suffix of executables... .exe
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether we are using the GNU C compiler... yes
checking whether x86_64-w64-mingw32-gcc.exe accepts -g... yes
checking for x86_64-w64-mingw32-gcc.exe option to accept ISO C89... none needed
checking for a sed that does not truncate output... /usr/bin/sed
checking for grep that handles long lines and -e... /usr/bin/grep
checking for egrep... /usr/bin/grep -E
checking for fgrep... /usr/bin/grep -F
checking for ld used by x86_64-w64-mingw32-gcc.exe... ld
checking if the linker (ld) is GNU ld... yes
checking for BSD- or MS-compatible name lister (nm)... /usr/bin/nm -B
checking the name lister (/usr/bin/nm -B) interface... BSD nm
checking whether ln -s works... yes
checking the maximum length of command line arguments... 8192
checking how to convert x86_64-pc-cygwin file names to x86_64-pc-cygwin format... func_convert_file_noop
checking how to convert x86_64-pc-cygwin file names to toolchain format... func_convert_file_noop
checking for ld option to reload object files... -r
checking for objdump... objdump
checking how to recognize dependent libraries... file_magic ^x86 archive import|^x86 DLL
checking for dlltool... dlltool
checking how to associate runtime and link libraries... func_cygming_dll_for_implib
checking for ar... ar
checking for archiver @FILE support... @
checking for strip... strip
checking for ranlib... ranlib
checking for gawk... gawk
[...]

Issue #5628, opened 2023-08-08
`make cold` on Windows: inconsistent assumptions over implementation Stdlib__Sys


talex5:
I'm trying to follow the instructions at https://opam.ocaml.org/blog/opam-2-2-0-alpha2/ to install opam 2.2 on Windows (as there didn't seem to be any Windows binaries at https://github.com/ocaml/opam/releases/tag/2.2.0-alpha2).

I started with a VM with OCaml 4.14 and cygwin pre-installed. I used the cygwin GUI to install the requested packages and then ran `make cold`. That failed after a bit with:

```
[...]
checking for x86_64-w64-mingw32-i686-w64-mingw32-gcc... no
checking for i686-w64-mingw32-gcc... no
configure: error: amd64 C compiler not found - opam-putenv cannot be built
```

I then used the cygwin GUI to install the main gcc package and also `mingw64-i686-gcc-g++` (it wasn't clear to me if both were needed or not, but the configure script was checking for both so I installed both).

Then I did `make clean` and `make cold`, which ran for a while and then failed with:
```
../../ocamlopt.opt.exe -g -nostdlib -I ../../stdlib -strict-sequence -principal -absname -w +a-4-9-40-41-42-44-45-48-66-70 -warn-err
or +A -bin-annot -safe-string -strict-formats -I byte -I native -pack -o native/dynlink_compilerlibs.cmx dynlink_compilerlibs/binuti
ls.cmx dynlink_compilerlibs/config.cmx dynlink_compilerlibs/build_path_prefix_map.cmx dynlink_compilerlibs/misc.cmx dynlink_compiler
libs/identifiable.cmx dynlink_compilerlibs/numbers.cmx dynlink_compilerlibs/arg_helper.cmx dynlink_compilerlibs/clflags.cmx dynlink_
compilerlibs/profile.cmx dynlink_compilerlibs/consistbl.cmx dynlink_compilerlibs/terminfo.cmx dynlink_compilerlibs/warnings.cmx dynl
ink_compilerlibs/local_store.cmx dynlink_compilerlibs/load_path.cmx dynlink_compilerlibs/int_replace_polymorphic_compare.cmx dynlink
_compilerlibs/lazy_backtrack.cmx dynlink_compilerlibs/location.cmx dynlink_compilerlibs/longident.cmx dynlink_compilerlibs/docstring
[...]

PR #5635, opened 2023-08-15
Fix "make cold" on Windows when gcc is available

kit-ty-kate:
Partially f.i.x.e.s. #5628
Fixes https://github.com/ocaml/opam/issues/5600


rjbou (approved):
After discussing it, seems ok from my perspective

Issue #5602, opened 2023-07-13
Variable expansion in env fields on Windows


rjbou:
If we take the example of pkg-config, there is 
```
setenv: PKG_CONFIG_PATH += "%{lib}%/pkgconfig"
```
On `cmd`, `opam env` gives
```
set PKG_CONFIG_PATH=C:\opamroot\switch\lib/pkgconfig
```
opam should handle present separators in opam files?

MisterDA:
Duplicate of #4690?

kit-ty-kate:
@AltGr and I discussed that at the meeting and we can’t really find a totally satisfying solution that doesn’t involve changing the opam syntax. The best we could find would be to automatically change `/` to `\` on Windows if:
* the string is in `build-env` or `setenv`
* the string is not a triple-double-quote type string (requires a small change in the API of `opam-file-format` but not to the syntax)

Another solution i just thought of after the meeting, would be to use a builtin extra `x-*` field. For example:
```
setenv: PKG_CONFIG_PATH += "%{lib}%/pkgconfig"
x-allow-env-path-rewrite-on-windows: true
```
This way the rewriting is made explicit (`x-allow-env-path-rewrite-on-windows` would be false by default). What do you think @AltGr?

Overall discussion also related to https://github.com/ocaml/opam/pull/2927

AltGr:
Hm the additional field feels slightly clunky, but has the good point of being completely explicit and allowing a reasonable workaround to unblock this.

We might consider making it `true` by default though ? While less satisfactory in that it allows some magic "by default", that remains limited to the Windows scope where we can assume the package wouldn't have worked anyway...

avsm:
This seems like a good reason to bump `opam-version` and adopt the new semantics if the package adopts the higher version number.

kit-ty-kate:
> This seems like a good reason to bump `opam-version` and adopt the new semantics if the package adopts the higher version number.

We can’t do that because opam 2.0 and 2.1 don’t support having opam files with unsupported versions (https://github.com/ocaml/opam/issues/5631), and doing a full repository transition like it was done for the transition from opam 1.2 to opam 2.0 requires tools, resources and time we do not have.

rjbou:
`setenv` is also defined for `switch-config` files.

PR #5636, opened 2023-08-17
Path rewriting for `setenv:` and `build-env:`

rjbou:
Implementation of proposal described in #5602
+ add lint to check that the `x-` field is well defined

todo:
* [ ] add proper test
* [ ] handle & check no modification on parse/print
* [ ] update changes
* [ ] check for another name


src/format/opamFile.ml
```diff
@@ -3203,6 +3203,60 @@ module OPAMSyntax = struct
     in
     Pp.pp parse print
 
+  let rewrite_xfield = "x-allow-env-path-rewrite-on-windows"
+  let handle_env_paths =
+    let pp_constraint =
+      pp_minimal_opam_version (OpamVersion.of_string "2.2")
```
rjbou, line 3209: Not sure if we need that constraint, as it won't be syntaxically broken if read by opam < 2.2, just may not work on windows, what is the current status of 2.1.

rjbou: From dev meeting: remove this constraint, have a case by case availability guard in repo opam files.

src/client/opamAdminRepoUpgrade.ml
```diff
@@ -372,6 +372,12 @@ let do_upgrade repo_root =
         "CAML_LD_LIBRARY_PATH", PlusEq, "%{lib}%/stublibs", None;
         "OCAML_TOPLEVEL_PATH", Eq, "%{toplevel}%", None;
       ] |>
+      O.with_extensions
+        (OpamStd.String.Map.singleton OpamFile.OPAM.rewrite_xfield
+           (OpamTypesBase.nullify_pos (OpamParserTypes.FullPos.Bool true))) |>
+      O.with_available
+        (FOp (FIdent ([], OpamVariable.of_string "opam-version", None),
+              `Geq, FString "2.2")) |>
```
rjbou, line 380: should we also have that addition for 1.2 comp -> 2.0 opam files ? see https://github.com/ocaml/opam/blob/8a23fc031b80abcaaf0e8c96059d02eea60d3fc6/src/state/opamFormatUpgrade.ml#L1396-L1398

rjbou: from dev meeting: to remove for ocaml package on opam admin upgrade, defined variables are package/switch variables.

src/state/opamFileTools.ml
```diff
@@ -875,6 +875,42 @@ let t_lint ?check_extra_files ?(check_upstream=false) ?(all=false) t =
...
+               t.extensions, env with
+       | Some {pelem = Bool _; _}, _ -> []
+       | (Some _| None), _::_ ->
+         List.filter (fun (_, _, content, _) ->
+             OpamStd.String.contains_char content '/')
+           env
+       | _, _ -> []
+     in
+     cond 69 `Warning
+       "setenv or build-env contains paths and no rewriting rule defined"
```
rjbou, line 890: better wording ?

src/state/opamFileTools.ml
```diff
@@ -875,6 +875,42 @@ let t_lint ?check_extra_files ?(check_upstream=false) ?(all=false) t =
...
+             | _ -> acc)
+           true f
+       in
+       match OpamStd.String.Map.find_opt OpamFile.OPAM.rewrite_xfield
+               t.extensions with
+       | Some {pelem = Bool true; _} -> opamv t.available
+       | _ -> false
+     in
+     cond 70 `Warning
+       "Rewrite x- field set to true, but no availability guard set"
```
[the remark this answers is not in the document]
kit-ty-kate:
```suggestion
       "x-allow-env-path-rewrite-on-windows is set to true but opam-version >= "2.2" is missing from the available field"
```

rjbou:
Updated

src/client/opamAdminRepoUpgrade.ml
rjbou: Finally, i've added an `x-allow-rewrite: false` to avoid new lint warning.

src/format/opamFile.ml
```diff
@@ -3203,6 +3203,68 @@ module OPAMSyntax = struct
...
+      in
+      let b2f (_v,_op,content,_comment) =
+        _v, _op,
+        replace ~from:'/' ~to_:"\\" content,
+        _comment
+      in
+      let f2b (_v,_op,content,_comment) =
+        _v, _op,
+        replace ~from:'\\' ~to_:"/" content,
+        _comment
```
kit-ty-kate, line 3231: Why not simply use the functions from `OpamSystem` that are already there?
```suggestion
      let b2f (v, op, content, comment) =
        (v, op, OpamSystem.back_to_forward content, comment)
      in
      let f2b (v, op, content, comment) =
        (v, op, OpamSystem.forward_to_back content, comment)
```

rjbou: The real answer is: i momentarily forget them.
Another answer is that those functions doesn't handle escaped slash or backslash. Maybe worth adding that too ?

kit-ty-kate:
```suggestion
    let back_to_forward, forward_to_back =
      let replace ~from ~to_ content =
        let re =
          Re.(compile @@
              alt [
                seq [ bos; char from ];
                seq [ group (diff any (char '\\')); char from ]
              ])
        in
        let f gr =
          Printf.sprintf "%s%s"
            (try Re.Group.get gr 1 with Not_found -> "") to_
        in
        Re.replace re ~f content
      in
      let b2f (v, op, content, comment) =
        v, op,
        replace ~from:'/' ~to_:"\\" content,
        comment
      in
      let f2b (v, op, content, comment) =
        v, op,
        replace ~from:'\\' ~to_:"/" content,
        comment
```

kit-ty-kate (approved):

src/format/opamFile.ml
AltGr: Hm here we are handling an already parsed string, so I don't think escapes should be an issue ?

src/state/opamFileTools.ml
```diff
@@ -875,6 +875,47 @@ let t_lint ?check_extra_files ?(check_upstream=false) ?(all=false) t =
...
+           true f
+       in
+       match OpamStd.String.Map.find_opt OpamFile.OPAM.rewrite_xfield
+               t.extensions with
+       | Some {pelem = Bool true; _} -> opamv t.available
+       | _ -> false
+     in
+     cond 70 `Warning
+       "x-allow-env-path-rewrite-on-windows is set to true but \
+        opam-version >= \"2.2\" is missing from the available field"
```
AltGr, line 917: IIRC the convention is to write `fieldname:` with the colon ?

AltGr (approved):
LGTM, but... where is the doc ?
In it it may also be appropriate to mention the reason for an `x-*` field (the manual specifies that it's normally for external tools IIRC)

rjbou:
To not change the opam file syntax, opam need to use it's own external tool features ^^

src/format/opamFile.ml
rjbou: The idea is to be able to rewrite `first/still\/second/third` into `first\still/second\third`

rjbou:
Doc updated!

src/format/opamFile.ml
```diff
@@ -3203,6 +3203,49 @@ module OPAMSyntax = struct
     in
     Pp.pp parse print
 
+  let rewrite_xfield = "x-allow-env-path-rewrite-on-windows"
+  let handle_env_paths =
+    let guard_2_1 t k =
+      if OpamVersion.(compare t.opam_version (of_string "2.1") > 0) then t
```
kit-ty-kate, line 3209:
```suggestion
    let guard_2_2 t k =
      if OpamVersion.(compare t.opam_version (of_string "2.2") > 0) then t
```

rjbou:
```suggestion
    let guard_2_2 t k =
      if OpamVersion.(compare t.opam_version (of_string "2.2") >= 0) then t
```

kit-ty-kate (commented):
The manual and the changelog still refer to the old `x-allow-env-path-rewrite-on-windows` field instead of the new one.

src/client/opamConfigCommand.ml
```diff
@@ -638,10 +648,14 @@ let switch_allowed_fields, switch_allowed_sections =
...
+                 let open OpamTypes in
                  let env =
-                   List.filter (fun (vr,op,vl,_) ->
-                       None = OpamStd.List.find_opt (fun (vr',op',vl',_) ->
-                           vr = vr' && op = op' && vl = vl') nc.env) c.env
+                   List.filter (fun upd ->
+                       None = OpamStd.List.find_opt (fun upd' ->
+                           upd.envu_var = upd'.envu_var
+                           && upd.envu_op = upd'.envu_op
+                           && upd.envu_value = upd'.envu_value)
```
kit-ty-kate, line 657:
```suggestion
                           String.equal upd.envu_var upd'.envu_var
                           && (upd.envu_op : OpamParserTypes.env_update_op) = upd'.envu_op
                           && String.equal upd.envu_value upd'.envu_value)
```

src/format/opamFile.ml
```diff
@@ -551,26 +551,97 @@ module Environment = LineFile(struct
...
+        (Pp.of_pair "env_update_op"
+           (OpamLexer.FullPos.env_update_op, OpamPrinter.env_update_op_kind))
+      in
+      let comment_and_sepformat =
+        (separator ^+ host_or_target ^+ (Pp.last -| Pp.identity))
+        -| (Pp.of_pair "comment-sep-ht"
+              ((fun (sep, (ht, comment)) ->
+                  (Some comment, Some (sep, ht))),
+               (function | Some comment, Some (sep, ht) -> sep, (ht, comment)
+                         | _, _ -> failwith "")))
```
kit-ty-kate, line 587: there should probably be a proper message here, or `assert false` if it should not happend.

src/client/opamConfigCommand.ml
kit-ty-kate: I would also be more in favour of matching on the record to make sure that the compiler can give us a warning if a field is added to the record for example.

Something like:
```
let {envu_var; envu_op; envu_value;
     envu_comment = _; envu_separator = _; envu_format = _} = upd in
```

src/format/opamFile.ml
rjbou: for the record: it can't be assert false here as we use `Pp.fallback` that errors on `Assertion_failure`. Added a message to failwith.

rjbou: maybe it need simply `Pp.unexpected ()`

src/format/opamFile.ml
```diff
@@ -3203,6 +3335,66 @@ module OPAMSyntax = struct
     in
     Pp.pp parse print
 
+  let rewrite_xfield = "x-env-path-rewrite"
+  let handle_env_paths =
+    let guard_2_1 t k =
+      if OpamVersion.(compare t.opam_version (of_string "2.1") > 0) then t
+      else k t
+    in
```
kit-ty-kate, line 3343: what is the purpose of this?

doc/pages/Manual.md
```diff
@@ -652,6 +651,54 @@ other system).
...
+<environement-rewrites> ::= { <environement-rewrite> }*
+<environement-rewrite>  ::= <ident> <bool>
+                          | <ident> <separator-formula> <format-formula>
+
+<separator-formula>     ::= <separator-formula> <logop> <separator-formula>
+                          | ( <separator-formula> )
+                          | <separator> { { <distrib-formula> }* }
+<format-formula>        ::= <format-formula> <logop> <format-formula>
+                          | ( <format-formula> )
+                          | <format> { { <distrib-formula> }* }
```
kit-ty-kate, line 671: As far as i could see, only `separator-formula = <separator>` and `format-formula = <format>` can be shown in the tests. I think it would be nice to have the other syntaxes tested for completeness.

rjbou: added

rjbou:
Final update!

doc/pages/Manual.md
```diff
@@ -652,6 +651,61 @@ other system).
 
 `FOO += ""`, `FOO := ""`, etc. are all ignored - i.e. opam never adds empty segments to an existing variable.
 
+#### Environment update portability
+<a id="env-update-rewrite"></a>
+
+Some fields define an environment update portability specification. In the opam
+file it is the [`x-env-path-rewrite:`](#opamfield-x-env-path-rewrite) in the
```
kit-ty-kate, line 658:
```suggestion
file it is the [`x-env-path-rewrite:`](#opamfield-x-env-path-rewrite) field, of the
```

doc/pages/Manual.md
```diff
@@ -652,6 +651,61 @@ other system).
...
+  Windows).
+- host-quoted: use the *host* interpretation of entries and double-quote any
+  entries which include the separator character.
+- target: use the *target* interpretation of entries (i.e. rewrite slashes to
+  backslashes on Windows).
+- target-quoted: use the *target* interpretation of entries and double-quote
+  any entries which include the separator character.
+
+Default separator is `";" | ";" { os = "win32"}` (`;` on Windows, `:` on Cygwin
+or any other system) and default format is `target` with no slashes rewrite.
```
kit-ty-kate, line 694: The wording is a bit confusing. Reading the definition for `target` above, it feels pointless to say the default format is `target` if it’s not going to do anything.
Does that actually do nothing or is there something to say for variables? In which case it might be worth being more explicit.

doc/pages/Manual.md
```diff
@@ -652,6 +651,61 @@ other system).
...
+
+For example, on Windows:
+- `[FOO false]`: `FOO` won't be translated nor rewritten, and default separator is used if needed
+- `[FOO true]`: `FOO` is rewritten using defaults `;` and `target` with slash rewriting
+  - `FOO = "a:/path/to"` -> FOO=a:\path\to
+- `[FOO ":" "target-quoted"]: `FOO` will be append using `:` separator, if the added path contains '/' they are transformed into `\`, and if the added path contains a `:`, it will be quoted
```
kit-ty-kate, line 700:
```suggestion
- `[FOO ":" "target-quoted"]: `FOO` will be appended using `:` separator, if the added path contains '/' they are transformed into `\`, and if the added path contains a `:`, the added path will be quoted
```

doc/pages/Manual.md
```diff
@@ -652,6 +651,61 @@ other system).
...
+  - `FOO += "a/path/to"` -> FOO=a\path\to:R:\previous\path
+  - `FOO += "a:path/to"` -> FOO="a:path\to":R:\previous\path
+- `[FOO ";" "target-quoted"]: `FOO` will be append using `;` separator, if the added path contains '/' they are transformed into `\`, and if the added path contains a `;`, it will be quoted
```
kit-ty-kate, line 703:
```suggestion
- `[FOO ";" "target-quoted"]: `FOO` will be appended using `;` separator, if the added path contains '/' they are transformed into `\`, and if the added path contains a `;`, the added path will be quoted
```

doc/pages/Manual.md
```diff
@@ -652,6 +651,61 @@ other system).
...
+  - `FOO += "a/path/to"` -> FOO=a\path\to;R:\previous\path
+  - `FOO += "a;path/to"` -> FOO="a;path\to";R:\previous\path
+- `[FOO ";" "host"]: `FOO` will be append using `:`, and its path will be translated according host translator, on Windows:
+  - `FOO += "A:\path\to"` -> FOO=/cygdrive/a/path/to:/previous/path
```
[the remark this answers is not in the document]
kit-ty-kate: I’m unsure what’s a host translator though

doc/pages/Manual.md
```diff
@@ -1152,6 +1212,12 @@ files.
        has changed. Even then, this won't unpin any packages that would have
        been removed from `pin-depends:`.
 
+- <a id="opamfield-x-env-path-rewrite">`x-env-path-rewrite: [ <environmenet-update-rewwrite> ... ]`</a>:
```
kit-ty-kate, line 1215:
```suggestion
- <a id="opamfield-x-env-path-rewrite">`x-env-path-rewrite: [ <environment-update-rewrite> ... ]`</a>:
```

kit-ty-kate (commented):
Final review. Looks good otherwise

src/client/opamAction.ml
```diff
@@ -521,31 +521,78 @@ let prepare_package_source st nv dir =
 
 let compilation_env t opam =
   let open OpamParserTypes in
+  let empty = Some (SPF_Resolved None) in
   let build_env =
     List.map (OpamEnv.env_expansion ~opam t) (OpamFile.OPAM.build_env opam)
+    |> List.map OpamEnv.resolve_separator_and_format
```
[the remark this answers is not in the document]
rjbou:
```suggestion
    List.map
      (fun env ->
         OpamEnv.resolve_separator_and_format
           (OpamEnv.env_expansion ~opam t env))
      (OpamFile.OPAM.build_env opam)
```

doc/pages/Manual.md
```diff
@@ -652,6 +651,61 @@ other system).
...
+- `[FOO false]`: `FOO` won't be translated nor rewritten, and default separator is used if needed
+- `[FOO true]`: `FOO` is rewritten using defaults `;` and `target` with slash rewriting
+  - `FOO = "a:/path/to"` -> FOO=a:\path\to
+- `[FOO ":" "target-quoted"]: `FOO` will be appended using `:` separator, if the added path contains '/' they are transformed into `\`, and if the added path contains a `:`, the added path will be quoted
+  - `FOO += "a/path/to"` -> FOO=a\path\to:R:\previous\path
+  - `FOO += "a:path/to"` -> FOO="a:path\to":R:\previous\path
+- `[FOO ";" "target-quoted"]: `FOO` will be appended using `;` separator, if the added path contains '/' they are transformed into `\`, and if the added path contains a `;`, the added path will be quoted
+  - `FOO += "a/path/to"` -> FOO=a\path\to;R:\previous\path
+  - `FOO += "a;path/to"` -> FOO="a;path\to";R:\previous\path
+- `[FOO ":" "host"]: `FOO` will be appended using `:`, and its path will be translated according to the host translator, on Windows:
```
rjbou, line 706:
```suggestion
- `[FOO ":" "host"]: `FOO` will be appended using `:`, and its path will be translated according to the host translator, i.e `cygpath <path>`:
```

src/format/opamFile.ml
```diff
@@ -3203,6 +3335,63 @@ module OPAMSyntax = struct
...
+      in
+      Pp.V.map_list @@
+      Pp.fallback
+        (norewrite @@ Pp.V.map_pair Pp.V.ident Pp.V.bool)
+        (rewrite @@ Pp.V.map_triple Pp.V.ident
+           (filtered Pp.V.separator) (filtered Pp.V.path_format))
+    in
+    let parse ~pos t =
+      if OpamVersion.(compare t.opam_version (of_string "2.0") > 0) then t
+      else
```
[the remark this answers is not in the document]
rjbou: It is the guard to do no transformation if an opam file is 1.2.
Do we want to enable this "feature" for 1.2 files ?

doc/pages/Manual.md
rjbou: `target` not explicitly specified is a no-op indeed. I've worded it like that because in the code the default is `target` with no rewrite. Maybe not worth keep it in that way in the doc.

rjbou: Should we add a word on specific case ? `PKG_CONFIG_PATH` (`:`, `target-quoted`) and `PATH` on windows (`;`, `target_quoted`)?

src/core/opamSystem.mli
```diff
@@ -194,7 +194,7 @@ val get_cygpath_function: command:string -> (string -> string) lazy_t
     if in a functioning Cygwin or MSYS2 environment, translating the Windows or a
     Unix PATH variable into a Unix PATH variable. Returns the identity function
     otherwise. *)
-val get_cygpath_path_transform: (string -> string) lazy_t
+val get_cygpath_path_transform: (pathlist:bool -> string -> string) lazy_t
```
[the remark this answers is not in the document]
rjbou: done

src/format/opamFile.ml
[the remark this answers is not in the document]
kit-ty-kate: > It is the guard to do no transformation if an opam file is 1.2.

Mmh, maybe I'm missing something but the code doesn't seem to do that. The code does no transformation if the `opam-version` field is strictly greater than `2.0`. In practice this means that opam files with `opam-version: "2.1"` (as allowed by opam-file-format) will skip this transformation.

doc/pages/Manual.md
kit-ty-kate: > Maybe not worth keep it in that way in the doc.

I think so. In my opinion documentations should aim to provide a pointer to behaviours of a system as far removed from the code as possible.

> Should we add a word on specific case ?

sure!

doc/pages/Manual.md
```diff
@@ -1152,6 +1212,12 @@ files.
        has changed. Even then, this won't unpin any packages that would have
        been removed from `pin-depends:`.
 
+- <a id="opamfield-x-env-path-rewrite">`x-env-path-rewrite: [ <environment-update-rewrite> ... ]`</a>:
+  a specific extra field, used to specify if environment variables defined in
+  [`setenv:`](#opamfield-setenv) and [`build-env`](#opamfield-build-env)
+  rewrite rules. See [`environment update portability`](#env-update-rewrite)
+  for syntax.
```
dra27, line 1219:
```suggestion
  a specific extra field, used to specify rewrite rules for environment variables
  defined in [`setenv:`](#opamfield-setenv) and [`build-env`](#opamfield-build-env).
  See [`environment update portability`](#env-update-rewrite) for syntax.
```

doc/pages/Manual.md
```diff
@@ -652,6 +651,61 @@ other system).
...
+  Windows).
+- host-quoted: use the *host* interpretation of entries and double-quote any
+  entries which include the separator character.
+- target: use the *target* interpretation of entries (i.e. rewrite slashes to
+  backslashes on Windows).
+- target-quoted: use the *target* interpretation of entries and double-quote
+  any entries which include the separator character.
+
+Default separator is `";" | ";" { os = "win32"}` (`;` on Windows, `:` on Cygwin
+or any other system) and default format is `target` with no slashes rewrite.
```
rjbou, line 694:
```suggestion
If a variable is not mentioned in `x-env-path-rewrite`, the separator is assumed to be `;` on Windows and `:` on all other systems; no slash or quoting transformations are performed. There is two special default cases:
* `PKG_CONFIG_PATH` uses `:` separator and is `target-quoted`
* `PATH` on Windows uses `;` separator and is `target-quoted`
```

doc/pages/Manual.md
```diff
@@ -652,6 +651,61 @@ other system).
...
+For example, on Windows:
+- `[FOO false]`: `FOO` won't be translated nor rewritten, and default separator is used if needed
+- `[FOO true]`: `FOO` is rewritten using defaults `;` and `target` with slash rewriting
+  - `FOO = "a:/path/to"` -> FOO=a:\path\to
+- `[FOO ":" "target-quoted"]: `FOO` will be appended using `:` separator, if the added path contains '/' they are transformed into `\`, and if the added path contains a `:`, the added path will be quoted
+  - `FOO += "a/path/to"` -> FOO=a\path\to:R:\previous\path
+  - `FOO += "a:path/to"` -> FOO="a:path\to":R:\previous\path
+- `[FOO ";" "target-quoted"]: `FOO` will be appended using `;` separator, if the added path contains '/' they are transformed into `\`, and if the added path contains a `;`, the added path will be quoted
+  - `FOO += "a/path/to"` -> FOO=a\path\to;R:\previous\path
+  - `FOO += "a;path/to"` -> FOO="a;path\to";R:\previous\path
```
rjbou, line 705:
```suggestion
```

src/format/opamFile.ml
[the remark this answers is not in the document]
rjbou: Maybe it should be, like that we ensure that it is only for `2.0` files ?
```suggestion
      if not OpamVersion.(equal t.opam_version (of_string "2.0")) then t
      else
```

src/format/opamFile.ml
```diff
@@ -3203,6 +3335,63 @@ module OPAMSyntax = struct
...
+              | None -> upd)
+            env
+        in
+        { t with
+          env = update t.env;
+          build_env = update t.build_env;
+        }
+    in
+    let print t = t in (* extension field is rewritten as is *)
+    Pp.pp parse print
```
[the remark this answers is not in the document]
kit-ty-kate: In this case I think the more explicit version is better as it allows to know exactly what is being left untouched when reading the code

src/client/opamAction.ml
```diff
@@ -521,31 +521,81 @@ let prepare_package_source st nv dir =
...
   let cygwin_env =
     match OpamSysInteract.Cygwin.cygbin_opt t.switch_global.config with
     | Some cygbin ->
-      [ "PATH", EqPlus, OpamFilename.Dir.to_string cygbin, Some "Cygwin path" ]
+      [{ envu_var = "PATH";
+         envu_op = EqPlus;
+         envu_value = OpamFilename.Dir.to_string cygbin;
+         envu_comment = Some "Cygwin path";
+         envu_rewrite = empty;
+       }]
```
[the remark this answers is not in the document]
rjbou: Added in a commit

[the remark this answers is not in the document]
rjbou: I don't know which version is more easily readable

src/format/opamFile.ml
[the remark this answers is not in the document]
rjbou: agree, plus it permits to keep the `parse/print` explicit functions of pps

doc/pages/Manual.md
rjbou: cf https://github.com/ocaml/opam/pull/5636/files#r1389505791

doc/pages/Manual.md
```diff
@@ -652,6 +651,61 @@ other system).
...
+- host: use the *host* interpretation of PATHs (i.e. convert via `cygpath` on
+  Windows).
+- host-quoted: use the *host* interpretation of entries and double-quote any
+  entries which include the separator character.
+- target: use the *target* interpretation of entries (i.e. rewrite slashes to
+  backslashes on Windows).
+- target-quoted: use the *target* interpretation of entries and double-quote
+  any entries which include the separator character.
+
+Default separator is `";" | ";" { os = "win32"}` (`;` on Windows, `:` on Cygwin
```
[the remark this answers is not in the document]
rjbou: cf [comment](https://github.com/ocaml/opam/pull/5636/files#r1389505791)

src/client/opamAction.ml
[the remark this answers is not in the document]
dra27: Sorry - looking at it again, I now realise that I'd forgotten that a GADT was involved and that we can't convey that in optional parameters! (yet... it's been on my wish list for quite a few years...)

Wouldn't it be OK not to label those three arguments, though> I think `env_update_resolved "PATH" EqPlus (OpamFilename.Dir.to_string cygbin` quite clearly conveys what it's doing to give:

```ocaml
      [ OpamTypesBase.env_update_resolved ~comment:"Cygwin path"
          "PATH" EqPlus (OpamFilename.Dir.to_string cygbin) ]
```

?

doc/pages/Manual.md
dra27: Looks nice - highly pedantically "There _are_ two special default cases"

src/client/opamAction.ml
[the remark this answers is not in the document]
rjbou: It is possible. The labels came at first because of var & values are string, and like that there is no confusion between them.
I'm ok to remove them and have then
```ocaml
val env_update_resolved:
  ?comment:string -> ?rewrite:spf_resolved separator_path_format option
  -> string -> env_update_op_kind -> string
  -> spf_resolved env_update
```

doc/pages/Manual.md
```diff
@@ -652,6 +651,61 @@ other system).
...
+For example, on Windows:
+- `[FOO false]`: `FOO` won't be translated nor rewritten, and default separator is used if needed
+- `[FOO true]`: `FOO` is rewritten using defaults `;` and `target` with slash rewriting
+  - `FOO = "a:/path/to"` -> FOO=a:\path\to
+- `[FOO ":" "target-quoted"]: `FOO` will be appended using `:` separator, if the added path contains '/' they are transformed into `\`, and if the added path contains a `:`, the added path will be quoted
+  - `FOO += "a/path/to"` -> FOO=a\path\to:R:\previous\path
+  - `FOO += "a:path/to"` -> FOO="a:path\to":R:\previous\path
+- `[FOO ";" "target-quoted"]: `FOO` will be appended using `;` separator, if the added path contains '/' they are transformed into `\`, and if the added path contains a `;`, the added path will be quoted
+  - `FOO += "a/path/to"` -> FOO=a\path\to;R:\previous\path
+  - `FOO += "a;path/to"` -> FOO="a;path\to";R:\previous\path
```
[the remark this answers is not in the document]
rjbou:
```suggestion
- `[FOO ":" "target-quoted"]: `FOO` will be appended using `:` separator, if the added path contains '/' they are transformed into `\`, and if the added path contains a `:`, the added path will be quoted
  - `FOO += "a/path/to"` -> FOO=a\path\to:R:\previous\path
  - `FOO += "a:path/to"` -> FOO="a:path\to":R:\previous\path
```

src/format/opamFile.ml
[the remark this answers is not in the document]
rjbou: one is opam-version > 2.0, the other opam-version <> 2.0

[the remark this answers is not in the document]
kit-ty-kate: i meant when using `opam-version: "2.1"`

kit-ty-kate:
Thanks!

PR #5639, opened 2023-08-22
Fix extra-files handling on linting

rjbou:
\+ add a lint test for that

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -46,6 +46,7 @@ users)
 ## Source
 
 ## Lint
+  * Fix extra-files handling when linting packages from repositories, see #5068 [#5639 @rjbou]
 
 ## Repository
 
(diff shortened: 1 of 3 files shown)


kit-ty-kate (approved):

master_changes.md
```diff
@@ -46,6 +46,7 @@ users)
 ## Source
 
 ## Lint
+  * Fix extra-files handling when linting packages from repository with tarred repositories enabled, see #5068 [#5639 @rjbou]
```
kit-ty-kate, line 49: Actually according to the tests you can remove that section. It was also failing when tarring was disabled
```suggestion
  * Fix extra-files handling when linting packages from repositories, see #5068 [#5639 @rjbou]
```

rjbou: true, it was copy pasting :)

kit-ty-kate:
Thanks

PR #5642, opened 2023-08-24
reftests: add test for operators in opamfile filters

rjbou:


tests/reftests/filter-operators.test
```diff
@@ -0,0 +1,327 @@
...
+]
+### opam install gteq
+The following actions would be performed:
+=== install 5 packages
+  - install a    1 [required by gteq]
+  - install b    1 [required by gteq]
+  - install c    1 [required by gteq]
+  - install gteq 1
+  - install ok   1 [required by gteq]
+### :: Combination ::
```
dra27, line 237: I think it's definitely good to have some tests for this, but I'm not sure that reftests are the correct harness? Wouldn't it be better to be testing the filter parser then evaluation of filters separately (alcotest-style or some such), rather than invoking the whole of opam machinery to test this?

I'm pretty sure the tests above this comment would _definitely_ be clearer as a standard test case harness and I'm fairly even the tests below would be?

(i.e. I think the reftests framework is excellent for full-blown tests, but these feel more like unit tests and the framework's overkill for that?)

kit-ty-kate: I think it's fine to have in the reftest since it's something that is usable and most used from the command line tool

kit-ty-kate (approved):

kit-ty-kate:
Thanks!

PR #5643, opened 2023-08-24
reftest: add test for variables resolutions in filter, for all fields

rjbou:


src/core/opamSystem.ml
```diff
@@ -1637,7 +1637,9 @@ let gpatch = lazy begin
   let rec search_gpatch = function
     | [] -> None
     | patch_cmd::patch_cmds ->
-      match OpamProcess.run (make_command ~name:"patch" patch_cmd ["--version"]) with
+      match OpamProcess.run
+              (make_command ~verbose:OpamCoreConfig.(!r.verbose_level >= 3)
+                 ~name:"patch" patch_cmd ["--version"]) with
```
dra27, line 1642: It doesn't feel like a good to stop having `-vv` display _all_ commands run... is there anywhere else we already do this?

dra27 (commented):
I haven't gone through the test properly, but unlike #5642, I think this one is a good candidate for a reftest. I half wondered with the tests which produce messages if it would be worth piping the whole thing through an appropriate grep just to keep the output smaller?

src/core/opamSystem.ml
kit-ty-kate: I don't think this change should be part of this PR

kit-ty-kate: I've removed that part from the PR. Feel free to revert to the previous commit if you prefer to have the previous version

kit-ty-kate (approved):

src/core/opamSystem.ml
rjbou: After checking, not really, the other instances are for verbose level >= 2. The only instance where it is >=3 is for commands run by depext, to avoid the big log.

tests/reftests/filter-variable-resolution.test
```diff
@@ -0,0 +1,1111 @@
...
++ patch "--version"
+- GNU patch
+- Copyright (C) 2003, 2009-2012 Free Software Foundation, Inc.
+- Copyright (C) 1988 Larry Wall
+- 
+- License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>.
+- This is free software: you are free to change and redistribute it.
+- There is NO WARRANTY, to the extent permitted by law.
+- 
+- Written by Larry Wall and Paul Eggert
```
rjbou, line 64: Is that chunk always the same cross platform? At least on ubuntu, macos & cygwin from CI run

kit-ty-kate: it looks like yeah

tests/reftests/run.ml
```diff
@@ -151,13 +151,17 @@ type filt_sort =
...
   let buf = Buffer.create (String.length s * 2) in
   String.iter (function
     | ('|' | '(' | ')' | '*' | '+' | '?'
     |  '[' | '.' | '^' | '$' | '{' | '\\') as c -> Buffer.add_char buf '\\'; Buffer.add_char buf c
     | c -> Buffer.add_char buf c
   ) s;
   Buffer.contents buf
+  in
+  OpamConsole.error "%S -> %S" s r;
+  r
```
kit-ty-kate, line 164: this section should be reindented. Same for the similar changes in this file

tests/reftests/filter-variable-resolution.test
```diff
@@ -0,0 +1,1087 @@
...
+### <pkg:patches.1>
+opam-version: "2.0"
+patches: [
+  "to-apply.patch" { ?smtg }
+  "to-not-apply.patch" { !?smtg }
+]
+build: [
+  [ "grep" "ploplop" "some-content" ]
+  [ "grep" "bloblob" "some-content" ]
+]
```
kit-ty-kate, line 301: This section is already defined in line 18 of the same file

tests/reftests/filter-variable-resolution.test
```diff
@@ -0,0 +1,1087 @@
...
+-> compiled  patches.1
+-> installed patches.1
+Done.
+### :II:2: build :::
+### <pkg:build.1>
+opam-version: "2.0"
+build: [
+  [ "true" { ?smtg } ]
+  [ "false" { !?smtg } ]
+]
```
kit-ty-kate, line 331: already defined above

tests/reftests/filter-variable-resolution.test
```diff
@@ -0,0 +1,1087 @@
...
+-> compiled  build.1
+-> installed build.1
+Done.
+### :II:3: install :::
+### <pkg:install.1>
+opam-version: "2.0"
+install: [
+  [ "true" { ?smtg } ]
+  [ "false" { !?smtg } ]
+]
```
kit-ty-kate, line 349: same

tests/reftests/filter-variable-resolution.test
```diff
@@ -0,0 +1,1087 @@
...
++ true (CWD=${BASEDIR}/OPAM/switch-variable/.opam-switch/build/install.1)
+-> installed install.1
+Done.
+### :II:4: runtest :::
+### <pkg:runtest.1>
+opam-version: "2.0"
+run-test: [
+  [ "true" { ?smtg } ]
+  [ "false" { !?smtg } ]
+]
```
kit-ty-kate, line 366: same

tests/reftests/filter-variable-resolution.test
```diff
@@ -0,0 +1,1087 @@
...
+-> compiled  runtest.1
+-> installed runtest.1
+Done.
+### :II:5: remove :::
+### <pkg:remove.1>
+opam-version: "2.0"
+remove: [
+  [ "true" { ?smtg } ]
+  [ "false" { !?smtg } ]
+]
```
kit-ty-kate, line 384: same

tests/reftests/filter-variable-resolution.test
```diff
@@ -0,0 +1,1087 @@
...
++ true (CWD=${BASEDIR}/OPAM/switch-variable/.opam-switch/remove/remove.1)
+-> removed   remove.1
+Done.
+### :II:6: depends :::
+### <pkg:depends.1>
+opam-version: "2.0"
+depends: [
+ "foo" { ?smtg }
+ "undefined-smtg" { !?smtg }
+]
```
kit-ty-kate, line 409: same

tests/reftests/filter-variable-resolution.test
```diff
@@ -0,0 +1,1087 @@
...
+=== install 2 packages
+  - install depends 1
+  - install foo     1 [required by depends]
+### :II:7: depopts :::
+### <pkg:depopts.1>
+opam-version: "2.0"
+depopts: [
+ "bar" { ?smtg }
+ "undefined-smtg" { !?smtg }
+]
```
kit-ty-kate, line 421: same

tests/reftests/filter-variable-resolution.test
```diff
@@ -0,0 +1,1087 @@
...
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> removed   depopts.1
+Done.
+### :II:8: conflicts :::
+### <pkg:conflicts.1>
+opam-version: "2.0"
+conflicts: [
+  "baz" { ?smtg }
+  "undefined-smtg" { !?smtg }
+]
```
kit-ty-kate, line 454: same

tests/reftests/filter-variable-resolution.test
```diff
@@ -0,0 +1,1087 @@
...
+The following actions would be performed:
+=== install 1 package
+  - install undefined-smtg 1
+### :II:9: msg :::
+### <pkg:msg.1>
+opam-version: "2.0"
+messages: [
+  "Hej!" { ?smtg }
+  ">>smtg is undefined<<" { !?smtg }
+]
```
kit-ty-kate, line 479: same

tests/reftests/filter-variable-resolution.test
```diff
@@ -0,0 +1,1087 @@
...
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> installed msg.1
+Done.
+### :II:10: post-msg :::
+### <pkg:post-msg.1>
+opam-version: "2.0"
+post-messages: [
+  "Bye!" { ?smtg }
+  ">>smtg is undefined<<" { !?smtg }
+]
```
kit-ty-kate, line 494: same

tests/reftests/filter-variable-resolution.test
```diff
@@ -0,0 +1,1087 @@
...
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> installed post-msg.1
+Done.
+
+<><> post-msg.1 installed successfully ><><><><><><><><><><><><><><><><><><><><>
+=> Bye!
+### :II:11: avail :::
+### <pkg:avail.1>
+opam-version: "2.0"
+available: [ ?smtg ]
```
kit-ty-kate, line 509: same

tests/reftests/filter-variable-resolution.test
```diff
@@ -0,0 +1,1087 @@
...
+-pioupiou
++ploplop
+ bloblob
+### <pkg:patches.1:to-not-apply.patch>
+--- a/some-content	2020-12-02 14:22:55.364620832 +0100
+++ b/some-content	2020-12-02 14:23:05.668686881 +0100
+@@ -2,3 +2,3 @@
+ ploplop
+-bloblob
++noooooo
```
kit-ty-kate, line 554: these files don't have to be redefined

tests/reftests/filter-variable-resolution.test
```diff
@@ -0,0 +1,1087 @@
...
```
kit-ty-kate, line 799: these files don't have to be redefined

tests/reftests/filter-variable-resolution.test
```diff
@@ -0,0 +1,1087 @@
...
+### :III:11: avail :::
+### <pkg:avail.1>
+opam-version: "2.0"
+available: [ ?v:smtg ]
+### opam install avail
+[ERROR] avail: unmet availability conditions: '?v:smtg'
+# Return code 5 #
+### :::::::::::::::::::::::::::
+### :: Self package variable ::
+### :::::::::::::::::::::::::::
```
kit-ty-kate, line 778: for this section i think it would be also be useful to show the behaviour of opam when the package has already been installed. Like, for each fields, show the behaviour as is done currently, then force-install the package and try to install it again and show if there is a difference of behaviour

tests/reftests/run.ml
rjbou: oups i left debugging code

tests/reftests/filter-variable-resolution.test
rjbou: updated with these tests in a separate commit

rjbou (commented):
On file definition redundancy, the main idea is to have all information about the given test in its section. But there is 2 different cases:
 * patches that never changes, are always redefined, and their content it not so much important to understand the given test
 * opam files that are the same for a subset of the tests, they are more important to understand why it fails or not (how is used the variable).
I think that patches can be grouped at the beginning of the test, but opam files add more readability if they are duplicated in their own section.

kit-ty-kate:
i think this PR is the perfect way to start implementing the new testsuite schema we discussed about. We can use it to see how it works on a small scale, avoid duplication, and start the new testsuite work little by little

kit-ty-kate (commented):
lgtm otherwise

rjbou:
rebased/fixed/updated

tests/reftests/filter-variable-resolution.test
```diff
@@ -0,0 +1,1469 @@
...
+opam-version: "2.0"
+patches: [
+  "to-apply.patch" { ?smtg }
+  "to-not-apply.patch" { !?smtg }
+]
+build: [
+  [ "grep" "ploplop" "some-content" ]
+  [ "grep" "bloblob" "some-content" ]
+]
+### opam install patches -vv | sed-cmd patch | 'patch-[^"]+' -> 'patch' | sed-cmd grep | '^- GNU patch .*' -> '- GNU patch' | grep -v "Copyright\|License\|software\|WARRANTY\|Written" | unordered
```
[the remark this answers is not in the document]
rjbou: updated.

kit-ty-kate (approved):

kit-ty-kate:
Thanks!

PR #5564, opened 2023-05-26
Stop populating opam file with extra-files

rjbou:
Current behaviour is at lint stage, if an extra-files is found in `files` directory, but not in the opam file, it is added in the opam file automatically.
This behaviour is removed, all extra-files need to be declared in opam file.

/cc @hannesm @reynir 

* [x] queued on #5560
* [x] queued on #5561
* [x] queued on #6184

PR #5649, opened 2023-09-01
admin: add `update-extrafiles` command

rjbou:
This is needed with changes in #5564.
Repositories can contain opam file with no extra-files field, as opam automatically adds them. With changes in #5564, this is no more done. This command ease updating the repository to fill all extra-files fields, and permit also with the combination with the option `--hash` to choose & update used hash in a repository.

### DIFF
--- a/src/client/opamAdminCommand.ml
+++ b/src/client/opamAdminCommand.ml
@@ -331,6 +331,100 @@ let packages_with_prefixes repo_root packages =
             (List.rev missing_pkgs)));
   pkgs_map
 
+let update_extrafiles_command_doc =
+  "Add extra-files to an opam repository."
+let update_extrafiles_command cli =
+  let command = "update-extrafiles" in
+  let doc = update_extrafiles_command_doc in
+  let man = [
+    `S Manpage.s_description;
+    `P "This command scans through package definitions, and add extrafiles \
+        field as requested."
+  ]
+  in
+  let hash_kinds = [`MD5; `SHA256; `SHA512] in
+  let hash_type_arg =
+    OpamArg.mk_opt ~cli OpamArg.(cli_from cli2_2) ["hash"]
+      "HASH_ALGO" "The hash, or hashes to be added"
+      Arg.(some (enum
+                   (List.map (fun k -> OpamHash.string_of_kind k, k)
+                      hash_kinds))) None
+  in
+  let packages =
+    OpamArg.mk_opt ~cli OpamArg.(cli_from cli2_2) ["p";"packages"]
+      "PACKAGES" "Only add extra files for the given packages"
+      Arg.(list OpamArg.package) []
+  in
+  let cmd global_options hash_type packages () =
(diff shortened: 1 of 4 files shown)


src/client/opamAdminCommand.ml
```diff
@@ -331,6 +331,100 @@ let packages_with_prefixes repo_root packages =
...
+               OpamStd.Option.default [] (OpamFile.OPAM.extra_files opam)
+             in
+             let extra_files = OpamFilename.Base.Map.of_list extra_files in
+             let xfiles =
+               List.fold_left (fun extra (file, base) ->
+                   match OpamFilename.Base.Map.find_opt base extra_files with
+                   | Some hash ->
+                     let kind = OpamHash.kind hash in
+                     let xhash = compute ~kind file in
+                     if OpamHash.compare hash xhash <> 0 then
```
kit-ty-kate, line 396:
```suggestion
                     if not (OpamHash.equal hash xhash) then
```

kit-ty-kate (commented):
Aside from this minor stylistic suggestion, I’m not sure to understand the tests. Is there no test showing the reason for the command being there in the first place? i.e. "a file exists in `files/` but no `extra-files` is defined".

I’m only seeing the "a file exists in `files` but the `extra-files` section mismatches" case (which I didn’t really realized this was adding but is really welcome too!)

On that last point, given i feel like most usage of this will be for updating hashes instead of adding missing ones, maybe the command should be called `update-extrafiles` instead of `add-extrafiles`.

rjbou:
The test show several cases:
* a file exists in `files/` but no present in `extra-files:`
* a file exists in `files/` but `extra-files:` mismatch (for md5 & sha256)
* a file exists in `files/` and we want to update the hash (md5 -> sha256).

The name is taken from `add-hashes` command, but it can be changed.

AltGr (approved):

kit-ty-kate:
> a file exists in files/ but no present in extra-files

In that case i think a more explicit test case should be added to make sure adding `extra-files` when missing entirely happens:
```suggestion
### <packages/with-xf/with-xf.1/opam>
opam-version: "2.0"
### <packages/with-xf/with-xf.1/files/missing-hash>
content
### openssl md5 packages/with-xf/with-xf.1/files/missing-hash | '.*= ' -> '' >$ XF_MD5
### opam admin add-extrafiles
### opam-cat packages/with-xf/with-xf.1/opam | "${XF_MD5}" -> "good-md5"
extra-files: ["missing-hash" "md5=good-md5"]
opam-version: "2.0"
```

rjbou:
test added & name updated!

src/client/opamAdminCommand.ml
rjbou: equal is already a comparison with 0. Or you propose this change for another purpose?

kit-ty-kate (approved):

kit-ty-kate:
Thanks a lot!

Issue #5651, opened 2023-09-04
`opam lint`: a command line flag to make `opam lint` return 1 on warnings


Niols:
`opam lint` can report either errors or warnings. Currently, the return code of the command is 1 if there were errors and 0 if there weren't any (including if there were warnings). I would like a command line flag to either make warnings become errors or to make the return code be 1 also on warnings.

I “discovered” this behaviour when I realised that my CI running `opam lint` wasn't catching warnings and that I needed a fancier test involving reading the standard output of `opam lint`.

kit-ty-kate:
I've open a PR to allow marking warnings as errors in https://github.com/ocaml/opam/pull/5652

Niols:
Oh wow, very nice, thank you!

Niols:
Impressive; thank you! :tada:

PR #5652, opened 2023-09-04
opam lint: Allow to mark a set of warnings as errors

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/5651

This PR allows to mark a set of warnings as error and make the command fail.

Examples:
* `opam lint -W@0..999` enables all warnings and errors and fail if any is detected.
* `opam lint -W@62` enables W62 and fail if this warning is detected but ignore any other warnings or errors

The syntax is inspired from the `ocamlc -w` option.

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -3904,6 +3904,7 @@ let lint cli =
                 ) ([], false) warnings
             in
             let warnings =
+              if warnings_sel = [] then warnings else
               List.map (fun ((n, _, s) as warn) ->
                   match OpamStd.IntMap.find_opt n warnings_sel_map with
                   | Some `EnableError -> (n, `Error, s)
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -3890,13 +3890,10 @@ let lint cli =
                   stdin_f stdin,
                 None
             in
-            let enabled =
-              let map =
-                List.fold_left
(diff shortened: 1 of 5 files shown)


rjbou (approved):

rjbou:
I've added a test for --warn option (to complete ?) and a commit that adds the change in display (warning -> error).

src/client/opamCommands.ml
```diff
@@ -3885,21 +3890,25 @@ let lint cli =
...
             in
-            let warnings = List.filter (fun (n, _, _) -> enabled n) warnings in
-            let failed =
-              List.exists (function _,`Error,_ -> true | _ -> false) warnings
+            let warnings =
+              List.map (fun ((n, _, s) as warn) ->
+                  match OpamStd.IntMap.find_opt n warnings_sel_map with
+                  | Some `EnableError -> (n, `Error, s)
+                  | Some (`Disable | `Enable) | None -> warn)
+                warnings
```
rjbou, line 3911:
```suggestion
            let warnings =
              if warnings_sel = [] then warnings else
              List.map (fun ((n, _, s) as warn) ->
                  match OpamStd.IntMap.find_opt n warnings_sel_map with
                  | Some `EnableError -> (n, `Error, s)
                  | Some (`Disable | `Enable) | None -> warn)
                warnings
```
tiny optim

PR #5654, opened 2023-09-05
Resolve real path on `OpamSystem` temporary directory creation

rjbou:


src/core/opamSystem.ml
```diff
@@ -63,7 +86,7 @@ let rec mk_temp_dir ?(prefix="opam") () =
   if Sys.file_exists s then
     mk_temp_dir ()
   else
-    s
+    real_path s
```
kit-ty-kate, line 89: I think the only reason we have a wrapper around `OpamCompat.Unix.realpath` is because the OCaml < 4.13 version of it doesn't support files (as opposed to directories). In this case we know it is always a directory so it's safe to use that function directly.
```suggestion
    OpamCompat.Unix.realpath s
```

kit-ty-kate:
Ah... `mk_temp_dir` does not actually create a directory. I missed that fact.
```
/home/runner/local/bin/opam: "realpath" failed on /tmp/opam-4933-f80ff1: No such file or directory
```
If we want to keep that behaviour as-is then the previous version of this PR is good to me, otherwise we probably want to add a `mkdir s` in the else case and remove the mkdirs when that function is used.

rjbou:
Better revert to previous version :)

kit-ty-kate (approved):

kit-ty-kate:
Thanks! (ignoring failure in ocaml-benchmarks due to the machine being out of disk space)

PR #5657, opened 2023-09-08
reftest: update commands prefix replacement

rjbou:
Now it handles sandbox prefix, and command prefixes that are not a print of a subprocess (in an error message for ex).


tests/reftests/run.ml
```diff
@@ -421,18 +421,59 @@ module Parse = struct
...
+                ];
+                (* Unix & Mac command prefix
+                >[...] /usr/bin/cmd <
+                >[...] /usr/bin/cmd <
+                -->
+                >[...] cmd <
+                *)
+                seq [
+                  opt @@ char '"';
+                  repn @@ seq [ char '/'; repn @@ diff any (alt [ space; char '/']) ];
```
rjbou, line 452: It relies on no space on command full path

tests/reftests/run.ml
```diff
@@ -421,18 +421,59 @@ module Parse = struct
           | "|" :: "unordered" :: r ->
             get_rewr (true, acc) r
           | "|" :: "sed-cmd" :: cmd :: r ->
+            let repn r = repn r 1 None in
```
AltGr, line 424: This is `Re.rep1`

tests/reftests/run.ml
```diff
@@ -421,18 +421,59 @@ module Parse = struct
...
+              alt [
+              (* Sandbox prefix
+              >[...] /tmp/build_592d92_dune/opam-reftest-2b89f9/OPAM/opam-init/hooks/sandbox.sh "build" "cmd" <
+              >[...] ${BASEDIR}/opam-init/hooks/sandbox.sh "build" "cmd" <
+              -->
+              >[...] cmd <
+              *)
+                seq [
+                  alt [ char '/'; Re.str "${"; ];
+                  repn any; Re.str "sandbox.sh";
```
AltGr, line 435: maybe at least `nongreedy (rep1 any)` ?

tests/reftests/run.ml
```diff
@@ -421,18 +421,59 @@ module Parse = struct
...
+                  space;
```
AltGr, line 436: `rep1 space` ?

tests/reftests/run.ml
AltGr: you should probably exclude `"` and `'` as well

tests/reftests/run.ml
rjbou: true

tests/reftests/run.ml
rjbou: maybe the non greedy will permit to not have the need to exclude space

kit-ty-kate:
Thanks!

PR #5659, opened 2023-09-11
Clarify documentation for `enable` pseudo-variable

gridbugs:
The documentation for variables says that for boolean variables the form `name1+name2+name3:var` is equivalent to
`name1:var & name2:var & name3:var` but it's possible to write `name1+name2:enable` despite `enable` being the string "enable" or "disable". This change clarifies that `enable` isn't really a variable but syntactic sugar instead.

### DIFF
--- a/doc/pages/Manual.md
+++ b/doc/pages/Manual.md
@@ -512,9 +512,11 @@ the package being defined.
   resolved direct dependencies of the package
 - <a id="pkgvar-installed">`installed`</a>:
   whether the package is installed
-- <a id="pkgvar-enable">`enable`</a>:
-  takes the value "enable" or "disable" depending on whether the package is
-  installed
+- <a id="pkgvar-enable">`enable`</a>: is not a variable.
+  It takes the value "enable" or "disable" depending on whether the package is
+  installed. Used with the combination operator `+`, 
+  `name1+name2+name3:enable` is syntactic sugar for
+  `name1+name2+name3:installed?enable:disable`.
 - <a id="pkgvar-pinned">`pinned`</a>: whether the package is pinned
 - <a id="pkgvar-dirs">`bin`, `sbin`, `lib`, `man`, `doc`, `share`, `etc`</a>:
   the corresponding directories for this package (similar to
--- a/master_changes.md
+++ b/master_changes.md
@@ -149,6 +149,7 @@ users)
   * Fix a typo in the default man page [#6267 @fccm2]
   * Point users to ways to override the default opam root location in the opam init manpage [#6251 @kit-ty-kate]
   * Manual: Document the stamp field from repo files [#6306 @kit-ty-kate]
+  * Clarify documentation for `enable` pseudo-variable [#5659 @gridbugs]
 
 ## Security fixes
 


doc/pages/Manual.md
```diff
@@ -492,9 +492,11 @@ the package being defined.
...
 - <a id="pkgvar-installed">`installed`</a>:
   whether the package is installed
-- <a id="pkgvar-enable">`enable`</a>:
-  takes the value "enable" or "disable" depending on whether the package is
-  installed
+- <a id="pkgvar-enable">`enable`</a>: is not a variable.
+  `pkg1+pkg2+...+pkgN:enable` is syntactic sugar for
+  `pkg1+pkg2+...+pkgN:installed?enable:disable`. That is, it behaves like a
+  variable that takes the value "enable" or "disable" depending on whether the
+  package(s) is installed
```
kit-ty-kate, line 499:
```suggestion
  package(s) are installed
```

kit-ty-kate (approved):

doc/pages/Manual.md
```diff
@@ -492,9 +492,11 @@ the package being defined.
...
 - <a id="pkgvar-installed">`installed`</a>:
   whether the package is installed
-- <a id="pkgvar-enable">`enable`</a>:
-  takes the value "enable" or "disable" depending on whether the package is
-  installed
+- <a id="pkgvar-enable">`enable`</a>: is not a variable.
+  `pkg1+pkg2+...+pkgN:enable` is syntactic sugar for
+  `pkg1+pkg2+...+pkgN:installed?enable:disable`. That is, it behaves like a
+  variable that takes the value "enable" or "disable" depending on whether the
+  package(s) are installed
```
rjbou, line 499:
```suggestion
- <a id="pkgvar-enable">`enable`</a>:
  takes the value "enable" or "disable" depending on whether the package is
  installed. Used with combination operator `+`, 
  `name1+name2+...+nameN:enable` is syntactic sugar for
  `name1+name2+...+nameN:installed?enable:disable`.
```

rjbou:
test in #5642 updated

rjbou (changes requested):
Proposed a rewording. The combination part is a specific note on `enabled` variable usage with combination; It should be put after the definition.

doc/pages/Manual.md
kit-ty-kate:
```
- Used with combination operator `+`,
+ Used with the combination operator `+`,
```

doc/pages/Manual.md
```diff
@@ -494,7 +494,9 @@ the package being defined.
   whether the package is installed
 - <a id="pkgvar-enable">`enable`</a>:
   takes the value "enable" or "disable" depending on whether the package is
-  installed
```
gridbugs, line 497: A key point that was in the original change that's now removed here is that `enable` is not a variable. Since it's listed under the heading "Package variables", I think it needs to be clarified that it is not a variable. The original reason I opened this PR is I mistakenly tried to use it as a variable which led to confusion.

doc/pages/Manual.md
```diff
@@ -494,7 +494,9 @@ the package being defined.
   whether the package is installed
 - <a id="pkgvar-enable">`enable`</a>:
   takes the value "enable" or "disable" depending on whether the package is
```
kit-ty-kate, line 496:
```suggestion
- <a id="pkgvar-enable">`enable`</a>: is not a variable.
  It takes the value "enable" or "disable" depending on whether the package is
```

doc/pages/Manual.md
kit-ty-kate: Done. What do you think?

gridbugs: Looks good, thanks!

rjbou (approved):

PR #5663, opened 2023-09-11
GHA: Make sure opam init works in the absence of OPAMROOT

kit-ty-kate:
Adds a custom test for the issue fixed in https://github.com/ocaml/opam/pull/5662

### DIFF
--- a/.github/scripts/main/test.sh
+++ b/.github/scripts/main/test.sh
@@ -26,3 +26,9 @@ eval $(opam env)
 opam install lwt
 opam list
 opam config report
+
+# Make sure opam init (including the sandbox script) works in the absence of OPAMROOT
+unset OPAMROOT
+rm -r ~/.opam
+opam init --bare -vvv
+rm -r ~/.opam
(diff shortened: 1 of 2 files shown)


.github/scripts/main/test.sh
```diff
@@ -26,3 +26,9 @@ eval $(opam env)
 opam install lwt
 opam list
 opam config report
+
+# Make sure opam init (including the the sandbox script) works in the absence of OPAMROOT
```
rjbou, line 30:
```suggestion
# Make sure opam init (including the sandbox script) works in the absence of OPAMROOT
```

.github/scripts/main/test.sh
```diff
@@ -26,3 +26,9 @@ eval $(opam env)
...
+unset OPAMROOT
+rm -r ~/.opam
+opam init --bare
```
rjbou, line 33: Do we want to use the default repository (download archive for each job, then extracting it, etc.)? But i don't know if it is a good idea to work with a fake local repo, it will change default behaviour.

rjbou, line 33: `-vvv` to see sandbox check ? I don't know if we want to, maybe to debug in case of failure ?

kit-ty-kate: done

kit-ty-kate: i think we should use the default repository to make sure the default behaviour works

rjbou (approved):

Issue #5631, opened 2023-08-14
"opam repo add" fails if one of the packages in the repository has an unsupported version


kit-ty-kate:
```
$ tree
.
├── packages
│   └── a
│       └── a.1
│           └── opam
└── repo

4 directories, 2 files
```
```
$ cat packages/a/a.1/opam
opam-version: "3.0"
```
```
$ cat repo
opam-version: "2.0"
```
```
$ opam repo add test .
[test] Initialised
[ERROR] Could not update repository "test": In
        $OPAMROOT/.opam/repo/test/packages/a/a.1/opam:
        unsupported or missing file format version; should be 2.0 or older
[ERROR] Initial repository fetch failed
```

The expected behaviour would be do silently mark the package as unavailable and only fail when the `repo` file’s version is unsupported.

PR #5665, opened 2023-09-12
Silently mark packages requiring an unsupported version of opam as unavailable

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/5631

This makes it possible to have a smoother upgrade path for the opam repository when opam adds a new syntax for example.

In the future it would be nice to have a specific error message to tell the user why this package is not available, but I feel like we can do this later (TODO: open a ticket for it if this is merged first)

### DIFF
--- a/src/state/opamFileTools.ml
+++ b/src/state/opamFileTools.ml
@@ -1176,7 +1176,7 @@ let lint_gen ?check_extra_files ?check_upstream ?(handle_dirname=false)
       [0, `Error, "File does not exist"], None
     | OpamLexer.Error _ | Parsing.Parse_error ->
       [1, `Error, "File does not parse"], None
-    | OpamPp.Bad_version bf | OpamPp.Bad_format bf -> [warn_of_bad_format bf], None
+    | OpamPp.Bad_version (bf, _) | OpamPp.Bad_format bf -> [warn_of_bad_format bf], None
     | OpamPp.Bad_format_list bfl -> List.map warn_of_bad_format bfl, None
   in
   let check_extra_files = match check_extra_files with
@@ -1409,6 +1409,23 @@ let read_opam dir =
       (OpamPp.string_of_bad_format (OpamPp.Bad_format (snd err)));
     None
   | None, None -> None
+  | exception OpamPp.Bad_version ((_, _errmsg), Some version) ->
+    let sversion = OpamVersion.to_string version in
+    let scurrent = OpamVersion.to_string OpamVersion.current_nopatch in
+    log "opam-version %S unsupported on %s. Added as dummy unavailable package."
+      sversion (OpamFile.to_string opam_file);
+    Some
+      (OpamFile.OPAM.empty
+       |> OpamFile.OPAM.with_available
+         (FOp (FIdent ([], OpamVariable.of_string "opam-version", None),
+               `Geq, FString sversion))
+       |> OpamFile.OPAM.with_descr_body
+         (Printf.sprintf
+            "This package uses opam %s file format which opam %s cannot \
+             read.\n\n\
+             In order to install or view information on this package, please \
+             upgrade your opam installation to at least version %s."
+            sversion scurrent sversion))
 
 let read_repo_opam ~repo_name ~repo_root dir =
   let open OpamStd.Option.Op in
--- a/tests/reftests/repository.test
+++ b/tests/reftests/repository.test
@@ -637,12 +637,10 @@ some-field-that-do-not-exist: true
 <><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
 [repo] no changes from file://${BASEDIR}/REPO
 [versions] synchronised from file://${BASEDIR}/VRS
-[ERROR] In ${BASEDIR}/OPAM/repo/versions/packages/bad-opam-version/bad-opam-version.1/opam:
-        unsupported or missing file format version; should be 2.0 or older
-[ERROR] Strict mode: aborting
-[ERROR] Could not update repository "versions": OpamStd.OpamSys.Exit(30)
-# Return code 40 #
+opam-file                       opam-version "5.0" unsupported on ${BASEDIR}/OPAM/repo/versions/packages/bad-opam-version/bad-opam-version.1/opam. Added as dummy unavailable package.
+Now run 'opam upgrade' to apply any package updates.
 ### opam list -A --all-versions -s
+bad-opam-version.1
 foo.1
 foo.2
 foo.3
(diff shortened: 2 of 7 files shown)


src/state/opamFileTools.ml
```diff
@@ -1192,6 +1192,10 @@ let read_opam dir =
       (OpamPp.string_of_bad_format (OpamPp.Bad_format (snd err)));
     None
   | None, None -> None
+  | exception (OpamPp.Bad_version _bf) ->
+    log "opam-version unsupported on %s. Add as dummy unavailable package."
```
dra27, line 1196: `snd _bf` I think is the version, which is worth logging, so I think
```suggestion
  | exception (OpamPp.Bad_version (_, version) ->
    log "opam-version %S unsupported on %s. Added as dummy unavailable package." version
```
(s/Add/Added to describe what it has done, rather than what opam might be expecting the user to do)

src/state/opamFileTools.ml
```diff
@@ -1192,6 +1192,10 @@ let read_opam dir =
...
+      (OpamFile.to_string opam_file);
+    Some (OpamFile.OPAM.with_available (FBool false) OpamFile.OPAM.empty)
```
dra27, line 1198: I would tweak this slightly: we have the version above, so given:

```opam
opam-version: "3.0"
some-field-that-does-not-exist: true
```
we effectively translate it to:
```opam
opam-version: "2.0"
available: opam-version >= "3.0"
synopsis: "Requires opam 3.0"
description: """
This package uses opam 3.0 file format which opam 2.2 cannot read.

In order to install or view information on this package, please upgrade
your opam installation to at least version 3.0."""
```
This has the benefit that if the user attempts to install it, they get a better error message about the lack of availability, and if they do `opam show` or `opam list`, then they get some description about why the package is missing.

dra27 (commented):
I think this PR is an excellent idea - it concludes groundwork I laid in https://github.com/ocaml/opam-file-format/pull/44 in opam 2.1.

src/format/opamFormat.ml
```diff
@@ -912,6 +912,23 @@ module I = struct
...
+        OpamPp.t
+      = (check ~name ~errmsg f)
+    in
+    let _ :
+      (opamfile_item_kind with_pos list,
+       OpamVersion.t option * opamfile_item_kind with_pos list)
+        OpamPp.t
+      =   field name (parse opam_v) -|
+          map_fst (check ~name ~errmsg f)
+    in
```
rjbou, line 931:
```suggestion
```

rjbou:
Updated the PR

kit-ty-kate:
rebased

src/state/opamFileTools.ml
```diff
@@ -1328,6 +1328,35 @@ let read_opam dir =
...
+        "Unexpected failure: failed to parse version %S \
+         but is below the current version %S"
+        sversion (OpamVersion.to_string OpamVersion.current);
+      Some
+        (OpamFile.OPAM.empty
+         |> OpamFile.OPAM.with_available (FBool false)
+         |> OpamFile.OPAM.with_descr_body
+           (Printf.sprintf
+              "Failed to parse version %S but is below the current version %S"
+              sversion (OpamVersion.to_string OpamVersion.current)))
```
rjbou, line 1346: how can we reach that branch ?

kit-ty-kate: i don't know but i think it's worth the special case if it ever happens. As the log shows it is an "unexpected failure". Although if you prefer we can replace this by an assert

rjbou: I'm fine with the current code, it's just that i tried to raise that unexpected failure and i wasn't able to :)

kit-ty-kate: I've removed that section. Actually we do similar things in both cases and the description of the package added in the normal case does encompass this first one so it's not needed

rjbou: In fact, we should keep the version check, and if it is an unexpected error fail with internal error, we shouldn't act as it is known normal error.
From what i saw, there is no path via opam binary that lead to that path, but via lib, we can't be sure.

kit-ty-kate: I don't think it needs it. In both cases it's an unsupported version so i think we should have the same behaviour in both cases.

src/state/opamFileTools.ml
```diff
@@ -1328,6 +1328,23 @@ let read_opam dir =
...
+      (OpamFile.OPAM.empty
+       |> OpamFile.OPAM.with_available
+         (FOp (FIdent ([], OpamVariable.of_string "opam-version", None),
+               `Geq, FString sversion))
+       |> OpamFile.OPAM.with_descr_body
+         (Printf.sprintf
+            "This package uses opam %s file format which opam %s cannot \
+             read.\n\n\
+             In order to install or view information on this package, please \
+             upgrade your opam installation to at least version %s."
```
rjbou, line 1346: We should also redirect to a FAQ link, where we explain the issue and how to resolve it. It is easier to update the FAQ section with new "errors" (the futures ones that we try to handle _a priori_) than update the message in the package definition.
```suggestion
             upgrade your opam installation to at least version %s.\
             See https://opam.ocaml.org/doc/FAQ.html#Why-opam-tells-me-I-have-a-dummy-package."
```

kit-ty-kate: I'm not sure we want to do this here. This message will only appear in the description of the package and i think the rest of the message is descriptive enough.

However, future improvement could be made elsewhere in the code (somewhere more user-facing) as mentioned in the PR description:
> In the future it would be nice to have a specific error message to tell the user why this package is not available, but I feel like we can do this later (TODO: open a ticket for it if this is merged first)

tests/reftests/repository.test
```diff
@@ -607,3 +607,185 @@ opam-version: "2.0"
...
+opam-version: "2.0"
+name: "bad-opam-version"
+version: "2"
+synopsis: ""
+description: """\
+This package uses opam 5.0 file format which opam CURRENTMAJOR cannot read.
+
+In order to install or view information on this package, please upgrade your opam installation to at least version 5.0."""
+available: opam-version >= "5.0"
+### :: Parse errors
```
rjbou, line 725: @dra27 to look at those results

tests/reftests/repository.test
```diff
@@ -607,3 +607,185 @@ opam-version: "2.0"
...
+[WARNING] Could not read file ${BASEDIR}/OPAM/repo/versions/packages/two-one/two-one.1/opam. skipping:
+          At ${BASEDIR}/OPAM/repo/versions/packages/two-one/two-one.1/opam:3:0-3:0::
+          Parse error
+Now run 'opam upgrade' to apply any package updates.
+### opam show two-one --raw
+[ERROR] No package matching two-one found
+# Return code 5 #
+### rm -r VRS/packages
+### <VRS/packages/two-two/two-two.1/opam>
+opam-version: "2.2"
```
dra27, line 756: This - and the 2.3 one below - are the ones producing the wrong output (which is tracked in #6188 and which a PR for that will therefore fix)

src/state/opamFileTools.ml
rjbou: Discussed, no need to add it in FAQ

rjbou:
Once rebased on top of 6199, good to go!

dra27 (approved):
LGTM!

PR #5680, opened 2023-09-20
Add ./configure --enable-static to compile the opam binary statically

### LINKED ISSUE #5647: Create a static binary on Windows
With both MSVC and MinGW currently opam would fail to execute a binary compiled on another machine without the appropriate dlls pre-installed or shipped together. For MinGW it currently even wouldn't finish compiling without those DLLs (e.g. libstdc++) available somewhere in the PATH or if opam is compiled with `CONFIGURE_ARGS=--with-private-runtime`

Create a proper Windows Installer executable seems to be typical for Windows programs but to me it looks much simpler to simply include the needed dlls statically into the executable be default like we do for releases on Linux with musl.

Related to https://github.com/ocaml/opam/issues/5628

kit-ty-kate:
Fixes #5647


rjbou (approved):
thanks!

Issue #5331, opened 2022-10-27
[Feature request] opam install --verbose-on=<pkg1>,... 


kit-ty-kate:
Currently `--verbose` shows all package commands for a given request. However I feel like this is a bit too verbose for most use-case.
It would be nice to have a command line equivalent of setting `flags: verbose` for a given package (usually the ones requested) that would only show the package commands for the given set of packages instead of all.

rjbou:
Closing as merged

PR #5682, opened 2023-09-20
Add a verbose-on option.

desumn:
Implementation of #5331 

Add a `verbose-on=PACKAGES` parameters to build options. For the time being, it's equivalent to a simple `-v` for a list of package.

I'm not sure how I'd be able to simulate a modification of the opam file of a specific package to change add the verbose flag, so it's not exactly what #5331 asked for, however if there's a way I'd be glad to implement it!

### DIFF
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -304,6 +304,8 @@ let environment_variables =
       "see option `--skip-updates'.";
       "STATS", cli_original, (fun v -> STATS (env_bool v)),
       "display stats at the end of command.";
+      "VERBOSEON", cli_from cli2_2, (fun v -> VERBOSEON (env_string_list v)),
+      "see option --verbose-on";
       "WORKINGDIR", cli_original, (fun v -> WORKINGDIR (env_bool v)),
       "see option `--working-dir'.";
     ] in
@@ -647,19 +649,20 @@ type build_options = {
   lock_suffix   : string;
   assume_depexts: bool;
   no_depexts    : bool;
+  verbose_on    : name list option;
 }
 
 let create_build_options
     keep_build_dir reuse_build_dir inplace_build make no_checksums
     req_checksums build_test build_doc dev_setup show dryrun skip_update
     fake jobs ignore_constraints_on unlock_base locked lock_suffix
-    assume_depexts no_depexts
+    assume_depexts no_depexts verbose_on
     =
   {
     keep_build_dir; reuse_build_dir; inplace_build; make; no_checksums;
     req_checksums; build_test; build_doc; dev_setup; show; dryrun; skip_update;
     fake; jobs; ignore_constraints_on; unlock_base; locked; lock_suffix;
-    assume_depexts; no_depexts;
+    assume_depexts; no_depexts; verbose_on;
   }
 
 let apply_build_options cli b =
@@ -697,6 +700,10 @@ let apply_build_options cli b =
     ?fake:(flag b.fake)
     ?skip_dev_update:(flag b.skip_update)
     ?assume_depexts:(flag (b.assume_depexts || b.no_depexts))
+    ?verbose_on:
+      (b.verbose_on >>= function
+        | [] -> None
+        | vo -> Some (OpamPackage.Name.Set.of_list vo))
     ~scrubbed_environment_variables
     ()
 
@@ -1525,12 +1532,18 @@ let build_options cli =
        you installed the required dependency by hand. Implies \
        $(b,--assume-depexts), and stores the exceptions upon success as well."
   in
+  let verbose_on =
+    mk_opt ~cli (cli_from cli2_3) ~section ["verbose-on"] "PACKAGES"
+      "Be more verbose on specific packages.\
+       This is equivalent to setting $(b,\\$OPAMVERBOSE)."
+      Arg.(some (list package_name)) None
+  in
   Term.(const create_build_options
         $keep_build_dir $reuse_build_dir $inplace_build $make
         $no_checksums $req_checksums $build_test $build_doc $dev_setup $show
         $dryrun $skip_update $fake $jobs_flag ~section cli cli_original
         $ignore_constraints_on $unlock_base $locked $lock_suffix
-        $assume_depexts $no_depexts)
+        $assume_depexts $no_depexts $verbose_on)
 
 (* Option common to install commands *)
 let assume_built ?section cli =
--- a/src/core/opamStd.ml
+++ b/src/core/opamStd.ml
@@ -1854,6 +1854,9 @@ module Config = struct
   let env_string var =
     env (fun s -> s) var
 
+  let env_string_list var =
+    env (fun s -> OpamString.split s ',') var
+
   let env_float var =
     env float_of_string var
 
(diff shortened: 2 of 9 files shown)


src/state/opamStateConfig.ml
```diff
@@ -195,7 +201,11 @@ let initk k =
     ?no_env_notice:(E.noenvnotice ())
     ?locked:(E.locked () >>| function "" -> None | s -> Some s)
     ?no_depexts:(E.nodepexts ())
-
+    ?verbose_on:
+     (E.verbose () >>| fun s ->
+      OpamStd.String.split s ',' |>
+      List.map OpamPackage.Name.of_string |>
+      OpamPackage.Name.Set.of_list)
```
rjbou, line 208: It is better to add an environment variable converter in `OpamStd.Config` than splitting the string here. You can see how it is done for `OPAMDEBUGSECTIONS` with `OpamStd.Config.env_sections`, or more simple float variables with `OpamStd.Config.env_float`.
You can add a function that handles string with `,` separator, that converts it to a string list ; and convert that string list into names set here.

rjbou (commented):
Thanks a lot for this first PR!
I've put some comments, but the general idea on how you add it is good!
Please tell me if you need more precision!

rjbou:
I've added a test (`tests/reftests/verbose-on.test`), like that you can it is possible to test your feature, and we'll keep it as a new test in the testsuite, to be sure to detect regression in PRs.

I've also rebased your PR, we prefer rebasing on master than merging master for PRs, it eases history/commits lookup.

rjbou:
No worry! We'll try to review the changes in the next weeks. We are also focused on the 2.2 release on our side :)

desumn:
It's been some time, do you have any update on this?

src/client/opamArg.ml
```diff
@@ -304,6 +304,8 @@ let environment_variables =
       "see option `--skip-updates'.";
       "STATS", cli_original, (fun v -> STATS (env_bool v)),
       "display stats at the end of command.";
+      "VERBOSE_ON", cli_from cli2_2, (fun v -> VERBOSEON (env_string_list v)),
```
rjbou, line 307: Opam environment variable are all with a format without `_`
```suggestion
      "VERBOSEON", cli_from cli2_2, (fun v -> VERBOSEON (env_string_list v)),
```

src/client/opamArg.ml
```diff
@@ -1505,12 +1512,17 @@ let build_options cli =
        you installed the required dependency by hand. Implies \
        $(b,--assume-depexts), and stores the exceptions upon success as well."
   in
+  let verbose_on =
+    mk_opt ~cli (cli_from cli2_2) ~section ["verbose-on"] "PACKAGES"
+      "Be more verbose on specific packages."
```
rjbou, line 1517:
```suggestion
      "Be more verbose on specific packages.\
       This is equivalent to setting $(b,\\$OPAMVERBOSE)."
```

src/core/opamStd.ml
```diff
@@ -1678,6 +1678,8 @@ module Config = struct
 
   let env_int var = env int_of_string var
 
+  let env_string_list var = env (fun s -> OpamString.split s ';') var
```
rjbou, line 1681: In CLI, elements are separated with `,`. Any thoughts @dra27 @kit-ty-kate ?
```suggestion
  let env_string_list var = env (fun s -> OpamString.split s ',') var
```

rjbou:
Sorry for the delay, we are still diving into the release of 2.2 (that is in its beta2 since 2 days). I've put some comments, with these modification and an update of master_changes (points to add in Client, Reftests, and API), the PR should be nearly complete.

desumn:
That's fine, thanks for your reviews, commited the changes.

src/client/opamArg.ml
```diff
@@ -1505,12 +1510,17 @@ let build_options cli =
...
+      Arg.(some (list package)) None ~vopt:(Some [])
```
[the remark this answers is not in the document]
kit-ty-kate: I think it would be better to allow full package name + version as it is currently. This way the feature would be much simpler to use, especially in CI systems where this would be most useful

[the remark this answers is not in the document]
rjbou: If we have name.version, it can be used only if user knows in advance which version of the package will be installed, which is not always the case.
In usual CI, you may want to have install log of the package you are testing (usually yours), and its inner version can change.
What do you have in mind about disadvantages of only name?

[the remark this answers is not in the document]
kit-ty-kate: ok fair, let's do name only for now, we can always add name.version later if we want or if requested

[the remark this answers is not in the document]
rjbou: It could evolve into name.optional-version, like that we have all advantages :)

src/client/opamArg.ml
```diff
@@ -627,19 +629,20 @@ type build_options = {
   lock_suffix   : string;
   assume_depexts: bool;
   no_depexts    : bool;
+  verbose_on : (name * version option) list option
```
rjbou, line 632:
```suggestion
  verbose_on : name list;
```

src/client/opamArg.ml
```diff
@@ -1505,12 +1512,18 @@ let build_options cli =
        you installed the required dependency by hand. Implies \
        $(b,--assume-depexts), and stores the exceptions upon success as well."
   in
+  let verbose_on =
+    mk_opt ~cli (cli_from cli2_2) ~section ["verbose-on"] "PACKAGES"
+      "Be more verbose on specific packages.\
+       This is equivalent to setting $(b,\\$OPAMVERBOSE)."
+      Arg.(some (list package)) None ~vopt:(Some [])
```
rjbou, line 1519:
```suggestion
      Arg.(list package_name) []
```

rjbou:
Now opam 2.2 is out, we can take a look again to other PRs @desumn. 

I've updated the PR, squashed the intermediate commits, cleaned it.

src/client/opamArg.ml
```diff
@@ -697,6 +700,10 @@ let apply_build_options cli b =
     ?fake:(flag b.fake)
     ?skip_dev_update:(flag b.skip_update)
     ?assume_depexts:(flag (b.assume_depexts || b.no_depexts))
+    ?verbose_on:
+      (b.verbose_on >>= function
+        | [] -> None
+        | vo -> Some (OpamPackage.Name.Set.of_list vo))
```
rjbou, line 706: Current behaviour: is flag list is empty, it is considered as nothing was given. Same for environment variable.

tests/reftests/verbose-on.test
```diff
@@ -0,0 +1,312 @@
...
+- bar
+-> compiled  bar.1
+-> installed bar.1
++ echo "foo" (CWD=${BASEDIR}/OPAM/verbose-on/.opam-switch/build/foo.1)
+- foo
+-> compiled  foo.1
+-> installed foo.1
+Done.
+### : unknown package
+### opam reinstall qux --verbose-on unknown-pkg | unordered
```
rjbou, line 131: On unknown package, nothing is printed, should we?

kit-ty-kate: I think the current behaviour is fine

kit-ty-kate (approved):
I've remove a couple of local opens (which in my opinion we should avoid for readability and maintainability purpose). lgtm now

rjbou (approved):

PR #5686, opened 2023-09-25
OPAMVERBOSE: Fix `level` by flipping the `env_level` implementation

smorimoto:
This literally flips the implementation. As you can see in the implementation of `bool_of_string`, the original implementation does not seem correct.

https://github.com/ocaml/opam/blob/b0cb137edebc5e7d7ac1e650086e3be5800ae743/src/core/opamStd.ml#L1666-L1670

In fact, I found this when debugging a GitHub Actions issue, and you can easily reproduce the issue on your computer by just try each of the following commands:

```sh
OPAMVERBOSE=no opam update
```

```sh
OPAMVERBOSE=yes opam update
```

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -14,6 +14,7 @@ users)
   * Bump to 2.2.0~alpha3~dev [#5615 @rjbou]
 
 ## Global CLI
+  * Fix `OPAMVERBOSE` setting, 0 and 1 levels was inverted: eg, "no" gives level 1, and "yes" level 0 [#5686 @smorimoto]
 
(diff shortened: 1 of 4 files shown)


smorimoto:
I was not sure where to put this PR in the changelog.

rjbou:
Well spotted! I've updated the PR with some tests & filled changes. Thanks!

rjbou (approved):

rjbou:
Thanks!

smorimoto:
Likewise!

PR #5690, opened 2023-09-29
Proposal: allow adding npm package as depexts

haochenx:

### DIFF
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -377,17 +377,103 @@ let yum_cmd = lazy begin
     raise (OpamSystem.Command_not_found "yum or dnf")
 end
 
+let npm_cmd = lazy begin
+  if OpamSystem.resolve_command "npm" <> None then
+    ("npm", ["install"; "-g"], ["update"; "-g"])
+  else if OpamSystem.resolve_command "yarn" <> None then
+    ("yarn", ["global"; "add"], ["global"; "upgrade"])
+  else
+    raise (OpamSystem.Command_not_found "npm or yarn")
+end
+
+type npm_package_coord = {
+    scope : string option;
+    name : string;
+  }
+
+type 'coord package_spec' =
+  | System_package : string -> string package_spec'
+  | Npm_package : npm_package_coord -> npm_package_coord package_spec'
+
+type package_spec = Package : _ package_spec' -> package_spec
+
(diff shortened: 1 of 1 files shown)


rjbou:
Thanks for opening a PR. Usually, it is better to open an issue to open the discussion, like that we can guide you on what is possible and what isn't.

I understand the issue, but for the moment opam depext mechanism relies only on system package managers. There is a proposal to let the user choose its own package manager, and have the depext mechanism based on that instead of the distribution package manager. Once that done, it'll be possible to have filters on package manager, and integrate then `npm` support.
In the meantime, i'm afraid that there is no solution to install automatically npm packages.

As [mentioned](https://github.com/ocaml/opam-repository/pull/24492#issuecomment-1729864267) in opam repo PR, it is possible to add a `conf-bun` conf packages, that has a failing post-message *Please install bun using that command: xxx*.

haochenx:
>Usually, it is better to open an issue to open the discussion, like that we can guide you on what is possible and what isn't.

Ouch, I was creating a parallel issue after this PR, but apparently, the issue didn't get submitted (presumably due to a bad network connection as I was, and am still on travel.) Should I still open the issue or continue the discussion here (I still have the text cached in my browser so it's a matter of clicking the submission button)?

>I understand the issue, but for the moment opam depext mechanism relies only on system package managers. 

I've figured out that after reading the manual and looking at the source code.

>In the meantime, i'm afraid that there is no solution to install automatically npm packages.

Understood.

>There is a proposal to let the user choose its own package manager, and have the depext mechanism based on that instead of the distribution package manager. Once that done, it'll be possible to have filters on package manager, and integrate then npm support.

Do you have a reference to it? I did a preliminary search on the issues & prs of https://github.com/ocaml/opam but did not find it. I am happy to help if it may accelerate the process.

>As mentioned in opam repo PR, it is possible to add a conf-bun conf packages, that has a failing post-message Please install bun using that command: xxx.

Sure thing. I was at a conference last week and was planning to submit a `conf-bunjs` package doing exactly that after going back to my office (towards where I'm now traveling.)

haochenx:
As the parallel issue I have drafted is wider in scope, I decided to submit it anyway: https://github.com/ocaml/opam/issues/5692

haochenx:
PR for `conf-bunjs` created: https://github.com/ocaml/opam-repository/pull/24557

PR #5691, opened 2023-09-29
Directory tracking: in presence of .install file, track also intermediate directories

### LINKED ISSUE #5688: Opam does not remove some directories during uninstallation
In Frama-C we have noticed a surprising change between Opam 2.0 and 2.1 (yes, we notice it only now), on the uninstallation of `META` files and their directories. Currently I have only been able to produce it for Frama-C plug-ins.

With a small Frama-C plug-in like [empty.zip](https://github.com/ocaml/opam/files/12727067/empty.zip) (requires Frama-C installed).

With commands:
```
opam install .
opam uninstall --debug frama-c-empty
```

We can see in the following difference in the uninstallation log:

Opam 2.0.8
```
...  ACTION                 Removing the misc files
...  SYSTEM                 rm (switch)/.opam-switch/install/frama-c-empty.install
...  SYSTEM                 rm (switch)/lib/frama-c/plugins/empty/META
...  SYSTEM                 rmdir (switch)/lib/frama-c/plugins/empty
...  SYSTEM                 rm (switch)/.opam-switch/install/frama-c-empty.changes
```

Opam 2.1.5
```
...  ACTION                 Removing the misc files
...  SYSTEM                 rm (switch)/.opam-switch/install/frama-c-empty.install
...  SYSTEM                 rm (switch)/lib/frama-c/plugins/empty/META
...  SYSTEM                 rm (switch)/.opam-switch/install/frama-c-empty.changes
```

This can cause trouble with `dune-site` when loading plug-ins.

Is it a bug or is it intended (if it is intended, I guess we'll have to fix dune site) ?

rjbou:
Since #4494, directory tracking is simplified when a .install file is present: only cited files/directories are tracked and marked as changed if necessary.
Intermediate directories were not tracked, so directories are created but never removed.

fix #5688

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -114,7 +114,7 @@ users)
   * Add download test, to check `OPAMCURL/OPAMFETCH` handling [#5607 @rjbou]
   * Add `core/opamSystem.ml` specific tests, to test command resolution [#5600 @rjbou]
   * Add test for `OpamCoreConfig`, to check `OPAMVERBOSE` values [#5686 @rjbou]
-  * dot-install: generalise inner script [#5691 @rjbou]
+  * dot-install: generalise inner script & use less generic filenames [#5691 @rjbou]
 
 ### Engine
   * With real path resolved for all opam temp dir, remove `/private` from mac temp dir regexp [#5654 @rjbou]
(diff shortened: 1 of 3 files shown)


kit-ty-kate (approved):

kit-ty-kate:
Thanks!

PR #5708, opened 2023-10-23
Fix formatting in the Manual

kit-ty-kate:
The line with the note would make the next item of the list seem disconnected from the previous list
Screenshot 2023-10-23 at 12 12 58


rjbou (approved):
thanks!

Issue #5644, opened 2023-08-24
[windows] opam hints to use `eval` after install


Alizter:
I was using opam on windows with dkml and noticed that `opam switch` would correctly write the windows version of `eval $(opam env)` however the hint you get after installing a package still stayed the unix version.

rjbou:
Which version of opam do you have ?
Here is the output in cmd with opam 2.2.0~alpha2
![Screenshot at 2023-08-28 16-56-29](https://github.com/ocaml/opam/assets/37832926/a9b481ae-2974-4ca8-a607-685493656f38)

Alizter:
Maybe this was fixed recently:
![image](https://github.com/ocaml/opam/assets/8614547/23b8696e-5003-40a8-80e0-2b943264be40)

kit-ty-kate:
Which terminal are you using? `cmd.exe` or something else?

Alizter:
This is windows terminal which I believe runs PowerShell.

kit-ty-kate:
I just tried locally and there is definitely something funky going on with Windows Terminal.
On my machine running opam.exe will raise an `Invalid_argument("Filename.chop_prefix")` coming from https://github.com/ocaml/opam/blob/8a23fc031b80abcaaf0e8c96059d02eea60d3fc6/src/core/opamStd.ml#L1091

Looking at the values returned by `OPAMW_GetProcessAncestry` I get, in order:
```
opam.exe
windowsterminal.exe
sihost.exe
.
.
.
```
the obvious cause are the three last elements of the list `"."` but I'm not versed in Windows API enough to understand what's going on.

cc @jonahbeckford @dra27 do you have any idea?

Alizter:
In the windows terminal, you do not need to give the `.exe` extension. I'm not sure if that makes a difference.

kit-ty-kate:
> In the windows terminal, you do not need to give the `.exe` extension. I'm not sure if that makes a difference. This is a feature of PowerShell.

i tried it already but it does not make a difference. The issue is the last three `"."` elements in my case. I think in yours it somehow succeeded but didn't manage to detect Powershell either and defaulted on `sh`. I'm not sure how it did that though. Do you have the `SHELL` environment variable set by any chance?

Alizter:
Well I am running all this with `with-dkml` so that probably makes a difference?

jonahbeckford:
Sounds like two issues.

[...]

PR #5714, opened 2023-11-02
Harden `windows_get_shell`

dra27:
In Windows native containers, the `Filename.chop_suffix` call in `windows_get_shell` was occasionally failing. `OpamStubs.getProcessAncestry` can return empty strings. `windows_get_shell` now explictly handles the empty string case and `.exe` is now _optionally_ stripped.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -105,7 +105,7 @@ users)
 ## Internal
 
 ## Internal: Windows
-  * Fix sporadic crash in shell detection (seen in native containers) [#5714 @dra27]
+  * Fix sporadic crash and segfault in shell detection (seen in native containers) [#5714 @dra27]
 
 ## Test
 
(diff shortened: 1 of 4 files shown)


rjbou (approved):

rjbou:
related to https://github.com/ocaml/opam/issues/5644#issuecomment-1697443468

master_changes.md
```diff
@@ -105,6 +105,7 @@ users)
 ## Internal
 
 ## Internal: Windows
+  * Fix sporadic crash in shell detection (seen in native containers) [#5714 @dra27]
```
rjbou, line 108: note for dev: Missing api changelog

dra27: Done

dra27:
I pushed two commits fixing errors in the C stub for determining the shell, one of which was heinous (the pointer into the array would end up pointing at the _old_ array following `realloc`)

kit-ty-kate:
Thanks!

Issue #5617, opened 2023-07-25
opam 2.2+ and Git on Windows


jonahbeckford:
(Follow-up issue from opam-dev conversation)

### Context

Git has:
- a Credentials Manager (used for private repositories ... ie. GitHub forks)
- a set of trusted certificate authorities
- Git configuration

### Implications

- We can't expose `git.exe` from Cygwin (or MSYS2) to a Windows user. Cygwin git has its own Credentials Manager (none of the Windows credentials will work), a different set of trusted certificate authorities (rarely important, except in a managed or corporate environment), and a different Git configuration (the different `core.autocrlf` configuration settings are quite important for interoperability, etc.).
- `git` is required for opam ... early in its initialization

Two suggestions given:
1. Ask user to install Git for Windows before installing opam
2. Have opam use Cygwin `git` for opam initialization, but after that use `git` from the system (ie. Git for Windows)

### Context

The canonical Git on Windows is:
- Git for Windows

However, many Windows power users (ie. average Windows developers) use Chocolatey and Scoop to install their packages.

The real problem is that adding `git.exe` to the PATH is often dangerous.

Here is Git for Windows which is safe *except* if user said "Add Git Bash to PATH" during installation.

```batch
REM Having this in the PATH is safe!

C:> dir "C:\Program Files\Git\cmd\"


    Directory: C:\Program Files\Git\cmd


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a----         10/7/2022  12:51 PM         136752 git-gui.exe
-a----         10/7/2022  12:51 PM          45616 git.exe
-a----         10/7/2022  12:51 PM         136752 gitk.exe
-a----         10/7/2022  12:51 PM          45616 scalar.exe
-a----         10/7/2022  12:51 PM           3022 start-ssh-agent.cmd
-a----         10/7/2022  12:51 PM           2723 start-ssh-pageant.cmd
```

```batch
REM Having Git Bash in the PATH is dangerous because the bash.exe will conflict (sometimes silently) with Cygwin bash!

C:> dir "C:\Program Files\Git\bin"


    Directory: C:\Program Files\Git\bin


Mode                 LastWriteTime         Length Name
----                 -------------         ------ ----
-a----         10/7/2022  12:51 PM          46128 bash.exe
-a----         10/7/2022  12:51 PM          45616 git.exe
-a----         10/7/2022  12:51 PM          46128 sh.exe
```

### Implications

[...]

PR #5718, opened 2023-11-03
Git on windows: check and advertise to use Git for Windows

rjbou:
-- *outdated, to be updated* --

At initialisation, check the origin of resolved git command:
* if from Cygwin, ask to install Git for Windows one
* if from Git for Windows, check that it is the good directory in path (`\cmd`)
This is done by checking the existence of a bash executable near git.

+ no more install git in Cygwin initialisation if already present in path

related to #5617 

/cc @jonahbeckford 

todo
* [x] update changes

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -636,55 +636,62 @@ let init_checks ?(hard_fail_exn=true) init_config =
   else not (soft_fail || hard_fail)
 
 let git_for_windows_check =
-  if Sys.win32 then
-    fun cygbin ->
-      let gitcmd = OpamSystem.resolve_command "git" in
-      match gitcmd with
-      | None -> (* git *)
-        OpamConsole.note "no git!"
-      | Some gitcmd ->
-        let bindir = Filename.dirname gitcmd in
-        let bashcmd = Filename.concat bindir "bash.exe" in
-        (* We don't want to resolve it, just to check if it exists and is exe *)
-        let check_bash = OpamSystem.resolve_command ~env:[||] bashcmd in
-        match check_bash with
-        | None -> ()
-        | Some bashcmd -> (* found bash in git prefix *)
-          let is_cygwin =
-            OpamStd.Option.map_default (fun prefix ->
-                OpamStd.String.starts_with ~prefix bashcmd)
-              false cygbin
-          in
-          let git4cmd_bindir =
-            Filename.dirname bindir
-          in
-          let git4cmd_cmddir =
-            Filename.concat git4cmd_bindir "cmd"
-          in
-          let is_git4win =
-            let file = Filename.concat git4cmd_cmddir "git.exe" in
-            Sys.file_exists file
-          in
-          if is_cygwin then
-            OpamConsole.warning
-              "You are using Cygwin git, \
-               consider installing Git for Windows: https://gitforwindows.org \
-               and check that it is well in your Path"
-          else
-          if is_git4win then
-            OpamConsole.warning
-              "You are using Git for Windows, \
-               but your path contains the wrong binary directory.\n\
-               Consider reinstalling or add %s in your path instead of %s"
-              git4cmd_cmddir bindir
-          else
-            OpamConsole.warning
-              "A path in you Path contains Unix tools that doesn't come \
-               from Cygwin or Git for Windows: %s.\n\
-               Consider installing Git for Windows: https://gitforwindows.org \
-               and cleaning your Path"
-              bindir
-  else fun _ -> ()
+  if not Sys.win32 && not Sys.cygwin then fun _ -> None else
+  fun cygbin ->
+    let contains_git p =
+      OpamSystem.resolve_command ~env:[||] (Filename.concat p "git.exe")
+    in
+    let gits =
+      OpamStd.Env.get "PATH"
+      |> OpamStd.Sys.split_path_variable
+      |> OpamStd.List.filter_map (fun p ->
+          match contains_git p with
+          | Some git ->
+            Some (git, OpamSystem.bin_contains_bash p)
+          | None -> None)
+    in
+    (if gits = [] then
+       OpamConsole.warning
+         "No preinstalled git software found, we recommend using Git for \
+          Windows https://gitforwindows.org or winget. You can install it \
+          and relaunch opam initialisation. Otherwise, opam will install \
+          directly git from Cygwin."
+     else
+     let is_in_cygwin p =
+       match cygbin with
+       | None -> false
+       | Some cygbin -> OpamStd.String.starts_with ~prefix:cygbin p
+     in
+     OpamConsole.msg "Found those git binaries in path:\n%s\n"
+       (OpamStd.Format.itemize (fun (git, has_bash) ->
+            Printf.sprintf "%s%s"
+              git
+              (if has_bash
+               && not (is_in_cygwin git) then
+                 " but contains bash in it, which is dangerous" else ""))
+           gits));
+    let rec get_gitbin () =
+      match OpamConsole.read "Enter Git for Windows or winget binary path:" with
+      | None -> get_gitbin ()
+      | Some gitbin ->
+        match contains_git gitbin, OpamSystem.bin_contains_bash gitbin with
+        | Some _, false ->
+          OpamConsole.note "Using git from %s" gitbin;
+          gitbin
+        | Some _, true ->
+          OpamConsole.error
+            "A bash executable was found in %s, which will override \
+             usual bash. Please check you binary path"
+            gitbin;
+          get_gitbin ()
+        | None, _ ->
+          OpamConsole.error "No git executable found in %s" gitbin;
+          get_gitbin ()
+    in
+    if OpamConsole.confirm ~default:false
+        "Do you want to specify which git to use (non cygwin)?" then
+      Some (get_gitbin ())
+    else None
 
 let windows_checks ?cygwin_setup config =
   let vars = OpamFile.Config.global_variables config in
@@ -922,7 +929,15 @@ let windows_checks ?cygwin_setup config =
       >>| OpamFilename.Dir.to_string)
   in
   OpamCoreConfig.update ?cygbin ();
-  git_for_windows_check cygbin;
+  let gitbinpath = git_for_windows_check cygbin in
+  OpamCoreConfig.update ?gitbinpath ();
+  let config =
+    match gitbinpath with
+    | Some gitbin ->
+      OpamFile.Config.with_gitbinfield
+        (OpamFilename.Dir.of_string gitbin) config
+    | None -> config
+  in
   config
 
 let update_with_init_config ?(overwrite=false) config init_config =
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -635,6 +635,57 @@ let init_checks ?(hard_fail_exn=true) init_config =
   if hard_fail && hard_fail_exn then OpamStd.Sys.exit_because `Configuration_error
   else not (soft_fail || hard_fail)
 
+let git_for_windows_check =
+  if Sys.win32 then
+    fun cygbin ->
+      let gitcmd = OpamSystem.resolve_command "git" in
+      match gitcmd with
+      | None -> (* git *)
+        OpamConsole.note "no git!"
+      | Some gitcmd ->
+        let bindir = Filename.dirname gitcmd in
+        let bashcmd = Filename.concat bindir "bash.exe" in
+        (* We don't want to resolve it, just to check if it exists and is exe *)
+        let check_bash = OpamSystem.resolve_command ~env:[||] bashcmd in
+        match check_bash with
+        | None -> ()
+        | Some bashcmd -> (* found bash in git prefix *)
+          let is_cygwin =
+            OpamStd.Option.map_default (fun prefix ->
+                OpamStd.String.starts_with ~prefix bashcmd)
+              false cygbin
+          in
+          let git4cmd_bindir =
+            Filename.dirname bindir
+          in
+          let git4cmd_cmddir =
+            Filename.concat git4cmd_bindir "cmd"
+          in
+          let is_git4win =
+            let file = Filename.concat git4cmd_cmddir "git.exe" in
+            Sys.file_exists file
+          in
+          if is_cygwin then
+            OpamConsole.warning
+              "You are using Cygwin git, \
+               consider installing Git for Windows: https://gitforwindows.org \
+               and check that it is well in your Path"
+          else
+          if is_git4win then
+            OpamConsole.warning
+              "You are using Git for Windows, \
+               but your path contains the wrong binary directory.\n\
+               Consider reinstalling or add %s in your path instead of %s"
+              git4cmd_cmddir bindir
+          else
+            OpamConsole.warning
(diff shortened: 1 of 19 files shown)


src/client/opamClient.ml
```diff
@@ -635,6 +635,57 @@ let init_checks ?(hard_fail_exn=true) init_config =
   if hard_fail && hard_fail_exn then OpamStd.Sys.exit_because `Configuration_error
   else not (soft_fail || hard_fail)
 
+let git_for_windows_check =
+  if Sys.win32 then
```
kit-ty-kate, line 639: Shouldn’t this code be also ran on Cygwin?
```suggestion
  if Sys.win32 || Sys.cygwin then
```

jonahbeckford: I personally agree with @kit-ty-kate but I'm not a Cygwin user. I could see it going either way; Cygwin (not MSYS2) is meant as a full Unix replacement including userland tools.

src/client/opamClient.ml
```diff
@@ -635,6 +635,57 @@ let init_checks ?(hard_fail_exn=true) init_config =
...
+    fun cygbin ->
+      let gitcmd = OpamSystem.resolve_command "git" in
```
jonahbeckford, line 641: To deal with the common edge case where both Cygwin (or MSYS2) git and Git for Windows git are installed, I've been doing PATH modifications in DkML so that Git for Windows always is in front of the PATH (before `/usr/bin`). That may be overkill here but the problem remains what to do when they are both installed.

Currently this PR will search the PATH and only check the first `git` ... which will almost always be Cygwin/MSYS2 `/usr/bin/git`.

Not sure how, but could `OpamSystem.resolve_command` be adjusted to find the correct `git`?

jonahbeckford: FYI: The PATH hack for shells (`PATH="${PATH/:\/usr\/bin:\/bin:/:}"`) seen in https://learn.microsoft.com/en-us/vcpkg/users/platforms/mingw#how-to-avoid-mixing-different-installations seems like a common way to fix these sorts of Cygwin/MSYS2 problems. So my suggestion would be to have a `OpamSystem.resolve_command ~safe:()` mode which does the PATH hack in OCaml to find if there is a Git for Windows. And if that fails, then do a normal `OpamSystem.resolve_command` if needed.

rjbou:
PR update with new mechanism: at init, ask for user to specify its git path (git4win or winget), stores it, and put it at the beginning of build & other paths.

rjbou:
@jonahbeckford could you test the new git selection mechanism, if you have some time?

rjbou:
I can try to generate one.

jonahbeckford:
Don't worry.

I am using the `opam.exe` associated with [Build-Windows (x86_64-pc-windows, x86_64-pc-cygwin, 4.14.1)](https://github.com/jonahbeckford/opam/actions/runs/7188079781/job/19577097065#logs) based off this PR and https://github.com/ocaml/opam/pull/5762.

### Testing

- [ ] A. Does it run in Windows Sandbox. ⛔No

### Issue A

Looks like it is missing the VC Runtime redistributables (`VCRUNTIME140`):

```
with-dkml ldd 'Y:\source\dkml\build\pkg\bump\WindowsSandbox\tools\opam.exe'                                                                                                       
        ntdll.dll => /c/WINDOWS/SYSTEM32/ntdll.dll (0x7ff847410000)
        KERNEL32.DLL => /c/WINDOWS/System32/KERNEL32.DLL (0x7ff8454d0000)
        KERNELBASE.dll => /c/WINDOWS/System32/KERNELBASE.dll (0x7ff8447b0000)
        apphelp.dll => /c/WINDOWS/SYSTEM32/apphelp.dll (0x7ff8415d0000)
        ADVAPI32.dll => /c/WINDOWS/System32/ADVAPI32.dll (0x7ff845e80000)
        msvcrt.dll => /c/WINDOWS/System32/msvcrt.dll (0x7ff846630000)
        sechost.dll => /c/WINDOWS/System32/sechost.dll (0x7ff845160000)
        RPCRT4.dll => /c/WINDOWS/System32/RPCRT4.dll (0x7ff8452e0000)
        GDI32.dll => /c/WINDOWS/System32/GDI32.dll (0x7ff8452b0000)
        win32u.dll => /c/WINDOWS/System32/win32u.dll (0x7ff844780000)
        gdi32full.dll => /c/WINDOWS/System32/gdi32full.dll (0x7ff844f80000)
        msvcp_win.dll => /c/WINDOWS/System32/msvcp_win.dll (0x7ff844b60000)
        ucrtbase.dll => /c/WINDOWS/System32/ucrtbase.dll (0x7ff844e60000)
        USER32.dll => /c/WINDOWS/System32/USER32.dll (0x7ff846dd0000)
        SHELL32.dll => /c/WINDOWS/System32/SHELL32.dll (0x7ff8455a0000)
        WS2_32.dll => /c/WINDOWS/System32/WS2_32.dll (0x7ff845e00000)
        VERSION.dll => /c/WINDOWS/SYSTEM32/VERSION.dll (0x7ff839f20000)
        VCRUNTIME140.dll => /c/WINDOWS/SYSTEM32/VCRUNTIME140.dll (0x7ff8320d0000)
        VCRUNTIME140_1.dll => /c/WINDOWS/SYSTEM32/VCRUNTIME140_1.dll (0x7ff8317e0000)
        MSVCP140.dll => /c/WINDOWS/SYSTEM32/MSVCP140.dll (0x7ff831f50000)
```

That isn't the end of the world ... I believe you will be building completely static versions. In the meantime, I'll install https://aka.ms/vs/17/release/vc_redist.x64.exe

jonahbeckford:
I have a screen capture at https://vimeo.com/893933043. Most of it is smooth (thx!) but there are a few adjustments needed:

- [ ] 2:27. I think we should recommend `winget install Git.Git` first, before the URL https://gitforwindows.org, because the Git for Windows installation has **ten (10)** multiple choice pages! Especially if `winget` is in the PATH (I should do that too in DkML).
- [ ] 3:45. opam says "You can install it (Git) and relaunch opam initialization". In reality, I have install Git for Windows, then press Ctrl-C to get out of the opam initialization (not obvious), and then rerun `opam init`.
- [ ] 4:24. When redoing `opam init` the Cygwin is re-downloaded and re-installed (although I think this time this goes faster). (Low-priority)
- [ ] 5:34. I was confused what I was supposed to answer. The question "Do you want to specify which git to use? (non cygwin)" didn't give me a clue whether "y" or "n" would use the git.exe it found in the PATH. And I don't know what "(non cygwin)" implied.

rjbou:
Thanks a lot for this detailed feedback!!! The workflow needs to be updated:
* see if it is possible for some setup to have the git check before cygwin install 
* abort properly or propose to pause and give the path ?
* rewording, rewording, rewording

src/client/opamClient.ml
```diff
@@ -635,12 +635,118 @@ let init_checks ?(hard_fail_exn=true) init_config =
...
+        in
+        OpamConsole.menu "Which Git should opam use?"
+          ~default:`Default ~no:`Default ~options
+      in
+      match prompt () with
+      | `Default -> None
+      | `Specify -> loop ()
+      | `Location gitbin -> loop ~gitbin ()
+      | `Abort ->
+        OpamConsole.note "Once your choosen Git installed, relaunch opam init.";
```
jonahbeckford, line 704:
```suggestion
        OpamConsole.note "Once your choosen Git installed, open a new PowerShell or Command Prompt window, and relaunch opam init.";
```

because even after a successful `winget install Git.Git` the prompt keeps coming back. I suspect that Git is not visible in the PATH of the same PowerShell window.

jonahbeckford:
I have a small code suggestion, but otherwise LTGM. Thanks!

kit-ty-kate (commented):
I’m not convinced `gitbinfield` is a very descriptive name of the new option. What about `git-path` ?

rjbou:
> I’m not convinced `gitbinfield` is a very descriptive name of the new option. What about `git-path` ?

It's normal, it's not meant to stay :)

Looking for name ideas for:
* `gitbinfield` in config: the binary directory of git
* `gitbinpath` in `OpamCoreConfig.t`:  the binary directory of git
* `gitbin` in `OpamClient.init/reinit`:  the binary directory of git if given or no git setup
* `--git-location` & `--no-git-location`: options to specify git binary directory git & no git setup
* `bin_contains_bash` in `OpamSystem`: does the given directory (that should be a `/bin` one) contains `bash.exe`

rjbou:
> I have a small code suggestion, but otherwise LTGM. Thanks!

Thanks a lot! Indeed, better advertise to launch another terminal session to have the path updated.

src/client/opamClient.ml
dra27: No - for the same reason as we're advising users aiming for native Windows opam experience to use Git for Windows, users of Cygwin's opam should be using Cygwin's git (or at least, we shouldn't be advising the contrary). Cygwin's opam package is maintained, as well.

src/client/opamClient.ml
```diff
@@ -635,12 +635,118 @@ let init_checks ?(hard_fail_exn=true) init_config =
   if hard_fail && hard_fail_exn then OpamStd.Sys.exit_because `Configuration_error
   else not (soft_fail || hard_fail)
 
-let windows_checks ?cygwin_setup config =
+let git_for_windows_check =
+  if not Sys.win32 && not Sys.cygwin then fun ?gitbin:_ () -> None else
```
dra27, line 639:
```suggestion
  if Sys.unix then fun ?gitbin:_ () -> None else
```

src/client/opamClient.ml
```diff
@@ -635,12 +635,118 @@ let init_checks ?(hard_fail_exn=true) init_config =
...
+          | Some git ->
+            Some (git, OpamSystem.bin_contains_bash p)
+          | None -> None)
+    in
+    let get_gitbin ?gitbin () =
+      let bin =
+        match gitbin with
+        | Some _ -> gitbin
+        | None ->
+          OpamConsole.read "Enter winget or Git for Windows binary path (\\bin):"
```
dra27, line 659: I'm not sure what the "winget" bit means here? Is "Please enter the path containing git.exe (e.g. C:\Program Files\Git\cmd)" closer?

src/client/opamClient.ml
```diff
@@ -635,12 +635,118 @@ let init_checks ?(hard_fail_exn=true) init_config =
...
+      | None -> None
+      | Some gitbin ->
+        match contains_git gitbin, OpamSystem.bin_contains_bash gitbin with
+        | Some _, false ->
+          OpamConsole.msg "Using Git from %s" gitbin;
+          Some gitbin
+        | Some _, true ->
+          OpamConsole.error
+            "A bash executable was found in %s, which will override \
+             usual bash. Please check you binary path."
```
dra27, line 671:
```suggestion
             Cygwin's bash. Please check your binary path."
```

src/client/opamClient.ml
dra27: Yes, indeed. There are a couple of solutions, if one is assuming that the user really installed Git:
- Refresh `PATH` from the Registry, but I don't recommend that we do that. Git for Windows setup will (or should) have broadcast [`WM_SETTINGCHANGE`](https://learn.microsoft.com/en-us/windows/win32/winmsg/wm-settingchange) which is the indicator to do this, the PATH environment can then be reloaded from HKEY_CURRENT_USER\Environment and HKEY_LOCAL_MACHINE\System\CurrentControlSet\Control\Session Manager\Environment, but this is really fiddly and it doesn't respect that the user may have overridden PATH.
- Read the InstallPath value from HKEY_LOCAL_MACHINE\SOFTWARE\GitForWindows - I haven't double-checked what happens with a user-only installation, but after a system-wide installation, that contains the install location of Git for Windows

src/client/opamClient.ml
```diff
@@ -635,12 +635,118 @@ let init_checks ?(hard_fail_exn=true) init_config =
...
+      match gitbin with
+      | Some (Right ()) -> None
+      | Some (Left gitbin) ->
+        header ();
+        get_gitbin ~gitbin:(OpamFilename.Dir.to_string gitbin) ()
+      | None ->
+        if OpamStd.Sys.tty_out then
+          (header ();
+           OpamConsole.msg
+             "Cygwin Git is functional but have credentials issues, \
```
dra27, line 717:
```suggestion
             "Cygwin Git is functional but can have credentials issues, \
```
Possibly worth adding that Cygwin's git is fine for public repositories, but can have issues with the credentials for private repositories?

src/client/opamClient.ml
```diff
@@ -635,12 +635,118 @@ let init_checks ?(hard_fail_exn=true) init_config =
...
+              we recommend using:\n%s\n"
+             (OpamStd.Format.itemize (fun s -> s)
+                [ "winget with 'winget install Git.Git'";
+                  "Git for Windows available at https://gitforwindows.org" ]);
```
dra27, line 721:
```suggestion
                [ "Install via with 'winget install Git.Git'";
                  "Git for Windows can be downloaded and installed from https://gitforwindows.org" ]);
```

dra27 (commented):
Some comments on the message, and a few drive-by comments. As ever, I'm sorry to enter these discussions late. I think the general mechanism here is very good, and what I'm suggesting next reuses everything that's already here, just in a slightly different way.

I wonder if there is a possibly conceptually different approach to storing the path to Git's bin (with apologies if this has already been discussed and ruled out, because I know that @rjbou and @kit-ty-kate have discussed this extensively!).

The mechanism for adding Cygwin's `bin` directory to `PATH` is a necessary evil for Windows support, but the (admittedly very) long-term trajectory for Windows is that it should one day be unnecessary. Conversely, the "tricks" to get a Git installation's `bin` directory into `PATH` is something we'd likely always need to do.

I'd be tempted, therefore, to make it the job of opam init to determine if it can find the _exact_ path to a `git.exe` which it likes (Git for Windows) and store that - so internally `opam` stores `C:\Program Files\Git\cmd\git.exe` (that path can be determined using the Registry). opam should obviously offer the choice, but I think `opam init` should attempt to find that version of git, regardless of `PATH` (the menu could of course _offer_ the `git.exe` found in PATH).

The rest of the check (bash in PATH, etc.) is more general. We should be checking the _entire_ PATH to see if Cygwin's bash is going to be shadowed anyway, and we can also _warn_ that the `git` found in PATH does not match the one opam wants to use, or has that the user has done something which causes `git.exe` and `bash.exe` to be next to each other (non-recommended Git-for-Windows configuration, other package managers, etc.)

So, in summary: I think opam should very precisely determine the `git.exe` it wants to use (same idea as OPAMFETCH, etc.) and store that - all its git operations should do that. opam shouldn't then do any magic PATH manipulation where git is concerned - it should simply warn the user (possibly at each call - i.e. at `opam install`, etc.) that the git in PATH may cause problems for package installations. That approach is similar to what Git-for-Windows itself does - it provides the ability to put git-bash in PATH, but strongly recommends not doing so.

src/client/opamClient.ml
kit-ty-kate:
```suggestion
          OpamConsole.read "Please enter the path containing git.exe (e.g. C:\Program Files\Git\cmd):"
```

kit-ty-kate (approved):

kit-ty-kate:
Thanks a lot!

PR #5723, opened 2023-11-09
reftest engine tweaks for Windows

dra27:
On Windows, when run in GitHub Actions, the `opam env` in reftests is using a `sh` as the default shell (so we get `ENV='value'; export ENV` format). When run locally, I get `cmd` as the default shell, which causes conflicts as we get `set ENV=value` notation instead.

I had a deeper dive into this as part of looking at #5636 and #5714. Process hierarchy is a slightly strange thing on Windows - it's already the case that the ancestry returned gets "chopped" (process groups, etc. can do this). The ancestry being returned is not incorrect, or at least it is consistent - tools such as Process Explorer see the same thing.

For whatever reason, the GitHub Actions get Cygwin's `sh.exe` as one of the ancestor processes. Locally, it's `make.exe` for me 🤷 I'm not totally sure why, but it doesn't particularly matter - the ancestry detection _is_ correct when run from cmd or powershell, which is when it's needed. When native opam is run from a Cygwin shell, `SHELL` is set, and opam uses that.

So the only issue is the reftests, which don't propagate the `SHELL` environment variable. The first commit here rather than propagating it, simply sets it. We might consider doing this for Unix as well - I'm not sure whether running `make tests` from `fish` has the same problem. I think this could also be done with `OPAMSHELL`, but given that we weren't propagating the `SHELL` variable before, I don't think an issue with just setting it instead.

Also while looking at things for #5636, it's quite easy now to end up with the opam root or the temporary directory being translated to Cygwin notation which doesn't then get caught by the regex for conversion to `BASEDIR` or `OPAMTMP`. I've factored out the code which was doing the previous forward/backslash manipulation and exposed `OpamSystem.apply_cygpath` so now those directories are captured with backslashes, forward-slashes *or* in Cygwin notation.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -128,6 +128,7 @@ users)
   * Add environment variables path addition [#5606 @rjbou]
   * Remove duplicated environment variables in environmenet [#5606 @rjbou]
   * Add `PATH` to replaceable variables [#5606 @rjbou]
+  * Set `SHELL` to `/bin/sh` in Windows to ensure `opam env` commands are consistent [#5723 @dra27]
 
 ## Github Actions
   * Add coreutils install for cheksum validation tests [#5560 @rjbou]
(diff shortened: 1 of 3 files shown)


rjbou:
fixes #5377

rjbou (approved):
thanks!
I modified `apply_cygpath` comment, lgtm!

PR #5727, opened 2023-11-14
Update the release script according to the opam 2.2.0~alpha3 release process

kit-ty-kate:
Following changes from https://gitlab.com/kit-ty-kate/qemu-base-images

Related to https://github.com/ocaml/opam/pull/5616/files#r1273867852 for the change in the Makefile

Tested on top of https://github.com/ocaml/opam/pull/5383


kit-ty-kate:
Tested successfully on top of #5737. I also spedup the script and it now takes only 1h37 (vs more than 7h before)

PR #5732, opened 2023-11-15
Bump the version number after the release of 2.2.0~alpha3 and update the release steps

kit-ty-kate:


rjbou:
is there a reason for choosing alpha4~dev instead of beta1~dev ?

kit-ty-kate:
> is there a reason for choosing `alpha4~dev` instead of `beta1~dev` ?

we can’t predict the future so it’s better if we need to urgently release another alpha for whatever reason.

rjbou:
fair enough :+1:

rjbou (approved):

PR #5715, opened 2023-11-02
opam env: fix carriage return on Cygwin shell

rjbou:
fix #5684

PR #5743, opened 2023-11-27
`OPAMSWITCH` and `OPAMROOT` reverts

### LINKED ISSUE #5696: opam 2.2.0~alpha2: OPAMSWITCH does not work anymore
The reason is that the shell hook resets `OPAMSWITCH` to the empty string as soon as it is set. This is because `opam env --shell=zsh --readonly` (my shell is zsh) prints `OPAMSWITCH=''; export OPAMSWITCH;` as the first line.

Before, the same command did not print `OPAMSWITCH`. This seems to be a consequence of https://github.com/ocaml/opam/pull/5417.

rjbou:
~Store old `OPAMSWITCH` in last env file, in order to revert it properly~
Don't store `OPAMSWITCH` & `OPAMROOT` in the last environment file, like that they are not reverted. They are set with `--set-switch` and `--set-root`, they are meant to remain in the shell environment.

fixes #5696

queued on #5715 for windows ci

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -107,6 +107,7 @@ users)
 ## Reftests
 ### Tests
   * Add some additional test to tree, for `--dev` && `--no-switch` [#5687 @rjbou]
+  * switch-set: add test that checks unsetting `OPAMSWITCH` when it was set by `opam env --set-switch` on an already set `OPAMSWITCH` variable in environment [#5742 rjbou]
 
 ### Engine
   * Set `SHELL` to `/bin/sh` in Windows to ensure `opam env` commands are consistent [#5723 @dra27]
--- a/src/client/opamConfigCommand.ml
+++ b/src/client/opamConfigCommand.ml
@@ -297,7 +297,15 @@ let ensure_env_aux ?(base=[]) ?(set_opamroot=false) ?(set_opamswitch=false)
         not (String.equal upd.envu_var "OPAM_LAST_ENV"))
       updates
   in
-  let last_env_file = write_last_env_file gt switch updates in
+  let last_env_file =
+    write_last_env_file gt switch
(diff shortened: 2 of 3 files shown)


rjbou:
CI issue
```diff
@@ -96,11 +96,20 @@ opam sw show
 ### sh unset-opamswitch.sh opam
 foo
 OPAMSWITCH='bar'; export OPAMSWITCH;
-OPAMSWITCH=''; export OPAMSWITCH;
-bar
+unset-opamswitch.sh: line 13: $'\r': command not found
+unset-opamswitch.sh: line 13: $'\r': command not found
+unset-opamswitch.sh: line 13: $'\r': command not found
+unset-opamswitch.sh: line 13: $'\r': command not found
+unset-opamswitch.sh: line 13: $'\r': command not found
+foo
```
will be resolved by #5715

rjbou:
Other solution could be to not store `OPAMSWITCH` in last env file at all

rjbou:
Updated!

src/core/opamConsole.ml
```diff
@@ -511,13 +521,13 @@ let clear_status =
 
 let print_message =
   if Sys.win32 then
-    fun ch fmt ->
+    fun ?(stdout=stdout) ch fmt ->
       flush (if ch = `stdout then stderr else stdout);
       clear_status ();
       (* win32_print_message *always* flushes *)
       Printf.ksprintf (win32_print_message ch) fmt
```
dra27, line 528: `win32_print_message` needs a `~stdout` parameter, too (sorry - these functions have ended up a bit of a mess over time... I think just adding that parameter here will be easier than trying to straighten out all the old VT100 emulation code)

kit-ty-kate (approved):

kit-ty-kate:
Thanks!

PR #5757, opened 2023-12-08
basic tracing

c-cube:
This is WIP.

I'm adding tracing to find why opam is so slow on my machine. It gives excellent insights on where time is spent (e.g after `opam upd`, all 30k files are read one by one, to update the cache).

I've been instrumenting things as I explore but it needs to be more principled in general I think.

The overhead when tracing is not active (ie `TRACE_FILE` is not set) should be fairly low.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -97,6 +97,7 @@ users)
 ## Shell
 
 ## Internal
+  * Add tracing for opam itself, enabled if `TRACE_FILE` is set in the environment [#5757 @c-cube]
 
 ## Internal: Windows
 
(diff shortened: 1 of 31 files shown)


c-cube:
- the tracing format is the classic JSON file for `chrome://tracing`, which can also be read by https://ui.perfetto.dev/
 (see also past example of this [1](https://discuss.ocaml.org/t/ann-runtime-events-tools-0-5-0/13235). So you can easily visualize it.

   It can look like this:
2023-12-08-153651_2042x736_scrot

- I think we could use `Unix.gettimeofday()` instead, I think, we don't need that high precision in general. I got a bit over enthusiastic.

kit-ty-kate (commented):
LGTM on first glance

kit-ty-kate:
We’re trying to get 2.2.0 out the door asap and i’d prefer to merge this after the release to avoid unnecessary time spent in extra maintenance and CI work (don’t worry the CI failure is unrelated to you changes).
Is that ok for you?

c-cube:
Of course, no worries! This is really a tool to debug/improve opam itself, not something end users urgently need, I think.

kit-ty-kate (commented):
I've squashed the PR and cleaned a couple trailing whitespaces. It looks good overall but there are a couple of things to split off into their own PRs I think

PR #5762, opened 2023-12-12
Upload and retain opam.exe artifact

jonahbeckford:
Makes it easy for code reviewers (so they don't have to rebuild the source code). Makes it easy to compare-and-constrast old version.

Helps https://github.com/ocaml/opam/pull/5718

### DIFF
--- a/.github/workflows/ci.ml
+++ b/.github/workflows/ci.ml
@@ -326,8 +326,65 @@ let main_build_job ~analyse_job ~cygwin_job ?section runner start_version ~oc ~w
     else
       (matrix, []) in
   let matrix = ((platform <> Windows), matrix, includes) in
-  let needs = if platform = Windows then [analyse_job; cygwin_job] else [analyse_job] in
+  let needs =
+    [analyse_job]
+    @ (if platform = Windows then [cygwin_job] else [])
+  in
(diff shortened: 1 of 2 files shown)


rjbou:
Thanks a lot for the idea!
I'll update the PR with a generation for unix & macos too, triggered via a label.

rjbou:
Updated the PR with:
* uploading artefact for linux, mac & windows binaries
* uploading artefact if the pR have label `CI:BINARIES`
* artefact available for 15 days

rjbou:
oups, there is no gh cli on windows jobs.

kit-ty-kate:
Windows artifacts are already uploaded since https://github.com/ocaml/opam/pull/5718/commits/9ccb2d12b012f7166bc06da317165d5811f6b446 but i'm not sure macOS needs it too given how fast opam takes to compile on there.

I'd be incline to close this if everyone agrees.

jonahbeckford:
The artifacts were generated because of how difficult it was to reproduce the exact environment (C compiler, Unix shell, 3rd party downloads) that generated the opam.exe binary of a PR.  Much more difficult on Windows than other OS-es. The compile speed was unimportant.

If no one has complained on other OS-es, then okay to drop.

PR #5769, opened 2023-12-19
Upgrade to mccs.1.1+17

kit-ty-kate:
Opening this PR in prevision of the release.
See https://github.com/ocaml-opam/ocaml-mccs/pull/50 for detail of the fix.


kit-ty-kate:
Queued on https://github.com/ocaml/opam-repository/pull/24966. Once it is merged only the opam-repository commit should be changed and it should be good to go.

rjbou (approved):
thanks!

PR #5775, opened 2024-01-03
Update Install.md FreeBSD instructions

lukstafi:
If someone is a complete FreeBSD greenhorn newbie, they might not realize that the opam package is `ocaml-opam`, and installing from source is painful.


kit-ty-kate (approved):

rjbou:
Thanks!

Issue #4958, opened 2021-12-10
a possible regression on macOS /var/folders permission problem


ivg:
It looks like that #3576 re-appeared in 2.1.2, we have the "operation not permitted" error on our [CI builds][1] after 2.1.2 has been released
```
 - make: error: couldn't create cache file '/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/xcrun_db-cZTiPW1A' (errno=Operation not permitted)
```

Here's the [log][2] for the same action that passes without problems on 2.1.1 (and older versions), and here is the [action][3] itself for your reference. Any suggestions on workarounds are very welcome!

[1]: https://github.com/BinaryAnalysisPlatform/bap/runs/4473828489?check_suite_focus=true#step:3:397
[2]: https://github.com/BinaryAnalysisPlatform/bap/runs/4413569831?check_suite_focus=true
[3]: https://github.com/BinaryAnalysisPlatform/bap/blob/master/.github/workflows/build-dev-repo.yml

kit-ty-kate:
This is puzzling. I’m not seeing anythin related change in 2.1.2 compared to 2.1.1: https://github.com/ocaml/opam/compare/2.1.1...2.1.2
Are you still having the issue? Was it reproducible?

ivg:
Yep, it's reproducible on each run, and yes it's puzzling, as I checked every single line of the change and didn't find anything relevant.

kit-ty-kate:
@smorimoto is there a related change in setup-ocaml by any chance?

ivg:
Those error messages turned to be red herrings. Apparently, they existed all the time, even when we were passing the installation test. The true problem was the below those error messages and is a manifestation of the https://github.com/ocaml/ocaml/pull/9895 so the workaround is `rm $(opam var lib)/ocaml/VERSION`

P.S. it is still interesting what those messages mean and are they a problem or not, but closing for now.

kit-ty-kate:
mmmh, it seems to be a manifestation of the problem that might be fixed by https://github.com/ocaml/opam/pull/4797: macOS tools have a bug in which it expects the cache directories to always be writable if it can read them (obviously not true when sandboxed). `xcrun_db-*` files seem to come from xcode so it seems that they all have common code with `/usr/bin/security`.
I’ve create a bug report on Apple’s bugtracker last year but I still did not get any answers.

[...]

PR #5780, opened 2024-01-12
Fix macos sandbox script incorrectly assuming that `getconf` doesn't exist

ElectreAAS:
As noted in #5778, this is a proper bugfix:
- `[ -z /usr/bin/getconf ]` tests if the literal string "/usr/bin/getconf" is empty - which is always false - and doesn't probe the filesystem at all.
- `[ -n "$(which getconf)" ]` actually asks the system if that command exists. The place where it is defined (/usr/bin or anything else) doesn't matter since we call it directly below.

It should be noted that if `$TMPDIR` is not set *and* `getconf` doesn't exist, we carry on the program silently. Is that intended?

I do not have a macos machine on hand and cannot test this, but this could be related to #4958

### DIFF
--- a/src/state/shellscripts/sandbox_exec.sh
+++ b/src/state/shellscripts/sandbox_exec.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# This script is only used on macOS, where /bin/bash is both guaranteed to exist and
+# This script is only used on macOS, where /bin/bash is both guaranteed to exist
 # and to be for the native architecture, which is why /usr/bin/env bash is not used.
 # See https://github.com/ocaml/opam/issues/5450
 set -ue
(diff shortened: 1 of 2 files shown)


src/state/shellscripts/sandbox_exec.sh
```diff
@@ -20,10 +20,10 @@ add_mounts() {
 }
 
 if [ -z ${TMPDIR+x} ]; then
-  # Others applications obtain the per-user temporary
+  # Other applications obtain the per-user temporary
   # directory differently; the latter should be made readable/writable
   # too and getconf seems to be a robust way to get it
-  if [ -z /usr/bin/getconf ]; then
+  if [ -n "$(which getconf)" ]; then
```
kit-ty-kate, line 26: Using the `command` builtin is a bit faster and more future proof (in case Apple wants to remove `which`). It is also what is used in the rest of the file (see line 38 for example)
```suggestion
  if command -v getconf > /dev/null; then
```

kit-ty-kate (commented):
Note that by default macOS already defines `TMPDIR` during startup so this code is almost never executed in practice (i think the only way would be if you manually unset that variable or if you start your shell with `env -i`, which is definetely possible but unlikely)

src/state/shellscripts/sandbox_exec.sh
ElectreAAS: Oh that's what `command` does!
Thanks for the tip :)
Fixed

kit-ty-kate (approved):

PR #5788, opened 2024-01-18
Workflow fixes

dra27:
- depexts workflows have been broken since #5769
- The hygiene check that the workflows were properly generated didn't have the correct trigger
- We should run _all_ workflows on pushes to master or the maintenance branches


kit-ty-kate:
Thanks!

PR #5793, opened 2024-01-22
Add winsymlinks:native to the CYGWIN environment variable when running a Cygwin executable on Windows

### LINKED ISSUE #5782: Cygwin internal installation should add `winsymlinks:native` to `CYGWIN`
This is a first step towards improved symlink support. We should ultimately be advising that developers enable Developer Mode (which allows symlinks to be used without elevation), and we should also have a better fallback for tarballs which include symlinks, but in the meantime, using the environment setup from #5781:

```
C:\>opam install ocamlbuild
=== install 1 package
  * ocamlbuild 0.14.2

#=== ERROR while fetching sources for ocamlbuild.0.14.2 =======================#
OpamSolution.Fetch_fail("C:\\Devel\\opam-3\\opam.exe: \"lstat\" failed on C:\\Users\\DRA\\AppData\\Local\\Temp\\opam-126020-bae9bd\\ocamlbuild-0.14.2\\examples\\07-dependent-projects\\libdemo: No such file or directory")


│ ▼ fetch ocamlbuild 0.14.2
└─
╶─ No changes have been performed
```

The issue is that Cygwin will have used its own symlink format (most likely a WSL symlink) which native Windows opam can't read. The workaround is easy:

```
C:\>set CYGWIN=winsymlinks:native

C:\>opam install ocamlbuild
=== install 1 package
  * ocamlbuild 0.14.2

▼ retrieved ocamlbuild.0.14.2  (cached)
* installed ocamlbuild.0.14.2
Done.
```
At the same point as we add the internal Cygwin bin to the PATH, we should ensure that `winsymlinks:native` (or `winsymlinks:nativestrict`, if the user has elected to set that) are set and remove either `winsymlinks:sys` or `winsymlinks:lnk` if found (i.e. it's part of scrubbing the environment.). Cygwin's `CYGWIN` variable is documented [here](https://cygwin.com/cygwin-ug-net/using-cygwinenv.html) and its symlink support is documented [here](https://cygwin.com/cygwin-ug-net/using.html#pathnames-symlinks).

kit-ty-kate:
~Fixes #5782~ 

Actually no, at least not yet. Upon creating the PR I realized another point in the code also modifies `CYGWIN` (in a destructive manner)

@dra27 I’m reading the comments for `OpamProcess.cygwin_create_process_env` and the use of `win_create_process` (which i’m also realizing does not exist in OCaml >= 5.0, which means opam can’t be compiled on Windows if you’re using ocaml 5) and I’m not sure if it’s just a relic of the past that can now be replaced by `Unix.create_process_env` or not? Was the issue encountered with `Unix.create_process_env` ever got fixed or was it only an issue in opam and thus not something that could actually be fixed upstream?

If it is a relic of the past I’ll go ahead and remove it and together with the current code, it should work.

### DIFF
--- a/doc/pages/Manual.md
+++ b/doc/pages/Manual.md
@@ -1175,6 +1175,7 @@ files.
     - `OPAM_PACKAGE_VERSION=<ver>` (`<ver>` is the version of the package being built/installed/removed)
     - `OPAMCLI=2.0` (since opam 2.1)
     - `TMP` and `TMPDIR` are set by the sandbox script (bubblewrap), but should not be relied on since the sandbox is not used on all platforms and can be disabled by the user.
+    - `CYGWIN=winsymlinks:native` on Windows, or `CYGWIN=$CYGWIN winsymlinks:native` if `CYGWIN` is defined and not empty and `CYGWIN` does not contain either `winsymlinks:native` or `winsymlinks:nativestrict` already (since opam 2.2). In some cases `noglob` can also be added to this variable, such that the default value becomes `CYGWIN=winsymlinks:native noglob` (since opam 2.1)
 
   See [`x-env-path-rewrite:`](#opamfield-x-env-path-rewrite)
   for path portability of environment variables on Windows.
--- a/master_changes.md
+++ b/master_changes.md
@@ -74,6 +74,7 @@ users)
 ## Build
   * Do not check for cppo in the configure script (not used directly anymore since #5498) [#5794 @kit-ty-kate]
   * Upgrade vendored cmdliner to 1.2.0 [#5797 @kit-ty-kate]
+  * Add winsymlinks:native to the CYGWIN environment variable when installing a package on Windows [#5793 @kit-ty-kate - fix #5782]
 
 ## Infrastructure
   * Fix depexts CI workflow and ensure all workflows run on master push [#5788 @dra27]
--- a/src/core/opamProcess.ml
+++ b/src/core/opamProcess.ml
@@ -61,6 +61,9 @@ let cygwin_create_process_env prog args env fd1 fd2 fd3 =
    *        time in OPAM.
    *
    * [This stray " is here to terminate a previous part of the comment!]
+   *
+   * Automatically setting CYGWIN to at least contain winsymlinks:native was
+   * also added later to make sure the native links are used.
    *)
   let make_args argv =
     let b = Buffer.create 128 in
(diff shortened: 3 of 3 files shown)


dra27:
In somewhat haste, nothing's been fixed in OCaml (nor is ever likely to be), so I believe the code is still needed. The stubs are still there, they are called `caml_unix_create_process` and `caml_unix_create_process_native` in OCaml 5.0+

dra27 (commented):
Thanks for looking at this! The part in `OpamProcess` is completely orthogonal - that's all to do with getting the escaping of arguments correct. It doesn't destructively change `CYGWIN` though - it's editing it (in what's looking to me 9 years later to be a possibly over-pedantic fashion, but perhaps the rules for specifying options in the `CYGWIN` variable used to be more strict).

Regardless, it's definitely not interfering because my local workaround has been to `set CYGWIN=winsymlinks:strict` in my shell prior to invoking `opam` and it's not overriding it 🙂

kit-ty-kate:
@dra27 done

dra27:
A thought which didn't occur to me before - does this also apply when we're running commands from within opam (e.g. quite critically, `tar`)?

doc/pages/Manual.md
```diff
@@ -1175,6 +1175,7 @@ files.
     - `OPAM_PACKAGE_VERSION=<ver>` (`<ver>` is the version of the package being built/installed/removed)
     - `OPAMCLI=2.0` (since opam 2.1)
     - `TMP` and `TMPDIR` are set by the sandbox script (bubblewrap), but should not be relied on since the sandbox is not used on all platforms and can be disabled by the user.
+    - `CYGWIN=winsymlinks:native` (or `CYGWIN=$CYGWIN winsymlinks:native` if CYGWIN is defined and not empty and CYGWIN does not contain either `winsymlinks:native` or `winsymlinks:nativestrict` already) on Wind    ows (since opam 2.2)
```
rjbou, line 1178:
```suggestion
    - `CYGWIN=winsymlinks:native` (or `CYGWIN=$CYGWIN winsymlinks:native` if `CYGWIN` is defined and not empty and `CYGWIN` does not contain either `winsymlinks:native` or `winsymlinks:nativestrict` already) on Windows (since opam 2.2)
```

kit-ty-kate:
ah i think you're right. That code should be in `OpamProcess.cygwin_create_process_env` instead

rjbou (approved):
lgtm, thanks!

PR #5794, opened 2024-01-22
Do not check for cppo in the configure script (not used directly anymore)

kit-ty-kate:
Cppo was used directly until #5498. It is still used by dependencies (namely `extlib`) however, so it should stay vendored.


rjbou (approved):
thanks!

Issue #5781, opened 2024-01-12
Cygwin internal installation should set `noacl` in `/etc/fstab` for `/cygdrive` mount


dra27:
By default, Cygwin attempts to preserve POSIX permissions on files in mount points. This generally results in chaos, and doesn't make sense for native Windows opam switches (we may be using Cygwin commands to build and install things, but the _result_ is native Windows).

It can also cause "interesting" errors. The issue is typically side-stepped by the default ACL on a user's profile directory (so the _default_ will _usually_ work). The following repro case demonstrates the issue, which can also be seen with opam 2.0.10 from "OCaml for Windows" ("OCaml for Windows" mostly side-steps the problem because the opam root is by default within a Cygwin `/home` directory).

```
C:\>opam config report
# opam config report
# opam-version         2.2.0~alpha4~dev (a358d94e412be5ae22054a4dc920cca2fda49821)
# self-upgrade         no
[ERROR] Opam has not been initialised, please run `opam init'
# read-state           OpamStd.OpamSys.Exit(50)

C:\>md Test

C:\>icacls Test
Test BUILTIN\Administrators:(I)(OI)(CI)(F)
     NT AUTHORITY\SYSTEM:(I)(OI)(CI)(F)
     BUILTIN\Users:(I)(OI)(CI)(RX)
     NT AUTHORITY\Authenticated Users:(I)(M)
     NT AUTHORITY\Authenticated Users:(I)(OI)(CI)(IO)(M)

Successfully processed 1 files; Failed processing 0 files
```

So `C:\Test` is a root directory with default permissions (Windows 10/11). Make that `OPAMROOT` and the init:

<details>
<summary>Environment setup</summary>

```
C:\>set OPAMROOT=C:\Test

C:\>opam init --cygwin-internal-install --no-git-location -a --bare sunset git+https://github.com/ocaml-opam/opam-repository-mingw.git#sunset
[WARNING] Flag --cygwin-internal-install is experimental, there is no guarantee that it will be kept; avoid using it in scripts.
[WARNING] Flag --no-git-location is experimental, there is no guarantee that it will be kept; avoid using it in scripts.
No configuration file found, using built-in defaults.

<><> Unix support infrastructure ><><><><><><><><><><><><><><><><><><><><><><><>
[NOTE] Configured with internal Cygwin install for depexts
Checking for available remotes: git, mercurial.
  - you won't be able to use rsync and local repositories unless you install the rsync command on your system.
[...]

PR #5796, opened 2024-01-22
Disable ACL in Cygwin to avoid permission missmatch errors

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/5781


dra27:
In the grand scheme of the amount of work going on in `opam init`, the time overhead is low, but I think it would be better to use OCaml to read the text file and update that line with a ocaml-re, rather than shelling out to `sed` (the cost of the Cygwin shell invocations in the `eval-variables` is perceptible, for example, when running `opam config list`)

dra27:
Looks good, thanks! I just pushed a tweak to open the file using native paths. @rjbou - LGTY?

PR #4816, opened 2021-08-30
Environments for PowerShell and Windows command line

jonahbeckford:
Splitting up https://github.com/ocaml/opam/pull/4813:

> For 0f0bd98 ("Show eval expression for PowerShell on native Windows"), opam 2.2 should finally see the activation of parent_putenv (and the opam-putenv auxiliary)... on Windows, there's no need to output shell instructions - on Windows, opam env can actually update the environment variables of the shell (see dra27@2934dc4#diff-e21e37a0339dbaafa95fd0b6d57221f656f18d0771c06f6f837bb98c3a71b91aR193). However, that doesn't necessarily preclude merging this first!

For continuity this initial PR is as-is from the original PR.

PR #5797, opened 2024-01-23
Fix shell detection on Windows when opam is called via Cygwin's /usr/bin/env even though cmd/powershell is used

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/5644

This line was added in https://github.com/ocaml/opam/pull/4816 but looking at the relevant discussion (https://github.com/ocaml/opam/pull/4816#discussion_r927799712) I’m not sure if the change makes much sense to me (but I’m not a Windows user so most of the discussions around Windows doesn’t make sense to me anyway :D )

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -72,6 +72,7 @@ users)
 
 ## Build
   * Do not check for cppo in the configure script (not used directly anymore since #5498) [#5794 @kit-ty-kate]
+  * Upgrade vendored cmdliner to 1.2.0 [#5797 @kit-ty-kate]
 
 ## Infrastructure
   * Fix depexts CI workflow and ensure all workflows run on master push [#5788 @dra27]
--- a/master_changes.md
+++ b/master_changes.md
@@ -57,6 +57,7 @@ users)
 ## Clean
 
 ## Env
+  * Fix shell detection on Windows when opam is called via Cygwin's /usr/bin/env even though cmd/powershell is used [#5797 @kit-ty-kate]
 
 ## Opamfile
 
--- a/src/core/opamStd.ml
+++ b/src/core/opamStd.ml
@@ -1093,7 +1093,6 @@ module OpamSys = struct
         Some (Accept (SH_pwsh Powershell))
       | "pwsh.exe" -> Some (Accept (SH_pwsh Powershell_pwsh))
       | "cmd.exe" -> Some (Accept SH_cmd)
-      | "env.exe" -> Some (Accept SH_sh)
       | "" -> None
       | name ->
         Option.map
--- a/src_ext/Makefile.sources
+++ b/src_ext/Makefile.sources
@@ -18,8 +18,8 @@ MD5_re = 68c427f8b55507f8b61c613ee6a5b3da
 
 $(call PKG_SAME,re)
 
-URL_cmdliner = https://erratique.ch/software/cmdliner/releases/cmdliner-1.1.1.tbz
-MD5_cmdliner = 26b9af55a7456290b710e1c96a0ca231
+URL_cmdliner = https://erratique.ch/software/cmdliner/releases/cmdliner-1.2.0.tbz
+MD5_cmdliner = b860881cc90c68b703dca0f35bdd4cdb
 
 $(call PKG_SAME,cmdliner)
 


dra27 (approved):
This line was there for Diskuv (I don't know why Diskuv invokes `opam` using MSYS2's `env`, rather than via `Unix.create_process_env` et al, but I expect there's a reason @jonahbeckford?)

The idea behind the original line was that if we see "env" in the process chain, opam assumes that its in Unix-land, and guesses `sh`. That said, it turns out the only situation where this really matters is when `env` is used from a _non-login_ terminal. In a login shell, Cygwin exports `SHELL` by default, so with no special handling for `env` (i.e. your change here), it still correctly identifies the shell. So with this PR, the time that it would be "wrong" is these steps:

- From Powershell, or Cmd, run MSYS2/Cygwin's bash _without_ `--login` (e.g. `C:\cygwin64\bin\bash`)
- `echo $SHELL` should give `/bin/bash`, but `cmd /c echo %SHELL%`  should give `%SHELL%` (from a login shell, you'd get `/bin/bash` for both)
- `opam env` should give sh-output
- `/usr/bin/env opam env` will now give cmd-style output
- `export SHELL`
- `/usr/bin/env opam env` will now give sh-style output

Given that this line _has_ caused the wrong shell for actual users (and while using the system for which the line was originally there anyway) coupled with what I think is a fairly contrived failure mode, I think we should delete the line 🙂

jonahbeckford:
> This line was there for Diskuv (I don't know why Diskuv invokes opam using MSYS2's env, rather than via Unix.create_process_env et al, but I expect there's a reason @jonahbeckford?)

I think because `opam --help` does not work on native Windows (no PAGER environment variable if I remember correctly; there may be more). I hate that hack, but `--help` was kinda important. And I also think that hack predated the https://github.com/dbuenzli/cmdliner/issues/166 hang issue.

kit-ty-kate:
> I think because `opam --help` does not work on native Windows (no PAGER environment variable if I remember correctly; there may be more). I hate that hack, but `--help` was kinda important. And I also think that hack predated the [dbuenzli/cmdliner#166](https://github.com/dbuenzli/cmdliner/issues/166) hang issue.

Does this mean upgrading to cmdliner 1.2.0 should fix together with this PR in the current state would fix both issues?

If so I'll upgrade the lower-bounds and vendor of cmdliner in this PR.

jonahbeckford:
There were at least two separate problems and only one has been fixed. (That linked issue expects "less" and "groff" to be in the PATH, which it is not in native Windows. That might be a third and fourth problem.)

Sorry but I don't do a full investigation to discover the extent of the problems with the pager support on Windows, and I didn't file issues, and do all the other time-consuming steps needed for a full solution. I just know that using "env" as a hack is sufficient for a working pager, and that cmdliner.1.2.0 (which is used in DkML) is insufficient.

Vendoring to 1.2.0 won't solve the full problem, but is better than nothing.
________________________________

kit-ty-kate:
> There were at least two separate problems and only one has been fixed. (That linked issue expects "less" and "groff" to be in the PATH, which it is not in native Windows. That might be a third and fourth problem.)

cc @dbuenzli

dbuenzli:
I don't think it does. If it can't find a pager it [outputs a plain text version](https://github.com/dbuenzli/cmdliner/blob/165450112eb13130e20bd34d4f5978fc45e00999/src/cmdliner_manpage.ml#L446-L462) (which does not need groff).

kit-ty-kate:
I've updated the vendored cmdliner but in any case, with or without, I was unable to reproduce your issue @jonahbeckford. `--help` works just fine for me in both cmd and powershell

jonahbeckford:
It works with a pager? How did a pager get into your path?

(Normal users should not have the “/usr/bin/less” pager in their PATH, nor “groff”, and it would be scrolling 10 pages of help plaintext all at once.)

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

kit-ty-kate:
> It works with a pager? How did a pager get into your path? (Normal users should not have the “/usr/bin/less” pager in their PATH, nor “groff”, and it would be scrolling 10 pages of help plaintext all at once.)

It does not. As far as I can see cmdliner does not support `PAGER` to be a path on Windows as it uses `where` which doesn't seem to support full-paths (whereas `command -v` does)

But I don't think it's too much of a problem in the short term. Is there a bug report somewhere on cmdliner for this?

jonahbeckford:
Okay.  I don't know of any feature request for pager support on native Windows for cmdliner.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

kit-ty-kate:
Then I opened one here: https://github.com/dbuenzli/cmdliner/issues/185

rjbou:
Shouldn't we update dependencies to cmdliner >= 1.2.0 ?

kit-ty-kate:
> Shouldn't we update dependencies to cmdliner >= 1.2.0 ?

I don't personally think it's necessary

rjbou (approved):
thanks!

Issue #5798, opened 2024-01-23
opam doesn't find any compiler for a switch creation on Windows


F-Loyer:
I have just installed opam-full-2.2.0-beta1 on Windows (`make cold; make cold-install; opam init git+https://github.com/ocaml-opam/opam-repository-mingw`).

But I can't create a switch:

```
$opam switch list-available
# Listing available compilers from repositories: default
# No matches found
```
Also:
```
$opam config report
# opam config report
# opam-version         2.2.0~beta1
# self-upgrade         no
# system               arch=x86_64 os=win32 os-distribution=cygwin os-version=10.0.22621
# solver               builtin-mccs+glpk
# install-criteria     -removed,-count[avoid-version,changed],-count[version-lag,request],-count[version-lag,changed],-count[missing-depexts,changed],-changed
# upgrade-criteria     -removed,-count[avoid-version,changed],-count[version-lag,solution],-count[missing-depexts,changed],-new
# jobs                 7
# current-switch       none set
```

kit-ty-kate:
Could you show us what’s the output of `opam option` and `opam repo -a` ?

F-Loyer:
```
$opam option

<><> Global configuration <><><><><><><><><><><><><><><><><><><><><><><><><>  🐫
archive-mirrors                {}
best-effort-prefix-criteria    {}
depext                         true
depext-bypass                  {}
depext-cannot-install          false
depext-run-installs            true
download-command               {}
download-jobs                  3
git-location                   {}
jobs                           {}
post-build-commands            {}
post-install-commands          {}
post-remove-commands           {}
post-session-commands          {}
pre-build-commands             {}
pre-install-commands           {}
pre-remove-commands            {}
pre-session-commands           {}
repository-validation-command  {}
solver                         {}
solver-criteria                {}
solver-fixup-criteria          {}
solver-upgrade-criteria        {}
swh-fallback                   true
sys-pkg-manager-cmd            ["cygwin" "C:\\Cygwin\\bin\\cygcheck.exe"]
wrap-build-commands            ["%{hooks}%/sandbox.sh" "build"] {os = "linux" | os = "macos"}
wrap-install-commands          ["%{hooks}%/sandbox.sh" "install"] {os = "linux" | os = "macos"}
wrap-remove-commands           ["%{hooks}%/sandbox.sh" "remove"] {os = "linux" | os = "macos"}

<><> Switch configuration <><><><><><><><><><><><><><><><><><><><><><><><><>  🐫
[...]

PR #5799, opened 2024-01-23
opam config report: add the list of all installed repositories

kit-ty-kate:
Looking at https://github.com/ocaml/opam/issues/5798 we can’t see the list of repositories in the output of `opam config report` because `repositories` only shows when the current switch is set. Having the list of all the repositories at all time is in my opinion very useful for this kind of issues.

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -1487,12 +1487,9 @@ let config cli =
             if n <> 0 then [Printf.sprintf "%d (%s)" n label]
             else [] in
           print "jobs" "%d" (Lazy.force OpamStateConfig.(!r.jobs));
-          match OpamStateConfig.get_switch_opt () with
-          | None -> print "current-switch" "%s" "none set"; `Ok ()
-          | Some switch ->
-            OpamSwitchState.with_ `Lock_none ~switch gt @@ fun state ->
-            print "repositories" "%s"
-              (let repos = state.switch_repos.repositories in
+          let print_repositories name rt =
+            print name "%s"
+              (let repos = rt.repositories in
                let default, nhttp, nlocal, nvcs =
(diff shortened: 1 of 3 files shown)


src/client/opamCommands.ml
```diff
@@ -1520,6 +1517,14 @@ let config cli =
...
                | None -> ""
               );
+          in
+          OpamRepositoryState.with_ `Lock_none gt @@ fun rt ->
+          print_repositories "all-repositories" rt;
+          match OpamStateConfig.get_switch_opt () with
+          | None -> print "current-switch" "%s" "none set"; `Ok ()
+          | Some switch ->
+            OpamSwitchState.with_ `Lock_none ~switch gt @@ fun state ->
+            print_repositories "repositories" state.switch_repos;
```
rjbou, line 1527: I don't understand the difference between `all-repositories` and `repositories` fields

kit-ty-kate: `repositories` only shows the list of repositories from the current switch.
`all-repositories` shows all of them (as in: `opam repository list -a`)

rjbou: Hmm... It is then the wrong value used: `rt` = `state.switch_repos`, to access to switch state repos, it is `state.switch_config.repos`

rjbou (changes requested):
Previous `repository` field was already listing all repositories, the pointed issue is that it is not shown if no switch is present.
We can change the behaviour (in another PR as it is not part of the fix), and have the 2 fields: one listing all repositories, and the other current switch repositories. In that case, it's better to minisie the change of behaviour by keeping `repositories` for all repositories and have a new one, for example `switch-repositories`, for switch ones.

kit-ty-kate:
Superseded by https://github.com/ocaml/opam/pull/6116

PR #5803, opened 2024-01-24
Shell integration: properly handle filenames starting with tilde

### LINKED ISSUE #5801: Double tilde when manually entering filename for shell configuration
## Issue description
Running a reinit and giving it a path starting with a tidle `~` bugs out.

## Details
```shell
> pwd
/home/me/foo
> opam init --reinit
[...]
Do you want opam to configure bash?
1. Yes, update ~/.profile
2. ...
3. ...
4. Specify another config file to update instead
> 4
Enter the name of the file to update:
```

- If you enter a simple filename `opam-thingy.sh`, it asks to update `~/opam-thingy.sh` :white_check_mark:
- If it's a qualified path like `/etc/foobar/opam-thingy.sh`, it asks to update exactly that :white_check_mark:
- If it's a relative path like `./opam-thingy.sh`, it expands the relative path: `~/foo/opam-thingy.sh` :white_check_mark:
- If it's a path starting with a tilde like `~/.camels/opam-thingy.sh`, it doubles the tilde: `~/~/.camels/opam-thingy.sh` :negative_squared_cross_mark:
- Bonus: if it's a path starting with "$HOME" it won't expand the variable: `~/'$HOME'/opam-thingy.sh`. But really, who would do that.

ElectreAAS:
Fixes #5801.

### DIFF
--- a/src/state/opamEnv.ml
+++ b/src/state/opamEnv.ml
@@ -1235,13 +1235,9 @@ let setup
           menu shell (OpamStd.Option.map OpamFilename.of_string (OpamStd.Sys.guess_dot_profile shell))
             default
         | `Change_file ->
-          let open OpamStd.Option.Op in
           let dot_profile =
-            (OpamConsole.read "Enter the name of the file to update:"
-             >>| (fun f ->
-                 if Filename.is_implicit f then Filename.concat (OpamStd.Sys.home ()) f
-                 else f)
-             >>| OpamFilename.of_string)
+            OpamStd.Option.map OpamFilename.of_string @@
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
One thing that the original code did and the new code doesn’t do is if you just type `.bashrc` for example, it used to default to `$HOME/.bashrc` instead of `$PWD/.bashrc`.

I personally prefer the code without such magic (your version) but it’s one thing to keep in mind. The new code has only been included in opam 2.2.0~alpha1 and up for now so it’s not a big deal to change its behaviour.

PR #5808, opened 2024-01-28
Install rsync to the internal Cygwin installation

dra27:
The VCS backends all require rsync, and this enables the local backend by default.


kit-ty-kate (approved):
Looks straightforward to me

rjbou (approved):

kit-ty-kate:
Thanks!

PR #5818, opened 2024-02-01
Fix the documentation of opam lint --warnings

kit-ty-kate:
The documentation for `opam lint --warnings=WARN` currently states that when WARN starts with a `+` the default warnings are disables, however this is also the case with `@` (added in #5652)

cc @Khady maybe that would have helped with your problem where it wasn't clear to you that `@` is a replacement to `+` and not an addition (not `-43+@21` but `-43@21`)

Or maybe should should have a description more like ocamlc's as it is basically the same thing:
```
  -w <list>  Enable or disable warnings according to <list>:
        +<spec>   enable warnings in <spec>
        -<spec>   disable warnings in <spec>
        @<spec>   enable warnings in <spec> and treat them as errors
     <spec> can be:
        <num>             a single warning number
        <num1>..<num2>    a range of consecutive warning numbers
        <letter>          a predefined set
     default setting is "+a-4-7-9-27-29-30-32..42-44-45-48-50-60-66..70"
```


Khady:
looks good to me, thanks. Now that I re-read the explanation carefully it's obvious that `@` is an "operator" like `+` or `-`. I wasn't careful enough. The description from ocamlc is maybe easier to understand for someone who just gives a very quick look to the doc.

rjbou (approved):

Issue #5792, opened 2024-01-22
install.sh doesn’t support platforms without sudo (e.g. doas)


kit-ty-kate:
Platforms such as OpenBSD or Alpine Linux are switching to doas instead of sudo by default. Currently the install script will fail if sudo is not found or if the current user is not root.

I think it would be nice to support `doas` or at least `su` so the script doesn’t fail

PR #5820, opened 2024-02-02
shell/install.sh: Add support for doas as an alternative to sudo

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/5792


rjbou (approved):
lgtm, thanks!

PR #5828, opened 2024-02-11
Disable Windows critical error handler

dra27:
When the WIndows loader cannot find DLLs, the default, very very legacy behaviour is to display a dialog such as this:

image

This dialog is problematic, as it's part of the loader, so the application itself is not able to handle this itself. It's doubly problematic for us, as it's blocking.

The convention is that _calling_ applications are expected to disable this error handler. New processes inherit this setting, which causes the loader to suppress the dialog box and instead return an exit code. This is documented best practice (see the notes for `SEM_FAILCRITICALERRORS` in [SetErrorMode on MSDN](https://learn.microsoft.com/en-us/windows/win32/api/errhandlingapi/nf-errhandlingapi-seterrormode)).

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -96,6 +96,7 @@ users)
 ## Internal
 
 ## Internal: Windows
+  * Ensure that the system critical error dialog is disabled when opam starts [#5828 @dra27]
 
 ## Test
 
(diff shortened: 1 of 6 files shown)


rjbou (approved):
As there is no simple way to have more information on the missing DLL, lgtm! thanks

dra27:
Let's get it merged before that screenshot of the dialog grows any further!

PR #5829, opened 2024-02-11
Various `eval-variables` fixes and `sys-ocaml-system`

dra27:
#3900 introduced `sys-ocaml-cc`, `sys-ocaml-libc` and `sys-ocaml-arch` which followed from some very early demonstrations I gave of the Windows opam packaging proposals at MSR Cambridge in 2015.

These variables have been quietly sat on new opam 2.1 installations since 2021. This PR addresses three issues:

- If you upgrade from a 2.0 root to a 2.1, you don't get them (you can see this in ocaml/opam Docker Hub images).
- At present, they use Unix sh commands, and they don't evaluate at all with the internal Cygwin installation on Windows
- A bug (which will be fixed in Cygwin 3.5.1) means that if the sh commands are used with an OCaml 5.1.1 compiler without the runtime DLLs (which is **VERY** easy to do), opam hangs on a popup.

I've therefore:
- Attempted some initial refactoring to ensure that the format upgrade re-applies eval-variables. As a side-note, I think the presence of `eval-variables` in `opamrc` is a mistake - they belong to the repository, not to the configuration, but we are where we are and that's something we can look at improving in opam 3.0
- Allowed the eval-variables to use the internal Cygwin installation (this is just a missing case)
- Further tweaked the opam 2.2 versions of these variables not to require Posix tools on Windows. A small side-effect is that because I use `-config-var`, system compilers must be 4.08+ on Windows, but I think this will be fine (it's not like we have old distributions to worry about here - system compilers will have been set-up on purpose to avoid recompilation).

Finally, while doing the Windows compiler packaging updates, the variables added in #3900 are working beautifully (as expected), but I propose adding `sys-ocaml-system` so that at some point similar support can be extended to Unix system compilers. `sys-ocaml-system` is simply the value of `system` from `ocamlc -config` and permits precise detection of the system compiler, especially in a mixed-architecture environment. It's in the same kind of area as #5587 - trying to probe the compiler itself, rather than the present very weak assumption of global variables correlating with all runnable compilers (remembering that an arm64 Windows machine is capable of running 4 different architectures...)


tests/reftests/opamroot-versions.test
```diff
@@ -2739,11 +2517,12 @@ roots: ["i-am-package.2" "i-am-sys-compiler.2"]
...
 RSTATE                          LOAD-REPOSITORY-STATE @ ${BASEDIR}/OPAM
-RSTATE                          No cache found
-RSTATE                          loaded opam files from repo default in 0.000s
+RSTATE                          Cache found
 
 <><> Updating repositories ><><><><><><><><><><><><><><><><><><><><><><><><><><>
-[default] no changes from file://${BASEDIR}/default
+Fatal error:
+Not_found
+# Return code 99 #
```
kit-ty-kate, line 2525: Something went wrong here

kit-ty-kate: I figured out, i ran the reftests without an empty `src/client/no-git-version`

dra27:
Thanks for sorting out the tests, @kit-ty-kate. I’m happy with the filtering out, unless it would be better just to sed off the values (i.e. be able to see that the variables were defined, but not worry about their differing values), @rjbou?

I think the only remaining thing is to put the various definitions into OpamEnv and reference them from there in both OpamFormatUpgrade and OpamInitDefaults.

kit-ty-kate:
> I think the only remaining thing is to put the various definitions into OpamEnv and reference them from there in both OpamFormatUpgrade and OpamInitDefaults.

How does https://github.com/ocaml/opam/pull/5829/commits/01576d068342f38bbbf68ea7d0be6e238119e6df sound?

tests/reftests/opamroot-versions.test
```diff
@@ -744,13 +744,12 @@ RSTATE                          LOAD-REPOSITORY-STATE @ ${BASEDIR}/OPAM
 RSTATE                          Cache found
 STATE                           LOAD-SWITCH-STATE @ sw-comp
 STATE                           Switch state loaded in 0.000s
-### opam-cat $OPAMROOT/config | '"${OPAMROOTVERSION}"' -> "current" | grep -v sys-pkg-manager-cmd | grep -v global-variables
+### opam-cat $OPAMROOT/config | '"${OPAMROOTVERSION}"' -> "current" | grep -v sys-pkg-manager-cmd | grep -v global-variables | grep -v eval-variables:
```
rjbou, line 747: I'm not sure about this `eval-variables` removal in the whole test, we may want to test that the `eval-variables` upgrade is well done, as it is the purpose of this test.

rjbou: Checked, it can be removed. I've added a specifically for those eval variables.

rjbou:
I've pushed all my review changes, separated in tiny commits. Once no more changes are needed, I'll do a final commit history cleaning.

kit-ty-kate (approved):

src/core/opamProcess.ml
```diff
@@ -14,12 +14,13 @@ let log ?level fmt =
...
 let default_env =
   let f () = lazy (
-    let env = OpamStd.Env.raw_env () in
     match OpamCoreConfig.(!r.cygbin) with
-    | Some cygbin -> OpamStd.Env.cyg_env ~env ~cygbin ~git_location:OpamCoreConfig.(!r.git_location)
-    | None -> env
+    | Some cygbin ->
+      OpamStd.Env.cyg_env ~cygbin
+        ~git_location:OpamCoreConfig.(!r.git_location) ()
+    | None -> OpamStd.Env.raw_env ()
```
dra27: I don't mind, but the thing I did like before is that it tried to make it clearer that we are always looking at `OpamStd.Env.raw_env ()`, it's just whether it's had `cygbin`/`git_location` added to it.

src/core/opamStd.ml
```diff
@@ -837,8 +837,7 @@ module Env = struct
     let lazy_env = lazy (to_list (raw_env ())) in
     fun () -> Lazy.force lazy_env
 
-  let cyg_env ~cygbin ~git_location =
-    let env = raw_env () in
+  let cyg_env ~env ~cygbin ~git_location =
```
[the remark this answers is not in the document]
dra27: My rationale was that the dev _should_ know 🙂 The exact environment in use at any given point can be quite tricky to determine, so I was figuring that being explicit was potentially better. I don't overly mind, though.

dra27 (commented):
Tests look good too, thanks!

rjbou:
> My rationale was that the dev should know 🙂 The exact environment in use at any given point can be quite tricky to determine, so I was figuring that being explicit was potentially better. I don't overly mind, though.

Fair enough!

I've updated the PR:
 * Reworked history to clean it for better readability
 * Added more information in master changes
 * Changed `OpamEnv.sys_ocaml_eval_variables` code to ensure that unix and win variables are defined.

dra27:
It's fine thing to have a single PR number with all three of us as authors against it 🙂

Approving the bits I didn't write, thank you!

kit-ty-kate:
Thanks!

PR #5830, opened 2024-02-11
Tweaks to internal Cygwin setup

dra27:
Related to #5793, this adds `--symlink-type native` to the Cygwin setup invocation if the user has the ability to create them. This is more to disable the use of WSL symlinks than to actually enable any additional binaries (Cygwin names binaries `.exe`, but it doesn't do that for a symlink to an executable, so they're not any more available than they were before).

We can save a small amount of time and a potential error message by passing `--no-version-check` - that simply tells Cygwin not to report if the version of setup is older than the latest reported in `setup.ini`.

Hot off the presses in Setup 2.929 is an option to reduce the noise of the installation. We should definitely use the `noinput` version at all times, but the `hidden` version is nice as it suppresses the GUI completely. The caveat, and why I'm keeping this as draft for now, is that there is _no_ UI for Cygwin at all, then, so we need to display some kind of "Please wait..." indicator and possibly spinner ourselves.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -66,6 +66,7 @@ users)
 ## Opamfile
 
 ## External dependencies
+  * Pass --symlink-type native to Cygwin setup if symlinks are available [#5830 @dra27]
 
 ## Format upgrade
   * Handle init OCaml `sys-ocaml-*` eval variables during format upgrade from 2.0 -> 2.1 -> 2.2 [#5829 @dra27]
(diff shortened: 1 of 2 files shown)


dra27:
Right, sorting out the work for `hidden` can wait - we should merge the interim version with `noinput`

rjbou (approved):
thanks!

kit-ty-kate:
Thanks!

PR #5831, opened 2024-02-11
Recommend Developer Mode on Windows

dra27:
Apropos note in #5782. The motivation behind this is to have symlinks available as an optimisation (for example, to allow caches, compiler installations, etc. to use `ln` instead of `cp` to save space, just as on Unix, but not to be a _requirement_ for any packages). There are other ways to enable this, but "Developer Mode" is by far the simplest, and is recommended (or even required) by Microsoft's own SDKs.


kit-ty-kate (approved):

kit-ty-kate:
Thanks!

PR #5832, opened 2024-02-11
Resolve the Cygwin bin directory above Windows system32 directory

dra27:
`OpamProcess.cygwin_create_process_env` contains some ad hoc logic which ensures that the Cygwin `bin` directory appears before the Windows `system32` directory in `Path` and adjusts it if necessary. This allows the user to put `c:\cygwin64\bin` routinely at the _end_ of `Path` so that Cygwin utilities don't shadow anything (including, importantly, the Microsoft linker `link.exe` if using the MSVC port of OCaml). However, when invoking a Cygwin executable directly, this results in `/usr/bin` appearing at the end of `PATH`, which is confusing - especially if the executable (or script) expects to invoke `bash`. Hence the workaround.

There are, however, more things than just `bash.exe` where Cygwin executables may want their own version to apply when executing and it's possible for other directories to shadow Cygwin as well.

Enter #5545, and opam's internally managed Cygwin. This inserts a `PATH =+ ` update to append the Cygwin `bin` directory to the end when executing `build` and `install` commands, etc. This environment update interacts surprisingly with `OpamProcess.cygwin_create_process_env`. `["bash" "-c" "..."]` "executes" using WSL yet `["sh" "-c" "bash -c \"foo\""] executes using Cygwin's `bash`, for example. The resolution problems can be clearly seen using `where` - `["where" "bash"]` shows `C:\Windows\system32\bash.exe` being first but `["sh" "-c" "where bash"]` shows Cygwin's `bash.exe` being first.

This PR generalises the adjustment being made to `Path` by introducing a new kind of environment update. This update is only available internally via the constructor `Cygwin` (i.e. it cannot appear in `opam` files, nor can it be written/read from `environment` files). It places the directory given as far down the list of directories in `Path` so that `bash.exe`, `tar.exe` and `sort.exe` (and `git.exe`, if Cygwin's `git` is being used) still resolve to the Cygwin copy. In practice, this means that the `bin` directory will never be placed after `C:\Window\system32` (as before) but also means, for example, that it would not be moved after a Scoop `bin` directory (which may contain `bash.exe` or `git.exe`), etc.

With this slightly more principled mechanism in place, the ad hoc adjustment in `OpamProcess.cygwin_create_process_env` is removed (i.e. if the user puts `C:\cygwin64\bin` into `Path`, they are now responsible for putting it in the right place; when opam puts its own `.cygwin\root\bin` diectory into `Path`, it ensures that it puts it into the correct place.)


src/format/dune
```diff
@@ -12,3 +13,9 @@
   (wrapped     false))
 
 (ocamllex opamLineLexer opamInterpLexer)
+
+(test
```
kit-ty-kate, line 17: shouldn't this live in somewhere under the `tests` directory?

dra27:
First two commits are optional (and could be moved elsewhere). Could do with some more comments, possibly a test. I've been testing with this:
```
opam-version: "2.0"
synopsis: "Show the environment"
description: "Yet another environment debug package"
maintainer: "David Allsopp <david.allsopp@metastack.com>"
authors: "David Allsopp"
license: "ISC"
homepage: "https://opam.ocaml.org"
bug-reports: "https://github.com/ocaml/opam/issues"
flags: conf
build: [
  ["cmd" "/c" "echo %%PATH%%"] # *** 1 ***
  ["cmd" "/c" "echo ============================================"]
  ["sh" "-c" "env | grep '^PATH='"] # *** 2 ***
  ["cmd" "/c" "echo ============================================"]
  ["cmd" "/c" "env | findstr /R ^^PATH="] # *** 3 ****
  ["cmd" "/c" "echo ============================================"]
  ["where" "bash"] # *** 4 ***
]
```
but that's not readily convertable to a test script. Before this PR:
1. Has cygbin at the end of PATH
2. Has cygbin just before `C:\Windows\system32\` in PATH
3. Is the same as 1 (so cygbin at the end of PATH)
4. Has C:\Windows\system32\bash.exe **first** if you have WSL installed 💀

With this PR, 1, 2 and 3 _all_ have cygbin just before `C:\Windows\system32\` and 4 has `bash.exe` in cygbin first, regardless of whether you have WSL installed 😇

kit-ty-kate (commented):
Aside from the question of where the `TypeGymnastics` module/test should live (I would personally prefer for it to be in a new `tests/libtests` directory or something like that), LGTM.

dra27:
Yes, I think in my mind the "type gymnastics" test was going to be part of the (dev) build to be sure at all times. Then I used a `test` stanza and didn't think just to move it to tests! I'll do that 🙂

kit-ty-kate (approved):

src/format/opamFile.mli
```diff
@@ -1026,18 +1034,32 @@ module Repos_config: sig
 end
 
 module Switch_config: sig
-  type t = {
+  type t = private {
```
rjbou, line 1037: This is not correlated to this PR, it should go on its own PR.

tests/lib/dune
```diff
@@ -0,0 +1,10 @@
+(test
+  (name      patcher)
```
rjbou, line 2: Patcher move is not directly correlated to this PR, but as it is needed for addition of type gymnastics, to not delay (have a PR from which this one depends) we could keep it.

tests/lib/dune
```diff
@@ -0,0 +1,10 @@
...
+  (modules   patcher)
+  (libraries opam-core))
+
+(test
```
rjbou, line 6: Should we add in CI at least a run of it ? Maybe not in all jobs

rjbou (commented):
On the idea and the code, LGTM! I've put some comments in the review.
and... what a type gymnastic xD

rjbou:
Tested locally, it works well. For a CI test, it needs its proper jobs, to have an environment 100% controlled (git, bash, cygwin, etc.), it can't be on reftests.

src/format/opamTypes.mli
```diff
@@ -435,10 +435,22 @@ type _ separator_path_format =
       * (path_format * filter) generic_formula
       -> spf_unresolved separator_path_format
 
+type euok_writeable = [ `writeable ]
```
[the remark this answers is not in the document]
dra27: So are `spf_resolved` and `spf_unresolved` 😉

src/state/opamEnv.ml
```diff
@@ -241,10 +241,29 @@ let apply_op_zip ~sepfmt op arg (rl1,l2 as zip) =
       if eqcol then l@[""], [arg] else l, [""; arg]
     | l -> l, [arg]
   in
+  let cygwin path =
+    let contains_in dir item =
+      Sys.file_exists (Filename.concat dir item)
+    in
+    let shadow_list =
+      List.filter (contains_in arg) ["bash.exe"; "tar.exe"; "git.exe"]
```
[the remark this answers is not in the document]
dra27: I don't think it should be - that list is a decision of opam's (i.e. opam-state), not part of a standard library (`OpamStd` in opam-core).

tests/lib/dune
dra27: I don't follow - this test is run in CI (as part of `make tests`) isn't it?

src/state/opamEnv.mli
```diff
@@ -29,26 +29,7 @@ val get_full:
...
-    With [base], apply the modifications to the specified base environment *)
-val get_opam_raw:
-  set_opamroot:bool -> set_opamswitch:bool -> ?base:env ->
-  force_path:bool ->
-  dirname -> switch -> env
-
-(** Like [get_opam_raw], but returns the list of updates instead of the new
-    environment. *)
+(** Returns the list of updates from the switch's cache file. *)
 val get_opam_raw_updates:
```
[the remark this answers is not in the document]
dra27: Commit removed

src/format/opamFile.mli
dra27: Commit removed (and conflict resolved)

tests/lib/dune
dra27: As a general thing, I think having a separate commit is enough - I don't think we'd need to require that to go to a separate PR, unless the reviewing's become too large?

src/client/opamConfigCommand.ml
```diff
@@ -210,6 +210,9 @@ let check_writeable l =
     match envu_op with
     | Eq | PlusEq | EqPlus | ColonEq | EqColon | EqPlusEq as envu_op ->
       {update with envu_op}
+    | _ ->
```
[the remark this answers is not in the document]
dra27: Oops 🙂

dra27:
For the testing, my feeling was indeed that to do it via reftests would be very contrived. What we might be able to do in CI in general is to put a "poisoned" `tar.exe` in PATH in a directory _after_ the Visual Studio bin directories. The MSVC build would then fairly naturally test this - if opam puts Cygwin's `bin` _before_ the Visual Studio directories, `link` would fail and if it puts it _after_ the poisoned directory, extraction will fail (so a canary-style test rather than a unit test)

tests/lib/dune
rjbou: True, I had in mind that it is `make reftests` that is run in CI.

src/format/opamTypesBase.ml
```diff
@@ -264,6 +264,14 @@ let env_update_unresolved ?comment:envu_comment ?rewrite
         rewrite;
   }
 
+external op_of_raw:
+  OpamParserTypes.FullPos.env_update_op_kind
+    -> euok_writeable env_update_op_kind = "%identity"
+external raw_of_op:
+  euok_writeable env_update_op_kind
+    -> OpamParserTypes.FullPos.env_update_op_kind = "%identity"
+(* cf. typeGymnastics.ml *)
```
rjbou, line 273:
```suggestion
(* cf. tests/lib/typeGymnastics.ml *)
```

rjbou (approved):
LGTM!

kit-ty-kate:
Thanks!!

PR #5837, opened 2024-02-14
Two (more) environment fixes

dra27:
Two fixes here, both relatively subtle:

- If the environment contained `FOO=bar`, the update `FOO += ""` would cause opam to execute _without_ `FOO` in the environment
- A small regression in #5356 - in the refactoring for case-preserving updates, a boolean check got accidentally inverted

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -60,6 +60,7 @@ users)
 ## Env
   * Fix shell detection on Windows when opam is called via Cygwin's /usr/bin/env even though cmd/powershell is used [#5797 @kit-ty-kate]
   * Fix incorrect deduplication of environment variables on update. Effect was that FOO += "" would occlude the value of FOO in the environment [#5837 @dra27]
+  * Fix regression from #5356 on the detection of out-of-date environment variables. As part of a refactoring, a filter predicate got inverted [#5837 @dra27]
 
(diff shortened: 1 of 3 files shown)


rjbou (commented):
It is missing a test for the case *add an empty string to an undefined variable*. Nothing changes in the environment as the variable is not defined beforehand.
```diff
diff --git a/tests/reftests/env.test b/tests/reftests/env.test
index 0a8a7a210..781341de3 100644
--- a/tests/reftests/env.test
+++ b/tests/reftests/env.test
@@ -80,6 +80,7 @@ setenv: [
   [ NV_VARS3 := ""    ]
   [ NV_VARS3 := "foo" ]
   [ NV_VARS4  = ""    ]
+  [ NV_VARS5 += ""    ] # undefined
 ]
 flags: compiler
 ### opam switch create emptyvar nv
```
An update of the documentation to be more precise, and lgtm!

dra27:
I think the manual’s correct, isn’t it? The issue was that opam wasn’t previously doing what it was supposed to (by any definition of `+=`)

rjbou (approved):
thanks!

kit-ty-kate:
Thanks

PR #5840, opened 2024-02-15
Hijack the %{?val_if_true:val_if_false}% syntax to support extending the variables of packages with + in their name

kit-ty-kate:
~Per our own documentation, '+' should not be a valid character for packages. However packages like conf-g++ or conf-c++ now exist in the wild. This solution is a compromise in hope for {name1+name2+name3:var} to be parsable again~

Related to https://github.com/ocaml/opam-file-format/issues/59

### DIFF
--- a/src/format/opamTypesBase.ml
+++ b/src/format/opamTypesBase.ml
@@ -122,14 +122,19 @@ let string_of_filter_ident (pkgs,var,converter) =
    | Some (it,ifu) -> "?"^it^":"^ifu
    | None -> "")
 
-let filter_ident_of_string_t ~interpolation s =
+exception Parse_variable of string * string
+
+let filter_ident_of_string_t ~interpolation ?(accept=true) s =
   match OpamStd.String.rcut_at s ':' with
   | None -> [], OpamVariable.of_string s, None
   | Some (p,last) ->
     let get_names s =
-      List.map
-        (function "_" -> None | s -> Some (OpamPackage.Name.of_string s))
-        (OpamStd.String.split s '+')
+      List.filter_map
+        (function
+          | "_" -> Some None
+          | "" -> if accept then None else raise (Parse_variable (s, last))
+          | s -> Some ( Some (OpamPackage.Name.of_string s)))
+        (OpamStd.String.split_delim s '+')
     in
     match OpamStd.String.rcut_at p '?' with
     | None ->
@@ -155,10 +160,10 @@ let filter_ident_of_string_t ~interpolation s =
         get_names packages, OpamVariable.of_string var, converter
 
 let filter_ident_of_string s =
-  filter_ident_of_string_t ~interpolation:false s
+  filter_ident_of_string_t ~interpolation:false ~accept:true s
 
-let filter_ident_of_string_interp s =
-  filter_ident_of_string_t ~interpolation:true s
+let filter_ident_of_string_interp ?accept s =
+  filter_ident_of_string_t ~interpolation:true ?accept s
 
 let all_package_flags = [
   Pkgflag_LightUninstall;
--- a/src/format/opamTypesBase.ml
+++ b/src/format/opamTypesBase.ml
@@ -122,7 +122,7 @@ let string_of_filter_ident (pkgs,var,converter) =
    | Some (it,ifu) -> "?"^it^":"^ifu
    | None -> "")
 
-let filter_ident_of_string s =
+let filter_ident_of_string_t ~interpolation s =
   match OpamStd.String.rcut_at s ':' with
   | None -> [], OpamVariable.of_string s, None
   | Some (p,last) ->
@@ -134,6 +134,18 @@ let filter_ident_of_string s =
     match OpamStd.String.rcut_at p '?' with
     | None ->
       get_names p, OpamVariable.of_string last, None
+    | Some ("",val_if_true) when interpolation ->
+      (* TODO: Remove in opam 3.0.
+         Hack added in opam 2.2. This is a compatible syntax with opam 2.0 and
+         2.1 but supports + in the package name.  See
+         https://github.com/ocaml/opam-file-format/issues/59 *)
+      (match OpamStd.String.rcut_at val_if_true ':' with
+       | None ->
+         (* behaviour from opam 2.0 and 2.1 *)
+         [], OpamVariable.of_string last, None
+       | Some (p, var) ->
+         [Some (OpamPackage.Name.of_string p)],
+         OpamVariable.of_string var, None)
     | Some (p,val_if_true) ->
       let converter = Some (val_if_true, last) in
       match OpamStd.String.rcut_at p ':' with
@@ -141,6 +153,12 @@ let filter_ident_of_string s =
         [], OpamVariable.of_string p, converter
       | Some (packages,var) ->
         get_names packages, OpamVariable.of_string var, converter
+
+let filter_ident_of_string s =
+  filter_ident_of_string_t ~interpolation:false s
+
+let filter_ident_of_string_interp s =
+  filter_ident_of_string_t ~interpolation:true s
 
 let all_package_flags = [
   Pkgflag_LightUninstall;
--- a/tests/reftests/dune.inc
+++ b/tests/reftests/dune.inc
@@ -927,6 +927,24 @@
    (run ./run.exe %{exe:../../src/client/opamMain.exe.exe} %{dep:lock.test} %{read-lines:testing-env}))))
 
 (rule
+ (alias reftest-opam-pkgvar-with-plus)
+ (action
+  (diff opam-pkgvar-with-plus.test opam-pkgvar-with-plus.out)))
+
+(alias
+ (name reftest)
+ (deps (alias reftest-opam-pkgvar-with-plus)))
+
+(rule
+ (targets opam-pkgvar-with-plus.out)
+ (deps root-N0REP0)
+ (package opam)
+ (action
+  (with-stdout-to
+   %{targets}
+   (run ./run.exe %{exe:../../src/client/opamMain.exe.exe} %{dep:opam-pkgvar-with-plus.test} %{read-lines:testing-env}))))
+
+(rule
  (alias reftest-opamroot-versions)
  (action
   (diff opamroot-versions.test opamroot-versions.out)))
--- a/doc/pages/Manual.md
+++ b/doc/pages/Manual.md
@@ -247,7 +247,7 @@ and _options_:
 <section>       ::= <ident> [ <string> ] "{" <file-contents> "}"
 <ident>         ::= { <identchar> }* <letter> { <identchar> }*
 <varident>      ::= [ ( <ident> | "_" ) { "+" ( <ident> | "_" ) }* : ] <ident>
-<identchar>     ::= <letter> | <digit>  | "_" | "-"
+<identchar>     ::= <letter> | <digit>  | "_" | "-" | "+"
 <letter>        ::= "a".."z" | "A".."Z"
 <digit>         ::= "0".."9"
 <value>         ::= <bool> | <int> | <string> | <ident> | <varident> | <operator> | <list> | <option> | "(" <value> ")"
@@ -383,7 +383,14 @@ insert different strings depending on the boolean value of a variable.
 Additionally, boolean package variables may be combined using the following
 form: `name1+name2+name3:var` is the conjunction of `var` for each of `name1`,
 `name2` and `name3`, _i.e_ it is equivalent to `name1:var & name2:var &
-name3:var`
+name3:var`.
+
+**Warning**: if the package name contains a `+` character (e.g. `conf-g++`),
+their variables may only be accessed using opam 2.2 via string interpolation,
+with the following syntax:
+```
+"%{?conf-g++:your-variable:}%"
+```
 
 #### Scopes
 
--- a/master_changes.md
+++ b/master_changes.md
@@ -62,6 +62,7 @@ users)
 ## Source
 
 ## Lint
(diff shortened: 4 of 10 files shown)


kit-ty-kate:
Instead of this, which is not backward compatible with opam 2.0 and 2.1, I would suggest hijacking the `"%{?pk+*g1:var:}%"` (one or more `+` in a string expension, with `?` at the beginning and `:` at the end) syntax and adding a lint check when this syntax is present, that the package is marked with at least `available: opam-version >= "2.2.0~beta2"` (it should only be a warning, not an error as I expect cases where @dra27 would like to use this syntax in Windows context where the package itself should be compatible with opam < 2.2 but the value should be only reachable with opam 2.2)

See the definition for the details:

https://github.com/ocaml/opam/blob/87a88a52e4175aee874b3dbd9f4d43c4e30e1479/src/format/opamTypesBase.ml#L125

With 2.0 and 2.1 the syntax will output the empty string because the `variable` `<empty string>` is undefined and thus will fall into the else-or-undefined clause (the `:` at the end) and output `<empty string>`.
In 2.2 we can hijack this and take the core of the `true` case and reparse it without making `+` a separator. The only thing that we loose is be able to express the `name1+name2+name3:var` form but this is only a syntactic sugar for `name1:var & name2:var & name3:var` anyway and the `name1+name2+name3:enable` syntactic sugar which we can also replace by a more explicit version if needed.

dra27:
The workaround syntax sounds good to me!

dra27 (commented):
Looks good, thanks, apart from what I think is a missing colon in the to_string function? This is working for the `conf-g++` adaptations I'd made (wohoo - the package is no longer called `conf-mingw-w64-gplusplus-x86_64`!)

For the opam 3.0 reminder, another possibility might be to start the function in `OpamFormatUpgrade`, but the TODO comment also works 🙂

> The only thing that we loose is be able to express the `name1+name2+name3:var` form but this is only a syntactic sugar for `name1:var & name2:var & name3:var` anyway and the `name1+name2+name3:enable` syntactic sugar which we can also replace by a more explicit version if needed.
EDIT: we also can't use the ternary `var?true:false-or-undef` inside the new clause due to the parsing rule but this is the same thing (can be desugared)

I just want to check I'm properly understanding - I think we're not _losing_ the `pkg+pkg+var` or ternary `var?true:false-or-undef` here, because they don't presently work if pkg contains a `+`? It's that we're "only" gaining the ability to retrieve variable values as a string?

src/format/opamTypesBase.ml
```diff
@@ -115,12 +115,20 @@ let env_array l =
...
-  (match converter with
-   | Some (it,ifu) -> "?"^it^":"^ifu
-   | None -> "")
+  (* TODO: Remove in opam 3.0.
+     Hack added in opam 2.2. This is a compatible syntax with opam 2.0 and 2.1
+     but supports + in the package name.
+     See https://github.com/ocaml/opam-file-format/issues/59 *)
+  match pkgs, converter with
+  | [Some pkg], None when OpamStd.String.contains_char (OpamPackage.Name.to_string pkg) '+' ->
+    "?"^OpamPackage.Name.to_string pkg^":"^OpamVariable.to_string var
```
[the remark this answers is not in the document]
kit-ty-kate: Good catch. Fixed.

kit-ty-kate:
> I just want to check I'm properly understanding - I think we're not _losing_ the `pkg+pkg+var` or ternary `var?true:false-or-undef` here, because they don't presently work if pkg contains a `+`? It's that we're "only" gaining the ability to retrieve variable values as a string?

yes. What i meant was that we can't use those syntaxes inside the new one (e.g. `?pkg1+pkg2:var:` or `?pkg:var?x:y:`) because it would clash with the way `+` and `?` are parsed in this new context. So yes, there is no loss of ability, this PR only adds a way to refer to variables of packages with `+` in their name.

tests/reftests/var-syntax.test
```diff
@@ -0,0 +1,208 @@
...
+### <pkg:var-with-plus.1>
+opam-version: "2.0"
+build: [
+  ["echo" "%{?pkg++:installed:}%"]
+  ["echo" "%{?pkg+:installed:}%"]
+  ["echo" "%{?pkg+a:installed:}%"]
+  ["echo" "%{?pkg++a:installed:}%"]
+  ["echo" "%{?pkg1+pkg2:installed:}%"]
+  ["echo" "%{pkg1+pkg2:installed}%"]
+]
```
rjbou, line 23:
```suggestion
build: [
  ["echo" "'?pkg++:installed:'"     "%{?pkg++:installed:}%"]
  ["echo" "'?pkg+:installed:'"      "%{?pkg+:installed:}%"]
  ["echo" "'?pkg+a:installed:'"     "%{?pkg+a:installed:}%"]
  ["echo" "'?pkg++a:installed:'"    "%{?pkg++a:installed:}%"]
  ["echo" "'?pkg1+pkg2:installed:'" "%{?pkg1+pkg2:installed:}%"]
  ["echo" "'pkg1+pkg2:installed'"   "%{pkg1+pkg2:installed}%"]
]
```
It permits to locate directly what is the true/false echo

tests/reftests/var-syntax.test
```diff
@@ -0,0 +1,208 @@
...
+  - install var-with-plus 1
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+Processing  2/3: [var-with-plus: echo false]
++ echo "false" (CWD=${BASEDIR}/OPAM/test/.opam-switch/build/var-with-plus.1)
++ echo "false" (CWD=${BASEDIR}/OPAM/test/.opam-switch/build/var-with-plus.1)
++ echo "false" (CWD=${BASEDIR}/OPAM/test/.opam-switch/build/var-with-plus.1)
++ echo "false" (CWD=${BASEDIR}/OPAM/test/.opam-switch/build/var-with-plus.1)
++ echo "false" (CWD=${BASEDIR}/OPAM/test/.opam-switch/build/var-with-plus.1)
++ echo "false" (CWD=${BASEDIR}/OPAM/test/.opam-switch/build/var-with-plus.1)
```
rjbou, line 47: If you don't want to have all the echo for each run, it is possible to define a package for each `opam install pkg+`
```
### <pkg:var-with-plus.2>
opam-version: "2.0"
build: ["echo" "'?pkg++:installed:'" "%{?pkg++:installed:}%"]
depopts: [ "pkg++" ]
### opam install pkg++
### opam install --verbose var-with-plus.2 | sed-cmd echo
```
it'll also remove the unordered outputs

rjbou: Even defining each package at that moment :
```
### : End with ++ :
### <pkg:pkg++.1>
opam-version: "2.0"
### <pkg:var-with-plus.2>
opam-version: "2.0"
build: ["echo" "'?pkg++:installed:'" "%{?pkg++:installed:}%"]
depopts: [ "pkg++" ]
### opam install pkg++
### opam install --verbose var-with-plus.2 | sed-cmd echo

File: tests/reftests/dune.inc [hunk 4]
```diff
@@ -1427,6 +1427,24 @@
    %{targets}
    (run ./run.exe %{exe:../../src/client/opamMain.exe.exe} %{dep:var-option.test} %{read-lines:testing-env}))))
 
+(rule
+ (alias reftest-var-syntax)
```
rjbou, line 1431: The name is not explicit, maybe `opam-var-++-syntax` or `opam-var-filter-syntax`

tests/reftests/var-syntax.test
rjbou: My previous comment about labeling echoes is irrelevant if those one are implemented.

tests/reftests/var-syntax.test
```diff
@@ -0,0 +1,208 @@
...
++ echo "true" (CWD=${BASEDIR}/OPAM/test/.opam-switch/build/var-with-plus.1)
++ echo "true" (CWD=${BASEDIR}/OPAM/test/.opam-switch/build/var-with-plus.1)
+Processing  2/3: [var-with-plus: echo false]
++ echo "false" (CWD=${BASEDIR}/OPAM/test/.opam-switch/build/var-with-plus.1)
+Processing  2/3: [var-with-plus: echo true]
++ echo "true" (CWD=${BASEDIR}/OPAM/test/.opam-switch/build/var-with-plus.1)
+-> compiled  var-with-plus.1
+Installing var-with-plus.1.
+-> installed var-with-plus.1
+Done.
```
rjbou, line 208: It also needs a test for printing this new syntax, maybe with opam show?

rjbou (commented):
We also need to update the doc for that particular case.
Should we add a lint to warn that it shouldn't be acceptable to have `++` at the end of a package name (even if supported by this hack)? Users installing already present packages won't see it until they pin it, and it'll minimise new packages addition.

tests/reftests/dune.inc
kit-ty-kate: opam-var-filter-syntax then

src/format/opamTypesBase.ml
```diff
@@ -115,12 +115,20 @@ let env_array l =
...
-  (match converter with
-   | Some (it,ifu) -> "?"^it^":"^ifu
-   | None -> "")
+  (* TODO: Remove in opam 3.0.
+     Hack added in opam 2.2. This is a compatible syntax with opam 2.0 and 2.1
+     but supports + in the package name.
+     See https://github.com/ocaml/opam-file-format/issues/59 *)
+  match pkgs, converter with
+  | [Some pkg], None when OpamStd.String.contains_char (OpamPackage.Name.to_string pkg) '+' ->
+    "?"^OpamPackage.Name.to_string pkg^":"^OpamVariable.to_string var^":"
```
rjbou, line 124: Some indentation
```suggestion
  | [Some pkg], None
    when OpamStd.String.contains_char (OpamPackage.Name.to_string pkg) '+' ->
    "?"^OpamPackage.Name.to_string pkg^":"^OpamVariable.to_string var^":"
```

src/format/opamTypesBase.ml
```diff
@@ -134,6 +142,16 @@ let filter_ident_of_string s =
...
+    | Some ("",val_if_true) ->
+      (* TODO: Remove in opam 3.0.
+         Hack added in opam 2.2. This is a compatible syntax with opam 2.0 and 2.1
+         but supports + in the package name.
+         See https://github.com/ocaml/opam-file-format/issues/59 *)
+      (match OpamStd.String.rcut_at val_if_true ':' with
+       | None ->
+         [], OpamVariable.of_string last, None (* behaviour from opam 2.0 and 2.1 *)
+       | Some (p, var) ->
+         [Some (OpamPackage.Name.of_string p)], OpamVariable.of_string var, None)
```
rjbou, line 154: Some indentation
```suggestion
    | Some ("",val_if_true) ->
      (* TODO: Remove in opam 3.0.
         Hack added in opam 2.2. This is a compatible syntax with opam 2.0 and
         2.1 but supports + in the package name.  See
         https://github.com/ocaml/opam-file-format/issues/59 *)
      (match OpamStd.String.rcut_at val_if_true ':' with
       | None ->
         (* behaviour from opam 2.0 and 2.1 *)
         [], OpamVariable.of_string last, None
       | Some (p, var) ->
         [Some (OpamPackage.Name.of_string p)],
         OpamVariable.of_string var, None)
```

kit-ty-kate:
> Should we add a lint to warn that it shouldn't be acceptable to have `++` at the end of a package name (even if supported by this hack)?

I don't think so. This PR basically officializes that we do accept packages with pluses in their names.

> Users installing already present packages won't see it until they pin it, and it'll minimise new packages addition.

I'm not sure to understand this.

rjbou:
>> Should we add a lint to warn that it shouldn't be acceptable to have ++ at the end of a package name (even if supported by this hack)?

> I don't think so. This PR basically officializes that we do accept packages with pluses in their names.

Having a closer look, I'm afraid that the current PR is not enough to have a full support of `pkg++` packages, it needs parser change too: `{ pkg++:installed }` and its variations is not accepted by the parser, it is only available via string interpolation, in the form `"%{?pkg++:installed:}%"`.
Either we have a full support, and then it needs a parser change, or it is a workaround to handle existing ones and then we need to propagate that (lint, manual).
In the second case (workaround), the changes then need to be in a specific function that is called by interpolation functions, only for the transformation from a string (no need to print, as the more global string will be reprinted as is). Current changes are misleading as it affects also `OpamFormat.filter_ident` parser/printer that is used for `{ pkg:installed }` form.


>> Users installing already present packages won't see it until they pin it, and it'll minimise new packages addition.

> I'm not sure to understand this.

When a package is installed from repository, there is no display of lint warning/errors. So having a lint won't affect people installing `conf-g++`, but it'll at least warn people creating packages and opam repo maintainers about packages ending with `++`.

rjbou:
I've updated the PR, details are in commit messages. Roughly there is
* addition of a lint when a package variable in a string have its package name containing several `+`, warn to use the new syntax (a single `+` is indistinguishable from the `name1+name2:var` form)
* Some code transformation to access functions to permit the lint
* Split `filter_ident_of_string` to have the good ones used when needed (no hijack for not in string variables)
* Remove the `to_string` modification as it is not affected by the hijack

As usual, there is a lot of tiny commits to show/explain the diff with current PR, if accepted, i'll clean commits history in order to have it in a mergeable state.

dra27 (approved):
LGTM, thanks! I also double-checked that the mingw-w64-shims (which is now in opam-repository) is still working with it.

kit-ty-kate:
We're missing the addition of the new syntax to the documentation (with the appropriate warnings ofc). I'll add it and rebase the branch

rjbou (approved):
Another 2 hands PR ^^ Thanks!

kit-ty-kate:
^^ thanks!

Issue #5804, opened 2024-01-24
opam init create a weird .bash_profile from a Cygwin point of view


F-Loyer:
When trying an `opam init` from a Cygwin bash, it appends to my `.bash_profile`:

```
# BEGIN opam configuration
# This is useful if you're using opam as it adds:
#   - the correct directories to the PATH
#   - auto-completion for the opam binary
# This section can be safely removed at any time if needed.
test -r C:\Users\frede\AppData\Local\opam\opam-init\init.sh && . C:\Users\frede\AppData\Local\opam\opam-init\init.sh > /dev/null 2> /dev/null || true
# END opam configuration
```
Unfortunately, this doesn't follow the Cygwin conventions about paths (`/` vs. `\`, `cygdrive` root).

The `init.sh` file is not better:

```
if [ -t 0 ]; then
  test -r C:\Users\frede\AppData\Local\opam\opam-init\complete.sh && . C:\Users\frede\AppData\Local\opam\opam-init\complete.sh > /dev/null 2> /dev/null || true

  test -r C:\Users\frede\AppData\Local\opam\opam-init\env_hook.sh && . C:\Users\frede\AppData\Local\opam\opam-init\env_hook.sh > /dev/null 2> /dev/null || true
fi

test -r C:\Users\frede\AppData\Local\opam\opam-init\variables.sh && . C:\Users\frede\AppData\Local\opam\opam-init\variables.sh > /dev/null 2> /dev/null || true
```

```
# opam config report
# opam-version         2.2.0~beta1
# self-upgrade         no
# system               arch=x86_64 os=win32 os-distribution=cygwin os-version=10.0.22621
# solver               builtin-mccs+glpk
# install-criteria     -removed,-count[avoid-version,changed],-count[version-lag,request],-count[version-lag,changed],-count[missing-depexts,changed],-changed
# upgrade-criteria     -removed,-count[avoid-version,changed],-count[version-lag,solution],-count[missing-depexts,changed],-new
# jobs                 7
# repositories         1 (version-controlled)
# pinned               0
# current-switch       default
# invariant            ocaml-variants = 4.14.1+mingw64c
# compiler-packages    ocaml-variants.4.14.1+mingw64c
# ocaml:native         true
# ocaml:native-tools   true
# ocaml:native-dynlink true
[...]

PR #5036, opened 2022-02-02
init scripts: revert environment before settings new one, on session loading

rjbou:
fix #4649 + add a test that display init scripts

PR #5841, opened 2024-02-15
Quote all the paths to OPAMROOT when creating the init scripts on Unix

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/5804

I personally think keeping the Windows style paths on Windows when using cygwin shell is fine, however none of the paths were previously quoted which is a big issue and the reason why the user in #5804 looked in the file in the first place as such command will always fail on Windows.

On cmd and powershell double quoting is apparently enough to make sure the string can't escape its context (?) (e.g. `"` and `$` are not valid file characters)

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -100,6 +100,7 @@ users)
 ## Client
 
 ## Shell
+  * Quote all the paths to OPAMROOT when creating the init scripts on Unix in case OPAMROOT contains spaces, backslashes or special characters [#5841 @kit-ty-kate - fix #5804]
 
 ## Internal
 
--- a/master_changes.md
+++ b/master_changes.md
@@ -63,6 +63,7 @@ users)
   * Fix shell detection on Windows when opam is called via Cygwin's /usr/bin/env even though cmd/powershell is used [#5797 @kit-ty-kate]
   * Fix incorrect deduplication of environment variables on update. Effect was that FOO += "" would occlude the value of FOO in the environment [#5837 @dra27]
   * Fix regression from #5356 on the detection of out-of-date environment variables. As part of a refactoring, a filter predicate got inverted [#5837 @dra27]
+  * Unixify Windows paths in init shells scripts (sh, bash, zsh, fish & tsh) [#5797 @rjbou]
 
 ## Opamfile
 
--- a/src/core/opamSystem.mli
+++ b/src/core/opamSystem.mli
@@ -202,8 +202,7 @@ val get_cygpath_function: command:string -> (string -> string) lazy_t
 val get_cygpath_path_transform: (pathlist:bool -> string -> string) lazy_t
 
 (** [apply_cygpath path] applies the `cygpath` command to [name] using
-    `cygpath -- name`. `cygpath` is resolved with default environment
-    in the function. *)
+    `cygpath -- name`. *)
 val apply_cygpath: string -> string
 
 (** [command cmd] executes the command [cmd] in the correct OPAM
--- a/src/state/opamEnv.ml
+++ b/src/state/opamEnv.ml
@@ -875,19 +875,25 @@ let env_hook_script shell =
 
 let source root shell f =
   let fname = OpamFilename.to_string (OpamPath.init root // f) in
+  let unix_transform ?using_backslashes () =
+    let cygpath = Lazy.force OpamSystem.get_cygpath_path_transform in
+    cygpath ~pathlist:false fname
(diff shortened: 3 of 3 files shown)


rjbou (approved):
Updated accordingly, it then needs another review :)

kit-ty-kate:
Why did you remove the calls to `OpamStd.Env.escape_single_quotes` ? Is the cygpath function doing this job instead?

rjbou:
There is no need to escape them for printing, as cygpath will output the path with slashes. And cygpath function don't need to have a single escaped input.

kit-ty-kate:
what happens on unix though? (when cygpath is not present)

rjbou:
tested on cygwin shell
```
------------------
./opam/opam-init/init.cmd
------------------
if exist "C:\Users\opam\.opam\opam-init\variables.cmd" call "C:\Users\opam\.opam\opam-init\variables.cmd" >NUL 2>NUL

------------------
./opam/opam-init/init.csh
------------------
if ( $?prompt ) then
  if ( -f '/cygdrive/c/Users/opam/.opam/opam-init/env_hook.csh' ) source '/cygdrive/c/Users/opam/.opam/opam-init/env_hook.csh' >& /dev/null
endif

if ( -f '/cygdrive/c/Users/opam/.opam/opam-init/variables.csh' ) source '/cygdrive/c/Users/opam/.opam/opam-init/variables.csh' >& /dev/null

------------------
./opam/opam-init/init.fish
------------------
if isatty
  source '/cygdrive/c/Users/opam/.opam/opam-init/env_hook.fish' > /dev/null 2> /dev/null; or true
end

source '/cygdrive/c/Users/opam/.opam/opam-init/variables.fish' > /dev/null 2> /dev/null; or true

------------------
./opam/opam-init/init.ps1
------------------
. "C:\Users\opam\.opam\opam-init\variables.ps1" *> $null

------------------
./opam/opam-init/init.sh
------------------
if [ -t 0 ]; then
  test -r '/cygdrive/c/Users/opam/.opam/opam-init/complete.sh' && . '/cygdrive/c/Users/opam/.opam/opam-init/complete.sh' > /dev/null 2> /dev/null || true

  test -r '/cygdrive/c/Users/opam/.opam/opam-init/env_hook.sh' && . '/cygdrive/c/Users/opam/.opam/opam-init/env_hook.sh' > /dev/null 2> /dev/null || true
fi

test -r '/cygdrive/c/Users/opam/.opam/opam-init/variables.sh' && . '/cygdrive/c/Users/opam/.opam/opam-init/variables.sh' > /dev/null 2> /dev/null || true

------------------
./opam/opam-init/init.zsh
------------------
if [[ -o interactive ]]; then
  [[ ! -r '/cygdrive/c/Users/opam/.opam/opam-init/complete.zsh' ]] || source '/cygdrive/c/Users/opam/.opam/opam-init/complete.zsh' > /dev/null 2> /dev/null

  [[ ! -r '/cygdrive/c/Users/opam/.opam/opam-init/env_hook.zsh' ]] || source '/cygdrive/c/Users/opam/.opam/opam-init/env_hook.zsh' > /dev/null 2> /dev/null
fi

[[ ! -r '/cygdrive/c/Users/opam/.opam/opam-init/variables.sh' ]] || source '/cygdrive/c/Users/opam/.opam/opam-init/variables.sh' > /dev/null 2> /dev/null
```
reminds me of #5036, that tried to add init scripts tests

rjbou:
> what happens on unix though? (when cygpath is not present)

[`get_cygpath_path_transform`](https://github.com/ocaml/opam/blob/1507b232f46b0bdbafc69baf355780e35a52cb06/src/core/opamSystem.ml#L570) is a no-op (identity)

kit-ty-kate:
> > what happens on unix though? (when cygpath is not present)
> 
> [`get_cygpath_path_transform`](https://github.com/ocaml/opam/blob/1507b232f46b0bdbafc69baf355780e35a52cb06/src/core/opamSystem.ml#L570) is a no-op (identity)

in that case this is a problem and `OpamStd.Env.escape_single_quotes` is still required

rjbou:
Oh, i didn't see that it was an addition for the quoting. Yes, it is needed!

rjbou:
Updated

rjbou (approved):

rjbou:
thanks!

rjbou:
Bench tests failing as service was on available.

Issue #5681, opened 2023-09-20
Broken use of curl in 2.2.0~alpha3 on native Windows


jonahbeckford:
### Assets

* `opam.exe`: https://github.com/diskuv/dkml-component-opam/releases/tag/2.2.0-alpha3-20230918 has a tarball that contains MSVC-compiled Windows 64-bit `opam.exe` (and other operating systems but those are easy to build).
* `with-dkml.exe`: Sadly there is not an easy way to get this except to install DkML 2.0.3 from https://gitlab.com/dkml/distributions/dkml/-/releases/2.0.3. But for this issue all that it should be doing is placing MSYS2 into the PATH.

### Problem

Context: `curl` has always been broken as a download method on native Windows. So after `opam init` I always do a `opam option --yes --global download-command=wget` in DkML.

With `opam.2.2.0~alpha0` (commits until 2022-12-28) I could do the following without a problem:

```sh
$ with-dkml opam init --yes --no-setup --bare --disable-sandboxing default https://opam.ocaml.org

# The following command is executed only if repo/default or repo/default.tar.gz was not created
# - It is created in 2.2.0~alpha0 so this won't execute. Some earlier versions of opam didn't though.
$ with-dkml opam repository add default https://opam.ocaml.org --yes --dont-select --rank=3

$ with-dkml opam option --yes --global download-command=wget
```

Now with `opam.2.2.0~alpha3` (commits until 2023-09-18) I have to use `--no-cygwin-setup` (which is fine) but I get an error:

```sh
$ with-dkml opam init --yes --no-setup --bare --disable-sandboxing --no-cygwin-setup
[WARNING] Flag --no-cygwin-setup is experimental, there is no guarantee that it will be kept; avoid using it in scripts.
No configuration file found, using built-in defaults.
Checking for available remotes: rsync and local, git.
  - you won't be able to use mercurial repositories unless you install the hg command on your system.
  - you won't be able to use darcs repositories unless you install the darcs command on your system.


<><> Fetching repository information ><><><><><><><><><><><><><><><><><><><>  🐫
[ERROR] Could not update repository "default": OpamDownload.Download_fail(_, "curl: code %http_coden while downloading https://opam.ocaml.org/index.tar.gz")
[ERROR] Initial download of repository failed.
```

jonahbeckford:
Blocks verification of #5673

jonahbeckford:
### Root cause

The `opam.2.2.0~alpha0` version had debug logs:

```
[...]

PR #5843, opened 2024-02-15
Opam 2.2 init with MSYS2

### LINKED ISSUE #5683: opam init with MSYS2
I am quite fine with `OPAMFETCH=wget` ... it is only for the initial `opam init`.

And it gets past the `curl` issue, so technically this issue can be closed.

However, `opam init` fails at a later point, and I do now think that `--msys2-location=DIR` is unavoidable. I'll open a new issue for that.

_Originally posted by @jonahbeckford in https://github.com/ocaml/opam/issues/5681#issuecomment-1728073761_

rjbou:
Opam 2.2 init on Windows is focused on resolving a pre-existent Cygwin install, or install opam internal one. Functions that detect if a Cygwin install is present are strictly checking a Cygwin one. This behaviour broke opam usage under MSYS2. As MSYS2 Cygwin is a Cygwin variant, it is not detected, and no specific handling of paths, syscall, etc. are done.

MSYS2 is not officially supported (not testing it in a regular basis), we should ensure that we can keep MSYS2 users to have a way to use opam.

This PR fixes that behaviour by relaxing Cygwin install checks to accept variants. As for Cygwin setup for `opam init`, a check of the install is done, and to keep the information, the config file is appended with
```
global-variables: [os-distribution "msys2" "Set by opam init"]
```
to be sure that further invocations of opam we don't loose that we are in an MSYS2 context (related to #5348).

*ftm in the PR, there is no `sys-package-manager-cmd` setting, user can set it by hand using `opam option`. It is an open question.*

fixes #5683

/cc @jonahbeckford

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -133,6 +133,7 @@ users)
 # API updates
 ## opam-client
   * `OpamClient.windows_checks`: On existing cygwin install, permit to detect msys2 and store `os-distribution=msys2` in `global-variables` config file field [#5843 @rjbou]
+  * `OpamClient.windows_checks`: When updating config file for msys2, resolve `pacman` path and store it in `sys-pkg-manager-cmd` for msys2 [#5843 @rjbou]
 
 ## opam-repository
 
--- a/master_changes.md
+++ b/master_changes.md
@@ -134,6 +134,7 @@ users)
 ## opam-client
   * `OpamClient.windows_checks`: On existing cygwin install, permit to detect msys2 and store `os-distribution=msys2` in `global-variables` config file field [#5843 @rjbou]
   * `OpamClient.windows_checks`: When updating config file for msys2, resolve `pacman` path and store it in `sys-pkg-manager-cmd` for msys2 [#5843 @rjbou]
+  * `OpamArg.apply_global_options`: load MSYS2 Cygwin binary path too [#5843 @rjbou]
 
 ## opam-repository
 
@@ -141,6 +142,7 @@ users)
   * `OpamEnv.env_expansion`: Fix detection of out-of-date environment variables, a filter predicate was inverted [#5837 @dra27]
   * `OpamSysInteract.Cygwin.check_install`: add `variant` argument to permit checking that it is an Cygwin-like install if it is set to true, keep checking that it is a strictly Cygwin install if false [#5843 @rjbou]
   * `OpamSysInteract.Cygwin.check_install`: look for `cygcheck.exe` in `usr/bin` also as MSYS2 doesn't have "bin" [#5843 @rjbou]
+  * `OpamGlobalState.load_config`: load MSYS2 Cygwin binary path too at config file loading [#5843 @rjbou]
 
 ## opam-solver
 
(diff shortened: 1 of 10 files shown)


kit-ty-kate:
I have trouble testing this PR.
* `opam init --bare` won't work because setting `C:\msys64` as a manual cygwin install fails with `cygcheck.exe not found`
* `opam init --bare --no-cygwin-setup` won't work because of https://github.com/ocaml/opam/issues/5681

How are users of MSYS2 expected to use opam?

jonahbeckford:
Thanks!

I am going to set expectations that I won't be able to look at this for at least a week. Most likely two weeks. But I will get to it!

jonahbeckford:
Yep, same issues that @kit-ty-kate reported.

```
WDAGUtilityAccount@37bfe102-8339-4958-b1e5-11ac3588b96d CLANG64 /c/Users/WDAGUtilityAccount/Tools/opam
# ./opam init --yes --no-setup --bare --disable-sandboxing '--git-location=/c/Program Files/Git/cmd' --cygwin-location=/c/msys64 default https://opam.ocaml.org
[WARNING] Flag --cygwin-location is experimental, there is no guarantee that it will be kept; avoid using it in scripts.
[WARNING] Flag --git-location is experimental, there is no guarantee that it will be kept; avoid using it in scripts.

<><> Git ><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫
Using Git from C:\Program Files\Git\cmdYou can change that later with 'opam option "git-location=C:\A\Path\bin"'
<><> Unix support infrastructure ><><><><><><><><><><><><><><><><><><><><><>  🐫
[ERROR] Error while checking Cygwin install (C:\cygwin64): bin\cygcheck.exe not found in C:\msys64
```

It clearly though is using `cygpath` to translate paths which is the main thing we wanted.

<details>

<summary>Test Apparatus (for me)</summary>

* https://gitlab.com/dkml/distributions/dkml
* `Package-WindowsSandbox` target from `pkg\bump\WindowsSandbox\CMakeLists.txt`

```
.\install-vcruntime.cmd
.\install-msys2.cmd
.\install-winget.cmd
.\install-git.cmd
cd opam # Unpack the .zip from GitHub Actions
C:\msys64\msys2_shell.cmd -here -clang64
```

</details>

rjbou:
Updated PR, previous 2 points resolved.

kit-ty-kate (approved):
Thanks a lot! It works really well as far as i tested! (being: `opam init --bare`, `opam switch create --empty`, `opam install conf-autoconf`)

kit-ty-kate:
Thanks!

s5bug:
When this makes it to a build, what will the procedure be for ripping out the OPAM Cygwin and configuring it to use my MSYS2?

kit-ty-kate:
@s5bug something like:
```
opam init '--cygwin-location=C:\msys64'
```
and without the option (a simple `opam init`) you'll be asked if you want to install the default Cygwin (recommended) or provide a custom location.

PR #5847, opened 2024-02-16
Add `sys-pkg-manager-cmd`  in init config

### LINKED ISSUE #5844: Init config file does not define `sys-pkg-manager-cmd`
It is be useful to define at `init` level the package manager path for MSYS2, cf #5348

related to #5843 

And can be useful in the future for when we want to be more permissive and let the user specify some of package manager paths.

rjbou:
fixes #5844


kit-ty-kate (approved):

kit-ty-kate:
Thanks!

rjbou:
Bench test failing as non avail

PR #5862, opened 2024-02-28
Windows: Remove use of deprecated function SHGetFolderPath and use SHGetKnownFolderPath instead

### LINKED ISSUE #5861: Opam for windows : make cold fails with `OPAMW_SHGetFolderPath`
I’m trying to compile opam for windows from scratch. I have an error in the command make cold, which comes after the compiler has been bootstrapped:

```
$ make cold OCAML_PORT=mingw64
…
Opam will be built WITH its default built-in solver

Executables will be installed in /usr/local/bin
Manual pages will be installed in /usr/local/share/man

Downloading vendored source dependencies...
done
Extracting vendored source dependencies in src_ext/... done
env PATH="`pwd`/bootstrap/ocaml/bin:$PATH" CAML_LD_LIBRARY_PATH= make
make[1]: Entering directory '/home/opam'
/home/opam/src_ext/dune-local/dune.exe build --profile=release --root .  --promote-install-files -- opam-installer.install opam.install
Fatal error: exception Failure("OPAMW_SHGetFolderPath")
Fatal error: exception Failure("OPAMW_SHGetFolderPath")
File "doc/man/dune", line 5, characters 0-142:
5 | (rule
6 |   (targets opam.1)
7 |   (deps opam-topics.inc opam-admin-topics.inc)
8 |   (action (with-stdout-to %{targets} (run %{bin:opam} --help=groff))))
Fatal error: exception Failure("OPAMW_SHGetFolderPath")
make[1]: *** [Makefile:144: build-opam] Error 1
make[1]: Leaving directory '/home/opam'
make: *** [Makefile:283: cold] Error 2
```

kit-ty-kate:
Fixes #5861 

Based on the 6 years old comment at the top of `OPAMW_SHGetFolderPath` we might probably want to use the new API and don't want to spend the time supporting such an old system


src/stubs/win32/opamWindows.c
```diff
@@ -549,24 +551,29 @@ CAMLprim value OPAMW_process_putenv(value pid, value key, value val)
...
 }
 
-/*
- * Somewhat against my better judgement, wrap SHGetFolderPath rather than
- * SHGetKnownFolderPath to maintain XP compatibility. OPAM already requires
- * Windows Vista+ because of GetCurrentConsoleFontEx, but there may be a
- * workaround for that for XP lusers.
- */
-CAMLprim value OPAMW_SHGetFolderPath(value nFolder, value dwFlags)
+value OPAMW_SHGetKnownFolderPath(REFKNOWNFOLDERID rfid)
```
dra27, line 554:
```suggestion
static value OPAMW_SHGetKnownFolderPath(REFKNOWNFOLDERID rfid)
```
(the function name could then be de-prefixed as well, if you want, e.g. `get_current_known_folder_path` or some such)

src/stubs/win32/opamWindows.c
```diff
@@ -549,24 +551,29 @@ CAMLprim value OPAMW_process_putenv(value pid, value key, value val)
...
-    caml_failwith("OPAMW_SHGetFolderPath");
+  PWSTR path = NULL;
+  value result;
+
+  if (SUCCEEDED(SHGetKnownFolderPath(rfid, (DWORD)0, NULL, &path))) {
+    result = caml_copy_string_of_utf16(path);
+    CoTaskMemFree((LPVOID)path);
+    return result;
+  } else {
+    CoTaskMemFree((LPVOID)path);
```
dra27, line 564: Entirely under style/choice, but the casts are just noise:
```suggestion
  if (SUCCEEDED(SHGetKnownFolderPath(rfid, 0, NULL, &path))) {
    result = caml_copy_string_of_utf16(path);
    CoTaskMemFree(path);
    return result;
  } else {
    CoTaskMemFree(path);
```

src/core/opamStd.ml
```diff
@@ -1007,8 +1006,7 @@ module OpamSys = struct
         r
 
   let system () =
-    (* CSIDL_SYSTEM = 0x25 *)
-    OpamStubs.(shGetFolderPath 0x25 SHGFP_TYPE_CURRENT)
+    OpamStubs.getPathToSystem ()
```
dra27, line 1009: This was already (my) bad, but this ought to be lazy (given the number of times we can end up calling cygpath, etc. - this function is used the Cygwin-processing in `OpamProcess`), something like:

```ocaml
let system =
  let system = Lazy.from_fun OpamStubs.getPathToSystem in
  fun () -> Lazy.force system
```

dra27 (commented):
Thanks for this - apart from getting rid of youthfully exuberent (= daft) comments of mine from the source code, the fact `SHGetKnownFolderPath` handles allocation itself will automatically deal with long paths (which wasn't a thing when I wrote the original code; I just still harboured an unhealthy desire to keep old OSes working...!). I have a hunch that long paths might be what's gone wrong in the original issue (although it's surprising).

My instinct would be to have a single C stub taking a variant (more closely wrapping one API function), but actually I think your instinct is better in having separate functions! 🙂

We need to link opam with ole32 (for `CoTaskMemFree`) and uuid (for the `FOLDERID_` "constants") - I'm pushing a commit which hopefully does that bit!

dra27 (approved):
LGTM, thanks!

PR #5864, opened 2024-03-04
Shell integration: Test if file exists before sourcing in fish + powershell

ElectreAAS:
This PR is an off-shoot from #5802. Same two first commits, to make it easier to review and produce a cleaner history


kit-ty-kate (approved):

rjbou (approved):

kit-ty-kate:
Thanks!

PR #5867, opened 2024-03-04
Fix the reftests when running in a directory containing spaces

kit-ty-kate:


rjbou:
What is your test case on which it fails?

kit-ty-kate:
> What is your test case on which it fails?

I haven't tested, but leaving `rm -rf` around with unquoted variable arguments is a recipe for disaster.

rjbou:
I understand the point of quoting `rm -rf` calls, and completely agree. Better be cautious.

In that specific case, `OPAMROOT` is safe for the moment. It is a temporary directory, composed by dune temporary directory `/tmp/build_xxxxxx_dune`, appending to it the string `/opam-reftest-xxxxxx` for the test current directory, and in the end `/OPAM` for the opam root directory. It ends up to a string that does not contain spaces, unless user changes its temporary directory to a path with space, or dune changes its temporary directory with a space.
Plus, as it is a variable, it is resolved at the last moment, so `$OPAMROOT`, even if it contains spaces, is a single string in the `args` list given to `Unix.create_process_env`.

If we want to avoid completely that unquoted `rm -rf` issue, there is other calls to it not on opamroot but on other variables (`BASEDIR` if i remember well).

I give it a try to run reftests in a directory that contains spaces, and it fails on other tests, for other reasons.

kit-ty-kate:
oh right i forgot run.ml didn't run sh but it's own command parser. Nevermind then, sorry for the wasted time

Issue #5770, opened 2023-12-24
[macOS] `opam` 2.1.5 fails to build: `sed: opam.install: No such file or directory` `make: *** [processed-opam.install] Error 1`


barracuda156:
Unable to build `opam` 2.1.5 on macOS:
```
make[1]: Leaving directory `/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_ocaml_opam/opam/work/opam-full-2.1.5/src_ext'
dune build --profile=release --root .   -- opam-installer.install opam.install
fatal: not a git repository (or any of the parent directories): .git
fatal: not a git repository (or any of the parent directories): .git
File "src/core/opamStubsTypes.ml", line 1:
Warning 70 [missing-mli]: Cannot find interface file.
File "src/tools/opam_installer.ml", line 1:
Warning 70 [missing-mli]: Cannot find interface file.
File "src/solver/opamCudfSolverSig.ml", line 1:
Warning 70 [missing-mli]: Cannot find interface file.
dune build --profile=release --root .   -- opam-installer.install opam.install
sed -f process.sed opam.install > processed-opam.install
sed: opam.install: No such file or directory
make: *** [processed-opam.install] Error 1
make: Leaving directory `/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_ocaml_opam/opam/work/opam-full-2.1.5'
Command failed:  cd "/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_ocaml_opam/opam/work/opam-full-2.1.5" && /usr/bin/make -j1 -w lib-ext all 
Exit code: 2
```

Tested on:

macOS 14.2.1 / arm64 / Apple clang 15
macOS 10.6 / ppc / gcc 13.2.0

Versions used:
`ocaml` 4.14.1
`dune` 3.12.1

See also: https://trac.macports.org/ticket/68586

kit-ty-kate:
what does `ocamlc -config` say?
How did you build opam? If you used the vendored ocaml compiler (using `make cold` or `make compiler`), what does `./bootstrap/ocaml/bin/ocamlc -config` say?

barracuda156:
@kit-ty-kate This is on a PowerPC machine:

```
36-29% /opt/local/ocamlc -config
zsh: no such file or directory: /opt/local/ocamlc
36-29% /opt/local/bin/ocamlc -config
version: 4.14.1
standard_library_default: /opt/local/lib/ocaml
[...]

PR #5868, opened 2024-03-04
Make sure to use the vendored dune when using ./configure --with-vendored-deps

kit-ty-kate:
Partially fixes https://github.com/ocaml/opam/issues/5770 (the remaining part is to be fixed in the 2.1 branch)


kit-ty-kate:
It looks like i didn't read correctly the description of the `--without-dune` option

PR #5869, opened 2024-03-04
Upgrade the vendored dune to 3.14.0 and other packages to their latest release

kit-ty-kate:
Fixes an issue reported in https://github.com/ocaml/opam/issues/5770 that happens on some architectures (e.g. PPC).
Prepares fix for https://github.com/ocaml/opam/issues/5861 (coming in dune 3.14.1)

PR #5871, opened 2024-03-05
Bump src_ext versions and patch Dune

dra27:
This should be a temporary fix for #5861, until Dune 3.14.1 is released.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -41,6 +41,7 @@ src_ext/seq/
 src_ext/sha/
 src_ext/spdx_licenses/
 src_ext/stdlib-shims/
+src_ext/swhid_core/
 src_ext/topkg/
 Opam.Runtime.*/
 *.tar.bz2
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ else
 endif
 
 ifeq ($(DUNE),)
-  DUNE_EXE = src_ext/dune-local/dune.exe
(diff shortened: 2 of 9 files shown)


kit-ty-kate:
We seemed to have duplicated some work between #5869 (and the subsequent PR for 3.14.1 i had ready). It looks like dune 3.14.1 is going to be released pretty soon so the patch might be obsolete quite quickly.
If it is released before we merge all the other blockers for 2.2.0~beta2 i'd personally rather go with the patch-less approach

dra27:
Oo, that's very silly of me - sorry! There are some bits of this one that are worth combining with yours (the not-strictly-related .gitignore part and the update to the other src_exts).

Let's see whether it becomes moot, but I disagree with the patch - I think it's better to have an inspectable patch in the repo (just as we used to) rather than a pointer a specific sha elsewhere.

kit-ty-kate:
> Let's see whether it becomes moot, but I disagree with the patch - I think it's better to have an inspectable patch in the repo (just as we used to) rather than a pointer a specific sha elsewhere.

oh no i agree with that. What i meant is that if dune 3.14.1 releases like tomorrow or something it would be better to use that instead directly

dra27:
Ohhhh 🤦‍♂️ Yes, we were arguing in total agreement 🙂

dra27:
This is all merged in #5869 or will be fixed in Dune 3.14.1

PR #5879, opened 2024-03-12
Support `avoid-version` and `deprecated` with the 0install-solver

dra27:
This plumbs in https://github.com/ocaml-opam/opam-0install-solver/pull/37, both of which I think we should sort out, but possibly not for 2.2.0.

### DIFF
--- a/src/solver/opamSolver.ml
+++ b/src/solver/opamSolver.ml
@@ -167,15 +167,25 @@ let opam2cudf_map universe version_map packages =
     OpamPackage.Set.fold (fun nv -> OpamPackage.Map.add nv true)
       (packages %% set) OpamPackage.Map.empty
   in
+  let avoid_versions =
+    OpamStd.Option.default OpamPackage.Set.empty @@
+    OpamStd.List.assoc_opt String.equal "avoid-version" universe.u_attrs
+  in
   let base_map =
     OpamPackage.Set.fold (fun nv ->
+        let pkg_extra = [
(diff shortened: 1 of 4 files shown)


dra27:
I don't think we had this in a different PR, @kit-ty-kate?

kit-ty-kate:
> I don't think we had this in a different PR, @kit-ty-kate?

not that i can remember i don't think.

dra27:
Just for context, this was a brief rabbithole from https://github.com/ocaml/opam-repository/pull/25441#discussion_r1521147331 🙂

src/solver/opamSolver.ml
```diff
@@ -167,15 +167,25 @@ let opam2cudf_map universe version_map packages =
...
+  in
   let base_map =
     OpamPackage.Set.fold (fun nv ->
+        let pkg_extra = [
+          OpamCudf.s_source, `String(OpamPackage.name_to_string nv);
+          OpamCudf.s_source_number, `String(OpamPackage.version_to_string nv);
+        ] in
+        let pkg_extra =
+          if OpamPackage.Set.mem nv avoid_versions then
+            (OpamCudf.s_avoid_version, `Bool true)::pkg_extra
```
kit-ty-kate, line 182: This change should not be necessary after https://github.com/ocaml-opam/opam-0install-solver/pull/37/files#r1697389138

kit-ty-kate:
Superseded by https://github.com/ocaml/opam/pull/6130

PR #5883, opened 2024-03-17
Windows-inspired tweaks the UI

dra27:
None of this is intended for direct merging - it's the result of some idle hacking while travelling at the moment.

There are four separable parts of this PR, which anyone should feel free to take and move to completion 🙂

- The display of command output (when `-v`) is specified is implemented on Unix using `Unix.alarm` and so was consequently disabled on Windows in #1915. It's an irritating limitation when debugging the compiler in opam on Windows, though! The approach is to use a thread instead of sigalrm, but the implementation needs to be hardened (the `XXX` notes are things I didn't have in my frontal lobe and couldn't be bothered to lookup!)
- Hidden in that first commit is also an actual bug fix. The code in #1915 uses `input_line` until it reaches `End_of_file`, but the final line may not actually be complete. Regardless, opam then displays it on a complete. Even on Linux, this causes lines in `configure` scripts to get split. There's a hack in the commit to fix it by using `seek_in`, but it would be better to use proper input functions and do the job properly (but it's fiddly). I'm fairly sure the statement of correctness in a comment in the code is _wrong_.
- Extracting tarballs with many files is slow on Windows... the next two commits were starts of an implementation to display an "extract" item at this point, which makes opam looks much less like it's crashed while it extracts either the compiler or Dune tarballs (I was amused when installing a package from npm earlier to note that it really ought to do the same thing.......)
- Finally, packages are slower on Windows to build at the moment (especially the compiler) and the last commit provides a PoC for displaying spinners. The commit message has a few more notes. This needs massively tidying up - `OpamProcess` should acquire a more structured way of determining "progress" (which might allow, say, the download jobs to tick when content is received and the tarball extractions to tick when a file has been extracted) and the type for the status bar in `OpamParallel` should be made richer so that it can be much less hackily annotated the a spinner. Obviously the spinner itself needs to get through the usual mechanisms in OpamConsole for UTF-8 detection, but I was hacking all this in Windows Terminal, with full Unicode joy...

The verbose fixes are optional for 2.2.0, but both the extract and spinners I think _are_ required for 2.2.0 because there are otherwise extremely long pauses on Windows which invite pressing CTRL+C.

### DIFF
--- a/src/core/dune
+++ b/src/core/dune
@@ -3,7 +3,7 @@
   (public_name opam-core)
   (synopsis    "OCaml Package Manager core internal stdlib")
   ; TODO: Remove (re_export ...) when CI uses the OCaml version that includes https://github.com/ocaml/ocaml/pull/11989
-  (libraries   re (re_export ocamlgraph) unix sha jsonm swhid_core uutf
+  (libraries   re (re_export ocamlgraph) unix sha jsonm swhid_core uutf threads
                (select opamACL.ml from
                        (opam-core.libacl -> opamACL.libacl.ml)
                        (                 -> opamACL.dummy.ml))
--- a/src/core/opamParallel.ml
+++ b/src/core/opamParallel.ml
@@ -98,42 +98,80 @@ module Make (G : G) = struct
       raise (Cyclic sccs)
     );
 
+    let rec limit_width acc rem_cols = function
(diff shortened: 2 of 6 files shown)


dra27:
It can be seen in action on Windows in:

PR #5885, opened 2024-03-18
Manual: add missing 'since opam 2.2' annotation when mentionning with-dev-setup

kit-ty-kate:


rjbou (approved):
LGTM!

In the manual, there was no specification of `since xx` for newly added features or fields. It was like that as there was the separation between documentation for 1.2, 2.0, 2.1 in the menu. But it is no more done, and master is the one deployed, which can be confusing for users.
We should do a whole review of the doc to put `since xxx` everywhere it is required. The current status (mix of some previous changes that introduced some `since xxx` and other without it) can confuse the reader that everything that is not annotated is always present/working (2.0 & 2.1 compliant).

PR #5893, opened 2024-03-21
Warn if GNU patch is not detected when using patch

kit-ty-kate:
Lightweight alternative to #5892 
See #5400 

In this PR, once per process if a patch is going to be applied (either a patch application in a package or a repository update) we first check which patch executable to use. If no GNU patch is detected we display a warning.

On some platforms the name of the `patch` command can change depending on the context (e.g. macOS where Homebrew provides `patch` but Macports provides `gpatch`), so to handle that we simply check both `patch` and `gpatch` and detect which one is GNU patch. On the platforms more likely to have `patch` as GNU patch we check `patch` first and default to it if none were found, and inversely on the platforms that are more likely to have `gpatch` as GNU patch.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -31,6 +31,7 @@ users)
   * Simplify computation of OCaml init default `sys-ocaml-*` eval variables on Unix [#5829 @dra27]
   * Add a init OCaml `sys-ocaml-system` eval variable [#5829 @dra27]
   * Mark the internal cygwin installation as recommended [#5903 @kit-ty-kate]
+  * Check for gpatch instead of patch on NetBSD and DragonFlyBSD [#5893 @kit-ty-kate]
 
 ## Config report
 
@@ -65,6 +66,7 @@ users)
  * Add warning 69: Warn for new syntax when package name in variable in string interpolation contains several '+' [#5840 @rjbou]
 
 ## Repository
+  * Warn if `GNU patch` is not detected during a repository update [#5893 @kit-ty-kate]
 
 ## Lock
 
@@ -126,6 +128,9 @@ users)
   * Quote all the paths to OPAMROOT when creating the init scripts on Unix in case OPAMROOT contains spaces, backslashes or special characters [#5841 @kit-ty-kate - fix #5804]
 
 ## Internal
+  * Warn if `GNU patch` is not detected when a patch is applied [#5893 @kit-ty-kate]
+  * Use `gpatch` by default instead of `patch` on NetBSD and DragonFlyBSD [#5893 @kit-ty-kate]
+  * Use `gpatch` if it exists and is detected as GNU patch when `patch` is not `GNU patch` [#5893 @kit-ty-kate]
 
 ## Internal: Windows
   * Ensure that the system critical error dialog is disabled when opam starts [#5828 @dra27]
@@ -191,3 +196,4 @@ users)
 ## opam-core
   * `OpamStd.Sys`: add `is_cygwin_variant_cygcheck` that returns true if in path `cygcheck` is from a Cygwin or MSYS2 installation [#5843 @rjbou]
   * `OpamStd.Env.cyg_env`: takes the environment to cygify, usually `OpamStd.Env.raw_env` [#5829 @dra27]
+  * `OpamSystem.patch` now displays a warning when GNU patch is not detected and looks for both patch and gpatch as a backup option depending on the OS [#5893 @kit-ty-kate]
--- a/src/client/opamInitDefaults.ml
+++ b/src/client/opamInitDefaults.ml
@@ -46,6 +46,8 @@ let linux_filter = os_filter "linux"
 let macos_filter = os_filter "macos"
 let openbsd_filter = os_filter "openbsd"
 let freebsd_filter = os_filter "freebsd"
+let netbsd_filter = os_filter "netbsd"
+let dragonflybsd_filter = os_filter "dragonfly"
 let not_open_free_bsd_filter =
   FNot (FOr (openbsd_filter,  freebsd_filter))
 let win32_filter = os_filter "win32"
@@ -53,7 +55,9 @@ let not_win32_filter =
   FOp (FIdent ([], OpamVariable.of_string "os", None), `Neq, FString "win32")
 let sandbox_filter = FOr (linux_filter, macos_filter)
 
-let gpatch_filter = FOr (openbsd_filter, freebsd_filter)
+let gpatch_filter =
+  FOr (FOr (openbsd_filter, netbsd_filter),
+       FOr (freebsd_filter, dragonflybsd_filter))
 let patch_filter = FNot gpatch_filter
 
 let gtar_filter = openbsd_filter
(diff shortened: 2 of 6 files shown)


src/core/opamSystem.ml
```diff
@@ -1633,6 +1633,31 @@ let translate_patch ~dir orig corrected =
...
+  | NetBSD
+  | DragonFly
+  | OpenBSD
+  | FreeBSD -> "gpatch"
+  | Darwin
+  | Linux
+  | Cygwin
+  | Win32
+  | Unix
+  | Other _ -> "patch"
```
rjbou, line 1647: Some sorting
```suggestion
  | DragonFly
  | FreeBSD
  | NetBSD
  | OpenBSD -> "gpatch"
  | Cygwin
  | Darwin
  | Linux
  | Unix
  | Win32
  | Other _ -> "patch"
```

rjbou (approved):
On the idea and the code, lgtm!
Can you explain in the main comment why this is applied only on updating opam repositories ?
Also, a precision in the changes/PR/commits, it is not specific to `opam update`, but more generally opam repositories updating. These updates can be launched by `opam update`, `opam switch`, and `opam init`.

src/core/opamSystem.ml
```diff
@@ -1633,6 +1633,31 @@ let translate_patch ~dir orig corrected =
...
   close_in ch
 
+let patch_cmd = lazy begin
+  match OpamStd.Sys.os () with
+  | DragonFly
+  | FreeBSD
+  | NetBSD
+  | OpenBSD -> "gpatch"
+  | Cygwin
+  | Darwin
```
kit-ty-kate, line 1643: mmmh, this is actually tricky. Macports provides `gpatch` but Homebrew provides `patch`

kit-ty-kate: Fixed by resolving both

hannesm:
I do understand the logic to look for "gpatch" .. "patch" -- but I'd do it differently:
- not look for gpatch on GNU/Linux (i.e. only look for it on BSD and macOS)
- unify the whole lookup to uniformly look for gpatch first and if not found (or not GNU patch) go for patch

Of course your PR is fine, but I wonder in terms of lines of code and code understandability whether it is worth it (my experience is if you have the least platform-specific conditionals, the easiest it is to debug).

But these are only my 2 cents... thanks for your work on this.

src/core/opamSystem.ml
```diff
@@ -1633,6 +1633,40 @@ let translate_patch ~dir orig corrected =
...
+    | DragonFly
+    | FreeBSD
+    | NetBSD
+    | OpenBSD -> ("gpatch", ["patch"])
+    | Cygwin
+    | Darwin
+    | Linux
+    | Unix
+    | Win32
+    | Other _ -> ("patch", ["gpatch"])
```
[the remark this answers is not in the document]
kit-ty-kate: the problem with that is that it changes the behaviour compared to 2.1.5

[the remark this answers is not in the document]
hannesm: yes, well, or even:
```OCaml
let default_cmd, other_cmds = "gpatch", [ "patch" ]
```

avoiding the match on the os type at all. But I hear your "changes the behaviour". With that in mind, I like your PR very much.

The only suggestion left is:
```OCaml
  let default_cmd, other_cmds =
    match OpamStd.Sys.os () with
    | DragonFly
    | FreeBSD
    | NetBSD
    | OpenBSD
    | Darwin -> ("gpatch", ["patch"])
    | Cygwin
    | Linux
    | Unix
    | Win32
    | Other _ -> ("patch", [])
```

so, put Darwin into the pot where gpatch is tried first, and don't try gpatch on Linux & Windows.

But as said, maybe the current state is best for avoiding too much trouble / regressions.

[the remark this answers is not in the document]
hannesm: Actually considering the code and platforms, maybe

```OCaml
let default_cmd, other_cmds = "patch", [ "gpatch" ]
```

is suitable? This would on OpenBSD and FreeBSD call `patch --version`, and then go to using `gpatch`...

master_changes.md
```diff
@@ -64,6 +67,7 @@ users)
 ## Lint
 
 ## Repository
+  * Warn if GNU patch is not detected during a repository update [#5893 @kit-ty-kate]
 
 ## Lock
```
rjbou, line 73: + an API changes as now the `OpamSystem.patch` display a warning

src/core/opamSystem.ml
```diff
@@ -1633,6 +1633,40 @@ let translate_patch ~dir orig corrected =
...
+      | r ->
+        (match OpamProcess.is_success r, r.OpamProcess.r_stdout with
+         | true, full::_ when
+             OpamStd.String.is_prefix_of ~from:0 ~full "GNU patch " ->
+           Some patch_cmd
+         | _ ->
+           search_gpatch patch_cmds)
+      | exception _ -> search_gpatch patch_cmds
+  in
+  let default_cmd, other_cmds =
```
rjbou, line 1650: Init defaults is not aligned with this selection
For the moment, it is
```
required-tools: [
 "patch" {} {!(os = "openbsd" | os = "freebsd")}
 "gpatch" {} {os = "openbsd" | os = "freebsd"}
...
```
Shouldn't we add DragonFly and NetBSD, or change to `os-family = "bsd"` ?

src/core/opamSystem.mli
```diff
@@ -304,6 +304,8 @@ val get_lock_fd: lock -> Unix.file_descr
 
 (** {2 Misc} *)
 
+val gpatch : string Lazy.t
+
```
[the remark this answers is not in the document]
kit-ty-kate: good catch. Removed

src/client/opamInitDefaults.ml
```diff
@@ -46,14 +46,16 @@ let linux_filter = os_filter "linux"
...
+let dragonflybsd_filter = os_filter "dragonfly"
 let not_open_free_bsd_filter =
   FNot (FOr (openbsd_filter,  freebsd_filter))
 let win32_filter = os_filter "win32"
 let not_win32_filter =
   FOp (FIdent ([], OpamVariable.of_string "os", None), `Neq, FString "win32")
 let sandbox_filter = FOr (linux_filter, macos_filter)
 
-let gpatch_filter = FOr (openbsd_filter, freebsd_filter)
+let gpatch_filter = FOr (FOr (openbsd_filter, netbsd_filter), FOr (freebsd_filter, dragonflybsd_filter))
```
rjbou, line 58:
```suggestion
let gpatch_filter =
  FOr (FOr (openbsd_filter, netbsd_filter),
       FOr (freebsd_filter, dragonflybsd_filter))
```

rjbou (approved):
LGTM! tiny formatting remark that you can ignore :)

rjbou:
related to #3639

src/client/opamInitDefaults.ml
rjbou: pushed, like that CI runs and good to go tomorrow

PR #5894, opened 2024-03-21
Use git diff instead of the system diff command

kit-ty-kate:
opam already harbour a patch parser used to preprocess CRLF in patch files. As (yet another) alternative to #5892 and #5893 we can take advantage of this preprocessing and output the files that we detect are going to be deleted.

This can only happen if the format in the given patch file is parseable, this is too tricky to do with the system diff commands (see https://github.com/hannesm/patch/issues/8) without a robust date parser so instead we can use `git diff --no-index` when we control the input patch (during opam update). When that's the case, we delete the empty straggler files after the application of the patch that we were able to detect.

This makes the system patch command on macOS usable when packages have been deleted from an opam repository.

See https://github.com/ocaml/opam/pull/5891
Fixes https://github.com/ocaml/opam/issues/3639

### DIFF
--- a/src/repository/opamRepositoryBackend.ml
+++ b/src/repository/opamRepositoryBackend.ml
@@ -84,8 +84,8 @@ let get_diff parent_dir dir1 dir2 =
   OpamSystem.make_command
     ~verbose:OpamCoreConfig.(!r.verbose_level >= 2)
     ~dir:(OpamFilename.Dir.to_string parent_dir) ~stdout:patch
-    "diff"
-    [ "-ruaN";
+    "git"
+    [ "diff"; "--no-index"; "-a"; "--";
       OpamFilename.Base.to_string dir1;
       OpamFilename.Base.to_string dir2; ]
   @@> function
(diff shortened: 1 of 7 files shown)


src/repository/opamRepositoryBackend.ml
```diff
@@ -84,8 +84,8 @@ let get_diff parent_dir dir1 dir2 =
   OpamSystem.make_command
     ~verbose:OpamCoreConfig.(!r.verbose_level >= 2)
     ~dir:(OpamFilename.Dir.to_string parent_dir) ~stdout:patch
-    "diff"
-    [ "-ruaN";
+    "git"
+    [ "diff"; "--no-index"; "-a"; "--";
```
kit-ty-kate, line 88:
```suggestion
    [ "diff"; "--no-index"; "--no-renames"; "-a"; "--";
```

kit-ty-kate:
Superseded by #5892 which i'm hoping to finish whenever the `patch` library is ready

PR #5903, opened 2024-04-02
Mark the internal cygwin installation as recommended

kit-ty-kate:


rjbou (approved):
lgtm! with a change log entry

Issue #5816, opened 2024-01-31
jobs variable reset only works when upgrading format from 2.0 -> 2.2, not 2.0 -> 2.1 -> 2.2


dra27:
When upgrading from a 2.0 root directly to 2.2, one gets:

```
[NOTE] The 'jobs' option was reset, its value was 255 and its new value will vary according to the current number of cores on your machine. You can restore the fixed
       value using:
           opam option jobs=255 --global
```

If the root has already been upgraded to 2.1, this check is skipped (and the old, fixed value, persists).

Not clear on the impact - it just probably just means a lot of users have a hard-coded wrong opam 2.1 value.

rjbou:
linked to #5284

PR #5904, opened 2024-04-02
Reset the "jobs" config variable when upgrading to opam 2.2

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/5816

### DIFF
--- a/src/state/opamFormatUpgrade.ml
+++ b/src/state/opamFormatUpgrade.ml
@@ -1110,7 +1110,18 @@ let sys_config_variables_unix_2_1 = [
   "Host C Runtime Library type of the OCaml compiler present on your system";
 ]
 
-let from_2_0_to_2_1 ~on_the_fly _ conf =
+let from_2_0_to_2_1 ~on_the_fly:_ _ conf =
+  apply_eval_variables conf [] sys_config_variables_unix_2_1, gtc_none
+
+let v2_2_alpha = OpamVersion.of_string "2.2~alpha"
+
+let from_2_1_to_2_2_alpha ~on_the_fly:_ _ conf =
+  apply_eval_variables conf sys_config_variables_unix_2_1
+    OpamEnv.sys_ocaml_eval_variables, gtc_none
+
+let v2_2_beta = OpamVersion.of_string "2.2~beta"
+
+let from_2_2_alpha_to_2_2_beta ~on_the_fly _ conf =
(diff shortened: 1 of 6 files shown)


kit-ty-kate:
Note from the dev meeting: a better fix would be to upgrade the root-version to `2.2~beta` and move the upgrade there instead to prevent issues for (rare) people who would have upgraded their opam all the way from 2.0 to 2.1 to 2.2.0~alpha (in which case this code would not have been run)

src/state/opamFormatUpgrade.ml
```diff
@@ -1130,12 +1134,7 @@ let from_2_0_to_2_1 ~on_the_fly _ conf =
    | Some prev_jobs when prev_jobs = max 1 (OpamSysPoll.cores () - 1) -> ()
    | Some prev_jobs -> info_jobs_changed ~prev_jobs
    | None -> info_jobs_changed ~prev_jobs:1);
-  let conf = apply_eval_variables conf [] sys_config_variables_unix_2_1 in
```
kit-ty-kate, line 1133: issue: this transformation is lost

rjbou (commented):
Thanks! Agree with your comment, we'll see when we upgrade the root number.

src/state/opamFormatUpgrade.ml
kit-ty-kate: fixed

kit-ty-kate:
Ready to be reviewed again

tests/reftests/opamroot-versions.test
```diff
@@ -2712,8 +2746,8 @@ compiler: ["i-am-sys-compiler.2"]
 installed: ["i-am-package.2" "i-am-sys-compiler.2"]
 opam-version: "2.0"
 roots: ["i-am-package.2" "i-am-sys-compiler.2"]
-### :V:6:e: reinit from 2.2~alpha
-### ocaml generate.ml 2.2~alpha
+### :V:6:e: reinit from 2.2~beta
```
rjbou: At each bump root, a new entry should be added, I updated the PR

rjbou (commented):
Once the reftest reviewed, good to go!
thanks!

tests/reftests/opamroot-versions.test
kit-ty-kate: Thanks! That looks good to me

rjbou (approved):
As the tests looks good, unblocking :)

Issue #5755, opened 2023-12-07
`opam tree` causes error when used with `--with-doc`


smorimoto:
Actual error:

```
❯ opam tree ./hello.opam --json=tmp.json --test --doc
Constructing initial basis...
Number of 0-1 knapsack inequalities = 5983
Constructing conflict graph...
Conflict graph has 1395 + 699 = 2094 vertices
[ERROR] Could not simulate installing the specified package(s) to this switch:
The actions to process have cyclic dependencies:
  - install odoc (= 25) -> install camlp-streams (= 2) -> install odoc-parser (= 6) -> install odoc (= 25)
```

This is likely related to: https://github.com/ocaml/opam/issues/4541

ElectreAAS:
Looking into it!

ElectreAAS:
Could you paste the contents of your `hello.opam` file (or a redacted subset) please? I have a hard time reproducing this

kit-ty-kate:
@ElectreAAS using
```
opam-version: "2.0"
depends: [
  "odoc" {with-doc}
]
```
in a switch where you have `odoc` installed i can reproduce it locally using the command above

kit-ty-kate:
I'm also seeing a different unexpected output when given `--no-switch` on top of that:
```
$ opam tree ./hello.opam --test --doc --no-switch
The following actions are simulated:
=== install 1 package
  ∗ hello dev

hello.dev
```
we expect `odoc` to be present in the tree but it's not currently.

PR #5919, opened 2024-04-10
Fix opam tree --test --doc assigning the with-{test,doc} variables to unrequested packages

kit-ty-kate:
Fix https://github.com/ocaml/opam/issues/5755
Fix https://github.com/ocaml/opam/issues/5920

`OpamSwitchState.universe` was called with `~requested:st.installed` so all the packages that were currently installed had their `with-doc`, `with-test`, `with-dev-setup` and `dev` variables assigned instead of just the packages that were given as input to the `opam tree` command.

~TODO:~
  - [x] ~add a test (see https://github.com/ocaml/opam/issues/5755#issuecomment-1921773756)~


src/client/opamTreeCommand.ml
```diff
@@ -378,19 +378,35 @@ let print_solution st new_st missing solution =
...
   match OpamSolver.resolve universe
           (OpamSolver.request ~install ()) with
   | Success solution ->
     let new_st = OpamSolution.dry_run st solution in
-    print_solution st new_st names solution;
-    new_st, get_universe tog new_st
+    print_solution st new_st
+      (OpamPackage.Name.Set.of_list (List.map fst install))
+      solution;
+    let select, missing = get_installed new_st install in
```
rjbou, line 406:
```suggestion
    let root = OpamFormula.packages_of_atoms new_st.install install in
```
with that, you don't need to move and externalise get_installed

kit-ty-kate: done in https://github.com/ocaml/opam/pull/5919/commits/25d2e2284fe707068df808245942b022705e934e

src/client/opamTreeCommand.ml
```diff
@@ -399,24 +402,25 @@ let simulate_new_state tog st universe install names =
...
         let installed =
           OpamPackage.Set.filter (OpamFormula.check atom) st.installed
         in
         if OpamPackage.Set.is_empty installed then
-          (select, atom :: missing)
-        else (installed ++ select, missing))
-      (OpamPackage.Set.empty, []) atoms
+          let missing_pkgs = OpamPackage.Set.filter (OpamFormula.check atom) st.packages in
+          if OpamPackage.Set.is_empty missing_pkgs then
+            assert false;
```
rjbou, line 415: Why this assert false ? it should be possible to query a non existing package, and it will fail when it tries to dry install it.

src/client/opamTreeCommand.ml
```diff
@@ -399,24 +402,25 @@ let simulate_new_state tog st universe install names =
...
-let dry_install tog st universe install =
-  simulate_new_state tog st universe install
-    (OpamPackage.Name.Set.of_list (List.map fst install))
-
 let run st tog ?no_constraint mode filter atoms =
   let open OpamPackage.Set.Op in
-  let select, missing =
-    List.fold_left (fun (select, missing) atom ->
+  let select, missing, requested =
+    List.fold_left (fun (select, missing, requested) atom ->
```
rjbou, line 408: Instead of computing requested here, what do you think of computing the universe conditionally (regarding missing status):
```
  let st, universe =
    let universe =
      let requested =
        OpamFormula.packages_of_atoms
          (if missing = [] then st.installed else st.packages) atoms
      in
      get_universe tog requested st
    in
```

kit-ty-kate: I like the idea. Added in https://github.com/ocaml/opam/pull/5919/commits/36cd0528f8f704489e8b215834cda738007295e1

src/client/opamTreeCommand.ml
kit-ty-kate: Ah yes, my bad, I completely forgot about non-existing packages. Removed in https://github.com/ocaml/opam/pull/5919/commits/36cd0528f8f704489e8b215834cda738007295e1

kit-ty-kate:
All cleaned up, squashed and ready to go.

rjbou (approved):
thanks!

PR #5921, opened 2024-04-11
Fix downloading URLs with invalid characters

dra27:
`OpamDownload.download` downloads a given URL using the basename of the URL as the filename. On Windows, where are there are many restrictions on valid filenames, this causes a problem if the URL includes any query string. Since OpamDownload passes the name used to the continuation, on Windows the illegal characters are simply replaced with underscores.

### DIFF
--- a/src/core/opamStd.mli
+++ b/src/core/opamStd.mli
@@ -532,7 +532,8 @@ module Sys : sig
       OS) *)
   val path_sep: char
 
-  (** [true] if a character may be used in a filename. *)
+  (** [true] if a character may be used in a filename, depending on the OS.
+      For example, on Windows, `:` and `?` can't be in the name. *)
   val is_valid_basename_char: char -> bool
 
   (** Splits a PATH-like variable separated with [path_sep]. More involved than
--- a/src/core/opamStd.mli
+++ b/src/core/opamStd.mli
@@ -532,6 +532,9 @@ module Sys : sig
       OS) *)
   val path_sep: char
 
+  (** [true] if a character may be used in a filename. *)
+  val is_valid_basename_char: char -> bool
+
   (** Splits a PATH-like variable separated with [path_sep]. More involved than
       it seems, because there may be quoting on Windows. By default, it returns
       the path cleaned (remove trailing, leading, contiguous delimiters).
--- a/src/repository/opamDownload.ml
+++ b/src/repository/opamDownload.ml
@@ -188,8 +188,18 @@ let download_as ?quiet ?validate ~overwrite ?compress ?checksum url dst =
       ()
 
 let download ?quiet ?validate ~overwrite ?compress ?checksum url dstdir =
+  let base =
+    let base = OpamUrl.basename url in
+    if Sys.win32 then
+      let f c =
+        if OpamStd.Sys.is_valid_basename_char c then c else '_'
+      in
+      String.map f base
+    else
+      base
+  in
   let dst =
-    OpamFilename.(create dstdir (Base.of_string (OpamUrl.basename url)))
+    OpamFilename.(create dstdir (Base.of_string base))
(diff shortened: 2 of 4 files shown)


kit-ty-kate (commented):
looks fine on its own, however I'm wondering how things interact if two files (e.g. extra-files) are downloaded and rewritten to the same name.

For example `file*.txt` and `file?.txt` would both rewrite to `file_.txt` and would overwrite each other. Is this detected by OpamDownload?

If so it would be nice to have a testcase for this, as well as for the base fix.

dra27:
It's worth the test case regardless, but this isn't disambiguating the _final_ name - it's the intermediate name which if I traced the code through correctly is always done in an empty temporary directory. For example, in:

```
patches: ["zstd-detection.patch"]
extra-source "zstd-detection.patch" {
  src: "https://github.com/ocaml/ocaml/commit/baf65b91c51bb04b09ecc98b94ddd4ba3b446912.patch?full_index=1"
  checksum: "sha256=958e061bc3b967e32a5606d5109ed7faacb9b793fe2de0e8f8697c23a178c5cf"
}
```

the final name (`zstd-detection.patch`) must be Windows-compatible and there's no automatic renaming taking place there. The issue is that while downloading, opam downloads https://github.com/ocaml/ocaml/commit/baf65b91c51bb04b09ecc98b94ddd4ba3b446912.patch?full_index=1 to a file called `baf65b91c51bb04b09ecc98b94ddd4ba3b446912.patch?full_index=1` before moving it to `zstd-detection.patch`

kit-ty-kate:
oh i see! Then I'm wondering, does the name have to be similar as the original? Why not simply have one static name that's always going to be the same?

kit-ty-kate (approved):

src/core/opamStd.mli
```diff
@@ -532,6 +532,9 @@ module Sys : sig
       OS) *)
   val path_sep: char
 
+  (** [true] if a character may be used in a filename. *)
```
rjbou, line 535: Maybe a little bit of context ?
```suggestion
  (** [true] if a character may be used in a filename, depending on the OS.
      For example, on Windows, `:` and `?` can't be in the name. *)
```

src/repository/opamDownload.ml
```diff
@@ -188,8 +188,18 @@ let download_as ?quiet ?validate ~overwrite ?compress ?checksum url dst =
...
+      let f c =
+        if OpamStd.Sys.is_valid_basename_char c then c else '_'
+      in
+      String.map f base
+    else
+      base
+  in
   let dst =
-    OpamFilename.(create dstdir (Base.of_string (OpamUrl.basename url)))
+    OpamFilename.(create dstdir (Base.of_string base))
```
rjbou, line 202: I'm wondering if we should have that check/change in `OpamFilename.Base` directly, like that we are sure that every file created is valid. But can we have that change (forbidden char -> `_`) for all files?

rjbou, line 202: In opam code, there is no collision with `_` change (as you pointed), but I'm wondering if we could have collision for library users.

dra27: I don't think we can ever automatically change it, but we may want to do things to cause better error messages. For example, if an opam file actually specifies `extra-source "zstd-detection?full_index=1.patch" {` then I don't think opam should ever do anything to make this work, but it perhaps fail with a better error message if this is encountered.

dra27: (the key point here is that this filename is internal to opam)

dra27: Possibly - on the plus-side, the fallout is limited to Windows, where the code would have been failing before!

rjbou (approved):
thanks!

PR #5922, opened 2024-04-11
Fix incorrect package installed state written on error

dra27:
If a package needs to be rebuilt because of upstream changes, an error in the rebuild causes a cache indicating that it's installed to be written. Something appears to be wrong in `OpamClient.parallel_apply`, but I haven't got to the bottom of exactly what.

What I have written is a test! Spotted while debugging #5921 - because errors during fetch also trigger this. Note in the commit history that erasing `.opam-switch/packages/cache` causes opam to do the right thing on the second upgrade.

### DIFF
--- a/tests/reftests/update-upgrade.test
+++ b/tests/reftests/update-upgrade.test
@@ -506,3 +506,38 @@ The following actions will be performed:
 -> removed   ongoing.dev
 -> installed ongoing.dev
 Done.
+### <REPO/packages/d-foo/d-foo.6/opam>
+opam-version: "2.0"
+build: [ "false" ]
+### opam update
+[NOTE] ongoing.dev has previously been updated with --working-dir, not resetting unless explicitly selected
+
+<><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
+[second] no changes from file://${BASEDIR}/REPO2
+[default] synchronised from file://${BASEDIR}/REPO
+
+<><> Synchronising development packages <><><><><><><><><><><><><><><><><><><><>
+[quux.dev5] synchronised (no changes)
+Now run 'opam upgrade' to apply any package updates.
+### opam upgrade
+The following actions will be performed:
+=== recompile 1 package
+  - recompile d-foo 6 [upstream or system changes]
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+[ERROR] The compilation of d-foo.6 failed at "false".
+
+
+
+
+<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
++- The following actions failed
+| - build d-foo 6
++- 
+- No changes have been performed
+# Return code 31 #
+### : Running opam upgrade a second time should do the same thing - the same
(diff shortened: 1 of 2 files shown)


dra27:
In the 2.2 project as I haven't yet determined if this is a regression or long-standing.

tests/reftests/update-upgrade.test
```diff
@@ -506,3 +506,51 @@ The following actions will be performed:
...
+
+
+
+
+<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
++- The following actions failed
+| - build d-foo 6
++- 
+- No changes have been performed
+# Return code 31 #
```
kit-ty-kate, line 538: To show more explicitly what the problem is:
```suggestion
# Return code 31 #
### opam upgrade
Nothing to do.
```

kit-ty-kate: If I'm guessing what the problem is wrongly, maybe there should be a comment in the test to tell what the cache file is about

dra27: Ah, the overall PR diff doesn't convey it, indeed. The point is that `opam upgrade` should repeatedly produce error 31 - we're trying to recompile a package which is going to fail. Commit-by-commit:
- 3030619036aed9336c781476e8536d26067560d8 has the second `opam upgrade` claiming "Nothing to do"
- d6d4167d4f586074a23f6d5f913fc3bd3d330159 is what the fix should then do to it (second attempt to install also fails)
- 85cd847a8de491257812d0321d337f16c76b21f4 shows that if we delete the cache file, then we get the desired behaviour (i.e. the cache is the problem)

kit-ty-kate: Maybe a comment could be added that says something like "test to make sure we return exit code 31"

dra27: A comment? A comment? Done 🙂

kit-ty-kate: Thanks!

tests/reftests/update-upgrade.test
```diff
@@ -506,3 +506,56 @@ The following actions will be performed:
...
+### opam update
+[NOTE] ongoing.dev has previously been updated with --working-dir, not resetting unless explicitly selected
+
+<><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
+[second] no changes from file://${BASEDIR}/REPO2
+[default] synchronised from file://${BASEDIR}/REPO
+
+<><> Synchronising development packages <><><><><><><><><><><><><><><><><><><><>
+[quux.dev5] synchronised (no changes)
+Now run 'opam upgrade' to apply any package updates.
```
rjbou, line 521: Is this `opam update` output needed?

kit-ty-kate: I think currently yes but we could remove it if the new version of `d-foo.6` was introduced using `<pkg:d-foo.6>` instead of `<REPO/...>`

kit-ty-kate:
Superseded by https://github.com/ocaml/opam/pull/6213

PR #5927, opened 2024-04-18
Relax warning 41 for package variables guarded by a :installed filter

dra27:
The compiler packages contain the following snippet:

```
depends: [
  "flexdll" {>= "0.42" & os = "win32"}
]
build: [
  [
    "./configure"
    "--with-flexdll=%{flexdll:share}%" {flexdll:installed}
  ]
]
```
but the use of `flexdll:share` triggers lint warning 41 because the flexdll dependency is guarded by `os`. However, this warning is a bit obtuse given the argument is guarded by `flexdll:installed` (and `:installed` explicitly doesn't trigger lint warning 41).

This PR alters lint warning 41 so that when it evaluates the list of variables used in commands, it scans the filters for instances of `package:installed` and then performs a partial evaluation of the filter with that `package:installed` variable explicitly set to false (but nothing else set in the environment). If the filter reduces to `false`, then all instances of `package:var` guarded by the filter are ignored.

The first commit adds a test case showing the present behaviour, then the fix shows two of the packages warned being eliminated.

### DIFF
--- a/src/state/opamFileTools.ml
+++ b/src/state/opamFileTools.ml
@@ -174,19 +174,35 @@ let unguarded_commands_variables commands =
     let guarded_packages, filter_variables =
       unguarded_packages_from_filter guarded_packages filter
     in
-    (filter_guarded (OpamFilter.simple_arg_variables argument) guarded_packages)
-    @ filter_variables
+    let variables_from_arguments =
+      filter_guarded (OpamFilter.simple_arg_variables argument) guarded_packages
+    in
+    guarded_packages, variables_from_arguments @ filter_variables
   in
-  let unguarded_command_variables (command, filter) =
-    let guarded_packages, filter_variables =
+  let unguarded_command_variables guarded_packages (command, filter) =
+    let filter_guarded_packages, filter_variables =
       unguarded_packages_from_filter OpamPackage.Name.Set.empty filter
     in
-    let add_argument acc argument =
-      unguarded_argument_variables guarded_packages argument @ acc
+    let add_argument (guarded_packages, acc) argument =
+      let guarded_packages, unguarded_variables =
+        unguarded_argument_variables guarded_packages argument
+      in
+      guarded_packages, unguarded_variables @ acc
     in
-    List.fold_left add_argument filter_variables command
+    let command_guarded_packages, unguarded_variables =
+      List.fold_left add_argument (filter_guarded_packages, filter_variables)
+        command
+    in
+    OpamPackage.Name.Set.union guarded_packages command_guarded_packages,
+    unguarded_variables
   in
-  List.fold_left (fun acc c -> unguarded_command_variables c @ acc) [] commands
+  let f (guarded_packages, acc) c =
+    let guarded_packages, unguarded_variables =
+      unguarded_command_variables guarded_packages c
+    in
+    guarded_packages, (unguarded_variables @ acc)
+  in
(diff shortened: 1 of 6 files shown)


dra27:
@kit-ty-kate rightly points out a timing issue which concretely invalidates one of the test cases. The filter above should repeat the guard of the dependency, i.e. instead of:

```
    "--with-flexdll=%{flexdll:share}%" {flexdll:installed}
```
it should be:
```
    "--with-flexdll=%{flexdll:share}%" {flexdll:installed & os = "win32"}
```
this also concretely rules out this line from the test:
```
  ["%{foo:share}%%{bar:share}%" {bar:installed}] {foo:installed}
```
since `bar` is not referred to in the dependency formula in any way.

I'm wondering if this is in fact more related to #5928 - in order to use the variables, we need _both_ that it's unequivocally guarded by `:installed` (as here) and that the atom appears in some way in _either_ `depends` or `depopt` (to enforce the ordering).

Regardless, the test case of `bar` is a deficiency of this PR, so I'm putting it back to draft status for now.

dra27:
Additional commit added which tweaks the check slightly. Now:
- `foo:share` still doesn't trigger warning 41 because it's guarded by `foo:installed` **and** `foo` is referred to in `depends`
- Conversely, `bar:share` now triggers warning 41 because although it's guarded by `bar:installed`, `bar` is **not** referred to at all in `depends` on `depopts`
- `baz:share` continues to trigger warning 41 because it's neither protected by `bar:installed` nor referred to in `depends`/`depopts`
- But note that it now triggers warning 41 even for the `:installed` variable if the package has not been referred to `depends`/`depopts` (which I think is better than the present behaviour of always allowing `:installed` to be referred to, but that's open to further debate!)

src/state/opamFileTools.ml
```diff
@@ -119,6 +119,86 @@ let map_all_filters f t =
...
+   ["--with-foo"] {foo:installed & foo:bar != "baz"} not to trigger warning 41
+   if the package is not explicitly depended on. *)
+
+let is_installed_variable filter guarded_packages v =
+  match OpamVariable.Full.package v with
+  | None -> guarded_packages
+  | (Some name) as package ->
+    if OpamVariable.to_string (OpamVariable.Full.variable v) = "installed" then
+      let env v =
+        if OpamVariable.Full.package v = package &&
```
kit-ty-kate, line 137:
```suggestion
        if OpamPackage.equal (OpamVariable.Full.package v) package &&
```

kit-ty-kate (approved):

src/state/opamFileTools.ml
```diff
@@ -119,6 +119,86 @@ let map_all_filters f t =
...
+   ["--with-foo"] {foo:installed & foo:bar != "baz"} not to trigger warning 41
+   if the package is not explicitly depended on. *)
+
+let is_installed_variable filter guarded_packages v =
+  match OpamVariable.Full.package v with
+  | None -> guarded_packages
+  | (Some name) as package ->
+    if OpamVariable.to_string (OpamVariable.Full.variable v) = "installed" then
+      let env v =
+        if OpamPackage.equal (OpamVariable.Full.package v) package &&
```
kit-ty-kate, line 137:
```suggestion
        if Option.equal OpamPackage.Name.equal (OpamVariable.Full.package v) package &&
```

rjbou (approved):
Thanks!

Issue #5839, opened 2024-02-15
Cygwin depext doesn't report errors for non-existent packages


dra27:
Cygwin setup doesn't return an error code for a package name which doesn't exist, there's simply a line in the console output:

```
Package 'does-not-exist' not found.
```

We should already have the information available to be able to display an error at this stage:

```
C:\Users\DRA>opam install conf-zstd
The following actions will be performed:
=== install 2 packages
  ∗ conf-mingw-w64-zstd-i686 1     [required by conf-zstd]
  ∗ conf-zstd                1.3.8

Proceed with ∗ 2 installations? [y/n] y

The following system packages will first need to be installed:
    xxx-mingw64-i686-zstd
```

instead of:

```
<><> Handling external dependencies <><><><><><><><><><><><><><><><><><><><>  🐫

+ C:\Users\DRA\AppData\Local\opam\.cygwin\setup-x86_64.exe "--root" "C:\\Users\\DRA\\AppData\\Local\\opam\\.cygwin\\root" "--quiet-mode" "hidden" "--no-shortcuts" "--no-startmenu" "--no-desktop" "--no-admin" "--packages" "xxx-mingw64-i686-zstd" "--symlink-type" "native" "--upgrade-also" "--only-site" "--no-version-check" "--site" "https://cygwin.mirror.constant.com/" "--local-package-dir" "C:\\Users\\DRA\\AppData\\Local\\opam\\.cygwin\\cache"
- Starting cygwin install, version 2.930
- User has NO backup/restore rights
- User has NO symlink creation right
- Current Directory: C:\Users\DRA\AppData\Local\opam\.cygwin\cache
- root: C:\Users\DRA\AppData\Local\opam\.cygwin\root user
- Changing gid back to original
- Selected local directory: C:\Users\DRA\AppData\Local\opam\.cygwin\cache
- net: Preconfig
- site: https://cygwin.mirror.constant.com/
- Package 'xxx-mingw64-i686-zstd' not found.
- solving: 0 tasks, update: yes, use test packages: no
- solving: 0 tasks, update: no, use test packages: no
- Augmented Transaction List: is empty
- Registry value set: HKEY_CURRENT_USER\Software\Cygwin\setup\rootdir = "C:\Users\DRA\AppData\Local\opam\.cygwin\root"
- running: C:\Users\DRA\AppData\Local\opam\.cygwin\root\bin\dash.exe "/etc/postinstall/0p_000_autorebase.dash"
[...]

PR #5930, opened 2024-04-23
Add CLI option to specify additional packages for internal Cygwin.

### LINKED ISSUE #5834: Add CLI option to specify additional packages for internal Cygwin at `opam init`
This is to mitigate #5833, principally for CI, as it would allow a workflow which _knows_ it's creating a mingw-w64 switch to speed up the init process by adding `mingw64-x86_64-gcc-core` or `mingw64-i686-gcc-core` to the installation when it is created (indeed, a CI workflow might choose to add additional depext packages).

At a rough punt, this could be either a dedicated `--cygwin-internal-packages=package1,package2` but, given that that would need checking as it would conflict with `--cygwin-local-install`, it could just be made as an optional argument for `--cygwin-internal-install`?

In principle, we might want to consider having general maintenance commands for the internal Cygwin installation (although it's possible to do that by hand simply by running the setup program against the root), but that's a much bigger feature and the whole thing is still experimental anyway.

moyodiallo:
Fixes #5834.

Instead of reusing `--cygwin-internal-install`, it is more suitable to have new option (`--cygwin-packages=CYGWIN_PACKAGES`), to avoid ending up checking the exclusion between `--cygwin-internal-install` and `--cygwin-local-install`.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -20,6 +20,7 @@ users)
 ## Plugins
 
 ## Init
+  * ◈ New option `opam init --cygwin-extra-packages=CYGWIN_PKGS --cygwin-internal-install`, to specify additional packages for internal Cygwin [#5930 @moyodiallo - fix #5834]
 
 ## Config report
 
@@ -123,6 +124,7 @@ users)
 
 # API updates
 ## opam-client
+  * `OpamClient.init` and `OpamClient.reinit`: now can have additional cygwin packages to install [#5930 @moyodiallo]
 
 ## opam-repository
 
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -850,15 +850,19 @@ let windows_checks ?cygwin_setup ?git_location config =
            OpamFilename.(Dir.to_string (dirname_dir (dirname_dir cygbin))));
     config
   in
-  let install_cygwin_tools () =
-    let packages =
+  let install_cygwin_tools packages =
+    let default_packages =
       match OpamSystem.resolve_command "git" with
       | None -> OpamInitDefaults.required_packages_for_cygwin
       | Some _ ->
         List.filter (fun c -> not OpamSysPkg.(equal (of_string "git") c))
           OpamInitDefaults.required_packages_for_cygwin
     in
-    OpamSysInteract.Cygwin.install ~packages
+    (* packages comes last so that the user can override any potential version
+       constraints in default_packages (although, with the current version of
+       setup, and with the list of default_packages in OpamInitDefaults, this at
+       present doesn't matter too much). *)
+    OpamSysInteract.Cygwin.install ~packages:(default_packages @ packages)
   in
   let header () = OpamConsole.header_msg "Unix support infrastructure" in
 
@@ -956,7 +960,7 @@ let windows_checks ?cygwin_setup ?git_location config =
         match prompt () with
         | `Abort -> OpamStd.Sys.exit_because `Aborted
         | `Internal ->
-          let cygcheck = install_cygwin_tools () in
+          let cygcheck = install_cygwin_tools [] in
           let config = success cygcheck in
           config
         | `Specify ->
@@ -975,7 +979,7 @@ let windows_checks ?cygwin_setup ?git_location config =
   let config =
     match cygwin_setup with
     | Some `no -> config
-    | (Some (`internal | `default_location | `location _) | None)
+    | (Some (`internal _ | `default_location | `location _) | None)
       as cygwin_setup ->
       if OpamSysPoll.os env = Some "win32" then
         match OpamSysPoll.os_distribution env with
@@ -995,7 +999,7 @@ let windows_checks ?cygwin_setup ?git_location config =
              header ();
              let cygcheck =
                match setup with
-               | `internal -> install_cygwin_tools ()
+               | `internal pkgs -> install_cygwin_tools pkgs
                | (`default_location | `location _ as setup) ->
                  let cygroot =
                    match setup with
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -319,6 +319,16 @@ let init cli =
     else
       Term.const `none
   in
+  let cygwin_extra_packages =
+    if Sys.win32 then
+      mk_opt ~cli (cli_from ~experimental:true cli2_2)
+        ["cygwin-extra-packages"] "CYGWIN_PACKAGES"
+        "Specify additional packages to install \
+         with $(b,--cygwin-internal-install)"
+        Arg.(some (list string)) None
+    else
+      Term.const None
+  in
   let cygwin_location =
     if Sys.win32 then
       mk_opt ~cli (cli_from ~experimental:true cli2_2)
@@ -351,7 +361,8 @@ let init cli =
       interactive update_config completion env_hook no_sandboxing shell
       dot_profile_o compiler no_compiler config_file no_config_file reinit
       show_opamrc bypass_checks
-      cygwin_internal cygwin_location git_location no_git_location
+      cygwin_internal cygwin_location cygwin_extra_packages
+      git_location no_git_location
       () =
     apply_global_options cli global_options;
     apply_build_options cli build_options;
@@ -408,17 +419,28 @@ let init cli =
         OpamStd.Sys.guess_dot_profile shell >>| OpamFilename.of_string)
     in
     let cygwin_setup =
-      match cygwin_internal, cygwin_location with
-      | `internal, Some _ ->
+      let bad_arg arg1 arg2 =
         OpamConsole.error_and_exit `Bad_arguments
-          "Options --cygwin-internal-install and \
-           --cygwin-location are incompatible";
-      | `no, Some _ ->
-        OpamConsole.note "Ignoring argument --cygwin-location";
-        Some `no
-      | `none, None -> None
-      | (`default_location | `none), Some dir -> Some (`location dir)
-      | (`internal | `default_location | `no) as setup, None -> Some setup
+          "Options --%s and --%s are incompatible"
+          arg1 arg2
+      in
+      match cygwin_internal, cygwin_location, cygwin_extra_packages with
+      | `internal, Some _, _ ->
+        bad_arg "cygwin-internal-install" "cygwin-location"
+      | `default_location, _, Some _ ->
+        bad_arg "cygwin-local-install" "cygwin-extra-packages"
+      | `no, Some _, _ ->
+        bad_arg "no-cygwin-setup" "cygwin-location"
+      | `no,  _, Some _ ->
+        bad_arg "no-cygwin-setup" "cygwin-extra-packages"
+      | `none, Some _, Some _ ->
+        bad_arg "cygwin-location" "cygwin-extra-packages"
+      | (`internal | `none), None, pkgs ->
+        Some (`internal
+                (OpamStd.Option.default [] pkgs
+                 |> List.map OpamSysPkg.of_string))
+      | (`default_location | `none), Some dir, None -> Some (`location dir)
+      | (`default_location | `no) as setup, None, None -> Some setup
     in
     let git_location =
       match git_location, no_git_location with
@@ -527,7 +549,8 @@ let init cli =
           $setup_completion $env_hook $no_sandboxing $shell_opt cli
           cli_original $dot_profile_flag cli cli_original $compiler
           $no_compiler $config_file $no_config_file $reinit $show_default_opamrc
-          $bypass_checks $cygwin_internal $cygwin_location $git_location $no_git_location)
+          $bypass_checks $cygwin_internal $cygwin_location $cygwin_extra_packages
+          $git_location $no_git_location)
 
 (* LIST *)
 let list_doc = "Display the list of available packages."
--- a/src/client/opamClient.mli
+++ b/src/client/opamClient.mli
@@ -28,7 +28,7 @@ val init:
   ?env_hook:bool ->
   ?completion:bool ->
   ?check_sandbox:bool ->
-  ?cygwin_setup: [ `internal | `default_location | `location of dirname | `no ] ->
+  ?cygwin_setup: [ `internal of OpamSysPkg.t list | `default_location | `location of dirname | `no ] ->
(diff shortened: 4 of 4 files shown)


src/client/opamClient.ml
```diff
@@ -996,6 +996,7 @@ let windows_checks ?cygwin_setup ?git_location config =
              let cygcheck =
                match setup with
                | `internal -> install_cygwin_tools ()
+               | `internal_with_pkgs _ -> install_cygwin_tools ()
```
kit-ty-kate, line 999: I believe the arguments should be actually used here, otherwise they are never used anywhere and the option doesn't do anything besides enabling internal cygwin

src/client/opamCommands.ml
```diff
@@ -308,7 +308,7 @@ let init cli =
       mk_vflag ~cli `none [
         cli_from ~experimental:true cli2_2,
         `internal, ["cygwin-internal-install"],
-        "Let opam setup and manage an internal Cygwin install (recommended)";
+        "Let opam setup and manage an internal Cygwin install";
```
kit-ty-kate, line 311: This should be kept

kit-ty-kate (commented):
I like this approach, once the code is fixed and works it should be good to go for me at least

src/client/opamClient.ml
moyodiallo: Of course it is here.

src/client/opamCommands.ml
moyodiallo: I don't get the context. Why removing it here ?

kit-ty-kate: you're removing something (the ` (recommended)` part) for some reason and I think it should be kept as-is, as it was just added in https://github.com/ocaml/opam/pull/5903

src/client/opamClient.ml
```diff
@@ -850,14 +850,18 @@ let windows_checks ?cygwin_setup ?git_location config =
            OpamFilename.(Dir.to_string (dirname_dir (dirname_dir cygbin))));
     config
   in
+  let packages =
```
kit-ty-kate, line 853:
```suggestion
  let default_packages =
```
just to be more explicit

src/client/opamClient.ml
```diff
@@ -850,14 +850,18 @@ let windows_checks ?cygwin_setup ?git_location config =
...
-    let packages =
-      match OpamSystem.resolve_command "git" with
-      | None -> OpamInitDefaults.required_packages_for_cygwin
-      | Some _ ->
-        List.filter (fun c -> not OpamSysPkg.(equal (of_string "git") c))
-          OpamInitDefaults.required_packages_for_cygwin
-    in
+    OpamSysInteract.Cygwin.install ~packages
+  in
+  let install_cygwin_tools_with_pkgs ~pkgs =
```
kit-ty-kate, line 863: why not simply merge this new function with the existing `install_cygwin_tools` function?

src/client/opamCommands.ml
moyodiallo: Oh sorry I missed that, I thought is your proposition to remove it. This was not intended, thanks.

src/client/opamClient.ml
moyodiallo: Ok, that's fair.

kit-ty-kate (approved):

dra27:
I'm not sure what I'm missing, but why do we need a new constructor instead of just adding the list to ``` `internal```? Similarly, I'm not sure why having to display an error with `--cygwin-packages=CYGWIN_PACKAGES` specified on it own is better than `--cygwin-internal-install=[CYGWIN_PACKAGES]`?

moyodiallo:
> I'm not sure what I'm missing, but why do we need a new constructor instead of just adding the list to `` `internal ``? Similarly, I'm not sure why having to display an error with `--cygwin-packages=CYGWIN_PACKAGES` specified on it own is better than `--cygwin-internal-install=[CYGWIN_PACKAGES]`?

`--cygwin-internal-install`, `--cygwin-local-install` flags are already grouped together and we could let that as it is. Doing `--cygwin-internal-install=[CYGWIN_PACKAGES]` is going split them and make them separate options.

src/client/opamCommands.ml
```diff
@@ -408,8 +416,16 @@ let init cli =
         OpamStd.Sys.guess_dot_profile shell >>| OpamFilename.of_string)
     in
     let cygwin_setup =
+      let cygwin_internal =
+        match cygwin_internal, cygwin_packages with
+        | `internal, Some pkgs -> `internal_with_pkgs (List.map OpamSysPkg.of_string pkgs)
```
dra27, line 421: This should be:
```suggestion
        | (`internal | `none), Some pkgs -> `internal_with_pkgs (List.map OpamSysPkg.of_string pkgs)
```
(cf. the ```(`default_location | `none)``` case below, but...

src/client/opamCommands.ml
```diff
@@ -408,8 +416,16 @@ let init cli =
...
+        | _, Some _ ->
+          OpamConsole.error_and_exit `Bad_arguments
+            "Option --cygwin-packages is only compatible with --cygwin-internal-install option";
```
dra27, line 424: Note that `--cygwin-location` below is _ignored_ if ``` `none``` is given, but I think this is actually better (with the handling for ``` `none``` above) and it would be better to change the warning below to an error, but...

src/client/opamCommands.ml
```diff
@@ -408,8 +416,16 @@ let init cli =
...
+        | _ as setup, None -> setup
+      in
       match cygwin_internal, cygwin_location with
```
dra27, line 427: Rather than the above, I'd suggest:
```ocaml
      match cygwin_internal, cygwin_location, cygwin_packages with
      | `internal, Some _, _ ->
          "Options --cygwin-internal-install and \
           --cygwin-location are incompatible";
      | `no, Some _, _ ->
          "Options --no-cygwin-setup and \
           --cygwin-location are incompatible";
      | (`none | `internal), None, pkgs -> Some (`internal (Option.default [] pkgs))
      | `none, None -> None
      | (`default_location | `none), Some dir -> Some (`location dir)
      | (`default_location | `no) as setup, None -> Some setup
```
The key point is that `--cygwin-location` with nothing else already _implies_ `--cygwin-local-install`, so similarly `--cygwin-packages` with nothing else should _imply_ `--cygwin-internal-install`

src/client/opamClient.ml
```diff
@@ -850,15 +850,15 @@ let windows_checks ?cygwin_setup ?git_location config =
...
+  let install_cygwin_tools packages =
+    let default_packages =
       match OpamSystem.resolve_command "git" with
       | None -> OpamInitDefaults.required_packages_for_cygwin
       | Some _ ->
         List.filter (fun c -> not OpamSysPkg.(equal (of_string "git") c))
           OpamInitDefaults.required_packages_for_cygwin
     in
-    OpamSysInteract.Cygwin.install ~packages
+    OpamSysInteract.Cygwin.install ~packages:(default_packages @ packages)
```
dra27, line 861: It's worth a comment here that `packages` comes last, because it allows the command line to override version constraints in the default (were we to put any)

dra27 (commented):
I'm not sure that the code grouping with `vflag` is a compelling reason to guide the CLI, but it occurred to me on the ride in this morning that my suggestion is not good for a different reason: if `--cygwin-internal-install` takes an optional argument then `opam init --cygwin-internal-install git+https://github.com/ocaml/opam-repository.git` changes meaning, and that's bad.

My original issue, as ever, has a bad name suggestion... `--cygwin-packages` implies we are suggesting the _full_ list of packages, where we are in reality specifying _additional_ packages. Perhaps `--cygwin-extra-packages` is a better name?

kit-ty-kate:
> Perhaps `--cygwin-extra-packages` is a better name?

While it's a bit longer to write, i think i agree with that.

src/client/opamCommands.ml
moyodiallo: This sounds good.

master_changes.md
```diff
@@ -18,6 +18,7 @@ users)
 ## Plugins
 
 ## Init
+  * ◈ New option `opam init --cygwin-packages=CYGWIN_PKGS --cygwin-internal-install`, to specify additional packages for internal Cygwin [#5930 @moyodiallo - fix #5834]
```
kit-ty-kate, line 21:
```suggestion
  * ◈ New option `opam init --cygwin-extra-packages=CYGWIN_PKGS --cygwin-internal-install`, to specify additional packages for internal Cygwin [#5930 @moyodiallo - fix #5834]
```

kit-ty-kate (approved):
LGTM

src/client/opamCommands.ml
```diff
@@ -408,17 +416,31 @@ let init cli =
...
            --cygwin-location are incompatible";
-      | `no, Some _ ->
-        OpamConsole.note "Ignoring argument --cygwin-location";
-        Some `no
-      | `none, None -> None
-      | (`default_location | `none), Some dir -> Some (`location dir)
-      | (`internal | `default_location | `no) as setup, None -> Some setup
+      | `default_location, _, Some _ ->
+        OpamConsole.error_and_exit `Bad_arguments
+          "Options --cygwin-location-install and \
```
dra27, line 426:
```suggestion
          "Options --cygwin-location and \
```

dra27 (commented):
A few typos, but otherwise looks good to me, thanks

src/client/opamCommands.ml
moyodiallo: Indeed but it is `local` we need to change to have `--cygwin-local-install`.

src/client/opamCommands.ml
```diff
@@ -408,17 +416,31 @@ let init cli =
...
-      | (`default_location | `none), Some dir -> Some (`location dir)
-      | (`internal | `default_location | `no) as setup, None -> Some setup
+      | `default_location, _, Some _ ->
+        OpamConsole.error_and_exit `Bad_arguments
+          "Options --cygwin-local-install and \
+           --cygwin-extra-packages are incompatible";
+      | `no, Some _, _ ->
+        OpamConsole.error_and_exit `Bad_arguments
+          "Options --no-cygwin-setup and \
+           --cygwin-location are incompatible";
```
rjbou, line 431: In previous code, the behaviour was : _if `--no-cygwin-setup` is given, all other argument are ignored_. If there a reason to change that to bad arguments errors ?

src/client/opamCommands.ml
```diff
@@ -351,7 +359,7 @@ let init cli =
       interactive update_config completion env_hook no_sandboxing shell
       dot_profile_o compiler no_compiler config_file no_config_file reinit
       show_opamrc bypass_checks
-      cygwin_internal cygwin_location git_location no_git_location
+      cygwin_internal cygwin_location git_location no_git_location cygwin_extra_packages
```
rjbou, line 362: It is better to group the arguments : cygwin ones and git ones

rjbou (commented):
Thanks! I'm posted few comments, most of them are implemented in separate commits to squash once validated. Remain the behaviour change question.

src/client/opamCommands.ml
```diff
@@ -408,17 +416,31 @@ let init cli =
...
+      | `no,  _, Some _ ->
+        OpamConsole.error_and_exit `Bad_arguments
+          "Options --no-cygwin-setup and \
+           --cygwin-extra-packages are incompatible";
+      | `none, Some _, Some _ ->
+        OpamConsole.error_and_exit `Bad_arguments
+          "Options --cygwin-location and \
+           --cygwin-extra-packages are incompatible";
+      | (`internal | `none), None, pkgs ->
+        Some (`internal (OpamStd.Option.default [] pkgs |> List.map OpamSysPkg.of_string))
```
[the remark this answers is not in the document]
rjbou: my mistake

src/client/opamCommands.ml
dra27: It was from https://github.com/ocaml/opam/pull/5930#discussion_r1579012410 (and for discussion!) - I think the previous behaviour is "bad"... if the user has specified two conflicting options, we should display an error.

rjbou: fair enough

src/client/opamCommands.ml
```diff
@@ -408,17 +419,33 @@ let init cli =
...
+      | `none, Some _, Some _ ->
+        OpamConsole.error_and_exit `Bad_arguments
+          "Options --cygwin-location and \
+           --cygwin-extra-packages are incompatible"
+      | (`internal | `none), None, pkgs ->
+        Some (`internal
+                (OpamStd.Option.default [] pkgs
+                 |> List.map OpamSysPkg.of_string))
+      | (`default_location | `none), Some dir, None -> Some (`location dir)
+      | (`default_location | `no) as setup, None, None -> Some setup
```
rjbou, line 448:
```suggestion
      let bad_arg arg1 arg2 =
        OpamConsole.error_and_exit `Bad_arguments
          "Options --%s and --%s are incompatible"
          arg1 arg2
      in
      match cygwin_internal, cygwin_location, cygwin_extra_packages with
      | `internal, Some _, _ ->
        bad_arg "cygwin-internal-install" "cygwin-location"
      | `default_location, _, Some _ ->
        bad_arg "cygwin-local-install" "cygwin-extra-packages"
      | `no, Some _, _ ->
        bad_arg "no-cygwin-setup" "cygwin-location"
      | `no,  _, Some _ ->
        bad_arg "no-cygwin-setup" "cygwin-extra-packages"
      | `none, Some _, Some _ ->
        bad_arg "cygwin-location" "cygwin-extra-packages"
```

rjbou (approved):
Except the proposal, LGTM! Thanks!

moyodiallo:
I haven't seen no mention about testing. Did someone test it ?

rjbou:
I forgot to note but I've tested it when I reviewed installing packages (non installed, already installed, unknown) and some argument incompatibility.
@kit-ty-kate is doing further tests.

kit-ty-kate:
worked fine. It suffers from https://github.com/ocaml/opam/issues/5839 and https://github.com/ocaml/opam/issues/5952 but those are separate issues. Thanks a lot!

Issue #5926, opened 2024-04-18
Unexpected behaviour when prepending/appending to initially empty variables multiple times


gridbugs:
When a variable is initially unset or empty, appending to it and then prepending to it using the colon operators (`:=` and `=:`) causes an unexpected leading colon in the result. Prepending to an unset variable followed by appending to it produces the same unexpected result.

Here is a package reproducing the problem:
```
opam-version: "2.0"

build-env: [

 [ X := "a" ]
# X should be "a:"
 [ X =: "b" ]
# X shoulde be "a:b"
# (assuming opam doesn't add a separator when appending to a variable which ends with a separator)
# X unexpectedly has the value ":a:b"

 [ Y =: "a" ]
# Y should be ":a"
 [ Y := "b" ]
# Y should be "b:a"
# (assuming opam doesn't add a separator when prepending to a variable which starts with a separator)
# Y unexpectedly has the value ":b:a"

]

install: [
    [ "bash" "-c" "echo X=$X >> %{lib}%/foo" ]
    [ "bash" "-c" "echo Y=$Y >> %{lib}%/foo" ]
]
```

Reproduce by running `opam install ./foo.opam && cat _opam/lib/foo` which will produce the output:
```
X=:a:b
Y=:b:a
```

But according to the documentation the output should be:
```
X=a:b
Y=b:a
```

I'm on linux. Reproduced on both 2.1.5 and opam-2.2.0-beta2.

dra27:
As in #5935, cross-referencing:
- #3886 (fixing #3878, #2870 and #2255)
- #3404 (fixing #3391)
- #3325 (fixing #3319)

The documentation for `=:` and `:=` I think is not correct since #3404. I think the semantics of `:=` and `=:` should be as given for empty/unset values but also that they both preserve trailing and leading `:` in the current value.

Given those semantics:
- `[ X := "a" ]` with `X` empty gives `a:` (current behaviour is **correct**)
- `[ X =: "b" ]` with `X` set to `a:` gives `a:b:`, adding `b` at the end, but preserving the trailing colon (current behaviour is **incorrect** as `X=:a:b`)
- `[ Y =: "a" ]` with `Y` empty gives `:a` (current behaviour is **correct**)
- `[ Y := "b" ]` with `Y` set to `:a` gives `:b:a`, adding `b` at the start, but preserving the leading colon (current behaviour is **correct**)

[...]

Issue #5925, opened 2024-04-18
Using the `+=` opetator in `build-env` to prepend an empty (but set) variable adds a trailing seperator


gridbugs:
This is contrary to the documentation [here](https://opam.ocaml.org/doc/Manual.html#Environment-updates):

> += or := prepend. They differ when the variable is unset of empty, where := adds a trailing separator.

To reproduce, put the following in a file named `foo.opam`:
```
opam-version: "2.0"

build-env: [
 [ FOO = "" ]
 [ FOO += "foo" ]
]

install: [
 [ "bash" "-c" "echo FOO=$FOO >> %{lib}%/x" ]
]
```

Then run:
```
opam switch create .
opam install ./foo.opam
cat _opam/lib/x
```

The output will be be:
```
FOO=foo:
```

But according to the documentation the output should be:
```
FOO=foo
```

I'm on linux. Reproduced on both 2.1.5 and opam-2.2.0-beta2.

dra27:
OK, I've been looking at this in more detail with the preparatory work in #5935. Cross-referencing a slew of other previous issues in this area to be sure we don't accidentally regress them (because the fixes predate the reftests infrastructure):
- #3886 (fixing #3878, #2870 and #2255)
- #3404 (fixing #3391)
- #3325 (fixing #3319)

With that history paged in, I still think this issue is unequivocally a bug (i.e. the docs are correct and opam isn't adhering to them). It might be relevant to see whether OPAM 1.2.2 was doing this, in terms of assessing whether fixing this may cause a regression.

Issue #5838, opened 2024-02-14
Environment revert not working correctly for Unix-like variables on Windows


dra27:
```
C:\Users\DRA>opam switch create manpath --empty
# Run for /f "tokens=*" %i in ('opam env --switch=manpath') do @%i to update the current shell environment

C:\Users\DRA>opam env
set OPAM_LAST_ENV=C:\Users\DRA\AppData\Local\opam\manpath\.opam-switch\last-env\env-3fd90d6beff7ec39310fda00a45ad218-0
set OPAM_SWITCH_PREFIX=C:\Users\DRA\AppData\Local\opam\manpath
set "MANPATH=:"C:\Users\DRA\AppData\Local\opam\manpath\man""
set "Path=C:\Users\DRA\AppData\Local\opam\manpath\bin;..."

C:\Users\DRA>for /f "tokens=*" %i in ('opam env') do @%i

C:\Users\DRA>echo %MANPATH%
:"C:\Users\DRA\AppData\Local\opam\manpath\man"

C:\Users\DRA>opam env
set OPAM_LAST_ENV=C:\Users\DRA\AppData\Local\opam\manpath\.opam-switch\last-env\env-3fd90d6beff7ec39310fda00a45ad218-0
set OPAM_SWITCH_PREFIX=C:\Users\DRA\AppData\Local\opam\manpath
set "MANPATH="C:\Users\DRA\AppData\Local\opam\manpath\man":"C:\Users\DRA\AppData\Local\opam\manpath\man""
set "Path=C:\Users\DRA\AppData\Local\opam\manpath\bin;..."

C:\Users\DRA>opam switch
#  switch   compiler  description
→  manpath            manpath
```

With #5837, we now get the correct warning:

```
C:\Users\DRA>opam switch --debug
00:00.041  CLI                    Parsing CLI version 2.2
00:00.042  GSTATE                 LOAD-GLOBAL-STATE @ C:\Users\DRA\AppData\Local\opam
00:00.043  SWITCH                 list
#  switch   compiler  description
→  manpath            manpath
00:00.044  ENV                    Not up-to-date env variables: [MANPATH]

[WARNING] The environment is not in sync with the current switch.
          You should run: for /f "tokens=*" %i in ('opam env') do @%i
```

dra27:
[...]

PR #5935, opened 2024-04-24
Fixes to reversal of environment updates

dra27:
Fixes for #5838, #5925 and #5926. This PR has become quite an epic rabbithole, but the fixes and changes are very, very intertwined. It is possible to separate the final fix for #5926 to a subsequent PR, but the final fix itself is actually quite small.

The first commit adds a new `sort` operation to the reftest language. While working on this, the stability of the output of `opam env` and the stability of the output when using `unordered` made making the true effect of each commit tricky. I'm still not sure that `unordered` is behaving as well as it should when there's actually a change to the output (i.e. when one line changes, it's allowing some unchanged lines to be re-ordered noisily in the diff).

I then add a whole range of tests:
- Since opening the original issue for `MANPATH`, I noticed that `CAML_LD_LIBRARY_PATH` wasn't reverting correctly on Windows either, and it turns out that's because opam wasn't correctly handling updates with forward slashes (i.e. `CAML_LD_LIBRARY_PATH += "%{lib}%/stublibs` expands to `%OPAMROOT%\switch\lib/stublibs` but `opam env` normalises the `/stublibs` to `\stublibs`. When reverting, this normalisation is missing.
- A test for #5838 ensures MANPATH is actually reverted
- A test for #4861 ensures an update containing multiple directories and is then extended to include a quoted directory
- A test for #5925 tests `+=`, `=+`, `:=` and `=:` applied to a variable set to `""` in the same `setenv` block and shows `+=` and `=+` misbehaving.
- A comprehensive test (!!) for #5926 tests those four operators on:
  - `a:` and `:a`, both set in the environment, and synthesised by using `:=` or `=:` on an unset variable
  - `a1::a2` (a "middle" blank) entry
  - A variable set to a single `:`

There have been a lot of issues and PRs in the past to do with the treatment of `:=` and `=:` principally where `MANPATH` is concerned. I'm relatively confident that the approach for this before has not been strictly correctly (and is the cause of the issue being seen in #5926). For both `PATH` (as defined by Posix) and `MANPATH` (as defined by the tool itself), an unset value, an empty value and the value `:` are all equivalent. For `PATH`, that means the current directory and all of these entries are literally equivalent to `PATH=.`. For `MANPATH`, those three entries mean "use the default search path".

Now, as far as I can see the previous discussions have been about ensuring that `MANPATH` has a leading or trailing `:`, to ensure that opam never hides the default path. Thus, if `MANPATH` is not set, then we have `MANPATH =: "%{man}%"` giving us `:/home/dra/.opam/default/man`. However, the focus should not be on having a leading or trailing `:`, the focus should be on ensuring that a "blank" entry is created if needed (this is the rule for `FOO := "a"` and `FOO =: "a"` are `a:` or `:a` respectively if `FOO` is unset or empty) and but then _for all updates_ separately ensuring that the blank entry is _preserved_.

i.e. if `FOO` is `:a` then _both_ `FOO += "b"` and `FOO := "b"` should set `FOO` to `b::a`

The actual fixes then proceed as:
- The `CAML_LD_LIBRARY_PATH` issue arises from what appears to be an incorrect comment in `OpamEnv.apply_op_zip` indicating that the argument is already transformed, which may suggest why `OpamEnv.reverse_env_update` was missing the transformation. Essentially, the transformation ensures that `opam env --revert` reverts the same value as it applied in `opam env`
- In `OpamEnv.split_var`, the cases for splitting using the quoted rules were the wrong way around. Fixing that regresses #4861, though
- If an environment variable is empty or unset, opam internally synthesises that as `[]`. However, if `setenv` or `build-env` has `FOO = ""` then we end up with `[""]` as opam must ensure that this variable is overwritten. The easiest solution for #5925 is simply to allow for that - `[""]` can only arise from `FOO = ""` as an update, so `+=` and `=+` handle that case, and remove the `""` entry from the list. `:=` and `=:` at this point of course work correctly, because the `""` _should_ be there.
- The quoted splitting algorithm in `OpamEnv.split_var` is not quite right - it doesn't cope correctly with a separator quoted inside the double-quote marks. My initial approach therefore was to replace it with the PATH-splitter in `OpamStd.Env.split_path`. This, however, reveals a more fundamental problem - the PATH-splitter correctly splits the PATH, but it _removes_ all the quote marks. The user is at liberty to add quotes in a PATH even if they're not needed. For example, it's a common mistake on Windows to quote directories with spaces in them (having, for example, `Path` set to `C:\Windows\system32;"C:\Program Files\Something\bin"`). With the corrected split function, `PATH += "foo"` causes `Path` to become `foo;C:\Windows\system32;C:\Program Files\Something\bin` with the superfluous quotes removed. This isn't correct, given that the effect of `PATH += "foo"` should morally be much closer to `export PATH="foo;$PATH"`.
- Similarly to this dequoting, with `x-env-path-rewrite`, if two packages disagree over the separator, then we have the slightly odd situation where the variable is split according to the first package's separator and re-joined according to the second package's. i.e. if `Path` is `foo;bar` and the first package to be evaluated has `PATH += "baz"` then `Path` internally is `["baz"; "foo"; "bar"]` and will be exported as `baz;foo;bar`. However, if another package is then added which incorrectly has `x-env-path-rewrite: PATH ":" "target"` and executes `PATH += "broken"` then internally we have `["broken"; "baz"; "foo"; "bar"]` but the variable will be reconstituted as `broken:baz:foo:bar`, so from an initial value of `foo;bar`, `opam env` changed the existing separator.
- Combined with the previous issue with quoting, I've changed the internal representation. Instead of storing just the directory, we instead store the directory, the _raw_ unquoted version of the directory and the separator. This has the benefit that concatentation simply uses the given separators and _raw_ values. The cygpath translations, if used, happen in the _raw_ version, with the native directories manipulated internally. This change of representation naturally fixes several of the issues in #5926 and restores the superfluous quotes in environment variables.
- The final part of the fix for #5838 is to correct the rule used in `Opam.unzip_to` to split the value in `FOO += value`. I've commented (hopefully clearly) the three cases in the commit. At this point, the #4861 test passes again and #5838 is fixed.
- The next commit then handled #3404 in a slightly different way, creating `::` as required (see the changes in the reftests in that commit)
- Finally, there's a long-standing logic error `OpamEnv.reverse_env_update` for `=+` and `=:` where one list was reversed which shouldn't be. Again, the effect is visible in the reftests! The point is that a reversed environment variable was unzipped, and needs to reversed, but there was one reversal too many, given how environment variable updates are stored.

TODO
- [ ] Update master_changes.md in full
- [ ] Update documentation (once semantics presented here fully agreed)


dra27:
OK, the fix I was proposing I didn't like because it didn't preserve unnecessarily escaped parts of existing variables. I then further found it regressed #4861 and consequently broke the msvs-detect package.

I have a plan (which I'm currently implementing) - a failing test for the main issue (#5838) added and for the first bit of the fix, which I am certain is right. The change I think is required for #5838 without breaking #4861 I think will also address #5925 and #5926, so I've added tests showing the failure behaviour for them as well.

dra27:
The change in c4167f6a7ee8248b7aae0fe65f3695703e80c73b requires `environment` files to be reset - @rjbou... does that happen naturally with a format upgrade?

That said, it's making me wonder if we should ever be "putting" `rewrite_default` in the `environment` cache files anyway?

src/state/opamEnv.ml
```diff
@@ -267,11 +328,11 @@ let apply_op_zip ~sepfmt op arg (rl1,l2 as zip) =
     or empty lists is returned if the variable should be unset or has an unknown
     previous value. *)
 let reverse_env_update ~sepfmt var op arg cur_value =
-  let arg = transform_format ~sepfmt arg in
-  if String.equal arg  "" && op <> Eq then None else
+  let (arg, _, _) = transform_format ~sepfmt var arg in
```
kit-ty-kate, line 362: to make the code clearer, I feel like having a record instead of a triple sum type would make things clearer in the future and potentially better at avoiding any future bugs

tests/reftests/env.win32.test
```diff
@@ -85,13 +137,14 @@ setenv: [
   [ RCTQ_ENVSET_ADD_WITH_COL += "a:/nother/gi;ven/path" ]
 ]
 build-env: [
+  [ LC_ALL = "C" ]
```
kit-ty-kate, line 140: I feel like we'd want this all the time, like globally. Couldn't we add this in `run.ml` instead?

dra27: I debated this, too - but the problem then is that it's really not visible from the test that we've done that, and I anticipated that being confusing in future... especially as it would then be really unclear if it ever got changed or broken how and when it might then fail.

src/core/opamStd.ml
```diff
@@ -710,6 +710,26 @@ module OpamString = struct
...
+  let split_quoted path sep =
+    let length = String.length path in
+    let rec f acc index current last normal =
+      if (index : int) = length then
+        let current = current ^ String.sub path last (index - last) in
+        List.rev (if current <> "" then current::acc else acc)
+      else
+      let c = path.[index]
+      and next = succ index in
+      if c = sep && normal || c = '"' then
```
kit-ty-kate, line 722:
```suggestion
      if (c : char) = sep && normal || c = '"' then
```

rjbou (approved):
Thanks!

kit-ty-kate (approved):
Good to go once rebased

kit-ty-kate:
Thanks!

PR #5938, opened 2024-04-25
Fix GitHub Actions for macOS

dra27:
The default macOS runners are now Ventura (14) which has added a realpath utility which causes brew's coreutils package to install realpath as grealpath. Follow brew's configuration advice for the PATH variable.

### DIFF
--- a/.github/scripts/main/preamble.sh
+++ b/.github/scripts/main/preamble.sh
@@ -10,6 +10,9 @@ fi
 echo "Cache -> $CACHE"
 OCAML_LOCAL=$CACHE/ocaml-local
 OPAM_LOCAL=$CACHE/opam-local
+if [ "$RUNNER_OS" = 'macOS' ]; then
+  PATH="/opt/homebrew/opt/coreutils/libexec/gnubin:$PATH"
+fi
 PATH=$OPAM_LOCAL/bin:$OCAML_LOCAL/bin:$PATH; export PATH
 
 OPAM_COLD=${OPAM_COLD:-0}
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
why do we need coreutils in CI on macOS?

kit-ty-kate:
answering to myself: the reftests
```
diff --git a/_build/default/tests/reftests/archive.test b/_build/default/tests/reftests/archive.out
index 69c2eaa..58ff033 100644
--- a/_build/default/tests/reftests/archive.test
+++ b/_build/default/tests/reftests/archive.out
@@ -938,7 +938,9 @@ The following actions will be performed:
 -> installed good-sha256-good-md5.1
 Done.
 ### sh check-cache.sh
-MD5: link, to sha256 archive
+realpath: illegal option -- m
+usage: realpath [-q] [path ...]
```
I feel like we could get rid of that usage instead so we're sure we're testing macOS correctly the way users actually use it

dra27:
Sure - in the meantime, we ought to get CI back up and running, though!

dra27:
Although there are two caveats to possibly not changing it: firstly, prior to Ventura IIUC, macOS did not have a `realpath` utility (it's just added it), but also the usage here is only (I think) in a checking function (i.e. the use of `realpath` is part of the test harness, not part of the opam being tested)

kit-ty-kate:
fair enough. I opened a TODO in https://github.com/ocaml/opam/issues/5940 to think about that later

kit-ty-kate (approved):

PR #5942, opened 2024-05-01
intial draft NixOS depext support

RyanGibb:
This addresses https://github.com/ocaml/opam/issues/5124 and stems from conversations in https://github.com/ocaml/opam/pull/5332.

This is an initial draft to get a conversation started.

Nix's declarative deployment model doesn't match the semantics of the other package managers opam invokes, so it requires some special treatment to be supported.

From conversations on https://github.com/ocaml/opam/pull/5332 the most flexible approach is to create a shell.nix and allow the user to decide how to provide the dependencies, such as invoking `nix-shell`.

This has some limitations, like if the shell is destroyed, and the packages are garbage collected, things will break. For more permanent deployment https://github.com/tweag/opam-nix (or others) can be used to create Nix derivations for opam projects. But this should be sufficient to get depexts on NixOS working for a development environment using an opam switch.

This could possibly also be adapted for Guix.

### DIFF
--- a/src/client/opamSolution.ml
+++ b/src/client/opamSolution.ml
@@ -1179,24 +1179,32 @@ let install_depexts ?(force_depext=false) ?(confirm=true) t packages =
       manual_install t sys_packages
   and menu t sys_packages =
     let answer =
-      let pkgman =
-        OpamConsole.colorise `yellow
-          (OpamSysInteract.package_manager_name ~env config)
-      in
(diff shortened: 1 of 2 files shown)


RyanGibb:
I'm going to close this in favour of creating a Nix shell like environment which Opam can set the environment variables for: https://github.com/RyanGibb/nix.opam

PR #5943, opened 2024-05-02
Allow to parse opam 2.1 switch import files containing extra-files

### LINKED ISSUE #5941: import with opam 2.2 of an export file created with 2.1.2 fails
```
$ opam sw import ./opam-switch-export-20240424-111440.txt --sw=.
Switch initialisation failed: clean up? ('n' will leave the switch partially installed)
[y/n] y
Fatal error:
Invalid_argument("Wrong padding")
```

The file was generated with
```
opam switch export --full opam-switch-export-20240424-111440.txt
```

I'll see if I can share the file without leaking details.

kit-ty-kate:
Fixes #5941 

opam 2.1 used the Base64 module from extlib which does not seem to pad encoded outputs. In contrast, opam 2.2 uses the Base64 module from the base64 library (after extlib's Base64 module was removed in extlib.1.7.8) which by default pads outputs and requires padded inputs when decoding them.

cc-ing base64 implementations maintainers @dinosaure @ygrek 

- Question for @dinosaure: is this fix fine? Should `Base64.encode` be also called with `~pad:false`? What are the pitfalls of disabling padding?

- Question for @ygrek: Just to double-check, am i correct in assuming the output of extlib's base64 implementation is always not padded?

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -28,6 +28,7 @@ users)
 ## Remove
 
 ## Switch
+  * Allow to parse opam 2.1 switch import files containing extra-files [#5943 @kit-ty-kate - fix #5941]
 
 ## Config
 
--- a/src/client/opamSwitchCommand.ml
+++ b/src/client/opamSwitchCommand.ml
@@ -629,7 +629,7 @@ let export rt ?(freeze=false) ?(full=false)
               List.fold_left (fun (hmap,err) (file, base, hash) ->
                   if OpamFilename.exists file &&
                      OpamHash.check_file (OpamFilename.to_string file) hash then
-                    let value = Base64.encode_string (OpamFilename.read file) in
+                    let value = Base64.encode_string ~pad:false (OpamFilename.read file) in
                     OpamHash.Map.add hash value hmap, err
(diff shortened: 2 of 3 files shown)


dinosaure:
As far as `Base64.decode_exn` is concerned, the default setting is that values should have a pad as if there were a '\000' for a C string. Of course, in reality, this is not necessarily the case. The pad becomes useful if you concatenate base64 values and want to decode them cleanly without there being a mismatch between the values (since you have this pad as a delimiter).

If the base64 value exists on its own (or is delimited by something else such as `"`), the pad is not necessary. What `Base64.decode_exn` tries to do is check for the existence of this pad by default, but if you are manipulating base64 values where the pad is not needed, `pad:false` is indeed advisable.

In other words, `pad:true` is for a _paranoid_ mode of retrieving values.

kit-ty-kate:
Thanks! In that case I will use this option for the encoder as well.
I'll also add some tests in a minute

rjbou (approved):
Tiny comment in the test (see commit), otherwise ltgm! Thanks!

PR #5944, opened 2024-05-03
tree: fix `--with-test/doc` when packages are missing

### LINKED ISSUE #5920: opam tree --{test,doc} does not work in combination with --no-switch
Report extracted from https://github.com/ocaml/opam/issues/5755#issuecomment-1921795899

rjbou:
When packages need to be installed (simulate install), either on normal mode or with `--no-switch` option, dependencies where not retrieved with the good selection (with-test,, etc.). The constructed universe was base on `installed` field instead of all packages.

fix #5920

### DIFF
--- a/src/client/opamTreeCommand.ml
+++ b/src/client/opamTreeCommand.ml
@@ -378,10 +378,10 @@ let print_solution st new_st missing solution =
 
 (** Setting states for building *)
 
-let get_universe tog st =
+let get_universe ~all tog st =
   let OpamListCommand.{doc; test; dev_setup; dev; _} = tog in
   OpamSwitchState.universe st ~doc ~test ~dev_setup ~force_dev_deps:dev
-    ~requested:st.installed
+    ~requested:(if all then st.packages else st.installed)
     Query
 
 let simulate_new_state tog st universe install names =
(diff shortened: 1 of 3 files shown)


src/client/opamTreeCommand.ml
```diff
@@ -378,10 +378,10 @@ let print_solution st new_st missing solution =
 
 (** Setting states for building *)
 
-let get_universe tog st =
+let get_universe ~all tog st =
   let OpamListCommand.{doc; test; dev_setup; dev; _} = tog in
   OpamSwitchState.universe st ~doc ~test ~dev_setup ~force_dev_deps:dev
-    ~requested:st.installed
+    ~requested:(if all then st.packages else st.installed)
```
rjbou, line 384: tiny optimisation, i don't know if it worth keeping it.

kit-ty-kate: This conflicts with https://github.com/ocaml/opam/pull/5919

rjbou: PRs need to be queued or merged together (their content, not on master :)). this one resolve only the `--with-test --no-switch` issue while it keeps the faulty behaviour of retrieving with-test/doc dependencies for all packages.

rjbou:
integrated in #5919

PR #5946, opened 2024-05-06
doc: fix pinning instruction in readme

### LINKED ISSUE #5945: The instructions for compiling the repo using `opam pin` appear to be out of date?
The README currently advises

https://github.com/ocaml/opam/blob/b3d2f5c554e6ef3cc736a9f97e252486403e050f/README.md#L18-L19

But to get a working install I actually had to pin all the dependencies of the various opam libraries in the right order. Based on my notes, I think this command should currently give a working install:

```sh
opam pin opam-core --dev-repo -y &&\
    opam pin opam-format --dev-repo -y &&\
    opam pin opam-repository --dev-repo -y &&\
    opam pin opam-state --dev-repo -y &&\
    opam pin opam-solver --dev-repo -y &&\
    opam pin opam-client --dev-repo -y &&\
    opam pin add opam-devel --dev -y
```

I kind of doubt we want to replace the current readme line with this recipe, but maybe we could do something like...

1. Put it in a script or a phony make target
2. Change https://github.com/ocaml/opam/tree/b3d2f5c554e6ef3cc736a9f97e252486403e050f/opam-devel.opam so that it includes all the libraries as pin-depends and advise people to pin this package from the source rather than from the opam repo. (I think this would work around the non-transitive nature of pins?)

Hopefully there is a nicer option (I'm happy to help with that), but in the meantime this issue may offer a recipe to help others along who hit the same snag.

rjbou:
fix #5945

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,8 +15,9 @@ and [Usage](http://opam.ocaml.org/doc/Usage.html) guides.
 
 ## Compiling This Repo
 
-Either from an existing opam installation, use `opam pin add opam-devel
---dev`, or:
+Either from an existing opam installation, use
+`opam pin add .` if you cloned locally the repository,
+`opam pin add git+https://github.com/ocaml/opam` otherwise, or:
 
 * Make sure you have the required dependencies installed:
   - GNU make
--- a/master_changes.md
+++ b/master_changes.md
@@ -147,6 +147,7 @@ users)
 
 ## Doc
   * Remove the ppa from the installation instructions on Ubuntu [#5988 @kit-ty-kate - fix #5987]
+  * Fix pinning instructions in readme [#5946 @rjbou - fix #5945]
 
 ## Security fixes
 


README.md
```diff
@@ -15,8 +15,8 @@ and [Usage](http://opam.ocaml.org/doc/Usage.html) guides.
 
 ## Compiling This Repo
 
-Either from an existing opam installation, use `opam pin add opam-devel
---dev`, or:
+Either from an existing opam installation, use
+`opam pin add git+https://github.com/ocaml/opam`, or:
```
rjbou, line 19:
```suggestion
Either from an existing opam installation, use
`opam pin add .` if you cloned locally the repository, 
`opam pin add git+https://github.com/ocaml/opam`otherwise, or:
```

kit-ty-kate: modulo the missing required space before "otherwise", lgtm

rjbou: updated

kit-ty-kate (approved):

kit-ty-kate:
Thanks!

PR #5950, opened 2024-05-10
Fix the value of the 'arch' variable when the current OS is 32bit on a 64bit machine

### LINKED ISSUE #5949: `opam var arch` reports kernel rather than user arch on Raspberry Pi OS
32-bit Raspberry Pi OS runs with a 64-bit kernel and 32-bit userspace. This means that `opam var arch` returns `arm64`, but since the C compiler is 32-bit, it would be much more helpful it returned `arm32` (as it does if run through `linux32`).

Certainly for Raspberry Pi OS, but possibly in general on Unix, we could perhaps use `uname -m` to determine the machine architecture, but then refine it with `getconf LONG_BIT` and if that returns `32` then map a few of the architectures down to smaller userspaces?

kit-ty-kate:
Fixes #5949 

`getconf LONG_BIT` is POSIX-complient and thus should work on any Unix-like systems (including cygwin)

Tested successfully on Linux, macOS, Cygwin and MSYS2

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -61,6 +61,7 @@ users)
 ## Show
 
 ## Var/Option
+  * Fix the value of the 'arch' variable when the current OS is 32bit on a 64bit machine [#5950 @kit-ty-kate - fix #5949]
 
 ## Update / Upgrade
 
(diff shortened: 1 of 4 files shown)


dra27 (approved):
LGTM, thanks!

PR #5953, opened 2024-05-14
Mitigate copying/deletion of unrecognised symlinks

dra27:
Partially addresses the issue reported in https://github.com/ocaml/opam/issues/5782#issuecomment-2108726192

In these situations, Cygwin is most likely to create a JUNCTION, which can't be read by `Unix.lstat`. The fix for now I propose is simply to warn. This has the very big benefit of making ocamlbuild install without needing Developer Mode.

It has the downside that we fail to copy the file... I'm suggesting we might choose to live with that for now.

Note that `OpamSystem.remove_file_t` contained an unnecessary second call to `lstat`. I needed to re-use the actual removal logic for the exception case in `remove_dir_t`, but `lstat` is very expensive in C on Unix, even more expensive in OCaml and indescribably more expensive on Windows, so eliminating the double-call seems better!

### DIFF
--- a/src/core/opamSystem.ml
+++ b/src/core/opamSystem.ml
@@ -129,19 +129,15 @@ let win32_unlink fn =
     with _ -> raise e)
 
 let remove_file_t ?(with_log=true) file =
-  if
-    try ignore (Unix.lstat file); true with Unix.Unix_error _ -> false
-  then (
-    try
-      if with_log || log_for_file_management () then
-        log "rm %s" file;
-      if Sys.win32 then
-        win32_unlink file
-      else
-        Unix.unlink file
-    with Unix.Unix_error _ as e ->
-      internal_error "Cannot remove %s (%s)." file (Printexc.to_string e)
-  )
+  try
+    if with_log || log_for_file_management () then
+      log "rm %s" file;
+    if Sys.win32 then
+      win32_unlink file
+    else
+      Unix.unlink file
+  with Unix.Unix_error _ as e ->
+    internal_error "Cannot remove %s (%s)." file (Printexc.to_string e)
 
 let rec remove_dir_t dir =
   let files = get_files dir in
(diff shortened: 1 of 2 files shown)


dra27:
With this, the issue at least reduces to:

```
C:\Scratch>opam source ocp-indent
[WARNING] Warning: cannot copy C:\Users\DRA\AppData\Local\Temp\opam-54072-ca16dd\ocp-indent-1.8.1\tests\inplace\link.ml to
          C:\Scratch\ocp-indent.1.8.1\tests\inplace
Successfully extracted to C:\Scratch\ocp-indent.1.8.1
```

src/core/opamSystem.ml
```diff
@@ -129,19 +129,15 @@ let win32_unlink fn =
     with _ -> raise e)
 
 let remove_file_t ?(with_log=true) file =
-  if
-    try ignore (Unix.lstat file); true with Unix.Unix_error _ -> false
```
kit-ty-kate, line 133: `remove_file_t` on a non-existing file used to simply ignore it. This PR changes this behaviour to error instead.
Was the old behaviour used anywhere that we know?

dra27: I don't think so - it's only other use is in the cross-device fallback for `mv`. The only way it could be called with a file which didn't exist is if the file has been removed between the test in the calling function and the test here, and there was already a race for that anyway (between the test I've deleted here and the `internal_error` handler below).

kit-ty-kate: it is also being used in `remove_file` and `remove_dir` which are used everywhere in opam's code (e.g. via `OpamFilename.remove`)

dra27: `remove_file` doesn't use `remove_file_t` (LL202-215) and `remove_dir` already checks whether its argument exists?

kit-ty-kate: AH! I see. There are two toplevel versions of `remove_file` in the same file. Only the latter is exported. Sorry for missing that. This is all good then.

kit-ty-kate (approved):

kit-ty-kate:
maybe this also warrants an addition to the API section of the master changelog actually

src/core/opamSystem.ml
dra27: Actually, until you’d said, I’d missed that too! I was only sure that the code was ok because I’d deleted the first remove_file as a result of an unused value warning and already searched for other uses of remove_file_t 🙂

dra27:
The API hasn’t changed though?

dra27:
Ah, yes, indeed. I added:

```
  * `OpamSystem.copy_dir` and `OpamSystem.mv` may display a warning on Windows if an invalid symlink (e.g. an LXSS Junction) is found [#5953 @dra27]
```

kit-ty-kate:
Thanks!

PR #5957, opened 2024-05-16
Release script fixes and improvements

kit-ty-kate:
Matching what is in the Dockerfile and what is in the Makefile to make it compile on macOS for the 2.1 branch


dra27 (approved):
LGTM!

PR #5962, opened 2024-05-17
Move last-env `OPAM_LAST_ENV` files outside the switch

### LINKED ISSUE #5823: last-env files should be stored outside the switch
The original implementation of #5417 addressed two problems:
- #3181, which was the original inspiration; and
- #3411 (and #5423)

The first problem is solved and working correcting in that the `last-env` file contains the environment updates which were _previously_ applied to the environment. Previously, opam always used _the same_ `environment` file both to revert and apply updates, which was what resulted in the effective prohibition of packages using `setenv` (because after uninstalling a package which had `setenv` instructions, `environment` no longer contains the changes from that package, so they can't be reverted).

The second part was solved in the original version of the PR by proposing that the last-env files be stored in `/tmp`, but we changed this prior to merge. Unfortunately, the change at present breaks/removes the second part of the fix.

The last-env files must at least be stored outside the switch, or the removal of a global switch or removal/moving of a local switch causes the same problem (the fact that moving a local switch probably breaks the switch is not important here - it's still important to be able to revert the old environment). This becomes a serious concern for the MSVC compiler packages, which contain a large number of environment updates.

There are two choices:
1. As in the original, put the last-env files in `/tmp` (or equivalent)
  - This means that one can delete the opam root and still successfully revert previous environment updates
  - At first glance, that seems utterly niche (not sure anyone beyond opam devs operates with multiple opam roots), but it also hardens the case of `rm -rf ~/.opam && opam init` which is a more common user "fix" than we wish it was
2. Put the last-env files in `~/.opam/.last-env`
  - Completely fixes #3411, and also "fixes" `opam switch remove foo` run _before_ running `eval $(opam env --revert)`

I'm still kinda keen on 1, perhaps with a fallback to 2 if, for some reason, TMPDIR is not writeable, but the Windows implications mean I think we must do at least 2.

moyodiallo:
Fixes #5823 by storing the last-env `OPAM_LAST_ENV` files in the directory `~/.opam/.last-env` instead of in the switch as the solution described in the issue.

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -4205,7 +4205,6 @@ let clean cli =
            cleandir (OpamPath.Switch.build_dir root sw);
            cleandir (OpamPath.Switch.remove_dir root sw);
            cleandir (OpamPath.Switch.extra_files_dir root sw);
-           cleandir (OpamPath.Switch.last_env root sw);
            let pinning_overlay_dirs =
              List.map
                (fun nv -> OpamPath.Switch.Overlay.package root sw nv.name)
--- a/master_changes.md
+++ b/master_changes.md
@@ -36,6 +36,7 @@ users)
   * Allow to parse opam 2.1 switch import files containing extra-files [#5943 @kit-ty-kate - fix #5941]
 
 ## Config
+  * Move last-env `OPAM_LAST_ENV` files outside the switch to be in the `opam root` [#5962 @moyodiallo - fix #5823]
 
 ## Pin
 
@@ -152,6 +153,7 @@ users)
 ## opam-solver
 
 ## opam-format
+  * `OpamPath`: remove `OpamPath.Switch.last_env` function in favor to `OpamPath.last_env` as the files are no more stored in switch directory [#5962 @moyodiallo - fix #5823]
 
 ## opam-core
   * `OpamStd.String`: add `split_quoted` that preserves quoted separator [#5935 @dra27]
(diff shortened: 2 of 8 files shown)


moyodiallo:
@dra27  Do we need more test here ?

src/client/opamCommands.ml
```diff
@@ -4223,6 +4222,7 @@ let clean cli =
            List.iter (fun d ->
                if not (List.mem d keep_sources_dir) then rmdir d))
          switches);
+    if all_switches then cleandir (OpamPath.last_env root);
```
kit-ty-kate, line 4225: I'm unsure if we want to clean last-env in that case. What do you think @dra27 @rjbou?

kit-ty-kate (approved):
Aside from the one question, I think this looks pretty good.

src/client/opamCommands.ml
moyodiallo: yeah I was hesitating about this because it is not consequent to consider cleaning it.

dra27: Having discussed it this morning, let's leave the `.last-env` files as _not_ being cleaned:
```suggestion
```
and we'll open an issue to track sorting this out in opam 2.2.1

rjbou: issue created

rjbou (approved):
Thanks!
I've added an entry in API changes, otherwise, LGTM!

PR #5964, opened 2024-05-18
Tweak incorrect case in `--cygwin-extra-packages`

dra27:
An incorrectly translated case in #5930 means `opam init` with no arguments at the moment unconditionally installs an internal Cygwin! Easy to review based on the diff of the original commit in https://github.com/ocaml/opam/pull/5930/commits/8bc1214c171e588cd3e574c86c798e6a193ca073 - the ```| `none, None -> None``` deleted case at L419 has no equivalent in the new code (indeed, no case yields `None`).

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -21,7 +21,7 @@ users)
 ## Plugins
 
 ## Init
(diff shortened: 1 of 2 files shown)


kit-ty-kate (approved):

kit-ty-kate:
Thanks!

Issue #5741, opened 2023-11-24
opam.ocaml.org repository takes 10+ minutes to download on Windows


jonahbeckford:
This is a long-standing issue. But I realize I haven't communicated it so perhaps it could be tracked ...

The following takes a little over 10 minutes on GitLab SaaS Windows hardware (almost same timing on my desktop):

```sh
opam repository set-url default https://opam.ocaml.org --yes --all
```

Ditto for `opam init`.

kit-ty-kate:
This is indeed a long known standing issue. The problem appears when the cache needs to be rebuilt (when i remove it manually locally, `opam repo list` takes 4m21s on my local Windows machine), which involves reading the content of all the files in the opam repositories you have set, which, for some reason, is very fast on Unixes but takes a really long time on Windows.

What is the way to read a lot (~ 30_000) of files in quick successions on Windows?

Alizter:
For the record, in Dune package management we are using git as a revision store for opam repositories and it seems to be working well on windows. That is, it is reasonably performant.

kit-ty-kate:
@Alizter does it load/read all the packages in opam-repository once fetched? opam needs to preload every packages, i don’t think this is the case for dune so I’m not surprised the issue is not encountered there.

Alizter:
@kit-ty-kate We don't load all the files. We have a revision store mechanism that 0install uses to fetch opam files during solving. This lets us use a single repository to store all opam repositories at all revisions as needed. You're correct that we have no need to fetch all the files in an opam repository at once, only the ones the solver requests.

kit-ty-kate:
Update on this issue: I had initially hoped to have a fix before the release of opam 2.2.0 but the required change (https://github.com/ocaml/opam/issues/5648) is big enough that i am currently restarting in my 3rd fresh branch for the fix since January. https://github.com/ocaml/opam/issues/5824 would also help this issue but seems equally – if even more – hard to fix.

I'm still working on it in the background but it seems increasingly unlikely to make the cut, although it will most certainly be in the next release after.

kit-ty-kate:
I tried one last experiment in https://github.com/ocaml/opam/pull/5966 which worked at reducing the time `opam init/update` takes in half on Windows (which helps a lot but still isn't as fast as Unix), but given the amount of untested change that would have to go through I don't think it's very wise to merge at the moment.

NicoNekoru:
[...]

Issue #5648, opened 2023-08-30
Allow to use a "compressed" (one file) repository format for performance and sustainability purpose


kit-ty-kate:
opam repositories currently have a "one file per packages/versions" but as the number of packages grow it creates a sustainability problem for people with low number of inodes for their filesystems (e.g. see https://github.com/ocaml/opam/issues/5484) and a performance problem (you have to open each file on every `opam update`)

I'm not set on a particular format for that file but it could be the format that `opam switch export` already uses.
@mseri also suggested using SQLite

rjbou:
Some data: today repo concatened gives a file of 39M and 1 235 506 lines.

avsm:
One mechanism would be to use normal tar.xz files and use the OCaml libraries to parse them directly with unpacking. That has the benefits of making them easy to create, and there performance improvements from not having a lot of small files.

rjbou:
*From dev meeting*

We can have several formats in the opam repository itself:
* the *historical* one: directories and opam files per package,
* a aggregated file: concat of all opam files, human readable,
* a compressed file: xz-compressed version of the aggregated file.

The repo file can mention what it the format of the repo, but they can coexists in a simple repo. Opam can understand all those formats (backward compatibility), for API users, it is imperceptible as fetching repo functions remain the same. Opam can also try to retrieve compressed format, then fallback on aggregated format, then fallback to plain directory one.

These new formats can be served via a webserver, for example having opam2web generate them. For github main opam repository, one solution is to have an alternate branch, that serves the aggregated file. It would be automatically updated for each merge.

c-cube:
I have another suggestion: a `zip` file. It doesn't compress as well as .zst or .xz but it has a big advantage that it's randomly addressable, so you never need to actual unzip it. If you want `foo/foo.1.2/opam` you can directly get the corresponding entry without decompression.

kit-ty-kate:
xz is also randomely addressable. I use this feature in https://github.com/kit-ty-kate/opam-health-check-ng using the `pixz` external tool.

dinosaure:
[...]

Issue #5591, opened 2023-06-30
OpamFile.OPAM.read_from_string is not domain-safe


talex5:
It seems that trying to parse two opam files at the same time results in a crash:
```ocaml
let data = {|
opam-version: "2.0"
version: "2.1.5"
synopsis: "Format library for opam 2.1"
|}

let test () =
  try
    for _ = 1 to 1000 do
      let file = OpamFile.OPAM.read_from_string data in
      ignore file
    done
  with ex ->
    Printexc.print_backtrace stderr;
    Printf.eprintf "test.ml: %s\n%!" (Printexc.to_string ex)

let () =
  Printexc.record_backtrace true;               (* Does nothing?? *)
  let a = Domain.spawn test in
  let b = Domain.spawn test in
  Domain.join a;
  Domain.join b
```

```
$ dune exec ./test.exe
test.ml: At ./<none>:2:14-2:19::  
Parse error
fish: Job 1, 'dune exec ./test.exe' terminated by signal SIGSEGV (Address boundary error)
```

Different runs produce different errors, e.g.
```
test.ml: At ./<none>:3:7-3:8::         
Parse error
test.ml: Invalid_argument("index out of bounds")
```

I guess this is because it uses `Stdlib.Parsing`, which doesn't look thread-safe. Would be good to have some documentation on how to use the library safely with OCaml 5.

Tested with opam-format 2.1.5.

talex5:
Also, `opamStd.ml`'s `terminal_columns` installs a signal handler that uses lazy (and isn't atomic): https://github.com/ocaml/opam/blob/d1fe3e82b4889c7f24b83e3284720b3611e93574/src/core/opamStd.ml#L958-L961

This crashes under multicore if two threads try to log at the same time:
```
CamlinternalLazy.Undefined
Raised at CamlinternalLazy.force_gen_lazy_block in file "camlinternalLazy.ml", line 75, characters 9-24
Called from OpamStd.OpamFormat.reformat in file "src/core/opamStd.ml" (inlined), line 1280, characters 44-71
Called from OpamConsole.warning.(fun) in file "src/core/opamConsole.ml", line 584, characters 6-63
```

avsm:
The segv is possibly due to the Stdlib.Parsing module being thread-unsafe and involving C bindings (https://github.com/ocaml/ocaml/blob/trunk/stdlib/parsing.ml). A standalone testcase would be suitable to create an issue on the ocaml/ocaml repo.

[...]

PR #5966, opened 2024-05-21
Improve the performance of opam update/init on Windows

kit-ty-kate:
WIP attempt at (temporary) fixing #5741 by throwing more silicon^Wdomains at the problem while #5648 is being worked on.

This currently requires:
- https://github.com/ocaml/ocaml-re/issues/287 to be fixed or alternatively we could just use mutexes for now
- https://github.com/ocaml/opam-file-format/pull/60 to be merged and released
- to remove or add a mutex around `Stats.read_files`
- more testing as i'm sure there are more data races

Current result:
- On Linux: `opam update default` goes from 6 seconds to 4 seconds
- On Windows: `opam init --bare` goes from 6 and a half minutes to 3 minutes exactly. Of that 3 minutes, 1 minute is unchanged and taken by the extraction of the tar.gz archive.

Partly inspired by [NTFS really isn't that bad - Robert Collins (LCA 2020)](https://www.youtube.com/watch?v=qbKGw8MQ0i8)

Related to #5591

### DIFF
--- a/src/core/opamUrl.ml
+++ b/src/core/opamUrl.ml
@@ -43,7 +43,7 @@ let equal u v = compare u v = 0
 exception Parse_error of string
 let parse_error s = raise (Parse_error s)
 
-let split_url =
+let split_url u =
   let re =
     Re.(compile @@ whole_string @@ seq [
         (* Parse the scheme, which is either backend+protocol or just a protocol *)
@@ -66,7 +66,6 @@ let split_url =
         opt @@ seq [ char '#'; group @@ rep any ];
       ])
   in
-  fun u ->
     match Re.Group.all (Re.exec re u) with
     | [| _; vc; transport; path; suffix; hash |] ->
       let opt = function "" -> None | s -> Some s in
@@ -103,7 +102,7 @@ let backend_of_string = function
                 (OpamConsole.colorise `underline p))
 
 
-let looks_like_ssh_path =
+let looks_like_ssh_path path =
   (* ':' before any '/' : assume ssh, like git does. Exception for 'x:' with
      single char, because Windows *)
   let re =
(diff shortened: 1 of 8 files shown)


kit-ty-kate:
The PR is now fully working as far as i tested.
However given the untested nature of this change, its presence in such a central piece of code and its reliance on big changes in underlying software (opam-file-format) I don't think it should be in 2.2 but i think it a good place to start to think about:
- what is actually needed for opam to move to OCaml 5 (especially on windows as it is currently untested)
- making part of the engine parallel
- changing in API to make it data-race free (the change in `OpamConsole.log`, uses of `ocaml-re`, removal of `chdir`, introduction of `OpamLazy` from https://github.com/ocaml/opam/pull/5877)

I'll split this PR into several smaller ones once 2.2.0 is out.

src/core/opamUrl.ml
```diff
@@ -43,7 +43,7 @@ let equal u v = compare u v = 0
 exception Parse_error of string
 let parse_error s = raise (Parse_error s)
 
-let split_url =
+let split_url u =
   let re =
```
kit-ty-kate, line 47: actually it's much easier to do this using domain local storage
```suggestion
let split_url =
  let re =
    Domain.DLS.new_key @@ fun () ->
```

src/core/opamConsole.ml
```diff
@@ -560,7 +560,10 @@ let clear_pending debug_level =
   Queue.iter f pending;
   Queue.clear pending
 
-let log section ?(level=1) fmt =
+let log_mutex = Mutex.create ()
+let log section ?(level=1) k =
+  Mutex.protect log_mutex @@ fun () ->
+  k @@ fun fmt ->
```
kit-ty-kate, line 566: this might not be necessary if we replace the use of the default formatters by their OCaml5 `Format.get_*` counterparts.

Issue #5923, opened 2024-04-17
PKG_CONFIG_PATH is being double-quoted in the environment


jonahbeckford:
This is opam `2.2~beta2`. The problem did not manifest itself in opam `2.2~alpha0`.

The dump from `opam option` is:

```
setenv                 [[PKG_CONFIG_SYSTEM_LIBRARY_PATH = ""] [PKG_CONFIG_SYSTEM_INCLUDE_PATH = ""] [PKG_CONFIG_PATH =
                       "Y:\\source\\dkml\\build\\pkg\\bump\\msys64\\clang64\\lib\\pkgconfig"]]
```

Yet the environment variable available inside `opam install` (and in the opam `.env` log file) is:

```
PKG_CONFIG_PATH="Y:\source\dkml\build\pkg\bump\msys64\clang64\lib\pkgconfig"
```

jonahbeckford:
## Test Cases without double quotes

```sh
# Cygwin and pkg-config

$ env 'PKG_CONFIG_PATH=Y:\source\dkml\build\pkg\bump\msys64\clang64\lib\pkgconfig' pkg-config --cflags libffi
-IY:/source/dkml/build/pkg/bump/msys64/clang64/include

$ env 'PKG_CONFIG_PATH="Y:\source\dkml\build\pkg\bump\msys64\clang64\lib\pkgconfig"' pkg-config --cflags libffi
Package libffi was not found in the pkg-config search path.
Perhaps you should add the directory containing `libffi.pc'
to the PKG_CONFIG_PATH environment variable
Package 'libffi', required by 'virtual:world', not found
```

Conclusion: double-quotes are not needed

jonahbeckford:
### Test Cases with double quotes

```sh
# MSYS2 and pkgconf

$ env 'PKG_CONFIG_PATH="Y:\source\dkml\build\pkg\bump\msys64\clang64\lib\pkgconfig"' pkgconf --cflags libffi
-IZ:/msys64/clang64/bin/../include
```

Conclusion: adding double-quotes breaks both pkg-config and pkgconf

jonahbeckford:
### Correction for Cygwin examples

```
# Cygwin using Cygwin native /usr/bin/pkg-config

$ cygpath -aw .
Z:\cygwin64\home\beckf

$ /usr/bin/pkg-config --version
2.2.0

[...]

PR #5972, opened 2024-05-24
Change default rule for `PKG_CONFIG_PATH` and `MANPATH` to `target`

dra27:
Neither of these systems supports quoting (on Windows or Unix).

This should address the regression in #5923, but it's not the final fix for Windows, so that issue should stay open after this is merged.

cc @RyanGibb

### DIFF
--- a/tests/reftests/env.win32.test
+++ b/tests/reftests/env.win32.test
@@ -23,12 +23,12 @@ Done.
 ### : already contains a directory which needs escaping (e.g. building with
 ### : MSVC)
 ### opam env --shell=cmd | grep "PATH" | 'XXX:.*' -> '' | 'set "P' -> 'set P'
-set "MANPATH=:"${BASEDIR}/OPAM/rewriting/man""
+set MANPATH=:${BASEDIR}/OPAM/rewriting/man
 set PATH=${BASEDIR}/OPAM/rewriting/bin;C:\Devel\bin1;C:\Devel\bin2;"C:\Devel\bin3;";C:\Devel\bin4;ZZZ:\;
 ### opam exec -- opam env --shell=cmd --revert | grep 'ZZZ:' | 'ZZZ:\\.*' -> 'ZZZ:\' | 'set "P' -> 'set P' 
 ### : Test for #5838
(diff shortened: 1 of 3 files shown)


kit-ty-kate (approved):

rjbou (approved):
Thanks!

tests/reftests/env.win32.test
```diff
@@ -23,14 +23,13 @@ Done.
...
-set "MANPATH=:"${BASEDIR}/OPAM/rewriting/man""
+set MANPATH=:${BASEDIR}/OPAM/rewriting/man
 set PATH=${BASEDIR}/OPAM/rewriting/bin;C:\Devel\bin1;C:\Devel\bin2;"C:\Devel\bin3;";C:\Devel\bin4;ZZZ:\;
 ### opam exec -- opam env --shell=cmd --revert | grep 'ZZZ:' | 'ZZZ:\\.*' -> 'ZZZ:\' | 'set "P' -> 'set P' 
 ### : Test for #5838
 ### opam env | grep MANPATH
-MANPATH=':"${BASEDIR}/OPAM/rewriting/man"'; export MANPATH;
+MANPATH=':${BASEDIR}/OPAM/rewriting/man'; export MANPATH;
 ### opam exec -- opam env --revert | grep MANPATH
-MANPATH=''; export MANPATH;
```
kit-ty-kate, line 33: is this expected?

dra27: No, that was clumsy of me. I think the `Target` above should be `Host`, but I need to test that (especially w.r.t. MSYS2)

dra27:
The switch to `Host` restores the MANPATH revert and gets us to:

```
set MANPATH=:/cygdrive/c/Users/DRA/AppData/Local/opam/testing-mingw/man
```
on Cygwin, and
```
set MANPATH=:/c/Devel/Roots/msys2-testing/testing/man
```
on MSYS2, which should be correct. i.e. the directories are colon-separated, but are converted using cygpath.

kit-ty-kate:
Thanks!

Issue #5971, opened 2024-05-23
`OpamDownload` assertion failure is causing  opam-repo-ci builds to fail on arm32-ocaml-4.14


shonfeder:
First noticed (afaik) at https://github.com/ocaml/opam-repository/pull/25905#issuecomment-2119010020

The error we're seeing in CI is

```
/home/opam: (run (network host)
                 (shell "opam init --reinit --config .opamrc-sandbox -ni"))
Fatal error:
File "src/repository/opamDownload.ml", line 140, characters 2-8: Assertion failed
"/usr/bin/linux32" "/bin/sh" "-c" "opam init --reinit --config .opamrc-sandbox -ni" failed with exit status 99
```

which can be seen in, e.g., [this CI log](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/f4bd28ba789a542ddaadb40512a72c889061cc5e/variant/extras,arm32-ocaml-4.14,multicore-bench.0.1.3)

The failing assertion is at 

https://github.com/ocaml/opam/blob/391333d35bcdc8b55df709b876b8bafcf75f3452/src/repository/opamDownload.ml#L140

kit-ty-kate:
is it reproducible or does it only happen from time to time?

dbuenzli:
FWIW it also happened on the cmdliner release [here](https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/8b1bcef2701b746b5322949a3b4eed51eca511ef/variant/extras,arm32-ocaml-4.14,cmdliner.1.3.0).

shonfeder:
It's reproducible. E.g., every Jane Street package looks to be suffering the same fate currently: https://opam.ci.ocaml.org/github/ocaml/opam-repository/commit/b0fb4f8c144e4e78cd6de1972fc3453a2024d8a8

rjbou:
It seems to happen only on arm & freebsd images.
If it is at repository reloading stage, it shouldn't go through that code as in the image it is defined as a directory (file:///home/opam/opam-repository).
Is it possible to extract a backtrace and some logs (-vv | --debug)?

shonfeder:
I'll see about getting this reproducing net week. I also realized I didn't take into account the container caching when I claimed it is reproducible, and all of the CI jobs I've looked at so far are pulling that step from the cache.

kit-ty-kate:
[...]

PR #5975, opened 2024-05-27
Hardens the interface of OpamUrl.local_{file,dir}

kit-ty-kate:
Would help to debug #5971 

This change is done to be able to show an error message instead of a raw assert failure when an expected local file does not exist.

This is a breaking API change and will require a change in https://github.com/ocaml-opam/opam-publish/pull/138. While not strictly necessary, i believe this change is an overall improvement and will help track similar issues in the future.

One related issue i'm seeing with how these functions are used at the moment, is that if you have a url like `file+git://<your-git-repo>/some/file/possibly/modified/or/untracked/since/HEAD` this function would happily return the file in question, which it seems like it should not do. I'm not sure if this behaviour is exploited in the wild or in our code but i think this issue is separate and can be tracked in a later issue.


kit-ty-kate:
Superseded by #5979 and #5978

Issue #5956, opened 2024-05-16
TODO: have a look at the patches from the FreeBSD ports repository


kit-ty-kate:
https://cgit.freebsd.org/ports/tree/devel/ocaml-opam/files

* The two ocaml compiler patches can be fixed by simply upgrading to 4.14.2
* The addition of `CC="$(CC)"` to `make compiler` sounds ok to me and could be imported

hannesm:
ref https://github.com/ocaml/opam/issues/5584

kit-ty-kate:
I didn't manage to reproduce the issue mentioned in https://cgit.freebsd.org/ports/commit/devel/ocaml-opam/files/patch-Makefile?id=29410e20d1de876478fde6e4bc80b8b0823b1184

@hannesm do you have any hints on how to reproduce the issue?

hannesm:
@kit-ty-kate I do not have a reproduction. It is related to `tinderbox` (https://wiki.freebsd.org/Tinderbox) together with ccache. I never used tinderbox.

PR #5976, opened 2024-05-27
Upgrade the vendored OCaml compiler to 4.14.2

kit-ty-kate:
Partially f.i.x.e.s https://github.com/ocaml/opam/issues/5956


dbuenzli:
Does that change lower bound on OCaml to compile `opam` ? (Mainly interested to inform a corresponding reasonable `cmdliner` lower bound).

rjbou:
No, it's only the vendored one, used for `make cold`. Opam keeps ocaml lower bound `>= 4.08.0`.

PR #5977, opened 2024-05-27
Display a note when reloading a repository

kit-ty-kate:
First commit is extracted from https://github.com/ocaml/opam/pull/5883

This should at least mitigate the concerns from users thinking opam might have crashed while waiting for opam update/init to finish.

The second commit is temporary while https://github.com/ocaml/opam/issues/5741 is being fixed.


dra27 (approved):
This looks beautiful, thanks - things taking a long time now at least explain what they're doing while taking a long time!

rjbou (approved):
Thanks!

Issue #5978, opened 2024-05-28
TODO: remove OpamUrl.local_{file,dir}


kit-ty-kate:
Description will be updated once the related PR is opened

PR #5979, opened 2024-05-28
Display an appropriate error message when the file given to opam init --config does not exist

### LINKED ISSUE #5971: `OpamDownload` assertion failure is causing  opam-repo-ci builds to fail on arm32-ocaml-4.14

kit-ty-kate:
Fixes #5971 

This PR introduces `OpamUrl.kind` which i think is an improvement over `OpamUrl.local_{file,dir}` as way of detecting if a url is local or not. Most functions that use `OpamUrl.local_{file,dir}` already check later if the file exist and fail gracefully so there is no need to check twice.

I opened https://github.com/ocaml/opam/issues/5978 to think about removing `local_{file,dir}` in the future, but for now this would be a fairly intrusive change.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -127,6 +127,7 @@ users)
 ### Tests
   *  Add test cases to `update.test` for version-equivalent renames [#6774 @arozovyk fix #6754]
   * Fix a failure when two hashes start with the same two characters [#6793 @kit-ty-kate]
+  * Add a test showing the behaviour of `opam init --config` when the file given does not exist [#5979 @kit-ty-kate @rjbou]
 
 ### Engine
 
--- a/master_changes.md
+++ b/master_changes.md
@@ -158,3 +158,4 @@ users)
 
 ## opam-core
   * `OpamCmdliner` was added. It is accessible through a new `opam-core.cmdliner` sub-library [#6755 @kit-ty-kate]
+  * `OpamUrl`: rename and expose `local_path` as `looks_like_local_path` [#5979 @kit-ty-kate]
--- a/master_changes.md
+++ b/master_changes.md
@@ -19,6 +19,7 @@ users)
 ## Plugins
 
 ## Init
+  * ✘ Display an appropriate error message when the file given to `opam init --config` does not exist or is in a VCS. This changes the behaviour for VCS local urls that was previously retrieved. [#5979 @kit-ty-kate - fix #5971]
 
 ## Config report
 
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -131,9 +131,27 @@ let get_init_config ~no_sandboxing ~no_default_config_file ~add_config_file =
     (if no_default_config_file then []
      else List.filter OpamFile.exists (OpamPath.init_config_files ()))
     @ List.map (fun url ->
-        match OpamUrl.local_file url with
-        | Some f -> OpamFile.make f
-        | None ->
+        match url.OpamUrl.backend with
+        | #OpamUrl.version_control ->
+          OpamConsole.error_and_exit `Bad_arguments
+            "Version control url not supported for %S"
+            (OpamUrl.to_string url);
+        | `rsync ->
+          begin match OpamUrl.looks_like_local_path url with
+          | Some path ->
+            let file = OpamFilename.of_string path in
+            if OpamFilename.exists file then
+              OpamFile.make file
+            else
+              OpamConsole.error_and_exit `Bad_arguments
+                "File does not exist: %s"
+                (OpamFilename.to_string file)
+          | None ->
+            OpamConsole.error_and_exit `Bad_arguments
+              "SSH url not supported for %S"
+              (OpamUrl.to_string url)
+          end
+        | `http ->
           let f = OpamFilename.of_string (OpamSystem.temp_file "conf") in
           OpamProcess.Job.run (OpamDownload.download_as ~overwrite:false url f);
           let hash = OpamHash.compute ~kind:`SHA256 (OpamFilename.to_string f) in
--- a/src/core/opamUrl.ml
+++ b/src/core/opamUrl.ml
@@ -209,7 +209,7 @@ let base_url url =
   | "" -> url.path
   | tr -> Printf.sprintf "%s://%s" tr url.path
 
-let local_path = function
+let looks_like_local_path = function
   | { transport = ("file"|"path"|"local"|"rsync"); path;
       hash = _; backend = (#version_control | `rsync); }
     when looks_like_ssh_path path = None ->
@@ -218,14 +218,14 @@ let local_path = function
 
 let local_dir url =
   let open OpamStd.Option.Op in
-  local_path url >>|
+  looks_like_local_path url >>|
   OpamFilename.Dir.of_string >>= fun d ->
   if OpamFilename.exists_dir d then Some d
   else None
 
 let local_file url =
   let open OpamStd.Option.Op in
-  local_path url >>|
+  looks_like_local_path url >>|
   OpamFilename.of_string >>= fun f ->
   if OpamFilename.exists f then Some f
   else None
--- a/tests/reftests/init.test
+++ b/tests/reftests/init.test
@@ -328,15 +328,13 @@ Parse error
 ### :II:e:1: Non existing local file
 ### rm -rf $OPAMROOT
 ### opam init --bypass-checks --bare --no-setup --config some-non-existant-file
-Fatal error:
-File "src/repository/opamDownload.ml", line 198, characters 2-8: Assertion failed
-# Return code 99 #
+[ERROR] File does not exist: ${BASEDIR}/some-non-existant-file
+# Return code 2 #
 ### :II:e:2: Non existant vcs local file
 ### rm -rf $OPAMROOT
 ### opam init --bypass-checks --bare --no-setup --config git+file://$BASEDIR/config-files/config
-Fatal error:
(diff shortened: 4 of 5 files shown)


rjbou (changes requested):
Discussed in dev meeting, it needs more design. At current stage, it is not possible to have a local path that is a vsc.
I've added some `opam init --config` tests

src/core/opamUrl.ml
```diff
@@ -211,20 +211,42 @@ let base_url url =
...
+  | LocalPath
+  | VersionControl of version_control
+  | Http
+  | Ssh
+
+let kind = function
+  | {backend = (#version_control as vcs); _} -> VersionControl vcs
+  | url ->
+    match local_path url, url with
+    | Some _, {backend = `rsync; _} -> LocalPath
```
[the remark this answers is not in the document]
kit-ty-kate: no because there is no guarantee that the files are in the state that you want on disk (it could point to a different branch, be a bare git repo and thus have no files checked-out, …)

src/repository/opamDownload.ml
```diff
@@ -171,16 +171,25 @@ let really_download
...
+    if OpamFilename.equal src dst then Done () else
       (if OpamFilename.exists dst then
          if overwrite then OpamFilename.remove dst else
            OpamSystem.internal_error "The downloaded file will overwrite %s."
              (OpamFilename.to_string dst);
        OpamFilename.copy ~src ~dst;
        Done ())
-  | None ->
+  | VersionControl vcs ->
+    OpamConsole.error_and_exit `Internal_error
```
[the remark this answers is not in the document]
kit-ty-kate: i strongly think this should be an error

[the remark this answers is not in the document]
kit-ty-kate: That's outside of the scope of this PR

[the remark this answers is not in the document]
rjbou: Even if it is outside the scope, it is something that we could discuss, at least to see if it worth opening an issue

src/core/opamUrl.ml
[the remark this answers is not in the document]
rjbou: For several functions, we need to have a local path that is a version control. Sometimes we just need to check if it is a local vcs-ed path, sometimes just local path, sometimes local path non vcs-ed.

src/repository/opamDownload.ml
```diff
@@ -229,16 +229,25 @@ let really_download
   Done ()
 
 let download_as ?quiet ?validate ~overwrite ?compress ?checksum url dst =
-  match OpamUrl.local_file url with
-  | Some src ->
-    if src = dst then Done () else
+  match OpamUrl.kind url with
```
rjbou: This change is not a change related to the PR purpose

src/client/opamCommands.ml
```diff
@@ -131,9 +131,24 @@ let get_init_config ~no_sandboxing ~no_default_config_file ~add_config_file =
...
+              "File does not exist: %s."
+              (OpamUrl.to_string url)
+        | VersionControl _ ->
+          OpamConsole.error_and_exit `Bad_arguments
+            "Version control url not supported for %S"
+            (OpamUrl.to_string url);
+        | Ssh ->
+          OpamConsole.error_and_exit `Bad_arguments
+            "Ssh url is not supported for %S"
+            (OpamUrl.to_string url);
```
rjbou: I think that we should go through the entire list and then fail gracefully with a complete error message. Here, if some have as argument `--config non-existent --config vcs-file --config ssh-file`, they'il need to run opam init 3 times to remove all errors.

rjbou: At the moment, parsing failure are reported as is
```
$ opam init --config /tmp/parse-error
Configuring from /tmp/parse-error and then from built-in defaults.
[ERROR] Error in configuration file, fix it, use '--no-opamrc', or check your '--config FILE' arguments:
At /tmp/t:1:136-1:137::
Parse error
```
Do we want to handle more gracefully that in that PR ?

rjbou: It is possible to factorise these cases
```suggestion
        | VersionControl _ | Ssh as kind ->
          OpamConsole.error_and_exit `Bad_arguments
            "%s url not supported for %S"
            (OpamUrl.string_of_kind kind)
            (OpamUrl.to_string url);
```

src/repository/opamDownload.ml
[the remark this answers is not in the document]
rjbou: I may have put my comment not in the good place, cf comment in https://github.com/ocaml/opam/pull/5979#discussion_r2547210374.

rjbou (commented):
The purpose of the PR and the general way to do it looks good to me.
However, i have several comments.
On the main topic of this PR:
* should we require that all config files are good, or should we skip the ones that are not compliant (bc not found, or backend, or even parsing failures). Currently, parsing errors are erroring at the first failure (related to comment XXX)
* at the moment, parsing failure are reported as is. Do we want to handle more gracefully that in that PR ? 
```
$ opam init --config /tmp/parse-error
Configuring from /tmp/parse-error and then from built-in defaults.
[ERROR] Error in configuration file, fix it, use '--no-opamrc', or check your '--config FILE' arguments:
At /tmp/t:1:136-1:137::
Parse error
```

On the addition of the type `OpamUrl.kind`.
> I opened #5978 to think about removing local_{file,dir} in the future, but for now this would be a fairly intrusive change.

I don't agree. I had to make those changes to be able to precise what are the needs of the type (https://github.com/ocaml/opam/pull/5979#discussion_r2547135194) and confirm the intuition i had.
Having a first PR that adds this type and use it through the codebase without `local_file` and `local_dir` will permit us to refine its definition and usage. Then, we'll be able to use it directly in this PR.
I have a dirty/draft local branch with it, and what i saw is having this type helps to have a more fine grained control (checks) and information (when reading code) about what we expect from those files/directories.

src/client/opamCommands.ml
kit-ty-kate: I don't think this is worth spending the time to design this just to support such a niche corner-case

src/core/opamUrl.ml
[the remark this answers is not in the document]
kit-ty-kate: The problem with local VCS paths, is that there are no telling that a path to a file is the correct one. Given that we do not know the current branch of whatever is checked-out locally, or even if it is checked-out at all (e.g. bare git repositories)

src/repository/opamDownload.ml
kit-ty-kate: Fair enough, i removed that commit

tests/reftests/init.test
```diff
@@ -303,3 +303,66 @@ wrap-remove-commands: ["%{hooks}%/sandbox.sh" "remove"] {os = "linux" | os = "ma
...
+# Return code 2 #
+### :III:d:5: Non existant distant VCS file
+### rm -rf $OPAMROOT
+### opam init --bypass-checks --bare --no-setup --config git+file://som.e/distant/config
+[ERROR] Version control url not supported for "git+file://som.e/distant/config"
+# Return code 2 #
+### :II:d:1: Non existant http file
+### rm -rf $OPAMROOT
+### opam --version >$ OPAMVERSION
+### opam init --bypass-checks --bare --no-setup --config http://som.e/url/to/config | sed-cmd curl | "${OPAMVERSION}" -> "current" | "[^ ]*build_[^ ]*\.part" -> file.part
```
kit-ty-kate:
```suggestion
### opam init --bypass-checks --bare --no-setup --config http://som.e/url/to/config | sed-cmd curl | "${OPAMVERSION}" -> "current" | "[^ ]*\.part" -> file.part
```

src/client/opamCommands.ml
rjbou: Fair enough. Opened #6814 to track it.

master_changes.md
```diff
@@ -18,6 +18,7 @@ users)
 ## Plugins
 
 ## Init
+  * Display an appropriate error message when the file given to `opam init --config` does not exist or is in a VCS [#5979 @kit-ty-kate - fix #5971]
```
```suggestion
  * ✘ Display an appropriate error message when the file given to `opam init --config` does not exist or is in a VCS. This changes the behaviour for VCS local urls that was previously retrieved. [#5979 @kit-ty-kate - fix #5971]
```

rjbou (approved):
I've updated the PR with some tests (several errors and parsing errors) and fixed their numbering, and removed a trailing `.` at the end of the non existent file message.
Apart the precision of behaviour change in changes, lgtm!

kit-ty-kate:
ocaml-benchmarks is hitting a 502 bad gateway so i'm guessing it won't start. Ignoring.

PR #5981, opened 2024-05-29
Add a "trust cache" repository option

AltGr:


AltGr:
This is an attempted quickfix for the automated repo cache at https://gitlab.ocamlpro.com/OCamlPro/opam-repository/-/tree/cached : the repository fetched from CI has git LFS handles for the cached files, but not the full files, so their hashes can't be recomputed.

AltGr:
The OCamlPro repo cache (https://gitlab.ocamlpro.com/OCamlPro/opam-repository/-/tree/cached) still needs this feature

kit-ty-kate:
Sorry for the lack of feedback. We had discussed it in a meeting a while back and forgot to write the conclusion here. I personally think having this environment variable available for everyone is too dangerous. I think having an argument limited to opam admin cache would be just fine although i'm not sure what's the best API to have for this, maybe just a new `?unsafe_trust_cache:bool` argument to `fetch_from_cache` would be enough

PR #5982, opened 2024-05-30
Support Nix depexts with `opam env`

RyanGibb:
Nix doesn't install packages in the traditional sense -- instead it puts them in a store and makes them available through environment variables. I.e., `nix-shell -p gcc` will drop us into a shell with gcc in the `$PATH`.

We can set appropriate environment variables with Opam to make system dependencies (depexts) available with Nix.
Similar to how `nix-shell` works under the hood, we create a Nix derivation such as
```
{ pkgs ? import <nixpkgs> {} }:
with pkgs;
let
  packages = [ gmp ];
  inputs = with buildPackages; packages ++ [ pkg-config ];
in
stdenv.mkDerivation {
  name = "opam-nix-env";
  nativeBuildInputs = inputs;

  phases = [ "buildPhase" ];

  buildPhase = ''
vars=("NIX_CC" "NIX_CC_FLAGS" "NIX_CFLAGS_COMPILE" "NIX_CC_WRAPPER_TARGET_HOST_x86_64_unknown_linux_gnu" "NIX_LDFLAGS" "PKG_CONFIG_PATH")
for var in "''${vars[@]}"; do
  escaped="$(echo "''${!var}" | sed -e 's/^$/@/' -e 's/ /\\ /g')"
  echo "$var    =       $escaped        Nix" >> $out
done
echo "PATH      =       $PATH   Nix"
  '';

  preferLocalBuild = true;
}
```

Which we can build to output a file with environment variables that make depexts available, in Opam's environment variable format. This file is a Nix store root, so it's dependencies won't be garbage collected by Nix until the file is removed.

This has been tested and successfully creates an environment to provide packages `conf-gmp` and `conf-libseccomp`, as well as depexts directly.


src/state/opamSysInteract.ml
```diff
@@ -1009,14 +1026,45 @@ let install_packages_commands_t ?(env=OpamVariable.Map.empty) config sys_package
     [`AsUser (Commands.msys2 config),
      "-Su"::"--noconfirm"::packages], None
   | Netbsd -> [`AsAdmin "pkgin", yes ["-y"] ("install" :: packages)], None
+  | Nix ->
+    let open OpamFilename in
+    let dir = OpamPath.Switch.meta OpamStateConfig.(!r.root_dir) switch in
+    let packages_str = String.concat " " (List.rev (OpamSysPkg.Set.fold (fun p l -> OpamSysPkg.to_string p :: l) sys_packages [])) in
```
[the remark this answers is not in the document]
RyanGibb: That makes sense, I don't remember why I put that there. Thanks kate!

RyanGibb:
Rebased on master

RyanGibb:
Looks like I'm running into https://github.com/NixOS/nix/issues/10587 with the Docker image

src/state/opamSysInteract.ml
```diff
@@ -1080,14 +1092,51 @@ let install_packages_commands_t ?(env=OpamVariable.Map.empty) config sys_package
...
+{|{ pkgs ? import <nixpkgs> {} }:
+with pkgs;
+stdenv.mkDerivation {
+  name = "opam-nix-env";
+  nativeBuildInputs = with buildPackages; [ |} ^ packages ^ {| ];
+
+  phases = [ "buildPhase" ];
+
+  buildPhase = ''
+vars=("NIX_CC" "NIX_CC_FLAGS" "NIX_CFLAGS_COMPILE" "NIX_CC_WRAPPER_TARGET_HOST_x86_64_unknown_linux_gnu" "NIX_LDFLAGS" "PKG_CONFIG_PATH")
```
RyanGibb, line 1118: We might want to make these environment variables configurable in order to support new dependencies from Nix, which might require additional environment variables, without cutting a new release of Opam.

RyanGibb:
Rebased on master -- and could do with another depext CI run if it needs re-approval!

RyanGibb:
Okay! The CI is failing only due to https://github.com/ocaml/opam-repository/pull/26261 (I've got it working locally with my branch of opam-repository)

master_changes.md
```diff
@@ -68,6 +68,8 @@ users)
  * Use --quiet-mode noinput for the internal Cygwin installation (which is definitely a fully-specified command line) and --quiet-mode unattended for external Cygwin installations (in case the user does need to select something, e.g. a mirror) [#6046 @dra27]
   * [BUG] Fix apt/debian lookup for installed packages [#6054 @rjbou]
 
+[NEW] Support providing external dependencies with Nix
```
kit-ty-kate, line 71:
```suggestion
  * [NEW] Support providing external dependencies with Nix [#5982 @RyanGibb]
```

src/client/opamSolution.ml
```diff
@@ -1291,23 +1294,23 @@ let install_depexts ?(force_depext=false) ?(confirm=true) t packages =
                                  f status.OpamSysPkg.s_available }
               sys_map
           | None -> sys_map)
-        packages
+        new_packages
         (Lazy.force t.sys_packages)
     in
-    { t with sys_packages = lazy sys_packages }
+    Some ({ t with sys_packages = lazy sys_packages })
```
kit-ty-kate, line 1300:
```suggestion
    Some { t with sys_packages = lazy sys_packages }
```

src/state/opamSysInteract.ml
```diff
@@ -1093,14 +1105,51 @@ let install_packages_commands_t ?(env=OpamVariable.Map.empty) config sys_package
...
+for var in "''${vars[@]}"; do
+  escaped="$(echo "''${!var}" | sed -e 's/^$/@/' -e 's/ /\\ /g')"
+  echo "$var	=	$escaped	Nix" >> $out
+done
+echo "PATH	+=	$PATH	Nix" >> $out
```
kit-ty-kate, line 1136: I'm not sure to understand when the bash script starts and end but if those lines are indeed a bash script then these variables need to be quoted
```suggestion
  echo "$var	=	$escaped	Nix" >> "$out"
done
echo "PATH	+=	$PATH	Nix" >> "$out"
```

src/state/opamSysInteract.ml
```diff
@@ -1093,14 +1105,51 @@ let install_packages_commands_t ?(env=OpamVariable.Map.empty) config sys_package
     [`AsUser (Commands.msys2 config),
      "-Su"::"--noconfirm"::packages], None
   | Netbsd -> [`AsAdmin "pkgin", yes ["-y"] ("install" :: packages)], None
+  | Nix ->
+    let open OpamFilename in
```
kit-ty-kate, line 1109: I'd personally rather avoid opens. It makes it harder to understand the code
```suggestion
```

src/state/opamEnv.ml
```diff
@@ -658,7 +658,14 @@ let compute_updates ?(force_path=false) st =
     in
     List.map resolve_separator_and_format updates
   in
-  switch_env @ pkg_env @ man_path @ [path]
+  let nix_env =
+    let open OpamFilename in
```
kit-ty-kate, line 662: same here
```suggestion
```

src/state/opamEnv.ml
```diff
@@ -658,7 +658,14 @@ let compute_updates ?(force_path=false) st =
...
+    create (OpamPath.Switch.meta OpamStateConfig.(!r.root_dir) st.switch) (basename (raw "nix.env"))
+    |> OpamFile.make
+    |> OpamFile.Environment.read_opt
+    |> Option.fold ~none:[] ~some:(List.map resolve_separator_and_format)
```
kit-ty-kate, line 666: I'm unsure this is the right place for this code. The function was only dealing with the build environment but this section now reads a file unconditionally (should it be done only when nix is detected?)

src/client/opamSolution.ml
```diff
@@ -1269,19 +1271,20 @@ let install_sys_packages ~map_sysmap ~confirm env config sys_packages t =
...
     OpamConsole.header_msg "Handling external dependencies";
     OpamConsole.msg "\n";
-    entry_point t sys_packages
+    entry_point t sys_packages required
   with Sys.Break as e -> OpamStd.Exn.finalise e give_up_msg
 
-let install_depexts ?(force_depext=false) ?(confirm=true) t packages =
+let install_depexts ?(force_depext=false) ?(confirm=true) t ~new_packages ~all_packages =
   let map_sysmap f t =
+    let t = Option.get t in
```
kit-ty-kate, line 1287: could you explain why all these `Option.get` are needed?

src/client/opamSolution.ml
```diff
@@ -1128,37 +1129,38 @@ let get_depexts ?(force=false) ?(recover=false) t packages =
...
+  let avail, required, nf =
+    OpamPackage.Set.fold (fun pkg (avail,req,nf) ->
         match OpamPackage.Map.find_opt pkg sys_packages with
         | Some sys ->
           OpamSysPkg.(Set.union avail sys.s_available),
+          OpamSysPkg.(Set.union req sys.s_required),
           OpamSysPkg.(Set.union nf sys.s_not_found)
-        | None -> avail, nf)
-      packages (OpamSysPkg.Set.empty, OpamSysPkg.Set.empty)
+        | None -> avail, req, nf)
```
kit-ty-kate, line 1145: i think it might be worth using a dedicated record instead of a tuple at this point. Otherwise it could become confusing very quickly with all those similar types around

src/client/opamSolution.ml
RyanGibb: Thanks! Is there a formatter that I can run on my changes at all?

kit-ty-kate: no we're not using ocamlformat. But i don't think it's needed anyway, your formatting looks fine otherwise

src/state/opamSysInteract.ml
RyanGibb: `$out` will be a [Nix store path](https://nix.dev/manual/nix/2.23/store/store-path), so shouldn't need quoted, but I don't see any harm in adding them.

src/client/opamSolution.ml
kit-ty-kate: As for formatters in general, we use ocp-indent but it would not have caught that as it is only a simple formatter, not a re-format tool

src/state/opamEnv.ml
RyanGibb: I think this fails gracefully if the file isn't there with the `Option.fold`, but perhaps we could add a check to only if `os-family = nixos`? I can also refactor this into a separate function `compute_nix_update` called my `OpamEnv.updates`?

src/client/opamSolution.ml
RyanGibb: It's because I don't have access to the switch state here https://github.com/ocaml/opam/pull/5982/commits/edd90302e11813b3ac51f37277fb40fd9f3159a9#diff-6426799f367266b2a65325c870931b51df35852e12ddbe8580e74f4eeeb58177R1606 afaik because it hasn't been created at this point. We need to pass it in to `OpamSysInteract.install*` in order for Nix to write to files in the switch, but we also need to support installing system packages before this switch has been created in the existing `OpamClient` logic (which will do nothing for the Nix backend).

src/state/opamSysInteract.ml
RyanGibb: I used them as I noticed their use in a fair bit of the rest of the codebase, but I'm happy to remove them if you prefer!

src/client/opamSolution.ml
RyanGibb: We have `opamSysPkg.status`?

RyanGibb: Though, we're also passing around the couple of `available` and `required` a lot, which perhaps warrants a separate record.

src/state/opamSysInteract.ml
```diff
@@ -1093,14 +1105,50 @@ let install_packages_commands_t ?(env=OpamVariable.Map.empty) config sys_package
...
+done
+echo "PATH	+=	$PATH	Nix" >> "$out"
+  '';
+
+  preferLocalBuild = true;
+}
+|} in
+    OpamFilename.write drvFile contents;
+    let envFile = OpamFilename.create dir (OpamFilename.basename (OpamFilename.raw "nix.env")) |> OpamFilename.to_string in
+    [`AsUser "nix-build", [ OpamFilename.to_string drvFile; "--out-link"; envFile ] ], None)
```
RyanGibb, line 1143: It would be good if we could trigger a re-build of any packages that depend on depexts if the drv path changes, as if they link against a previous derivation it could be GC'd. An alternative might be to keep the old `nix.env` files around as Nix GC roots.

src/state/opamSysInteract.ml
```diff
@@ -1093,14 +1105,50 @@ let install_packages_commands_t ?(env=OpamVariable.Map.empty) config sys_package
...
+    let packages = String.concat " "
+      (OpamSysPkg.Set.fold (fun p l -> OpamSysPkg.to_string p :: l)
+      OpamSysPkg.Set.Op.(sys_packages ++ required) [])
+    in
+    let contents =
+{|{ pkgs ? import <nixpkgs> {} }:
+with pkgs;
+stdenv.mkDerivation {
+  name = "opam-nix-env";
+  nativeBuildInputs = with buildPackages; [ |} ^ packages ^ {| ];
```
rjbou, line 1125: Is it possible to update that file for each installed package, like that there is no need to retrieve and pass the whole list of already installed packages and their system packages ?

.github/scripts/depexts/generate-actions.sh
```diff
@@ -102,6 +102,12 @@ RUN apt install -y $mainlibs $ocaml
 RUN apt install -y g++
 EOF
     ;;
+  nix)
+    cat >$dir/Dockerfile << EOF
+FROM nixos/nix
+RUN nix-channel --update
+RUN nix-env -i gnum4 git rsync patch bzip2 gnumake wget ocaml ocaml5.1.1-ocaml-compiler-libs unzip gcc diffutils patch getconf-glibc gnused
```
rjbou, line 109: while trying locally, i had a not found for `ocaml5.1.1-ocaml-compiler-libs`, it workd with `ocaml5.2.1-ocaml-compiler-libs`, but we can't change the script for each new version of the package. Is it possible to not have it version dependant?

src/state/opamSysInteract.ml
```diff
@@ -1093,14 +1105,50 @@ let install_packages_commands_t ?(env=OpamVariable.Map.empty) config sys_package
...
      "-Su"::"--noconfirm"::packages], None
   | Netbsd -> [`AsAdmin "pkgin", yes ["-y"] ("install" :: packages)], None
+  | Nix ->
+    (match switch with
+    | None ->
+        log "Nix depext must be passed switch";
+        [], None
+    | Some switch ->
+    let dir = OpamPath.Switch.meta OpamStateConfig.(!r.root_dir) switch in
+    let drvFile = OpamFilename.create dir (OpamFilename.basename (OpamFilename.raw "env.nix")) in
```
rjbou, line 1115: At least `nix.env` needs to be added in OpamPath as it is used in another place in the code.

src/state/opamSysInteract.ml
```diff
@@ -1093,14 +1105,50 @@ let install_packages_commands_t ?(env=OpamVariable.Map.empty) config sys_package
     [`AsUser (Commands.msys2 config),
      "-Su"::"--noconfirm"::packages], None
   | Netbsd -> [`AsAdmin "pkgin", yes ["-y"] ("install" :: packages)], None
+  | Nix ->
+    (match switch with
```
rjbou, line 1109: To avoid passing switch option, is it possible to add as optional argument a function that returns the needed prefix?

RyanGibb: Yes, I think so

src/state/opamSysInteract.ml
RyanGibb: Right!

src/state/opamSysInteract.ml
RyanGibb: I think this is more ergonomic, as it allows system packages to be uninstalled too. It's bridging opam's imperative mode to nix's declarative model.

src/state/opamSysInteract.ml
RyanGibb: I think capnproto might use a flag that isn't here.

RyanGibb:
David made an observation in today's Opam dev meeting that this PR will not remove depexts from the Nix derivation until the next install. This might effect reproducibility in some circumstances, but this isn't a huge deal as if Nix users want full reproducibility they can use opam2nix.

.github/scripts/depexts/generate-actions.sh
kit-ty-kate: querying and use it sounds like the way to go

kit-ty-kate:
```suggestion
RUN nix-env -iA nixpkgs.gnum4 nixpkgs.git nixpkgs.rsync nixpkgs.patch nixpkgs.bzip2 nixpkgs.gnumake nixpkgs.wget nixpkgs.ocaml nixpkgs.ocamlPackages.ocaml-compiler-libs nixpkgs.unzip nixpkgs.gcc nixpkgs.diffutils nixpkgs.patch nixpkgs.getconf nixpkgs.gnused
```

kit-ty-kate (commented):
while tracking down https://github.com/NixOS/nixpkgs/pull/388471 i tried this branch and it worked really well in a nixos docker container. I was able to build `opam-publish` which requires `openssl` and `gmp` without any issues at all, like magic.

src/client/opamSolution.ml
kit-ty-kate: I managed to find a solution to avoid the `Option.get` in https://github.com/ocaml/opam/pull/5982/commits/cecc358447457a17eb9c18ab170643c334a9e7fc (pushed as a fixup commit)

src/state/opamSysInteract.ml
kit-ty-kate: Done in https://github.com/ocaml/opam/pull/5982/commits/241d29cbc3594c9bee6126f29d713c32eb425c22 (pushed as a fixup commit)

src/state/opamSysInteract.ml
kit-ty-kate: After review i found that instead the switch state was more indicated if at least to avoid uses of the unsightly `OpamStateConfig.(!r.root_dir)`. This is implemented in the fixup commit https://github.com/ocaml/opam/pull/5982/commits/241d29cbc3594c9bee6126f29d713c32eb425c22

src/state/opamSysInteract.ml
```diff
@@ -1102,14 +1114,59 @@ let install_packages_commands_t ?(env=OpamVariable.Map.empty) config sys_package
...
+  | Nix ->
+    (match st with
+     | None ->
+       log "Nix depext must be passed switch";
+       [], None
+     | Some (st : _ OpamStateTypes.switch_state) ->
+       let dir = OpamPath.Switch.meta st.switch_global.root st.switch in
+       let drvFile =
+         OpamFilename.create dir
+           (OpamFilename.basename (OpamFilename.raw "env.nix"))
```
kit-ty-kate, line 1126: This file seems to be there only as a sort of temporary file for `nix-build`. I also find the naming convention a bit confusing when taken together with `nix.env`.
`env.nix` and `nix.env` are way too close in terms of name and it's easy to confuse between them.

Wouldn't some sort of nameless temporary file be more indicated if a file has to be used?
Couldn't we use a constant string together with environment variables for `packages` (and possibly `vars`) instead?

.github/scripts/depexts/generate-actions.sh
```diff
@@ -83,35 +83,41 @@ EOF
...
 RUN apt update
 RUN apt install -y $mainlibs $ocaml
 RUN apt install -y g++
 EOF
     ;;
+  nix)
+    cat > "$dir/Dockerfile" << EOF
+FROM nixos/nix
+RUN nix-channel --update
+RUN nix-env -iA nixpkgs.gnum4 nixpkgs.git nixpkgs.rsync nixpkgs.patch nixpkgs.bzip2 nixpkgs.gnumake nixpkgs.wget nixpkgs.ocaml nixpkgs.ocamlPackages.ocaml-compiler-libs nixpkgs.unzip nixpkgs.gcc nixpkgs.diffutils nixpkgs.patch nixpkgs.getconf nixpkgs.gnused
```
kit-ty-kate, line 111:
```suggestion
RUN nix-env -iA nixpkgs.gnum4 nixpkgs.git nixpkgs.rsync nixpkgs.patch nixpkgs.bzip2 nixpkgs.gnumake nixpkgs.wget nixpkgs.ocaml nixpkgs.ocamlPackages.ocaml-compiler-libs nixpkgs.unzip nixpkgs.gcc nixpkgs.diffutils nixpkgs.patch nixpkgs.getconf nixpkgs.gnused nixpkgs.gawk
```

.github/scripts/depexts/generate-actions.sh
```diff
@@ -83,35 +83,41 @@ EOF
...
 RUN apt update
 RUN apt install -y $mainlibs $ocaml
 RUN apt install -y g++
 EOF
     ;;
+  nix)
+    cat > "$dir/Dockerfile" << EOF
+FROM nixos/nix
+RUN nix-channel --update
+RUN nix-env -iA nixpkgs.gnum4 nixpkgs.git nixpkgs.rsync nixpkgs.patch nixpkgs.bzip2 nixpkgs.gnumake nixpkgs.wget nixpkgs.ocaml nixpkgs.ocamlPackages.ocaml-compiler-libs nixpkgs.unzip nixpkgs.gcc nixpkgs.diffutils nixpkgs.patch nixpkgs.getconf nixpkgs.gnused nixpkgs.gawk
```
kit-ty-kate, line 111: haha `nixpkgs.ocamlPackages.ocaml-compiler-libs` is in fact JaneStreet's `compiler-libs` wrapper library, not the compiler's compiler-libs as a separate package (which is otherwise frequently split as a separate package in some distributions for some reason)
```suggestion
RUN nix-env -iA nixpkgs.gnum4 nixpkgs.git nixpkgs.rsync nixpkgs.patch nixpkgs.bzip2 nixpkgs.gnumake nixpkgs.wget nixpkgs.ocaml nixpkgs.unzip nixpkgs.gcc nixpkgs.diffutils nixpkgs.patch nixpkgs.getconf nixpkgs.gnused nixpkgs.gawk
```

src/state/opamSysInteract.ml
```diff
@@ -1163,6 +1220,7 @@ let update ?(env=OpamVariable.Map.empty) config =
     | Macports -> Some (`AsAdmin "port", ["sync"])
     | Msys2 -> Some (`AsUser (Commands.msys2 config), ["-Sy"])
     | Netbsd -> None
+    | Nix -> None
```
kit-ty-kate, line 1223: What about:
```suggestion
    | Nix -> Some (`AsUser "nix-channel", ["--update"])
```

.github/scripts/depexts/generate-actions.sh
RyanGibb: Ah, magic!

src/state/opamSysInteract.ml
RyanGibb: This wouldn't work for flakes, but is perhaps better than nothing

RyanGibb:
Thanks for the comments and commits @kit-ty-kate ! I'm glad it was useful :-) Apologies I haven't had a lot of time to work on this PR recently

src/state/opamSysInteract.ml
kit-ty-kate: is there a way to get that list of variables in a more dynamic way?

RyanGibb: An `env` picked up variables that broke things. E.g. for the Nix derivation,
```
{ pkgs ? import <nixpkgs> {} }:
with pkgs;
stdenv.mkDerivation {
  name = "opam-nix-env";
  nativeBuildInputs = with buildPackages; [ pkg-config gmp capnproto ];

  phases = [ "buildPhase" ];

  buildPhase = ''
env >> "$out"
  '';

  preferLocalBuild = true;
}
```

The environment is:
```
SHELL=/nix/store/8vpg72ik2kgxfj05lc56hkqrdrfl8xi9-bash-5.2p37/bin/bash
NIX_BUILD_CORES=8
configureFlags=
mesonFlags=
PKG_CONFIG_PATH=/nix/store/11szpaxyi7n640kc3n3jhs779a7jhnb9-gmp-with-cxx-6.3.0-dev/lib/pkgconfig:/nix/store/alygr3k4q0sygjvq82aiwg692r9lz0m7-capnproto-1.0.2/lib/pkgconfig:/nix/store/a5vsp7s1xa0fz2f0sc2gqk84lcwwgnak-openssl-3.3.3-dev/lib/pkgconfig:/nix/store/ccrjbcfl5zcdp0n5mh69f4airrb9g8m4-zlib-1.3.1-dev/lib/pkgconfig
shell=/nix/store/8vpg72ik2kgxfj05lc56hkqrdrfl8xi9-bash-5.2p37/bin/bash
depsHostHost=
STRINGS=strings
depsTargetTarget=
stdenv=/nix/store/szfi0xd0cwdaldhkp8vlgi7jfv662nmd-stdenv-linux
builder=/nix/store/8vpg72ik2kgxfj05lc56hkqrdrfl8xi9-bash-5.2p37/bin/bash
phases=buildPhase
PWD=/build
SOURCE_DATE_EPOCH=315532800
NIX_ENFORCE_NO_NATIVE=1
CXX=g++
TEMPDIR=/build
system=x86_64-linux
TZ=UTC
HOST_PATH=/nix/store/9m68vvhnsq5cpkskphgw84ikl9m6wjwp-coreutils-9.5/bin:/nix/store/vc2d1bfy1a5y1195nq7k6p0zcm6q89nx-findutils-4.10.0/bin:/nix/store/lr3cvmq5ahqcj29p8c6m0fdl1y8krc86-diffutils-3.10/bin:/nix/store/3ks7b6p43dpvnlnxgvlcy2jaf1np37p2-gnused-4.9/bin:/nix/store/qjsj5vnbfpbg6r7jhd7znfgmcy0arn8n-gnugrep-3.11/bin:/nix/store/3p3fwczck2yn1wwfjnymzkz8w11vbvg7-gawk-5.3.1/bin:/nix/store/pmg7hw4ar16dhx5hk6jswvxy349wzsm7-gnutar-1.35/bin:/nix/store/nc394xps4al1r99ziabqvajbkrhxr5b7-gzip-1.13/bin:/nix/store/yyfzan4mn874v885jy6fs598gjb31c4l-bzip2-1.0.8-bin/bin:/nix/store/m3da56j3r7h4hp0kr8v1xsnk16x900yf-gnumake-4.4.1/bin:/nix/store/8vpg72ik2kgxfj05lc56hkqrdrfl8xi9-bash-5.2p37/bin:/nix/store/zfjv48ikkqn3yhi8zi7lvqvxyscbg87n-patch-2.7.6/bin:/nix/store/rbns8mzghhqxih4hj2js4mb4s6ivy5d1-xz-5.6.3-bin/bin:/nix/store/p751fjd81h3926ivxsq0x20lz5j7yscc-file-5.45/bin
doInstallCheck=
HOME=/homeless-shelter
NIX_BINTOOLS=/nix/store/1mv8pj4nxwnd9bbxshljc9p4cnl3rakj-binutils-wrapper-2.43.1
GZIP_NO_TIMESTAMPS=1
depsTargetTargetPropagated=
cmakeFlags=
NIX_SSL_CERT_FILE=/no-cert-file.crt
outputs=out
NIX_STORE=/nix/store
TMPDIR=/build
LD=ld
NIX_ENFORCE_PURITY=1
buildPhase=env >> "$out"

READELF=readelf
doCheck=
NIX_LOG_FD=2
depsBuildBuild=
TERM=xterm-256color
SIZE=size
propagatedNativeBuildInputs=
strictDeps=
AR=ar
AS=as
TEMP=/build
NIX_BINTOOLS_WRAPPER_TARGET_HOST_x86_64_unknown_linux_gnu=1
SHLVL=1
NIX_BUILD_TOP=/build
NM=nm
NIX_CFLAGS_COMPILE= -frandom-seed=813r1c5rl2 -isystem /nix/store/11szpaxyi7n640kc3n3jhs779a7jhnb9-gmp-with-cxx-6.3.0-dev/include -isystem /nix/store/alygr3k4q0sygjvq82aiwg692r9lz0m7-capnproto-1.0.2/include -isystem /nix/store/a5vsp7s1xa0fz2f0sc2gqk84lcwwgnak-openssl-3.3.3-dev/include -isystem /nix/store/ccrjbcfl5zcdp0n5mh69f4airrb9g8m4-zlib-1.3.1-dev/include -isystem /nix/store/11szpaxyi7n640kc3n3jhs779a7jhnb9-gmp-with-cxx-6.3.0-dev/include -isystem /nix/store/alygr3k4q0sygjvq82aiwg692r9lz0m7-capnproto-1.0.2/include -isystem /nix/store/a5vsp7s1xa0fz2f0sc2gqk84lcwwgnak-openssl-3.3.3-dev/include -isystem /nix/store/ccrjbcfl5zcdp0n5mh69f4airrb9g8m4-zlib-1.3.1-dev/include
patches=
buildInputs=
preferLocalBuild=1
SSL_CERT_FILE=/no-cert-file.crt
NIX_PKG_CONFIG_WRAPPER_TARGET_HOST_x86_64_unknown_linux_gnu=1
depsBuildTarget=
OBJCOPY=objcopy
out=/nix/store/813r1c5rl2va9jgby6xx00zqrklxi2fk-opam-nix-env
STRIP=strip
XDG_DATA_DIRS=/nix/store/ry0993a35awd6d74pss5gjdwj5dsrf1p-pkg-config-wrapper-0.29.2/share:/nix/store/vpg96mfr1jw5arlqg831i69g29v0sdb3-zlib-1.3.1/share:/nix/store/9wwkzvmrlv43y71rbw8sbh9bamc62a16-patchelf-0.15.0/share
TMP=/build
OBJDUMP=objdump
PATH=/nix/store/ry0993a35awd6d74pss5gjdwj5dsrf1p-pkg-config-wrapper-0.29.2/bin:/nix/store/alygr3k4q0sygjvq82aiwg692r9lz0m7-capnproto-1.0.2/bin:/nix/store/lx93h1j6wwbwj7pfsjbl21lskl2pji67-openssl-3.3.3-bin/bin:/nix/store/9wwkzvmrlv43y71rbw8sbh9bamc62a16-patchelf-0.15.0/bin:/nix/store/gnd8f9h2ycxrfrvrga508c4n9cxy6720-gcc-wrapper-13.3.0/bin:/nix/store/yg4ahy7gahx91nq80achmzilrjyv0scj-gcc-13.3.0/bin:/nix/store/2kiskq24j06g4qw3xs8zsy2dkawh4gxk-glibc-2.40-66-bin/bin:/nix/store/9m68vvhnsq5cpkskphgw84ikl9m6wjwp-coreutils-9.5/bin:/nix/store/1mv8pj4nxwnd9bbxshljc9p4cnl3rakj-binutils-wrapper-2.43.1/bin:/nix/store/22qnvg3zddkf7rbpckv676qpsc2najv9-binutils-2.43.1/bin:/nix/store/9m68vvhnsq5cpkskphgw84ikl9m6wjwp-coreutils-9.5/bin:/nix/store/vc2d1bfy1a5y1195nq7k6p0zcm6q89nx-findutils-4.10.0/bin:/nix/store/lr3cvmq5ahqcj29p8c6m0fdl1y8krc86-diffutils-3.10/bin:/nix/store/3ks7b6p43dpvnlnxgvlcy2jaf1np37p2-gnused-4.9/bin:/nix/store/qjsj5vnbfpbg6r7jhd7znfgmcy0arn8n-gnugrep-3.11/bin:/nix/store/3p3fwczck2yn1wwfjnymzkz8w11vbvg7-gawk-5.3.1/bin:/nix/store/pmg7hw4ar16dhx5hk6jswvxy349wzsm7-gnutar-1.35/bin:/nix/store/nc394xps4al1r99ziabqvajbkrhxr5b7-gzip-1.13/bin:/nix/store/yyfzan4mn874v885jy6fs598gjb31c4l-bzip2-1.0.8-bin/bin:/nix/store/m3da56j3r7h4hp0kr8v1xsnk16x900yf-gnumake-4.4.1/bin:/nix/store/8vpg72ik2kgxfj05lc56hkqrdrfl8xi9-bash-5.2p37/bin:/nix/store/zfjv48ikkqn3yhi8zi7lvqvxyscbg87n-patch-2.7.6/bin:/nix/store/rbns8mzghhqxih4hj2js4mb4s6ivy5d1-xz-5.6.3-bin/bin:/nix/store/p751fjd81h3926ivxsq0x20lz5j7yscc-file-5.45/bin
propagatedBuildInputs=
CC=gcc
NIX_CC=/nix/store/gnd8f9h2ycxrfrvrga508c4n9cxy6720-gcc-wrapper-13.3.0
depsBuildTargetPropagated=
depsBuildBuildPropagated=
NIX_CC_WRAPPER_TARGET_HOST_x86_64_unknown_linux_gnu=1
CONFIG_SHELL=/nix/store/8vpg72ik2kgxfj05lc56hkqrdrfl8xi9-bash-5.2p37/bin/bash
__structuredAttrs=
RANLIB=ranlib
NIX_HARDENING_ENABLE=bindnow format fortify fortify3 pic relro stackprotector strictoverflow zerocallusedregs
NIX_LDFLAGS=-rpath /nix/store/813r1c5rl2va9jgby6xx00zqrklxi2fk-opam-nix-env/lib  -L/nix/store/nn7nj7xq8gdfyxqhchphzia7i34rwb0v-gmp-with-cxx-6.3.0/lib -L/nix/store/alygr3k4q0sygjvq82aiwg692r9lz0m7-capnproto-1.0.2/lib -L/nix/store/mqvgaq2bnb1701l7zcn0d6vd7w8b0z7l-openssl-3.3.3/lib -L/nix/store/vpg96mfr1jw5arlqg831i69g29v0sdb3-zlib-1.3.1/lib -L/nix/store/nn7nj7xq8gdfyxqhchphzia7i34rwb0v-gmp-with-cxx-6.3.0/lib -L/nix/store/alygr3k4q0sygjvq82aiwg692r9lz0m7-capnproto-1.0.2/lib -L/nix/store/mqvgaq2bnb1701l7zcn0d6vd7w8b0z7l-openssl-3.3.3/lib -L/nix/store/vpg96mfr1jw5arlqg831i69g29v0sdb3-zlib-1.3.1/lib
nativeBuildInputs=/nix/store/ry0993a35awd6d74pss5gjdwj5dsrf1p-pkg-config-wrapper-0.29.2 /nix/store/11szpaxyi7n640kc3n3jhs779a7jhnb9-gmp-with-cxx-6.3.0-dev /nix/store/alygr3k4q0sygjvq82aiwg692r9lz0m7-capnproto-1.0.2
name=opam-nix-env
PKG_CONFIG=pkg-config
depsHostHostPropagated=
_=/nix/store/9m68vvhnsq5cpkskphgw84ikl9m6wjwp-coreutils-9.5/bin/env
```

Things like `SHELL` and `HOME` are obviously not things we want to set.

So it looks like `nix develop` has a [list of variables to ignore](https://github.com/NixOS/nix/blob/e4bda20918ad2af690c2e938211a7d362548e403/src/nix/develop.cc#L308-L325), rather than a list to include (though both are configurable). Perhaps we could go with this exclusion approach instead?

src/client/opamSolution.ml
RyanGibb: I've switched to using `OpamSysPkg.status` in this function!

kit-ty-kate:
@RyanGibb Do you know why `autopin.test` fails? Looking briefly at the changes i can't really find the cause

RyanGibb:
Hmm, I'm not sure. It seems the depext mechanism isn't being triggered when the test thinks it should, but I can't figure out why. I think this test is using the `Dummy` depext mechanism which (I think?) doesn't list installed packages as installed so retriggers the depext mechanism every time. I don't know why these changes would effect that behaviour, since we don't touch the `Dummy` path. It's also strange that this is the only failing test.

RyanGibb:
I'm wondering if it could be something to do with the handling of available/required/installed packages, and am thinking of changing [`packages_status`](https://github.com/ocaml/opam/pull/5982/files#diff-a657e0822f7ed4a1177c0ea985eeab221d568a0911c27f8ab07ea9e5c6e6eb23L19) to return `OpamSysPkg.status`.

kit-ty-kate:
I found the issue. One of the functions used `all_packages` instead of `new_packages`.
I gotta say that i find these (`old_packages`, `new_packages`, `all_packages`) naming scheme pretty confusing.
What do they refer to? What's new, old and all?

src/client/opamClient.ml
```diff
@@ -1613,8 +1614,13 @@ let check_for_sys_packages config system_packages =
         |> OpamVariable.Map.of_list
       in
       (*Lazy.force header;*)
-      OpamSolution.print_depext_msg set;
-      OpamSolution.install_sys_packages ~confirm:true env config missing ()
+      OpamSolution.print_depext_msg (available, missing);
+      let _ : _ option =
+        OpamSolution.install_sys_packages
```
kit-ty-kate, line 1619: why does a "check" need to install packages?

src/client/opamClient.ml
```diff
@@ -1602,9 +1602,10 @@ let update_with_init_config ?(overwrite=false) config init_config =
 
 let check_for_sys_packages config system_packages =
   if system_packages <> [] then
-    let ((missing, _) as set) =
+    let (available, required, missing) =
```
kit-ty-kate, line 1605: fun fact: the original code got it backwards. I'm not sure what it means in terms of bugs though. We might want to have a new test case for this.

src/client/opamClient.ml
RyanGibb: I might have to defer to dra27 on that one :-) It was added in this commit: https://github.com/ocaml/opam/commit/fa4d39f178253db70e1e33baf4ed48357bcc5560#diff-6426799f367266b2a65325c870931b51df35852e12ddbe8580e74f4eeeb58177R1565-R1580

src/client/opamClient.ml
RyanGibb: Oh dear 😅 Thanks for fixing this Kate!

RyanGibb:
> What do they refer to? What's new, old and all?

So to support Nix depexts we have to pass the not only the `packages` we want to install, but also the `old_packages` which are already in the derivation, since we recreate it from scratch each time. https://github.com/ocaml/opam/pull/5982/files#diff-6c0a6590c49c5a229f469d627b9e9484f1b20c1fb5fb1c8102d64a3d2c1d09b6R476

We change install_depexts to also pass in all the depexts in the switch https://github.com/ocaml/opam/pull/5982/files#diff-e8aa4752d602eeda81985e2ee34e48c4d6441eef00b51f5461c9cab855a0c0c3R1387 (which requires plumbing switch state through things), and we set old_packages to the difference between `all_packages` and `new_packages`.

I think we could perhaps create a new record to hold this 3-tuple, and the existing 2-tuple (which would have prevented this bug).

.github/scripts/depexts/generate-actions.sh
```diff
@@ -83,35 +83,44 @@ EOF
...
 RUN apt update
 RUN apt install -y $mainlibs $ocaml
 RUN apt install -y g++
 EOF
     ;;
+  nix)
+    cat > "$dir/Dockerfile" << EOF
+FROM nixos/nix
+RUN nix-channel --update
+RUN nix-env -iA nixpkgs.gnum4 nixpkgs.git nixpkgs.rsync nixpkgs.patch nixpkgs.bzip2 nixpkgs.gnumake nixpkgs.wget nixpkgs.ocaml nixpkgs.unzip nixpkgs.gcc nixpkgs.diffutils nixpkgs.patch nixpkgs.getconf nixpkgs.gnused nixpkgs.gawk
```
kit-ty-kate, line 111: Something is seriously seriously wrong with this nixos image, i've sent a bug-report upstream: https://github.com/NixOS/nix/issues/13023

In the meantime we can keep this as-is

kit-ty-kate (changes requested):
Using the latest commit (8e5d628) of this PR, i'm getting some serious issues when the current `os-family` is unknown:
```
$ opam init
[...]
Fatal error: External dependency handling not supported for OS family 'linux'.
```
which didn't happen before

kit-ty-kate:
Given that `opam env` isn't changed when a depext has been installed (for good reasons, you don't want to have your environment replaced by another one just to get access to some tool), i'm wondering whether we should notify the users (when the use nix) that their current environment won't be modified, so that they are not surprised when they install `conf-pkg-config` or something and do not see it in their current environment.

For example if someone wants to work on a local project that use tools such as `pkg-config`, `opam install . --deps-only` won't be enough to get a working environment to build it locally, they would have to install system dependencies manually too.

So to sum up, the method chosen by this PR is great to install packages, but doesn't help when developing (which users arguably use opam more for). I think this is fine, but users definitely need to get notified when a depext is being installed.

kit-ty-kate:
@RyanGibb regarding my last comment, is there a nix command that one could run to enter a clean new shell with an environment as if opam was building the package? Would `nix-shell` with one `-p` for each package that was installed be sufficient? (e.g. `nix-shell -p pkg-config -p openssl -p gmp`)

If this is reasonable, maybe we cold display it as a hint for users to run after an `opam install --deps-only <local dir>` build where the dependency tree contains depexts?

RyanGibb:
Oh, good point! `nix-shell -p pkg-config openssl gmp etc...` should work (though requires a channel to be set up). An annoyance with this approach is packages from these ephemeral shells can be GC'd, breaking built artifacts.

kit-ty-kate:
wouldn't these packages be part of the `env.nix` file and thus be listed in the Nix root thingy that would make them not GCable? (cf. the PR description)

RyanGibb:
Ah, yes, you're correct. Assuming they haven't changed the channel, that should work quite nicely!

.github/scripts/depexts/generate-actions.sh
```diff
@@ -96,6 +96,19 @@ RUN yum install -y $mainlibs
...
+  mainlibs=${mainlibs/tar/}
+  mainlibs=$(echo "$mainlibs" | sed 's/\([[:alnum:]]\+\)/nixpkgs.\1/g')
+  additionallibs="gcc diffutils getconf gnused gawk"
+  additionallibs=$(echo "$additionallibs" | sed 's/\([[:alnum:]]\+\)/nixpkgs.\1/g')
+  # We don't use $ocaml as nix compiler lib is not main ocaml compiler
+    cat > "$dir/Dockerfile" << EOF
+FROM nixos/nix
+RUN nix-channel --update
+RUN nix-env -iA $mainlibs $additionallibs nixpkgs.ocaml
+EOF
```
kit-ty-kate, line 111:
```suggestion
EOF
  ;;
```

.gitignore
```diff
@@ -100,3 +100,4 @@ src/x_build_libs.ocp
 # dev setup
 /.ocamlinit
 /release/out
+.github/actions
```
kit-ty-kate, line 103: Without the `/` this can match any prefix:
```suggestion
/.github/actions
```

src/state/opamSysInteract.ml
```diff
@@ -1040,42 +1100,114 @@ let install_packages_commands_t ?(env=OpamVariable.Map.empty) ~to_show config sy
...
+  -u TEMP \
+  -u TEMPDIR \
+  -u TERM \
+  -u TMP \
+  -u TMPDIR \
+  -u TZ \
+  -u UID \
+  -u PATH \
+  -u XDG_DATA_DIRS \
+  -u  self-referential \
```
kit-ty-kate, line 1182:
```suggestion
  -u self-referential \
```

src/state/opamSysInteract.ml
```diff
@@ -966,29 +965,90 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
...
+  | Dummy _
+  | Freebsd
+  | Gentoo
+  | Homebrew
+  | Macports
+  | Msys2
+  | Netbsd
+  | Openbsd
+  | Suse
+    -> false
```
kit-ty-kate, line 1007: Maybe it's preferable to not to that much vertical space
```suggestion
  | Alpine | Altlinux | Arch | Centos | Cygwin | Debian | Dummy _
  | Freebsd | Gentoo | Homebrew | Macports | Msys2 | Netbsd
  | Openbsd | Suse -> false
```

src/state/opamSysInteract.ml
```diff
@@ -966,29 +965,90 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
...
+         but that would involve an expensive Nixpkgs evaluation.
+         Saying no packages are installed results in a warning that
+         conf packages depend on a 'system package that can no longer
+         be found.' But omitting them will mean that they won't be
+         added to the Nix derivation.
+      *)
+      let s_available = packages in
+      let s_not_found = OpamSysPkg.Set.empty in
+      let open OpamSysPkg in
+      { s_available; s_not_found }
```
kit-ty-kate, line 980:
```suggestion
      { OpamSysPkg.s_available; s_not_found }
```

src/state/opamSysInteract.ml
```diff
@@ -966,29 +965,90 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
...
     let sys_installed =
       run_query_command "pkg_info" ["-qP"]
       |> package_set_of_pkgpath
     in
     compute_sets sys_installed
 
+let stateless_install ?(env=OpamVariable.Map.empty) () =
+  match family ~env () with
+  | exception Failure _ -> false (* no depexts *)
+  | Nix -> false
```
kit-ty-kate, line 991: Shouldn't the function name be reversed? Nix is a stateless install, as well as no depext, but everything else should be stateful. So either the boolean needs to be flipped or the function needs to be renamed to `stateful_install`

src/state/opamSwitchState.ml
```diff
@@ -198,21 +198,23 @@ let depexts_status_of_packages_raw
...
         else
-          avail, not_found
+          status
       in
       OpamPackage.Map.map (fun set ->
-          { OpamSysPkg.s_available = set %% avail;
-            OpamSysPkg.s_not_found = set %% not_found}
-        ) syspkg_map
+          { OpamSysPkg.s_available = set %% status.s_available;
+            OpamSysPkg.s_not_found = set %% status.s_not_found})
```
kit-ty-kate, line 216:
```suggestion
            OpamSysPkg.s_not_found = set %% status.s_not_found })
```

src/client/opamSolution.ml
```diff
@@ -1247,31 +1266,33 @@ let install_sys_packages ~map_sysmap ~confirm env config sys_packages t =
...
-    let installed = sys_packages -- still_missing in
+    let still_missing = status.s_available ++ status.s_not_found in
+    let installed = sys_packages.ti_new -- still_missing in
     let t =
       map_sysmap (fun sysp -> OpamSysPkg.Set.diff sysp installed) t
     in
     if OpamSysPkg.Set.is_empty still_missing then t else
-    if OpamSysPkg.Set.(not (is_empty notfound) && is_empty needed) then
+    if OpamSysPkg.Set.(not (is_empty status.s_not_found)
+                       && is_empty status.s_available) then
```
kit-ty-kate, line 1287:
```suggestion
    if not (OpamSysPkg.Set.is_empty status.s_not_found)
    && OpamSysPkg.Set.is_empty status.s_available then
```

master_changes.md
```diff
@@ -116,6 +116,12 @@ users)
   * Change probing tool for SUSE-based distributions from `zypper` to `rpm` [#6464 @kit-ty-kate]
   * Disable the detection of available system packages on SUSE-based distributions [#6464 @kit-ty-kate]
 
+[NEW] Support providing external dependencies with Nix
+ * Always pass --no-version-check and --no-write-registry to Cygwin setup [#6046 @dra27]
+ * Use --quiet-mode noinput for the internal Cygwin installation (which is definitely a fully-specified command line) and --quiet-mode unattended for external Cygwin installations (in case the user does need to select something, e.g. a mirror) [#6046 @dra27]
+  * [BUG] Fix apt/debian lookup for installed packages [#6054 @rjbou]
+  * [NEW] Support providing external dependencies with Nix [#5982 @RyanGibb]
```
kit-ty-kate, line 123: todo: these need to be cleaned before merge

src/client/opamSolution.ml
```diff
@@ -1129,58 +1129,76 @@ let syspkgs_to_string spkgs =
...
+              s_available = acc.s_available ++ sys.s_available;
+              s_not_found = acc.s_not_found ++ sys.s_not_found;
+            }
+          | None -> acc)
+        pkg_to_install OpamSysPkg.status_empty
+    in
+    print_depext_msg status;
+    let ti_required =
+      let bypass = t.switch_config.OpamFile.Switch_config.depext_bypass in
+      OpamPackage.Set.fold OpamSysPkg.Set.Op.(fun nv req ->
```
kit-ty-kate, line 1184: This open can be removed if the same open just above was moved a couple lines above in the function
```suggestion
      OpamPackage.Set.fold (fun nv req ->
```

rjbou (approved):

kit-ty-kate:
Thanks a lot!

RyanGibb:
Wow, awesome that this got merged at long last! Thank you both for all the work you put in :-)

PR #5141, opened 2022-05-11
Filter unresolved wariables in constraint

rjbou:
fix #5139

PR #5983, opened 2024-06-01
Don't display undefined variable warnings when evaluating the `?` operator

dra27:
#2921 added the `?` operator to allow filters to have expressions like `?foo & foo = "foo"`, where `?foo` is true if and only if `foo` is a defined variable (or an expression where all the variables have been successfully resolved).

However, it doesn't seem to have been working, which I'd spotted in the ocaml-system updates in https://github.com/ocaml/opam-repository/pull/25861. Unfortunately, the failure means that the repository can't be used with `--strict`.

The fundamental problem is that the warning about undefined variables is displayed when the variables are being resolved, which sadly includes the moment they are being tested with the `?` operator. However, when a filter is partially evaluated, the resulting filter will only contain undefined variables (and the `post` variable), which allows the check added in #5141 to be co-opted to display the unresolved variables _after_ the filter has been reduced, rather than during the reducing.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -51,6 +51,7 @@ users)
 ## Var/Option
 
 ## Update / Upgrade
+  * [BUG] Stop triggering "Undefined filter variable variable" warning for `?variable` [#5983 @dra27]
 
 ## Tree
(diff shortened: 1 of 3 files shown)


kit-ty-kate (approved):

rjbou (approved):
thanks!

kit-ty-kate:
Thanks!

PR #5984, opened 2024-06-02
Harden curl output parsing

dra27:
Issue originally reported on [Discuss](https://discuss.ocaml.org/t/windows-compiler-support-in-opam-2-2-0-beta2/14648/28).

The OP was using a native Windows `curl.exe` from Git-for-Windows git-bash (MSYS2 has curl both for its "Cygwin" main install, but also compiled natively - git-bash installs the mingw64 native-compiled version). This yields an exception during `opam init` while attempting to setup the internal Cygwin installation:

```
Fatal error:
OpamDownload.Download_fail(_, "curl: code  while downloading https://cygwin.com/sha512.sum")
```

Note that there is no code. Running the command manually, everything appears to be OK, but if the output is piped to a file, the issue becomes apparent (the various `\r` effects removed):

```
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   291  100   291    0     0    635      0 --:--:-- --:--:-- --:--:--   638200

```
There is a blank line at the end of the file and the 200 is at the end of the penultimate line. There appears to be an issue with interleaving stderr and stdout when the output is being piped. At this point I could wave my hands and spout Posix, line-buffering and all sorts of explanations, but this PR is a somewhat simpler approach - the status code now has a newline printed before and after it and the parser attempts to convert the first _non-blank_ line from the end of the output to an integer.

I was tempted to add `--no-progress-meter`, but it turns out that's too recent an option. I'm wondering why `--silent` has never been added (same for the others), but perhaps that's because if something goes wrong, the display might be useful. We _could_ add `--silent --show-error` which would add errors only to the log files (`--silent` is there since the dawn of time, and `--show-error` was added in 5.9 on 22-May-1999).

### DIFF
--- a/src/repository/opamDownload.ml
+++ b/src/repository/opamDownload.ml
@@ -22,15 +22,15 @@ let user_agent =
 
 let curl_args = [
   CString "--write-out", None;
-  CString "%%{http_code}\\n", None;
-  CString "--retry", None; CIdent "retry", None;
-  CString "--retry-delay", None; CString "2", None;
+  CString "%%{http_code}\\n", None; (* 6.5 13-Mar-2000 *)
+  CString "--retry", None; CIdent "retry", None; (* 7.12.3 20-Dec-2004 *)
+  CString "--retry-delay", None; CString "2", None; (* 7.12.3 20-Dec-2004 *)
   CString "--compressed",
-  Some (FIdent (OpamFilter.ident_of_string "compress"));
-  CString "--user-agent", None; user_agent, None;
-  CString "-L", None;
-  CString "-o", None; CIdent "out", None;
-  CString "--", None; (* End list of options *)
+  Some (FIdent (OpamFilter.ident_of_string "compress")); (* 7.10 1-Oct-2002 *)
+  CString "--user-agent", None; user_agent, None; (* 4.5.1 12-Jun-1998 *)
+  CString "-L", None; (* 4.9 7-Oct-1998 *)
+  CString "-o", None; CIdent "out", None; (* 2.3 21-Aug-1997 *)
+  CString "--", None; (* End list of options; 5.0 1-Dec-1998 *)
   CIdent "url", None;
 ]
 
--- a/src/repository/opamDownload.ml
+++ b/src/repository/opamDownload.ml
@@ -132,7 +132,9 @@ let download_command ~compress ?checksum ~url ~dst () =
       OpamConsole.error_and_exit `Configuration_error
         "Empty custom download command"
   in
-  OpamSystem.make_command ~allow_stdin:false cmd args @@> tool_return url
+  let stdout = OpamSystem.temp_file ~auto_clean:false "dl" in
+  OpamProcess.Job.finally (fun () -> OpamSystem.remove_file stdout) @@ fun () ->
+  OpamSystem.make_command ~allow_stdin:false ~stdout cmd args @@> tool_return url
 
 let really_download
     ?(quiet=false) ~overwrite ?(compress=false) ?checksum ?(validate=true)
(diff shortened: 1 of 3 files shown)


src/repository/opamDownload.ml
```diff
@@ -108,13 +110,17 @@ let tool_return url ret =
...
-      else Done ()
+      let l = List.rev l in
+      try
+        let code = List.find ((<>) "") l in
+        let num = try int_of_string code with Failure _ -> 999 in
+        if num >= 400 then
+          fail (Some ("curl error code " ^ code),
+                    Printf.sprintf "curl: code %s while downloading %s"
+                      code (OpamUrl.to_string url))
+        else Done ()
```
rjbou, line 121:
```suggestion
        try
          if int_of_string code >= 400 then
            fail (Some ("curl error code " ^ code),
                  Printf.sprintf "curl: code %s while downloading %s"
                    code (OpamUrl.to_string url))
          else Done ()
        with Failure _ ->
          fail (Some "No error code",
                Printf.sprintf "curl: no error code while downloading %s"
                  (OpamUrl.to_string url))
```
Like that we give the 'no code' information instead of an unknown 999 code

dra27: Hah - I debated whether the 999 thing should remain as well, so the fact you also thought it suggests it really ought to go!

I pushed an extra commit, but switching it to have fewer `try` blocks (not necessarily elegantly, though). It also removes the separation of "no output" from "no code parsed", which I think is also fine, as we don't really care about the distinction.

kit-ty-kate (approved):

src/repository/opamDownload.ml
```diff
@@ -132,7 +132,8 @@ let download_command ~compress ?checksum ~url ~dst () =
       OpamConsole.error_and_exit `Configuration_error
         "Empty custom download command"
   in
-  OpamSystem.make_command ~allow_stdin:false cmd args @@> tool_return url
+  let stdout = OpamSystem.temp_file "dl" in
```
rjbou, line 135: The file need to be cleaned. It can be done explicitely (see `OpamGit.diff` or `OpamSysInteract.run_command`, 2 different way to do it), or let to the log cleaner to do it. We want to ensure that everything that is created in user temp directory is cleaned by opam, as far as possible.

kit-ty-kate: done

src/repository/opamDownload.ml
```diff
@@ -132,7 +132,9 @@ let download_command ~compress ?checksum ~url ~dst () =
       OpamConsole.error_and_exit `Configuration_error
         "Empty custom download command"
   in
-  OpamSystem.make_command ~allow_stdin:false cmd args @@> tool_return url
+  let stdout = OpamSystem.temp_file ~auto_clean:false "dl" in
+  OpamProcess.Job.finally (fun () -> OpamSystem.remove_file stdout) @@ fun () ->
```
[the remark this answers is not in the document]
kit-ty-kate: I'm not sure to understand. finally does not drop the exception, it reraise it. See the documentation: https://github.com/ocaml/opam/blob/89a43e2551bdc9991645ba5c655900e59e9d0c5d/src/core/opamProcess.mli#L197

rjbou (approved):
Thanks!

src/repository/opamDownload.ml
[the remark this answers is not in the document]
rjbou: my mistake, i misread something. lgtm!

kit-ty-kate:
Thanks!

PR #5992, opened 2024-06-06
Extend OpamConsole.menu to support 35 options

dra27:
The Git-for-Windows and Cygwin options mean at least one Windows user is capable of blowing the previous limit of 9 menu items quite trivally. 1-9 used as before with a-z used for the remaining items.

### DIFF
--- a/src/core/opamConsole.ml
+++ b/src/core/opamConsole.ml
@@ -964,9 +964,13 @@ let print_table ?cut oc ~sep table =
   List.iter (fun l -> print_line (cleanup_trailing l)) table
 
 let menu ?default ?unsafe_yes ?yes ~no ~options fmt =
-  assert (List.length options < 10);
+  assert (List.length options < 36);
   let options_nums =
-    List.mapi (fun n (ans, _) -> ans, string_of_int (n+1)) options
+    let option_of_index n =
+      (* NOTE: 1..9 (starts at ASCII 49) & a..z (starts at ASCII 97) *)
+      (String.make 1 (char_of_int (if n < 9 then n+49 else n+97)))
+    in
+    List.mapi (fun n (ans, _) -> ans, option_of_index n) options
   in
   let nums_options = List.map (fun (a, n) -> n, a) options_nums in
   let rec prev_option a0 = function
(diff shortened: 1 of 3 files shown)


src/core/opamConsole.ml
```diff
@@ -962,9 +962,12 @@ let print_table ?cut oc ~sep table =
   List.iter (fun l -> print_line (cleanup_trailing l)) table
 
 let menu ?default ?unsafe_yes ?yes ~no ~options fmt =
-  assert (List.length options < 10);
+  assert (List.length options < 36);
   let options_nums =
-    List.mapi (fun n (ans, _) -> ans, string_of_int (n+1)) options
+    let option_of_index n =
+      (String.make 1 (char_of_int (if n < 9 then n+49 else n+88)))
```
kit-ty-kate, line 968:
```suggestion
      (String.make 1 (char_of_int (if n < 9 then n+49 else n+97)))
```

dra27: Doesn't everyone know them from memory 😂

src/core/opamConsole.ml
```diff
@@ -962,9 +962,13 @@ let print_table ?cut oc ~sep table =
   List.iter (fun l -> print_line (cleanup_trailing l)) table
 
 let menu ?default ?unsafe_yes ?yes ~no ~options fmt =
-  assert (List.length options < 10);
+  assert (List.length options < 36);
```
rjbou, line 965: It needs also a change in the function description
```diff
--- a/src/core/opamConsole.mli
+++ b/src/core/opamConsole.mli
@@ -132,7 +132,7 @@ val confirm:
     options are set.
     [no] is the option to choose otherwise, when non interactive, on [escape].
     [default] is the option to choose on an active empty input ("\n").
-    Max 9 options. *)
+    Max 35 options. *)
 val menu:
   ?default:'a -> ?unsafe_yes:'a -> ?yes:'a -> no:'a ->
   options:('a * string) list ->

kit-ty-kate: done

kit-ty-kate (approved):

rjbou (approved):
Thanks!

PR #5994, opened 2024-06-06
API: Allow explicit installation of system packages

dra27:
This is a refactoring in `OpamSolution` so that the depext menu can be displayed either with a list of opam packages (as `OpamSolution.install_depexts` presently does, or slightly lower-level where the set of system packages to install is already calculated by the caller (the new `OpamSolution.install_sys_packages`).

Additionally, the existing `OpamSolution.print_depext_msg` is exposed (ultimately to allow it be used by `OpamClient`).


kit-ty-kate (approved):

rjbou (approved):
Thanks!

kit-ty-kate:
Thanks!

PR #5995, opened 2024-06-06
Always pass `--no-write-registry` to Cygwin setup

dra27:
This flag is already passed when initially installing Cygwin, but any subsequent depext installs which don't have it cause the internal installations to be put in the registry (when possible).

From my testing so far, including it with an existing registered installation does not cause Cygwin to _delete_ the existing registration.


rjbou (approved):
Thanks!

PR #5996, opened 2024-06-06
Don't allow make to become shadowed

dra27:
This should deal with the problem being seen with flexdll on setup-ocaml. The issue there is that the Cygwin bin directory gets pushed past a manual installation of mingw-w64 installed on the GHA runners (in `C:\mingw64`) which includes gcc and which quite rightly confuses OCaml.


rjbou (approved):
Thanks!

PR #5998, opened 2024-06-06
Wind CI forwards to the Windows compilers

dra27:


kit-ty-kate (approved):

rjbou (approved):
Thanks!

rjbou:
Is it needed to wait for #6000 CI to validate it ?

dra27:
Actually, I think it's the other way around - if the failure seen on my fork is real, then #6000 is at fault. Let's merge this and I'll rebase #6000 on to it.

PR #5997, opened 2024-06-06
Rename and slightly repurpose `OpamClient.git_for_windows_check`

dra27:
The diff for this PR is horrific in the GH interface, but with `--ignore-space-change`, it is just:

```diff
diff --git a/src/client/opamClient.ml b/src/client/opamClient.ml
index 879b2a770..e768ca879 100644
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -654,9 +654,7 @@ let is_git_for_windows git =
     end
   | _ -> false

-let git_for_windows_check =
-  if not Sys.win32 then fun ?git_location:_ () -> None else
-  fun ?git_location () ->
+let git_for_windows ?git_location () =
   let header () = OpamConsole.header_msg "Git" in
   let contains_git p =
     OpamSystem.resolve_command ~env:[||] (Filename.concat p "git.exe")
@@ -693,7 +691,8 @@ let git_for_windows_check =
            warning suggesting that the machine be reconfigured to enable them
            in PATH, but also gives the opportunity to use the git-location
            mechanism to select it for opam's internal use. *)
-          let test_for_installation ((gits, gfw_message, abort_action) as acc) (hive, key) =
+        let test_for_installation ((gits, gfw_message, abort_action) as acc)
+                                  (hive, key) =
           let process root =
             let git_location = Filename.concat root "cmd" in
             let git = Filename.concat git_location "git.exe" in
@@ -764,7 +763,9 @@ let git_for_windows_check =
       let options =
         (`Default, "Use default Cygwin Git")
         :: (List.filter_map (fun (git, bash) ->
-              if bash then None else
+            if bash then
+              None
+            else
               let bin = Filename.dirname git in
               Some (`Location bin, "Use found git in "^bin))
             gits)
@@ -851,7 +852,12 @@ let windows_checks ?cygwin_setup ?git_location config =
         (OpamFilename.Dir.to_string gl_cli) ;
       Some (Left gl_cli)
   in
-  let git_location = git_for_windows_check ?git_location () in
+  let git_location =
+    if Sys.win32 then
+      git_for_windows ?git_location ()
+    else
+      None
+  in
   OpamCoreConfig.update ?git_location ();
   let config =
     match git_location with
```

The idea is to make the diff of the large `opam init` PR which is coming a bit easier to handle. The semantic change is instead of having a no-op function is to make it the responsibility of the caller not to call this function on "not Windows".

PR #5991, opened 2024-06-06
Disentangle `OpamProcess.resolve_command` and `OpamSystem.resolve_command`

dra27:
The `opam init` menu overhaul wants to be able to use `resolve_command` from within `OpamStd.Sys` (to be used with some low-level functions), but this isn't possible at present because the default for `OpamSystem.resolve_command`'s `~env` argument is `OpamProcess.default_env`.

However, since #3348 there's already been a slightly unholy mess (added by me...) to allow OpamProcess to use `resolve_command`. This PR disentagles it slightly, by moving the bulk of the logic to `OpamStd.Sys` and leaving the default argument handling in `OpamProcess` and then simply making `OpamSystem.resolve_command` an alias for `OpamProcess.resolve_command`. Not totally sure why I didn't do that in 2018...

Issue #5952, opened 2024-05-13
opam init --cygwin-internal-install is ignored from cygwin terminal


kit-ty-kate:
It should at the very least show a warning.
Detected while testing https://github.com/ocaml/opam/pull/5930

PR #6000, opened 2024-06-06
Small refactoring and fixes to `opam init` on Windows.

dra27:
At present includes #5991, #5994 and #5997. The diff becomes slightly less hideous with #5997 merged. The PR includes the behaviour of #5996 and may need #5998 for CI to pass properly.

This PR considerably extends the ideas from #5963. The primary goal is to extend `opam init` on Windows to recognise as many scenarios for getting either Cygwin or MSYS2 as we at present know to exist, so that users are presented with meaningful choices. In particular, this addresses running `opam init` from with a Cygwin or MSYS2 bash shell (in particular, #5952 is fixed).

I think it's helpful to start with some scenarios with beta2, and then explain the fixes from there:
1. From cmd/pwsh, with nothing other than `opam` installed, we already have an excellent experience which recommends pausing to install Git for Windows (with helpful advice on that added #5963) and which installs Cygwin.
2. If the user has installed Git for Windows using its (non-default) "Use Git and optional Unix tools from the Command Prompt" setting (or run `choco install git /GitAndUnixToolsOnPath`), then beta2 displays rather more Git options than might be nice, but #5963 also makes this a quite elegant experience by default (the `bash.exe` included on PATH is correctly shadowed by the internal Cygwin installation, which is recommended).
3. If the user installs things with [Scoop](https://scoop.sh/) - e.g. `scoop install cygwin msys2 git` - then `opam init` presently works, but neither Git for Windows, MSYS2 or Cygwin are recognised by opam.
4. Similarly, if the user installs with [Chocolatey](https://chocolatey.org/) - e.g. `choco install cyg-get msys2 git` - then `opam init` does detect Git for Windows, but does not detect either the MSYS2 or Cygwin installations.
5. Similarly, if the user installs with [winget](https://learn.microsoft.com/en-us/windows/package-manager/winget/) - e.g. `winget install Cygwin.Cygwin MSYS2.MSYS2 Git.Git` - then we have an experience similar to Chocolatey, although as Cygwin is installed to the default C:\cygwin64, that is recognised by opam.
6. If the user runs `opam init` from Git Bash (not a recommended thing to do), then we get hit by #5984, but if we work around that then the experience is OK. This is a scenario where there are tools on the PATH, and running with `--no-cygwin-setup` is not as OK - `opam var os-distribution` reporting `win32` instead of `msys2`. That said, using Git Bash to run `opam init` is never going to be a good idea (it has no package manager - it's intended for interaction with Git only).
7. If the user has elected to configure their `PATH` to have MSYS2's main bin directory permanently installed, then the default behaviour of `opam init` is pretty good, _unless_ MSYS2's `curl` shadows Windows `curl`.
8. If the user runs `opam init` from within MSYS2's `bash` the experience both with and without `--no-cygwin-setup` is pretty good.
9. If the user has elected to configure their `PATH` to have Cygwin's main bin directory permanently installed, then the default behaviour is affected by #5952.
10. Running `opam init` from within Cygwin's `bash` is similarly affected by #5952.

In all of these scenarios where the user has preinstalled either MSYS2 or Cygwin, we do have the issue that `opam init --no-cygwin-setup` (i.e. just using what's found) will typically fail because `patch`, `unzip` and so forth are not installed by default.

The changes here seek to make all of those scenarios as awesome as possible.

At a very high-level the changes are:
- MSYS2 and Cygwin are now treated equally where `sys-pkg-manager-cmd` is concerned (if both have been defined, Cygwin takes priority). In particular that means that MSYS2 gets added to PATH in exactly the same way as Cygwin.
- `os-distribution` is no longer explicitly set by `opam init`. Instead, `os-distribution` is always determined by looking at `cygpath.exe`. The internal or `--cygwin-location` treatment merely affects PATH resolution (i.e. the internal cygbin addition) - once cygpath.exe is resolved, it becomes the source of truth. Thus, from an MSYS2 shell, `opam var os-distribution` is now `msys2` _by default_.
- For now, I've removed the prompting for Cygwin setup. The idea here is that we treat Cygwin setup as being equal between both an internal and external installation of Cygwin. In particular, this fixes a bug in depext treatment when initialising with `--cygwin-location` as Cygwin setup was never downloaded.
- The final, somewhat horrific, refactor in OpamClient attempts to deal with the fact that the Git menu and the Unix tools decisions are not independent. The Unix tools are now determined first, and that feeds into the decision about what may need to be done for Git.
- Various tricks are now exploited to find Cygwin and MSYS2 installations:
  - Cygwin's two registry keys are inspected
  - The default locations of C:\cygwin64 and C:\msys64 are used
  - A path search for msys2.exe provides an heuristic for seeeking an MSYS2 installation
  - Special handling for Scoop is added, both to identify its Git for Windows and to identify its MSYS2 installation
- A key feature of all of these changes is that we only ever cygcheck and cygpath while inspecting things. In particular, we have to be very careful about running `bash` since we have no idea what it may do.
- Chocolatey and winget both correctly initialise their MSYS2 installations, but unfortunately Scoop leaves this as an instruction to the user. Again, I've added some special handling to detect this - but this is treated just as running a depext command, so opam asks about it first.
- The depext system is now integrated into `opam init` so opam will offer to add the required packages to both Cygwin and MSYS2 for existing installations, rather than simply erroring.

Put together, the changes mean that the user is presented with meaningful choices in all 10 of those scenarios and, in particular, they will see _all_ of the packages they have already installed. If they agree with everything opam suggests, they should then always end up with initialised opam with a compiler 🥳

On my system, when added to #5992, `opam init` run from Cygwin's bash now offers:

```
DRA@Tau /cygdrive/c/Devel/opam
$ ./opam init

<><> Unix support infrastructure ><><><><><><><><><><><><><><><><><><><><><>  🐫

opam and the OCaml ecosystem in general require various Unix tools in order to operate correctly. At present, this requires the installation of Cygwin to provide these tools.

How should opam obtain Unix tools?
> 1. Use tools found in PATH (Cygwin installation at C:\cygwin64)
  2. Automatically create an internal Cygwin installation that will be managed by opam (recommended)
  3. Use Cygwin installation found in C:\cygwin64
  4. Use Cygwin installation found in C:\OCaml64
  5. Use Cygwin installation found in C:\cygwin64-2
  6. Use Cygwin installation found in C:\Users\DRA\AppData\Local\opam\.cygwin\root
  7. Use Cygwin installation found in C:\Devel\Roots\beta2-testing\.cygwin\root
  8. Use Cygwin installation found in C:\Cygwin64-beta2-testing
  9. Use Cygwin installation found in C:\cygwin64-throw
  a. Use Cygwin installation found in C:\Devel\Roots\opam-testing\.cygwin\root
  b. Use Cygwin installation found in C:\Devel\Roots\final-test\.cygwin\root
  c. Use Cygwin installation found in C:\Devel\Roots\env-testing\.cygwin\root
  d. Use Cygwin installation found in C:\cygwin64-throw2
  e. Use Cygwin installation found in C:\Devel\Roots\msys2-native\.cygwin\root
  f. Use Cygwin installation found in C:\Devel\Roots\os-dist-cygwin\.cygwin\root
  g. Use MSYS2 installation found in C:\msys64
  h. Use another existing Cygwin/MSYS2 installation
  i. Abort initialisation

[1/2/3/4/5/6/7/8/9/a/b/c/d/e/f/g/h/i]
```

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -1003,9 +1003,8 @@ let windows_checks ?cygwin_setup ?git_location config =
           in
           (* Check for default cygwin installation path *)
           let default =
-            match OpamSysInteract.Cygwin.(check_install
-                                            ~variant:true default_cygroot) with
-            | Ok cygcheck ->
+            match OpamSysInteract.Cygwin.(analyse_install default_cygroot) with
+            | Ok (_, cygcheck) ->
               let prompt_cygroot () =
                 let options = [
                   `manual,
@@ -1036,10 +1035,10 @@ let windows_checks ?cygwin_setup ?git_location config =
               | None -> None
               | Some entry ->
                 let cygcheck =
-                  OpamSysInteract.Cygwin.check_install ~variant:true entry
+                  OpamSysInteract.Cygwin.analyse_install entry
                 in
                 match cygcheck with
-                | Ok cygcheck -> Some cygcheck
+                | Ok (_, cygcheck) -> Some cygcheck
                 | Error msg -> OpamConsole.error "%s" msg; None
           in
           (* And finally ask for setup.exe *)
@@ -1113,9 +1112,8 @@ let windows_checks ?cygwin_setup ?git_location config =
                    | `default_location -> OpamSysInteract.Cygwin.default_cygroot
                    | `location dir -> OpamFilename.Dir.to_string dir
                  in
-                 (match OpamSysInteract.Cygwin.check_install ~variant:true
-                          cygroot with
-                 | Ok cygcheck -> cygcheck
+                 (match OpamSysInteract.Cygwin.analyse_install cygroot with
+                 | Ok (_, cygcheck) -> cygcheck
                  | Error msg ->
                    OpamConsole.error_and_exit `Not_found
                      "Error while checking %sCygwin install (%s): %s"
@@ -1131,9 +1129,9 @@ let windows_checks ?cygwin_setup ?git_location config =
           (* We check that current install is good *)
           (match OpamSysInteract.Cygwin.cygroot_opt config with
            | Some cygroot ->
-             (match OpamSysInteract.Cygwin.check_install ~variant:true
+             (match OpamSysInteract.Cygwin.analyse_install
                       (OpamFilename.Dir.to_string cygroot) with
-             | Ok cygcheck ->
+             | Ok (_, cygcheck) ->
                OpamSysInteract.Cygwin.check_setup None;
                success cygcheck
              | Error err -> OpamConsole.error "%s" err; get_cygwin None)
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -955,12 +955,12 @@ let windows_checks ?cygwin_setup ?git_location config =
     config
   in
   let install_cygwin_tools packages =
+    let default_packages = OpamInitDefaults.required_packages_for_cygwin in
     let default_packages =
-      match OpamSystem.resolve_command "git" with
-      | None -> OpamInitDefaults.required_packages_for_cygwin
-      | Some _ ->
-        List.filter (fun c -> not OpamSysPkg.(equal (of_string "git") c))
-          OpamInitDefaults.required_packages_for_cygwin
+      if OpamSystem.resolve_command "git" = None then
+        OpamSysPkg.of_string "git" :: default_packages
+      else
+        default_packages
     in
     (* packages comes last so that the user can override any potential version
        constraints in default_packages (although, with the current version of
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -1096,7 +1096,10 @@ let windows_checks ?cygwin_setup ?git_location config =
              because there are commands opam requires which are only provided
              using it (patch, etc.). MSYS2 avoids this by requiring
              os-distribution to be set. *)
-          let cygcheck = OpamSysInteract.Cygwin.cygcheck_opt config in
+          let cygcheck =
+            OpamStd.String.Map.find_opt "cygwin"
+                    (OpamFile.Config.sys_pkg_manager_cmd config)
+          in
           (match cygwin_setup with
            | None -> get_cygwin cygcheck
            | Some setup  ->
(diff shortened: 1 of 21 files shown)


src/state/opamSysPoll.ml
```diff
@@ -114,14 +114,15 @@ let poll_os_distribution () =
...
-      |> (fun cygbin -> OpamStd.Sys.is_cygwin_cygcheck ~cygbin)
+    let kind =
+      OpamStd.Sys.get_windows_executable_variant
+        ?search_in_first:(OpamCoreConfig.(!r.cygbin)) "cygpath.exe" 
     in
-    if cygwin then Some "cygwin" else os
+    begin match kind with
+    | `Msys2 -> Some "msys2"
+    | `Cygwin -> Some "cygwin"
+    | _ -> os
```
kit-ty-kate, line 124:
```suggestion
    | `Native | `Tainted _ -> os
```

src/client/opamClient.ml
```diff
@@ -652,497 +652,878 @@ let is_git_for_windows git =
...
+        None
+      | None, _ ->
+        OpamConsole.error "No Git executable found in %s." git_location;
+        None
+  in
+  let options =
+    (List.filter_map (fun (git, bash) ->
+        if bash then None else
+        let bin = Filename.dirname git in
+        Some (`Location bin, "Use found git in "^bin))
```
kit-ty-kate, line 804:
```suggestion
        if bash then
          None
        else
          let bin = Filename.dirname git in
          Some (`Location bin, "Use found git in "^bin))
```

src/client/opamClient.ml
```diff
@@ -652,497 +652,878 @@ let is_git_for_windows git =
...
+         include Cygwin/MSYS2. *)
+      let env = OpamStubs.get_initial_environment () in
+      let cygcheck = 
+        OpamSystem.resolve_command ~env:(Array.of_list env) "cygcheck.exe"
+        |> Option.map OpamSysInteract.Cygwin.analyse_install 
+      in
+      begin match cygcheck with
+      | Some (Ok (kind2, root2)) ->
+        let root2 = OpamFilename.Dir.to_string root2 in
+        if kind = kind2 && String.equal root root2 then
```
kit-ty-kate, line 1082:
```suggestion
        if (kind : [`Cygwin | `Msys2]) = kind2 && String.equal root root2 then
```

kit-ty-kate (approved):

rjbou (approved):
What a small refactoring!! :D it's working smoothly, thanks!

kit-ty-kate:
Thanks!

PR #6005, opened 2024-06-09
Fix bug in recursion case in `OpamSystem.mk_temp_dir`

dra27:
The `~prefix` argument wasn't passed when recursing. Very niche failure mode, but having spotted it...


rjbou (approved):
well spotted, thanks!

PR #6015, opened 2024-06-11
Various fixes

kit-ty-kate:
Noticed when debugging https://github.com/kit-ty-kate/opam-repository/pull/26

The first commit is my fault from #6000. I noticed it when running opam locally on Windows and saw the choices `1/2/3/4/5/6/7/8/9/j` instead of the expected `1/2/3/4/5/6/7/8/9/a`.

The second commit helped to locate the issue i was having as `[ERROR] Could not update repository "default": rsync failed` was not very helpful. With this change I now get `Error:  Could not update repository "default": Not available: Directory D:/a/opam-repository/opam-repository/opam-repository/ does not exist`


kit-ty-kate:
Split into #6026 and #6027

PR #6018, opened 2024-06-12
ocp-indent: disable unindented else block

kit-ty-kate:
```
if ... then a else
b
```
A good portion of the code are indented to use this code pattern and this makes reading that code really hard to me. Forcing the indentation of `b` in the example above makes it much more readable, especially for large functions.
New indented code would look like:
```
if ... then a else
  b
```

### DIFF
--- a/.ocp-indent
+++ b/.ocp-indent
@@ -1,2 +1,1 @@
 normal
-strict_else=auto


rjbou (approved):
Thanks!
```
A new
  era

rjbou:
Does it need to be backported in 2.2 ?

kit-ty-kate:
> Does it need to be backported in 2.2 ?

I don't think so since most of the backports are cherry-picks from master i don't expect the .ocp-indent from that branch to be actually used by us very much

PR #6020, opened 2024-06-14
Update URL for ocamlgraph source

mtelvers:
Using `curl`, ocamlgraph.lri.fr is redirecting to GH which returns HTTP 200 and a page with `Not found` whereas `wget` returns 404.


dra27:
Borrowed a commit from https://github.com/ocaml/opam/pull/5384 to try to get CI to pass!

dra27:
CI will fail because it would need to be updated, but discussed today in the dev meeting - we'll leave updating CI until a 2.0.11 release is required.

mtelvers:
Adding a link to the issue https://github.com/ocurrent/docker-base-images/issues/283

PR #6024, opened 2024-06-17
Fix a wrong use of OpamFilename.of_string

kit-ty-kate:
Related to https://github.com/ocaml/opam/pull/5822

This line of code defines a dummy default value to pass to the two pretty-printers below.
However the `OpamFilename.of_string` function is going to try to expend this path as much as possible, which it should not do.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -90,6 +90,7 @@ users)
 ## Shell
 
 ## Internal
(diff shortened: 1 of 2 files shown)


dra27 (approved):
LGTM!

kit-ty-kate:
Benchmark failed because of https://github.com/ocaml/ocaml/issues/13237

PR #6026, opened 2024-06-17
Fix OpamConsole.menu > 9 options

kit-ty-kate:
Extracted from #6015


dra27 (approved):

kit-ty-kate:
Benchmark was failing because of https://github.com/ocaml/ocaml/issues/13237

PR #6029, opened 2024-06-18
Nullify positions of the extensions fields in OpamFile.OPAM.effective_part

### LINKED ISSUE #6028: opam upgrade keeps wanting to recompile ocaml
Hi,
I'm using `2.2.0~beta3` on linux.
I'm encountering an issue where `opam upgrade` keeps wanting to recompile `ocaml` and `ocaml-base-compiler`.

This looks like:

```
$ opam update
$ opam upgrade
[...]
  ↻ ocaml                       4.14.1          [upstream or system changes]
  ↻ ocaml-base-compiler         4.14.1          [upstream or system changes]
[...]
Proceed with ↻ 141 recompilations? [y/n] y
[...]
$ opam upgrade
[...]
  ↻ ocaml                       4.14.1          [upstream or system changes]
  ↻ ocaml-base-compiler         4.14.1          [upstream or system changes]
[...]
Proceed with ↻ 141 recompilations? [y/n] n
```

`opam remote` has just the default http remote. Here's part of the output of `opam config report`:

    # repositories         1 (http), 10 (version-controlled) (default repo at 5d6f1cf0)

Let me know what extra info I can provide.

kit-ty-kate:
Fixes #6028 

Prior to https://github.com/ocaml/opam/pull/5636, opam did not traverse the extension fields when testing equality.
Prior to https://github.com/ocaml/opam-repository/pull/25861, very few packages actually had the `x-env-path-rewrite` field so the problem did not occur.
Users with a fresh install cache saved after an install will also not immediately notice this issue (see https://github.com/ocaml/opam/issues/6031)

The issue was that the extension field carries all the positions of the original file and opam files of installed packages are saved in their normalised form which changes the positions of the fields, so when `OpamFile.OPAM.effectively_equal` is called it would see a difference in the extension field.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -99,6 +99,7 @@ users)
 
 ## Reftests
 ### Tests
+  * add a complete test to make sure effectively_equal does not take the location of the fields into account [#6029 @kit-ty-kate]
 
 ### Engine
 
--- a/master_changes.md
+++ b/master_changes.md
@@ -40,6 +40,7 @@ users)
 ## Var/Option
 
 ## Update / Upgrade
+  * Fix `opam upgrade` wanting to recompile opam files containing the `x-env-path-rewrite` field [#6029 @kit-ty-kate - fix #6028]
 
 ## Tree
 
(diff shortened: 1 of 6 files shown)


dra27 (approved):
Ta!

rjbou:
Failing CI is from benchmark test broken.

PR #6033, opened 2024-06-18
Fix `opam init -yn` asking questions on Windows

### LINKED ISSUE #6013: opam init -yn still asks for cygwin and git related questions on Windows if ran from a terminal
I don't think this should be the case. If ran from an automated tool without an open tty though, no question will be asked as expected.

dra27:
Fixes #6013. Mostly this was simply that `~yes` parameters were not provided when they should have been. The slightly more subtle fix is if MSYS2 is initialised which is requires `--confirm-level unsafe-yes` but lacked a `~yes`. In this case, the action for `-y` is simply is to ignore the change.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -19,6 +19,7 @@ users)
 ## Plugins
 
 ## Init
+  * Provide defaults so `opam init -y` no longer asks questions [#6033 @dra27 fix #6013]
 
 ## Config report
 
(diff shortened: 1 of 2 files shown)


rjbou (approved):
Thanks!

rjbou:
Red CI is from benchmark test broken.

PR #6034, opened 2024-06-18
Suppress Windows-specific menus with `opam init -ya`

dra27:
This at the moment sits on top of #6033. #6033 is a bug-fix and needs to go in 2.2.0 - the extra commit here follows a discussion last week with @kayceesrk regarding making `opam init -ya` to be potentially less confusing for students. In this mode, `opam init` looks more like:

```
C:\Users\David Allsopp>\\tau\devel\opam\opam init -ya
[NOTE] opam is storing files in C:\opamroot

<><> Windows Developer Mode <><><><><><><><><><><><><><><><><><><><><><><><>  🐫
opam does not require Developer Mode to be enabled on Windows, but it is
recommended, in particular because it enables support for symlinks without
requiring opam to be run elevated (which we do not recommend doing).

More information on enabling Developer Mode may be obtained from
https://learn.microsoft.com/en-gb/windows/apps/get-started/enable-your-device-for-development
[NOTE] opam will maintain an internal Cygwin installation
[NOTE] opam will add Git to Cygwin
```

Note (!) that each NOTE would otherwise be an entire menu. The change is quite noisy, and I'm not totally sure that we should do this for 2.2.0, just given the late hour.

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -691,7 +691,7 @@ let string_of_kind = function
   | `Msys2 -> "MSYS2"
   | `Cygwin -> "Cygwin"
 
-let git_for_windows kind mechanism cygwin_is_tweakable =
+let git_for_windows kind mechanism ~interactive ~cygwin_is_tweakable =
   let resolve_git_in p =
     OpamSystem.resolve_command ~env:[||] (Filename.concat p "git.exe")
   in
@@ -890,7 +890,7 @@ let git_for_windows kind mechanism cygwin_is_tweakable =
       | None -> false
       | Some git -> is_git_for_windows git
     in
-    if not git_found && OpamStd.Sys.tty_out then
+    if not git_found && interactive && OpamStd.Sys.tty_out then
       (OpamConsole.header_msg "Git";
        OpamConsole.msg
          "Cygwin Git is functional but can have credentials issues for private \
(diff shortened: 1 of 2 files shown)


src/client/opamClient.ml
```diff
@@ -1821,7 +1848,7 @@ let init
         let config, mechanism, system_packages, msys2_check_root =
           if Sys.win32 then
             determine_windows_configuration ?cygwin_setup ?git_location
-                                            ~bypass_checks config
+                                            ~bypass_checks ~interactive config
```
rjbou, line 1851: It's missing one in reinit

rjbou (approved):
I like the idea, it give the information on what opam is doing even if automatic.

src/client/opamClient.ml
```diff
@@ -1148,20 +1152,31 @@ let rec cygwin_menu ~bypass_checks header =
...
-  OpamConsole.msg
-    "\n\
-     opam and the OCaml ecosystem in general require various Unix tools in \
-     order to operate correctly. At present, this requires the installation \
-     of Cygwin to provide these tools.\n\n";
-  match OpamConsole.menu "How should opam obtain Unix tools?"
-          ~default ~no:default ~options with
+  let result =
+    if interactive then begin
+      Lazy.force header;
```
rjbou, line 1157: Maybe keep the header outside the interactive ? Like that it give more information about the future notes.

dra27: Yes - nice idea!

src/client/opamClient.ml
dra27: It's there, isn't it? (L1618?)

rjbou: It is in fact, my mistake

rjbou:
Tested the PR, working as it should be :+1:

kit-ty-kate:
Thanks!

PR #6045, opened 2024-06-21
Bump version to 2.3.0~alpha~dev

rjbou:
* [x] update tests
* [x] Bump cli version/add cli 2.3


dra27:
Doesn't master now become 2.3.0~dev?

tests/reftests/empty-conflicts-002.test
```diff
@@ -1,6 +1,8 @@
 11ea1cb
 ### OPAMYES=1 OPAMSTRICT=0
 ### OPAMVAR_arch=x86_64 OPAMVAR_os=linux OPAMVAR_os_family=arch OPAMVAR_os_distribution=archarm
+### # We need to set this version to keep the test as is
+### OPAMVAR_opam_version=2.2.0
```
rjbou, line 5: I don't know if we should update the test (and possibly alter it?) or freeze it like that...

dra27: I agree it looks a bit weird, but I think it's OK - the only thing which might be worth double-checking is whether the bounds actually matter i.e. if something actually needs fixing in opam-repository? But this could be tracked in a follow-up issue/note, rather than blocking the version bump.

dra27 (approved):
LGTM, thanks!

rjbou:
Discussed in meeting: seems that setting the `opam-version` variable is the good fix, to keep tests as is.

kit-ty-kate:
Thanks!

PR #6046, opened 2024-06-21
Adjust common flags passed to Cygwin setup

### LINKED ISSUE #6041: Cywing setup hangs
Trying to use opam 2.2 from ocaml-setup
Setup-ocaml.v3alpha uses a separate cygwin root so that it can cache it separately from opam.
After restoring a cygwin cache and the opam one. cygwin-setup (called by opam) hangs without any logs.

See https://github.com/ocaml/Zarith/actions/runs/9596506376/job/26463523796

dra27:
Fixes #6041 - an additional option might be to explicitly abort when run non-interactively as well, but it's getting a bit niche because if you're non-interactive and expecting depext to work with an external Cygwin then clearly you must have specified `--confirm-level unsafe-yes`.

**Backport into 2.2 in #6057**

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -59,6 +59,7 @@ users)
 ## Opamfile
 
 ## External dependencies
+ * Always pass --no-version-check and --no-write-registry to Cygwin setup [#6046 @dra27]
 
 ## Format upgrade
 
--- a/master_changes.md
+++ b/master_changes.md
@@ -60,6 +60,7 @@ users)
 
 ## External dependencies
  * Always pass --no-version-check and --no-write-registry to Cygwin setup [#6046 @dra27]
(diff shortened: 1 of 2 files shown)


dra27:
It turns out that this can be solved in setup-ocaml by ensuring that Cygwin's setup registry key is restored as part of the action. I'm therefore not certain that we _have_ to do this for 2.2.0

dra27:
This is no longer required for 2.2.0 - the adjustments in it are probably worth including in 2.2.1, however.

dra27:
Oh dear, I appear to need to U-turn! There is an issue with this which probably is worth addressing for 2.2.0 - if the user happens to end up even interactively with an external Cygwin which needs input, then we've turned the dialogues off, which is bad.

I've added a commit to use `--quiet-mode unattended` for external Cygwin, keeping `--quiet-mode noinput` for the internal installation only, where we know that all the inputs have been provided.

rjbou (approved):
Thanks!

PR #6047, opened 2024-06-21
Restore signature of OpamStateConfig.opamroot

dra27:
#5457 changed the signature of `OpamStateConfig.opamroot` between beta2 and beta3 which has broken the opam-dune-lint plugin along with a few other packages. This seems a shame, given that it's quite trivial and indeed natural to have two functions and simply have the provenance-returning version have a new name.

Given that ~~it looks like we'll need an rc2 to deal with the Cygwin hang~~it's a demonstrably safe internal API restoration, I'm hoping we can slip this API change in ~~at the same time~~to the final release. If we do, https://github.com/ocaml/opam-repository/pull/26066 can be reverted.

**Backported into 2.2 in #6058**

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -114,6 +114,7 @@ users)
 ## opam-repository
 
 ## opam-state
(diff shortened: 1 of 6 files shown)


rjbou (approved):
Thanks!

PR #6054, opened 2024-06-25
depext: fix Debian/apt lookup for installed packages

rjbou:
dpkg-query output have the information in the first column, but only the second character needed to be looked up.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -66,6 +66,7 @@ users)
 ## External dependencies
  * Always pass --no-version-check and --no-write-registry to Cygwin setup [#6046 @dra27]
  * Use --quiet-mode noinput for the internal Cygwin installation (which is definitely a fully-specified command line) and --quiet-mode unattended for external Cygwin installations (in case the user does need to select something, e.g. a mirror) [#6046 @dra27]
+  * [BUG] Fix apt/debian lookup for installed packages [#6054 @rjbou]
 
 ## Format upgrade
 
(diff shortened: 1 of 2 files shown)


master_changes.md
```diff
@@ -59,6 +59,7 @@ users)
 ## Opamfile
 
 ## External dependencies
+  * [BUG] Fix apt/debian lookup for installed packages [#XXX @rjbou]
```
dra27, line 62:
```suggestion
  * [BUG] Fix apt/debian lookup for installed packages [#6054 @rjbou]
```

src/state/opamSysInteract.ml
```diff
@@ -774,12 +774,13 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
     let get_installed str_pkgs =
       (* ouput:
          >ii  uim-gtk3                 1:1.8.8-6.1  amd64    Universal ...
-         >ii  uim-gtk3-immodule:amd64  1:1.8.8-6.1  amd64    Universal ...
+         >ri  uim-gtk3-immodule:amd64  1:1.8.8-6.1  amd64    Universal ...
```
dra27, line 777: Worth expanding this a bit (looking at the `dpkg-query` man page) that the first code can be `[uihrp]` but that we only care about the second character?

dra27 (approved):
I'd possibly expand the comment to be more explicit that we're ignoring the first character (and why), but it was very easy to answer that from the dpkg-query man page!

src/state/opamSysInteract.ml
dra27: New comment added below looks perfect, thanks!

kit-ty-kate:
Queued on https://github.com/ocaml/opam/pull/6063 for CI to succeed

PR #6063, opened 2024-07-02
GHA: Remove centos when testing depexts as the docker images no longer work

kit-ty-kate:
Should fix the Github Action issues we've seen recently

### DIFF
--- a/.github/scripts/depexts/generate-actions.sh
+++ b/.github/scripts/depexts/generate-actions.sh
@@ -2,7 +2,7 @@
 
 set -eu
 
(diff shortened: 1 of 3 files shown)


rjbou:
Main centos image is [deprecated](https://hub.docker.com/_/centos/), there is no available image that can replace it?

kit-ty-kate:
to my knowledge, there is not

kit-ty-kate:
Superseded by https://github.com/ocaml/opam/pull/6079

Issue #6064, opened 2024-07-02
Opam admin cache doesn't create a symlink for sha256 to sha512


psafont:
I've run into a couple issues when building [xapi's opam repository](https://github.com/xapi-project/xs-opam) as an RPM in an environment without internet access:
1. When running opam admin cache in opam 2.2.0, the sha256 file for [crc.2.2.0](https://github.com/xapi-project/xs-opam/blob/d2c0aa683efe03e90a801a0d5628820644e55a54/packages/xs/crc.2.2.0/opam) is not created, only the sha512 one. The former should a symlink to the latter.
2. When installing the crc.2.2.0 package, opam 2.1.4 finds that there is no file for its sha256 hash and tries to use internet do fetch it. First from opam.ocaml.org, and after with github.com, after which it fails. I believe it should try to use the sha512 hash to fetch the file from the cache before using other sources.

I'm updating the opam version for issue 2. to see if this behaviour has changed. In any case I've been able to work around issue 1 by manually creating the symlink in the sha256 folder to the sha512 file. Another workaround is removing the sha512 hash from the checksum list: it makes opam create the file created correctly in the sha256 folder.

I don't quite understand why this is the only package in the whole repository with this behaviour, but it looks like a bug.

kit-ty-kate:
Has issue 2 has been fixed in opam 2.1.5 by https://github.com/ocaml/opam/pull/5538 ?

psafont:
I'll test building an opam 2.1.6 rpm to be used in the build

psafont:
I've made a github runner to package a repo tarball, it uses ubuntu's opam package: 2.1.2. The resulting tarball contains the file in the sha256 folder, but _not_ a file in the sha512 folder

kit-ty-kate:
Actually I'm getting the same problem with opam 2.2.0 while using https://github.com/kit-ty-kate/opam-health-check-ng, which means a large number of packages fail and I'm unable to check them.

I'm looking into it.

rjbou:
Do you have a previously populated cache or it is an new one?
Opam creates the symlinks, taking the strongest checksum as real archive and creates links to that one for others checksums.
In the case where the cache is already present, especially if the strongest checksum archive file is already present it can alter this mechanism.

psafont:
> Do you have a previously populated cache or it is an new one?

I can reproduce this locally starting from an empty cache. It's strange this only seems to happen for a single package in the repository.

kit-ty-kate:
[...]

PR #6068, opened 2024-07-03
Add every listed checksums of a given archive to the cache when fetching from it

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/6064

Backported in 2.2 by https://github.com/ocaml/opam/pull/6069

### DIFF
--- a/src/repository/opamRepository.ml
+++ b/src/repository/opamRepository.ml
@@ -44,6 +44,11 @@ let cache_file cache_dir checksum =
   in
   aux cache_dir (OpamHash.to_path checksum)
 
+let link_files ~target f l =
+  List.iter (fun x ->
+      OpamFilename.link ~relative:true ~target ~link:(f x))
+    l
+
 let fetch_from_cache =
   let currently_downloading = ref [] in
   let rec no_concurrent_dls key f x =
@@ -93,21 +98,27 @@ let fetch_from_cache =
       failwith "Version control not allowed as cache URL"
   in
   try
-    let hit_file =
-      OpamStd.List.find_map (fun ck ->
+    match
+      List.fold_left (fun (hit, misses) ck ->
           let f = cache_file cache_dir ck in
-          if OpamFilename.exists f then Some f else None)
-        checksums
-    in
-    if List.for_all
-        (fun ck -> OpamHash.check_file (OpamFilename.to_string hit_file) ck)
-        checksums
-    then Done (Up_to_date (hit_file, OpamUrl.empty))
-    else mismatch hit_file
+          if OpamFilename.exists f
+          then (if hit = None then (Some f, misses) else (hit, misses))
+          else (hit, f :: misses))
+        (None, []) checksums
+    with
+    | None, _ -> raise Not_found
+    | Some hit_file, miss_files ->
+      if List.for_all
+          (fun ck -> OpamHash.check_file (OpamFilename.to_string hit_file) ck)
+          checksums
+      then begin
+        link_files ~target:hit_file Fun.id miss_files;
+        Done (Up_to_date (hit_file, OpamUrl.empty))
+      end else
+        mismatch hit_file
   with Not_found -> match checksums with
     | [] -> let m = "cache miss" in Done (Not_available (Some m, m))
-    | checksum::_ ->
-      (* Try all cache urls in order, but only the first checksum *)
+    | checksum::other_checksums ->
       let local_file = cache_file cache_dir checksum in
       let tmpfile = OpamFilename.add_extension local_file "tmp" in
       let rec try_cache_dl = function
@@ -125,6 +136,7 @@ let fetch_from_cache =
               checksums
           then
             (OpamFilename.move ~src:tmpfile ~dst:local_file;
+             link_files ~target:local_file (cache_file cache_dir) other_checksums;
              Done (Result (local_file, root_cache_url)))
           else mismatch tmpfile
       in
@@ -151,14 +163,9 @@ let validate_and_add_to_cache label url cache_dir file checksums =
     (let checksums = OpamHash.sort checksums in
      match cache_dir, checksums with
      | Some dir, best_chks :: others_chks ->
-       OpamFilename.copy ~src:file ~dst:(cache_file dir best_chks);
-       List.iter (fun checksum ->
-           let target = cache_file dir best_chks in
-           let link = cache_file dir checksum in
-           try
-             OpamFilename.link ~relative:true ~target ~link
-           with Sys_error _ -> ())
-         others_chks;
+       let target = cache_file dir best_chks in
+       OpamFilename.copy ~src:file ~dst:target;
+       link_files ~target (cache_file dir) others_chks;
      | _ -> ());
     true
 
--- a/src/repository/opamRepository.mli
+++ b/src/repository/opamRepository.mli
@@ -58,7 +58,8 @@ val pull_file:
   unit download OpamProcess.job
 
 (** Same as [pull_file], but without a destination file: just ensures the file
-    is present in the cache. *)
+    is present in the cache and that every listed checksums are correctly linked
+    to the archive, otherwise it adds the missing links. *)
 val pull_file_to_cache:
   string -> cache_dir:dirname -> ?cache_urls:url list ->
   OpamHash.t list -> url list -> string download OpamProcess.job
--- a/tests/reftests/admin.test
+++ b/tests/reftests/admin.test
@@ -651,6 +651,10 @@ opam-version: "2.0"
 ### opam-cat packages/with-xf/with-xf.2/opam | "${XF_MD5}" -> "good-md5" | "${XF_SHA256}" -> "good-sha"
 extra-files: ["missing-hash" "md5=good-md5"]
 opam-version: "2.0"
+### ::::::::::::::::::::::::::::::::::::::::::::::::::::::
+### :: Cache is tested in admin-cache.test ::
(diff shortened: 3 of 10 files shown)


tests/reftests/admin.test
```diff
@@ -600,9 +600,52 @@ opam-version: "2.0"
...
+### sh -c "find cache | wc -l"
+24
+### : check strong and weak hashes presence :
+### # we can't check linking status as on Windows it is not linked but copied
+### test -f cache/md5/$PRE_OC_MD5/$OC_MD5
+### test -f cache/sha512/$PRE_OC_SHA512/$OC_SHA512
+### test -f cache/md5/$PRE_DL_MD5/$DL_MD5
+### test -f cache/sha256/$PRE_DL_SHA256/$DL_SHA256
+### test -f cache/sha512/$PRE_DL_SHA512/$DL_SHA512
+### : remove strongest hashes and recompute :
```
kit-ty-kate, line 634:
```suggestion
### : remove weakest hashes and recompute :
```

src/repository/opamRepository.ml
```diff
@@ -44,6 +44,13 @@ let cache_file cache_dir checksum =
   in
   aux cache_dir (OpamHash.to_path checksum)
 
+let link_files ~target f l =
+  List.iter (fun x ->
+      try
+        OpamFilename.link ~relative:true ~target ~link:(f x)
+      with Sys_error _ -> ()) (* Can happen on Windows *)
```
kit-ty-kate, line 51:
```suggestion
      OpamFilename.link ~relative:true ~target ~link:(f x)
```

src/repository/opamRepository.ml
```diff
@@ -44,6 +44,11 @@ let cache_file cache_dir checksum =
   in
   aux cache_dir (OpamHash.to_path checksum)
 
+let link_files ~target f l =
+  List.iter (fun x ->
+      OpamFilename.link ~relative:true ~target ~link:(f x)
```
kit-ty-kate, line 49:
```suggestion
      OpamFilename.link ~relative:true ~target ~link:(f x))
```

rjbou:
I've updated 
* the test (in admin-cache.*.test now) for linux/mac and windows: they show more easily the different behaviour, the links, etc.
  * add an md5 corrupted/collision test too
* the changelog
* commit messages to be more accurate with current PR content.

tests/reftests/admin-cache.unix.test
```diff
@@ -0,0 +1,406 @@
...
+pkg=$1
+kind=$2
+hash=$3
+
+pre=`echo $hash | cut -c 1-2`
+arch=cache/$kind/$pre/$hash
+msg="$pkg : $kind :"
+
+if [ -L $arch ]; then
+  link=`readlink "$arch"`
```
rjbou, line 87: I'm not sure if readlink is portable enough. I need here to retrieve the target of the link, not the fully resolved one

tests/reftests/archive.test
```diff
@@ -1028,15 +1028,19 @@ The following actions will be performed:
...
+[ERROR] Conflicting file hashes, or broken or compromised cache!
+          - sha256=99a0bdda5cc48eef1f53c6831c06c4b1d51bc746e68a84efc71785f00d7e927e (MISMATCH)
+          - md5=2c729f293240d43f46058280c2ac00e3 (MISMATCH)
+
+-> retrieved good-sha256-good-md5.1  (file://${BASEDIR}/archive.tgz)
 -> installed good-sha256-good-md5.1
 Done.
 ### sh check-cache.sh
-MD5: archive, with mismatching checksum
+MD5: link, to sha256 archive
```
rjbou, line 1039: it fixed that!

rjbou:
there is a regression in extrasource
```diff
--- a/_build/default/tests/reftests/extrasource.test
+++ b/_build/default/tests/reftests/extrasource.out
@@ -1177,16 +1177,12 @@ The following actions will be performed:
   - install good-sha256-good-md5 1
 
 <><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
-[ERROR] Conflicting file hashes, or broken or compromised cache!
-          - sha256=hash (MISMATCH)
-          - md5=hash (MISMATCH)
-
--> retrieved good-sha256-good-md5.1  (file://${BASEDIR}/p.patch)
+-> retrieved good-sha256-good-md5.1  (cached)
 -> installed good-sha256-good-md5.1
 Done.
 ### sh check-cache.sh
-MD5: link, to sha256 patch
-SHA256: patch, with matching checksum
+MD5: patch, with matching checksum
+SHA256: patch, with mismatching checksum

File: tests/reftests/admin-cache.unix.test [hunk 6]
```diff
@@ -0,0 +1,406 @@
...
+### :V: corrupted md5:
+### ::::::::::::::::::
+### <packages/lorem/lorem.1/opam>
+opam-version: "2.0"
+### <hex1>
+4dc968ff0ee35c209572d4777b721587d36fa7b21bdc56b74a3dc0783e7b9518afbfa200a8284bf36e8e4b55b35f427593d849676da0d1555d8360fb5f07fea2
+### <hex2>
+4dc968ff0ee35c209572d4777b721587d36fa7b21bdc56b74a3dc0783e7b9518afbfa202a8284bf36e8e4b55b35f427593d849676da0d1d55d8360fb5f07fea2
+### xxd -r -p hex1 arch-lorem.tgz
+### xxd -r -p hex2 arch-evil.tgz
```
rjbou, line 360: for the md5 collision test, I used `xxd`, but there is maybe a simpler way?

kit-ty-kate:
> there is a regression in extrasource

i don't think this is a regression. Only a change of behaviour stemming from the fact that now the chosen file from the cache is the last one in the list instead of the first one. We can keep the old behaviour but it requires more code. I don't think it's broken per-say.

src/repository/opamRepository.ml
```diff
@@ -93,21 +98,27 @@ let fetch_from_cache =
...
-          if OpamFilename.exists f then Some f else None)
-        checksums
-    in
-    if List.for_all
-        (fun ck -> OpamHash.check_file (OpamFilename.to_string hit_file) ck)
-        checksums
-    then Done (Up_to_date (hit_file, OpamUrl.empty))
-    else mismatch hit_file
+          if OpamFilename.exists f
+          then (Some f, misses)
```
kit-ty-kate, line 105: If you want to keep the old behaviour:
```suggestion
          then (if hit = None then (Some f, misses) else (hit, misses))
```

tests/reftests/admin-cache.unix.test
kit-ty-kate: it looks fine by me and looks reasonably portable. If it has issues it should be fine just replacing it later

tests/reftests/admin-cache.unix.test
```diff
@@ -0,0 +1,413 @@
...
+### :IV: add a other hashes afterwards :
+### ::::::::::::::::::::::::::::::::::::
+### : ipsum : only md5 then adding sha512
+### : dolor : md5 & sha256 then adding sha512
+### : sit   : sha256 then adding md5 & sha512
+### <sedi.sh>
+regexp=$1
+file=$2
+sed "$regexp" $file > $file.tmp
+mv $file.tmp $file
```
kit-ty-kate, line 264: with its present use it's not a problem but if it's used togethr with ${BASEDIR} or whatever in the future it can become a problem and fly under the radar
```suggestion
sed "$regexp" "$file" > "$file.tmp"
mv "$file.tmp" "$file"
```

src/repository/opamRepository.ml
```diff
@@ -44,6 +44,11 @@ let cache_file cache_dir checksum =
   in
   aux cache_dir (OpamHash.to_path checksum)
 
+let link_files ~target f l =
+  List.iter (fun x ->
+      OpamFilename.link ~relative:true ~target ~link:(f x))
+    l
```
rjbou, line 50: `OpamSystem.link` removes the existing link to create a new one. Do we want that behaviour or check beforehand here if the symlink checksum -> target exists, otherwise create it ?

rjbou (commented):
* `pull_file_to_cache` need an update doc as it no more a no op if the file exists and there is no links.
* update of test (full reproducibility depends on #6103)
Otherwise, lgtm!

src/repository/opamRepository.ml
kit-ty-kate: i think we should do this separately in `OpamSystem.link` directly. I've opened https://github.com/ocaml/opam/issues/6113 to track this

src/repository/opamRepository.mli
```diff
@@ -58,7 +58,8 @@ val pull_file:
   unit download OpamProcess.job
 
 (** Same as [pull_file], but without a destination file: just ensures the file
-    is present in the cache. *)
+    is present in the cache and that every listed checksums are correctly linked
+    to the archive. *)
```
rjbou, line 62:
```suggestion
    is present in the cache and that every listed checksums are correctly linked
    to the archive, otherwise add missing links. *)
```

kit-ty-kate:
> * `pull_file_to_cache` need an update doc as it no more a no op if the file exists and there is no links.
> 
>     * update of test (full reproducibility depends on [admin hashes cache: move under opam root #6103](https://github.com/ocaml/opam/pull/6103))
>       Otherwise, lgtm!

both done

src/repository/opamRepository.mli
kit-ty-kate: done

rjbou (approved):
lgtm!

PR #6078, opened 2024-07-08
Add a benchmark showing the current performance of OpamVersionCompare

kit-ty-kate:
This PR aims at showing whether or not https://github.com/ocaml/opam/pull/5518 actually improves performance.
The idea is similar to https://github.com/ocaml/opam/pull/5900


kit-ty-kate:
After some testing #6094 seems necessary

rjbou (approved):
Thanks!

PR #6079, opened 2024-07-09
GHA: Several fixes on depexts workflow

rjbou:
Major ones:
* Fix centos by faking on with almalinux
* Fix conf-* packages install and check
* Add packages with `os-version` to check
* More caching in docker build

_I'll backport once green_

### DIFF
--- a/.github/scripts/depexts/generate-actions.sh
+++ b/.github/scripts/depexts/generate-actions.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 
 set -eu
 
--- a/.github/scripts/depexts/generate-actions.sh
+++ b/.github/scripts/depexts/generate-actions.sh
@@ -111,11 +111,11 @@ cp binary/opam $dir/opam
 
 cat >>$dir/Dockerfile << EOF
 RUN test -d /opam || mkdir /opam
-ENV OPAMROOTISOK 1
-ENV OPAMROOT /opam/root
-ENV OPAMYES 1
-ENV OPAMCONFIRMLEVEL unsafe-yes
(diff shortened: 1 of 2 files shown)


.github/scripts/depexts/generate-actions.sh
```diff
@@ -42,9 +42,12 @@ EOF
...
    OCAML_CONSTRAINT=' & < "5.0"'
     cat >$dir/Dockerfile << EOF
-FROM centos:7
+FROM almalinux:9.4
+RUN dnf install 'dnf-command(config-manager)' -y
+RUN dnf config-manager --set-enabled crb
 RUN yum install -y $mainlibs $ocaml
-RUN yum install -y gcc-c++
+RUN yum install -y gcc-c++ diffutils
+RUN sed -i 's/ID="almalinux"/ID="centos"/' /etc/os-release
```
kit-ty-kate, line 50: i don't think we should try to emulate centos

rjbou: why not?

kit-ty-kate: centos is effectively a dead platform as far as the community goes. Trying to emulate a distribution that is diverging is a bad idea

rjbou: It is a dead platform true, but there are still users on it, and it is still supported by opam.
Almalinux is meant to be a replacement to centos, providing backward compatibility. Until it diverges completely, it is a good testing platform for centos jobs.
We could add a full support for almalinux & rockylinux, but that's another issue, and they should have their own jobs.

kit-ty-kate (approved):

PR #6082, opened 2024-07-09
[2.1 backport] Port the release script from master

kit-ty-kate:
This would allow us to use the release script as-is from the tag instead of from master. This would help reproducibility

Related to https://github.com/ocaml/opam/issues/5958
Backports https://github.com/ocaml/opam/pull/6084

PR #6084, opened 2024-07-09
Remove opam 2.1 support from the release script

kit-ty-kate:
Simplifies the release script and makes it less likely to have errors
Related to https://github.com/ocaml/opam/pull/6082 and https://github.com/ocaml/opam/pull/6083

Backported to 2.2 in https://github.com/ocaml/opam/pull/6085


rjbou (commented):
Shouldn't we also add in the release readme the fact that now you need to be in the branch you want to release (and not just fetch the tag)

kit-ty-kate:
done

rjbou (approved):
Thanks!

PR #6091, opened 2024-07-11
Consistent accepted repository links 

### LINKED ISSUE #4673: doesn't work: custom repo when creating switch
In the [manual](https://opam.ocaml.org/doc/man/opam-switch.html):

> Prepare a local switch for building the packages defined in ./. This scans the current directory for package definitions, chooses a compatible compiler, creates a local switch and installs the local package dependencies.
> ```
> opam switch create trunk --repos default,beta=https://github.com/ocaml/ocaml-beta-repository.git ocaml-variants.4.10.0+trunk
> ```

This command fails with:

```
Creating repository beta...
[ERROR] Could not update repository "beta": OpamDownload.Download_fail(_, "curl: code 404 while
        downloading https://github.com/ocaml/ocaml-beta-repository.git/index.tar.gz")
[ERROR] Initial fetch of these repositories failed: beta
```

I also tried with my own custom repo and it fails with the same error.

(opam version 2.0.8)

Keryan-dev:
`opam switch create foo --repos bar=[url]` should no longer require `git+` to be accepted. Consistent with `opam repository add`.

fixes #4673

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -2557,7 +2557,7 @@ let repository cli =
 (* From a list of strings (either "repo_name" or "repo_name=URL"), configure the
    repos with URLs if possible, and return the updated repos_state and selection of
    repositories *)
-let with_repos_rt gt repos f =
+let with_repos_rt gt cli repos f =
   OpamRepositoryState.with_ `Lock_none gt @@ fun rt ->
   let repos, rt =
   match repos with
@@ -2574,9 +2574,10 @@ let with_repos_rt gt repos f =
       OpamStd.List.filter_map (function
           | (_, None) -> None
           | (n, Some url) ->
+            let handle_suffix = OpamCLIVersion.Op.(cli @>= cli2_3) in
             let repo =
               OpamStd.Option.Op.(
-                OpamUrl.parse_opt ~handle_suffix:false ~from_file:false url
+                OpamUrl.parse_opt ~handle_suffix ~from_file:false url
                 >>| fun u -> n, u)
             in
             if repo = None then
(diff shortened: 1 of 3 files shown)


kit-ty-kate (approved):

src/client/opamCommands.ml
```diff
@@ -2565,7 +2565,7 @@ let with_repos_rt gt repos f =
           | (n, Some url) ->
             let repo =
               OpamStd.Option.Op.(
-                OpamUrl.parse_opt ~handle_suffix:false ~from_file:false url
+                OpamUrl.parse_opt ~from_file:false url
```
dra27, line 2568: cf. https://github.com/ocaml/opam/issues/4673#issuecomment-872888360 - this ought to be CLI-version guarded, but that shouldn’t be too bad, as the version check _is_ then the value passed as `handle_suffix`

src/client/opamCommands.ml
```diff
@@ -2563,9 +2563,10 @@ let with_repos_rt gt repos f =
       OpamStd.List.filter_map (function
           | (_, None) -> None
           | (n, Some url) ->
+            let handle_suffix = OpamCLIVersion.(current >= (2,2)) in
```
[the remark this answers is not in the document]
Keryan-dev: This isn't right anyway so I'll repatch it. 
But `OpamCLIVersion` redefined comparison so I reckon I should make sure to use that.

tests/reftests/switch-creation.test
```diff
@@ -49,7 +49,7 @@ Done.
 ${BASEDIR}/OPAM/undef-switch
 ### # switch with repo link
 ### # expects a warning, useful to check chosen backend
-### opam switch create w-repo --repositories=from-link=git+https://github.com/ocaml/opam-repository.git --dry-run --empty
+### opam switch create w-repo --repositories=from-link=https://github.com/ocaml/opam-repository.git --dry-run --empty
```
kit-ty-kate, line 52: The reason we split tests and changes in separate commits is to the what has changed easier.
I think it would be better to not change the tests and show the command fail in the previous commit, then in this one you'd be able to see what the actual change is.

src/client/opamCommands.ml
```diff
@@ -2710,7 +2711,7 @@ let switch cli =
           compatible compiler, creates a local switch and installs the local \
           package dependencies.";
       `Pre "    opam switch create trunk --repos \
-            default,beta=git+https://github.com/ocaml/ocaml-beta-repository.git \
+            default,beta=https://github.com/ocaml/ocaml-beta-repository.git \
             ocaml-variants.4.10.0+trunk";
```
rjbou, line 2715: This should also be guarder by a cli version, no ?
```ocaml
      `Pre (Printf.sprintf
              "    opam switch create trunk --repos \
               default,beta=https://github.com/ocaml/ocaml-beta-repository.git \
               ocaml-variants.4.10.0+trunk"
              (if OpamCLIVersion.Op.(cli @>= cli2_3) then "" else "git+"));

rjbou (approved):
Thanks!
I've updated it, squashed commits and set manpage documentation dependent on requested cli version

kit-ty-kate (approved):

PR #6094, opened 2024-07-12
Make the benchmark setup process faster and the benchmark itself more stable

kit-ty-kate:
Required for https://github.com/ocaml/opam/pull/6078 as it makes the new benchmarks stable enough to be merged.

### DIFF
--- a/bench.Dockerfile
+++ b/bench.Dockerfile
@@ -1,6 +1,10 @@
 FROM debian:bullseye
 RUN apt-get update -qq && apt-get install -qq -yy make curl gcc g++ patch bzip2 git unzip
 RUN adduser --disabled-password --gecos '' --shell /bin/bash opam
+ENV OPAMREPOSHA 26770281fa1ea8b13aab979c1dfbd326e9ab512c
+RUN git clone https://github.com/ocaml/opam-repository --depth 1 /rep/opam-repository
+RUN git -C /rep/opam-repository fetch origin $OPAMREPOSHA
+RUN git -C /rep/opam-repository checkout $OPAMREPOSHA
 USER opam
 COPY --chown=opam:opam . /src
 WORKDIR /src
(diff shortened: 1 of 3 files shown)


bench.Dockerfile
```diff
@@ -1,16 +1,17 @@
...
 USER opam
-ENV OPAMREPOSHA 26770281fa1ea8b13aab979c1dfbd326e9ab512c
-RUN git clone https://github.com/ocaml/opam-repository --depth 1
-RUN git -C opam-repository fetch origin $OPAMREPOSHA
-RUN git -C opam-repository checkout $OPAMREPOSHA
-RUN opam init -n --disable-sandboxing ./opam-repository
-RUN find "$(pwd)/opam-repository" -name opam -type f > /home/opam/all-opam-files
+RUN opam init --bare -n --disable-sandboxing /rep/opam-repository
+RUN opam switch create --fake default 4.14.0
+RUN find /rep/opam-repository -name opam -type f > /home/opam/all-opam-files
```
rjbou, line 17: Reading file bench test depends on `OPAMREPOSHA`, that can evolve in the future.

tests/bench/bench.ml
```diff
@@ -20,37 +20,45 @@ let () =
...
   in
   let time_install_cmd_w_invariant =
     (* NOTE: https://github.com/ocaml/opam/issues/5502 *)
+    Gc.compact ();
     launch "opam switch create two --empty";
     launch "opam switch set-invariant core -n --sw two";
     time_cmd ~exit:0 (fmt "%s install magic-trace -y --fake --sw two" bin)
   in
-  let time_OpamSystem_read_10 =
+  let time_OpamSystem_read_100 =
```
rjbou, line 39: This reading bench test concerns reading a lot of opam files. Do we want to have also a test for reading huge file too? Currently (with this hash), `all-opam-files` file is 1.3M.

kit-ty-kate: i don't think so given the overwhelming present use-case of `OpamSystem.read`. Also i don't think this should be done in this PR if we'd want to do that. This PR is only for stability and performance improvements.

bench.Dockerfile
kit-ty-kate: Sure but we'd see that evolution on the graph associated with the commit changing it so i don't think this is an issue.

rjbou: The idea of these benches is to have them stable over the time, to only see changes related to test target, no?

kit-ty-kate: stable over runs yes, not necessarily stable over time, i would say. Benchmark stable over time is an extremely sticky thing to do and i don't think we should meddle with that (e.g. at some point in the future the machine running those benchmarks will die and the result of the bench will change)

rjbou: Of course we can't rely on it indefinitely, as it depends on a machine, but i think that we shouldn't have a benchmark that changes if an external value changes (here repo hash). In fact, it is the same for other opam install bench tests too, they depend son that repo hash too.
Anyway, in the future if we need to change that hash, we can update the test accordingly (e.g. have 2 repo hashes).

rjbou (approved):
Thanks!

tests/bench/bench.ml
rjbou: fair enough

PR #6098, opened 2024-07-15
opam switch list-available will not display compilers flagged with avoid-version/deprecated unless --all is given

### LINKED ISSUE #6089: opam switch list-available should only display packages that aren't flagged with avoid-version/deprecated
https://discuss.ocaml.org/t/error-on-opam-install-utop-fedora/14948/6

I think that command could be improved by only returning the list of packages without those flags by default and adding a `--all` argument to get the previous behaviour

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/6089

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -2651,7 +2651,9 @@ let switch cli =
     cli_original, "list-available", `list_available, ["[PATTERN]"],
     "Lists all the possible packages that are advised for installation when \
      creating a new switch, i.e. packages with the $(i,compiler) flag set. If \
-     no pattern is supplied, all versions are shown.";
+     no pattern is supplied, all versions are shown. By default, packages \
+     flagged with $(i,avoid-version) or $(i,deprecated) will not appear \
+     unless $(b,--all) is given.";
     cli_original, "show", `current, [],
     "Prints the name of the current switch.";
     cli_from cli2_1, "invariant", `show_invariant, [],
@@ -2807,6 +2809,12 @@ let switch cli =
        right away: wait for the next $(i,install), $(i,upgrade) or similar \
        command."
   in
+  let all =
+    mk_flag ~cli (cli_from cli2_3) ["all"]
+      "Only for $(i,list-available): show all available compilers, \
+       regardless of whether they are flagged with $(i,avoid-version) \
+       or $(i,deprecated)."
+  in
   (* Deprecated options *)
   let d_alias_of =
     mk_opt ~cli (cli_between cli2_0 cli2_1)
@@ -2818,7 +2826,7 @@ let switch cli =
   let switch
       global_options build_options command print_short
       no_switch packages formula empty descr full freeze no_install deps_only repos
-      force no_action
+      force no_action all
       d_alias_of d_no_autoinstall params () =
     if d_alias_of <> None then
       OpamConsole.warning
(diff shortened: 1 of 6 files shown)


rjbou (changes requested):
Is it possible to add a test to show the new option?

kit-ty-kate:
> Is it possible to add a test to show the new option?

done

src/client/opamCommands.ml
```diff
@@ -2894,6 +2907,10 @@ let switch cli =
            order = `Custom order;
         }
         compilers;
+      if not all then
+        OpamConsole.note
+          "Some compilers may have been hidden (e.g. pre-releases). \
+           If you want to display them, run: opam switch list-available --all";
```
rjbou, line 2913:
```suggestion
           If you want to display them, run: 'opam switch list-available --all'";
```

src/client/opamCommands.ml
```diff
@@ -2894,6 +2907,10 @@ let switch cli =
            order = `Custom order;
         }
         compilers;
+      if not all then
```
rjbou, line 2910: It is better to show the message only if there was some hidden packages.
Ftm, we don't have the information. If we retrieve the complete `compilers` set and filter it afterwards compilers flagged with `avoid version` or `deprecated`, we can access to the diff.

src/client/opamCommands.ml
```diff
@@ -2803,7 +2808,7 @@ let switch cli =
   let switch
       global_options build_options command print_short
       no_switch packages formula empty descr full freeze no_install deps_only repos
-      force no_action
+      force no_action all
```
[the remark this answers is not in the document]
kit-ty-kate: done

src/client/opamListCommand.mli
```diff
@@ -57,6 +57,7 @@ type selector =
   | Pattern of pattern_selector * string
   | Atoms of atom list
   | Flag of package_flag
+  | NotFlag of package_flag
```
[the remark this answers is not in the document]
kit-ty-kate: done

src/client/opamCommands.ml
kit-ty-kate: done

src/client/opamCommands.ml
```diff
@@ -2815,6 +2821,9 @@ let switch cli =
     if d_no_autoinstall then
       OpamConsole.warning "Option %s is deprecated, ignoring it."
         (OpamConsole.colorise `bold "--no-autoinstall");
+    if all && command <> Some `list_available then
+      OpamConsole.warning "Option %s is not used"
```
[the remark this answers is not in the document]
rjbou: I'm also wondering if we should add note about `--all` option in `list-available` section.

rjbou:
I've updated the PR with a more complete test for witch list-available (that highlighted #6152) and squashed some commits.

src/client/opamCommands.ml
```diff
@@ -2909,6 +2928,10 @@ let switch cli =
            order = `Custom order;
         }
         compilers;
+      if not all && not (OpamPackage.Set.equal all_compilers compilers) then
+        OpamConsole.note
+          "Some compilers may have been hidden (e.g. pre-releases). \
```
kit-ty-kate, line 2933:
```suggestion
          "Some compilers have been hidden (e.g. pre-releases). \
```

rjbou (approved):
thanks!

PR #6100, opened 2024-07-16
Avoid double symlink in archive cache

rjbou:
pro: have a cleaner cache by avoiding double symlinks (e.g. `md5 -> sha512` & `sha256 -> sha512` instead of `md5 -> sha256 -> sha512`)
cons: when the cache contains a `sha256 -> md5` link, installing 2 packages that refer to the same archive but one with an md5 and the other with a sha256 ; the one with a sha256 won't be found in cache and redownloaded.

Another solution is to have opam always compute and store all checksums linked (from sha*'s), like proposed by @kit-ty-kate. 

* [ ] add a test
* [ ] update master changes
* [ ] queued on #6068

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -28,6 +28,7 @@ users)
 
 ## Install
   * Fix package name display for no agreement conflicts [#6055 @rjbou - fix #6030]
+  * Make fetching an archive from cache add missing symlinks [#6068 @kit-ty-kate - fix #6064]
 
 ## Remove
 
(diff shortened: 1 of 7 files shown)


rjbou:
consider closing ?

kit-ty-kate:
> consider closing ?

i think it's fine to leave it open for now. It's not critical but might be nice to have in the future if we're satisfied with the compromises of the implementation

PR #6102, opened 2024-07-17
Stop using polymorphic comparison when comparing switch_selections

kit-ty-kate:
Related to https://github.com/ocaml/ocaml/pull/9928


src/format/opamTypesBase.mli
```diff
@@ -112,3 +112,6 @@ val raw_of_op: euok_writeable env_update_op_kind -> OpamParserTypes.FullPos.env_
 (* Path transformers & separator functions *)
 val string_of_path_format: path_format -> string
 val char_of_separator: separator -> char
+
```
rjbou, line 115:
```suggestion

(** Comparators **)
(* Switch selections *)
```

kit-ty-kate: done

rjbou (approved):
Thanks!

PR #6103, opened 2024-07-17
admin hashes cache: move under opam root

rjbou:
`opam admin add-hashes` creates a cache that is actually stored under `~/.cache/hash-cache`. Move it in `<opamroot>/download-cache/hash-cache`.

### DIFF
--- a/tests/reftests/admin.test
+++ b/tests/reftests/admin.test
@@ -513,6 +513,16 @@ for p in $@; do
   fi
   echo "}" >> $file
 done
+### <check-hash-ca.sh>
+hc=$OPAMROOT/download-cache/hash-cache
+echo "root: $(basename $(dirname $(dirname "$hc")))"
+if [ -d "$hc" ]; then
+  for f in $(ls "$hc") ; do
+    echo "$f: $(wc -l "$hc/$f" | cut -f 1 -d ' ')"
+  done
+else
+  echo "no hash-cache"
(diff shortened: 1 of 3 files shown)


tests/reftests/admin.test
```diff
@@ -513,6 +513,16 @@ for p in $@; do
   fi
   echo "}" >> $file
 done
+### <check-hash-ca.sh>
+hc=$OPAMROOT/download-cache/hash-cache
+echo "root: $(basename $(dirname $(dirname $hc)))"
+if [ -d "$hc" ]; then
+  for f in $(ls $hc) ; do
```
kit-ty-kate, line 520:
```suggestion
  for f in "$(ls "$hc")" ; do
```

tests/reftests/admin.test
```diff
@@ -513,6 +513,16 @@ for p in $@; do
...
+    echo "$f: $(wc -l $hc/$f | cut -f 1 -d ' ')"
```
kit-ty-kate, line 521:
```suggestion
    echo "$f: $(wc -l "$hc/$f" | cut -f 1 -d ' ')"
```

kit-ty-kate (approved):
LGTM. I've renamed the `hash_tables` function into `create_hash_tables` otherwise it's way too confusing between the `hash_tables` function and the `hash_tables` value.

PR #6104, opened 2024-07-17
Add support for wget2

kit-ty-kate:
Currently on system that have wget2 (e.g. Fedora 40), downloading archives do not work if wget is used.
Per wget2's man page:
```
   --content-disposition
       If this is set to on, experimental (not fully-functional) support for “Content-Disposition” headers is enabled.  This can currently result in extra round-trips to the server for a “HEAD” request, and is
       known to suffer from a few bugs, which is why it is not currently enabled by default.

       This option is useful for some file-downloading CGI programs that use “Content-Disposition” headers to describe what the name of a downloaded file should be.
```
I don't think we want to enable an experimental argument. Given we use `-O` already I'm not sure what the goal of it was.
This was original added 9 years ago in https://github.com/ocaml/opam/commit/c49430d79cffc0180e2d125456f1360bdc894884

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -25,6 +25,7 @@ users)
 ## Config report
 
 ## Actions
+  * Add support for wget2 [#6104 @kit-ty-kate]
 
 ## Install
   * Fix package name display for no agreement conflicts [#6055 @rjbou - fix #6030]
--- a/src/repository/opamDownload.ml
+++ b/src/repository/opamDownload.ml
@@ -35,7 +35,6 @@ let curl_args = [
 ]
 
 let wget_args = [
-  CString "--content-disposition", None;
   CString "-t", None; CIdent "retry", None;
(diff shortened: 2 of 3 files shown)


kit-ty-kate:
> Given we use `-O` already I'm not sure what the goal of it was.
> This was original added 9 years ago in [c49430d](https://github.com/ocaml/opam/commit/c49430d79cffc0180e2d125456f1360bdc894884)

it looks like it was simply a remnant of a time when opam wasn't using -O with wget or -o with curl. So it should be totally safe to merge

rjbou (approved):
It was introduced in #271 ([change](https://github.com/ocaml/opam/pull/271/files#diff-dedcd3b96141f5b4cddeae81206cc50659d8b3911de1dc4f1007332874faa2e5R435)).

PR #6106, opened 2024-07-18
Fix most unhelpful conflict explanation message by merging all formulas together

### LINKED ISSUE #4373: Unhelpful conflict messages

### LINKED ISSUE #5422: Diverse solver functions take too long to check some requests
Mentioned in #5407 
Seen in https://github.com/ocaml/opam-repository/pull/23020

Steps to reproduce:
```
cd $(mktemp -d)
git clone --recursive "https://github.com/ocaml/opam-repository.git" && cd "opam-repository" && git fetch origin "refs/pull/23020/head" && git reset --hard 4eff4643
git fetch origin master
git merge --no-edit 8e66fa6235625556c21fe7273ab6517d3f2fc565
cat > ../Dockerfile <<'END-OF-DOCKERFILE'
FROM ocaml/opam:debian-11-ocaml-4.07@sha256:83c1e1e1eaae44453b8a44b940b5972d0781b2680a5a4b928046862208a4e629
USER 1000:1000
WORKDIR /home/opam
RUN opam init --reinit -ni
ENV OPAMDOWNLOADJOBS="1"
ENV OPAMERRLOGLEN="0"
ENV OPAMSOLVERTIMEOUT="500"
ENV OPAMPRECISETRACKING="1"
RUN rm -rf opam-repository/
COPY --chown=1000:1000 . opam-repository/
RUN opam repository set-url --strict default opam-repository/
ENV OPAMCRITERIA="-removed,-count[avoid-version,changed],-count[version-lag,request],-count[version-lag,changed],-count[missing-depexts,changed],-changed"
ENV OPAMFIXUPCRITERIA="-removed,-count[avoid-version,changed],-count[version-lag,request],-count[version-lag,changed],-count[missing-depexts,changed],-changed"
ENV OPAMUPGRADECRITERIA="-removed,-count[avoid-version,changed],-count[version-lag,request],-count[version-lag,changed],-count[missing-depexts,changed],-changed"
RUN opam pin add -k version -yn curly.0.3.0 0.3.0
RUN opam update --depexts
RUN opam reinstall curly.0.3.0
RUN opam reinstall --with-test curly.0.3.0
END-OF-DOCKERFILE
```

kit-ty-kate:
Fixes #4373 

This PR fixes it by:
- using every available hints given by dose3 instead of just taking the first hint for a given set of packages (removal of `ChainSet.precludes`)
- merging all the extra information together

In some cases this makes the message as verbose but at least it is correct and helpful.
We can try to make them even more readable in a later work but at least this fixes the biggest issue at hand: conflict messages telling lies and being unreadable. 

~Built on top of https://github.com/ocaml/opam/pull/5210~
Also fixes https://github.com/ocaml/opam/issues/5422 (alcotest now only takes 45s to print the explanation instead of +infinit)


src/solver/opamCudf.ml
```diff
@@ -1037,19 +1028,121 @@ let extract_explanations packages cudfnv2opam reasons : explanation list =
...
-            explanations, ct_chains
+            explanations
         with Not_found ->
-          explanations, ct_chains)
-      ([], ct_chains) reasons
+          explanations)
+      [] reasons
+  in
+
+  let rec is_included_in_formula formula1 formula2 =
```
kit-ty-kate, line 1045:
```suggestion
  let rec simplify_formula formula1 formula2 =
```

tests/reftests/deps-only.test
```diff
@@ -85,15 +85,15 @@ depends: "b"
 ### opam install a c --deps-only
 [ERROR] Package conflict!
   * Missing dependency:
-    - a
+    - deps-of-c -> b -> a
```
rjbou, line 88: For packages `deps-of-*`, it should be displayed differently. It is an internal package created by opam in the deps-only mechanism. Maybe `(deps of c)`?

rjbou, line 88: Here, the message was good as `a` is requested in the cli.

src/solver/opamCudf.ml
```diff
@@ -1144,20 +1144,13 @@ let extract_explanations packages cudfnv2opam reasons : explanation list =
...
-        | _ -> false)
-  in
   log ~level:3 "Explanations: %a" Pp_explanation.pp_explanationlist explanations;
   match explanations with
   | [] ->
     OpamConsole.error_and_exit `Internal_error
       "Internal error while computing conflict explanations:\n\
        sorry about that. Please report how you got here in \
        https://github.com/ocaml/opam/discussions/5130 if possible."
-  | `Missing (_, sdeps, fdeps) :: rest when same_depexts sdeps fdeps rest ->
```
rjbou, line 1067: Is it possible to display differently the dependency path/conflict? Like that the focus remain at first look on the missing package, and the information of the dependency is available.

src/solver/opamCudf.ml
```diff
@@ -850,8 +838,8 @@ let extract_explanations packages cudfnv2opam reasons : explanation list =
   in
   let print_set pkgs =
```
rjbou, line 839: It is no more `print_set`, but something like `construct_set` ?

src/solver/opamCudf.ml
```diff
@@ -864,10 +852,10 @@ let extract_explanations packages cudfnv2opam reasons : explanation list =
...
             OpamFormula.formula_of_version_set all_versions versions
           in
-          OpamFormula.to_string (Atom (name, formula)))
+          Atom (name, formula))
         nvs
     in
-    String.concat ", " (OpamPackage.Name.Map.values strs)
+    `Formula (false, OpamFormula.ors (OpamPackage.Name.Map.values strs))
   in
   let cs_to_string ?(hl_last=true) cs =
```
rjbou, line 860: same here, `construct_cs` ?

src/solver/opamCudf.ml
```diff
@@ -1037,19 +1028,120 @@ let extract_explanations packages cudfnv2opam reasons : explanation list =
...
+      | `Request, `Request -> f
+      | `Formula (bold, f), `Formula (bold', f') ->
+        if Bool.equal bold bold' then
+          match simplify_formula f f' with
+          | Some simplified -> `Formula (bold, simplified)
+          | None -> raise (Invalid_argument "")
+        else
+          raise (Invalid_argument "")
+      | `Invariant, _ | `Request, _ | `Formula _, _ ->
+        raise (Invalid_argument "")
```
rjbou, line 1083: If these case should not happen, it should be an internal error with a message. If it is sure that it won't happen, it should be then be `assert false` with a comment.

rjbou: For the `bold <> bold'` case, if we are sure that it won't happen, it is possible to take directly the first `bold`.

src/solver/opamCudf.ml
```diff
@@ -850,8 +838,8 @@ let extract_explanations packages cudfnv2opam reasons : explanation list =
   in
   let print_set pkgs =
     if Set.exists is_artefact pkgs then
-      if Set.exists is_opam_invariant pkgs then "(invariant)"
-      else "(request)"
+      if Set.exists is_opam_invariant pkgs then `Invariant
```
rjbou, line 841: It's worth adding a comment for the result type. It is obvious for `'Invariant` and `'Request` what they are for, but `'Formula of bool * formula` the boolean is not obvious what it is.

src/solver/opamCudf.ml
```diff
@@ -732,24 +732,12 @@ module ChainSet = struct
...
-
   let length cs = fold (fun l acc -> min (List.length l) acc) cs max_int
 end
 
 type explanation =
   [ `Conflict of string option * string list * bool
-  | `Missing of string option * string *
-                (OpamPackage.Name.t * OpamFormula.version_formula)
-                  OpamFormula.formula
+  | `Missing of string option * string * OpamFormula.t
```
rjbou, line 740: Is it worth updating this type to include in it the `'NoLongerAvailable` variant, and even keep the high level information of formula, invariant, request in it; keeping all the work of *stringification* to the `string_of_explanation` function?

src/solver/opamCudf.ml
```diff
@@ -1037,35 +1028,129 @@ let extract_explanations packages cudfnv2opam reasons : explanation list =
...
+        simplify_formula x1 y1 >>= fun z1 ->
+        simplify_formula x2 y2 >>= fun z2 ->
+        Some (Or (z1, z2))
+    | Empty, _
+    | Atom _, _
+    | And _, _
+    | Or _, _ -> None
+  in
+
+  let explanations =
```
rjbou, line 1070: It may worth adding some comment of the explanations constructions, as there is several steps to have the final explanation list?

rjbou (commented):
I've put some remarks on the code (non linear) but the idea is good, it is working well!

With the third commit, the title of the PR should change to something more general like "Conflict messages: fix most of unhelpful explanations and display full dependency path"

tests/reftests/conflict-4373.test
```diff
@@ -1,10 +0,0 @@
-c1d23f0e
-### OPAMVAR_enable_ocaml_beta_repository=1 opam switch create --fake 4.12.0
-[ERROR] No compiler matching `4.12.0' found, use `opam switch list-available' to see what is available, or use `--packages' to select packages explicitly.
-# Return code 5 #
-### opam install --show expect.0.0.6
-[ERROR] No switch is currently set. Please use 'opam switch' to set or install a switch
-# Return code 50 #
-### opam install --show ppx_expect.v0.14.0
-[ERROR] No switch is currently set. Please use 'opam switch' to set or install a switch
-# Return code 50 #
```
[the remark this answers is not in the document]
kit-ty-kate: Yes, i mentioned it in https://github.com/ocaml/opam/pull/5210#discussion_r1681663194 but i forgot to port it to this PR too

tests/reftests/deps-only.test
kit-ty-kate: I agree, however the problem already existed before and i feel like it's a bit outside of the scope of this PR.
The virtual `deps-of-*` package is a hack introduced in `OpamClient` and getting rid of it seems tedious. I'd rather do that separately. If you're ok with that I'll open a new ticket to track it

kit-ty-kate: Not really. `opam install --deps-only a` is totally valid so the missing dependency is really via `c`

src/solver/opamCudf.ml
kit-ty-kate: i don't understand that

tests/reftests/deps-only.test
rjbou: I was not proposing to remove the `deps-of-*` mechanism, but to display it differently as now it will be more shown. But yes, it can be an improvement of this RP that we can track with an issue.

src/solver/opamCudf.ml
rjbou: before we had:
```
  * Missing dependency:
    - a
```
Now we will have
```
  * Missing dependency:
    - c -> b -> a
```
I was proposing to display differently `a` from the prefixed path `c -> b ->`. Like that at first look it is directly visible that the main information is `a`.
It is kind of UI trick, i didn't see on a terminal how it is, I though about it while looking at the test, which are not colored.

tests/reftests/deps-only.test
kit-ty-kate: The thing is that the mechanism is linked to the way it's displayed. You cannot do one without the other as a package named `deps-of-something` is perfectly valid and cannot be distinguished from the `deps-of-*` virtual package hack

src/solver/opamCudf.ml
kit-ty-kate: In that case i disagree. I think the reason for the failure should match the actual reason. Adding special cases like this would complicate the code and confuse users.

rjbou: I don't want to hide, but to highlight. For example is `a` (the main missing) is bold, we would have
image
We have all the information (the full path), with a highlight on the main reason (missing package).

src/solver/opamCudf.ml
kit-ty-kate: I'm not sure to understand what you mean. `` `NoLongerAvailable `` is not defined and only locally used. The change highlighted here is only cosmetic (`OpamFormula.t = (OpamPackage.Name.t * OpamFormula.version_formula) OpamFormula.formula`)

src/solver/opamCudf.ml
kit-ty-kate: The use of `Invalid_argument` here is used because this function is called using `List.fold_left2` and these cases are valid but we want to raise early if the two conflict do not match

src/solver/opamCudf.ml
kit-ty-kate: fixed in 3b9d8032e

src/solver/opamCudf.ml
kit-ty-kate: the return type of this function is still a string. Also `cs` is different from `cst/construct` and is a different concept. I've improved the naming difference in 3b9d8032e

src/solver/opamCudf.ml
kit-ty-kate: fixed in 3b9d8032e

src/solver/opamCudf.ml
kit-ty-kate: added in https://github.com/ocaml/opam/pull/6106/commits/b591657225fe9c5b5904788e3c107b567470c89d. What do you think?

src/solver/opamCudf.ml
kit-ty-kate: I'm not sure i follow. The highlight is already done, the last item on the list is already bold and red when displayed in an interactive terminal

rjbou (approved):
All comments resolved in meeting, thanks!

kit-ty-kate:
squashed

PR #6107, opened 2024-07-18
Rework download failures for better error reporting 

Keryan-dev:
Early WIP for eventual discussions.

General wishlist so far:
- Better early error diagnostic
- Reduce generic error format uses to a minimum
- Eliminate raw exception printing in error messages
- Eliminate assumptions about error string content

### DIFF
--- a/tests/reftests/extrasource.test
+++ b/tests/reftests/extrasource.test
@@ -563,7 +563,7 @@ The following actions will be performed:
           got      md5=hash
 [ERROR] Failed to get extra source "i-am-a-patch" of multiple-md5.1: Checksum mismatch
 
-OpamSolution.Fetch_fail("Checksum mismatch")
+Checksum mismatch
 
 
 <><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
@@ -583,7 +583,7 @@ The following actions will be performed:
           got      md5=hash
 [ERROR] Failed to get extra source "i-am-a-patch" of multiple-md5.1: Checksum mismatch
 
-OpamSolution.Fetch_fail("Checksum mismatch")
+Checksum mismatch
 
 
 <><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
@@ -618,7 +618,7 @@ The following actions will be performed:
           got      md5=hash
 [ERROR] Failed to get extra source "i-am-a-patch" of bad-md5.1: Checksum mismatch
 
-OpamSolution.Fetch_fail("Checksum mismatch")
+Checksum mismatch
 
 
 <><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
@@ -638,7 +638,7 @@ The following actions will be performed:
           got      md5=hash
 [ERROR] Failed to get extra source "i-am-a-patch" of bad-md5.1: Checksum mismatch
 
-OpamSolution.Fetch_fail("Checksum mismatch")
+Checksum mismatch
 
 
 <><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
@@ -690,7 +690,7 @@ The following actions will be performed:
           got      sha256=hash
 [ERROR] Failed to get extra source "i-am-a-patch" of good-md5-bad-sha256.1: Checksum mismatch
 
-OpamSolution.Fetch_fail("Checksum mismatch")
+Checksum mismatch
 
 
 <><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
@@ -740,7 +740,7 @@ The following actions will be performed:
           got      sha256=hash
 [ERROR] Failed to get extra source "i-am-a-patch" of good-md5-bad-sha256.1: Checksum mismatch
 
-OpamSolution.Fetch_fail("Checksum mismatch")
+Checksum mismatch
 
 
 <><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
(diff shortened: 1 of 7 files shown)


src/format/opamTypes.mli
```diff
@@ -48,13 +48,33 @@ type std_path =
   | Lib | Bin | Sbin | Share | Doc | Etc | Man
   | Toplevel | Stublibs
 
+(** Download failure explanation *)
+type dl_fail_reason = string option * string
+(** Respectively the short and long version of an error message. The
+    usage is: the first argument is displayed on normal mode (nothing
+    if [None]), and the second one on verbose mode. *)
```
kit-ty-kate, line 55: Using a record instead here would make it easier to read, use and reason about i think
```suggestion
type dl_fail_reason = {
  short_reason : string option;
  long_reason : string;
}
```

kit-ty-kate (commented):
apart from the one suggestion, the design looks great

Keryan-dev:
Trimmed up, and undrafted.

Some tests changes, still not ideal but there should be no loss of information.

src/repository/opamDownload.ml
```diff
@@ -94,27 +95,35 @@ let tool_return url ret =
...
+                  long_reason = Printf.sprintf "Curl failed: %s"
+                      (OpamProcess.result_summary ret); })
+      else match ret.OpamProcess.r_stdout with
+        | [] -> Some Curl_empty_response
+      | l  ->
+        let code = List.hd (List.rev l) in
+        let num = try int_of_string code with Failure _ -> 999 in
+        if num >= 400
+        then Some (Curl_error_response code)
+        else None
```
kit-ty-kate, line 117: Missing indent:
```suggestion
        | l  ->
          let code = List.hd (List.rev l) in
          let num = try int_of_string code with Failure _ -> 999 in
          if num >= 400
          then Some (Curl_error_response code)
          else None
```

src/client/opamAction.ml
```diff
@@ -290,27 +290,29 @@ let download_shared_source st url nvs =
   OpamProcess.Job.catch (fun e ->
       let na =
         match e with
-        | OpamDownload.Download_fail (s,l) -> (s,l)
-        | e -> (None, Printexc.to_string e)
+        | OpamDownload.Download_fail (Generic_failure reason) -> reason
+        | e -> {short_reason = None; long_reason = Printexc.to_string e }
```
kit-ty-kate, line 294:
```suggestion
        | e -> {short_reason = None; long_reason = Printexc.to_string e}
```

src/client/opamAdminCommand.ml
```diff
@@ -10,6 +10,7 @@
 (**************************************************************************)
 
 open OpamTypes
+open OpamTypesBase
```
kit-ty-kate, line 13:
```suggestion
```

src/client/opamAdminCommand.ml
```diff
@@ -284,8 +286,11 @@ let cache_command cli =
            (OpamPackage.Map.keys errors));
       OpamConsole.errmsg "%s"
         (OpamStd.Format.itemize (fun (nv,el) ->
+             let reasons =
+               List.map (fun e -> (get_dl_failure_reason e).long_reason) el
```
kit-ty-kate, line 290:
```suggestion
               List.map (fun e -> (OpamTypesBase.get_dl_failure_reason e).long_reason) el
```

kit-ty-kate (commented):
I'd recommend to avoid opens.
They make the codebase harder to maintain, read and reason about.

tests/reftests/extrasource.test
```diff
@@ -583,7 +583,7 @@ The following actions will be performed:
           got      md5=hash
 [ERROR] Failed to get extra source "i-am-a-patch" of multiple-md5.1: Checksum mismatch
 
-OpamSolution.Fetch_fail("Checksum mismatch")
+Checksum mismatch
```
kit-ty-kate, line 586: this message feels redundant with the ones above

Keryan-dev: In what way ? The lack of information should be the object of a patch for checksum related errors.

kit-ty-kate:
```
[ERROR] multiple-md5.1/i-am-a-patch: Checksum mismatch for file://${BASEDIR}/p.patch:
          expected md5=hash
          got      md5=hash
[ERROR] Failed to get extra source "i-am-a-patch" of multiple-md5.1: Checksum mismatch

Checksum mismatch
```
"checksum mismatch" is displayed three times one after the other.

I believe the difference is that both errors above are displayed as the error happens and the last one is displayed when the actions finished.

Since this PR is no longer just an API improvement, i think it would be nicer to make it so that such redundant error messages are squashed into one

leviroth:
I found a case where this change makes the error considerably less descriptive:

```bash
# On opam 2.2.1
$ opam repository add example http://example.com
[ERROR] Could not update repository "example": OpamDownload.Download_fail(_, "curl: code 404 while downloading http://example.com/index.tar.gz")
[ERROR] Initial repository fetch failed

# On this branch
$ dune exec src/client/opamMain.exe -- repository add example http://example.com
[ERROR] Could not update repository "example": OpamDownload.Download_fail(_)
[ERROR] Initial repository fetch failed
```

I believe the issue arises because of the use of `Printexc.to_string` in various places. As I understand it, the default exception printer is unable to represent variants with arguments. Thus, changing an exception like `Download_fail of string option * string` into `Download_fail of dl_failure` means that none of the error description can be rendered by `Printexc` (as, indeed, the first argument is currently rendered).

PR #6116, opened 2024-07-24
Always list all the repositories in 'opam config report' regardless of whether or not a switch is currently set

kit-ty-kate:
Supersedes https://github.com/ocaml/opam/pull/5799 as it was its original intent behind that PR.
I don't see a reason why `repositories` should only be shown when a switch is set, so this PR makes it so that it is always displayed.


rjbou (approved):
There is no reason, thanks!

Issue #6118, opened 2024-07-25
Comparing package versions with opam cli


mbarbin:
Hello!

I don't think this is an issue with opam, it's more like a question: What is the syntax in opam for a package version that denote a pre-release version?

I was trying to follow https://semver.org/#spec-item-9 but was surprised by this:

```shell
$ opam install ppxlib.0.33.0
The following actions will be performed:
=== downgrade 1 package
  ↘ ppxlib                      0.33.0-preview.1 to 0.33.0
```

(I would have expected that `0.33.0-preview.1` < `0.33.0`, and thus this install being an upgrade, not a downgrade.

I sort of recall seeing perhaps '\~' used as a separator. Is this the way? (as in, I should have named the package `0.33.0~preview.1` instead). I couldn't find '\~' in the semver doc. Is this a different convention?

Sorry if it's a duplicate. I am not well versed in pre-release packages with opam, and trying to establish habits that will work well.
Thank you!

kit-ty-kate:
The version ordering is explain in the manual: https://opam.ocaml.org/doc/Manual.html#version-ordering
but in a nutshell:
- opam follows Debian versioning scheme, not semver
- `~` (tilde) is indeed the right separator for what you are trying to achieve

kit-ty-kate:
You can also find examples of that very same package in https://github.com/ocaml/opam-repository/blob/master/packages/ppxlib/ppxlib.0.32.1~5.2preview/opam
Out of curiosity `ppxlib.0.33.0-preview.1` is not present in opam-repository (although the stable 0.33.0 release is), does this package come from a custom repository?

mbarbin:
Thanks a lot!

I was confused a bit by https://github.com/ocaml/opam?tab=readme-ov-file#versioning but I understand now this statement is about the version of the `opam` binary itself, not a statement about the version of user packages.

> Out of curiosity ppxlib.0.33.0-preview.1 is not present in opam-repository (although the stable 0.33.0 release is), does this package come from a custom repository?

Yes, I had published a preview [here](https://github.com/mbarbin/opam-repository) while working on some PR with the ppxlib devs.

I will adjust my process to use the '~' separator then, thanks again!

mbarbin:
Thinking about this a bit @kit-ty-kate, what do you think about adding a quick sanity check opam sub command allowing some validation and comparison of 2 versions, exposing the same logic as used by the rest of the opam system?

I'm thinking something as simple as:

```bash
[...]

PR #6119, opened 2024-07-26
doc: add brief note about version ordering and OCaml REPL example

mbarbin:
Minor documentation update (README & User Manual) about version ordering in opam:

- Add a note to distinguish the version scheme used for the opam binary from that of user packages
- Add an example of how to execute the function implemented in opam for quick sanity checks

This follows a discussion in #6118 where I had some questions about this.

OCaml REPL snippet curtesy of @kit-ty-kate in https://github.com/ocaml/opam/issues/6118#issuecomment-2252387409_

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,10 @@ that's not an option for you, you can use `git format-patch` and email us.
 
 ## Versioning
 
-The release cycle respects [Semantic Versioning](http://semver.org/).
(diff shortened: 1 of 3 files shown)


kit-ty-kate (approved):
Thanks a lot!

kit-ty-kate:
Thanks again for the contribution. The new manual should deployed on https://opam.ocaml.org/doc/Manual.html by tomorrow

PR #6121, opened 2024-07-29
Fix "opam install --check pkg" when pkg depends on a non-existing package

kit-ty-kate:
As the test show non-existing packages (or atoms requiring a version that does not exist yet for example) does not currently work when using `opam install --check`


rjbou (approved):
Thanks!

Issue #6097, opened 2024-07-15
opam install --check does not check whether dependencies are installed recursively


kit-ty-kate:
If a dependency is added or changed in the dependency tree of the package being tested with `--check`, then the change won't be detected.

Example:
```
$ cat ./a.opam
opam-version: "2.0"
depends: ["b"]
$ cat ./b.opam
opam-version: "2.0"
$ cat ./c.opam
opam-version: "2.0"
$ opam pin add -yn .
$ opam install ./b.opam
$ opam install --check ./a.opam
All dependencies installed
$ echo 'depends: ["c"]' >> ./b.opam
$ opam pin add -yn .
$ opam install --check ./a.opam
```

PR #6122, opened 2024-07-29
Make "opam install --check" check if all dependencies are installed recursively

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/6097
Queued on top of #6121 
Queued on https://github.com/ocaml/opam/pull/6123

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -34,6 +34,7 @@ users)
   * [BUG] Fix `opam install --deps-only` set direct dependencies as root packages [#6125 @rjbou]
   * [BUG] Fix `opam install --check pkg` when pkg depends on a non-existing package [#6121 @kit-ty-kate]
   * Disable shallow clone by default except for opam repositories [#6146 @kit-ty-kate - fix #6145]
+  * Improve performance of `opam install --check` [#6122 @kit-ty-kate]
 
 ## Build (package)
   * ◈ Add `--verbose-on` option to enable verbose mode on specified package names [#5682 @desumn @rjbou]
--- a/master_changes.md
+++ b/master_changes.md
@@ -229,6 +229,7 @@ users)
   * Add `OpamTypesBase.switch_selections_{compare,equal}`: proper comparison functions for `OpamTypes.switch_selections` [#6102 @kit-ty-kate]
   * `OpamFormula`: add `exists` [#5317 @kit-ty-kate]
   * `OpamTypes.universe`: make `u_available` and `u_attrs` lazy to speedup actions that do not require availiblity information [#5317 @kit-ty-kate - fix #5314]
+  * `OpamFormula`: add some missing comparison functions for `relop`, `version_constraint` and `atom` (`compare_relop`, `compare_version_constraint` and `compare_atom` respectively) [#6122 @kit-ty-kate]
 
 ## opam-core
   * `OpamStd.Env`: add `env_string_list` for parsing string list environment variables (comma separated) [#5682 @desumn]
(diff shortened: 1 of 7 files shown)


src/client/opamClient.ml
```diff
@@ -1945,10 +1945,9 @@ let init
   gt, rt, default_compiler
 
 let check_installed ~build ~post t atoms =
-  let available = (Lazy.force t.available_packages) in
   let pkgs =
     OpamPackage.to_map
-      (OpamFormula.packages_of_atoms available atoms)
+      (OpamFormula.packages_of_atoms t.packages atoms)
```
rjbou: The performance improvement needs to be in its own PR.

rjbou: In fact, as the code is removed in the rest of the PR, shouldn't we not keep the commit?

rjbou (commented):
lgtm on the idea ans its proper code.
I've added a test for basic cases for install check, like that we are sure that these changes don't change them. I don't know if it should be in this PR or #6121.

tests/reftests/install-check.test
```diff
@@ -0,0 +1,69 @@
...
+The following actions will be faked:
+=== upgrade 1 package
+  - upgrade a 1 to 3
+=== install 1 package
+  - install b 1      [required by a]
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+Faking installation of b.1
+Faking installation of a.3
+Done.
```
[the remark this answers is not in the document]
kit-ty-kate: no other reason other than the force of habit and just in case it makes it faster

src/client/opamClient.ml
kit-ty-kate: the code is moved elsewhere in the final commit and is here to offset the performance cost of looking at the dependencies recursively

kit-ty-kate:
The benchmark failure is real and is caused by an infinit loop. I'll have a look

master_changes.md
```diff
@@ -212,6 +212,7 @@ users)
 
 ## opam-format
   * Add `OpamTypesBase.switch_selections_{compare,equal}`: proper comparison functions for `OpamTypes.switch_selections` [#6102 @kit-ty-kate]
+  * `OpamFormula`: add some missing comparison functions for `relop`, `version_constraint` and `atom` (`compare_relop`, `compare_version_constraint` and `compare_atom` respecively [#6122 @kit-ty-kate]
```
rjbou, line 216:
```suggestion
  * `OpamFormula`: add some missing comparison functions for `relop`, `version_constraint` and `atom` (`compare_relop`, `compare_version_constraint` and `compare_atom` respectively) [#6122 @kit-ty-kate]

File: src/client/opamClient.ml (hunk 1 above)
Comment (rjbou) [reply src/client/opamClient.ml]: lgtm

**Review by rjbou — APPROVED** [review]
I've added some test from an old branch that I have, and some format nitpicking. Feel free to squash if it's ok for you.
Otherwise, lgtm!

File: src/client/opamClient.ml [hunk 4]
```diff
@@ -1962,62 +1957,109 @@ let check_installed ~build ~post t atoms =
...
+            then (new_map, seen)
+            else loop new_map seen found_conj
+      ) (map, seen) atoms
+  in
+  fst (loop OpamPackage.Name.Map.empty SeenSet.empty atoms)
 
 let assume_built_restrictions ?available_packages t atoms =
-  let missing = check_installed ~build:false ~post:false t atoms in
+  let missing =
+    check_installed ~build:false ~post:false ~recursive:false t atoms
```
rjbou, line 2044: Assume built does not check recursively, is it to not change the known behaviour?

kit-ty-kate: no, the previous behaviour of this function did not check recursively so the behaviour of `assume-built` hasn't changed either

rjbou (approved):
Thanks!

PR #6124, opened 2024-07-30
Add an admin command to compare 2 package versions

mbarbin:
Add a new CLI `opam admin compare-versions` to compare package versions for sanity checks.

This command has 2 modes:

1. By default - user interactive.

In this mode, we provide two versions, and opam outputs a user-friendly expression with the result of the comparison, spelled out. For example:

```sh
 0.0.9 < 0.0.10
```

2. For use in scripts - if needed. Adding the argument `--assert=OP`.

In this mode, the output is suppressed (the command is silent). The result of the command will be encoded in the exit code, to be used in bash conditionals:

exit 0: the comparison holds
exit 1: it doesn't

For example:

```sh
$ opam admin compare-versions 0.0.9 0.0.10 --assert='<'
[0]

$ opam admin compare-versions 0.0.9 0.0.10 --assert='>='
[1]
```

### Original notes

This is meant for quick sanity checks for example.

- Added some basic tests to cover the command. The tests I added do not duplicate the tests for the actual comparison function, rather that meant to cover the basic cases encountered by the command.

- [x] Please update `master_changes.md` file with your changes.

This is following a discussion with @kit-ty-kate in #6118.

Note to reviewers:

I wasn't too sure where to expose the new command. It's not really an admin command, but I couldn't find another suitable section.

This is my first PR on opam and I am not familiar with the code base, as well as some dependencies (e.g. cmdliner). Please feel free to request for as many changes as you like. Thank you!

### DIFF
--- a/src/client/opamAdminCommand.ml
+++ b/src/client/opamAdminCommand.ml
@@ -830,8 +830,8 @@ let check_command cli =
   Term.(const cmd $ global_options cli $ ignore_test_arg $ print_short_arg
         $ installability_arg $ cycles_arg $ obsolete_arg)
 
-let compare_package_versions_command_doc = "Compare 2 package versions"
-let compare_package_versions_command cli =
+let compare_versions_command_doc = "Compare 2 package versions"
+let compare_versions_command cli =
   let version_arg n =
     let doc =
       Arg.info
@@ -840,12 +840,12 @@ let compare_package_versions_command cli =
     in
     Arg.(required & pos n (Arg.some' OpamArg.package_version) None & doc)
   in
-  let command = "compare-package-versions" in
-  let doc = compare_package_versions_command_doc in
+  let command = "compare-versions" in
+  let doc = compare_versions_command_doc in
   let man = [
     `S Manpage.s_description;
     `P "This command compares 2 package versions for quick sanity checks, and prints the result of the comparison to the console. For example:";
-    `I ("For example:", "opam admin compare-package-versions 0.0.9 0.0.10");
+    `I ("For example:", "opam admin compare-versions 0.0.9 0.0.10");
     `I ("outputs:", "0.0.9 < 0.0.10");
     `S Manpage.s_arguments;
     `S Manpage.s_options;
@@ -1259,7 +1259,7 @@ let admin_subcommands cli =
     upgrade_command cli;
     lint_command cli;
     check_command cli;
-    compare_package_versions_command cli;
+    compare_versions_command cli;
     list_command cli;
     filter_command cli;
     add_constraint_command cli;
--- a/src/client/opamAdminCommand.ml
+++ b/src/client/opamAdminCommand.ml
@@ -854,8 +854,7 @@ let compare_package_versions_command cli =
   let cmd global_options v1 v2 () =
     OpamArg.apply_global_options cli global_options;
     let result = OpamPackage.Version.compare v1 v2 in
-    OpamConsole.formatted_msg "%d (%s %s %s)\n"
-      result
+    OpamConsole.formatted_msg "%s %s %s\n"
       (OpamPackage.Version.to_string v1)
       (if result < 0 then "<" else if result = 0 then "=" else ">")
       (OpamPackage.Version.to_string v2);
--- a/src/client/opamAdminCommand.ml
+++ b/src/client/opamAdminCommand.ml
@@ -889,8 +889,8 @@ let compare_versions_command cli =
   let doc = compare_versions_command_doc in
   let man = [
     `S Manpage.s_description;
-    `P "This command compares 2 package versions for quick sanity checks, \
-        and by default prints the result of the comparison to the console. \
+    `P "This command compares 2 package versions for quick sanity checks. \
+        By default it prints the result of the comparison to the console. \
         You may optionally control the exit-code with '--assert=OP'. \
         For example:";
     `Pre "\n\
--- a/src/client/opamAdminCommand.ml
+++ b/src/client/opamAdminCommand.ml
@@ -846,7 +846,7 @@ let compare_package_versions_command cli =
     `S Manpage.s_description;
     `P "This command compares 2 package versions for quick sanity checks, and prints the result of the comparison to the console. For example:";
     `I ("For example:", "opam admin compare-package-versions 0.0.9 0.0.10");
-    `I ("outputs:", "-1 (0.0.9 < 0.0.10)");
+    `I ("outputs:", "0.0.9 < 0.0.10");
     `S Manpage.s_arguments;
     `S Manpage.s_options;
   ]
--- a/src/client/opamAdminCommand.ml
+++ b/src/client/opamAdminCommand.ml
@@ -830,39 +830,11 @@ let check_command cli =
   Term.(const cmd $ global_options cli $ ignore_test_arg $ print_short_arg
         $ installability_arg $ cycles_arg $ obsolete_arg)
 
-module Compare_version_operator = struct
-  type t =
-    [ `Lt
-    | `Let
-    | `Eq
-    | `Gt
-    | `Get
-    ]
-
-  let all = [ `Lt ; `Let ; `Eq ; `Gt ; `Get ]
-
-  let eval : t -> int -> int -> bool = function
-    | `Lt -> ( < )
-    | `Let -> ( <= )
-    | `Eq -> ( = )
-    | `Gt -> ( > )
-    | `Get -> ( >= )
-
-  let assert_ t result = eval t result 0
-
-  let to_string : t -> string = function
-    | `Lt -> "<"
-    | `Let -> "<="
-    | `Eq -> "="
-    | `Gt -> ">"
-    | `Get -> ">="
-end
-
 let compare_versions_command_doc = "Compare 2 package versions"
 let compare_versions_command cli =
   let operators =
-    List.map (fun op -> Compare_version_operator.to_string op, op)
-      Compare_version_operator.all
+    List.map (fun op -> OpamFormula.string_of_relop op, op)
+      OpamFormula.all_relop
   in
   let assert_result =
     let doc =
@@ -908,16 +880,16 @@ let compare_versions_command cli =
   in
   let cmd global_options v1 v2 assert_result () =
     OpamArg.apply_global_options cli global_options;
-    let result = OpamPackage.Version.compare v1 v2 in
     match assert_result with
     | None ->
+      let result = OpamPackage.Version.compare v1 v2 in
       OpamConsole.formatted_msg "%s %s %s\n"
         (OpamPackage.Version.to_string v1)
         (if result < 0 then "<" else if result = 0 then "=" else ">")
         (OpamPackage.Version.to_string v2)
     | Some op ->
       OpamStd.Sys.exit_because
-        (if Compare_version_operator.assert_ op result
+        (if OpamFormula.eval_relop op v1 v2
          then `Success
          else `False)
   in
--- a/src/client/opamAdminCommand.ml
+++ b/src/client/opamAdminCommand.ml
@@ -830,8 +830,54 @@ let check_command cli =
   Term.(const cmd $ global_options cli $ ignore_test_arg $ print_short_arg
         $ installability_arg $ cycles_arg $ obsolete_arg)
 
+module Compare_version_operator = struct
+  type t =
+    [ `Lt
+    | `Let
+    | `Eq
+    | `Gt
+    | `Get
+    ]
+
+  let all = [ `Lt ; `Let ; `Eq ; `Gt ; `Get ]
+
+  let eval : t -> int -> int -> bool = function
+    | `Lt -> ( < )
+    | `Let -> ( <= )
+    | `Eq -> ( = )
+    | `Gt -> ( > )
+    | `Get -> ( >= )
+
+  let assert_ t result = eval t result 0
+
+  let to_string : t -> string = function
+    | `Lt -> "<"
+    | `Let -> "<="
+    | `Eq -> "="
+    | `Gt -> ">"
+    | `Get -> ">="
+end
+
 let compare_versions_command_doc = "Compare 2 package versions"
 let compare_versions_command cli =
+  let operators =
+    List.map (fun op -> Compare_version_operator.to_string op, op)
+      Compare_version_operator.all
+  in
+  let assert_result =
+    let doc =
+      Arg.info
+        ~docv:"OP"
+        ~doc:(Printf.sprintf
+             "When supplied, the output is suppressed and the result of the \
+              comparison is checked againts the provided operator. The command exits 0 \
+              if the comparison holds, and 1 otherwise. \
+              $(docv) must be %s.\n" (Arg.doc_alts_enum ~quoted:true operators))
+        [ "assert" ]
+    in
+    let op_conv = Arg.enum operators in
+    Arg.(value & opt (Arg.some' op_conv) None doc)
+  in
   let version_arg n =
     let doc =
       Arg.info
@@ -844,23 +890,40 @@ let compare_versions_command cli =
   let doc = compare_versions_command_doc in
   let man = [
     `S Manpage.s_description;
-    `P "This command compares 2 package versions for quick sanity checks, and prints the result of the comparison to the console. For example:";
-    `I ("For example:", "opam admin compare-versions 0.0.9 0.0.10");
-    `I ("outputs:", "0.0.9 < 0.0.10");
+    `P "This command compares 2 package versions for quick sanity checks, and by default prints the result of the comparison to the console. You may optionally control the exit-code with '--assert=OP'. For example:";
+    `Pre "\n\
+       \\$ opam admin compare-versions 0.0.9 0.0.10\n\
+       0.0.9 < 0.0.10\n\
+       \n\
+       \\$ opam admin compare-versions 0.0.9 0.0.10 --assert '<'\n\
+       [0]\n\
+       \n\
+       \\$ opam admin compare-versions 0.0.9 0.0.10 --assert '>='\n\
+       [1]";
     `S Manpage.s_arguments;
     `S Manpage.s_options;
   ]
   in
-  let cmd global_options v1 v2 () =
+  let cmd global_options v1 v2 assert_result () =
     OpamArg.apply_global_options cli global_options;
     let result = OpamPackage.Version.compare v1 v2 in
-    OpamConsole.formatted_msg "%s %s %s\n"
-      (OpamPackage.Version.to_string v1)
-      (if result < 0 then "<" else if result = 0 then "=" else ">")
-      (OpamPackage.Version.to_string v2);
+    if Option.is_none assert_result then
+      OpamConsole.formatted_msg "%s %s %s\n"
+        (OpamPackage.Version.to_string v1)
+        (if result < 0 then "<" else if result = 0 then "=" else ">")
+        (OpamPackage.Version.to_string v2);
+    let exit_reason =
+      match assert_result with
+      | None -> `Success
+      | Some op ->
+        if Compare_version_operator.assert_ op result
+        then `Success
+        else `False
+    in
+    OpamStd.Sys.exit_because exit_reason
   in
   OpamArg.mk_command  ~cli OpamArg.cli_original command ~doc ~man
-  Term.(const cmd $ global_options cli $ version_arg 0 $ version_arg 1)
+  Term.(const cmd $ global_options cli $ version_arg 0 $ version_arg 1 $ assert_result)
 
 let pattern_list_arg =
   OpamArg.arg_list "PATTERNS"
--- a/src/client/opamAdminCommand.ml
+++ b/src/client/opamAdminCommand.ml
@@ -869,10 +869,10 @@ let compare_versions_command cli =
       Arg.info
         ~docv:"OP"
         ~doc:(Printf.sprintf
-             "When supplied, the output is suppressed and the result of the \
-              comparison is checked againts the provided operator. The command exits 0 \
-              if the comparison holds, and 1 otherwise. \
-              $(docv) must be %s.\n" (Arg.doc_alts_enum ~quoted:true operators))
+                "When supplied, the output is suppressed and the result of \
+                 the comparison is checked againts the provided operator. \
+                 The command exits 0 if the comparison holds, and 1 otherwise. \
+                 $(docv) must be %s.\n" (Arg.doc_alts_enum ~quoted:true operators))
         [ "assert" ]
     in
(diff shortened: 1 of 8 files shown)


src/client/opamAdminCommand.ml
```diff
@@ -830,6 +830,39 @@ let check_command cli =
...
+    `I ("outputs:", "-1 (0.0.9 < 0.0.10)");
+    `S Manpage.s_arguments;
+    `S Manpage.s_options;
+  ]
+  in
+  let cmd global_options v1 v2 () =
+    OpamArg.apply_global_options cli global_options;
+    let result = OpamPackage.Version.compare v1 v2 in
+    OpamConsole.formatted_msg "%d (%s %s %s)\n"
+      result
```
kit-ty-kate, line 858: what is the goal of showing the raw int in the message? Is it to simplify parsing if someone were to use this command in a script? If so `cut '-d ' -f2` would already give that information
```suggestion
    OpamConsole.formatted_msg "%s %s %s\n"
```

mbarbin: Hi @kit-ty-kate, and thank you for starting the review already!

This is reminiscent of a distant (and perhaps obsolete) past where some `compare` functions would return non-normalized results, and I wasn't sure whether the internal int would be of interest here.

Your question prompted me to look at `OpamVersionCompare` more closely, and indeed it uses a `normalize_comp_result` step.

Given this, I'm happy to remove the int from the output.

By the way, I appreciate the use of `Ordering`-like modules nowadays and feel enthusiastic about seeing it integrated into more projects over time (perhaps a good candidate for an addition to `OpamStd` later down the road?).

mbarbin: Sorry I had accepted your commit suggestion through the GitHub interface as a separate commit but ended up amending it by accident. I guess you'll squash all commits at the end?

avsm:
I recommend following the design of `dpkg --compare-versions` instead, where the `op` can also be specified. The return value is then a simple true (0) or false (non-zero).

https://manpages.ubuntu.com/manpages/oracular/en/man1/dpkg.1.html

```
      --compare-versions ver1 op ver2
           Compare version numbers, where op is a binary operator.  dpkg returns true (0) if the
           specified condition is satisfied, and false (1) otherwise.  There are two groups of
           operators, which differ in how they treat an empty ver1 or ver2.  These treat an empty
           version as earlier than any version: lt le eq ne ge gt.  These treat an empty version
           as later than any version: lt-nl le-nl ge-nl gt-nl.  These are provided only for
           compatibility with control file syntax: < << <= = >= >> >.  The < and > operators are
           obsolete and should not be used, due to confusing semantics.  To illustrate: 0.1 < 0.1
           evaluates to true.
```

mbarbin:
Hello @avsm so nice to have you join the conversation!

Thank you for pointing out that command, I'll have a look. At quick glance, this can suggest that this new opam command could be named `compare-versions` (and the "-package-" middle part doesn't add much).

About the design I'd be curious what usage do you have in mind. I sort of see the appeal to re-using existing designs, but in this case, I feel:

1. dpkg sounds like it is bound by some historical deprecation constraints, which would be nice not to be tied to in opam from the get-go, unless there are applicable reasons
2. dpkg offers an interface which I believe, due to the 3 cases of the expected results, requires at least 2 calls to determine the outcome with no ambiguity, if the result of your query returns `false` (for example, if you are surprised that a strict-less-than doesn't hold, you still don't know whether you have an equal, or strictly-greater-than). Having the command return a result equivalent to an ordering.i just seems more simple, doesn't it? As a mere mortal, it's also friendly to be shown the plain ordering expression spelled out (I can tell it is going to take me some time to completely mentally digest that --compare-versions spec !).

The part about empty versions of dpkg just sounds very confusing to me. Is this due to a command parsing technicality, or are there really support for empty package versions out there? Does opam support empty package versions too?

mbarbin:
Adding a note about empty versions:

> The part about empty versions of dpkg just sounds very confusing to me. Is this due to a command parsing technicality, or are there really support for empty package versions out there? Does opam support empty package versions too?

As of the current tip of this PR:

```bash
$ ./_build/default/src/client/opamMain.exe admin compare-package-versions '' '1'
opam admin: VERSION argument: Package version can't be empty
Usage: opam admin compare-package-versions [OPTION]… VERSION VERSION
Try 'opam admin compare-package-versions --help' or 'opam admin --help' for more information.
```

So if we wanted to support empty version, some more changes would be required.

avsm:
I have no strong opinion, but outputting shell operators like `>` or `<` is usually the worst possible interface to a shell interface where escaping is a dark art at the best of times ;-)

Two other options that are more "unix-like":
- unix commands usually exit with `0` for success and non-zero for inequality, so you could exit with `-1` or `1` to indicate the results of the comparison
- you could take a list of versions on the command line argument and return them in ascending sorted order on stdout. That plays well with xargs/IFS style manipulations from shell scripts.

mbarbin:
@avsm Thanks a lot for this input, this is really helpful. I am going to explore ideas base on your suggestions, reverting the PR as a draft for now.

kit-ty-kate:
I don't personally see much value in a ""unix-style"" command. If people wanted that it'd be easier for them to just create a dedicated program using the `opam-core` library. opam admin commands are not the kind of command meant to be used as part of a script in this way. The current state of the PR looked fine by me

mbarbin:
@kit-ty-kate I think there could be a middle ground: if we add an optional `--assert (<,>,=,...)` command parameter to the state of this RP, with the understanding that when supplied, it affects the exit code accordingly, I think we combine the benefits of the approaches. Do you like it?

How about the name 'compare-versions' vs 'compare-package-versions' ? I have renamed locally, let me know if I should push this commit or not.

dbuenzli:
Not following the discussion but just note that `exit (-1)` is [going to](https://pubs.opengroup.org/onlinepubs/007904875/functions/exit.html) be 255 in your shell.

kit-ty-kate:
> How are they meant to be used, then?

As mentioned in https://github.com/ocaml/opam/issues/6118 this would be used as part of a double-check for packages unfamiliar with opam/debian version ordering, so always interactively.

What use-case would there be in an automated context?

avsm:
I use something similar in my personal opam repo (full of my homebrew shell scripts) to find the latest package revision from a directory to choose which one to install (just from looking at the opam files; there is no initialised opam here for various reasons yet and so no solver).  And more generally, I've used almost every command exposed by opam at some point in a shell script while doing something-or-other with the repositories -- it's just how CLIs are used!

I'm not saying there's anything _wrong_ with the design proposed in the original PR, except for a preference to not output known-shell-fragments (but even that's probably fine these days). But the ability to sort a list of package names+versions would definitely be something I find useful -- I've written OpamVersionCompare fragments for that several times in different contexts. Thinking about @dbuenzli's comment a bit more -- we never have a meaningful case where two versions are the same opam versions but lexically different, do we?  So the ordering logic just needs to sort in ascending order, and then equal versions would be distinguishable by the fact they're the same string. No need for exit codes then.

kit-ty-kate:
> we never have a meaningful case where two versions are the same opam versions but lexically different, do we?

we do. For example: `1.00 = 1.0`

avsm:
Bah, good example! I thought those were different, but what I was thinking of was the common mistake of "1.0.0" <> "1.0" in package formula.  So checking for equality is significant too, so it's not enough to just output a sorted list of package versions.

mbarbin:
I added a few commit to explore having a mode where the command can exit with 0 or 1, based on the OP input from @avsm. This is simply to add more context to the conversation. #6128 
This let you preview also what the change looks like with the shorter name `compare-versions`.

mbarbin:
I updated the PR based on the previous comments and will proceed to re-opening for another round of review. Thanks!

mbarbin:
Hello. I have rebased the PR.

I wanted to add a note about the addition of the `--assert` flag. After considering @avsm's input, I later imagined a reasonable use case, which I am including below.

This use case involves a setup using a `release.sh` script to release opam packages. I could imagine adding a new section to perform a sanity check to prevent backward releases, like this:
```
#!/bin/sh -e

# Some initial steps to extract the versions
OPAM='./_build/default/src/client/opamMain.exe'
PREVIOUS_VERSION=$1
NEW_VERSION=$2
IS_POINT_RELEASE=${3:-false} # This variable indicates if we are doing a point release
# ...

echo "Checking version ordering ..."

if [ "$IS_POINT_RELEASE" = false ]; then
  if ! $OPAM admin compare-versions $PREVIOUS_VERSION --assert '<' $NEW_VERSION; then
    echo "Unexpected version ordering - aborting release"
    exit 1
  fi
fi

echo "Carrying on with the release script ..."
# ...
```

--

J'ai pris la liberté, cher lecteur, d'orner le présent travail d'une petite chansonnette - elle m'a accompagné en travaillant sur ce programme.

Le bon roi Dagobert
A fait sa release à l'envers ;
Le grand saint Éloi
Lui dit: Ô mon roi ! Et
Le script de Votre Majesté
Est si faiblement typé.
C'est vrai, lui dit le roi,
je vais remettre tout ça à l'endroit.

src/client/opamAdminCommand.ml
```diff
@@ -830,6 +830,100 @@ let check_command cli =
...
+  let assert_ t result = eval t result 0
+
+  let to_string : t -> string = function
+    | `Lt -> "<"
+    | `Let -> "<="
+    | `Eq -> "="
+    | `Gt -> ">"
+    | `Get -> ">="
+end
+
```
kit-ty-kate, line 860: The operator is already defined by `Cudf_types.relop` and mirrored in `OpamParserTypes.FullPos.relop_kind`, `OpamTypes` and `OpamFormula`. I think those operation should probably be defined in `OpamFormula` instead

src/client/opamAdminCommand.ml
```diff
@@ -830,6 +830,100 @@ let check_command cli =
...
+        ~doc:"Package version to compare" []
+    in
+    Arg.(required & pos n (some OpamArg.package_version) None & doc)
+  in
+  let command = "compare-versions" in
+  let doc = compare_versions_command_doc in
+  let man = [
+    `S Manpage.s_description;
+    `P "This command compares 2 package versions for quick sanity checks, \
+        and by default prints the result of the comparison to the console. \
```
kit-ty-kate, line 893:
```suggestion
    `P "This command compares 2 package versions for quick sanity checks. \
        By default it prints the result of the comparison to the console. \
```

kit-ty-kate (commented):
otherwise that looks good to me

src/client/opamAdminCommand.ml
```diff
@@ -830,6 +830,100 @@ let check_command cli =
...
+  let all = [ `Lt ; `Let ; `Eq ; `Gt ; `Get ]
+
+  let eval : t -> int -> int -> bool = function
+    | `Lt -> ( < )
+    | `Let -> ( <= )
+    | `Eq -> ( = )
+    | `Gt -> ( > )
+    | `Get -> ( >= )
+
+  let assert_ t result = eval t result 0
```
kit-ty-kate, line 851: actually this section is already defined in `OpamFormula.check_relop`

mbarbin: Great, thanks! I updated the code to use the relops instead of redefining them.

src/client/opamAdminCommand.ml
mbarbin: Done. Thanks!

mbarbin:
Thanks a lot @kit-ty-kate for the review and for pointing me to the existing comparison operators. I updated as you suggested.

kit-ty-kate (approved):

dra27:
Slightly late to the party, but especially given the subsequent perfectly reasonable example:

```sh
  if ! $OPAM admin compare-versions $PREVIOUS_VERSION --assert '<' $NEW_VERSION; then
```

I'm left wondering why we've ended up with the `--assert` and not just gone with the 2 argument and 3 argument versions @avsm originally proposed, especially given that we have also ended up using operators that do need quoting anyway.

(I'd prefer to have the feature than not, though!)

kit-ty-kate:
I'm ok with having both a 2 arguments and a 3 arguments version of that command

mbarbin:
Hi @dra27 and thank you for joining the party!

I may have missed the specific proposal you are referring to. To make sure I understand it, in this proposal, the arguments are all positional, and depending on the number of arguments (2 or 3), you select the mode? Such as:

```sh
 0.0.9 < 0.0.10

$ opam admin compare-versions 0.0.9 <= 0.0.10
[0]

$ opam admin compare-versions 0.0.9 >= 0.0.10
[1]
```

Like this?

I would have to check a bit more what this entails to in `cmdliner`. My vague impression was that positional argument must be typed "statically" according to their position in the command line, but perhaps parsing them all as strings and deferring until the command handler should allow to make something like this work. For this reason, I tend to prefer the named `--assert` version. What do you think?

If you are thinking about something else entirely, please let me know!

mbarbin:
@dra27 thinking a bit more about the modes, I should add that it didn't even occur to me (until now) that we could design a syntax that makes use of the infix operators placed in postfix position. I'm inclined to think that it decreases somewhat the readability, but that may just be a personal preference.

That is, I think I prefer:

```sh
$ opam admin compare-versions 0.0.9 --assert '<=' 0.0.10
```

over:

```sh
$ opam admin compare-versions 0.0.9 0.0.10 <=
```

But I can probably get over it if that's what you're thinking about ! :-)

kit-ty-kate:
> @kit-ty-kate I am attempting to rebase this PR and am encountering a build issue, related to vendored opam-0install-cudf lib:
> 
> How do you update vendored libs in the opam repository? Is this something I have to do locally as well (such as a git submodule sync or smth), or are the updates committed and pushed to the repo?

It all depends on your setup. If you've used opam to install the dependencies then `opam update && opam upgrade opam-0install-cudf` will do. If you've used the vendored method (aka. `./configure --with-vendored-deps` or `make cold`) then doing another `./configure --with-vendored-deps` or `make cold` (depending on what you used) would fix your issue as well

kit-ty-kate:
> over:
> 
> ```shell
> $ opam admin compare-versions 0.0.9 0.0.10 <=
> ```

what about:
```
opam admin compare-versions 0.0.9 '<=' 0.0.10
```

dra27:
Definitely not proposing having them in postfix position, no! As in https://github.com/ocaml/opam/pull/6124#issuecomment-2258285695 and https://github.com/ocaml/opam/pull/6124#issuecomment-2329149715, with the operator as an infix.

I think cmdliner can just give a list of positional arguments or at least regard one of three as optional - so still having the operator-less version you originally proposed (displaying the result) but also having the scripted version

mbarbin:
> I think cmdliner can just give a list of positional arguments or at least regard one of three as optional - so still having the operator-less version you originally proposed (displaying the result) but also having the scripted version

That sounds interesting to me. I thought this would be rejected (doesn't it open the door for specs that quickly become ambiguous?). I want to explore this a bit more, in particular see what happens when you "squeeze" optional positional args in the middle of required positional args. I created https://github.com/mbarbin/cmdlang/issues/6 to track.

Short of that:

1. matching on the number of args equates to the power of a monad, so that's not standard I suppose.

```ocaml
match%bind Arg.count_positional with
| 2 ->
  let%map v1 = pos 0 version
  and v2 = pos 1 version in
  `Compute (v1, v2)
| 3 ->
  let%map v1 = pos 0 version
  and op = pos 1 operator
  and v2 = pos 2 version in
  `Assert (v1, op, v2)
```

2. Doing something more dynamic with a `pos_all` and parsing arguments in the body of the command is going to cut down on the ability for `cmdliner` to generate a precise usage page automatically. It is not obvious to me how to estimate the overall gain then, compared to `--assert`.

As a side note, selecting the mode of a multi-mode commands with a named arguments feels more natural to me rather than by the presence/absence of positional arguments, but I don't think this argument is tenable, as there are probably countless counter-examples in the Unix literature. Plus, @dra27 's input seem to convey the opposite preference, at least in this instance, so I guess I'm happy not to account for this personal preference here.

I'm sort of stuck there. Is there another combinator I am not thinking of from `cmdliner`?

kit-ty-kate:
> I'm sort of stuck there. Is there another combinator I am not thinking of from `cmdliner`?

Isn't that possible with 3 `Cmdliner.Arg.pos`, where the first and third one are given to `Cmdliner.Arg.required` and the second is given to `Cmdliner.Arg.value` ?

mbarbin:
> Isn't that possible with 3 Cmdliner.Arg.pos, where the first and third one are given to Cmdliner.Arg.required and the second is given to Cmdliner.Arg.value ?

That's what I want to explore a bit more. I am used to hitting [this limitation](https://github.com/janestreet/core/blob/6d8a862aacdbce87eeca0ac3d942001d546d4845/command/src/command.ml#L1017) when attempting this sort of things.

Let me play a bit with it and I'll get back to you. Thanks for your help!

kit-ty-kate:
Mmh, yeah it looks like cmdliner doesn't support that given positional arguments are absolute and do not change depending on optional parameters
```ocaml
let test arg1 arg2 arg3 =
  Printf.printf "arg1 = %s\narg2 = %s\narg3 = %s\n"
    arg1 (match arg2 with None -> "None" | Some x -> "Some "^x) arg3

module Arg = Cmdliner.Arg
module Cmd = Cmdliner.Cmd
module Term = Cmdliner.Term

let arg1 = Arg.(required & pos 0 (some string) None & info [])
let arg2 = Arg.(value & pos 1 (some string) None & info [])
let arg3 = Arg.(required & pos 2 (some string) None & info [])

let cmd =
  let info = Cmd.info "test" ~version:"dev" in
  Cmd.v info Term.(const test $ arg1 $ arg2 $ arg3)

let main () = Stdlib.exit (Cmd.eval cmd)
let () = main ()
```
```
$ ./a.out a b c
arg1 = a
arg2 = Some b
arg3 = c
$ ./a.out a c
test: a required argument is missing
Usage: test [OPTION]… ARG [ARG] ARG
Try 'test --help' for more information.
```
let's keep `--assert`. I'll open a ticket in cmdliner in case upstream wants to do have variable positional arguments support in the future if it's ever useful for anyone else.

dbuenzli:
Honestly I think you are all a bit overthinking that :-) Simply ask for three positional arguments make the last one optional, resolve that in code with a proper match and write the `SYNOPSIS` section yourself.

mbarbin:
> Honestly I think you are all a bit overthinking that :-) Simply ask for three positional arguments make the last one optional, resolve that in code with a proper match and write the SYNOPSIS section yourself.

Thanks for the input @dbuenzli!

Personally, I wouldn't know how to defend doing that, versus keeping withing a well supported use case and keeping the `--assert` flag. This sort of feels arbitrary to me.

> I'll open a ticket in cmdliner in case upstream wants to do have variable positional arguments support in the future if it's ever useful for anyone else.

My inclination is for `cmdliner` to align with the `core.command` limitation and cleanly reject these kinds of specs. I also don't intend to support this use case in `cmdlang` and I doubt `climate` will support it either.

dbuenzli:
Also I often find it that when people try to do subtle things with cli, it's often not in the interest of the end-users.

In this particular case, why not simply always require the operator ? I don't think having a default operator brings much here except that I will have to remember it.

dbuenzli:
> let's keep `--assert`. I'll open a ticket in cmdliner in case upstream wants to do have variable positional arguments support in the future if it's ever useful for anyone else.

I closed it. The problem is that in general this leads to ambiguous parses. There are already enough of these to confuse users, let's not add more.

mbarbin:
> In this particular case, why not simply always require the operator ? I don't think having a default operator brings much here except that I will have to remember it.

Are you proposing to add a syntax for a special "don't know" operator? Like:

```bash
$ opam admin compare-versions 0.0.9 '?' 0.0.10
0.0.9 < 0.0.10
```

mbarbin:
> The operators <, =, <=, with exit 0 if the assertion holds and 1 if it does not is all you need.

If the command also prints on stderr the corrected inequality, spelled out, in case of an error, I'd be OK with this. This is very close to my initial motivating use case.

```bash
$ opam admin compare-versions 0.33.0-preview.1 '<' 0.33.0
0.33.0-preview.1 > 0.33.0
[1]
```

mbarbin:
@kit-ty-kate what do you think? I pushed a preview [here](https://github.com/mbarbin/opam/tree/compare-versions-requires-operator) (keeping the existing operators).

dbuenzli:
> If the command also prints on stderr the corrected inequality, spelled out, in case of an error, I'd be OK with this.

I would find this a rather confusing cli. Why not simply output on stdout `true` or `false` which a `--quiet` will suppress.

dbuenzli:
> As a reminder, I was trying to introduce a command that allows comparing versions from the command line, for which you are not sure `a priori` of the result

Mmh sorry I misunderstood what you were trying to provide because I inferred it from the command name you proposed, namely `compare-versions`. Personally I think the command name you proposed is not a good name for this. 

I would rather suggest `order-versions`. I think having both is fine and useful. Basically:

```
admin compare-versions V0 OP V1  # Assert the equation both on stdout and with exit
admin order-versions V0 V1       # Output V0 OP V1 with OP such that the equation holds
```
Note that `order-versions` should never fail (unless `V0` and `V1` can't be ordered or parsed into versions).

dbuenzli:
> I suppose it depends on the perspective you adopt

Sure, I'm not attached to the names I suggested, I just needed another name to make the point about not cramming the two behaviours in the same command.

mbarbin:
I created a new draft trying to account for the various feedback received here. Please see : https://github.com/ocaml/opam/pull/6196

Thanks!

mbarbin:
If we prefer non-quoted operators, there's also the possibility of using the operator names as flags (rather than "--assert") and being back to having 1 single command. I created a version based on in https://github.com/ocaml/opam/pull/6197 as an additional data point.

I'll have some time to look again into this PR and the end of next week. Thanks a lot!

kit-ty-kate:
Looking at all 3 PRs i think i prefer https://github.com/ocaml/opam/pull/6197 (`opam admin compare-versions V1 [--eq|--lt|...] V2`) so far, but we'll talk about it at the next meeting on monday so it's not a final answer or anything yet

mbarbin:
> Looking at all 3 PRs i think i prefer #6197 

In retrospect that's the one I prefer as well so far. Thank you for following up and letting me know about your meeting. I'll be ready to help moving things forward as you see fit. Thanks a lot!

kit-ty-kate:
Dev meeting note: https://github.com/ocaml/opam/pull/6197 looks preferable for all 3 of us

mbarbin:
> Dev meeting note: after discussion, #6197 looks preferable for all 3 of us (Raja, David, Kate)

Great! I'll close the other ones, and do a little pass at it, so we can do a first round of review. Thank you!

PR #6125, opened 2024-07-30
Fix regression: deps-only setting direct dependencies as root packages by default

rjbou:
It was added in #4564, accidentally removed in #4975.

Backported in 2.2 in #6126

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -2301,21 +2301,34 @@ let install_t t ?ask ?(ignore_conflicts=false) ?(depext_only=false)
         (OpamSolution.install_depexts ~force_depext:true ~confirm:false t
            (OpamSolver.all_packages solution)), None
       else
-      let add_roots =
-        OpamStd.Option.map (function
-            | true ->
-              OpamPackage.Name.Set.union requested @@
-              OpamPackage.Name.Set.of_list (List.rev_map fst deps_atoms)
-            | false -> OpamPackage.Name.Set.empty)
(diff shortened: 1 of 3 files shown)


src/client/opamClient.ml
```diff
@@ -2301,21 +2301,35 @@ let install_t t ?ask ?(ignore_conflicts=false) ?(depext_only=false)
...
+          match add_to_roots, deps_only with
+          | (Some true | None), true ->
+            let pkgs_solution = OpamSolver.all_packages solution in
+            let requested_deps =
+              OpamPackage.Set.fold (fun nv acc ->
+                  OpamFormula.ors [
+                    OpamPackageVar.all_depends t (OpamSwitchState.opam t nv)
+                      ~depopts:false ~build:true ~post:false;
+                    acc
+                  ])
```
kit-ty-kate, line 2314: No need to call `ors` with only 2 parameters as it will always reduce to `Or (x.(1), x.(2))`
```suggestion
                  OpamFormula.Or (
                    OpamPackageVar.all_depends t (OpamSwitchState.opam t nv)
                      ~depopts:false ~build:true ~post:false,
                    acc))
```

kit-ty-kate (approved):
lgtm otherwise

src/client/opamClient.ml
rjbou: true

kit-ty-kate:
LGTM'd by @AltGr 

Thanks!

PR #6127, opened 2024-07-30
ocp-indent: make sure 'begin match' doesn't add an extra indentation

kit-ty-kate:
Makes the indentation style switch from:
```ocaml
let () =
  begin match x with
    | Some _ -> ()
    | None -> ()
  end
```
to the more common:
```ocaml
let () =
  begin match x with
  | Some _ -> ()
  | None -> ()
  end
```
Some of our use of `begin match` already use this indentation style manually

### DIFF
--- a/.ocp-indent
+++ b/.ocp-indent
@@ -1,1 +1,2 @@
 normal
+strict_with=auto


rjbou (approved):
Thanks! There is not so much `begin match` (and I personally not in `begin` coding style), but for the few ones that are in opam codebase, it makes sense!

rjbou:
@dra27, an opinion?

dra27 (approved):
LGTM! For me, the rationale is that if we actually had delimited `match`, then we'd happily put:

```ocaml
match x with
| `Foo -> ()
| `Bar -> ()
end
```
(with no indentation)

PR #6130, opened 2024-07-30
Add support for opam-0install-cudf 0.5.0

kit-ty-kate:
Requires https://github.com/ocaml-opam/opam-0install-cudf/pull/2
Requires https://github.com/ocaml-opam/opam-0install-cudf/pull/3
Supersedes https://github.com/ocaml/opam/pull/5879

This PR also adds support for unordered criteria with 0install, as well as `-changed` and `-count[avoid-version,solution]` to make the solver prefer to keep packages installed at their current version and avoid packages marked with `avoid-version` respectively.


src/solver/opamBuiltin0install.ml
```diff
@@ -90,25 +91,53 @@ let reconstruct_universe universe selections =
...
-    default
+  let default =
+    {
+      drop_installed_packages = false;
+      prefer_oldest = false;
+      handle_avoid_version = false;
+      prefer_installed = false;
+    }
+  in
+  let rec parse default = function
```
rjbou, line 107: `OpamBuiltinZ3` contains a `Syntax` module that parses criteria. We can move it into `OpamCudfSolverSig` and use it directly instead of having a parsing function per solver.

kit-ty-kate: done. I don't think this should be in `OpamCudfSolverSig` as that module is meant for signatures not implementations but it fits really well into a new `OpamCudfCriteria` module

rjbou (approved):
I've added the new module to index.html (it would have been a good test for the doc job!) and some comments in the nnew module mli.
lgtm, i'll let you check the new additions.
thanks!

Issue #6131, opened 2024-07-31
failure to fetch git submodules are not errors


kit-ty-kate:
https://github.com/ocaml/opam/commit/c030dcbff6aada1c8fa32e868268d4eda676d0c7 (in opam 2.0) added support for git submodules when pulling a source repository.
However currently failures while fetching the submodules are not errors but only displays a warning. I think this should be an error instead, as not stopping early can change the behaviour of building a such a package drastically.

hhugo:
related to https://github.com/ocaml/opam/issues/6021

kit-ty-kate:
good point, although in your case Windows' `MAX_PATH` is the real culprit but indeed failing early would show a better error message for sure

PR #6132, opened 2024-07-31
Fail when a git submodule fails to update instead of ignoring the error

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/6131

### DIFF
--- a/src/repository/opamGit.ml
+++ b/src/repository/opamGit.ml
@@ -179,9 +179,7 @@ module VCS : OpamVCS.VCS = struct
       if OpamFilename.exists (repo_root // ".gitmodules") then
         git repo_root [ "submodule"; "update"; "--init"; "--recursive" ]
         @@> fun r ->
-        if OpamProcess.is_failure r then
-          OpamConsole.warning "Git submodule update failed in %s"
-            (OpamFilename.Dir.to_string repo_root);
+        OpamSystem.raise_on_process_error r;
         Done ()
       else Done ()
 
--- a/tests/reftests/git.test
+++ b/tests/reftests/git.test
@@ -51,24 +51,22 @@ The following actions will be performed:
 <><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
 -> removed   submodule.1
 Done.
-### opam install submodule
+### opam install submodule | '".+[/\\]git(\.exe)? ' -> '"git '
 The following actions will be performed:
 === install 1 package
   - install submodule 1
 
 <><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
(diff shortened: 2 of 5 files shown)


src/repository/opamGit.ml
```diff
@@ -216,7 +214,12 @@ module VCS : OpamVCS.VCS = struct
                    @ List.map OpamFilename.SubPath.to_string
                      (OpamStd.Option.to_list subpath))
     @@> function
-    | { OpamProcess.r_code = 0; _ } -> Done true
+    | { OpamProcess.r_code = 0; _ } ->
+      git repo_root ["submodule"; "status"; "--recursive"] @@> fun r ->
+      if r.r_code = 0 &&
+         List.for_all (fun s -> String.length s > 0 && s.[0] = ' ') r.r_stdout
```
dra27, line 220: I haven't checked the full context of how this function is used, but is an uninitialised submodule here possible - is this used for pinning?

rjbou, line 220: It worth a comment here, to explain the whitespace check that is done. The manpage states
```quote
Each SHA-1 will possibly be prefixed with - if the submodule is not initialized,
+ if the currently checked out submodule commit does not match the SHA-1 found in
the index of the containing repository and U if the submodule has merge conflicts.
```
which explains why we require only first character to be a whitespace.

rjbou: It is called transitively from each `OpamRepository.pull_*` functions.

rjbou (changes requested):
On the code itself, lgtm!
Some remarks:
* The title of the PR doesn't fit the content, the second commit it more submodules handling than the failure itself. It needs to be updated by something more general, like "git submodules enhancements".
* Add a test to highlight the modifications, and follow them

kit-ty-kate:
I've split off the second commit in https://github.com/ocaml/opam/pull/6153 and added a test

rjbou (approved):
Thanks!

tests/reftests/git.test
```diff
@@ -0,0 +1,72 @@
...
+### git -C ./submodule config core.autocrlf false
+### touch ./submodule/some-file
+### git -C ./submodule add ./some-file
+### git -C ./submodule commit -qm "first commit"
+### mkdir use-submodule
+### git -C ./use-submodule init -q
+### git -C ./use-submodule config core.autocrlf false
+### ### This test may fail locally depending on your git version
+### ### File protocol is removed locally since git 2.38,
+### ### to fix a CVE: https://www.cve.org/CVERecord?id=CVE-2022-39253
```
rjbou, line 14: I added that note

kit-ty-kate: That note isn't quite true. Only a couple of versions have file protocol disabled. More recent git versions work just fine

kit-ty-kate: Ah nevermind, what i'm saying is not true. My memory failed me.

Issue #5567, opened 2023-06-02
opam install . --deps-only --working-dir incorrectly reports "Nothing to do"


jonsterling:
As far as I can tell, the `--working-dir` option of `opam` is not functional. I cannot tell if this is my own misunderstanding of what `--working-dir` or `--deps-only` are supposed to do, but this has happened often enough for me that I thought I would report it as a bug.

### What happens
I have a program I am developing, and I have added a new dependency to my `.opam` file. In this case, I added a dependency on `ptime`. I now wish to bring my current switch into alignment with the dependencies specified by my project. Note that I have not committed this change in `git`. Therefore, I run the command `opam install . --deps-only --working-dir`, which the documentation has led me to believe will do the following: it should install the dependencies of the package specified in the current directory, and it should read the version of the `.opam` file that is currently in the working directory rather than the one that is committed to version control.

**Here is what actually happens:**

```
$ opam install . --deps-only --working-dir
Nothing to do.
```

I believe this behavior is either wrong or highly unintuitive; in the latter case, maybe we could come up with something to add to the documentation to clarify the intended function of these options.

### What if I commit the change and don't use `--working-dir`?

When I commit the change to my `.opam` file and run `opam install . --deps-only`, then `opam` correctly offers to install `ptime`.

#### My `.opam` file

```
# This file is generated by dune, edit dune-project instead
opam-version: "2.0"
version: "2.1"
synopsis: "A tool for tending mathematical forests"
maintainer: ["Jonathan Sterling"]
authors: ["Jonathan Sterling"]
license: "GPL-3.0-or-later"
homepage: "https://github.com/jonsterling/ocaml-forester"
bug-reports: "https://github.com/jonsterling/ocaml-forester/issues"
depends: [
  "menhir"
  "ocamlgraph"
  "ocaml"
  "dune" {>= "3.7"}
  "xmlm"
  "ppx_deriving"
  "domainslib"
  "ptime"
  "yuujinchou" {>= "5.0.1"}
  "algaeff"
  "odoc" {with-doc}
]
build: [
  ["dune" "subst"] {dev}
  [
    "dune"
    "build"
    "-p"
    name
    "-j"
    jobs
    "@install"
    "@runtest" {with-test}
    "@doc" {with-doc}
  ]
]
dev-repo: "git+https://github.com/jonsterling/ocaml-forester.git"
```

You can also find the source repository here if you wish to reproduce on your own: https://github.com/jonsterling/ocaml-forester

[...]

PR #6135, opened 2024-07-31
Reftest: add more exhaustive pin command test

rjbou:
This is separated in 2 commits for readability/review:
* first that adds a test for all pin subcommands and options;
* second that update previous test (that were only special cases and new subcommands) to make them independent, each one is run in a fresh switch.

I don't know if I added a test for each case (exhaustiveness).

**Backported for 2.3 in #6229**

~Queued on #6219~ *Unqueued* it


tests/reftests/pin.test
```diff
@@ -1,7 +1,811 @@
...
+OPAM/targets/lib/nip-git2/nip-git2.ved.t
+OPAM/targets/lib/nip-path/nip-path.ved.t
+OPAM/targets/lib/nip-path2/nip-path2.ved.t
+OPAM/targets/lib/nip-path3/nip-path3.ved.t
+OPAM/targets/lib/nip-v/nip-v.2.t
+OPAM/targets/lib/nip2/nip2.ved.t
+OPAM/targets/lib/no-url/no-url.ved.t
+### :::::::::::::::::::::::
+### :II: Pin kinds
+### :::::::::::::::::::::::
```
rjbou, line 201: Add also a test for OPAMPINKIND ?

tests/reftests/pin.test
```diff
@@ -175,25 +996,101 @@ echo "another-inexistant" "inexistant"
 -> installed qux.dev
 -> installed bar.dev
 Done.
-### : Test opam pin remove <pkg>.<version>
-### opam pin remove --no-action bar.dev
-Ok, bar is no longer pinned to file://${BASEDIR}/bar (version dev)
+### :XI:c: Test opam pin remove <pkg>.<version>
```
rjbou, line 999: Should this test & remove --all be in its proper section above ?

tests/reftests/pin.test
```diff
@@ -175,25 +996,101 @@ echo "another-inexistant" "inexistant"
...
+nip is now pinned to file://${BASEDIR}/nip (version ved)
 ### opam pin remove --no-action nip.wrong-version
 [ERROR] nip is pinned but not to version wrong-version. Skipping.
 # Return code 2 #
-### : Test opam pin remove --all
-### OPAMNODEPEXTS=1
-### opam pin add --no-action bar.dev ./bar
+### opam pin remove --no-action nip2.dev
+[NOTE] nip2 is not pinned.
+### :XI:d: Test opam pin remove --all
```
rjbou, line 1014: I kept this test here, "as is": be the most similar to previous one with several kind of pins, etc. I don't know if it is intended or not?

tests/reftests/pin.test
```diff
@@ -1,7 +1,811 @@
...
+nip-empty2 is now pinned to file://${BASEDIR}/nip-empty2 (version dev)
+
+The following actions will be performed:
+=== install 1 package
+  - install nip-empty2 dev (pinned)
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> retrieved nip-empty2.dev  (file://${BASEDIR}/nip-empty2)
+-> installed nip-empty2.dev
+Done.
```
rjbou, line 640: Is it a bug?

tests/reftests/pin.test
```diff
@@ -1,7 +1,811 @@
...
+The following actions will be performed:
+=== recompile 1 package
+  - recompile nip-fo dev2 (pinned)
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> removed   nip-fo.dev2
+-> installed nip-fo.dev2
+Done.
+### opam show nip-fo --field=version
+dev2
```
rjbou, line 804: Is it a bug?

tests/reftests/pin.test
```diff
@@ -1,7 +1,811 @@
...
+  nv=$n
+  v=ved
+  nv=$n.$v
+fi
+git=""
+if [ $# -eq 2 ]; then
+  git=$2
+fi
+mkdir -p $n
+cat << EOF > $n/$n.opam
```
kit-ty-kate, line 17:
```suggestion
mkdir -p "$n"
cat << EOF > "$n/$n.opam"
```

tests/reftests/pin.test
```diff
@@ -1,7 +1,811 @@
...
+synopsis: "A word"
+description: "Two words."
+authors: "the testing team"
+homepage: "egapemoh"
+maintainer: "maint@tain.er"
+license: "MIT"
+dev-repo: "hg+https://pkg@op.am"
+bug-reports: "https://nobug"
+EOF
+cat << EOF > $n/$n.install
```
kit-ty-kate, line 31:
```suggestion
cat << EOF > "$n/$n.install"
```

tests/reftests/pin.test
```diff
@@ -1,7 +1,811 @@
...
+lib: ["$nv.t"]
+EOF
+echo piou >> $n/$nv.t
```
kit-ty-kate, line 34:
```suggestion
echo piou >> "$n/$nv.t"
```

tests/reftests/pin.test
```diff
@@ -1,7 +1,811 @@
...
+if [ "x$git" = "xgit" ]; then
+  test -d $n/.git || git -C ./$n init -q --initial-branch=master
+  git -C ./$n config core.autocrlf false
+  git -C ./$n add $n.opam $nv.t $n.install
+  git -C ./$n commit -qm "add $nv package"
```
kit-ty-kate, line 39:
```suggestion
  test -d "$n/.git" || git -C "./$n" init -q --initial-branch=master
  git -C "./$n" config core.autocrlf false
  git -C "./$n" add "$n.opam" "$nv.t" "$n.install"
  git -C "./$n" commit -qm "add $nv package"
```

tests/reftests/pin.test
```diff
@@ -1,7 +1,811 @@
...
+### sh add-pin.sh nip-git3 git
+### sh add-pin.sh nip-git4 git
+### # local dir with path
+### sh add-pin.sh nip-path
+### sh add-pin.sh nip-path2
+### sh add-pin.sh nip-path3
+### sh add-pin.sh nip-path4
+### # repo
+### <add-repo.sh>
+for nv in $@ ; do
```
kit-ty-kate, line 53:
```suggestion
for nv in "$@" ; do
```

tests/reftests/pin.test
```diff
@@ -1,7 +1,811 @@
...
+  n=${nv%.*}
+  pre=REPO/packages/$n/$nv
+  mkdir -p $pre/files
+  echo piou >> $nv.t
+  tar czf $nv.tgz $nv.t
+  MD5=$(openssl md5 $nv.tgz | cut -d' ' -f2)
+  cat << EOF > $pre/opam
```
kit-ty-kate, line 60:
```suggestion
  mkdir -p "$pre/files"
  echo piou >> "$nv.t"
  tar czf "$nv.tgz" "$nv.t"
  MD5=$(openssl md5 "$nv.tgz" | cut -d' ' -f2)
  cat << EOF > "$pre/opam"
```

kit-ty-kate (commented):
LGTM aside from those two things and the first commit message

tests/reftests/pin.test
kit-ty-kate: yes it is a known bug, i've seen it several times but i can't find the issue that i used to redirect people to when they got that issue

tests/reftests/pin.test
kit-ty-kate: the fact that linting is not done if the name of the package is not specified? yeah probably although it doesn't seem critical

tests/reftests/pin.test
kit-ty-kate: I don't think this is necessary but if you want go ahead

tests/reftests/pin.test
kit-ty-kate: i think it's fine here

tests/reftests/pin.test
kit-ty-kate: i'm not sure to understand what you mean here

master_changes.md
```diff
@@ -143,6 +143,7 @@ users)
   * clean: Add to check cleaning of sources directories [#5474 @rjbou]
   * Add reftest for `--verbose-on` option [#5682 @rjbou]
   * Add a test for --deps-only setting direct dependencies as root packages [#6125 @rjbou]
+  * Add pin command test [#6135 @rjbou]
```
kit-ty-kate, line 146:
```suggestion
  * Add some more tests for the pin command [#6135 @rjbou]
```

tests/reftests/pin.test
kit-ty-kate: https://github.com/ocaml/opam/issues/5567#issuecomment-2139447699 maybe?

tests/reftests/pin.test
```diff
@@ -1,9 +1,9 @@
 N0REP0
 ### OPAMYES=1
 ### <add-pin.sh>
-nv=$1
-n=${nv%.*}
-v=${nv#*.}
+nv="$1"
+n="${nv%.*}"
+v="${nv#*.}"
```
kit-ty-kate, line 6: just so you know, double quotes for variable definitions are not necessary.
For example:
```
x=$y/$z
```
is always safe. But:
```
x="$y $z"
```
does require quotes

tests/reftests/pin.test
rjbou: Is it intended to have different kind of pins to remove ? or it was the case because the current switch had already several kind of pins?

master_changes.md
```diff
@@ -107,6 +107,8 @@ users)
 
 ## Reftests
 ### Tests
+  * Add more tests for lint W59 [#6219 @rjbou]
```
kit-ty-kate, line 110:
```suggestion
```

tests/reftests/pin.test
```diff
@@ -1057,24 +1070,100 @@ echo "another-inexistant" "inexistant"
 -> installed bar.dev
 Done.
 ### :XI:c: Test opam pin remove <pkg>.<version>
-### opam pin remove --no-action bar.dev
-Ok, bar is no longer pinned to file://${BASEDIR}/bar (version dev)
+### opam switch create removal --empty
+### opam pin ./nip --no-action
+nip is now pinned to file://${BASEDIR}/nip (version ved)
+### opam pin remove --no-action nip.dev
```
kit-ty-kate, line 1076: This command is duplicated with the nip.wrong-version test below
```suggestion
```

tests/reftests/pin.test
```diff
@@ -1057,24 +1070,100 @@ echo "another-inexistant" "inexistant"
...
+[ERROR] nip is pinned but not to version dev. Skipping.
+# Return code 2 #
+### opam pin ./nip --no-action
+[NOTE] Package nip is already pinned to file://${BASEDIR}/nip (version ved).
+nip is now pinned to file://${BASEDIR}/nip (version ved)
 ### opam pin remove --no-action nip.wrong-version
 [ERROR] nip is pinned but not to version wrong-version. Skipping.
 # Return code 2 #
+### opam pin remove --no-action nip2.dev
```
kit-ty-kate, line 1085: The version of nip is ved not dev. The test is different from originally intended
```suggestion
### opam pin remove --no-action nip.ved
```

tests/reftests/pin.test
```diff
@@ -1057,24 +1070,100 @@ echo "another-inexistant" "inexistant"
...
-Ok, bar is no longer pinned to file://${BASEDIR}/bar (version dev)
+### opam switch create removal --empty
+### opam pin ./nip --no-action
+nip is now pinned to file://${BASEDIR}/nip (version ved)
+### opam pin remove --no-action nip.dev
+[ERROR] nip is pinned but not to version dev. Skipping.
+# Return code 2 #
+### opam pin ./nip --no-action
+[NOTE] Package nip is already pinned to file://${BASEDIR}/nip (version ved).
+nip is now pinned to file://${BASEDIR}/nip (version ved)
```
kit-ty-kate, line 1081: this is not needed since the pinned package hasn't been removed
```suggestion
```

kit-ty-kate (changes requested):

tests/reftests/pin.test
rjbou: In fact, it is 
```
### opam pin remove --no-action nip.ved
```
The logic
* test with a good version of pinned packages
* test with a wrong version of pinned packages
* test with a not pinned package

tests/reftests/pin.test
rjbou: cf above

tests/reftests/pin.test
kit-ty-kate: it's not. It was:
* remove good version of pkg1
* remove wrong version of pkg2

Thus we should simplify into:
* remove wrong version of pkg1
* remove good version of pkg1

rjbou: In the original test i agree it was : remove good version then remove wrong version. I've updated the test to add also removal of unpinned one.

"The logic" i was talking about is on the wanted final test (my modif), not the original.

kit-ty-kate: i don't think the extra test makes sense and i don't think we should add new tests here. The commit changing this part is here just to add separation not adding extra tests. It's already complicated to review changes in the reftests as it is.

rjbou: In the final form (not yet pushed, i'm still debugging cygwin), i've separated the addition of unpinned package test in first commit like that the second commit only contains the switch change.

kit-ty-kate (approved):
I'm personally not a huge fan of editing the testsuite. I would've preferred to have a new file or to rename the old one if you prefer, but changing it is hard to review and can potentially break test-cases (the `opam pin remove <pkg>.<version>` section was a good example). In my opinion our testsuite should be append-only whenever possible in the future.

In any case i won't block this. It looks fine as far as i could review.
Thanks!

tests/reftests/pin.test
```diff
@@ -1,401 +1,911 @@
...
+  - install nip-git3 ved (pinned)
 
-The following system packages will first need to be installed:
-    another-inexistant inexistant
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> installed nip-git2.ved
+-> installed nip-git3.ved
+Done.
+### # shell on Cygwin ads a trialing `\r` in the middle of the string
+### sh -c 'opam pin add $(opam pin scan ./lot_of_pkgs --normalise | grep path | sed "s/\\r//g") ' | unordered
```
kit-ty-kate, line 639: What about:
```suggestion
### sh -c 'opam pin scan ./lot_of_pkgs --normalise | xargs opam pin add'
```

rjbou: it doesn't work neither. If there is 2 packages, the `\\r` issue is present

kit-ty-kate: interesting

kit-ty-kate (approved):

kit-ty-kate:
Thanks a lot!

PR #6138, opened 2024-08-05
Update the installation documentation after the release of opam 2.2

kit-ty-kate:
We forgot to update that page after the release of opam 2.2


rjbou (approved):
Thanks!

PR #6141, opened 2024-08-07
Fix markup for inline code in section on `Environment update portability`

shonfeder:
The broken formatting is pretty apparent on the published docs currently: https://opam.ocaml.org/doc/Manual.html#Environment-update-portability


kit-ty-kate (approved):

kit-ty-kate:
Thanks a lot!

PR #6142, opened 2024-08-07
Add curl to the list of required cygwin packages to avoid issues with Windows' curl

### LINKED ISSUE #6120: error : [OPAM INIT] - Platform: Windows 11 HOME - Version: 23h2
Steps:

- Install cygwin
- Install packages
 
Packages

- Download opam from releases
- Add to env
- Run 
```
opam init
```
Error

```
Can't execute opam config report becauses its not initialized
```

kit-ty-kate:
Fixes #6120


dra27:
I think this is probably the way it'll have to go - two things which ought to be checked:
- [ ] Does `opam init --reinit` catch this (e.g. external Cygwin or MSYS2 which _doesn't_ have curl installed; `opam init` with 2.2.0; does `opam init --reinit` prompt to add the curl package
- [ ] Is it still possible to use `OPAMFETCH` to force the use of either wget or the system's curl (it ought to be; I just haven't checked)

kit-ty-kate:
> * Does `opam init --reinit` catch this (e.g. external Cygwin or MSYS2 which _doesn't_ have curl installed; `opam init` with 2.2.0; does `opam init --reinit` prompt to add the curl package

I've tried this and it's not pulled by `opam init --reinit` if curl is missing. I'll have a look at a solution

> * Is it still possible to use `OPAMFETCH` to force the use of either wget or the system's curl (it ought to be; I just haven't checked)

`OPAMCURL=C:\Windows\System32\curl.exe` and `OPAMFETCH=wget` both worked as expected

kit-ty-kate:
The last commit makes `opam init --reinit` re-run setup.exe and after trying it locally i can confirm this will reinstall curl if it is not installed yet.

kit-ty-kate:
Note for reviewers: the diff is easier to read with the "Hide whitespace" option on.

dra27:
Thanks, @kit-ty-kate! I think the additional code for `opam init --reinit` will be worth checking and integrating regardless, but I just found https://github.com/curl/curl/issues/13845 ... which makes sense, since Windows 11 has cURL 8.8.0, but my Windows 10 box has 8.0.1.

kit-ty-kate:
> I'm not sure, but does the present commit mean that `opam init --reinit` now _always_ re-runs Cygwin setup?

It does but i personally view it as a feature. As far as i know we currently have no way to update cygwin itself except by running the setup.exe binary itself and give all the correct paths to it, or by installing a depext. I feel like having `opam init --reinit` do that would be fairly reasonable.
Although looking into it more it seems that it's not the case here either but it should be if `--upgrade-also` where given to the setup.exe incantation :magic_wand:

dra27:
Yes, indeed - I think re-running setup for the internal Cygwin installation is a good idea .. but, as you note, it needs `--upgrade-also`! However, the check higher up is also needed for the MSYS2/external-Cygwin case.

kit-ty-kate:
Converted to draft as i think the original issue is better fixed by https://github.com/ocaml/opam/pull/6168. However i'm keeping this around as we might need want to think about the upgrade mechanism again at some point

PR #6144, opened 2024-08-07
Reduce allocations in opamVersionCompare

talex5:
Test-case:

```ocaml
let ()
  let x0 = Gc.stat () in
  assert (OpamVersionCompare.compare "1.2.3" "1.02.3" == 0);
  let x1 = Gc.stat () in
  Printf.printf "minor_words: %.0f\n" (x1.minor_words -. x0.minor_words)
```

Before: minor_words: 156
After:  minor_words: 84

This just passes all the parameters to `loop` rather than getting them from the context. Then `loop` itself isn't needed.

Ideally, there would be no allocations at all. However, removing the others makes the code quite messy so I just did the easy change here.

See https://roscidus.com/blog/blog/2024/07/22/performance-2/#statmemprof for some real-world profiling of the CI solver service, showing this to be a problem.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -121,6 +121,7 @@ users)
 ## Internal
   * Stop using polymorphic comparison when comparing `OpamTypes.switch_selections` [#6102 @kit-ty-kate]
   * Remove the meta opam packages opam and opam-admin [#6115 @kit-ty-kate]
+  * Reduce allocations in OpamVersionCompare [#6144 @talex5]
 
 ## Internal: Windows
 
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
Thanks a lot for this contribution. Would you be able to rebase it on top of https://github.com/ocaml/opam/pull/5518 ? I believe with both it should be able to reduce allocation further

kit-ty-kate (approved):
Our automated benchmark shows a 18% performance improvement. lgtm

talex5:
I've pushed a commit just to test it on top of https://github.com/ocaml/opam/pull/5518. Will need squashing if we decide to keep it.

kit-ty-kate:
With both work it improves performance by 41%! :tada:

kit-ty-kate:
I've put back the previous state and will merge this as soon as CI finishes, i'll update my PR on top of master after. Thanks again!

PR #6154, opened 2024-08-13
Upgrade to (lang dune 3.0)

kit-ty-kate:
According to @rgrinberg:
> Older dune doesn’t have proper support for directory targets, so stale artifact deletion would run and delete the contents of the directory

~This is the case for `make reftest-gen` (aka. `dune build @reftest-gen`) which starts by deleting every opam roots and opam repositories in `_build/default/tests/reftests` previously created by `make test`, which makes that command take upwards of 70 seconds before actually running `reftest-gen`.~
**EDIT:** Actually this is a separate issue fixed in https://github.com/ocaml/opam/pull/6155

Currently this PR does not work and makes every tests fail with:
```
[WARNING] No switch is currently set, perhaps you meant '--set-default'?
Fatal error:
/<path>/_build/default/src/client/opamMain.exe: "open" failed on
/tmp/build_ff9d44_dune/opam-reftest-7c8979/OPAM/repo/lock: Permission denied
```
and i have no idea why. Does anyone have an idea?

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -1,7 +1,8 @@
-(lang dune 2.0)
+(lang dune 3.0)
 (name opam)
 
 (implicit_transitive_deps true)
+(using directory-targets 0.1)
 
 (package (name opam-core))
 (package (name opam-format))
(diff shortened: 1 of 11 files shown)


kit-ty-kate:
The original issue was fixed in https://github.com/ocaml/opam/pull/6155 so closing this. We'll see about the permission issue whenever we upgrade to `(lang dune 3.0)`

PR #6155, opened 2024-08-13
Speedup "make reftest-gen"

kit-ty-kate:
Opening https://github.com/ocaml/opam/pull/6154 made me realised that the reason `make reftest-gen` is slow is that we're removing `tests/reftests/dune.inc` which lists all the opam roots/opam repositories and tar.gz targets, so dune build will scan the directory and remove everything that it doesn't know about and given those targets are not defined anymore it will then remove everything.

Simply not emptying that file fixes the issue. If in the future we need to upgrade the syntax or whatever we can always remove that file manually and the error message should make it obvious. But in the meantime this change is extremely beneficial to the day-to-day development setup by not having to remove then re-download, unpack and recreate 5GB everytime that rule is called


rjbou (approved):
Thanks!

Issue #6158, opened 2024-08-15
Suggestion for command hints


ttamttam:
I’d suggest to always output command hints on one single separate line.

It would make it a lot easier to copy paste if needed.

By instance

```
Run (& opam env --switch=default) -split '\r?\n' | ForEach-Object { Invoke-Expression $_ } to update the current shell environment
```

Would become 

```
To update the current shelll environment, run
    (& opam env --switch=default) -split '\r?\n' | ForEach-Object { Invoke-Expression $_ }
```

Best regards

kit-ty-kate:
I've opened https://github.com/ocaml/opam/pull/6159 with a slight difference explained in the description. What do you think?

ttamttam:
Closing. See #6159

PR #6159, opened 2024-08-15
Make the shell environment update hint easier to copy/paste

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/6158

I chose to keep in a single line instead of 2 just because we'd want to have it on a line starting with `# ` and i'm worried people would also copy the prefix when copy/pasting

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -74,6 +74,7 @@ users)
 ## Clean
 
 ## Env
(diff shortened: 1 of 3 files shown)


ttamttam (approved):
OK. I understand the constraint. Your solution is certainly an improvement.

rjbou (approved):
Thanks!

PR #6164, opened 2024-08-20
reftest: add test for pin wrongly fetched

rjbou:


rjbou:
Test added and fixed in #6221

Issue #467, opened 2013-02-14
retry curl on timeout


avsm:
Jenkins seems to fail occasionally with curl timeouts. Is there some retry option in curl that would make this a little more robust to transient errors?

PR #6168, opened 2024-08-20
Mitigate problems with curl on Windows 11

### LINKED ISSUE #6120: error : [OPAM INIT] - Platform: Windows 11 HOME - Version: 23h2

dra27:
Fixes #6120 by altering the call to curl. curl/curl#13845 can be mitigated by avoiding the use of `--write-out` so, having detected exit code 43 (which should never happen using the curl binary, as `curl` itself should normally know how to call its own functions!), a second call is made to call using `--fail` instead of `--write-out`. `--fail` was previously considered in https://github.com/ocaml/opam/issues/467#issuecomment-13881022. I've done some digging, and the semantics of `--fail` are good enough for what we're after, though using `--write-out` as in @samoht's f2ee0ec6e35196be2c57e8c32cb310cd942c771f is superior in general because it conveys more information to the user when an error occurs (which was the point).

The bug in cURL is causing enough issues elsewhere that I imagine Microsoft will update the `curl.exe` binary shipped with Windows before long - if they don't, it's easy to extend the mitigation further to extract the actual http response code from the error message, as the format of that message has never altered, from a look at cURL's sources.

There are three reasons to prefer this slightly more complicated fix to just adding the requirement for cURL from Cygwin/MSYS2 as in #6142:
1. We're getting away with the workaround of adding Cygwin's `curl` because Cygwin's SSL certificate doesn't contain the offending OIDS. If that were to change, the workaround would fail (because we have to download Cygwin's setup using `C:\Windows\System32\curl.exe`)
2. Cygwin's curl is (correctly) built with the wrong SSL stack, so the workaround has the potential to create problems for some users (same issue as Cygwin git vs Git-for-Windows)
3. The general direction of travel is supposed to be using fewer Cygwin/MSYS2 tools, so it's a shame to have to go in the opposite direction!

Running `opam init --debug -vv` on my Windows 11 laptop with this binary we can now see:

```
00:03.923  XSYS                   Downloading Cygwin setup checksums
Downloading Cygwin setup from cygwin.com
00:03.924  SYSTEM                 mkdir C:\Users\DRA\AppData\Local\Temp\opam-27848-4d3c3d
+ C:\Windows\system32\curl.exe "--write-out" "%{http_code}\\n" "--retry" "3" "--retry-delay" "2" "--user-agent" "opam/2.3.0~alpha~dev" "-L" "-o" "C:\\Users\\DRA\\AppData\\Local\\Temp\\opam-27848-4d3c3d\\sha512.sum.part" "--" "https://cygwin.com/sha512.sum"
- 200
-   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
-                                  Dload  Upload   Total   Spent    Left  Speed
100   291  100   291    0     0    666      0 --:--:-- --:--:-- --:--:--   670
00:04.413  SYSTEM                 rm C:\Users\DRA\AppData\Local\Temp\opam-DRA-27848\dl-27848-2260b8
00:04.413  SYSTEM                 mv C:\Users\DRA\AppData\Local\Temp\opam-27848-4d3c3d\sha512.sum.part -> C:\Users\DRA\AppData\Local\Temp\opam-27848-4d3c3d\sha512.sum
00:04.422  XSYS                   Downloading setup-x86_64.exe
00:04.422  SYSTEM                 mkdir C:\Devel\Roots\broken-curl
00:04.422  SYSTEM                 mkdir C:\Devel\Roots\broken-curl\.cygwin
+ C:\Windows\system32\curl.exe "--write-out" "%{http_code}\\n" "--retry" "3" "--retry-delay" "2" "--user-agent" "opam/2.3.0~alpha~dev" "-L" "-o" "C:\\Devel\\Roots\\broken-curl\\.cygwin\\setup-x86_64.exe.part" "--" "https://cygwin.com/setup-x86_64.exe"
- 200
-   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
-                                  Dload  Upload   Total   Spent    Left  Speed
100 1374k  100 1374k    0     0  1194k      0  0:00:01  0:00:01 --:--:-- 1196k
```

downloading Cygwin with no issues at all, but then mitigation kicking in for opam.ocaml.org:

```
00:44.717  PARALLEL               Next task in job 0: C:\Windows\system32\curl.exe --write-out %{http_code}\n --retry 3 --retry-delay 2 --user-agent opam/2.3.0~alpha~dev -L -o C:\Users\DRA\AppData\Local\Temp\opam-27848-a08201\index.tar.gz.part -- https://opam.ocaml.org/index.tar.gz
Processing  1/1: [default: http]
+ C:\Windows\system32\curl.exe "--write-out" "%{http_code}\\n" "--retry" "3" "--retry-delay" "2" "--user-agent" "opam/2.3.0~alpha~dev" "-L" "-o" "C:\\Users\\DRA\\AppData\\Local\\Temp\\opam-27848-a08201\\index.tar.gz.part" "--" "https://opam.ocaml.org/index.tar.gz"
- 000
-   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
-                                  Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
- curl: (43) A libcurl function was given a bad argument
00:45.092  PARALLEL               Collected task for job 0 (ret:43)
00:45.092  CURL                   Attempting to mitigate curl/curl#13845
00:45.093  PARALLEL               Next task in job 0: C:\Windows\system32\curl.exe --fail --retry 3 --retry-delay 2 --user-agent opam/2.3.0~alpha~dev -L -o C:\Users\DRA\AppData\Local\Temp\opam-27848-a08201\index.tar.gz.part -- https://opam.ocaml.org/index.tar.gz
+ C:\Windows\system32\curl.exe "--fail" "--retry" "3" "--retry-delay" "2" "--user-agent" "opam/2.3.0~alpha~dev" "-L" "-o" "C:\\Users\\DRA\\AppData\\Local\\Temp\\opam-27848-a08201\\index.tar.gz.part" "--" "https://opam.ocaml.org/index.tar.gz"
-   % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
-                                  Dload  Upload   Total   Spent    Left  Speed
100 7223k  100 7223k    0     0  14.9M      0 --:--:-- --:--:-- --:--:-- 14.9M
00:45.610  PARALLEL               Collected task for job 0 (ret:0)
00:45.611  SYSTEM                 rm C:\Users\DRA\AppData\Local\Temp\opam-DRA-27848\dl-27848-5e0cc2
00:45.611  SYSTEM                 rm C:\Users\DRA\AppData\Local\Temp\opam-DRA-27848\dl-27848-f3db71
00:45.612  SYSTEM                 mv C:\Users\DRA\AppData\Local\Temp\opam-27848-a08201\index.tar.gz.part -> C:\Users\DRA\AppData\Local\Temp\opam-27848-a08201\index.tar.gz
```

### DIFF
--- a/src/repository/opamDownload.ml
+++ b/src/repository/opamDownload.ml
@@ -20,19 +20,28 @@ let fail (s,l) = raise (Download_fail (s,l))
 let user_agent =
   CString (Printf.sprintf "opam/%s" (OpamVersion.(to_string current)))
 
-let curl_args = [
-  CString "--write-out", None;
-  CString "%%{http_code}\\n", None; (* 6.5 13-Mar-2000 *)
-  CString "--retry", None; CIdent "retry", None; (* 7.12.3 20-Dec-2004 *)
-  CString "--retry-delay", None; CString "2", None; (* 7.12.3 20-Dec-2004 *)
-  CString "--compressed",
-  Some (FIdent (OpamFilter.ident_of_string "compress")); (* 7.10 1-Oct-2002 *)
-  CString "--user-agent", None; user_agent, None; (* 4.5.1 12-Jun-1998 *)
-  CString "-L", None; (* 4.9 7-Oct-1998 *)
-  CString "-o", None; CIdent "out", None; (* 2.3 21-Aug-1997 *)
-  CString "--", None; (* End list of options; 5.0 1-Dec-1998 *)
-  CIdent "url", None;
-]
+let curl_args =
+  let main_args = [
+    CString "--retry", None; CIdent "retry", None; (* 7.12.3 20-Dec-2004 *)
+    CString "--retry-delay", None; CString "2", None; (* 7.12.3 20-Dec-2004 *)
+    CString "--compressed",
+    Some (FIdent (OpamFilter.ident_of_string "compress")); (* 7.10 1-Oct-2002 *)
+    CString "--user-agent", None; user_agent, None; (* 4.5.1 12-Jun-1998 *)
+    CString "-L", None; (* 4.9 7-Oct-1998 *)
+    CString "-o", None; CIdent "out", None; (* 2.3 21-Aug-1997 *)
+    CString "--", None; (* End list of options; 5.0 1-Dec-1998 *)
+    CIdent "url", None;
+  ] in
+  fun ~with_mitigation ->
+    if with_mitigation then
+      (* --fail is as old as curl; though the assumption that it leads to exit
+         code 22 when there's an error is probably 5.3 21-Dec-1998 (prior to
+         that it led to exit code 21) *)
+      (CString "--fail", None) :: main_args
+    else
+      (CString "--write-out", None) ::
+      (CString "%%{http_code}\\n", None) :: (* 6.5 13-Mar-2000 *)
+      main_args
 
 let wget_args = [
   CString "-t", None; CIdent "retry", None;
@@ -56,14 +65,16 @@ let ftp_args = [
   CIdent "url", None;
 ]
 
-let download_args ~url ~out ~retry ?checksum ~compress () =
+let download_args ~url ~out ~retry ?(with_curl_mitigation=false)
+                  ?checksum ~compress () =
   let cmd, _ = Lazy.force OpamRepositoryConfig.(!r.download_tool) in
   let cmd =
     match cmd with
     | [(CIdent "wget"), _] -> cmd @ wget_args
     | [(CIdent "fetch"), _] -> cmd @ fetch_args
     | [(CIdent "ftp"), _] -> cmd @ ftp_args
-    | [_] -> cmd @ curl_args (* Assume curl if the command is a single arg *)
+      (* Assume curl if the command is a single arg *)
+    | [_] -> cmd @ curl_args ~with_mitigation:with_curl_mitigation
     | _ -> cmd
   in
   OpamFilter.single_command (fun v ->
@@ -89,7 +100,28 @@ let download_args ~url ~out ~retry ?checksum ~compress () =
       | _ -> None)
     cmd
 
-let tool_return url ret =
+let download_command_t ~with_curl_mitigation ~compress ?checksum ~url ~dst c =
+  let cmd, args =
+    match
+      download_args
+        ~url
+        ~out:dst
+        ~retry:OpamRepositoryConfig.(!r.retries)
+        ~with_curl_mitigation
+        ?checksum
+        ~compress
+        ()
+    with
+    | cmd::args -> cmd, args
+    | [] ->
+      OpamConsole.error_and_exit `Configuration_error
+        "Empty custom download command"
+  in
+  let stdout = OpamSystem.temp_file ~auto_clean:false "dl" in
+  OpamProcess.Job.finally (fun () -> OpamSystem.remove_file stdout) @@ fun () ->
+  OpamSystem.make_command ~allow_stdin:false ~stdout cmd args @@> c
+
+let tool_return redownload_command url ret =
   match Lazy.force OpamRepositoryConfig.(!r.download_tool) with
   | _, `Default ->
     if OpamProcess.is_failure ret then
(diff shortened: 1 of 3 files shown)


kit-ty-kate (approved):
LGTM. ~I added a commit (feel free to remove it if you disagree) to remove the `with_curl_mitigation` argument which seems to me like something we probably want to do everytime. The other places that use it do not seem to use the error code at all.~ EDIT: oops sorry i'm tired, nevermind ^^"

To me it looks good to go once rebased (I would also suggest squashing everything into one commit as the first commit makes more sense combined with the second)

dra27:
Thanks, @kit-ty-kate! FWIW, the point in two commits is entirely because the first commit makes sense on its own (and simplifies the diff of the actual diff in the second):

```ocaml
if foo then
  raise Bar;
baz ()
```
is functionally inferior to:
```ocaml
if foo then
  raise Bar
else
  baz ()
```
especially when considering an alteration to the first branch of the `if` which is what this PR is doing)

src/repository/opamDownload.ml
```diff
@@ -34,6 +32,17 @@ let curl_args = [
...
+let curl_args ~with_mitigation =
+  if with_mitigation then
+    (* --fail is as old as cURL; though the assumption that it leads to exit
+       code 22 when there's an error is probably 5.3 21-Dec-1998 (prior to
+       that it led to exit code 21) *)
+    (CString "--fail", None) :: curl_args
+  else
+    (CString "--write-out", None) ::
+    (CString "%%{http_code}\\n", None) :: (* 6.5 13-Mar-2000 *)
+    curl_args
```
rjbou, line 44:
```suggestion
let curl_args =
  let main_args = [
    CString "--retry", None; CIdent "retry", None; (* 7.12.3 20-Dec-2004 *)
    CString "--retry-delay", None; CString "2", None; (* 7.12.3 20-Dec-2004 *)
    CString "--compressed",
    Some (FIdent (OpamFilter.ident_of_string "compress")); (* 7.10 1-Oct-2002 *)
    CString "--user-agent", None; user_agent, None; (* 4.5.1 12-Jun-1998 *)
    CString "-L", None; (* 4.9 7-Oct-1998 *)
    CString "-o", None; CIdent "out", None; (* 2.3 21-Aug-1997 *)
    CString "--", None; (* End list of options; 5.0 1-Dec-1998 *)
    CIdent "url", None;
  ] in
  fun ~with_mitigation ->
    if with_mitigation then
      (* --fail is as old as cURL; though the assumption that it leads to exit
         code 22 when there's an error is probably 5.3 21-Dec-1998 (prior to
         that it led to exit code 21) *)
      (CString "--fail", None) :: main_args
    else
      (CString "--write-out", None) ::
      (CString "%%{http_code}\\n", None) :: (* 6.5 13-Mar-2000 *)
      main_args
```
Better have it in the function than redefine it

src/repository/opamDownload.ml
```diff
@@ -56,14 +65,16 @@ let ftp_args = [
   CIdent "url", None;
 ]
 
-let download_args ~url ~out ~retry ?checksum ~compress () =
+let download_args ~url ~out ~retry ?(with_curl_mitigation=false)
```
rjbou, line 68:
```suggestion
let download_args_t ~url ~out ~retry ?(with_curl_mitigation=false)
```

src/repository/opamDownload.ml
```diff
@@ -133,7 +119,58 @@ let download_command ~compress ?checksum ~url ~dst () =
...
+        let code = List.hd (List.rev l) in
+        let num = try int_of_string code with Failure _ -> 999 in
+        if num >= 400 then
+          fail (Some ("curl error code " ^ code),
+                    Printf.sprintf "curl: code %s while downloading %s"
+                      code (OpamUrl.to_string url))
+        else Done ()
+
+let download_command ~compress ?checksum ~url ~dst () =
+  let download_command = download_command ~compress ?checksum ~url ~dst in
```
rjbou, line 172:
```suggestion
  let download_command = download_command_t ~compress ?checksum ~url ~dst in
```

src/repository/opamDownload.ml
```diff
@@ -133,7 +119,58 @@ let download_command ~compress ?checksum ~url ~dst () =
...
+              fail (Some "cURL failed owing to a server-side issue",
+                      Printf.sprintf "cURL failed with server-side error: %s"
+                        (OpamProcess.result_summary ret))
+            else
+              fail (Some "cURL failed",
+                      Printf.sprintf "cURL failed: %s"
+                        (OpamProcess.result_summary ret))
+          else Done ()
+      end else
+        fail (Some "Curl failed", Printf.sprintf "Curl failed: %s"
```
rjbou, line 154:
```suggestion
        fail (Some "cURL failed", Printf.sprintf "cURL failed: %s"
```

src/repository/opamDownload.ml
```diff
@@ -133,7 +119,58 @@ let download_command ~compress ?checksum ~url ~dst () =
...
+                    (OpamProcess.result_summary ret))
+    else
+      match ret.OpamProcess.r_stdout with
+      | [] ->
+        fail (Some "curl empty response",
+                  Printf.sprintf "curl: empty response while downloading %s"
```
rjbou, line 160:
```suggestion
        fail (Some "cURL empty response",
                  Printf.sprintf "cURL: empty response while downloading %s"
```

src/repository/opamDownload.ml
```diff
@@ -133,7 +119,58 @@ let download_command ~compress ?checksum ~url ~dst () =
...
+                    (OpamUrl.to_string url))
+      | l  ->
+        let code = List.hd (List.rev l) in
+        let num = try int_of_string code with Failure _ -> 999 in
+        if num >= 400 then
+          fail (Some ("curl error code " ^ code),
+                    Printf.sprintf "curl: code %s while downloading %s"
```
rjbou, line 167:
```suggestion
                    Printf.sprintf "cURL: code %s while downloading %s"
```

src/repository/opamDownload.ml
rjbou: Given the PR and new prints, it's worth harmonising with the good spelling

src/repository/opamDownload.ml
kit-ty-kate: The proper name is curl: https://curl.se/

rjbou: oh ok

dra27:
Thanks, @rjbou - changes applied

src/repository/opamDownload.ml
rjbou: oups, no need for this one. I misread

rjbou (approved):
Thanks!

PR #6172, opened 2024-08-21
Release 2.2.1

kit-ty-kate:
~This is a draft until it is decided whether or not https://github.com/ocaml/opam/pull/6169 should be part of the release or whether it should be part of 2.2.2~


rjbou (approved):
Thanks!

CHANGES
```diff
@@ -3,6 +3,24 @@ repositories (changes that are automatically handled by the format upgrade tools
 are not marked). Those prefixed with "(+)" are new command/option (since
 2.1.0~alpha2).
 
+2.2.1:
+* (*) Fix a regression in `opam install --deps-only` where the direct
+  dependencies were not set as root packages [#6125 @rjbou]
+* (*) Fix a regression when building git packages where the build result
+  were sometimes inconsistent compared to outside opam, by disabling shallow
+  clone by default except when fetching an opam repositories
```
rjbou, line 11:
```suggestion
* (*) Fix a regression on git fetching by disabling shallow
  clone by default except when fetching an opam repositories
```
It is not only git package build that was affected

kit-ty-kate: Do you have an example that isn't a git package?

kit-ty-kate: oh you mean local (rsync), sure. I can replace by "packages using git"

kit-ty-kate: mh, actually no the issue wouldn't be triggered there. Do you have an example?

kit-ty-kate: The entry doesn't say it's failing. Only that the result is inconsistent between with and without opam. That does englobe all that in my opinion

kit-ty-kate: fixed

Issue #6174, opened 2024-08-22
`opam init` fails with `Syntax error: "done" unexpected` on v2.2.1


zoj613:
setting up the `setup-ocaml` github action using opam v2.2.1 fails with
the following output:
```
/opt/hostedtoolcache/opam/2.2.1/x86_64/opam init --auto-setup --bare --enable-shell-hook
  /opt/hostedtoolcache/opam/2.2.1/x86_64/opam: 1: checking: not found
  /opt/hostedtoolcache/opam/2.2.1/x86_64/opam: 2: OCaml: not found
  /opt/hostedtoolcache/opam/2.2.1/x86_64/opam: 3: OCaml: not found
  /opt/hostedtoolcache/opam/2.2.1/x86_64/opam: 4: checking: not found
  /opt/hostedtoolcache/opam/2.2.1/x86_64/opam: 5: checking: not found
  /opt/hostedtoolcache/opam/2.2.1/x86_64/opam: 6: checking: not found
  /opt/hostedtoolcache/opam/2.2.1/x86_64/opam: 7: checking: not found
  /opt/hostedtoolcache/opam/2.2.1/x86_64/opam: 8: checking: not found
  /opt/hostedtoolcache/opam/2.2.1/x86_64/opam: 9: checking: not found
  /opt/hostedtoolcache/opam/2.2.1/x86_64/opam: 10: checking: not found
  /opt/hostedtoolcache/opam/2.2.1/x86_64/opam: 11: checking: not found
  /opt/hostedtoolcache/opam/2.2.1/x86_64/opam: 12: checking: not found
  /opt/hostedtoolcache/opam/2.2.1/x86_64/opam: 13: checking: not found
  /opt/hostedtoolcache/opam/2.2.1/x86_64/opam: 14: checking: not found
  /opt/hostedtoolcache/opam/2.2.1/x86_64/opam: 15: checking: not found
  /opt/hostedtoolcache/opam/2.2.1/x86_64/opam: 16: checking: not found
  /opt/hostedtoolcache/opam/2.2.1/x86_64/opam: 17: checking: not found
  /opt/hostedtoolcache/opam/2.2.1/x86_64/opam: 18: checking: not found
  /opt/hostedtoolcache/opam/2.2.1/x86_64/opam: 19: checking: not found
  /opt/hostedtoolcache/opam/2.2.1/x86_64/opam: 20: checking: not found
[...]

PR #6175, opened 2024-08-22
release: fix the docker phase

kit-ty-kate:
Would prepend the output of the configure script before.

Fixes https://github.com/ocaml/opam/issues/6174
Backported to 2.2 in https://github.com/ocaml/opam/pull/6175
Backported to 2.1 in https://github.com/ocaml/opam/pull/6177
Backported to 2.0 in https://github.com/ocaml/opam/pull/6178


rjbou (approved):
Thanks!

PR #6182, opened 2024-08-27
Fix highly embarrassing mistake in patch transformation

dra27:
This has sadly been wrong since #3349 was fixed in #3449 (which equates to "it's always been wrong", as that was during 2.0.0's dev cycle).

Patches require transformation to deal with line-endings - patch can cope with patching CRLF files, but the diff needs to have the same line-endings as the files. Prior to applying a patch, `OpamSystem.translate_patch` checks the files to which the diff is being applied and transforms the patch. If no transformations are needed (which is pretty much always - most release tarballs use Unix line-endings, even on Windows), then the patch is simply copied.

The process works fine, except for a teeny detail. Owing to a badly made logging transformation in #3449, the transformation was only written to the patch file if debugging mode was enabled 🤦

On Windows, while working on compiler packaging, this manifested itself as:

```
∗ installed ocaml-env-mingw64.1
∗ installed system-mingw.1
⬇ retrieved ocaml-compiler.4.08.1  (cached)
∗ installed mingw-w64-shims.0.2.0

#=== ERROR while compiling ocaml-compiler.4.08.1 ==============================#
Sys_error("C:\\Users\\DRA\\AppData\\Local\\opam\\log\\processed-patch-25204-bf7b21: Permission denied")


<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫
│ λ build ocaml-compiler 4.08.1
```

but the patch is successfully transformed when running with `--debug`. What was happening was the the patch file is opened (`ch_out`); nothing is written to it so the patch of course "applies" perfectly but because `ch_out` was never closed, the removal fails after trying to apply the patch (as it happens, the removal is skipped when debugging, but would in fact succeed).

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -144,6 +144,7 @@ users)
   * Remove the meta opam packages opam and opam-admin [#6115 @kit-ty-kate]
   * Reduce allocations in OpamVersionCompare [#6144 @talex5]
   * Speedup OpamVersionCompare by 25% by removing the unused handling of epoch [#5518 @kit-ty-kate]
+  * Fix error in `OpamSystem.transform_patch` - patches were only applied when debugging [#6182 @dra27 regression since #3449]
 
 ## Internal: Windows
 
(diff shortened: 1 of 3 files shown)


rjbou (approved):
Thanks!

PR #6183, opened 2024-08-27
[2.2 backport] Fix highly embarrassing mistake in patch transformation

dra27:
2.2 version of #6182. This code was supposed to have been quite battle-tested, but of course it hasn't been owing to this error.

~~The fix in #6182 is for everywhere - I propose limiting it slightly in 2.2.2 to be for Windows-only, to remove the risk of 2.2.2 causing unexpected regressions on non-Windows platforms.~~

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -90,6 +90,7 @@ users)
 ## Shell
 
 ## Internal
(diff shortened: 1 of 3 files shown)


rjbou (approved):
Thanks!

dra27:
My logic was flawed - in 2.0 patches which need translating are silently skipped on Linux/macOS so it's arguably worse than the Windows case!

Issue #6152, opened 2024-08-13
`opam switch list-available` doesn't handle `pkg.version` pattern


rjbou:
```
$ opam switch list-available 'ocaml-b*'
# Listing available compilers from repositories: default
# Name              # Version     # Synopsis
ocaml-base-compiler 3.07          Official 3.07 release
ocaml-base-compiler 3.07+1        Official 3.07+1 release
[..]
ocaml-base-compiler 5.2.0~beta2   Second beta release of OCaml 5.2.0
ocaml-base-compiler 5.2.0~rc1     First release candidate of OCaml 5.2.0
ocaml-base-compiler 5.2.0         Official release 5.2.0

$ opam switch list-available 'ocaml-base-compiler'
# Listing available compilers from repositories: default
# Name              # Version     # Synopsis
ocaml-base-compiler 3.07          Official 3.07 release
ocaml-base-compiler 3.07+1        Official 3.07+1 release
[..]
ocaml-base-compiler 5.2.0~beta2   Second beta release of OCaml 5.2.0
ocaml-base-compiler 5.2.0~rc1     First release candidate of OCaml 5.2.0
ocaml-base-compiler 5.2.0         Official release 5.2.0

$ opam switch list-available 'ocaml-base-compiler.4.14.0'
# Listing available compilers from repositories: default
# No matches found

$ opam list 'ocaml-base-compiler.4.14.0'
# Packages matching: name-match(ocaml-base-compiler) & version-match(4.14.0) & (installed | available)
# Package                  # Installed # Synopsis
ocaml-base-compiler.4.14.0 4.14.2      Official release 4.14.0
```

rjbou:
fixed by #6186

PR #6186, opened 2024-09-02
Fix `opam switch list-available pkg.version` pattern

arozovyk:
Adresses #6152 
Queued on #6318

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -40,6 +40,7 @@ users)
 ## Switch
   * [BUG] Fix `opam switch remove <dir>` failure when it is a linked switch [#6276 @btjorge - fix #6275]
   * [BUG] Fix `opam switch list-available` when given several arguments [#6318 @kit-ty-kate]
+  * [BUG] Correctly handle `pkg.version` pattern in `opam switch list-available` [#6186 @arozovyk - fix #6152]
 
 ## Config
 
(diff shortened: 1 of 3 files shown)


rjbou (approved):
Thanks! As we discussed, I've reworked the change.

kit-ty-kate (changes requested):
I'm not sure `ands` made sense in general in the first place. I think we should change the behaviour to use `ors` instead.
Otherwise there is no use to allow several parameters. e.g.
```
$ opam switch list-available ocaml-base-compiler ocaml-variants
# Listing available compilers from repositories: kit-ty-kate, default
# No matches found
```
If we switch to `ors` instead we can also simply remove the addition of the `?concat`, which doesn't seem useful to me.

Ideally we should also have a test showing the behaviour of `list-available` with several parameters

rjbou:
If we switch from `ands` to `ors`, it is the purpose of another PR.

kit-ty-kate:
> If we switch from `ands` to `ors`, it is the purpose of another PR.

Done in https://github.com/ocaml/opam/pull/6318. This should be merged first. I've marked this here PR as queued on it.

kit-ty-kate:
rebased

kit-ty-kate:
new proposal queued on top of #6318 with one extra test

kit-ty-kate (approved):

kit-ty-kate (approved):

kit-ty-kate:
Thanks!

PR #6187, opened 2024-09-03
Revert "[2.2 backport] Fix highly embarrassing mistake in patch transformation"

kit-ty-kate:
Reverts ocaml/opam#6183

After discussion, since the new release cycle is now shorter we would like to reserve the backports to regressions and security fixes.


dra27 (approved):
:+1:

PR #6190, opened 2024-09-04
Env: Add install.exe to the shadow program list

dra27:
install.exe is installed by Vim for Windows, which can result in the install command being shadowed, e.g. for ocamlfind.

### DIFF
--- a/src/state/opamEnv.ml
+++ b/src/state/opamEnv.ml
@@ -307,7 +307,9 @@ let rezip_to_string ?insert z =
   join_var (rezip ?insert z)
 
 let cygwin_non_shadowed_programs =
-  ["bash.exe"; "make.exe"; "sort.exe"; "tar.exe"]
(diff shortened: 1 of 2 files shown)


kit-ty-kate (approved):

src/state/opamEnv.ml
```diff
@@ -307,7 +307,7 @@ let rezip_to_string ?insert z =
   join_var (rezip ?insert z)
 
 let cygwin_non_shadowed_programs =
-  ["bash.exe"; "make.exe"; "sort.exe"; "tar.exe"]
+  ["bash.exe"; "make.exe"; "sort.exe"; "tar.exe"; "install.exe"]
```
rjbou, line 310:
```suggestion
  [ "bash.exe"; "make.exe"; "sort.exe"; "tar.exe";
    "install.exe"; (* from Vim for Windows *)
  ]
```

rjbou (approved):
Thanks!

PR #6196, opened 2024-09-13
Add 2 admin commands to compare 2 package versions

mbarbin:
Add 2 new commands `opam admin {assert-}compare-versions` to compare package versions for sanity checks.

1. `compare-versions` - User interactive.

In this command, we provide two versions, and opam outputs a user-friendly expression with the result of the comparison, spelled out. For example:

```sh
 0.0.9 < 0.0.10
```

2. `assert-compare-versions` - For use in scripts, if needed.

This command is silent, and requires an operator {lt,le,eq}. The result of the command will be encoded in the exit code, to be used in bash conditionals:

exit 0: the comparison holds
exit 1: it doesn't

For example:

```sh
$ opam admin assert-compare-versions 0.0.9 lt 0.0.10
[0]

$ opam admin assert-compare-versions 0.2.2 le 0.2.0
[1]
```

This PR is an alternative to #6124 trying to account for the various feedback received.

### DIFF
--- a/doc/man/opam-admin-topics.inc
+++ b/doc/man/opam-admin-topics.inc
@@ -61,6 +61,26 @@
   (package opam))
 
 (rule
+  (with-stdout-to opam-admin-compare-versions.0 (echo "")))
+(rule
+  (targets opam-admin-compare-versions.1 opam-admin-compare-versions.err)
+  (deps using-built-opam)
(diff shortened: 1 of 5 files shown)


mbarbin:
Will go with #6197 instead.

Issue #6053, opened 2024-06-25
Windows/Cygwin: opam break the fish environement path


Chimrod:
I’m using opam on windows, with the fish environement. 

After the environment intialisation, my terminal broke, not reconignizing any basic commands like `ls`. After digging a bit, It look likes the path where replaced with a mix of unix and windows path.

```
$ echo $PATH
/usr/local/bin /bin /cygdrive/c/Program Files (x86)/Common Files/Oracle/Java/java8path /cygdrive/c/Program Files (x86)/Common Files/Oracle/Java/javapath /cygdrive/c/Windows/System32 /cygdrive/c/Windows /cygdrive/c/Windows/System32/wbem /cygdrive/c/Windows/System32/WindowsPowerShell/v1.0 /cygdrive/c/Windows/System32/OpenSSH /cygdrive/c/Program Files/PuTTY /cygdrive/c/Program Files/Azure Data Studio/bin /cygdrive/c/Program Files/dotnet /cygdrive/c/Program Files (x86)/GnuPG/bin /cygdrive/c/Program Files (x86)/Microsoft SQL Server/160/DTS/Binn /cygdrive/c/Program Files/PowerShell/7 /cygdrive/c/Program Files/Python39 /cygdrive/c/Users/sdai/AppData/Local/Microsoft/WindowsApps /usr/bin /cygdrive/c/Program Files/Azure Data Studio/bin /cygdrive/c/Users/sdai/AppData/Local/Pandoc /cygdrive/c/Program Files/Neovim/bin /cygdrive/c/Users/sdai/AppData/Local/Programs/GnuPG/bin
```

Path variable after sourcing 

```
$ source '/cygdrive/d/home/.opam/opam-init/env_hook.fish'
$ echo $PATH
[...]

PR #6198, opened 2024-09-16
Fix empty updates in dynamic init scripts

dra27:
An update of the form:

```
FOO += ""
```

is not supposed to do anything to the environment (#5350). However, these updates were being written to the init scripts in opam-init (`variables.sh`, etc.). This PR fixes that, and consequently the problem with mingw-w64-shims reported in https://github.com/dra27/mingw-w64-shims/issues/2 and https://github.com/ocaml/opam/issues/6053#issuecomment-2351043018.

Before this PR, for a switch with mingw-w64-shims and only the x86_64 installation, variables.sh has:

```sh
# Updated by package mingw-w64-shims
PATH='C:\Devel\Roots\test-empty\.cygwin\root\usr\x86_64-w64-mingw32\sys-root\mingw\bin':"$PATH"; export PATH;
# Updated by package mingw-w64-shims
PATH='':"$PATH"; export PATH;
```
which is wrong, but after just:
```
# Updated by package mingw-w64-shims
PATH='C:\Devel\Roots\test-empty\.cygwin\root\usr\x86_64-w64-mingw32\sys-root\mingw\bin':"$PATH"; export PATH;
```
Both before and after, the switch's `environment` file correctly has both entries:
```
PATH    +=      C:\\Devel\\Roots\\test-empty\\.cygwin\\root\\usr\\x86_64-w64-mingw32\\sys-root\\mingw\\bin      Updated\ by\ package\ mingw-w64-shims
PATH    +=      @       Updated\ by\ package\ mingw-w64-shims
```


kit-ty-kate (commented):
modulo the fixup commit i've just pushed (feel free to remove it or squash it if you disagree or agree with it), that looks ok to me, although i'm not an expert in this part of the code.

Could you add a test in the reftests to avoid future regressions? From afar it looks reasonably easy to craft a test for this particular case

kit-ty-kate (approved):
I added a test showing the behaviour and rebased the branch. LGTM

tests/reftests/env.test
```diff
@@ -651,3 +651,29 @@ OPAM_PACKAGE_NAME=another-package
 OPAM_PACKAGE_VERSION=another-version
 OPAM_SWITCH_PREFIX=${BASEDIR}/OPAM/bd
 PATH=${BASEDIR}/OPAM/bd/bin:another-path
+### : The effect of env updates containing empty strings on variables.sh
```
rjbou, line 654: I think that this should be in its own test file. It is indeed environment handling, but more specifically variables.sh environment variable handling, we can begin a test that contains this test in order to update it with other tests specific to variables.sh.

rjbou (approved):

master_changes.md
```diff
@@ -100,6 +100,7 @@ users)
 
 ## Env
   * Add the `OPAMSOLVERTOLERANCE` environment variable to allow users to fix solver timeouts for good [#5510 @kit-ty-kate - fix #3230]
+  * Don't write empty environment update segments to variables.sh (`FOO += ""` no longer adds `FOO='':"$FOO"; export FOO;`) [#6198 @dra27]
```
rjbou, line 103:
```suggestion
  * [BUG] Don't write empty environment update segments to variables.sh (`FOO += ""` no longer adds `FOO='':"$FOO"; export FOO;`) [#6198 @dra27]
```

kit-ty-kate:
Thanks!

PR #6199, opened 2024-09-16
Return parse errors for 2.2/2.3 formats

### LINKED ISSUE #6188: opam 2.2 misdiagnosing parse errors for `opam-version: "2.2"`
Slightly academic, as there were no changes for 2.2, but this affects #5665.

For
```
opam-version: "2.1"
GARBAGE
```
`opam lint` gives a parse error, which is correct.

For
```
opam-version: "2.2"
GARBAGE
```
it gives "unsupported or missing file format version; should be 2.0 or older", which is incorrect (it should still be a path error). The issue is failing to handling opam-file-format 2.1.x signalling a parse error for a future opam syntax.

dra27:
Fixes #6188 and addresses the note in https://github.com/ocaml/opam/pull/5665#discussion_r1742278703

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -93,6 +93,7 @@ users)
 
 ## Opamfile
   * Make all writes atomic [#5489 @kit-ty-kate]
+  * Propagate future opamfile parse errors correctly [#6199 @dra27]
 
 ## External dependencies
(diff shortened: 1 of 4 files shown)


kit-ty-kate (approved):

kit-ty-kate:
Thanks a lot!

PR #6203, opened 2024-09-20
Add 2.3.0~alpha1 to the install scripts

kit-ty-kate:


kit-ty-kate:
Tested on Linux/arm64, macOS/arm64 and Windows/x86_64

PR #6207, opened 2024-09-21
depext: Add Altlinux family support

RiderALT:
using rpm for a package manager and apt-rpm (fork of the apt fron debian wuth RPM support) for the repository manager.

### DIFF
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -115,6 +115,7 @@ end
    below pattern matching *)
 type families =
   | Alpine
+  | Altlinux
   | Arch
   | Centos
   | Cygwin
@@ -174,6 +175,7 @@ let family ~env () =
   | Some family ->
     match family with
     | "alpine" | "wolfi" -> Alpine
+    | "altlinux" -> Altlinux
     | "amzn" | "centos" | "fedora" | "mageia" | "oraclelinux" | "ol"
     | "rhel" -> Centos
     | "archlinux" | "arch" -> Arch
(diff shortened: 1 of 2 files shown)


src/state/opamSysInteract.ml
```diff
@@ -1142,6 +1146,7 @@ let update ?(env=OpamVariable.Map.empty) config =
   let cmd =
     match family with
     | Alpine -> Some (`AsAdmin "apk", ["update"])
+    | Altlinux -> Some (`AsAdmin "apt-get", ["update"])
```
kit-ty-kate, line 1149: This can be merged with the `Debian` case below

kit-ty-kate (commented):
Thanks for this PR.

This change looks reasonable modulo the one suggestion.
If you have some spare time, would you be able to add altlinux to our testsuite? (see the files `.github/workflows/depexts.yml` and `.github/scripts/depexts/generate-actions.sh`) It should just be a case of either merging the debian and altlinux case or copy-pasting. Altlinux is available as `alt` on Docker Hub.

If you don't have time we'll do it at some point in the future

rjbou:
It is not possible ftm to add a depext test as it there is no conf package that have Altlinux support on main opam repository. A PR is added for when it will be the case #6277

kit-ty-kate (approved):

kit-ty-kate:
Thanks!

RiderALT:
Thank you very much for the merging, unfortunately I could not get to the Docker and make tests for altlinux, but I remember about this task.

rjbou:
No worries, the docker test PR is already opened, see https://github.com/ocaml/opam/pull/6207#issuecomment-2444825451

PR #6208, opened 2024-09-23
Simplify the making of stripped binaries by introducing the 'make opam-stripped' target

kit-ty-kate:
Suggested in https://github.com/ocaml/opam/pull/6204#discussion_r1771256137


dra27 (approved):
LGTM, thanks!

PR #6209, opened 2024-09-24
Fix opam install on a local directory not updating pinned packages' metadata

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/5567
Fixes https://github.com/ocaml/opam/issues/4507
Queued on #6438 

I have no idea why `for_view` was introduced in https://github.com/ocaml/opam/commit/9b2a67f36d9d0e20b734ff85ddef470268aabb25 but i've tested `opam show` and i'm not seeing any breaking nor does the testsuite.

The bypass here was causing local packages to be put in the `already_pinned` set. However this set is only used to update the `<switch>/.opam-switch/sources/<pkg>` directory, not the `<...>/overlay` directory as the assumption with this set is that the package description hasn't change.
Removing that bypass makes the function much more consistent.

The `reinstall` bit is a bit easier to understand: when using `--deps-only` or `--show-action`, the codepath goes through `simulate_local_pinning` which didn't take into account that some packages being pinned are already pinned. In that case the `reinstall` field must be updated to ensure the rest of the code knows that package should be reinstalled.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -29,6 +29,7 @@ users)
 ## Install
   * Do not keep the build directory of the pinned packages [#6436 @kit-ty-kate]
   * [BUG] Fix `opam install <local_dir>` not updating and storing pinned packages' metadata [#6209 @kit-ty-kate - fix #5567]
+  * [BUG] Fix `opam install --deps-only/--show-action <local_dir>` not updating (without storing) pinned packages' metadata [#6209 @kit-ty-kate - fix #5567]
 
 ## Build (package)
   * Patches are now applied using the `patch` OCaml library instead of GNU Patch [#5892 @kit-ty-kate - fix #6019 #6052]
@@ -251,6 +252,7 @@ users)
   * `OpamArg.InvalidCLI`: export exception [#6150 @rjbou]
   * `OpamArg`: export `require_checksums` and `no_checksums`, that are shared with `build_options` [#5563 @rjbou]
   * `OpamArg.hash_kinds`: was added [#5960 @kit-ty-kate]
+  * `OpamAuxCommands.{simulate_autopin,autopin ~simulate:true}`: now updates the `reinstall` field of the returned `switch_state` if necessary [#6209 @kit-ty-kate]
   * `OpamRepositoryCommand.switch_repos`: expose the function [#5014 @kit-ty-kate]
   * `OpamLockCommand.lock_opam`: add `~keep_local` argument to add local pins to pin-depends (and not resolve them) [#6411 @rjbou]
   * `OpamLockCommand.lock_opam`: make the `?only_direct` argument non-optional [#6411 @kit-ty-kate]
--- a/master_changes.md
+++ b/master_changes.md
@@ -28,6 +28,7 @@ users)
 
 ## Install
   * Do not keep the build directory of the pinned packages [#6436 @kit-ty-kate]
+  * [BUG] Fix `opam install <local_dir>` not updating and storing pinned packages' metadata [#6209 @kit-ty-kate - fix #5567]
 
 ## Build (package)
   * Patches are now applied using the `patch` OCaml library instead of GNU Patch [#5892 @kit-ty-kate - fix #6019 #6052]
--- a/master_changes.md
+++ b/master_changes.md
@@ -203,6 +203,7 @@ users)
   * Add admin filter subcommand test [#6166 @rjbou]
   * Add a test showing the behaviour of opam install when a local opam file changes while being pinned [#6209 @kit-ty-kate]
   * Add pin test to show stored overlay opam files [#6209 @rjbou]
+  * Add show test to highlight precedence of opam file selection and check that if an opam file is given it is always this one that is taken [#6209 @rjbou]
 
 ### Engine
 
--- a/master_changes.md
+++ b/master_changes.md
@@ -202,6 +202,7 @@ users)
   * Add tests for `--keep-build-dir` and `OPAMKEEPBUILDDIR` [#6436 @rjbou @kit-ty-kate]
   * Add admin filter subcommand test [#6166 @rjbou]
   * Add a test showing the behaviour of opam install when a local opam file changes while being pinned [#6209 @kit-ty-kate]
+  * Add pin test to show stored overlay opam files [#6209 @rjbou]
 
 ### Engine
 
--- a/master_changes.md
+++ b/master_changes.md
@@ -201,6 +201,7 @@ users)
   * Add a test for packages with subpath in a repository [#6439 @rjbou]
   * Add tests for `--keep-build-dir` and `OPAMKEEPBUILDDIR` [#6436 @rjbou @kit-ty-kate]
   * Add admin filter subcommand test [#6166 @rjbou]
+  * Add a test showing the behaviour of opam install when a local opam file changes while being pinned [#6209 @kit-ty-kate]
 
 ### Engine
 
--- a/src/client/opamAuxCommands.ml
+++ b/src/client/opamAuxCommands.ml
@@ -401,6 +401,29 @@ let simulate_local_pinnings ?quiet ?(for_view=false) st to_pin =
            st.switch_global st.switch st.switch_config ~pinned
            ~opams:local_opams)
     );
+    reinstall = lazy (
+      let open OpamPackage.Set.Op in
+      let installed_pinned = st.pinned %% st.installed in
+      OpamPackage.Set.fold (fun pinned_pkg reinstall ->
+          match
+            OpamPackage.Set.find_opt
+              (fun pkg ->
+                 OpamPackage.Name.equal
+                   (OpamPackage.name pinned_pkg)
+                   (OpamPackage.name pkg))
+              local_packages
+          with
+          | None -> reinstall
+          | Some local_pkg ->
+            let old_opam = OpamPackage.Map.find pinned_pkg st.installed_opams in
+            let new_opam = OpamPackage.Map.find local_pkg local_opams in
+            if OpamFile.OPAM.effectively_equal old_opam new_opam then
+              reinstall
+            else
+              OpamPackage.Set.add local_pkg
+                (OpamPackage.Set.remove pinned_pkg reinstall))
+        installed_pinned (Lazy.force st.reinstall)
+    );
     pinned;
   } in
   st, local_packages
--- a/src/client/opamAuxCommands.ml
+++ b/src/client/opamAuxCommands.ml
@@ -250,7 +250,7 @@ let resolve_locals ?(quiet=false) ?locked ?recurse ?subpath
              (OpamUrl.to_string nf.pin.pin_url))
           duplicates)
 
-let autopin_aux st ?quiet ?(for_view=false) ?recurse ?subpath ?locked
+let autopin_aux st ?quiet ?recurse ?subpath ?locked
     atom_or_local_list =
   let to_pin, atoms =
     resolve_locals ?quiet ?recurse ?subpath ?locked atom_or_local_list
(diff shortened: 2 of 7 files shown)


src/client/opamAuxCommands.ml
```diff
@@ -386,6 +386,29 @@ let simulate_local_pinnings ?quiet ?(for_view=false) st to_pin =
...
+          match
+            OpamPackage.Set.find_opt
+              (fun pkg ->
+                 OpamPackage.Name.equal
+                   (OpamPackage.name pinned_pkg)
+                   (OpamPackage.name pkg))
+              local_packages
+          with
+          | None -> reinstall
+          | Some to_pin ->
```
rjbou, line 402: It is already pinned (simulate state), not `to_pin`

master_changes.md
```diff
@@ -25,6 +25,7 @@ users)
 
 ## Install
   * Fix `opam install <local_dir>` not updating pinned packages' metadata [#6209 @kit-ty-kate - fix #5567]
+  * Fix `opam install --deps-only/--show-action <local_dir>` not updating pinned packages' metadata [#6209 @kit-ty-kate - fix #5567]
```
rjbou, line 28:
```suggestion
  * Fix `opam install <local_dir>` not updating and storing pinned packages' metadata [#6209 @kit-ty-kate - fix #5567]
  * Fix `opam install --deps-only/--show-action <local_dir>` not updating without storing pinned packages' metadata [#6209 @kit-ty-kate - fix #5567]
```
For commits messages too. it is a proposal, there is probably a better wording, I mainly wanted to highlight the difference.

rjbou, line 28: Also, the PR name should change to something more general, like *fix opam install sometimes not updating pinned packages' metadata*, it is no more only about `local dir as a target.

rjbou (changes requested):
I reviewed commit per commit, it's better to look at comments per commit.
I'm not sure about the third commit (reinstall packages), it raises a good question. we want to update the opam file internally, but do we want to upgrade the package itself? Usually, when users use `--deps-only` they don't want to launch an install of given package (no need, or don't want to double compile it). When the package is already installed, there is the "we want to have a consistent state" that leads us to want to upgrade the package. From CLI perspective, it is more consistent to not upgrade the package.
Maybe a note or warning to highlight the situation *package updated internally, if you really want to reinstall it, launch opam install pkg* is better than a full reinstall triggered by `--deps-only`.

master_changes.md
kit-ty-kate: why is it not only about local dir?

rjbou: My mistake, I meant that it is not only the command `opam install <local-dir>` but also combined with option `--dep-only` and `--show-action` that is fixed here. I updated the proposal.

tests/reftests/pin.test
```diff
@@ -399,3 +399,101 @@ echo GARBAGE>>"$1"
...
+### <pkg:dependency.1>
+opam-version: "2.0"
+### <pkg:dependency.2>
+opam-version: "2.0"
+### <pkg:dependency.3>
+opam-version: "2.0"
+### mkdir pin-change
+### git -C pin-change init -q --initial-branch=master
+### git -C pin-change config core.autocrlf false
+### <pin-change/opam>
```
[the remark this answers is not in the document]
kit-ty-kate: done

master_changes.md
kit-ty-kate: what about "Fix opam install on a local directory not updating pinned packages' metadata" ?

tests/reftests/pin.test
```diff
@@ -463,23 +464,41 @@ opam-version: "2.0"
...
+           warning 35: Missing field 'homepage'
+           warning 36: Missing field 'bug-reports'
+             error 57: Synopsis must not be empty
+           warning 68: Missing field 'license'
 [NOTE] Ignoring uncommitted changes in ${BASEDIR}/pin-change (`--working-dir' not specified or specified with no argument).
 [pin-change.dev] synchronised (no changes)
-### opam remove dependency
-[NOTE] dependency is not installed.
-
-Nothing to do.
```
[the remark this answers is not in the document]
kit-ty-kate: honestly, not sure. Probably an artfact of debugging. I've removed them from the previous commit as they do not provide any information

tests/reftests/pin.test
```diff
@@ -463,23 +464,41 @@ opam-version: "2.0"
...
-[NOTE] Package pin-change is already installed (current version is dev).
-### opam remove dependency
-[NOTE] dependency is not installed.
+The following actions will be performed:
+=== recompile 1 package
+  - recompile pin-change dev (pinned)
+=== install 1 package
+  - install   dependency 1            [required by pin-change]
 
-Nothing to do.
```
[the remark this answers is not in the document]
kit-ty-kate: same as above

src/client/opamAuxCommands.ml
```diff
@@ -303,10 +303,7 @@ let autopin_aux st ?quiet ?(for_view=false) ?recurse ?subpath ?locked
               | _ -> false)
            | None -> false)
           &&
-          (* For `opam show`, we need to check does the opam file changed to
-             perform a simulated pin if so *)
-          (not for_view ||
-           match
+          (match
              OpamSwitchState.opam_opt st pinned_pkg,
              OpamFile.OPAM.read_opt nf.pin.pin_file
```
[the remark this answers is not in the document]
kit-ty-kate: good catch, handling all those cases managed to fix that commit and remove the regression

src/client/opamAuxCommands.ml
kit-ty-kate: good point. renamed to `local_pkg`

src/client/opamAuxCommands.ml
```diff
@@ -303,16 +303,31 @@ let autopin_aux st ?quiet ?(for_view=false) ?recurse ?subpath ?locked
...
+             in
+             let opam =
+               OpamFile.OPAM.with_url
+                 (OpamFile.URL.create ?subpath:nf.pin.pin_subpath nf.pin.pin_url)
+                 opam
+             in
+             let opam =
+               (* This is required to avoid the case where locked opam files were
+                  stored with the added `available: opam-version >= "2.1"` *)
+               OpamFile.OPAM.with_available (OpamFile.OPAM.available opam0) opam
```
rjbou, line 328: It worth adding a comment where these fields are added for the internal stored opam, in order to update this piece of code when it is changed.
I don't know if we can have a test to follow that, keep the consistency between these fields added, and the ones checked here.

tests/reftests/pin.test
```diff
@@ -399,3 +399,105 @@ echo GARBAGE>>"$1"
...
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> retrieved pin-change.dev  (git+file://${BASEDIR}/pin-change#master)
+-> installed pin-change.dev
+Done.
+### <pin:pin-change/opam>
+opam-version: "2.0"
+depends: "dependency" {= "1"}
+### ::: behaviour when the package is not pinned
+### opam install ./pin-change
+pin-change is now pinned to git+file://${BASEDIR}/pin-change#master (version dev)
```
rjbou, line 446: We don't want the repin note, it is not repinned to another url (no source fetching as source_pin does), only the opam file updated/changed.

tests/reftests/pin.test
```diff
@@ -399,3 +399,105 @@ echo GARBAGE>>"$1"
...
+[NOTE] Ignoring uncommitted changes in ${BASEDIR}/pin-change (`--working-dir' not specified or specified with no argument).
```
rjbou, line 447: If we have a more fine grained handling of pin update (opam file vs source update), we can even remove that note if only the uncommitted opam file is updated, as now we have the information.

src/client/opamAuxCommands.ml
```diff
@@ -303,16 +303,31 @@ let autopin_aux st ?quiet ?(for_view=false) ?recurse ?subpath ?locked
               | _ -> false)
            | None -> false)
           &&
-          (* For `opam show`, we need to check does the opam file changed to
```
rjbou, line 306: It worth adding a test for `opam show` that highlights that there is no change in the behaviour

tests/reftests/pin.test
kit-ty-kate: i general i would say yes, however this is not the role of this PR. This behaviour is the one that opam currently has and this PR is not changing it.
See for example these two lines where a foo-core and foo-format were repinned and the "is now pinned to" message still appears:
- https://github.com/ocaml/opam/blob/f3466d96f876c522c75b543628e45898dc42ed7d/tests/reftests/autopin.test#L231
- https://github.com/ocaml/opam/blob/f3466d96f876c522c75b543628e45898dc42ed7d/tests/reftests/autopin.test#L259

tests/reftests/pin.test
kit-ty-kate: same as with https://github.com/ocaml/opam/pull/6209/files#r1815454356 i believe. I think those two improvements can be their own tickets to be fixed later.

master_changes.md
kit-ty-kate: PR title updated

src/client/opamAuxCommands.ml
```diff
@@ -303,10 +303,7 @@ let autopin_aux st ?quiet ?(for_view=false) ?recurse ?subpath ?locked
...
-             perform a simulated pin if so *)
-          (not for_view ||
```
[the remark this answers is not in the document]
rjbou: Discussed in a meeting, we need to have a repin, given the current code

kit-ty-kate:
rebased on master

src/client/opamAuxCommands.ml
rjbou: :+1:

src/client/opamAuxCommands.ml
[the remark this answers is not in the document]
rjbou: :+1:

src/client/opamAuxCommands.ml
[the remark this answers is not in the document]
rjbou: :+1:

tests/reftests/pin.test
rjbou: :+1:

rjbou:
> I'm not sure about the third commit (reinstall packages), it raises a good question. we want to update the opam file internally, but do we want to upgrade the package itself? Usually, when users use `--deps-only` they don't want to launch an install of given package (no need, or don't want to double compile it). When the package is already installed, there is the "we want to have a consistent state" that leads us to want to upgrade the package. From CLI perspective, it is more consistent to not upgrade the package.

We need to upgrade the package to keep in sync the switch.

src/client/opamAuxCommands.ml
```diff
@@ -303,16 +303,33 @@ let autopin_aux st ?quiet ?(for_view=false) ?recurse ?subpath ?locked
...
-           match
+          (match
              OpamSwitchState.opam_opt st pinned_pkg,
              OpamFile.OPAM.read_opt nf.pin.pin_file
            with
            | Some opam0, Some opam ->
+             (* Keep synchronised with [OpamFile.OPAM.pp_raw_fields] added
```
rjbou, line 311: i'm wondering if it would be better to have a function in `OpamFile.OPAM` that does all these additions in order to compare (or just returns the new opam file?), like that it's easier to keep synchronised as they would be in the same module / near in the code.

kit-ty-kate: I've added two fixup commits (to cleanup of course) exploring that idea.
The first one rewrites the comparison to be more efficient (2 read instead of 4), and more correct.
In particular the `available` field was bugging me. The `pp_minimal_opam_version` function would rewrite the available field but i can't find any good reason for this.

This function was originally added in https://github.com/ocaml/opam/commit/1d324776a846e1b12b4c86f4f3ec8c829cce0068 + https://github.com/ocaml/opam/commit/ca84c3ab11c1857d93ce62daee7f765e43152b17 but i can't find any reasoning for this change aside from "consistency". I personally think this makes the whole thing less consistent by introducing differences between original file and rewritten files, which is nearly impossible to detangle.

For this reason i chose to remove this part of the code which makes for a needed simplification, in my opinion.
What do you think?

tests/reftests/pin.test
```diff
@@ -1068,3 +1068,73 @@ opam-version: "2.0"
 
 No solution found, exiting
 # Return code 20 #
+### :C:e: Make sure opam install works when the opam file changes and the package is pinned
```
rjbou, line 1069: Depending on the order of merge, these header may need a change in the last rebase (`:C:e:` is also used in another PR)

master_changes.md
```diff
@@ -27,6 +27,8 @@ users)
 ## Actions
 
 ## Install
+  * Fix `opam install <local_dir>` not updating and storing pinned packages' metadata [#6209 @kit-ty-kate - fix #5567]
```
rjbou, line 30:
```suggestion
  * [BUG] Fix `opam install <local_dir>` not updating and storing pinned packages' metadata [#6209 @kit-ty-kate - fix #5567]
```

master_changes.md
```diff
@@ -27,6 +27,8 @@ users)
...
+  * Fix `opam install --deps-only/--show-action <local_dir>` not updating (without storing) pinned packages' metadata [#6209 @kit-ty-kate - fix #5567]
```
rjbou, line 31:
```suggestion
  * [BUG] Fix `opam install --deps-only/--show-action <local_dir>` not updating (without storing) pinned packages' metadata [#6209 @kit-ty-kate - fix #5567]
```

src/client/opamAuxCommands.ml
```diff
@@ -399,6 +399,29 @@ let simulate_local_pinnings ?quiet ?(for_view=false) st to_pin =
            st.switch_global st.switch st.switch_config ~pinned
            ~opams:local_opams)
     );
+    reinstall = lazy (
```
rjbou, line 402: It may worth having an entry in the API changelog ?

src/client/opamAuxCommands.ml
```diff
@@ -293,27 +293,37 @@ let autopin_aux st ?quiet ?(for_view=false) ?recurse ?subpath ?locked
...
-           with
-           | Some opam0, Some opam ->
-             let opam = OpamFile.OPAM.with_locked_opt nf.pin.pin_locked opam in
-             OpamFile.OPAM.equal opam0 opam
-           | _, _ -> false)
+          match OpamSwitchState.opam_opt st pinned_pkg with
+          | Some opam0 ->
+            (match OpamFile.OPAM.read_opt nf.pin.pin_file with
+             | Some opam ->
+               let opam = OpamFile.OPAM.with_locked_opt nf.pin.pin_locked opam in
```
rjbou, line 300: Commit 449436922834aa09e89efa58153a48c9fd72ab5c lost in the PR split

rjbou (approved):

src/client/opamAuxCommands.ml
kit-ty-kate: It was discarded on purpose as to me https://github.com/ocaml/opam/pull/6438 removes the actual change made to the opam definition that we have to think about. But i've readded it, i don't mind either way.

PR #6210, opened 2024-09-25
perf: optimize OpamString.split

kit-ty-kate:
this cuts the time spent on parsing `pacman -Si` significantly down

Commit by @c-cube split from https://github.com/ocaml/opam/pull/5757
Queued on #6212


kit-ty-kate:
After trying a couple of things, it looks like `opam show dune` shows the significant performance improvement (25% on my local Fedora), however Debian (the system used for benchmark) shows less improvement (6 or 7%). I'll add a benchmark nonetheless and see if it can be reproduced in a better controlled environment

kit-ty-kate:
ok i should've just started with benchmarking just the function in isolation. It now shows a 45x speedup

rjbou (approved):
Thanks! :rocket:

PR #6212, opened 2024-09-25
Add benchmarks for opam show and String.split

kit-ty-kate:
This PR is meant to show the performance gain from https://github.com/ocaml/opam/pull/6210

### DIFF
--- a/tests/bench/bench.ml
+++ b/tests/bench/bench.ml
@@ -99,6 +99,24 @@ let () =
     launch (fmt "%s switch create six --empty" bin);
     time_cmd ~exit:0 (fmt "%s list --installed --short --safe --color=never ocp-indent ocp-index merlin" bin)
   in
+  launch (fmt "%s switch create seven --empty" bin);
+  launch (fmt "%s install -y --fake dune" bin);
+  let time_show_installed =
+    Gc.compact ();
+    time_cmd ~exit:0 (fmt "%s show dune" bin)
+  in
+  let time_show_with_depexts =
+    Gc.compact ();
+    time_cmd ~exit:0 (fmt "%s show conf-llvm" bin)
+  in
+  let time_show_raw =
+    Gc.compact ();
+    time_cmd ~exit:0 (fmt "%s show --raw conf-llvm" bin)
+  in
+  let time_show_precise =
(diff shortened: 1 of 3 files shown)


kit-ty-kate:
https://github.com/ocaml/opam/pull/6210 is unable to show any gain. I'll convert this PR to a draft while this is being examined.

tests/bench/bench.ml
```diff
@@ -99,6 +99,39 @@ let () =
...
+      let ic = Stdlib.open_in_bin "/home/opam/all-opam-content" in
+      let rec loop files =
+        match Stdlib.input_line ic with
+        | file -> loop (file :: files)
+        | exception End_of_file -> files
+      in
+      loop []
+    in
+    let before = Unix.gettimeofday () in
+    List.iter (fun line -> ignore (OpamStd.String.split line ' ')) lines;
```
rjbou, line 132: Wouldn't be more precise if it is run 50, 100 times ? Current run/time is less than 1 second.

kit-ty-kate: the problem is that it is only 1s in #6210. In this PR it takes 45s which would be too long to run 50 or 100 times.

kit-ty-kate: i can run it 10 times maybe? That would be 4.5mins for this benchmark which seems still reasonable

rjbou: yes, the idea was to not have to diff times under 1s, to have more "accurate" data.

rjbou: The 5min bench is just for few launches

kit-ty-kate: done

tests/bench/bench.ml
```diff
@@ -148,6 +185,31 @@ let () =
...
+          "value": %f,
+          "units": "secs"
+        },
+        {
+          "name": "opam show --raw pkgname.version",
+          "value": %f,
+          "units": "secs"
+        },
+        {
+          "name": "OpamStd.String.split",
```
rjbou, line 210:
```suggestion
          "name": "OpamStd.String.split amortised over 10 runs",
```

rjbou (approved):

PR #6213, opened 2024-09-26
Fix the install cache storing the wrong version of the opam file after a build failure

kit-ty-kate:
Upgrades from 2.2 to 2.3 can trigger issues where a repository forgot to add the missing extra-files field.
opam tells the user they should upgrade as the opam file changed (as extra-files was added automatically), then the build fails and upon fixing the error, opam still says the package is to be rebuilt because the previous version of the opam file was stored in the install cache.

This should be backported to the 2.3 branch as it can easily be triggered during the upgrade from 2.2 to 2.3.
Backported to 2.3 via https://github.com/ocaml/opam/pull/6214

Potentially related to https://github.com/ocaml/opam/pull/5922

### DIFF
--- a/src/client/opamSolution.ml
+++ b/src/client/opamSolution.ml
@@ -824,15 +824,46 @@ let parallel_apply t
 
   (* 2/ Display errors and finalize *)
 
-  OpamSwitchState.Installed_cache.save
-    (OpamPath.Switch.installed_opams_cache t.switch_global.root t.switch)
-    (OpamPackage.Set.fold (fun nv opams ->
-         let opam =
-           OpamSwitchState.opam t nv |>
-           OpamFile.OPAM.with_metadata_dir None
-         in
-         OpamPackage.Map.add nv opam opams)
-        t.installed OpamPackage.Map.empty);
+  let save_installed_cache failed =
+    OpamSwitchState.Installed_cache.save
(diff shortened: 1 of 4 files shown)


src/client/opamSolution.ml
```diff
@@ -824,15 +824,45 @@ let parallel_apply t
...
+      (OpamPath.Switch.installed_opams_cache t.switch_global.root t.switch)
+      (OpamPackage.Set.fold (fun nv opams ->
+           let pkg_failed =
+             List.exists (function
+                 | `Fetch ps -> List.for_all (OpamPackage.equal nv) ps
+                 | `Build p
+                 | `Change (_, _, p)
+                 | `Install p
+                 | `Reinstall p
+                 | `Remove p -> OpamPackage.equal nv p)
```
rjbou, line 838: The failure is specific to build and fetch errors, maybe adding a comment have that information. It's permit to explain why we need to have that check on install cache update.

kit-ty-kate: I don't really think this part of the code needs a comment but i do agree that the section (filtering as a whole) needed one so i added one. What do you think?

rjbou: :+1:

rjbou (approved):
lgtm! If my test part is ok, good to merge.

Issue #6215, opened 2024-09-30
getconf: LONG_BIT: unknown variable


mndrix:
On OpenBSD 7.5 amd64 running opam version 2.3.0~alpha1, `opam upgrade` produces the following error:

```
getconf: LONG_BIT: unknown variable
'opam upgrade' failed.
Fatal error:
End_of_file
```

The output of `opam config report 2>&1` encounters the same error:

```
# opam config report
# opam-version         2.3.0~alpha1
# self-upgrade         no
getconf: LONG_BIT: unknown variable
# read-state           End_of_file
```

It doesn't look like I'll get around to sending a patch this week, but maybe this weekend :crossed_fingers:

dra27:
Cross-referencing https://github.com/ocaml/opam/pull/5950

kit-ty-kate:
It looks like OpenBSD is the only unix platform i could find where `uname -m` actually returns the "kernel architecture" instead of the "machine architecture" so the call to `getconf LONG_BIT` or equivalent is not required here.

Source:
- https://man.openbsd.org/uname.1 (tested in qemu)
- https://man.netbsd.org/uname.1
- https://man.freebsd.org/cgi/man.cgi?query=uname
- https://man.dragonflybsd.org/?command=uname&section=1
- https://man.archlinux.org/man/uname.1.en
- https://ss64.com/mac/uname.html

I'll open a PR in the next hour

kit-ty-kate:
That took a bit longer than expected as i had an idea on how to improve related code and discovered a bug elsewhere.
Anyway, this hasn't been tested outside of Linux yet but i believe it should work so feel free to test it: https://github.com/ocaml/opam/pull/6217

mndrix:
Confirming that #6217 fixes this problem for me

PR #6216, opened 2024-09-30
Fix the compilation of opam on Windows with OCaml >= 5.0 (again)

kit-ty-kate:
It seems that after a couple of rebase of #6189 we forgot to commit one change to `OpamStubs.ocaml5.ml` which actually means that the file doesn't compile (since the Windows+OCaml5 combination wasn't tested in our CI).
This here PR should fix this issue for good.

Backported to 2.3 in https://github.com/ocaml/opam/pull/6224

PR #6217, opened 2024-09-30
Use a C stub to call the uname function from the C standard library instead of calling the uname POSIX command

kit-ty-kate:
~Queued on top of #6216~
~F.i.x.e.s #6215~

~For the backport to 2.3, do we want to backport the entire PR or just the OpenBSD fix?~
~Partially backported to 2.3 in https://github.com/ocaml/opam/pull/6228~

~Queued on #6230~ merged

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -100,6 +100,9 @@ users)
 
 ## Internal
 
+## Internal: Unix
+  * Use a C stub to call the `uname` function from the C standard library instead of calling the `uname` POSIX command [#6217 @kit-ty-kate]
+
 ## Internal: Windows
 
 ## Test
@@ -132,6 +135,9 @@ users)
 
 ## opam-core
   * `OpamStd.Sys.{get_terminal_columns,uname,getconf,guess_shell_compat}`: Harden the process calls to account for failures [#6230 @kit-ty-kate - fix #6215]
-  * `OpamStd.Sys.{uname,getconf}`: now accepts only one argument as parameter, as per their documentation [#6230 @kit-ty-kate]
+  * `OpamStd.Sys.getconf`: was removed, replaced by `get_long_bit` [#6217 @kit-ty-kate]
+  * `OpamStd.Sys.get_long_bit`: was added, which returns the output of the `getconf LONG_BIT` command [#6217 @kit-ty-kate]
+  * `OpamStd.Sys.uname`: now returns the memoized result of the `uname` function from the C standard library [#6217 @kit-ty-kate]
+  * `OpamStd.Sys.get_freebsd_version`: was added, which returns the output of the `uname -U` command [#6217 @kit-ty-kate]
   * `OpamStubs.get_stdout_ws_col`: new Unix-only function returning the number of columns of the current terminal window [#6244 @kit-ty-kate]
   * `OpamSystem`: add `is_archive_from_string` that does the same than `is_archive` but without looking at the file, only analysing the string (extension) [#6219 @rjbou]
--- a/src/core/opamStd.ml
+++ b/src/core/opamStd.ml
@@ -991,20 +991,17 @@ module OpamSys = struct
 
   let etc () = "/etc"
 
-  let memo_command =
-    let memo = Hashtbl.create 7 in
-    fun cmd arg ->
-      try Hashtbl.find memo (cmd, arg) with Not_found ->
-        let r =
-          match process_in cmd [arg] with
-          | None -> None
-          | Some x -> Some (OpamString.strip x)
-        in
-        Hashtbl.add memo (cmd, arg) r;
-        r
+  let uname =
+    let uname = lazy (OpamStubs.uname ()) in
+    fun () ->
+      Lazy.force uname
 
-  let uname = memo_command "uname"
-  let getconf = memo_command "getconf"
+  (* We need to call [uname] here as the only way to get it
+     in C without calling a process or lookup files is to
+     use [__Freebsd_version] which is a define, so it is set
+     at compile time. *)
+  let get_freebsd_version () = process_in "uname" ["-U"]
+  let get_long_bit () = process_in "getconf" ["LONG_BIT"]
 
   let system =
     let system = Lazy.from_fun OpamStubs.getPathToSystem in
@@ -1026,14 +1023,14 @@ module OpamSys = struct
     let os = lazy (
       match Sys.os_type with
       | "Unix" -> begin
-          match uname "-s" with
-          | Some "Darwin"    -> Darwin
-          | Some "Linux"     -> Linux
-          | Some "FreeBSD"   -> FreeBSD
-          | Some "OpenBSD"   -> OpenBSD
-          | Some "NetBSD"    -> NetBSD
-          | Some "DragonFly" -> DragonFly
-          | _                -> Unix
+          match (uname ()).sysname with
+          | "Darwin"    -> Darwin
+          | "Linux"     -> Linux
+          | "FreeBSD"   -> FreeBSD
+          | "OpenBSD"   -> OpenBSD
+          | "NetBSD"    -> NetBSD
+          | "DragonFly" -> DragonFly
+          | _           -> Unix
         end
       | "Win32"  -> Win32
       | "Cygwin" -> Cygwin
--- a/src/core/opamStd.mli
+++ b/src/core/opamStd.mli
@@ -510,11 +510,15 @@ module Sys : sig
   (** Queried lazily *)
   val os: unit -> os
 
-  (** The output of the command "uname", with the given argument. Memoised. *)
-  val uname: string -> string option
+  (** The output of the command "uname -U". FreeBSD only. Reasoning:
+      https://github.com/ocaml/opam/pull/4274#issuecomment-659280485 *)
+  val get_freebsd_version: unit -> string option
 
-  (** The output of the command "getconf", with the given argument. Memoised. *)
-  val getconf: string -> string option
+  (** The output of the command "getconf LONG_BIT". *)
+  val get_long_bit: unit -> string option
+
+  (** The memoized result of the uname function from the C standard library *)
+  val uname : unit -> OpamStubs.uname
 
   (** Append .exe (only if missing) to executable filenames on Windows *)
   val executable_name : string -> string
--- a/src/core/opamStubs.mli
+++ b/src/core/opamStubs.mli
@@ -171,3 +171,14 @@ val get_stdout_ws_col : unit -> int
     linked with stdout. If stdout isn't linked to any terminal
     (e.g. redirection), then this function will return 0. A valid number
     of columns should be strictly above 0. *)
+
+type uname = {
+  sysname : string; (** uname -s *)
+  release : string; (** uname -r *)
+  machine : string; (** uname -m *)
+}
+(** A subset of the [struct utsname] C structure, as modified by uname(2),
+    converted to OCaml datatypes. *)
+
+val uname : unit -> uname
+(** Unix only. Returns info from uname(2) *)
--- a/src/core/opamCommonStubs.c
+++ b/src/core/opamCommonStubs.c
@@ -37,6 +37,7 @@
 
 #if OCAML_VERSION < 50000
 #define caml_unix_access unix_access
+#define caml_uerror uerror
 #endif
 
 CAMLprim value opam_is_executable(value path)
--- a/src/core/opamStubs.unix.ml
+++ b/src/core/opamStubs.unix.ml
@@ -48,3 +48,4 @@ let getVersionInfo = that's_a_no_no
 let get_initial_environment = that's_a_no_no
 
 external get_stdout_ws_col : unit -> int = "opam_stdout_ws_col"
+external uname : unit -> uname = "opam_uname"
--- a/src/core/opamStubsTypes.ml
+++ b/src/core/opamStubsTypes.ml
@@ -114,6 +114,12 @@ type win32_version_info = {
     (** Non-fixed string table. First field is a pair of Language and Codepage ID. *)
 }
 
+type uname = {
+  sysname : string;
+  release : string;
+  machine : string;
+}
+
 external is_executable : string -> bool = "opam_is_executable"
 (** faccessat on Unix; _waccess on Windows. Checks whether a path is executable
     for the current process. On Unix, unlike Unix.access, this is checked using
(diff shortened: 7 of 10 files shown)


avsm:
Works great on my OpenBSD desktop:

```
$ uname -a
OpenBSD 7.5 GENERIC.MP#82 amd64
```

```
arch              x86_64            # Inferred from system
exe                                 # Suffix needed for executable filenames (Windows)
jobs              1                 # The number of parallel jobs set up in opam configuration
make              gmake             # The 'make' command to use
opam-version      2.4.0~alpha1~dev  # The currently running opam version
os                openbsd           # Inferred from system
os-distribution   openbsd           # Inferred from system
os-family         bsd               # Inferred from system
os-version        7.5               # Inferred from system
root              /home/avsm/.opam2 # The current opam root directory
switch            default           # The identifier of the current switch
sys-ocaml-arch    x86_64            # Target architecture of the OCaml compiler present on your system
sys-ocaml-cc      cc                # Host C Compiler type of the OCaml compiler present on your system
sys-ocaml-libc    libc              # Host C Runtime Library type of the OCaml compiler present on your system
sys-ocaml-system  openbsd           # Target system of the OCaml compiler present on your system
sys-ocaml-version 4.14.1            # OCaml version present on your system independently of opam, if any
```

One test fails:
```
File "tests/reftests/dune.inc", line 2082, characters 0-243:
2082 | (rule
2083 |   (targets opam-repo-143dd2a2f59f5befbf3cb90bb2667f911737fbf8)
2084 |   (action
2085 |    (progn
2086 |     (run mkdir -p %{targets})
2087 |     (run tar -C %{targets} -xzf %{dep:opam-archive-143dd2a2f59f5befbf3cb90bb2667f911737fbf8.tar.gz} --strip-components=1))))
```

but looks unrelated (that `tar` needs to be `gtar` due to the `--strip-components`)

mndrix:
Thanks. This PR fixes #6215 for me.  My opinion may not matter on the patch, but that looked good to me too. I like the switch to `uname` via C stubs.

kit-ty-kate:
turns out we still need to call the `getconf` process. `sysconf(3)` depends on the libc architecture and doesn't have the necessary query with musl (`_SC_LONG_BIT` is glibc only).

We could rely on `Sys.word_size` as an alternative but it would depend on the architecture of the opam binary, not the userspace architecture (e.g. you can call an arm64 binary just fine on Raspberry Pi OS 32bit, which has a 32bit userspace but 64bit kernel)

I've also removed the C stub for FreeBSD as `__FreeBSD_version` is defined at compile-time and does not show the actual runtime version.

src/core/opamStd.mli
```diff
@@ -511,10 +511,12 @@ module Sys : sig
   val os: unit -> os
 
   (** The output of the command "uname", with the given argument. Memoised. *)
-  val uname: string -> string option
+  val uname_cmd: string -> string option
 
   (** The output of the command "getconf", with the given argument. Memoised. *)
   val getconf: string -> string option
+  (** The memoized result of uname(2) *)
+  val uname : OpamStubs.uname Lazy.t
```
rjbou, line 520: It's better to have a `uname: void -> OpamStubs.uname` to ease readability/writing when used

src/core/opamStd.ml
```diff
@@ -995,22 +995,9 @@ module OpamSys = struct
 
   let etc () = "/etc"
 
-  let memo_command =
```
rjbou, line 998: Is there a reason to remove memoisation for remaining uname cmd calls ?

src/core/opamStd.mli
```diff
@@ -510,11 +510,12 @@ module Sys : sig
   (** Queried lazily *)
   val os: unit -> os
 
-  (** The output of the command "uname", with the given argument. Memoised. *)
-  val uname_cmd: string -> string option
+  (** The output of the command "uname -U". FreeBSD only. *)
+  val uname_freebsd_version: unit -> string option
```
rjbou, line 514: It's not just uname, but uname -U to retrieve freebsd version. Maybe something like "freebsd_version"

src/core/opamStd.ml
```diff
@@ -903,19 +903,23 @@ module OpamSys = struct
...
-  let with_process_in cmd args f =
+  let process_in cmd args =
     if Sys.win32 then
       assert false;
-    let path = split_path_variable (Env.get "PATH") in
-    let cmd =
-      List.find Sys.file_exists (List.map (fun d -> Filename.concat d cmd) path)
-    in
-    let ic = Unix.open_process_in (cmd^" "^args) in
+    let args = Array.of_list (cmd :: args) in
```
rjbou, line 909:
```suggestion
```
cf below

src/core/opamStd.ml
```diff
@@ -903,19 +903,23 @@ module OpamSys = struct
...
-      let r = f ic in
-      ignore (Unix.close_process_in ic) ; r
-    with exn ->
-      ignore (Unix.close_process_in ic) ; raise exn
+      let path = split_path_variable (Env.get "PATH") in
+      let cmd =
+        List.find OpamStubs.is_executable
+          (List.map (fun d -> Filename.concat d cmd) path)
+      in
+      let (ic, _, _) as p = Unix.open_process_args_full cmd args env in
```
rjbou, line 917:
```suggestion
      let args = Array.of_list (cmd :: args) in
      let (ic, _, _) as p = Unix.open_process_args_full cmd args env in
```
`cmd` changed but not `args`.

src/core/opamStd.ml
```diff
@@ -903,19 +903,18 @@ module OpamSys = struct
       let split = if clean then OpamString.split else OpamString.split_delim in
       split path path_sep
 
-  let with_process_in cmd args f =
+  let process_in cmd args =
     if Sys.win32 then
       assert false;
```
rjbou, line 908: Should we keep that assert false or have a nicer error ?

src/core/opamStd.mli
```diff
@@ -510,11 +510,12 @@ module Sys : sig
   (** Queried lazily *)
   val os: unit -> os
 
-  (** The output of the command "uname", with the given argument. Memoised. *)
-  val uname_cmd: string -> string option
+  (** The output of the command "uname -U". FreeBSD only. *)
```
rjbou, line 513: Also, it's intended to be used only on freebsd, but no check is done for that. Proposal:
```
  (** FreeBSD version, probed via "uname -U".
      To use only on FreeBSD, otherwise returns [None]. *)

File: src/core/opamStubs.mli [hunk 8]
```diff
@@ -165,3 +165,12 @@ val getVersionInfo : string -> win32_version_info option
 
 val get_initial_environment : unit -> string list
 (** Windows only. Returns the environment which new processes would receive. *)
+
+type uname = {
+  sysname : string;
+  release : string;
+  machine : string;
+}
+
```
rjbou, line 174: Does these fields need to be documented ?

src/core/opamStd.mli
```diff
@@ -510,11 +510,14 @@ module Sys : sig
...
+  val uname_freebsd_version: unit -> string option
 
-  (** The output of the command "getconf", with the given argument. Memoised. *)
-  val getconf: string -> string option
+  (** The output of the command "getconf LONG_BIT". *)
+  val getconf_long_bit: unit -> string option
+
+  (** The memoized result of uname(2) *)
+  val uname : OpamStubs.uname Lazy.t
```
rjbou, line 520: uname should be before its specialised version for probing freebsd version

rjbou (commented):
* master changes needs to be updated with changes of the PR itself (freebsd uname, stubs, + API ones)
* `Fix opam on FreeBSD` commit
  * oneliner commit message is not enough explicit, maybe something like "Fix FreeBSD version probing"
  * commit message is explaining the process_in changes but not the uname call update
  * I don't know what is better, have everything on the same commit as freebsd fix, or two seperate commit each one having its purpose (process_in fix and uname call code simplification)
* I don't know if we should keep a backward compatiblity for the old `uname` and `getconf`, or just update the API and highlight the change in the release note.

src/core/opamStd.ml
kit-ty-kate: the assert false should stay since this function is not supposed to be called on windows and that it is not exported.

src/core/opamStd.ml
kit-ty-kate: the rest of the commands are called exactly at most once per opam calls so there is no reason to have a memorization function around

src/core/opamStd.mli
kit-ty-kate: `freebsd_version` doesn't give the impression that a command is going to be called so i'd rather keep the current name

src/core/opamStd.ml
kit-ty-kate: i don't think it really matters what value `argv[0]` has for those tools, but i can change that for the sake of having one less root value tracked by the GC for a short period of time

src/core/opamStd.mli
kit-ty-kate: i'm not sure about this suggestion. The fact that the function returns the freebsd version is already described in the function name. I prefer my version

kit-ty-kate:
I've split off the fix for OpenBSD to #6230 for simplification. Leaving this PR just for the `uname` improvement

src/core/opamStd.mli
rjbou: `uname` gives the impression that a command is called but it is not the case. This kind of information should be in the documentation.

src/core/opamStubs.mli
rjbou: The command helps, but it is better to have also some human understandable description :)

master_changes.md
```diff
@@ -131,7 +131,10 @@ users)
 ## opam-format
 
 ## opam-core
-  * `OpamStd.Sys.{get_terminal_columns,uname,getconf,guess_shell_compat}`: Harden the process calls to account for failures [#6230 @kit-ty-kate - fix #6215]
-  * `OpamStd.Sys.{uname,getconf}`: now accepts only one argument as parameter, as per their documentation [#6230 @kit-ty-kate]
+  * `OpamStd.Sys.{get_terminal_columns,guess_shell_compat}`: Harden the process calls to account for failures [#6230 @kit-ty-kate - fix #6215]
```
rjbou, line 134:
```suggestion
```

master_changes.md
```diff
@@ -131,7 +131,10 @@ users)
...
+  * `OpamStd.Sys.getconf`: was removed [#6217 @kit-ty-kate]
+  * `OpamStd.Sys.uname`: now returns the memoized result of `uname(2)` [#6217 @kit-ty-kate]
+  * `OpamStd.Sys.uname_freebsd_version`: was added, which returns the output of the `uname -U` command [#6217 @kit-ty-kate]
+  * `OpamStd.Sys.getconf_long_bit`: was added, which returns the output of the `getconf LONG_BIT` command [#6217 @kit-ty-kate]
```
rjbou, line 138:
```suggestion
  * `OpamStd.Sys.getconf`: was removed, replaced by `get_conf_long_bit` [#6217 @kit-ty-kate]
  * `OpamStd.Sys.getconf_long_bit`: was added, which returns the output of the `getconf LONG_BIT` command [#6217 @kit-ty-kate]
  * `OpamStd.Sys.uname`: now returns the memoized result of `uname(2)` [#6217 @kit-ty-kate]
  * `OpamStd.Sys.uname_freebsd_version`: was added, which returns the output of the `uname -U` command [#6217 @kit-ty-kate]
```

master_changes.md
```diff
@@ -131,7 +131,10 @@ users)
 ## opam-format
 
 ## opam-core
-  * `OpamStd.Sys.{get_terminal_columns,uname,getconf,guess_shell_compat}`: Harden the process calls to account for failures [#6230 @kit-ty-kate - fix #6215]
-  * `OpamStd.Sys.{uname,getconf}`: now accepts only one argument as parameter, as per their documentation [#6230 @kit-ty-kate]
+  * `OpamStd.Sys.{get_terminal_columns,guess_shell_compat}`: Harden the process calls to account for failures [#6230 @kit-ty-kate - fix #6215]
+  * `OpamStd.Sys.getconf`: was removed [#6217 @kit-ty-kate]
+  * `OpamStd.Sys.uname`: now returns the memoized result of `uname(2)` [#6217 @kit-ty-kate]
```
rjbou, line 136: In some distros, the `uname` C function is `uname(3)`. I think its better to refer it in plain text.
```suggestion
  * `OpamStd.Sys.uname`: now returns the memoized result of `uname` C binding [#6217 @kit-ty-kate]
```

master_changes.md
rjbou, line 134:
```suggestion
  * `OpamStd.Sys.{get_terminal_columns,uname,getconf,guess_shell_compat}`: Harden the process calls to account for failures [#6230 @kit-ty-kate - fix #6215]
```
No need to update it, uname & getconf are part of that PR

master_changes.md
```diff
@@ -131,7 +131,10 @@ users)
 ## opam-format
 
 ## opam-core
```
rjbou, line 133: Maybe it worth having an entry in `#Internal`, to highlight the better system polling change

kit-ty-kate: done

rjbou:
Except the naming, LGTM! We can discuss that monday.

src/core/opamStd.mli
rjbou:
```suggestion
  val get_freebsd_version: unit -> string option
```
As discussed on dev meeting

src/core/opamStd.ml
```diff
@@ -991,20 +991,13 @@ module OpamSys = struct
...
-        Hashtbl.add memo (cmd, arg) r;
-        r
+  let uname =
+    let uname = lazy (OpamStubs.uname ()) in
+    fun () ->
+      Lazy.force uname
 
-  let uname = memo_command "uname"
-  let getconf = memo_command "getconf"
+  let uname_freebsd_version () = process_in "uname" ["-U"]
```
rjbou, line 999:
```suggestion
  (* We need to call [uname] here as the only way to get it 
     in C without calling a process or lookup files is to 
     use [__Freebsd_version] which is a define, so it is set
     at compile time. *)
  let get_freebsd_version () = process_in "uname" ["-U"]
```
As discussed on dev meeting

rjbou:
As discussed on dev meeting, we should use `uname function` instead of `uname(2)` and `uname binary` or `tool` instead of `uname(1)`.

rjbou (approved):

PR #6219, opened 2024-10-01
Fix lint W59 with a local path that is not an archive

### LINKED ISSUE #6218: Lint W59 warn for missing checksum on local directory source url
When locally path pinning a package, the added url is local and the linting after edit warns that a checksum is missing:
```
[WARNING] The opam file didn't pass validation:
           warning 59: url doesn't contain a checksum
Proceed anyway ('no' will re-edit)? [y/n] y

### PR DESCRIPTION
**Pull request by rjbou**
fix #6218 

backported to 2.3 in https://github.com/ocaml/opam/pull/6220

### DIFF
--- a/src/core/opamSystem.ml
+++ b/src/core/opamSystem.ml
@@ -869,19 +869,17 @@ module Tar = struct
   let match_ext file ext =
     List.exists (Filename.check_suffix file) ext
 
+  let get_archive_extension file =
+    OpamStd.List.find_map_opt (fun (ext, t) ->
+        if match_ext file ext then Some t else None)
+      extensions
+
   let get_type file =
-    let ext =
-      List.fold_left
-        (fun acc (ext, t) -> match acc with
-           | Some t -> Some t
-           | None   ->
-             if match_ext file ext
-             then Some t
-             else None)
-        None
-        extensions in
     if Sys.file_exists file then guess_type file
-    else ext
+    else get_archive_extension file
+
+  let is_archive_from_string f =
+    get_archive_extension f <> None
 
   let is_archive file =
     get_type file <> None
@@ -930,6 +928,11 @@ end
 
 module Zip = struct
 
+  let extension = "zip"
+
+  let is_archive_from_string file =
+    Filename.check_suffix file extension
+
   let is_archive f =
     if Sys.file_exists f then
       try
@@ -944,11 +947,15 @@ module Zip = struct
         | _ -> false
       with Sys_error _ | End_of_file -> false
     else
-      Filename.check_suffix f "zip"
+      is_archive_from_string f
 
   let extract_command file =
     Some (fun dir -> make_command "unzip" [ file; "-d"; dir ])
 end
+
+let is_archive_from_string file =
+  Tar.is_archive_from_string file
+  || Zip.is_archive_from_string file
 
 let is_archive file =
   Tar.is_archive file || Zip.is_archive file
--- a/src/core/opamSystem.mli
+++ b/src/core/opamSystem.mli
@@ -233,8 +233,13 @@ val read_command_output: ?verbose:bool -> ?env:string array ->
 
 (** END *)
 
-(** Test whether the file is an archive, by looking as its extension *)
+(** Test whether the file is an archive, by looking at its magic number
+    if the file exists, otherwise by looking as its extension *)
 val is_archive: string -> bool
+
+(** Test whether the given path is an archive, only by looking at its
+    extension *)
+val is_archive_from_string: string -> bool
 
 (** [extract ~dir:dirname filename] extracts the archive [filename] into
     [dirname]. [dirname] should not exists and [filename] should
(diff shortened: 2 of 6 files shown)

### CODE REVIEW

File: src/state/opamFileTools.ml [hunk 1]
```diff
@@ -396,9 +396,15 @@ let t_lint ?check_extra_files ?(check_upstream=false) ?(all=false) t =
         | #OpamUrl.version_control -> true
         | _ -> false)
   in
+  let url_archive =
+    let open OpamStd.Option.Op in
+    t.url >>| OpamFile.URL.url >>| (fun u ->
+        OpamSystem.is_archive u.OpamUrl.path)
```
kit-ty-kate, line 402: i don't think that's right. `is_archive` will look up the file locally. If you have an url like `https:///etc/fstab`, the path will be looked up.

kit-ty-kate (changes requested):

src/state/opamFileTools.ml
rjbou: Good point, we should then add a function that checks if a path looks like an archive, like `looks_like_ssh_path`.

kit-ty-kate: i'm not sure we need a new function. Something like that would work just fine i think:
```suggestion
        OpamSystem.local_file u <> None)
```

rjbou: We don't want to check file existence on disk (which does `OpamUrl.local_file`). We want to know if given string seems like an archive path.

kit-ty-kate: What about using https://github.com/ocaml/opam/pull/5979 instead?

tests/reftests/lint.test
```diff
@@ -884,7 +885,53 @@ license: "ISC"
...
+authors: "the testing team"
+homepage: "egapemoh"
+maintainer: "maint@tain.er"
+license: "ISC"
+dev-repo: "hg+https://to@li.nt"
+bug-reports: "https://nobug"
+### <add-url.sh>
+basedir=$(echo $BASEDIR | sed "s/\\\\\\\\/\\\\\\\\\\\\\\\\/g")
+cat << EOF >> lint.opam
+url { src:"file://$basedir/an-archive" }
```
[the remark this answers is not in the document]
dra27: Slightly for readability, but I think that you can halve the number of backslashes by using a single-quoted string?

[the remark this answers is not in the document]
rjbou: Indeed

src/core/opamSystem.ml
```diff
@@ -869,19 +869,23 @@ module Tar = struct
...
+  let get_archive_extension file =
+    List.fold_left
+      (fun acc (ext, t) -> match acc with
+         | Some t -> Some t
+         | None   ->
+           if match_ext file ext
+           then Some t
+           else None)
+      None
+      extensions
```
kit-ty-kate, line 881: We don’t need to go through the whole list:
```suggestion
    OpamStd.List.find_map_opt (fun (ext, t) ->
        if match_ext file ext then Some t else None)
      extensions
```

src/core/opamSystem.mli
```diff
@@ -233,9 +233,14 @@ val read_command_output: ?verbose:bool -> ?env:string array ->
 
 (** END *)
 
-(** Test whether the file is an archive, by looking as its extension *)
+(** Test whether the file is an archive, by looking at its magic magic number
```
kit-ty-kate, line 236:
```suggestion
(** Test whether the file is an archive, by looking at its magic number
```

tests/reftests/lint.test
```diff
@@ -884,7 +885,53 @@ license: "ISC"
...
+synopsis: "A word"
+description: "Two words."
+authors: "the testing team"
+homepage: "egapemoh"
+maintainer: "maint@tain.er"
+license: "ISC"
+dev-repo: "hg+https://to@li.nt"
+bug-reports: "https://nobug"
+### <add-url.sh>
+basedir=`echo "$BASEDIR" | sed 's/\\\\/\\\\\\\\\\\\\\\\/g'`
```
kit-ty-kate, line 927: you used single quotes here instead of double-quotes and only halved the left part of the sed, not the right part.
```suggestion
basedir=`echo "$BASEDIR" | sed 's/\\\\/\\\\\\\\/g'`
```

tests/reftests/lint.test
```diff
@@ -884,7 +885,53 @@ license: "ISC"
...
```
kit-ty-kate, line 906: same here
```suggestion
basedir=`echo "$BASEDIR" | sed 's/\\\\/\\\\\\\\/g'`
```

tests/reftests/lint.test
kit-ty-kate: actually what is the goal of this sed? Is it just to escape every backslashes? If so, then the solution might be instead:
```suggestion
basedir=`echo "$BASEDIR" | sed 's/\\/\\\\/g'`
```
I know is everywhere in the reftests in its double-quoted form but i’ve never really understood it ^^"

rjbou: It's the different levels of escape that are needed to end with `s/\\/\\\\/g` launched.

kit-ty-kate (approved):

kit-ty-kate:
Thanks!

PR #6221, opened 2024-10-01
Pin: fix first local vcs fetch rsyncing all source

### LINKED ISSUE #5809: Initial pin of VCS directory includes untracked and ignored files [2.1 regression]
```Dockerfile
FROM ocaml/opam:debian-12-opam
RUN git clone https://github.com/ocaml/stdlib-shims.git
RUN sudo ln -f /usr/bin/opam-2.2 /usr/bin/opam
RUN rm -rf ~/.opam && opam init --bare -a ~/opam-repository
RUN touch stdlib-shims/stdlib-shims.install stdlib-shims/untracked
RUN opam switch create --empty stdlib-shims
RUN opam pin add stdlib-shims git+file:///home/opam/stdlib-shims -yn
RUN ls -a /home/opam/.opam/stdlib-shims/.opam-switch/sources/stdlib-shims > before
RUN opam update stdlib-shims
RUN ls -a /home/opam/.opam/stdlib-shims/.opam-switch/sources/stdlib-shims > after
RUN grep '\.git\|untracked\|install' before
RUN diff -u before after
```

Since opam 2.1, the initial copy of the pin directory incorrectly:
- Has _no_ `.git` folder
- Has copies of `stdlib.shims.install` (which is in `.gitignore`) and `untracked` (which should be an untracked file)

Replacing the `opam-2.2` with `opam-2.1` tests 2.1 and and removing that and the following line completely tests 2.0.

The first `opam update` of the VCS sources causes the `.git` directory to be added and the untracked/ignored files to be removed.

rjbou:
**Backported in 2.3 with  #6222**

The origin of the issue is from the magic to retrieve the opam file even if it is not committed added in #4300. To do that, the local vcs url was changed to local rsync, in order to get all source not only committed files. In the next synchronisation, as opam file contains the vcs url, the synchronisation fixed that anomaly.

Fix #5809
~~Queued on #6135~~

### DIFF
--- a/tests/reftests/pin.test
+++ b/tests/reftests/pin.test
@@ -90,7 +90,7 @@ Now run 'opam upgrade' to apply any package updates.
 ### : pin with url
 ### opam pin add nip-git git+file://$BASEDIR/nip-git
 Package nip-git does not exist, create as a NEW package? [y/n] y
-[nip-git.dev] synchronised (file://${BASEDIR}/nip-git)
+[nip-git.dev] synchronised (git+file://${BASEDIR}/nip-git#master)
 nip-git is now pinned to git+file://${BASEDIR}/nip-git#master (version ved)
 
 The following actions will be performed:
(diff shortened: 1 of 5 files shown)


tests/reftests/pin.test
```diff
@@ -1364,3 +1364,20 @@ echo GARBAGE>>"$1"
...
+### git -C vcs-local commit -qm 'init'
+### git -C vcs-local ls-files
+vcs-local.opam
+### <vcs-local/untracked>
+i shouldn't be retrieved
+### opam pin vcs-local ./vcs-local --no-action
+Package vcs-local does not exist, create as a NEW package? [y/n] y
+[vcs-local.dev] synchronised (file://${BASEDIR}/vcs-local)
+vcs-local is now pinned to git+file://${BASEDIR}/vcs-local#master (version dev)
+### test -f OPAM/vcs-local/.opam-switch/sources/vcs-local/untracked
```
kit-ty-kate, line 1383: I would also modify `vcs-local/vcs-local.opam` after the git add so it has untracked changes and add a cat after the `test -f` to make sure the file is correctly copied

kit-ty-kate (commented):
lgtm otherwise

kit-ty-kate (approved):

rjbou:
Updated

kit-ty-kate (approved):

kit-ty-kate:
Thanks a lot!

PR #6223, opened 2024-10-02
Update the vendored dependencies and unlock OCaml 5.3 compatibility

kit-ty-kate:
All the vendored dependencies were updated except re while https://github.com/ocaml/ocaml-re/issues/411 is being delt with.

Backported to 2.3 in https://github.com/ocaml/opam/pull/6225

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -71,6 +71,8 @@ users)
 
 ## Build
   * Bump the requirement for dune to 2.8 [#6204 @kit-ty-kate]
+  * Bump the vendored version of dune to 3.16.0, cppo to 1.7.0 and extlib to 1.8.0 [#6223 @kit-ty-kate]
+  * Fix compilation with OCaml 5.3 when using the vendored extlib by updating to the 5.3 compatible version (e.g. `make cold` or `./configure --with-vendored-deps`) [#6223 @kit-ty-kate]
 
 ## Infrastructure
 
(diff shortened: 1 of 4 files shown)


master_changes.md
```diff
@@ -71,6 +71,8 @@ users)
 
 ## Build
   * Bump the requirement for dune to 2.8 [#6204 @kit-ty-kate]
+  * Bump the vendored version of dune to 3.16.0, cppo to 1.7.0 and extlib to 1.8.0 [#6223 @kit-ty-kate]
+  * Fix compilation with OCaml 5.3 when using the vendored extlib (e.g. `make cold` or `./configure --with-vendored-deps`) [#6223 @kit-ty-kate]
```
rjbou, line 75: Is it the update of the vendored lib that is the fix ? If yes, I propose something like that.
```suggestion
  * Fix compilation with OCaml 5.3 when using the vendored extlib by depending on 5.3 compatible version (e.g. `make cold` or `./configure --with-vendored-deps`) [#6223 @kit-ty-kate]
```

kit-ty-kate: I don't think "depending on ..." makes sense in the context of vendored packages but something like that works for me:
```suggestion
  * Fix compilation with OCaml 5.3 when using the vendored extlib by updating to the 5.3 compatible version (e.g. `make cold` or `./configure --with-vendored-deps`) [#6223 @kit-ty-kate]
```

rjbou: lgtm!

rjbou (approved):
Thanks!

rjbou (approved):

PR #6226, opened 2024-10-03
Documentation: Update the command to install opam to point to the new simplified url on opam.ocaml.org

kit-ty-kate:
At present this is a redirection to the same script on github

See https://github.com/ocaml-opam/opam2web/pull/241


rjbou (approved):
Thanks!

PR #6230, opened 2024-10-08
System polling: Harden the process calls to account for failure

kit-ty-kate:
Split off from https://github.com/ocaml/opam/pull/6217
Fixes https://github.com/ocaml/opam/issues/6215

with_process_in was fairly brittle and would display stderr by default and ignore errors. The new version of that function (renamed process_in) now uses Unix.open_process_full and returns None if anything went wrong, fixing the issue on OpenBSD where `getconf LONG_BIT` would fail.

Backported to 2.3 in https://github.com/ocaml/opam/pull/6228

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -139,3 +139,5 @@ users)
 ## opam-format
 
 ## opam-core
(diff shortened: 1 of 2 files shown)


rjbou (approved):
Thanks!

Issue #5553, opened 2023-05-19
`opam update` that are noop should be instantaneous


Khady:
When an `opam update` sees that nothing changed in a repository the operation should be instantaneous instead of taking 20s.
Also opam should check headers like `last-modified` so that it doesn't actually download the index file if it wasn't changed since the last download.

```
louis@ahrefs-laptop1-popos1:~$ time opam update

<><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
[default] synchronised from https://opam.ocaml.org
Now run 'opam upgrade' to apply any package updates.

real	0m29.922s
user	0m11.714s
sys	0m9.216s
louis@ahrefs-laptop1-popos1:~$ time opam update --debug
00:00.034  CLI                    Parsing CLI version 2.1
00:00.035  GSTATE                 LOAD-GLOBAL-STATE @ /home/louis/.opam
00:00.035  CLIENT                 UPDATE 
00:00.035  RSTATE                 LOAD-REPOSITORY-STATE @ /home/louis/.opam
00:00.300  CACHE(repository)      Loaded /home/louis/.opam/repo/state-3AF31D16.cache in 0.264s
00:00.300  RSTATE                 Cache found
00:00.300  STATE                  LOAD-SWITCH-STATE @ default
00:00.302  CACHE(installed)       Loaded /home/louis/.opam/default/.opam-switch/packages/cache in 0.001s
00:00.803  STATE                  Switch state loaded in 0.503s

<><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
00:00.806  PARALLEL               Iterate over 1 task(s) with 3 process(es)
00:00.806  PARALLEL               Starting job 0 (worker 1/3): 0
00:00.806  SYSTEM                 mkdir /tmp/opam-2016727-26b9ac
00:02.395  REPOSITORY             update default from https://opam.ocaml.org
00:02.395  CURL                   pull-repo-update
00:02.395  SYSTEM                 mkdir /tmp/opam-2016727-26b9ac/default.new
00:02.395  SYSTEM                 mkdir /tmp/opam-2016727-f25dcf
00:02.395  PARALLEL               Next task in job 0: /usr/bin/curl --write-out %{http_code}\n --retry 3 --retry-delay 2 --user-agent opam/2.1.4 -L -o /tmp/opam-2016727-f25dcf/index.tar.gz.part -- https://opam.ocaml.org/index.tar.gz
Processing  1/1: [default: http]
[...]

PR #6249, opened 2024-10-17
Ensure each repositories stored in repos-config is associated with an URL

kit-ty-kate:
This is mostly cleanup work noticed as i was looking at implementing https://github.com/ocaml/opam/issues/5553.

Looking at the code of `OpamFile.Repos_config`, i got surprised having repositories without any url was allowed (`'a option`) so i digged a little and it seems that the `None` state is not used anywhere:
* the InitConfig parser would refuse it with `Missing repository URL`
* manually entering it in `repos-config` would be accepted but ignored internally and i have no idea why it would ever be used

In any case, this special case is not tested anywhere in the test-suite either so it looks reasonable to assume it was never meant to be handled, and reading the commit that added that possibility https://github.com/ocaml/opam/commit/667eacd9e75940e8eb5247d726054383ec68bf77, it seems to have been added with the idea that "it might be useful in the future", but 8 years later i don't think it turned out true at all.

https://github.com/ocaml-opam/opam-rt/pull/79 should be merged first right before merging this PR

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -119,6 +119,7 @@ users)
   * download tool: Use fetch on DragonFlyBSD and ftp on NetBSD [#6305 @kit-ty-kate]
   * Prefer curl over any other download tools on every systems, if available [#6305 @kit-ty-kate]
   * Avoid issues when using wget2 where the requested url might return an html page instead of the expected content [#6303 @kit-ty-kate]
+  * Ensure each repositories stored in repos-config is associated with an URL [#6249 @kit-ty-kate]
 
 ## Internal: Windows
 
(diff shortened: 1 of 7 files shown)


rjbou (commented):
Agree on the idea!
Even if it is not used, it needs a test to see the change and no change that are written in the main comment.

kit-ty-kate:
> Even if it is not used, it needs a test to see the change and no change that are written in the main comment.

I'm not sure to understand. Aside from crafting a custom repos-config file which can't happen in the wild i don't think there is a way to test this.

rjbou (approved):
I've added a test, it's mainly in the idea to thee the diff, even if it can't happen with the binary

Issue #6250, opened 2024-10-17
how to tell opam at init to use a different ~/.opam directory [question]


UnixJunkie:
Hello,

Is it possible to say something like:
```bash
opam init --dot-opam-dir=/tmp/this_directory
```
i.e. force opam to use an explicitly given directory to
store the dot opam dir?

Thanks,
F.

kit-ty-kate:
yes, using `--root` or the `OPAMROOT` environment variable

UnixJunkie:
thanks a lot Kate!

PR #6251, opened 2024-10-17
Doc: point users to ways to override the default opam root location in the opam init manpage

kit-ty-kate:
Might have helped https://github.com/ocaml/opam/issues/6250.

`--root` and `$OPAMROOT` aren't very well documented in `opam init`. The concept of an "opam root" is mentioned but the option is very far down the man page and the environment variable even more.
So i feel like, for such a useful command, it should be mentioned in the top summary.


rjbou (approved):
Thanks!

PR #5349, opened 2022-11-11
GHA: add doc job

rjbou:
Generation of documentation with `make doc`
* [x] add a check on generated doc ?
* [x] add a check on index.html update when new modules is added

PR #6252, opened 2024-10-18
Change example of non-letter in version ordering

gridbugs:
The "Package Formulas" grammar in the manual says that valid characters for version numbers are `<identchar> | "+" | "." | "~"`. Notably this does not include the "#" character. This change updates the example of a comparing a letter to a non-letter character to use a non-letter character that may appear in a version number.


rjbou (approved):
Thanks!
I've added a commit to remove a redundancy that I spotted while reviewing (`+` defined twice for `version`).
~~I propose to not merge it now, to queue it on #5349.~~ rebased

kit-ty-kate:
Thanks!

Issue #6270, opened 2024-10-28
opam upgrade pkg refuses to upgrade


kit-ty-kate:
```
$ opam upgrade cohttp
Everything as up-to-date as possible (run with --verbose to show unavailable upgrades).

The following packages are not being upgraded because the new versions conflict with other installed
packages:
  - cohttp.6.0.0~beta2
However, you may "opam upgrade" these packages explicitly, which will ask permission to downgrade or
uninstall the conflicting packages.
Nothing to do.
```
The message says that `opam upgrade <pkg>` should be able to upgrade, however what it meant is `opam upgrade <pkgname>.<version>`. This should be reflected in the message.

Example cohttp version: 5.3.1

PR #6272, opened 2024-10-28
Improve the messages when a package is not up-to-date on opam upgrade

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/6270 and also the confusing message about `--verbose` when it is unnecessary.

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -279,26 +279,29 @@ let upgrade_t
              ) latest OpamPackage.Map.empty
          in
 
-         if (OpamConsole.verbose ()) && not (OpamPackage.Set.is_empty unav) then
-           (OpamConsole.formatted_msg
-              "%s.\n\
-               The following newer versions couldn't be installed:\n"
-              hdmsg;
-            OpamConsole.msg "%s"
-              (OpamStd.Format.itemize (fun p ->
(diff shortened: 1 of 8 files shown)


src/client/opamClient.ml
```diff
@@ -330,9 +333,11 @@ let upgrade_t
                  ) (OpamPackage.Set.elements unopt))
            );
          OpamConsole.formatted_msg
-           "However, you may \"opam upgrade\" these packages explicitly, \
+           "However, you may \"opam upgrade\" these packages explicitly \
+            at these versions (e.g. \"opam upgrade %s\"), \
```
rjbou, line 337:
```suggestion
            at these versions (e.g. 'opam upgrade %s'), \
```
\+ some underline ?

rjbou (commented):
It's missing test update, otherwise, lgtm!

src/client/opamClient.ml
kit-ty-kate: for underline, i'm not sure if it's necessary. The rest of the opam example commands in our codebase uses either double-quotes like here or `bold`, the former being used more often.

rjbou: They are both used in fact, `"` and `'`.

rjbou (approved):
Thanks!

Issue #6271, opened 2024-10-28
avoid-version packages should be ignored when opam upgrade refuses to upgrade


kit-ty-kate:
```
$ opam upgrade cohttp
Everything as up-to-date as possible (run with --verbose to show unavailable upgrades).

The following packages are not being upgraded because the new versions conflict with other installed packages:
  - cohttp.6.0.0~beta2
However, you may "opam upgrade" these packages explicitly, which will ask permission to downgrade or uninstall the conflicting packages.
Nothing to do.
```
```
$ opam install cohttp.6.0.0~beta2
The following actions will be performed:
=== downgrade 3 packages
  ↘ conduit          7.1.0 to 6.2.3       [required by conduit-lwt]
  ↘ conduit-lwt      7.1.0 to 6.2.3       [required by cohttp-lwt-unix]
  ↘ conduit-lwt-unix 7.1.0 to 6.2.3       [required by cohttp-lwt-unix]
=== recompile 3 packages
  ↻ github           4.4.1                [uses cohttp]
  ↻ github-unix      4.4.1                [uses cohttp]
  ↻ opam-publish     2.4.0 (pinned)       [uses github]
=== upgrade 3 packages
  ↗ cohttp           5.3.1 to 6.0.0~beta2
  ↗ cohttp-lwt       5.3.0 to 6.0.0~beta2 [uses cohttp]
  ↗ cohttp-lwt-unix  5.3.0 to 6.0.0~beta2 [uses cohttp]
=== install 1 package
  ∗ http             6.0.0~beta2          [required by cohttp]

Proceed with ↘ 3 downgrades, ↻ 3 recompilations, ↗ 3 upgrades and ∗ 1 installation? [y/n] n
```
The message shown on `opam upgrade` should not be there as `cohttp.6.0.0~beta2` is flagged with `avoid-version`. The command should simply say that everything is up-to-date and should not mention the pre-release.

PR #6273, opened 2024-10-28
opam upgrade: Do not show the not-up-to-date message with packages tagged with avoid-version

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/6271

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -202,9 +202,27 @@ let upgrade_t
         else OpamPackage.packages_of_names t.installed requested
       in
       let latest =
-        OpamPackage.Name.Set.fold (fun name acc ->
-            OpamPackage.Set.add (OpamPackage.max_version t.packages name) acc)
-          (OpamPackage.names_of_packages to_check)
+        OpamPackage.Set.fold (fun pkg acc ->
+            let name = OpamPackage.name pkg in
+            let pkgs = OpamPackage.packages_of_name t.packages name in
+            let latest =
+              OpamPackage.Set.fold (fun pkg latest ->
+                  if OpamPackage.compare latest pkg < 0 then
+                    let opam = OpamPackage.Map.find pkg t.opams in
+                    let avoid_version =
+                      List.exists (function
+                          | Pkgflag_AvoidVersion | Pkgflag_Deprecated -> true
+                          | Pkgflag_LightUninstall | Pkgflag_Verbose
+                          | Pkgflag_Plugin | Pkgflag_Compiler
+                          | Pkgflag_Conf | Pkgflag_Unknown _ -> false)
+                        (OpamFile.OPAM.flags opam)
+                    in
+                    if avoid_version then latest else pkg
+                  else latest)
+                pkgs pkg
+            in
+            OpamPackage.Set.add latest acc)
+          to_check
           OpamPackage.Set.empty in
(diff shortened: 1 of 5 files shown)


src/client/opamClient.ml
```diff
@@ -202,9 +202,25 @@ let upgrade_t
...
+            let name = OpamPackage.name pkg in
+            let pkgs = OpamPackage.packages_of_name t.packages name in
+            let latest =
+              OpamPackage.Set.fold (fun pkg latest ->
+                  if OpamPackage.compare latest pkg < 0 then
+                    let opam = OpamPackage.Map.find pkg t.opams in
+                    let avoid_version =
+                      List.exists (function
+                          | Pkgflag_AvoidVersion | Pkgflag_Deprecated -> true
+                          | _ -> false)
```
rjbou, line 215: We should expand the default case, like that if we add another flag and this part is concerned there is an highlight.

src/client/opamClient.ml
```diff
@@ -202,9 +202,25 @@ let upgrade_t
...
       in
       let latest =
-        OpamPackage.Name.Set.fold (fun name acc ->
-            OpamPackage.Set.add (OpamPackage.max_version t.packages name) acc)
-          (OpamPackage.names_of_packages to_check)
+        OpamPackage.Set.fold (fun pkg acc ->
+            let name = OpamPackage.name pkg in
+            let pkgs = OpamPackage.packages_of_name t.packages name in
+            let latest =
+              OpamPackage.Set.fold (fun pkg latest ->
```
rjbou, line 209: this variable named `pkg` and the one defined 4 lines above are not the "same package" but have the same name, this can be misleading when reading the code.

src/client/opamClient.ml
```diff
@@ -202,9 +202,25 @@ let upgrade_t
         else OpamPackage.packages_of_names t.installed requested
       in
       let latest =
-        OpamPackage.Name.Set.fold (fun name acc ->
-            OpamPackage.Set.add (OpamPackage.max_version t.packages name) acc)
-          (OpamPackage.names_of_packages to_check)
+        OpamPackage.Set.fold (fun pkg acc ->
+            let name = OpamPackage.name pkg in
+            let pkgs = OpamPackage.packages_of_name t.packages name in
```
rjbou, line 207: I don't know if it worth working on the set of version instead of package one.

rjbou (commented):
I needs a test too.

src/client/opamClient.ml
kit-ty-kate: it simplifies the code greatly. Otherwise we would need to have an intermediate map of set and update it

src/client/opamClient.ml
kit-ty-kate: done

src/client/opamClient.ml
kit-ty-kate: I disagree. Since the name above doesn't have any use below i think it's better to shadow it. The same pattern is used in different places in our code already and i think it make sense to use it here

rjbou (approved):

src/client/opamClient.ml
rjbou: The same pattern is used true, usually it is a map that is populated through intertwined fold, and in that case it is more clear. But i'm sure that we can find this pattern for other case :)

PR #6278, opened 2024-10-29
Delay the creation of the context used when a package failed to build up until the package failed

kit-ty-kate:
Noticed while reviewing https://github.com/ocaml/opam/pull/6274.
It turns out that we try to fetch the revision of each packages being built even before executing the command, but that information is only used in rare cases so if many git packages are used this can cause some wasted time (not too much, up to a couple seconds on slow hardware with large installs, but not nothing)

### DIFF
--- a/src/core/opamProcess.ml
+++ b/src/core/opamProcess.ml
@@ -254,6 +254,12 @@ type t = {
   p_tmp_files: string list;
 }
 
+let equal p1 p2 =
+  (p1.p_pid : int) = (p2.p_pid : int)
+
+let compare p1 p2 =
+  Int.compare p1.p_pid p2.p_pid
+
 let output_lines oc lines =
   List.iter (fun line ->
     output_string oc line;
--- a/src/core/opamProcess.ml
+++ b/src/core/opamProcess.ml
@@ -208,7 +208,7 @@ type command = {
   cmd_stdout: string option;
   cmd_verbose: bool option;
   cmd_name: string option;
-  cmd_metadata: (string * string) list option;
+  cmd_metadata: (string * string) list Lazy.t option;
 }
 
 let string_of_command c = String.concat " " (c.cmd::c.args)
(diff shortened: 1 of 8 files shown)


src/core/opamProcess.ml
```diff
@@ -606,7 +612,7 @@ let set_verbose_f, print_verbose_f, isset_verbose_f, stop_verbose_f =
     (* implem relies on sigalrm, not implemented on win32.
        This will fall back to buffered output. *)
     if Sys.win32 then () else
-    let files = OpamStd.List.sort_nodup compare files in
+    let files = OpamStd.List.sort_nodup OpamStd.Compare.compare files in
```
rjbou, line 615:
```suggestion
    let files = OpamStd.List.sort_nodup String.compare files in
```
?

rjbou (commented):
The PR contains 2 things: the delay of context creation for package build, and the delay of metadata processes as the sumary is not always printed. PR title is good, but main comment and master changes should contains both. Ideally, 2 commits if it can be easily split.
Otherwise, lgtm!

src/core/opamProcess.ml
```diff
@@ -388,7 +395,9 @@ let create ?info_file ?env_file ?(allow_stdin=not Sys.win32) ?stdout_file ?stder
     | Some f ->
       let chan = open_out f in
       let info =
-        make_info ~cmd ~args ~cwd ~env_file ~stdout_file ~stderr_file ~metadata () in
+        make_info ~cmd ~args ~cwd ~env_file ~stdout_file ~stderr_file
+          ~metadata:(Lazy.force metadata) ()
```
kit-ty-kate, line 399: this is forced too early. I think there is a way to run it only when the process finished

kit-ty-kate (commented):
> The PR contains 2 things: the delay of context creation for package build, and the delay of metadata processes as the sumary is not always printed. PR title is good, but main comment and master changes should contains both. Ideally, 2 commits if it can be easily split.

I don't think those are two separate things. `cmd_metadata`, `p_metadata` and `r_info` are all storing the same things only in three different types. If anything i should be talking about metadata instead of context

PR #6289, opened 2024-11-07
Use an uppercase character again for the default in interactive questions

### LINKED ISSUE #6288: UI: since using 2.2.1 (likely 2.2), the default answer to interactive questions is barely visible
With opam 2.1, for any question that opam asked, there was a default answer indicated by an uppercase character `Do you want to continue with installation [Y/n]`.

Now, using opam 2.2, I can barely see what the default is --- yes, it is underlined and bold... but that's at least in my terminal barely visible:

opam-2 2-default-option

hannesm:
Fix #6288


kit-ty-kate (approved):
I'm all for this change. It fits with my ignored opinion in https://github.com/ocaml/opam/pull/5053#issuecomment-1068183200 except for the `<no character>` part which requires some further changes which can be done separately if we want (i'll open a PR anyway, it's linked to a small bug i noticed), but i don't care as much about that as the standard unix `[Y/n] ` question style

hannesm:
Thanks for the reftest fixes! Happy that there's an approval.

rjbou (approved):

PR #6296, opened 2024-11-14
CI: fiw macos jobs regarding gpatch

rjbou:
Update the github action scripts now that homebrew renamed the GNU patch binary to gpatch (see Homebrew/homebrew-core#174687)

Ported from #6292


kit-ty-kate:
i was hoping to have https://github.com/ocaml/opam/pull/6255 be the fix on master instead

kit-ty-kate (approved):
Actually if https://github.com/ocaml/opam/pull/5892 passes in 2.4.0 #6255 would be redundant so let's not stall this fix. Thanks!

PR #6304, opened 2024-11-18
Make curl the default download tool instead of wget

kit-ty-kate:
wget was the default despite not being in macOS because of #2006. However this was fixed and is not needed anymore


rjbou (approved):
I've added a commit to add some doc. Otherwise, lgtm!

PR #6307, opened 2024-11-20
Remove ocaml-system from the list of default compilers used during opam init

### LINKED ISSUE #3509: do not use `system` switch by default
Sorry to start a discussion about that when y'all are busy with opam2. I think the "system" switch is the number 1 source of issues for beginners that try to use opam and come ask for help on #ocaml @ freenode; my advice is almost always to use a non-system switch and it fixes their issue. Therefore I think opam shouldn't use the system switch at all unless explicitly asked so.

kit-ty-kate:
Fixes #3509

I don't think having `ocaml-system` in the list of default compiler is a good choice.
Most platforms actually won't work if you installed the `ocaml` package alone. They usually all require some sort of extra `ocaml-compiler-libs` to be complete and not end up in very confusing error messages for new users.

System compilers are also not tested in the CI for opam-repository as of today (see https://github.com/ocurrent/opam-repo-ci/issues/327) and end up breaking packages from time to time.

Experienced developers can always use `opam init -c ocaml-system` if they really want a system compiler and `ocaml-system` can always be installed explicitly later if needed, but as a default for everyone it seems to brittle to me.

### DIFF
--- a/tests/reftests/init.test
+++ b/tests/reftests/init.test
@@ -67,6 +67,8 @@ depends: [ "ocaml" {= "4.10.0" & post} ]
 flags: compiler
 conflict-class: "ocaml-core-compiler"
 ### :---:
+### :I: Default compiler selection
+### :I:1: No system compiler
 ### rm -rf ${OPAMROOT}
 ### <opamrc>
 eval-variables: [ sys-ocaml-version ["false"] "no system compiler" ]
@@ -87,6 +89,7 @@ Switch invariant: ["ocaml" {>= "4.05.0"}]
 Done.
 ### opam switch invariant
 ["ocaml" {>= "4.05.0"}]
+### :I:2: system compiler not compliant with default invariant
 ### rm -rf ${OPAMROOT}
 ### <opamrc>
 eval-variables: [ sys-ocaml-version ["echo" "4.02.3"] "old system compiler" ]
@@ -107,6 +110,7 @@ Switch invariant: ["ocaml" {>= "4.05.0"}]
 Done.
 ### opam switch invariant
 ["ocaml" {>= "4.05.0"}]
+### :I:3: System compiler = 4.07.0, should be selected default
 ### rm -rf ${OPAMROOT}
 ### <opamrc>
 eval-variables: [ sys-ocaml-version ["echo" "4.07.0"] "new system compiler" ]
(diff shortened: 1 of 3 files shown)


rjbou:
Discussion on dev meeting: For a transition phase, we should detect presence of an system ocaml and display a hint for the command to run to have by default ocaml-system

tests/reftests/init.test
```diff
@@ -116,36 +116,23 @@ Configuring from ${BASEDIR}/opamrc and then from built-in defaults.
 <><> Fetching repository information ><><><><><><><><><><><><><><><><><><><><><>
 [default] Initialised
 
-<><> Creating initial switch 'default' (invariant ["ocaml" {>= "4.05.0"}] - initially with ocaml-system)
```
rjbou, line 119: This test needs to be fixed with the new behaviour (add -c ocaml-system), and another one added for the change.

kit-ty-kate: done

kit-ty-kate:
> For a transition phase, we should detect presence of a system ocaml and display a hint for the command to run to have by default ocaml-system.

I'm not sure if we want that, reading such a message would be confusing to users. I would rather make it clear on the release note instead.

>  This change also affects default switch creation, not only init.

I'm not sure to understand what you mean. As far as i know `opam init` is the only subcommand affected

rjbou (commented):
I've added a commit to label the test to show their purpose, and a commit that contains a proposal to modify the tests : 
* I:3: show the defaut, so it is updated to show base compiler selection
* I:4: added to show the behaviour of system compiler explicit selection

tests/reftests/init.test
```diff
@@ -110,21 +110,23 @@ Done.
 ### rm -rf ${OPAMROOT}
 ### <opamrc>
 eval-variables: [ sys-ocaml-version ["echo" "4.07.0"] "new system compiler" ]
-### opam init --no-setup --bypass-checks default REPO/ --config opamrc | grep -v Cygwin
+### opam init --no-setup --bypass-checks default REPO/ --config opamrc -c ocaml-system | grep -v Cygwin
```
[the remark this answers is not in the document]
kit-ty-kate: I'm not sure to understand. This goes against what you asked for on your last review: https://github.com/ocaml/opam/pull/6307#discussion_r1856624911

rjbou (approved):

Issue #3199, opened 2018-01-27
opam pin fails adding new or multiple packages when --no-action is present


kit-ty-kate:
For now the opam pin command used with the --no-action/-n argument fails if the package pinned is new or if we try to add multiple packages from a single directory.
This behaviour does not really matches the intended purpose of --no-action, namely:
```
Just record the new pinning status, and don't prompt for
(re)installation or removal of affected packages.
```
Moreover the default behaviour in opam 1.2.2 in regards to adding new package was to add them anyway. --no-action didn't really play any role there but imho for scripts it is important to keep this behaviour and to be able to pin packages without installing them.

AltGr:
Could you give details on the interaction you are expecting ? By 'new', do you mean a package without metadata ?

This works:
```
% cat >opam
opam-version: "2.0"  
version: "test"
build: "false"
% opam pin -n .
Package foo does not exist, create as a NEW package ? [Y/n] y
[foo.~dev] synchronised from file:///tmp/foo
[WARNING] Failed checks on foo package definition from source at file:///tmp/foo:
    error 23: Missing field 'maintainer'
    error 25: Missing field 'authors'
    error 35: Missing field 'homepage'
  warning 36: Missing field 'bug-reports'
foo is now pinned to file:///tmp/foo (version test)
```

AltGr:
@kit-ty-kate Do you have any details on this ? I confirm that you are expecting the right thing from `-n`

kit-ty-kate:
Oh sorry I missed your first ping. Hum, well your example above doesn't work if you are using docker for instance. Of course if we add ```-y``` to the pin command it works but my point being that the behaviour is different between pining an existing package and pinning a new (that doesn't exist yet in the list of packages) or multiple packages because opam first asks questions like "do you want to continue" here.

AltGr:
Hmm, maybe it's the wording that is wrong then. The message should be understood as "don't prompt for **(re)installation or removal of affected packages**", not "**don't prompt** for (re)installation or removal of affected packages".

The important part is that it won't run the installation actions, not the fact that it will prompt (for other reasons). Was that the unclear part, and should we just clarify the wording, or do you think more is needed ?

rjbou:
ping @kit-ty-kate

dra27:
Would this be better, getting rid of the word "prompt" completely?

```
Only record the new pinning status; affected packages will
not be (re)installed or removed.
```

PR #6309, opened 2024-11-21
Don't ask confirmation when pinning a new package

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/3199

While reading https://github.com/ocaml/opam/issues/3199 i asked myself why this question was asked and i couldn't find any reason. It was added in https://github.com/ocaml/opam/pull/1335 with no reasoning behind it either.

In my opinion the question is redundant as the user is already asking to pin the package(s).

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -47,6 +47,7 @@ users)
 ## Pin
   * [NEW] Make it so pin list display the current revision of a pinned repository in a new column [#6274 @desumn - fix #5533]
   * [BUG] Stop double pin of packages located in ./opam/opam [#6343 @kit-ty-kate - fix #6342]
+  * Don't ask confirmation when pinning an unknown package (absent from repositories) [#6309 @kit-ty-kate @rjbou - fix #3199]
 
 ## List
 
--- a/src/client/opamPinCommand.ml
+++ b/src/client/opamPinCommand.ml
@@ -491,12 +491,10 @@ and source_pin
       version, None
   in
 
-  if not (OpamPackage.has_name st.packages name) &&
-     not (OpamConsole.confirm
-            "Package %s does not exist, create as a %s package?"
-            (OpamPackage.Name.to_string name)
-            (OpamConsole.colorise `bold "NEW"))
-  then raise Aborted;
+  if not (OpamPackage.has_name st.packages name) then
+    OpamConsole.note "Package %s does not exist in opam repositories \
+                      registered in the current switch."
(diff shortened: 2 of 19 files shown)


master_changes.md
```diff
@@ -43,6 +43,7 @@ users)
 ## Config
 
 ## Pin
+  * Don't ask confirmation when pinning a new package [#6309 @kit-ty-kate]
```
rjbou, line 46:
```suggestion
  * Don't ask confirmation when pinning a unknown package (absent from repositories) [#6309 @kit-ty-kate]
```

rjbou:
Discussion on dev meeting: add a note instead of a question

master_changes.md
kit-ty-kate: done

src/client/opamPinCommand.ml
```diff
@@ -491,12 +491,9 @@ and source_pin
...
   in
 
-  if not (OpamPackage.has_name st.packages name) &&
-     not (OpamConsole.confirm
-            "Package %s does not exist, create as a %s package?"
-            (OpamPackage.Name.to_string name)
-            (OpamConsole.colorise `bold "NEW"))
-  then raise Aborted;
+  if not (OpamPackage.has_name st.packages name) then
+    OpamConsole.note "Package %s was unknown until now"
```
kit-ty-kate, line 495: If anyone has a better idea feel free to change it

rjbou: Suggestion
```suggestion
    OpamConsole.note "Package %s does not exist in registered repositories."
```

kit-ty-kate: as a user i would find this message confusing. The repositories we refer to here are opam repositories but they might understand source repositories instead since the command does something with a source repository usually.

rjbou:
```suggestion
    OpamConsole.note "Package %s does not exist in opam repositories registered in the current switch."
```
?

kit-ty-kate: actually i'm not sure. What do we want to say? Maybe something like:
```suggestion
    OpamConsole.note "%s is a new package."
```

rjbou: The general idea is that we want to give the information that the package is not know (from loaded repositories), like that the user can check if it is not the intended behaviour: a typo in cli, in opam file, etc.

rjbou (approved):

PR #6313, opened 2024-11-26
Run shellcheck on shell/install.sh and make it pass, add a CI job to check next changes

kit-ty-kate:
This is a partial subset of https://github.com/ocaml/opam/pull/5778.
This PR is meant as a way to more easily kickstart the use of shellcheck to lint checkout shell scripts.

Details of reason behind the shell changes:
* [SC2006](https://www.shellcheck.net/wiki/SC2006): `$()` vs backticks
* [SC2268](https://www.shellcheck.net/wiki/SC2268): The x-prefix is actually unnecessary
* POSIX shell doesn't support `local`
* [SC2145](https://www.shellcheck.net/wiki/SC2145): `$@` in strings
* missing quotes
* [SC2162](https://www.shellcheck.net/wiki/SC2162): `read` without `-r` will interpret backslashes
* variables inside the format string for `printf` could contain `%` which could make the command fail

This PR also make sure the script keeps passing `shellcheck` by adding it to our CI.

### DIFF
--- a/.github/scripts/main/hygiene.sh
+++ b/.github/scripts/main/hygiene.sh
@@ -124,4 +124,17 @@ please run dune exec --root=. -- ./ci.exe from .github/workflows and fixup the c
 fi
 (set +x ; echo -en "::endgroup::check workflow generation\r") 2>/dev/null
 
+###
+# Shellcheck
+###
+
+(set +x ; echo -en "::group::check shell scripts using shellcheck\r") 2>/dev/null
+if shellcheck shell/install.sh ; then
+  (set +x; echo "shell/install.sh: OK") 2>/dev/null
+else
+  (set +x; echo -e "shell/install.sh: \e[31mERROR\e[0m") 2>/dev/null
+  ERROR=1
+fi
+(set +x ; echo -en "::endgroup::check shell scripts using shellcheck\r") 2>/dev/null
+
 exit $ERROR
--- a/.github/workflows/ci.ml
+++ b/.github/workflows/ci.ml
@@ -492,6 +492,7 @@ let hygiene_job (type a) ~analyse_job (platform : a platform) ~oc ~workflow f =
          "done";
        ]
     ++ run "Hygiene" ~cond:(Or[Predicate(true, Contains("steps.files.outputs.modified", "configure.ac"));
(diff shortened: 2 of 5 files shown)


ElectreAAS:
Nice to see this going forward!

.github/workflows/ci.ml
```diff
@@ -490,8 +490,9 @@ let hygiene_job (type a) ~analyse_job (platform : a platform) ~oc ~workflow f =
          "done";
        ]
     ++ run "Hygiene" ~cond:(Or[Predicate(true, Contains("steps.files.outputs.modified", "configure.ac"));
+                               Predicate(true, Contains("steps.files.outputs.modified", "shell/install.sh"));
                                Predicate(true, Contains("steps.files.outputs.all", "src_ext"));
-                               Predicate(true, Contains("steps.files.outputs.all", ".github/workflows"))])
+                               Predicate(true, Contains("steps.files.outputs.all", ".github"))])
```
rjbou, line 495: What is the reason of this change ?

.github/scripts/main/hygiene.sh
```diff
@@ -124,4 +124,17 @@ please run dune exec --root=. -- ./ci.exe from .github/workflows and fixup the c
...
 
+###
+# Shellcheck
+###
+
+(set +x ; echo -en "::group::check shell scripts using shellcheck\r") 2>/dev/null
+if shellcheck shell/install.sh ; then
+  (set +x; echo "shell/install.sh: OK") 2>/dev/null
+else
+  (set +x; echo "shell/install.sh: ERROR") 2>/dev/null
```
rjbou, line 135:
```suggestion
  (set +x; echo -e "shell/install.sh: [\e[31mERROR\e[0m]") 2>/dev/null
```

rjbou (commented):
On the content, lgtm! It's better to split on 2 commits: the first that updates shell/install.sh and the other that adds the CI check and confirms it passes.
Plus the PR title/comments should highlight the new shellcheck CI check.

.github/scripts/main/hygiene.sh
kit-ty-kate: done (although i didn't keep the brackets as they don't really make sense in this position)

rjbou (approved):
I've added an entry for the first commit in changelog. Otherwise, LGTM!!

PR #6315, opened 2024-11-28
Fix a typo in a documentation string

shym:
A PR so small the title says it all…


rjbou (approved):
Thanks! The PR is small, but helpful!

PR #6316, opened 2024-11-28
env: add a flag to output as environment variable bindings

### LINKED ISSUE #5791: Support new flag: `opam env --bare` for use with GitHub actions
Environment variables can be made available to steps in a GitHub workflow job by [writing the variable definitions to the `$GITHUB_ENV` file](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files).

- `opam env` prints an environment in terms of bash commands
- `opam env --sexp` prints an environment as an s-expression

If there existed a flag `--bare` (say) that printed an environment as exactly`VAR=VALUE` lines e.g.
```bash
  OPAM_SWITCH_PREFIX=/home/foo-user/.opam/default
  CAML_LD_LIBRARY_PATH=/home/foo-user/.opam/default/lib/ocaml/stublibs:/home/foo-user/.opam/default/lib/ocaml
  ...
```
that is, equivalent to
```bash
  opam env | sed -e "s/ export .*//g" -e "s/'//g" -e "s/\;//g"
```
then, a github action to configure an OPAM environment could be very simply written as just
```
  - run: |
    opam env --bare >> $GITHUB_ENV
```

rjbou:
fix #5791

The name is still to be determined (it's a sed-able name ftm)

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -71,6 +71,7 @@ users)
 ## Clean
 
 ## Env
+  * [NEW] Add `--raw` option to output env as environment variable binding, useful for CI environment populating [#6316 @rjbou - fix #5791]
 
 ## Opamfile
 
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -1302,9 +1302,15 @@ let option cli =
           $global_options cli $fieldvalue $global cli)
 
 module Common_config_flags = struct
-  let sexp cli =
-    mk_flag ~cli cli_original ["sexp"]
-      "Print environment as an s-expression rather than in shell format"
+  let env_format cli =
+    mk_vflag ~cli None [
+      cli_original, Some `sexp, ["sexp"],
+      "Print environment as an s-expression rather than in shell format";
+      cli_from cli2_4, Some `raw, ["raw"],
+      "Print environment as variable bindings rather than in shell format.\
+       Useful to populate CI environment."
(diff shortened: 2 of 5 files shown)


kit-ty-kate (commented):
I'm ok with the idea but i'm really not sure about the name of the new option.

`--bindings(s)` doesn't really describe anything to me personally. I haven't thought about it much but so far i prefer the original suggestion: `--bare`.

One other issue is that the expected usage is to add to the `$GITHUB_ENV` file in Github Actions. However the syntax for it is slightly more complex than we can first assume. In the rare case where a newline character is part of the value of an environment variable, the "binding" syntax is not correct anymore.
I think we should check that no newline character exists before printing and fail otherwise.

rjbou:
From dev meeting: Consensus over the name `--raw`. We won't handle newline in this PR, but prefer output a constructed json for CI handling.

rjbou:
Update with `--raw`

rjbou:
From (last week) dev meeting: we are going to a full json output for CI, containing the information of the full computed path, and also what to append and/or prepend. Like that CI scripts can use that information to change  their environment variables as needed.
For ex:
```
{ 
  "var" : "PATH"
  "full": "foo:bar:/usr/bin:/etc:baz";
  "prepend": ["foo"; "bar"];
  "append": ["baz"]
}

PR #6317, opened 2024-11-28
lint: add W74 to detect 'pin-depends' packages that are not present in 'depends' nor in 'depopts'

### LINKED ISSUE #5795: should `pin-depends` imply `depends` ?
Hi,

I realized that if I have the following `owi.opam` file:

```yml
depends: []
pin-depends: [
  [ "encoding.dev" "git+https://github.com/formalsec/encoding"]
]
```

Then, running `opam install -y ./*.opam --deps-only --with-test` would not install `encoding`.

But if I add it to the `depends` fields:

```yml
depends: [ "encoding" ]
pin-depends: [
  [ "encoding.dev" "git+https://github.com/formalsec/encoding"]
]
```

Then it would be properly installed with the same command.

I'm wondering if there is a reason why a package being present in `pin-depends` does not imply it is part of `depends` ?

Thanks!

rjbou:
fix #5795
This lint doesn't detect when the versions mismatch.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -71,6 +71,7 @@ users)
   * [NEW] Add options `--require-checksums` and `--no-checksums` to harmonise with `opam install` [#5563 @rjbou]
 
 ## Lint
+  * [NEW] Add W74 to detect `pin-depends` packages that are neither present in the `depends` nor `depopts` field [#6317 @rjbou - fix #5795]
 
 ## Repository
   * Accurately tag `curl` download command when loaded from global config file [#6270 @rjbou]
--- a/src/state/opamFileTools.ml
+++ b/src/state/opamFileTools.ml
@@ -1098,6 +1098,22 @@ let t_lint ?check_extra_files ?(check_upstream=false) ?(all=false) t =
        "Field 'extra-files' contains path with '..'"
        ~detail:relative
        (relative <> []));
+    (let missing =
+       let pkgs =
+         List.fold_left
+           (fun set (pkg, _) ->
+              OpamPackage.Name.Set.add (OpamPackage.name pkg) set)
+           OpamPackage.Name.Set.empty t.pin_depends
+       in
+       OpamPackage.Name.Set.fold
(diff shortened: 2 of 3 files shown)


kit-ty-kate (approved):

src/state/opamFileTools.ml
```diff
@@ -1098,6 +1098,21 @@ let t_lint ?check_extra_files ?(check_upstream=false) ?(all=false) t =
...
+         List.map fst t.pin_depends
+         |> OpamPackage.Set.of_list
+         |> OpamPackage.names_of_packages
+       in
+       OpamPackage.Name.Set.diff pkgs all_depends
+       |> OpamPackage.Name.Set.elements
+       |> List.map OpamPackage.Name.to_string
+     in
+     cond 74 `Warning
+       "Field 'pin-depends' contains packages that are not in 'depends' nor in \
```
kit-ty-kate, line 1112:
```suggestion
       "Field 'pin-depends' contains packages that are neither in 'depends' nor in \
```

master_changes.md
```diff
@@ -60,6 +60,7 @@ users)
 ## Source
 
 ## Lint
+  * [NEW] Add W74 to detect `pin-depends` packages that are not present in `depends` nor `depopts` field [#6317 @rjbou - fix #5795]
```
kit-ty-kate, line 63:
```suggestion
  * [NEW] Add W74 to detect `pin-depends` packages that are neither present in `depends` nor `depopts` field [#6317 @rjbou - fix #5795]
```

kit-ty-kate (commented):
The code can be improved by eliminating 3 out of the 6 datatypes traverses.
The gain is small in this particular case but out of principle it feels weird to not do better if the code is as short and maybe even more understandable

rjbou:
I find the set transformation by pipe more human understandable (you just follow the function name, not read the code of the function given to fold). But agree on the optim part.

kit-ty-kate (approved):

kit-ty-kate:
Thanks!

PR #6318, opened 2024-11-30
Fix opam switch list-available when given several arguments

kit-ty-kate:
Required for #6186 

Giving several arguments would never have given any results before. Furthermore it now matches the behaviour of `opam list`

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -39,7 +39,7 @@ users)
 
 ## Switch
   * [BUG] Fix `opam switch remove <dir>` failure when it is a linked switch [#6276 @btjorge - fix #6275]
-  * Fix `opam switch list-available` when given several arguments [#6318 @kit-ty-kate]
+  * [BUG] Fix `opam switch list-available` when given several arguments [#6318 @kit-ty-kate]
 
 ## Config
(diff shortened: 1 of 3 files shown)


rjbou:
Agree on the idea. Some history, it was introduced from the beginning as an `ands` (see #2650).
We need to advertise it as it is a change of behaviour. Maybe have it behind a CLI version guard ?

master_changes.md
```diff
@@ -39,6 +39,7 @@ users)
 
 ## Switch
   * [BUG] Fix `opam switch remove <dir>` failure when it is a linked switch [#6276 @btjorge - fix #6275]
+  * Fix `opam switch list-available` when given several arguments [#6318 @kit-ty-kate]
```
kit-ty-kate, line 42:
```suggestion
  * [BUG] Fix `opam switch list-available` when given several arguments [#6318 @kit-ty-kate]
```

kit-ty-kate:
> Maybe have it behind a CLI version guard ?

i don't think we need to do that for something that had no useful outcome

rjbou (approved):

PR #6319, opened 2024-11-30
Fail when trying to pin a package whose definition could not be found instead of forcing interactive edition

kit-ty-kate:
Closes https://github.com/ocaml/opam/issues/6322
This could happen with things like `opam pin <typo> <url>` or a typo in the `pin-depends` definition.

The interactive edition based on the dummy opam file template is now only allowed if the user is purposefully requesting it using `--edit` or `opam pin edit`.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -208,6 +208,7 @@ users)
   * Add pin test to show stored overlay opam files [#6209 @rjbou]
   * Add show test to highlight precedence of opam file selection and check that if an opam file is given it is always this one that is taken [#6209 @rjbou]
   * Add a reftest showing the effect of env updates containing empty strings on `variables.sh` [#6198 @kit-ty-kate]
+  * Add tests showing behaviour of `opam pin` when confronted with a missing opam description [#6319 @kit-ty-kate]
 
 ### Engine
 
--- a/master_changes.md
+++ b/master_changes.md
@@ -59,6 +59,7 @@ users)
   * Don't ask confirmation when pinning an unknown package (absent from repositories) [#6309 @kit-ty-kate @rjbou - fix #3199]
   * [BUG] Do not ask to install pin-depends twice [#6375 @kit-ty-kate - fix #6374]
   * [BUG] Ensure the right versions (the pinned one) of packages are used when simulating pinning [#6256 @rjbou @kit-ty-kate - fix #6248 #6379]
+  * ✘ Fail when trying to pin a package whose definition could not be found instead of forcing interactive edition (e.g. this could happen when making a typo in the package name of a pin-depends) [#6319 @kit-ty-kate - fix #6322]
 
 ## List
 
--- a/src/client/opamPinCommand.ml
+++ b/src/client/opamPinCommand.ml
@@ -569,37 +569,34 @@ and source_pin
     | _ -> opam_opt
   in
 
-  if not need_edit && opam_opt = None then
-    OpamConsole.note
-      "No package definition found for %s: please complete the template"
-      (OpamConsole.colorise `bold (OpamPackage.to_string nv));
-
-  let need_edit = need_edit || opam_opt = None in
-
   let opam_opt =
-    let opam_base = match opam_opt with
-      | None -> OpamFileTools.template nv
-      | Some opam -> opam
-    in
-    let opam_base =
-      OpamFile.OPAM.with_url_opt urlf opam_base
-    in
     if need_edit then
-      (if not (OpamFile.exists temp_file) then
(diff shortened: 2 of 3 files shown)


src/client/opamPinCommand.ml
```diff
@@ -571,32 +571,28 @@ and source_pin
...
-       edit_raw name temp_file >>|
-       (* Preserve metadata_dir so that copy_files below works *)
-       OpamFile.OPAM.(with_metadata_dir (metadata_dir opam_base))
-      )
+      let opam_base = match opam_opt with
+        | None ->
+          OpamConsole.note
+            "No package definition found for %s: please complete the template"
+            (OpamConsole.colorise `bold (OpamPackage.to_string nv));
+          OpamFileTools.template nv
```
kit-ty-kate, line 581: do we really want to keep this template around? In which situation is it useful? I feel like it doesn't really has any purpose and we might want to remove it entirely (probably in a separate PR)

kit-ty-kate:
@rjbou I'm not sure how to fix your test from https://github.com/ocaml/opam/pull/5471. Any ideas?

src/client/opamPinCommand.ml
```diff
@@ -569,32 +569,28 @@ and source_pin
...
+      if not (OpamFile.exists temp_file) then
+        OpamFile.OPAM.write_with_preserved_format
+          ?format_from:(OpamPinned.orig_opam_file st name opam_base)
+          temp_file opam_base;
+      edit_raw name temp_file >>|
+      (* Preserve metadata_dir so that copy_files below works *)
+      OpamFile.OPAM.(with_metadata_dir (metadata_dir opam_base))
     else
-      Some opam_base
+      Option.map (OpamFile.OPAM.with_url_opt urlf) opam_opt
```
rjbou, line 593: As we update url here, do we still need lines 600-604 ?
```ocaml
   let opam =
      match OpamFile.OPAM.get_url opam with
      | Some _ -> opam
      | None -> OpamFile.OPAM.with_url_opt urlf opam
    in
```

tests/reftests/pin.test
```diff
@@ -1068,3 +1068,68 @@ opam-version: "2.0"
...
+opam-version: "2.0"
+### sh ./pin-depends.sh missing-opam/pkg.opam pin-pkg.dev missing-opam
+### opam pin ./missing-opam
+[NOTE] Package pkg does not exist in opam repositories registered in the current switch.
+The following additional pinnings are required by pkg.dev:
+  - pin-pkg.dev at file://${BASEDIR}/missing-opam
+Pin and install them? [Y/n] y
+[NOTE] Package pin-pkg does not exist in opam repositories registered in the current switch.
+[pin-pkg.dev] synchronised (no changes)
+[ERROR] No valid package definition found
```
rjbou, line 1107: With the new behaviour, it may worth adding the package name here?

rjbou (commented):
On the general implem, LGTM! some small remarks

src/client/opamPinCommand.ml
kit-ty-kate: good catch. That was also the case in the code before this PR but it's always good to clean redundant code ^^

kit-ty-kate: done in follow up fixup commit

tests/reftests/pin.test
kit-ty-kate: done in follow up fixup commit

src/client/opamPinCommand.ml
rjbou: fair enough

master_changes.md
```diff
@@ -53,6 +53,7 @@ users)
   * Don't ask confirmation when pinning an unknown package (absent from repositories) [#6309 @kit-ty-kate @rjbou - fix #3199]
   * [BUG] Do not ask to install pin-depends twice [#6375 @kit-ty-kate - fix #6374]
   * [BUG] Ensure the right versions (the pinned one) of packages are used when simulating pinning [#6256 @rjbou @kit-ty-kate - fix #6248 #6379]
+  * [BUG] Fail when trying to pin a package whose definition could not be found instead of forcing interactive edition (e.g. this could happen when making a typo in the package name of a pin-depends) [#6319 @kit-ty-kate - fix #6322]
```
rjbou, line 56:
```suggestion
  * ✘ Fail when trying to pin a package whose definition could not be found instead of forcing interactive edition (e.g. this could happen when making a typo in the package name of a pin-depends) [#6319 @kit-ty-kate - fix #6322]
```
It was not a bug, it was a feature. Here we change the behaviour

PR #6320, opened 2024-12-01
Manual: add information when flags (avoid-version, deprecated) were introduced

hannesm:
As discussed with @kit-ty-kate. It would for sure make sense to add version information on more fields in the documentation. But for me that is pretty tedious since I'd have to manually look every field up. I'll do when I dig around more fields. But maybe someone else has the stamina to go through the entire manual.


kit-ty-kate (approved):

rjbou (approved):

rjbou:
Thanks!

PR #6321, opened 2024-12-01
Upgrade to opam-file-format 2.2.0~alpha1

kit-ty-kate:
now depends on menhir instead of ocamlyacc.

Requires https://github.com/ocaml/opam-file-format/pull/60

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -82,6 +82,8 @@ users)
 ## VCS
 
 ## Build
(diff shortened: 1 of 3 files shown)


kit-ty-kate:
opam commands can be affected negatively (average of 15% performace loss for commands that require to parse opam files). Are there some known menhir options to improve performance?

rjbou (approved):

PR #6326, opened 2024-12-05
sandbox: respect the `DUNE_CACHE_ROOT` environment variable if it exists

smorimoto:
Users can set `DUNE_CACHE_ROOT`, which has a higher priority than other environment variables.
Currently, there are some cases where the cache functionality does not work well when `DUNE_CACHE_ROOT` is set.

Ref: https://dune.readthedocs.io/en/stable/caching.html

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -85,6 +85,7 @@ users)
 ## Format upgrade
 
 ## Sandbox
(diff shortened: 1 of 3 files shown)


kit-ty-kate (approved):
the bubblewrap (linux) sandbox script also needed to be changed but i've done it. lgtm

smorimoto:
Thanks a lot!

kit-ty-kate:
Thanks!

PR #6330, opened 2024-12-10
Nushell support

benjamin-voisin:
This change make the command `opam env` work within [nushell](https://github.com/nushell/nushell) by typing
```
load-env (opam env | from json)
```

It does not add full nushell support, only for the `env` command, but it’s the first step. I don’t know that much about the code of opam and could not figure out where to make the other changes neccessary / when they are called, so this is open to improvments.

**NOTE**: nushell still has some breaking changes sometimes, so I don’t know how this will hold up to the future, but I am certain that it will be easier to change an existing implementation.

### DIFF
--- a/src/state/opamEnv.ml
+++ b/src/state/opamEnv.ml
@@ -868,6 +868,8 @@ let opam_env_invocation ?root ?switch ?(set_opamswitch=false) shell =
       Printf.sprintf " \"--%s=%s\"" argname
     | SH_sh | SH_bash | SH_zsh | SH_csh | SH_fish ->
       Printf.sprintf " '--%s=%s'" argname
+    | SH_nu ->
+      Printf.sprintf " '--%s=%s'" argname
     in
     if filepath_needs_quote pathval then
       quoted pathval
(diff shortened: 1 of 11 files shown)


rjbou:
Thanks for the PR!

> This change make the command `opam env` work within [nushell](https://github.com/nushell/nushell) by typing
> 
> ```
> load-env (opam env | from json)
> ```

If you don't want to add full support, just need a json output to be able to populate nushell environment, you can take a look to [this comment](https://github.com/ocaml/opam/pull/6316#issuecomment-2546070239), we have in mind to have something similar for CIs.

> It does not add full nushell support, only for the `env` command, but it’s the first step. I don’t know that much about the code of opam and could not figure out where to make the other changes neccessary / when they are called, so this is open to improvments.

The other places where a change is needed are for generating the init scripts. they are generated at `opam init` stage, in order to have some automatic loading in *rc file (completion, environment handling, automatic environment update, etc.).
You can find examples [here](https://github.com/ocaml/opam/blob/8f4837261a519aee6368d0066ab609e2f4e527cf/tests/reftests/init-scripts.unix.test#L29) in the init scripts test.

> **NOTE**: nushell still has some breaking changes sometimes, so I don’t know how this will hold up to the future, but I am certain that it will be easier to change an existing implementation.

If nushell users are willing to tell us when there is some update to do, and what are they, this is ok. It is hard for us to maintain a shell we don't use :).
Note that, ftm, opam has a release cycle of ~6 months. So breakages fix won't be immediately available.

src/state/opamEnv.ml
```diff
@@ -1065,6 +1075,8 @@ let if_interactive_script shell t e =
     Printf.sprintf "if ( $?prompt ) then\n  %s%sendif\n" t @@ ielse e
   | SH_fish ->
     Printf.sprintf "if status is-interactive\n  %s%send\n" t @@ ielse e
+  | SH_nu ->
+    Printf.sprintf "%sTODO%s" t @@ ielse e
```
kit-ty-kate, line 1079: @ferrisml do you have any idea how do those two TODOs with Nu?

PR #6333, opened 2024-12-10
Fix sandboxing support in NixOS

kit-ty-kate:
Upstreams changes from https://github.com/NixOS/nixpkgs/pull/363770

The issue is that in the bubblewrap sandbox `/run` is mounted as tmpfs so its content is emptied inside the sandbox and Nix makes use of `/run/current-system/sw` to store the whole system (everything symlinks to it if i understand correctly).

cc @RyanGibb just in case you can tell me if i've done/said something awful

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -86,6 +86,7 @@ users)
 
 ## Sandbox
   * Respect the `DUNE_CACHE_ROOT` environment variable if it exists [#6326 @smorimoto]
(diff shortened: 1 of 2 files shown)


eilvelia:
I think that `add_sys_mounts` depends on `dir` being set (that's why my nixpkgs pr changes `for dir in /*; do` instead of adding a separate line)

rjbou (approved):

Issue #5805, opened 2024-01-24
Fix opam admin check --obsolete


kit-ty-kate:
As noted in https://github.com/ocaml/opam/issues/5748, `opam admin check --obsolete` currently doesn’t work for its intended purpose.

Instead of ignoring `available` and filters (I haven’t looked at the code, I’m just guessing based on behaviour) it should probably:
* for each packages
  * list all variables they’re using
    * for each uses call the solver with the variable set to the possible values (the tricky part)
    * add all those possibilities to the result

This should also probably get rid of the need to have a `--ignore-test-doc` argument.
This could also fix https://github.com/ocaml/opam/issues/4541 since it’s sharing some code

kit-ty-kate:
cc @Armael @gasche since you both worked on something similar in the past to try and fix https://github.com/ocaml/opam/issues/4541.
Sorry the pseudo algorithm is very handwavy, this is very much a back of a napkin type thing and I’m not sure if the third point is even possible in particular, i think it’s probably too tricky to do precisely but an heuristic when you just evaluate all expression to `true` and `false` then solve N-time seems fine. Do you have any insights by any chance?

PR #6335, opened 2024-12-10
opam admin check: Set with-test and with-doc to false by default

kit-ty-kate:
Queued on top of #6331 

Currently `opam admin check` sets `with-test` and `with-doc` to true by default, which in my opinion makes little sense, especially in light of https://github.com/ocaml/opam/issues/4541 as it does not take into account user behaviour at all but some hypothetical "what if someone had OPAMWITHTEST=1 and OPAMWITHDOC=1 always set in their environment".

This draft PR flips the default for those two variables to use what the actual default everywhere else is (`with-test=false` and `with-doc=false`). It also gets rid of `-i` which has no purpose anymore.

Instead this PR introduces three more arguments: `--with-test`, `--with-doc` and `--with-dev-setup`. However in light of https://github.com/ocaml/opam/issues/4541 i'm not sure this is a good idea and i don't this anyone uses `opam admin check` without `-i` currently, which makes the introduction of those replacements seem useless. What do you think?

Related to https://github.com/ocaml/opam/issues/5805

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -118,6 +118,8 @@ users)
 ## Admin
   * ◈ Add `opam admin compare-versions` to compare package versions for sanity checks [#6197 @mbarbin]
   * [BUG] Fix `opam admin check` in the presence of the `with-dev-setup` variable [#6331 @kit-ty-kate - fix #6329]
+  * ✘ The `-i`/`--ignore-test-doc` argument has been removed from `opam admin check` [#6335 @kit-ty-kate]
+  * ✘ `opam admin check` now sets `with-test` and `with-doc` to `false` instead of `true` [#6335 @kit-ty-kate]
 
 ## Opam installer
 
--- a/src/client/opamAdminCommand.ml
+++ b/src/client/opamAdminCommand.ml
@@ -761,11 +761,6 @@ let check_command cli =
         are supposed to be available. By default, all checks are run."
   ]
   in
-  let ignore_test_arg =
-    OpamArg.mk_flag ~cli OpamArg.cli_original ["ignore-test-doc";"i"]
-      "By default, $(b,{with-test}) and $(b,{with-doc}) dependencies are \
-       included. This ignores them, and makes the test more tolerant."
-  in
   let print_short_arg =
     OpamArg.mk_flag ~cli OpamArg.cli_original ["s";"short"]
       "Only output a list of uninstallable packages"
@@ -782,7 +777,7 @@ let check_command cli =
     OpamArg.mk_flag ~cli OpamArg.cli_original ["obsolete"]
       "Analyse for obsolete packages"
   in
-  let cmd global_options ignore_test print_short
+  let cmd global_options print_short
       installability cycles obsolete () =
     OpamArg.apply_global_options cli global_options;
     let repo_root = checked_repo_root () in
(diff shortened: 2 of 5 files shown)


src/client/opamAdminCommand.ml
```diff
@@ -761,10 +761,17 @@ let check_command cli =
         are supposed to be available. By default, all checks are run."
   ]
   in
-  let ignore_test_arg =
-    OpamArg.mk_flag ~cli OpamArg.cli_original ["ignore-test-doc";"i"]
-      "By default, $(b,{with-test}) and $(b,{with-doc}) dependencies are \
-       included. This ignores them, and makes the test more tolerant."
+  let with_test_arg =
+    OpamArg.mk_flag ~cli OpamArg.(cli_from cli2_4) ["with-test"]
```
rjbou, line 765:
```suggestion
    OpamArg.mk_flag ~cli OpamArg.(cli_from cli2_4) ["t";"with-test"]

File: src/client/opamAdminCommand.ml [hunk 2]
```diff
@@ -761,10 +761,17 @@ let check_command cli =
...
+      "Set the $(b,with-test) variable to true for all packages."
+  in
+  let with_doc_arg =
+    OpamArg.mk_flag ~cli OpamArg.(cli_from cli2_4) ["with-doc"]
+      "Set the $(b,with-doc) variable to true for all packages."
+  in
+  let with_dev_setup_arg =
+    OpamArg.mk_flag ~cli OpamArg.(cli_from cli2_4) ["with-dev-setup"]
```
rjbou, line 773:
```suggestion
    OpamArg.mk_flag ~cli OpamArg.(cli_from cli2_4) ["s"; "with-dev-setup"]
```

i'm not sur for this one

src/client/opamAdminCommand.ml
rjbou: Adding short forms to be select to have them all more easily `-tds` instead of `--with-test --with-doc --with-dev-setup`

kit-ty-kate:
After reflection i chose to not add new arguments `--with-test`, `--with-doc` and `--with-dev-setup` as they would be highly misleading as they would set the variable globally instead of the intended per package, the former does not make sense and is not possible to have elsewhere in opam anyway.

The PR is now ready to review

master_changes.md
```diff
@@ -114,6 +114,8 @@ users)
 ## Admin
   * ◈ Add `opam admin compare-versions` to compare package versions for sanity checks [#6197 @mbarbin]
   * [BUG] Fix `opam admin check` in the presence of the `with-dev-setup` variable [#6331 @kit-ty-kate - fix #6329]
+  * ✘ The `-i`/`--ignore-test-doc` argument has been removed from `opam admin check` [#6335 @kit-ty-kate]
+  * ✘ `opam admin check` now sets `with-test` and `with-doc` to `false` instead of `true` [#6335 @kit-ty-kate]
```
rjbou, line 118:
```suggestion
  * ✘ `opam admin check` now sets `with-test`, `with-doc`, and `with-dev-setup` to `false` instead of `true` [#6335 @kit-ty-kate]
```

rjbou:
> After reflection i chose to not add new arguments `--with-test`, `--with-doc` and `--with-dev-setup` as they would be highly misleading as they would set the variable globally instead of the intended per package, the former does not make sense and is not possible to have elsewhere in opam anyway.

Agree. Plus, as there is no way to request check for a single package there is nothing else to add.

rjbou (approved):

kit-ty-kate:
rebased above master

Issue #5341, opened 2022-11-06
Add xz to the list of recommended tool checked during init


kit-ty-kate:

kit-ty-kate:
(oops keyboard typed enter too soon ^^")

progressive-galib:
done

progressive-galib:
@kit-ty-kate so are you gonna  close it or merge it ?

kit-ty-kate:
Since nobody had any issues in 2 years and i'm unable to remember or reproduce what was the problem with opam, lets close it. We can always reopen it later if needed.

PR #6340, opened 2024-12-18
Added xz to the list of recommended tool checked during init (As per #5341) 

progressive-galib:
Added xz to the list of recommended tool checked during opam init (As per #5341).


kit-ty-kate:
I'm not sure this is necessary. I can't remember what type of failure i was seeing at the time but looking around it seems that, of the system officially supported by opam, only OpenBSD tar does not support xz archives by default. And for this system we already use and require `gtar` instead.

kit-ty-kate:
Which platform(s) are you seeing this issue on?

kit-ty-kate:
Closing per https://github.com/ocaml/opam/issues/5341#issuecomment-2555929340

Issue #5520, opened 2023-04-22
`opam lock` does not exit with an error code after displaying an error


lefessan:
When some dependencies are requested in the `.opam` file, but are not yet installed in the current switch, `opam lock` will display an error about missing dependencies, but will still exit normally at the end.

Because of this behavior, scripts are not able to correctly detect the error when calling `opam lock`.

progressive-galib:
https://github.com/ocaml/opam/blob/master/src/client/opamLockCommand.ml#L31-L32

the problem seems to be here, i fixed it but gotta write tests for them now.

progressive-galib:
done

PR #6341, opened 2024-12-19
the opam lock command exits with a non-zero exit code when  dependencies  are not yet installed in the current switch

progressive-galib:
fixed https://github.com/ocaml/opam/issues/5520

### DIFF
--- a/src/client/opamLockCommand.ml
+++ b/src/client/opamLockCommand.ml
@@ -96,7 +96,7 @@ let select_packages atom_locs st =
                atoms
            in
            if missing <> [] then
-             (OpamConsole.error
+             (OpamConsole.error_and_exit
                 "Skipping %s, dependencies are not satisfied in this switch, \
                  not installed packages are:\n%s"
                 (OpamPackage.to_string nv)
--- a/src/client/opamLockCommand.ml
+++ b/src/client/opamLockCommand.ml
@@ -96,12 +96,13 @@ let select_packages atom_locs st =
                atoms
            in
            if missing <> [] then
-             (OpamConsole.error_and_exit
-                "Skipping %s, dependencies are not satisfied in this switch, \
-                 not installed packages are:\n%s"
-                (OpamPackage.to_string nv)
-                (OpamStd.Format.itemize OpamFormula.string_of_atom missing);
-              acc)
+            (OpamConsole.error_and_exit `Not_found
+              "Operation aborted: 
+              Dependencies for %s required in the current switch are not met. \  
+              \nPlease ensure the following packages are installed:\n%s"
+              (OpamPackage.to_string nv)
+              (OpamStd.Format.itemize OpamFormula.string_of_atom missing);
+          acc)
            else
              OpamPackage.Set.add nv acc)
          names OpamPackage.Set.empty


src/client/opamLockCommand.ml
```diff
@@ -96,7 +96,8 @@ let select_packages atom_locs st =
                atoms
            in
            if missing <> [] then
-             (OpamConsole.error
+             (OpamConsole.error_and_exit
+                `Aborted
```
kit-ty-kate, line 100: Not_found is usually the reason used in this type of case:
```suggestion
             OpamConsole.error_and_exit `Not_found
```

src/client/opamLockCommand.ml
```diff
@@ -96,7 +96,8 @@ let select_packages atom_locs st =
...
                 "Skipping %s, dependencies are not satisfied in this switch, \
                  not installed packages are:\n%s"
```
kit-ty-kate, line 102: The error message doesn't correspond to the action anymore (skipping vs. aborting)

kit-ty-kate (commented):
thanks for your contribution. Aside from the two individual comments, i can note that the if clause should change from as its return value isn't used in the `true` case anymore. Could you change the shape of the if to the following?:
```
if ... then
   error_and_exit... ;
OpamPackage.Set.add nv acc

src/client/opamLockCommand.ml
progressive-galib: done

progressive-galib:
can i talk to you about something else here or if you could msg me somewhere else ? i want to contribute more to opam more, i am just getting started so if you could suggest some easy/small issues and walk me through the source tree/code structure that would be nice

src/client/opamLockCommand.ml
progressive-galib: my new error msg might not make any sense grammatically,

 as English is not  first language help a bit, pls.

progressive-galib:
my new error msg might not make any sense grammatically,

as English is not first language help a bit, pls.

progressive-galib:
@kit-ty-kate done merge please

also can i talk to you about something else here or if you could msg me somewhere else ? i want to contribute more to opam more, i am just getting started so if you could suggest some easy/small issues and walk me through the source tree/code structure that would be nice

progressive-galib:
@rjbou @kit-ty-kate any update ?

kit-ty-kate:
Hi, sorry for the wait.
First of all, while i know English isn't your first language, i'd like to invite you to use a less pushy language when requesting action from us maintainers here and elsewhere, even more when requesting action during the winter break when the change isn't critical.

Regarding your request for discussion, both @rjbou and i have our email addresses displayed on our github profile, feel free to send us an email.

Regarding this here PR, i don't think the new behaviour matches the requested behaviour in https://github.com/ocaml/opam/issues/5520. I'm not sure it makes sense to stop as soon as a missing dependency is detected. Maybe keeping the same error message as before but simply returning the fact that an error occurred from this function, then use it to call `OpamStd.Sys.exit_because` if there was an error at the end of `OpamCommands.lock.lock` might be a better behaviour. This way the behaviour for (non-script) users of `opam lock` does not change while allowing script users to make sure the dependencies are all installed.
What do you think? As a user, how would you expect opam lock to work?

progressive-galib:
> Hi, sorry for the wait. First of all, while i know English isn't your first language, i'd like to invite you to use a less pushy language when requesting action from us maintainers here and elsewhere, even more when requesting action during the winter break when the change isn't critical.

i am learning English so next time my language wont be so pushy. and i will make no request on any holiday or sunday next time.


> Regarding your request for discussion, both @rjbou and i have our email addresses displayed on our github profile, feel free to send us an email.

thank you 

> Regarding this here PR, i don't think the new behaviour matches the requested behaviour in #5520. I'm not sure it makes sense to stop as soon as a missing dependency is detected. Maybe keeping the same error message as before but simply returning the fact that an error occurred from this function, then use it to call `OpamStd.Sys.exit_because` if there was an error at the end of `OpamCommands.lock.lock` might be a better behaviour. 

by same error msg do you mean this msg ?

`             (OpamConsole.error
                "Skipping %s, dependencies are not satisfied in this switch, \
                 not installed packages are:\n%s"
                (OpamPackage.to_string nv)
                (OpamStd.Format.itemize OpamFormula.string_of_atom missing);
              acc)
`
              
or this msg ?

``            (OpamConsole.error_and_exit `Not_found
              "Operation aborted: 
              Dependencies for %s required in the current switch are not met. \  
              \nPlease ensure the following packages are installed:\n%s"
              (OpamPackage.to_string nv)
              (OpamStd.Format.itemize OpamFormula.string_of_atom missing);
          acc)
``
> then use it to call OpamStd.Sys.exit_because if there was an error at the end of OpamCommands.lock.lock might be a better behaviour.
   
so the error is opamLockCommand then handled in OpamCommands.lock.lock ?

kit-ty-kate:
> by same error msg do you mean this msg ?
> 
> ` (OpamConsole.error "Skipping %s, dependencies are not satisfied in this switch, \ not installed packages are:\n%s" (OpamPackage.to_string nv) (OpamStd.Format.itemize OpamFormula.string_of_atom missing); acc)`

This one yes

> so the error is opamLockCommand then handled in OpamCommands.lock.lock ?

i meant i think the error message should stay where it is but the actual error (by which i mean the exception raised by `OpamStd.Sys.exit_because`) should be handled in `OpamCommands.lock.lock`

progressive-galib:
> i meant i think the error message should stay where it is but the actual error (by which i mean the exception raised by `OpamStd.Sys.exit_because`) should be handled in `OpamCommands.lock.lock`

if we were doing that do we still raise OpamConsole.error_and_exit or do we raise softer OpamConsole.error ?

PR #6343, opened 2024-12-19
Stop double pin of packages located in ./opam/opam

### LINKED ISSUE #6342: Double pin of package when in `opam directory`
cf [lock reftest](https://github.com/ocaml/opam/blob/master/tests/reftests/lock.test#L385)
```
### <pin:old/opam/opam>
opam-version: "2.0"
name: "silver"
depends: "foo"
### # ERROR double silver
### opam pin -n old
This will pin the following packages: silver, silver. Continue? [Y/n] y
Package silver does not exist, create as a NEW package? [Y/n] y
silver is now pinned to file://${BASEDIR}/old (version dev)
[NOTE] Package silver is already pinned to file://${BASEDIR}/old (version dev).
silver is now pinned to file://${BASEDIR}/old (version dev)

kit-ty-kate:
Fixes #6342 

The problem was that the `files_in_source` function would call its internal scanning function `files` twice on `<dir>` and `<dir>/opam`, however a side close of that subfunction would itself add `<dir>/opam/opam` if it is detected, which leads to the same file being detected twice.


rjbou (approved):
I've added a test in pin test (lock test is a side effect, it's not its purpose).
Otherwise, lgtm!

PR #6358, opened 2025-01-07
Display the version number of packages flagged with avoid-version/deprecated gray

### LINKED ISSUE #6354: opam show ocaml should not show unreleased versions
Similarly to https://github.com/ocaml/opam/pull/6098, it is surprising for users to see `opam show ocaml` display `5.4.0` (unreleased at the time of writing and only available via the `ocaml-variants.5.4.0+trunk` package which is tagged with `avoid-version`) as the latest available version (`all-versions` field).

To remedy that, i propose to hide unavailable (currently displayed red and crossed, if the terminal allows it) as well as versions only available using avoid-version/deprecated dependencies by default from the `all-versions` field unless `--all` is given.

kit-ty-kate:
Example of the new output with `opam show ocaml-base-compiler`:
scrn-2025-01-07-20-00-13

Fixes #6354, probably the less intrusive solution if the color scheme works for you. Compared to removing them completely as suggested in the ticket, it brings an issue of clarity of meaning between what yellow or red means, although this concern seems pretty minimal to me compared to the other UX concerns brought by a potential addition of `--all`.

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -921,7 +921,18 @@ let show cli =
         available packages as well as a short description for each.";
     `P "Paths to package definition files or to directories containing package \
         definitions can also be specified, in which case the corresponding \
-        metadata will be shown."
+        metadata will be shown.";
+    `P "Note: on terminal supporting it, version numbers can be colored the \
+        following ways:";
+    `Noblank;
+    `I ("-", "Magenta and bold: versions installed the current switch");
+    `Noblank;
+    `I ("-", "Bold: versions installed in other switches");
+    `Noblank;
+    `I ("-", "Red and crossed: versions not available on the current platform");
+    `Noblank;
+    `I ("-", "Gray: versions tagged with the $(i,avoid-version) or \
+              $(i,deprecated) flag");
   ] in
   let fields =
     mk_opt ~cli cli_original ["f";"field"] "FIELDS"
--- a/master_changes.md
+++ b/master_changes.md
@@ -213,6 +213,7 @@ users)
   * `OpamFile.Repos_config.t`: change the type to not allow repositories without an URL [#6249 @kit-ty-kate]
 
 ## opam-core
+  * `OpamConsole`: Replace `black` text style (unused and not very readable) by `gray` [#6358 @kit-ty-kate]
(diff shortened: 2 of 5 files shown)


rjbou (commented):
Now that there is 2 colors, maybe we should add a note in the manpage for the meaning of colors.

kit-ty-kate:
Per discussion at today's meeting, i've switched the color to gray instead of crossed+yellow to avoid confusion on whether it is possible to install the affected version or not

kit-ty-kate:
rebased

src/client/opamCommands.ml
```diff
@@ -921,7 +921,17 @@ let show cli =
...
+    `P "Note: on terminal supporting it, version numbers can be colored the \
+        following ways:";
+    `Noblank;
+    `I ("-", "Versions installed the current switch are displayed bold and magenta");
+    `Noblank;
+    `I ("-", "Versions installed in other switches are displayed bold");
+    `Noblank;
+    `I ("-", "Versions not available on the current platform are displayed crossed and red");
+    `Noblank;
+    `I ("-", "Versions tagged with the avoid-version or deprecated flag are displayed gray");
```
rjbou, line 934:
```suggestion
    `I ("-", "Versions tagged with the $(i, avoid-version) or $(i, deprecated) flag are displayed gray");
```

rjbou, line 934: I'm wondering if it is better to invert, like that information is available at first look:
```suggestion
    `I ("-", "Magenta and bold: versions installed the current switch");
    `Noblank;
    `I ("-", "Bold: versions installed in other switches");
    `Noblank;
    `I ("-", "Red and crossed: versions not available on the current platform");
    `Noblank;
    `I ("-", "Gray: versions tagged with the avoid-version or deprecated flag");
```

rjbou (commented):
The second commit should mention that the scope is opam show (prefix, in the text, etc.)

rjbou (approved):
:ok_hand:

rjbou:
If it is possible, can you add a picture with the gray version?

kit-ty-kate:
scrn-2025-01-20-18-21-32

PR #6367, opened 2025-01-18
Fix Debian badges and links and instructions

tasuki:
Seeing as these have been out of date for some ten years, perhaps it'd be wiser to not include links to anything other than unstable. But hey, the badges will look nice for a short while...

### DIFF
--- a/doc/pages/Install.md
+++ b/doc/pages/Install.md
@@ -85,13 +85,11 @@ yay -S opam-git
 
 #### Debian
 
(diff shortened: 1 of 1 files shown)


kit-ty-kate:
Thanks for your contribution. After review i had a look at another way to display all these information which might supersede your contribution in a way that would improve the maintenance burden for this documentation: https://github.com/ocaml/opam/pull/6372

tasuki:
Ok! I'll close this then...

PR #6372, opened 2025-01-20
Improve the installation documentation

kit-ty-kate:
While reviewing https://github.com/ocaml/opam/pull/6367 i realized the installation documentation is outdated and would require constant update not to be.

This PR does:
- Replace the individual badge for each distribution&version by repology's vertical-allrepos with its exclude_unsupported=1 option (see https://repology.org/project/referencing/badges) to only show the distribution version not EoL.
- Remove Exherbo from the list of supported distribution as it only provides the 11 years out-of-date opam 1.1.1
- Add Alpine Linux to the list
- Merge the Debian and Ubuntu steps
- Remove mentions of CentOS and RHEL which do not have the opam package
- Remove the Distribution.md file which is extremely out-of-date, redundant with the new automatically generated repology badge and requires a huge effort to maintain

as well as some general fixes and improvements


doc/pages/Install.md
```diff
@@ -60,114 +61,58 @@ sudo install <downloaded file> /usr/local/bin/opam
...
 ## Using your system's package manager
 
-This is generally the recommended way, **when available and up-to-date** (you
-can check [here](Distribution.html) the latest
-available release per distribution). Here is a list of supported distributions:
+This is generally the recommended way, **when available and up-to-date**. You
+can check the following table:
 
-#### Arch Linux
+[![repology package status](https://repology.org/badge/vertical-allrepos/opam.svg?exclude_unsupported=1)](https://repology.org/project/opam/versions)
```
rjbou, line 67: The list is quite long, i'm not sure that it should here but it's better than at the end i think...

doc/pages/Install.md
```diff
@@ -60,114 +61,58 @@ sudo install <downloaded file> /usr/local/bin/opam
...
 
-[![badge](https://repology.org/badge/version-for-repo/arch/opam.svg)](https://repology.org/project/opam/versions)
+If your distribution is up-to-date you can install opam the following way:
```
rjbou, line 69: Should this line be before the long table, in case people don't scroll down to see install instructions?

rjbou (commented):
The commit message should contains the details that is in the first comment

doc/pages/Install.md
kit-ty-kate: I think it makes much more sense at the beginning rather than the end.
The way users read it is:
1) what system do i have?
2) is the system package up-to-date?
  * If yes use that one (if you don't know how, the list would be right after)
  * If not, then the section just before on the binary installer will do

The per-distribution list is only useful for a minority of people (beginners), whereas the repology list is useful much more often in my opinion

doc/pages/Install.md
kit-ty-kate: done

rjbou (approved):

PR #6373, opened 2025-01-22
Fix regression on 'opam upgrade <package>' upgrading unrelated things

AltGr:
Regression introduced in 4db0c139ce184e7f49bdde63c8ffb6dbd1bb87d4, which is
dedicated to a different case (functionality of this patch is preserved)

### DIFF
--- a/tests/reftests/upgrade.test
+++ b/tests/reftests/upgrade.test
@@ -432,7 +432,6 @@ Done.
 => Note: This package is deprecated.
 ### :::::::::::::::::::::::
 ### : Make sure opam upgrade <pkg> only tries to upgrade <pkg> and not anything else
-### : TODO: broken
 ### :::::::::::::::::::::::
 ### opam upgrade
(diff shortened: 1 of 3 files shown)


tests/reftests/upgrade.test
```diff
@@ -437,6 +437,10 @@ The following actions would be performed:
 ### opam upgrade --show -a uninstalled
 uninstalled is not installed. Install it? [Y/n] y
 The following actions would be performed:
+=== upgrade 3 packages
+  - upgrade bar         2 to 3
+  - upgrade baz         1 to 2 [uses foo]
+  - upgrade foo         3 to 4
```
rjbou, line 443: Can you link the origin of the regression ?

kit-ty-kate: it's the same commit

rjbou (approved):
Reviewed by @AltGr too; lgtm!

PR #6376, opened 2025-01-24
Do not pre-write the answer to questions with the default anwser

kit-ty-kate:
Supersedes https://github.com/ocaml/opam/pull/6290
This implements https://github.com/ocaml/opam/pull/6289#pullrequestreview-2428097616

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -233,6 +233,7 @@ users)
 
 ## opam-core
   * `OpamConsole`: Replace `black` text style (unused and not very readable) by `gray` [#6358 @kit-ty-kate]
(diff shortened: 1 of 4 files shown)


rjbou (approved):
On the code itself, lgtm!
The commit should be split in 2 commits

kit-ty-kate:
> The commit should be split in 2 commits

why? I don't think this is reasonable without introducing new bugs in between the two commits

kit-ty-kate:
> I don't think this is reasonable without introducing new bugs in between the two commits

sorry nevermind, i was thinking of something else. Done.

rjbou (approved):

PR #6377, opened 2025-01-26
the opam lock command now exits with a non-zero exit code when dependencies are not yet installed in the current switch issue:#5520 reopened pr#6341

progressive-galib:
this might fix https://github.com/ocaml/opam/issues/5520

i mistakenly deleted the branch that was on https://github.com/ocaml/opam/pull/6341 thus i am re opening this pull request.

sorry for my silly mistake

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -4409,7 +4409,9 @@ let lock cli =
     let pkg_done =
       OpamPackage.Set.fold (fun nv msgs ->
           let opam = OpamSwitchState.opam st nv in
-          let locked = OpamLockCommand.lock_opam ~only_direct st opam in
+          let locked = try OpamLockCommand.lock_opam ~only_direct st opam 
+          with e -> exit 1
+          in
           let locked_fname =
             OpamFilename.add_extension
               (OpamFilename.of_string (OpamPackage.name_to_string nv))


progressive-galib:
@kit-ty-kate i am galib from https://github.com/ocaml/opam/pull/6341#issuecomment-2603328412

kit-ty-kate:
Hi, sorry but have you tested this code to make sure what you think it does? Reading it i can see this change doesn't do anything and i think it would be a better use of both of our time if things that you would like to work on were tested locally prior to sending it to us for review.

If you wish to learn OCaml from contributing to opam i think this is great, but i think most of the learning comes from reading the current code and testing what it does on your own, and only asking when you're stuck or the documentation is insufficient. If you jump to the last step without doing the former, there can only be so much actual learning from that.
For example from your question in https://github.com/ocaml/opam/pull/6341#issuecomment-2614328901 if you read the code or tested it, you would see that `OpamConsole.error` does not raise any exception, it only displays a message on screen.

progressive-galib:
ohh sorry this was due to me mistakenly deleting the branch and not taking my account my changes to opamLockCommand.ml i did in other branches.sorry for this.

i will fix it before Friday.

also please what type of exception should I raise ? i saw that opam has its own opamstd or something. does any specific internally defined exception should be raised or should a generic one do ?

progressive-galib:
@kit-ty-kate @rjbou done, and before friday as promised,

https://github.com/progressive-galib/opam/blob/galib-lock-exit-error-i5520/src/client/opamLockCommand.ml#L99-L107

currently due to exception the code never reaches the `acc)` before exeption is raised so should we do 

https://gist.github.com/progressive-galib/ef44f6ccf142e660c98abb62e9ec6284 this ?

progressive-galib:
done ignore what i said earlier benchmark is really negligible and also removed unreachable code @kit-ty-kate please merge

PR #6375, opened 2025-01-22
Do not ask to install pin-depends twice

kit-ty-kate:
Fixes #6374

Issue #6374, opened 2025-01-22
Refusing local pinnings on `opam switch create` repeats the question


AltGr:
Transcript:
```shell-session
$ opam switch create . ocaml.5.1.2 --ignore-constraints-on ocaml
The following additional pinnings are required by tryocaml.0.10:
  - js_of_ocaml.3.10.1 at git+https://github.com/AltGr/js_of_ocaml#fix-bytecode
  - js_of_ocaml-compiler.3.10.1 at git+https://github.com/AltGr/js_of_ocaml#fix-bytecode
  - js_of_ocaml-lwt.3.10.1 at git+https://github.com/AltGr/js_of_ocaml#fix-bytecode
  - js_of_ocaml-ppx.3.10.1 at git+https://github.com/AltGr/js_of_ocaml#fix-bytecode
  - js_of_ocaml-tyxml.3.10.1 at git+https://github.com/AltGr/js_of_ocaml#fix-bytecode
  - ocp-indent-nlfork.1.5.4 at git+https://git@github.com/OCamlPro/ocp-indent.git#nlfork
Pin and install them? [Y/n] n
Try to install anyway, assuming `--ignore-pin-depends'? [Y/n] y
tryocaml is now pinned to git+file:///home/lg/ocamlpro/tryocaml#tryocaml (version 0.10)
The following additional pinnings are required by tryocaml.0.10:
  - js_of_ocaml.3.10.1 at git+https://github.com/AltGr/js_of_ocaml#fix-bytecode
  - js_of_ocaml-compiler.3.10.1 at git+https://github.com/AltGr/js_of_ocaml#fix-bytecode
  - js_of_ocaml-lwt.3.10.1 at git+https://github.com/AltGr/js_of_ocaml#fix-bytecode
  - js_of_ocaml-ppx.3.10.1 at git+https://github.com/AltGr/js_of_ocaml#fix-bytecode
  - js_of_ocaml-tyxml.3.10.1 at git+https://github.com/AltGr/js_of_ocaml#fix-bytecode
  - ocp-indent-nlfork.1.5.4 at git+https://git@github.com/OCamlPro/ocp-indent.git#nlfork
Pin and install them? [Y/n] n
Try to install anyway, assuming `--ignore-pin-depends'? [Y/n] y
```

I'd be best not to ask the question twice

kit-ty-kate:
[...]

PR #6380, opened 2025-02-07
reftest: add pin depends special case test

rjbou:
relatd to #6375 & #6374


kit-ty-kate (approved):
otherwise lgtm

kit-ty-kate:
Thanks!

PR #6391, opened 2025-02-19
`opam admin`: enforce repository root check for all commands and add upgrade advice

rjbou:


src/client/opamAdminCommand.ml
```diff
@@ -16,13 +16,33 @@ open Cmdliner
...
   then
     OpamConsole.error_and_exit `Bad_arguments
       "No repository found in current directory.\n\
        Please make sure there is a \"packages%s\" directory" OpamArg.dir_sep;
+  (if check then
+     let default_current = "2.0" in
+     match repo_version_lt repo_root
+             (OpamVersion.of_string default_current) with
+     | Some v ->
+       OpamConsole.warning
```
kit-ty-kate, line 41: I think opam admin should simply refuse to handle 1.2 repositories
```suggestion
       OpamConsole.error_and_exit
```

rjbou: It can be done that way too. It need a proper cleaning of the code if there is some that handles 1.2 backward compatibility (expect upgrade ofc).

src/client/opamAdminCommand.ml
```diff
@@ -16,13 +16,33 @@ open Cmdliner
...
+
+let checked_repo_root ?(check=true)() =
   let repo_root = OpamFilename.cwd () in
   if not (OpamFilename.exists_dir (OpamRepositoryPath.packages_dir repo_root))
   then
     OpamConsole.error_and_exit `Bad_arguments
       "No repository found in current directory.\n\
        Please make sure there is a \"packages%s\" directory" OpamArg.dir_sep;
+  (if check then
+     let default_current = "2.0" in
```
kit-ty-kate:
```suggestion
     let default_current = OpamAdminRepoUpgrade.upgradeto_version_string in
```

src/client/opamAdminCommand.ml
```diff
@@ -16,13 +16,33 @@ open Cmdliner
...
+  let v' =
+    let open OpamStd.Option.Op in
+    (OpamFile.Repo.read_opt
+       (OpamRepositoryPath.repo repo_root)
+     >>= OpamFile.Repo.opam_version)
+    +! (OpamVersion.of_string "1.2")
+  in
+  if OpamVersion.compare v v' > 0 then Some v' else None
+
+let checked_repo_root ?(check=true)() =
```
kit-ty-kate:
```suggestion
let checked_repo_root ?(check=true) () =
```

src/client/opamAdminCommand.ml
```diff
@@ -16,13 +16,33 @@ open Cmdliner
...
 type command = unit Cmdliner.Term.t * Cmdliner.Cmd.info
 
-let checked_repo_root () =
+let repo_version_lt repo_root v =
+  let v' =
+    let open OpamStd.Option.Op in
+    (OpamFile.Repo.read_opt
+       (OpamRepositoryPath.repo repo_root)
+     >>= OpamFile.Repo.opam_version)
+    +! (OpamVersion.of_string "1.2")
```
kit-ty-kate: I think this should also probably be be extracted from OpamAdminRepoUpgrade line 479, put at the top of that module same as upgradeto_version_string, exported and used here instead.

kit-ty-kate (commented):
Looks good though I have some suggestions below.
I'm also wondering if we could move the test commit to be the first one of this PR to also see the result of the first commit

src/client/opamAdminCommand.ml
```diff
@@ -758,18 +778,19 @@ let upgrade_command cli =
   in
   let cmd global_options clear_cache create_mirror () =
     OpamArg.apply_global_options cli global_options;
+    let repo_root = checked_repo_root ~check:false () in
     if clear_cache then OpamAdminRepoUpgrade.clear_cache ()
     else match create_mirror with
```
kit-ty-kate: The repository doesn't need to be checked in `--clear-cache` mode
```suggestion
    if clear_cache then OpamAdminRepoUpgrade.clear_cache ()
    else
      let repo_root = checked_repo_root () in
      match create_mirror with
```

src/client/opamAdminCommand.ml
rjbou: issue created #6743

kit-ty-kate (approved):

kit-ty-kate:
ocaml-benchmarks failed due to a lack of disk space on the test machine

PR #6403, opened 2025-02-27
Set log dir as temporary directory on safe mode

### LINKED ISSUE #6402: External command run writes log on opam root with `--safe`
Extracted from #6400, see https://github.com/ocaml/opam/issues/6400#issuecomment-2686006317
Opam tries to write in log dir which is `<opamroot>/log`, making the sandbox fail to write.
There is two solutions: add `<opamroot>/log` as read-write in the sandbox script, or in safe mode set the log dir as temporary directory.

rjbou:
fix #6402 

TODO:
* [ ] changelog
* [ ] commit message to rewrite


tests/reftests/safe.test
```diff
@@ -0,0 +1,47 @@
...
+all:
+	echo \$(shell ${OPAM} var os-version --safe)
+	${OPAM} var os-version --safe
+	echo done
+EOF
+### sh cat-makefile.sh varos/Makefile
+### opam switch create varos --empty
+### # opam install foo bar
+### opam pin ./varos -n
+[NOTE] Package varos does not exist in opam repositories registered in the current switch.
```
rjbou, line 22: The previsou version (not fixed) is not portable

rjbou:
From dev meeting: see if it is possible to output stdout instead of writing for sys polling.

PR #6409, opened 2025-03-05
Fix opam file lookup when a pin is done with a local directory and a branch is specified 

### LINKED ISSUE #6408: `opam pin ./local#mybranch` does not take `mybranch` opam file
When we pin a local VCS directory, if the locally fetched branch is not the one that is specified in argument, opam look up for the local directory without taking into account the branch.
Example:
```
> git checkout snd
Switched to branch 'snd'
> git ls-files
snd.opam
> git checkout fst
Switched to branch 'fst'
> git ls-files
fst.opam
> cd ..
> opam pin snd ./mydir#snd -n
[NOTE] Package snd does not exist in opam repositories registered in the current switch.
[snd.dev] synchronised (git+file://path/to/mydir#snd)
[NOTE] No package definition found for snd.dev: please complete the template
[WARNING] The opam file didn't pass validation:
             error 22: Some fields are present but empty; remove or fill them: "homepage", "license", "bug_reports"
             error 57: Synopsis must not be empty
           warning 62: License doesn't adhere to the SPDX standard, see https://spdx.org/licenses/ : ""
Proceed anyway ('no' will re-edit)? [Y/n] y
You can edit this file again with "opam pin edit snd", export it with "opam show snd --raw"
snd is now pinned to git+file://path/to/local#snd (version dev)
```
This is due to the magic lookup of an opam file : it should be retrieved even if it is not committed.

rjbou:
fix #6408 
I'm not fully satisfied with that solution... But it may not be kept, superseded by #6419.

~This PR changes the behaviour of `opam pin add ./local-vcs` which would take the opam files from the VCS repository even if they had untracked changes. This behaviour exists since opam 2.1 (see https://github.com/ocaml/opam/pull/4300) but opam 2.0 didn't have the same behaviour.~
~The addition of a note about untracked changes can be added separately and is tracked by https://github.com/ocaml/opam/issues/6414~

### DIFF
--- a/src/client/opamPinCommand.ml
+++ b/src/client/opamPinCommand.ml
@@ -66,7 +66,7 @@ exception Fetch_Fail of string
 
 let get_source_definition ?version ?subpath ?locked st nv url =
   let root = st.switch_global.root in
-  let srcdir = OpamPath.Switch.pinned_package root st.switch nv.name in
+  let internal_pindir = OpamPath.Switch.pinned_package root st.switch nv.name in
   let fix opam =
     OpamFile.OPAM.with_url url @@
     (match version with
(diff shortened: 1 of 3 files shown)


src/client/opamPinCommand.ml
```diff
@@ -75,15 +75,31 @@ let get_source_definition ?version ?subpath ?locked st nv url =
...
+          let url = OpamUrl.of_string (OpamFilename.Dir.to_string d) in
+          OpamRepository.revision d
+            { url with
+              OpamUrl.transport = u.transport;
+              backend = u.backend}
+        in
+        get_branch dir @@+ fun local_branch ->
+        get_branch internal_pindir @@| fun distant_branch ->
+        (match local_branch, distant_branch with
+         | Some l, Some d when OpamPackage.Version.equal l d -> dir
```
kit-ty-kate, line 95: i'm not sure why `OpamRepository.revision` returns a version type but this might be something we might want to change to `string`. It is rather confusing to see it used like a version number.

kit-ty-kate:
After lengthy reviewing and tinkering i wasn't satisfied with the solution either so i came up with a different proposal, which while a bit radical i strongly believe to be a positive change for users:

Currently pinning a local vcs directory will override the opam files with whatever the version in the working directory is, regardless of whether or not `--working-dir` is given. I think this behaviour is too surprising to users. If they wish to use the working directory version then they can pass `--working-dir` explicitly. This also has the side-effect to simplify our code quite a lot which is always nice.

rjbou:
Putting back the PR to its original state as the fix of #6408. I've opened #6418 to discuss default behaviour change and #6419 the part of the PR that contains the wip.

kit-ty-kate (changes requested):
I strongly believe the magic should be removed all together instead as mentioned above

PR #6411, opened 2025-03-06
lock: add '--keep-local' argument to add local pins to 'pin-depends:' field

### LINKED ISSUE #4897: Add a flag to opam lock to keep local pins
By default, local pins are not kept in lock file. It can be useful to share a setup (after editing paths).

rjbou:
fix #4897

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -4392,9 +4392,10 @@ let lock cli =
         fixed and the same filter is on all dependencies that are added from \
         them";
     `P "- $(i,pin-depends) are kept and new ones are added if in the \
-        dependencies some packages are pinned ";
+        dependencies some packages are pinned";
     `P "- pins are resolved: if a package is locally pinned, opam tries to get \
-        its remote url and branch, and sets this as the target URL";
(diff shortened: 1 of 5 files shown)


kit-ty-kate:
I pushed two fixup commits fixing my comments. What do you think?

src/client/opamCommands.ml
```diff
@@ -4392,9 +4392,10 @@ let lock cli =
         fixed and the same filter is on all dependencies that are added from \
         them";
     `P "- $(i,pin-depends) are kept and new ones are added if in the \
-        dependencies some packages are pinned ";
+        dependencies some packages are pinned";
     `P "- pins are resolved: if a package is locally pinned, opam tries to get \
-        its remote url and branch, and sets this as the target URL";
+        its remote url and branch, and sets this as the target URL; otherwise \
+        it ignore them. You can use $(i,--keep-local) to keep the local url.";
```
kit-ty-kate, line 4398:
```suggestion
        it ignore them. You can use $(i,--keep-local) to keep the local urls.";
```

kit-ty-kate (approved):
LGTM otherwise

kit-ty-kate:
Thanks!

PR #6412, opened 2025-03-06
reftest: untie lock test from OPAMEDITOR hack, and fix lock VCS remote detection

rjbou:
The change in lock test is needed for #6319.
Queued on 
  * [x] #6411
  
edit: 
* untied from #6409 
* No more in the PR: While changing the test, i found the bug with git `remote get-url`, which is specific to `opam lock` handling, extracted in #6422


kit-ty-kate (approved):

kit-ty-kate:
Thanks a lot!

PR #6431, opened 2025-03-24
Allow detection of installed packages through their virtual names on ALT Linux, RHEL-based and SUSE-based distributions

### LINKED ISSUE #6426: regression: opem fails to recognize rpm Provides
`opam install conf-python-3` fails because for some reason `rpm -q --whatprovides python3 == python313-3.13.2-3.1.x86_64` is not recognized.

Is there some sort of `--debug-solver` knob to see what attempts opam 2.3.0 actually does to find something that provides `python3`?

This is a regression between 2.0.10 and 2.1.0~beta2:

````
opam init
opam install opam-devel.2.0.10
alias opam2="OPAMROOT=~/.opam2 ~/.opam/default/lib/opam-devel/opam"
rm -rf ~/.opam2 ; opam2 --version && opam2 init --disable-sandboxing && opam2 install conf-python-3
opam install opam-devel.2.1.0~beta2
rm -rf ~/.opam2 ; opam2 --version && opam2 init --disable-sandboxing && opam2 install conf-python-3
````

### PR DESCRIPTION
**Pull request by kit-ty-kate**
Fixes https://github.com/ocaml/opam/issues/6426
Queued on 
* https://github.com/ocaml/opam/pull/6464
* #6489

Per discussion in the ticket, getting the list of available packages on SUSE-based distributions is too costly.
Instead, the solution is one similar to https://github.com/ocaml/opam/pull/4791 which was chosen for RHEL-based distributions: simply disable the detection of available packages.

However, while with the original issue the user will be able to install `conf-python-3` on openSUSE Tumbleweed, opam will continue to detect that the package isn't installed and continuously bug the user about the missing `python3` package which is in fact already installed.

To fix that we use the `%{PROVIDES}` tag (documented in http://ftp.rpm.org/max-rpm/ch-queryformat-tags.html):
- Pros: adds support for virtual packages on RHEL-based distributions as well
- Cons: increases the time it takes to get the availability information by 18% (locally on Fedora: before takes 1.5s, now takes 1.77 seconds)

Related to https://github.com/ocaml/opam/issues/4759

### DIFF
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -785,8 +785,9 @@ let installed_packages_t ?(env=OpamVariable.Map.empty) config packages =
          slower to parse due to the extra number of package name but
          allows to detect virtual packages such as python3. *)
       run_query_command "rpm" ["-qa"; "--qf"; "%{NAME}\\n[%{PROVIDES}\\n]"]
-      |> List.map OpamSysPkg.of_string
-      |> OpamSysPkg.Set.of_list
+      |> List.fold_left (fun acc name ->
+          OpamSysPkg.Set.add (OpamSysPkg.of_string name) acc)
+        OpamSysPkg.Set.empty
     in
     get_relevant sys_installed
   | Cygwin ->
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -779,7 +779,12 @@ let installed_packages_t ?(env=OpamVariable.Map.empty) config packages =
        >python3-pip-wheel
     *)
     let sys_installed =
-      run_query_command "rpm" ["-qa"; "--qf"; "%{NAME}\\n"]
+      (* NOTE: In practice %{PROVIDES} seems to always contains %{NAME}
+         but this behaviour isn't documented, so just to be sure, it is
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/state/opamSysInteract.ml [hunk 1]
```diff
@@ -779,9 +779,13 @@ let installed_packages_t ?(env=OpamVariable.Map.empty) config packages =
        >python3-pip-wheel
     *)
     let sys_installed =
-      run_query_command "rpm" ["-qa"; "--qf"; "%{NAME}\\n"]
-      |> List.map OpamSysPkg.of_string
-      |> OpamSysPkg.Set.of_list
+      (* NOTE: In practice %{PROVIDES} seems to always contains %{NAME}
+         but this behaviour isn't documented, so just to be sure, it is
+         safer to add %{NAME} anyway. *)
```
```suggestion
         safer to add %{NAME} anyway.
         Using %{PROVIDES}% is a little bit slower but permit to detect
         virtual packages as python3. *)
```

rjbou (approved):

rjbou:
Affected

> Cons: increases the time it takes to get the availability information by 18% (locally on Fedora: before takes 1.5s, now takes 1.77 seconds)

Affected commands are `opam install`, `opam pin`, and `opam list`. The first two take already some time, so 0.2s would be unnoticed. It could be noticed for `opam list` though but the counterpart is to have a more precise depext detection.

PR #6435, opened 2025-03-26
Upgrade to cmdliner 2.0.0

### LINKED ISSUE #6425: cmdliner 2.0.0 support
It seems that the latest version of your package on the opam repository will not compile with the upcoming cmdliner 2.0.0 release.

This is either because you are using deprecated identifiers that are removed in 2.0.0 or because you still use argument converters as pairs which has been (sadly only silently) deprecated since 2017.

There are a few instruction [here](https://github.com/dbuenzli/cmdliner/issues/206) on how to make it compatible. If you run into problem or need more help use this issue, I'm listening.

kit-ty-kate:
Fixes #6425 

Issues to address:
- [ ] I'm personally unconvinced about the `CMDLINER_LEGACY_PREFIXES=true` escape hatch. I would much rather have a global reference (e.g. https://github.com/kit-ty-kate/cmdliner/commit/94ecc8a6f1cb3b5b07fa34bd0999e72349325f7c) or whatever to make sure the behaviour doesn't change between two versions while we weren't looking
- [ ] The escape hatch in combination with `?env` doesn't work with `Arg.enum`. Using the example patch above, this issue would be fixed.

If we decide to fork cmdliner with the following patch above master https://github.com/kit-ty-kate/cmdliner/commit/1565763d2af8515f3bc6212ec359fe4769891242, those issues wouldn't be there anymore at the cost of extra maintenance.

cc @dbuenzli

### DIFF
--- a/src/client/opamAdminCommand.ml
+++ b/src/client/opamAdminCommand.ml
@@ -1378,12 +1378,12 @@ let help =
     | None       -> `Help (`Pager, None)
     | Some topic ->
       let topics = "topics" :: cmds in
-      let conv, _ = Cmdliner.Arg.enum (List.rev_map (fun s -> (s, s)) topics) in
-      match conv topic with
-      | `Error e -> `Error (false, e)
-      | `Ok t when t = "topics" ->
+      let conv = Cmdliner.Arg.enum (List.rev_map (fun s -> (s, s)) topics) in
+      match Arg.Conv.parser conv topic with
+      | Error e -> `Error (false, e)
+      | Ok t when t = "topics" ->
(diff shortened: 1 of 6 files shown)


dbuenzli:
Note that the idea of https://github.com/ocaml/opam/issues/6425 was to simply suppress unabstract uses of converter and make it compile with 1.3.0 without deprecation warnings to allow for a smooth eco-system transition. This uses 2.0.0 APIs.

Regarding the introduction of a global reference the answer is still no: global references are not compositional. The last thing I want to see is libraries fighting to set that value or change it behind user's back.

Issue #4255, opened 2020-06-30
Build directories kept again for pinned packages


Blaisorblade:
#266 came up again. I have installed packages without the `--keep-build-dir` option (or corresponding environment option), yet build folders were kept and occupied 3G. Anything that could explain it?
I've learned to use `opam clean` as a workaround, but I thought I'd file this issue for future users anyway.

Also, FWIW, I don't know how to reproduce it. I'll try 

```
$ opam config report
# opam config report
# opam-version      2.0.7
# self-upgrade      no
# system            arch=x86_64 os=linux os-distribution=debian os-version=10
# solver            builtin-mccs+glpk
# install-criteria  -removed,-count[version-lag,request],-count[version-lag,changed],-changed
# upgrade-criteria  -removed,-count[version-lag,solution],-new
# jobs              1
# repositories      2 (http), 2 (version-controlled) (default repo at 24cd8ca3)
# pinned            1 (git), 2 (version)
# current-switch    ocaml-variants.4.07.1+flambda
```

```
$ du -h --ma=1 ~/.opam/ocaml-variants.4.07.1+flambda/.opam-switch/
8.0K	/home/artifact/.opam/ocaml-variants.4.07.1+flambda/.opam-switch/backup
8.0K	/home/artifact/.opam/ocaml-variants.4.07.1+flambda/.opam-switch/config
188K	/home/artifact/.opam/ocaml-variants.4.07.1+flambda/.opam-switch/packages
12K	/home/artifact/.opam/ocaml-variants.4.07.1+flambda/.opam-switch/overlay
2.9G	/home/artifact/.opam/ocaml-variants.4.07.1+flambda/.opam-switch/build
87M	/home/artifact/.opam/ocaml-variants.4.07.1+flambda/.opam-switch/sources
676K	/home/artifact/.opam/ocaml-variants.4.07.1+flambda/.opam-switch/install
3.0G	/home/artifact/.opam/ocaml-variants.4.07.1+flambda/.opam-switch/
$ opam pin
coq.8.11.2+flambda-byte    version  8.11.2+flambda-byte
coq-autosubst.dev.coq86    version  dev.coq86
dot-iris-builddep.~dev     git      git+file:///home/artifact/dot-iris#final-artifact
$ opam list
# Packages matching: installed
# Name                # Installed               # Synopsis
base-bigarray         base
base-threads          base
base-unix             base
conf-findutils        1                         Virtual package relying on findutils
conf-m4               1                         Virtual package relying on m4
[...]

PR #6436, opened 2025-03-27
Do not keep the build directory of the pinned packages

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/4255 for good
Partially f.i.x.e.s https://github.com/ocaml/opam/issues/5448

Reasoning:

The build directory is cleared anyway before each builds and is available in `sources` already so keeping it around doesn't have any purpose and only wastes precious disk space.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -198,6 +198,7 @@ users)
   * Untie lock with pin depend test from OPAMEDITOR behaviour [#6412 @rjbou]
   * Add test for lint E63 [#6438 @rjbou]
   * Add a test for packages with subpath in a repository [#6439 @rjbou]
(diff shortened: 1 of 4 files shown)


rjbou (commented):
I've exported the test to its own test file, to add more test for the `OPAMKEEPBUILDDIR` environment variable. It may worth to add some test with `--keep-build-dir`?

kit-ty-kate:
> It may worth to add some test with `--keep-build-dir`?

done. I've also extended your new reftest with the missing "repository package with local and vcs sources" cases

rjbou (approved):
Thanks!

PR #6440, opened 2025-03-29
Save disk space by only caching dev packages into the internal 'sources' directory

### LINKED ISSUE #4056: Feature request: reuse extracted sources
If I do "opam install X1...Xn", opam starts by downloading sources for packages X1..Xn, or fetching them from the download cache. Then, each archive (Xi.tar.gz) is extracted into `<opamroot>/<switch>/.opam-switch/sources/<pkg.ver>`.

Even if all the archives are in the download cache, extracting all the archives takes a lot of time if there are many packages.

If one kills the "opam install" invocation (with ^C), either during the "archive unpacking phase" or during the build, then restarting it will not reuse the already-extracted sources, and will instead start by unpacking again all the archives from scratch.

To get more incrementality, one (brittle) solution would be to use whatever files there are in a sources directory if it is present, instead of unpacking the archive. However, if opam is killed while unpacking an archive, then the corresponding sources directory will be in a corrupted state (not containing all the files most likely), and will corrupt the next run of opam.

A more robust solution would be to memorize a mapping `<directory where sources are extracted> -> <hash of the archive they come from>`. A new mapping would be added only after an archive has been successfully extracted. Then, after having downloaded an archive, and before extracting it, one would be able to check whether the source directory already corresponds to that archive -- in which case one could simply avoid extracting the archive again.

Does that sound like a reasonable idea in principle, and would this "more robust" solution work?
If yes, where would this mapping be stored?

### LINKED ISSUE #5448: space savings in opam switches

kit-ty-kate:
Fixes #5448 
Fixes #4056
~Queued on #6912~

The `.opam-switch/sources` directory currently stores the source of every installed packages in the following format:
* `<pkgname>`: are for pinned packages
* `<pkgname>.<version>`: are for the rest

In https://github.com/ocaml/opam/pull/2825 (opam 2.0.0~beta) the change from `packages.dev` to `sources` was introduced to support (as i understand it) non-pinned dev packages (e.g. `ocaml-variants.5.4.0+trunk`). This is a good change but this also had for side-effect to store the source of every packages regardless of whether it's a dev package or not, which i think isn't a good change given that it duplicates data that's already in the archive cache that could instead be extracted very easily (the cost of extraction vs. copy is negligible).
As a side note, extracting the sources from their tarball would be way faster on Windows that extraction + copy for the same reason detailed in #5741 

~As it currently stands this PR is a WIP. It compiles but still has many unsolved issues (in particular it breaks caching for non-pinned dev packages) and the reftests fail, but i'm opening it anyway to show the rough area of the code that needs to change and in case anyone wants to take over this work while i'm doing something else.~

Post-WIP edit: Further improvements can be made by extracting the tarball directly into the build directory and would also f.i.x https://github.com/ocaml/opam/issues/6693, but i think it is simpler to do that separately in a future PR.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -31,6 +31,7 @@ users)
 
 ## Actions
   * Reorder the list of actions by increased priority [#6864 @kit-ty-kate - fix #6863]
+  * The internal `sources` directory now only serves to cache dev packages instead of every packages, saving disk space. For other packages it is now removed during the build phase. [#6440 @kit-ty-kate - fix #4056 #5448]
 
 ## Install
   * Remove the build directory as soon as possible when installing a package [#6906 @kit-ty-kate - fix #5884]
@@ -332,6 +333,7 @@ users)
   * `OpamSysInteract`: add `available_packages` and `installed_packages` to be computed separately, redefine `packages_status` accordingly. These funct-ions are now no-op if the given system packages set is empty.  [#6489 @arozovyk]
   * `OpamGlobalState`: add `is_root_read_only` to check if we are in sandboxed environment [#6489 @rjbou]
   * `OpamSwitchState`: add `update_sys_packages` to update depexts status of a set of packages. [#6489 @arozovyk]
+  * `OpamSwitchState`: add `is_source_dir_temporary` [#6440 @kit-ty-kate]
   * `OpamSysInteract`: add `available_packages` and `installed_packages` to be computed separately, redefine `packages_status` accordingly [#6489 @arozovyk]
   * `OpamStateTypes`: add available system package status field `repos_syspkgs_available` (and its type `repo_syspkgs_available`) in `repos_state` for all the depexts declared in repo's packages. The new field is also added to the cache. [#6489 @arozovyk @rjbou]
   * `OpamRepositoryState.load`: load repo's available system packages [#6489 @arozovyk]
--- a/src/client/opamSolution.ml
+++ b/src/client/opamSolution.ml
@@ -661,8 +661,12 @@ let parallel_apply t
       in
       let source_dir = source_dir nv in
       (if OpamFilename.exists_dir source_dir
-       then (if not is_inplace then
-               OpamFilename.copy_dir ~src:source_dir ~dst:build_dir)
+       then
+         (if not is_inplace then
+            (if OpamSwitchState.is_source_dir_temporary t nv then
+               OpamFilename.move_dir ~src:source_dir ~dst:build_dir
+             else
+               OpamFilename.copy_dir ~src:source_dir ~dst:build_dir))
        else OpamFilename.mkdir build_dir;
        OpamAction.prepare_package_source t nv build_dir @@+ function
        | Some exn -> store_time (); Done (`Exception exn)
@@ -702,7 +706,11 @@ let parallel_apply t
          OpamFilename.rmdir d;
          let source_dir = source_dir nv in
          if OpamFilename.exists_dir source_dir
-         then OpamFilename.copy_dir ~src:source_dir ~dst:d
+         then
+           (if OpamSwitchState.is_source_dir_temporary t nv then
+              OpamFilename.move_dir ~src:source_dir ~dst:d
+            else
+              OpamFilename.copy_dir ~src:source_dir ~dst:d)
          else OpamFilename.mkdir d;
          OpamAction.prepare_package_source t nv d
        else Done None) @@+ fun _ ->
--- a/src/state/opamSwitchState.ml
+++ b/src/state/opamSwitchState.ml
@@ -808,6 +808,9 @@ let source_dir st nv =
   then OpamPath.Switch.pinned_package st.switch_global.root st.switch nv.name
   else OpamPath.Switch.sources st.switch_global.root st.switch nv
 
+let is_source_dir_temporary st nv =
+  not (OpamPackage.Set.mem nv st.pinned || is_dev_package st nv)
+
 let overlay_opam_file st name =
   if is_pinned st name then
     let file =
--- a/src/state/opamSwitchState.mli
+++ b/src/state/opamSwitchState.mli
@@ -152,8 +152,14 @@ val dev_packages: 'a switch_state -> package_set
 
 (** Returns the local source mirror for the given package
     ({!OpamPath.Switch.sources} or {!OpamPath.Switch.pinned_package}, depending
-    on wether it's pinned). *)
+    on whether it's pinned. If it's not pinned, be mindful that the directory
+    is temporary (see {!is_source_dir_temporary}). *)
 val source_dir: 'a switch_state -> package -> dirname
+
+(** Returns [true] if the {!source_dir} of the package is meant to be a
+    temporary directory: packages that are not pinned or "dev package"
+    as defined in {!is_dev_package}. *)
+val is_source_dir_temporary: 'a switch_state -> package -> bool
 
 (** Returns the opam file overlay stored internally for pinned packages *)
 val overlay_opam_file:
--- a/tests/reftests/sources-directory-cleaning.test
+++ b/tests/reftests/sources-directory-cleaning.test
@@ -49,7 +49,6 @@ The following actions will be performed:
 -> installed waste.1
 Done.
 ### ls OPAM/sources-dir/.opam-switch/sources/
-waste.1
 ### : at removal
 ### opam remove waste
 The following actions will be performed:
@@ -71,7 +70,6 @@ The following actions will be performed:
 -> installed waste.1
 Done.
 ### ls OPAM/sources-dir/.opam-switch/sources
-waste.1
 ### opam reinstall waste
 The following actions will be performed:
 === recompile 1 package
@@ -83,7 +81,6 @@ The following actions will be performed:
 -> installed waste.1
 Done.
 ### ls OPAM/sources-dir/.opam-switch/sources
-waste.1
 ### : at atomic reinstall
 ### opam remove waste
 The following actions will be performed:
@@ -119,7 +116,6 @@ The following actions will be performed:
 -> installed waste.1
 Done.
 ### ls OPAM/sources-dir/.opam-switch/sources
-waste.1
 ### opam remove waste
 The following actions will be performed:
 === remove 1 package
@@ -156,7 +152,6 @@ The following actions will be performed:
 -> installed waste.1
 Done.
 ### ls OPAM/sources-dir/.opam-switch/sources
-waste.1
 ### opam remove waste
 The following actions will be performed:
 === remove 1 package
--- a/tests/reftests/action-disk.test
+++ b/tests/reftests/action-disk.test
@@ -234,9 +234,7 @@ SYSTEM                          rmdir ${OPAMTMP}
 SYSTEM                          rm ${OPAMTMP}/x-source
 -> retrieved main-repo.1  (file://${BASEDIR}/archive.tgz)
 SYSTEM                          rmdir ${BASEDIR}/OPAM/install-from-repo/.opam-switch/build/main-repo.1
-SYSTEM                          copydir ${BASEDIR}/OPAM/install-from-repo/.opam-switch/sources/main-repo.1 -> ${BASEDIR}/OPAM/install-from-repo/.opam-switch/build/main-repo.1
-SYSTEM                          mkdir ${BASEDIR}/OPAM/install-from-repo/.opam-switch/build/main-repo.1
-SYSTEM                          copy ${BASEDIR}/OPAM/install-from-repo/.opam-switch/sources/main-repo.1/content -> ${BASEDIR}/OPAM/install-from-repo/.opam-switch/build/main-repo.1/content
+SYSTEM                          mv ${BASEDIR}/OPAM/install-from-repo/.opam-switch/sources/main-repo.1 -> ${BASEDIR}/OPAM/install-from-repo/.opam-switch/build/main-repo.1
 SYSTEM                          copy ${BASEDIR}/OPAM/download-cache/md5/pre/+xs-hash+ -> ${BASEDIR}/OPAM/install-from-repo/.opam-switch/build/main-repo.1/x-source.main-repo.1
 SYSTEM                          read ${BASEDIR}/OPAM/repo/default/packages/main-repo/main-repo.1/files/x-file.main-repo.1
 SYSTEM                          write ${BASEDIR}/OPAM/install-from-repo/.opam-switch/build/main-repo.1/x-file.main-repo.1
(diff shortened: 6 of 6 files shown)


kit-ty-kate:
As a demonstration, `opam clean --switch-cleanup --all` which primarily cleans up these unnecessary directories, just removed 20GB of disk space on my main machine.

kit-ty-kate:
xref live test results: https://github.com/ocaml/opam/pull/6906#issuecomment-4306603679

src/state/opamSwitchAction.ml
```diff
@@ -66,6 +66,7 @@ let create_empty_switch gt ?synopsis ?repos ?invariant switch =
     OpamFilename.mkdir (OpamPath.Switch.lib_dir root switch config);
     OpamFilename.mkdir (OpamPath.Switch.stublibs root switch config);
     OpamFilename.mkdir (OpamPath.Switch.toplevel root switch config);
+    OpamFilename.mkdir (OpamPath.Switch.sources_dir root switch);
```
rjbou: Why do we need to always create that directory ?

rjbou (changes requested):
On the idea, lgtm, it is a good enhancement for disk usage saving. On the implementation, i don't think it should be done that way. The `source_dir` function should keep returning the sources directory, and when we want to bypass that source directory to extract directly on the build dir, it should be done directly and explicitly.

src/state/opamSwitchAction.ml
kit-ty-kate: mainly for one of the tests. I've replaced that by a commit adding the necessary mkdir

rjbou: There is no need to change opam behaviour to comply a test. The test can be changed to no longer run an `ls` but another command that don't fail if the directory is not found. At the time, I used `ls` because `sources` directory is always created by opam when fetching sources. Another solution is to create the directory by hand `### mkdir OPAM/sources-dir/.opam-switch/sources/ ` at the beginning of the test. It doesn't change what is tested : we want to highlight what source directory is still present after the command.

src/client/opamAction.ml
```diff
@@ -269,8 +269,12 @@ let download_shared_source st url nvs =
   let nvs =
     (* filter out version-pinned packages since we already have their source *)
     List.filter (fun nv ->
-        let dir = OpamSwitchState.source_dir st nv in
         not (OpamPackage.Set.mem nv st.pinned &&
+             let dir = match OpamSwitchState.source_dir st nv with
+               | None ->
+                 assert false (* impossible given match on st.pinned above *)
```
rjbou: I liked the previous version that used directly `OpamPath.x` when needed. It avoided the `assert false`.

src/client/opamSolution.ml
```diff
@@ -645,21 +645,22 @@ let parallel_apply t
...
       let test, doc, dev_setup =
         let found = OpamPackage.Set.mem nv requested in
         OpamStateConfig.(!r.build_test) && found,
         OpamStateConfig.(!r.build_doc) && found,
         OpamStateConfig.(!r.dev_setup) && found
       in
-      let source_dir = source_dir nv in
       (if OpamFilename.exists_dir source_dir
        then (if not is_inplace then
                OpamFilename.copy_dir ~src:source_dir ~dst:build_dir)
```
rjbou: By not differentiating the cases, we end up with src = dst = build_dir if the package is not pinned

src/state/opamSwitchAction.ml
rjbou: I didn't see, you already made the mkdir

kit-ty-kate: >  Another solution is to create the directory by hand ### mkdir OPAM/sources-dir/.opam-switch/sources/ 

sorry if i wasn't clear, that's what i meant by: `I've replaced that by a commit adding the necessary mkdir` (see new first commit from 2 hours ago)

tests/reftests/sources-directory-cleaning.test
```diff
@@ -48,8 +48,9 @@ The following actions will be performed:
 -> retrieved waste.1  (file://${BASEDIR}/waste.tgz)
 -> installed waste.1
 Done.
+### : mkdir in case the sources directory hasn't been created
+### mkdir OPAM/sources-dir/.opam-switch/sources/
```
rjbou:
```suggestion

### : mkdir in case the sources directory hasn't been created
### mkdir OPAM/sources-dir/.opam-switch/sources/
### opam switch create sources-dir --empty
### : at install
### opam install waste
The following actions will be performed:
=== install 1 package
- install waste 1

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
-> retrieved waste.1  (file://${BASEDIR}/waste.tgz)
-> installed waste.1
Done.
```

src/state/opamSwitchAction.ml
rjbou: indeed, i misread, you was clear enough

src/state/opamSwitchState.mli
```diff
@@ -152,8 +152,12 @@ val dev_packages: 'a switch_state -> package_set
 
 (** Returns the local source mirror for the given package
     ({!OpamPath.Switch.sources} or {!OpamPath.Switch.pinned_package}, depending
-    on wether it's pinned). *)
-val source_dir: 'a switch_state -> package -> dirname
+    on wether it is pinned or simply a dev package).
+    Returns [None] if it is neither pinned or a dev package. *)
```
rjbou:
```suggestion
    Returns [None] if it is neither pinned or a dev package as defined
    in {!is_dev_package}.  *)
```

src/state/opamSwitchState.mli
```diff
@@ -152,8 +152,12 @@ val dev_packages: 'a switch_state -> package_set
...
+val source_dir: 'a switch_state -> package -> dirname option
+
+(** Returns [source_dir] or [build_dir] otherwise. *)
```
rjbou:
```suggestion
(** Returns {!source_dir} or {!OpamPath.Switch.build_dir} otherwise. *)
```

src/state/opamUpdate.ml
```diff
@@ -707,7 +711,9 @@ let download_package_source_t st url nv_dirs =
 
 let download_shared_package_source st url nvs =
```
rjbou, line 708: The documentation should be updated in the mli to highlight the new directory destinaation

tests/reftests/dot-install.test
```diff
@@ -128,7 +128,7 @@ ACTION                          prepare_package_source: dot.~dev at ${BASEDIR}/O
 ACTION                          Removing dot.~dev
 -> removed   dot.~dev
 Installing dot.~dev.
-TRACK                           after install: 0 elements, 0 added, scanned in 0.000s
+TRACK                           after install: 3 elements, 0 added, scanned in 0.000s
```
rjbou: How does it affects dirtrack added ? 
With full debug, the diff with master is
```diff
-@@ -120,17 +122,106 @@ added: [
+@@ -120,17 +122,103 @@ added: [
  ### unset OPAMDEBUGSECTIONS
  ### OPAMDEBUG=-5
  ### opam reinstall dot --dry-run
@@ -94,10 +94,6 @@
  ACTION                          download_package: dot.~dev
 PARALLEL                        Job 62929499 finished
 PARALLEL                        Starting job 798933370 (worker 1/15 -/1 -/1): build dot.~dev
-SYSTEM                          rmdir ${BASEDIR}/OPAM/inst/.opam-switch/build/dot.~dev
-SYSTEM                          rm ${BASEDIR}/OPAM/inst/.opam-switch/build/dot.~dev/dot.install
-SYSTEM                          rm ${BASEDIR}/OPAM/inst/.opam-switch/build/dot.~dev/a-file
-SYSTEM                          mkdir ${BASEDIR}/OPAM/inst/.opam-switch/build/dot.~dev
  ACTION                          prepare_package_source: dot.~dev at ${BASEDIR}/OPAM/inst/.opam-switch/build/dot.~dev
 PARALLEL                        Job 798933370 finished
 PARALLEL                        Starting job 743026333 (worker -/15 1/1 -/1): remove dot.~dev
@@ -107,11 +103,12 @@
  -> removed   dot.~dev
 PARALLEL                        Job 743026333 finished
 PARALLEL                        Starting job 699714656 (worker -/15 1/1 -/1): install dot.~dev
-FILE(.install)                  Cannot find ${BASEDIR}/OPAM/inst/.opam-switch/build/dot.~dev/dot.install
-TRACK                           before install: 0 elements scanned in 0.000s
+FILE(.install)                  Read ${BASEDIR}/OPAM/inst/.opam-switch/build/dot.~dev/dot.install in 0.000s
+FILE(.install)                  Wrote ${BASEDIR}/OPAM/inst/.opam-switch/install/dot.install atomically in 0.000s
+TRACK                           before install: 3 elements scanned in 0.000s
  Installing dot.~dev.
-[ERROR] tracked after {}
- TRACK                           after install: 0 elements, 0 added, scanned in 0.000s
+[ERROR] tracked after { share/dot/a-file:F:12fc204edeae5b57713c5ad7dcb97d39, share/dot:D, share:D }
+ TRACK                           after install: 3 elements, 0 added, scanned in 0.000s
  -> installed dot.~dev
 FILE(.config)                   Cannot find ${BASEDIR}/OPAM/inst/.opam-switch/config/dot.config
 PARALLEL                        Job 699714656 finished

File: src/client/opamAction.ml (hunk 2 above)
Comment (kit-ty-kate) [reply src/client/opamAction.ml]: done in the latest fixup commit

File: src/state/opamUpdate.ml (hunk 7 above)
Comment (kit-ty-kate) [reply src/state/opamUpdate.ml:708]: not relevant anymore

**Comment by kit-ty-kate** [comment]
Thanks for the review. Turns out part of design had bad assumptions and extracting directly into `build` broke a lot of invariants that wasn't in the path that i tested for.

I redid the design and now the testsuite look good. The change is that we now use the `sources` directory as temporary directory that will be moved (they're pretty much guaranteed to be on the same partition so the cost of moving is none) to the `build` directory.

Depending on your personal preference i've pushed two commits. The first one changes the type of `OpamSwitchState.source_dir` to return whether or not it is meant to be permanent. However this value isn't used most of the time and costs an extra call to `OpamSwitchState.is_dev_package` so i also pushed a second commit to put back the old return type but add an extra `is_source_dir_temporary` for when the information is really necessary.

So depending on your preference, feel free to squash the second one or remove it, i'm happy with both solutions.

File: src/state/opamSwitchState.ml [hunk 9]
```diff
@@ -808,6 +808,11 @@ let source_dir st nv =
   then OpamPath.Switch.pinned_package st.switch_global.root st.switch nv.name
   else OpamPath.Switch.sources st.switch_global.root st.switch nv
 
+let is_source_dir_temporary st nv =
+  if OpamPackage.Set.mem nv st.pinned then false
+  else if is_dev_package st nv then false
+  else true
```
rjbou:
```suggestion
  not (OpamPackage.Set.mem nv st.pinned || is_dev_package)
```
??

src/state/opamSwitchState.mli
```diff
@@ -152,9 +152,14 @@ val dev_packages: 'a switch_state -> package_set
 
 (** Returns the local source mirror for the given package
     ({!OpamPath.Switch.sources} or {!OpamPath.Switch.pinned_package}, depending
-    on wether it's pinned). *)
+    on whether it's pinned. If it's not pinned, be mindful that the directory
+    is temporary (see {!is_source_dir_temporary}). *)
 val source_dir: 'a switch_state -> package -> dirname
 
+(** Returns [true] if the {!source_dir} of the package is meant to be a
+    temporary directory. *)
```
rjbou:
```suggestion
    temporary directory: packages that are not pinned or "dev package"
    as defined in {!is_dev_package}.  *)
```

src/client/opamCommands.ml
```diff
@@ -4355,7 +4355,7 @@ let clean cli =
                (OpamPackage.Set.union st.pinned
                   (OpamPackage.Set.filter (OpamSwitchState.is_dev_package st)
                      st.installed))
-             |> List.map (OpamSwitchState.source_dir st)
+             |> List.map (fun nv -> OpamSwitchState.source_dir st nv)
```
rjbou: This change is no longer needed in the last version

```suggestion
             |> List.map (OpamSwitchState.source_dir st)
```

rjbou (approved):
I like the second solution, it's more clear, and we don't have extra code (extra `snd`).

master_changes.md
```diff
@@ -31,6 +31,7 @@ users)
 
 ## Actions
   * Reorder the list of actions by increased priority [#6864 @kit-ty-kate - fix #6863]
+  * The internal `sources` directory now only serves to cache dev packages instead of every packages, saving disk space [#6440 @kit-ty-kate - fix #4056 #5448]
```
```suggestion
  * The internal `sources` directory now only serves to cache dev packages instead of every packages, saving disk space. It is removed after install. [#6440 @kit-ty-kate - fix #4056 #5448]
```

kit-ty-kate:
Ignoring the macOS failure (#7037)

kit-ty-kate:
Noteworthy: the `opam install lot-io-source lot-io-source2` benchmark show a 24% speedup :tada:

PR #6464, opened 2025-04-07
Disable the detection of available system packages on SUSE-based distributions

kit-ty-kate:
Split off of https://github.com/ocaml/opam/pull/6431
See https://github.com/ocaml/opam/pull/6431#issue-2944043215 for explanation


rjbou (approved):
LGTM on the idea. From discussion, `rpm` is always present on SUSE-based distros, so the change should be safe.
Good to merge once the CI is green (it didn't, i don't know why)

PR #6467, opened 2025-04-08
Only download Cygwin's setup.exe when the command is actually going to be displayed or used

kit-ty-kate:
opam only look at `os-distribution` to know if it should download Cygwin's `setup.exe` or not:
https://github.com/ocaml/opam/blob/1458aa74921cd71b8a0f0182de308bd06e7146a6/src/client/opamSolution.ml#L1222
https://github.com/ocaml/opam/blob/1458aa74921cd71b8a0f0182de308bd06e7146a6/src/client/opamSolution.ml#L1254

The tests have been failing quite frequently over the past two days due to some network issues with `cygwin.com`. Setting `os-distribution` to not be `cygwin` should alleviate this.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -179,6 +179,7 @@ users)
   * Use a C stub to call the `uname` function from the C standard library instead of calling the `uname` POSIX command [#6217 @kit-ty-kate]
 
 ## Internal: Windows
+  * Only download Cygwin's `setup.exe` when the command is actually going to be displayed or used [#6467 @kit-ty-kate]
 
 ## Test
   * Add a library test for the new OCaml implementation of patch and diff [#5892 @rjbou]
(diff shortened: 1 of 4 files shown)


rjbou:
I think its better to move those check directly in `OpamSysInteract.install`, like that it is checked (and downloaded) only when really needed. 
https://github.com/ocaml/opam/blob/1458aa74921cd71b8a0f0182de308bd06e7146a6/src/state/opamSysInteract.ml#L1047-L1051

kit-ty-kate:
This is orthogonal to this PR though. Could we first deal with the CI failures and improve internals later? If you're fine with that please feel free to open a ticket so we can remember to do it

rjbou (approved):
Thanks!

PR #6468, opened 2025-04-09
depext: enforce dry run for non query commands

rjbou:


src/state/opamSysInteract.ml
```diff
@@ -73,15 +74,15 @@ let run_command
   OpamProcess.cleanup r;
   Done (code, out)
 
-let run_query_command ?vars cmd args =
+let run_query_command ?(nonzero=false) ?discard_err ?vars cmd args =
   let vars = (`set, ("LC_ALL","C"))::OpamStd.Option.to_list vars in
-  let code,out = run_command ~vars cmd args in
-  if code = 0 then out
+  let code,out = run_command ?discard_err ~vars cmd args ~dryrun:false in
+  if nonzero || code = 0 then out
```
kit-ty-kate, line 80: i don't understand why nonzero exit statuses should be allowed

src/state/opamSysInteract.ml
```diff
@@ -14,7 +14,8 @@ let log fmt = OpamConsole.log "XSYS" fmt
 (* Always call this function to run a command, as it handles `dryrun` option *)
 
 let run_command
-    ?vars ?(discard_err=false) ?allow_stdin ?verbose ?(dryrun=false) cmd args =
+    ?vars ?(discard_err=false) ?allow_stdin ?verbose
+    ?(dryrun=OpamStateConfig.(!r.dryrun)) cmd args =
   let clean_output =
     if not discard_err then
```
kit-ty-kate, line 20: it's very unclear what `discard_err` does

kit-ty-kate (changes requested):
Moving to the 2.7 milestone. I think this requires more discussion that we don't have time for atm.

PR #6481, opened 2025-04-22
Remove `ignore` use and some unnecessary open

rjbou:
Use instead `let _ : <typ> =` to ensure that no argument is missing.

### DIFF
--- a/src/solver/opamCudf.ml
+++ b/src/solver/opamCudf.ml
@@ -1342,7 +1342,6 @@ let dump_cudf_request ~version_map (_, univ,_ as cudf) criteria =
   function
   | None   -> None
   | Some f ->
-    ignore ( version_map: int OpamPackage.Map.t );
     incr solver_calls;
     let filename = Printf.sprintf "%s-%d.cudf" f !solver_calls in
     let oc = open_out filename in
@@ -1554,8 +1553,10 @@ let call_external_solver ~version_map univ req =
   if Cudf.universe_size univ > 0 then
     let criteria = OpamSolverConfig.criteria req.criteria in
     let chrono = OpamConsole.timer () in
-    ignore (dump_cudf_request ~version_map cudf_request
-              criteria OpamSolverConfig.(!r.cudf_file));
+    let _ : string option =
+      dump_cudf_request ~version_map cudf_request
+        criteria OpamSolverConfig.(!r.cudf_file)
+    in
     (* Wrap a return of exn Timeout through Depsolver *)
     let check_request_using ~call_solver ~explain req =
       let timed_out = ref false in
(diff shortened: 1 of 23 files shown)


src/solver/opamCudf.ml
```diff
@@ -1554,8 +1553,10 @@ let call_external_solver ~version_map univ req =
   if Cudf.universe_size univ > 0 then
     let criteria = OpamSolverConfig.criteria req.criteria in
     let chrono = OpamConsole.timer () in
-    ignore (dump_cudf_request ~version_map cudf_request
-              criteria OpamSolverConfig.(!r.cudf_file));
+    let _ : string option =
+      (dump_cudf_request ~version_map cudf_request
+         criteria OpamSolverConfig.(!r.cudf_file))
+    in
```
kit-ty-kate, line 1559:
```suggestion
      dump_cudf_request ~version_map cudf_request
        criteria OpamSolverConfig.(!r.cudf_file)
    in
```

src/state/opamFormatUpgrade.ml
```diff
@@ -853,13 +853,14 @@ let from_2_0_alpha_to_2_0_alpha2 ~on_the_fly:_ root conf =
...
               OpamFilename.move_dir ~src:(pkgdir nv) ~dst:(pkgdir new_nv);
-            OpamStd.Option.Op.(
-              OpamFilename.opt_file (pkgdir new_nv // "opam") >>|
-              OpamFile.make >>= fun f ->
-              OpamFile.OPAM.read_opt f >>|
-              opam_file_from_1_2_to_2_0 ~filename:f >>|
-              OpamFile.OPAM.write_with_preserved_format f
-            ) |> ignore;
+            let _ : unit option =
+              OpamStd.Option.Op.(
```
kit-ty-kate, line 857: To avoid the indentation change we could do:
```suggestion
              let open OpamStd.Option.Op in
```

kit-ty-kate (approved):
Apart from a couple of optional suggestions and a personal opinion that the title/label doesn't quite fit (for me, API means we're changing the interface, not just doing a cleanup of the internal implementation), LGTM.

src/solver/opamCudf.ml
```diff
@@ -1342,7 +1342,6 @@ let dump_cudf_request ~version_map (_, univ,_ as cudf) criteria =
   function
   | None   -> None
   | Some f ->
-    ignore ( version_map: int OpamPackage.Map.t );
```
rjbou, line 1345: reintroduce ?

kit-ty-kate: version_map is used bellow so i don't think this line makes sense

kit-ty-kate (approved):
Rebased and ensured complete using `ocamlgrep Stdlib.ignore`.

I've removed https://github.com/ocaml/opam/pull/6481/commits/e882a1896adf684a8c1e8e5545ddedc066927ff1 as i think this should be in its own dedicated PR (https://github.com/ocaml/opam/pull/7034)

kit-ty-kate:
Thanks a lot!

Issue #6474, opened 2025-04-16
Opam fails to download cygwin shasum


ajbt200128:
We've been getting dozens of these errors on our windows builds:

```
OPAMSOLVERTIMEOUT=1500 LWT_DISCOVER_ARGUMENTS="--use-libev true" LIBRARY_PATH="/lib:" opam install --confirm-level=unsafe-yes -y --depext-only ./ ./dev/required.opam 
[WARNING] Failed checks on required package definition from source at file://D:/a/semgrep-proprietary/semgrep-proprietary/OSS/dev:
  warning 68: Missing field 'license'
[WARNING] Opam packages conf-mingw-w64-curl-x86_64.1, conf-mingw-w64-gmp-x86_64.1, conf-mingw-w64-pcre-x86_64.1 and conf-mingw-w64-pcre2-x86_64.1 depend on the following system packages that are no longer installed: mingw64-x86_64-curl mingw64-x86_64-gmp mingw64-x86_64-pcre mingw64-x86_64-pcre2

The following system packages will first need to be installed:
    mingw64-x86_64-curl mingw64-x86_64-gmp mingw64-x86_64-pcre mingw64-x86_64-pcre2

<><> Handling external dependencies <><><><><><><><><><><><><><><><><><><><><><>

Fatal error:
OpamDownload.Download_fail(_, "curl failed: \"D:\\\\cygwin\\\\bin\\\\curl.exe --write-out %{http_code}\\\\n --retry 3 --retry-delay 2 --user-agent opam/2.3.0 -L -o C:\\\\Users\\\\runneradmin\\\\AppData\\\\Local\\\\Temp\\\\opam-2896-f71c65\\\\sha512.sum.part -- [https://cygwin.com/sha512.sum\](https://cygwin.com/sha512.sum/)" exited with code 16")
make[1]: *** [Makefile:281: install-opam-deps] Error 99
make[1]: Leaving directory '/cygdrive/d/a/semgrep-proprietary/semgrep-proprietary/OSS'
make: *** [Makefile:255: install-deps-for-semgrep-core] Error 2
make: Leaving directory '/cygdrive/d/a/semgrep-proprietary/semgrep-proprietary/OSS'
```

It seems like it may be an upstream issue with cygwin (or maybe libcurl?), but it would be awesome if opam retried this step a few times at the least.

kit-ty-kate:
[...]

PR #6482, opened 2025-04-23
Cygwin: Fallback to the existing setup-x86_64.exe if its upgrade failed to be fetched

### LINKED ISSUE #6495: `setup.exe`: download failure error instead of retrieving the local exe
When checking for cygwin install, we try to download last version of `setup.exe`. If there is an error while fetching `setup.exe` or its sha, opam fails, even if there is an already present `setup.exe` stored internally.
We should add a fallback to use that `setup.exe` instead of failing.

kit-ty-kate:
Partially f.i.x.e.s https://github.com/ocaml/opam/issues/6474
Fix #6495

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -63,6 +63,7 @@ users)
 ## Opamfile
 
 ## External dependencies
+  * Cygwin: Fallback to the existing `setup-x86_64.exe` if its upgrade failed to be fetched [#6482 @kit-ty-kate - fix #6495, partial fix #6474]
 
 ## Format upgrade
 
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -262,7 +262,7 @@ module Cygwin = struct
     let kind = `SHA512 in
     let dst_exists = OpamFilename.exists dst in
     let current_checksum =
-      if OpamFilename.exists dst then
+      if dst_exists then
         Some (OpamHash.compute ~kind (OpamFilename.to_string dst))
       else
         None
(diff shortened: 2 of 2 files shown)


master_changes.md
```diff
@@ -63,6 +63,7 @@ users)
 ## Opamfile
 
 ## External dependencies
+  * Cygwin: Fallback to the existing `setup-x86_64.exe` if its upgrade failed to be fetched [#6482 @kit-ty-kate - partial fix #6474]
```
rjbou, line 66:
```suggestion
  * Cygwin: Fallback to the existing `setup-x86_64.exe` if its upgrade failed to be fetched [#6482 @kit-ty-kate - fix #6495, partial fix #6474]
```

src/state/opamSysInteract.ml
```diff
@@ -460,8 +460,14 @@ module Cygwin = struct
   (* Set setup.exe in the good place, ie in .opam/.cygwin/ *)
   let check_setup ~update =
     let dst = cygsetup () in
-    if update || not (OpamFilename.exists dst) then
-      OpamProcess.Job.run @@ download_setupexe dst
+    let download () = download_setupexe dst in
```
rjbou, line 463: The fallback is added in `check_setup` and not in `donwload_setup_exe`.
There is another use of `donwload_setup_exe` in Cygwin internal `install` function. It is mainly called via an init, but it can be also called via a reinit. That one could fail too.

master_changes.md
```diff
@@ -63,6 +63,7 @@ users)
 ## Opamfile
 
 ## External dependencies
+  * Cygwin: Fallback to the existing `setup-x86_64.exe` if its upgrade failed to be fetched [#6482 @kit-ty-kate - fix #6495, partial fix #6474]
```
rjbou, line 66:
```suggestion
  * Cygwin: Fallback to the existing `setup-x86_64.exe` if its upgrade failed to be fetched (sha & exe) [#6482 @kit-ty-kate - fix #6495, partial fix #6474]
```

rjbou (approved):
Some suggestion, but absolutely not mandatory.
Thanks!

Issue #6484, opened 2025-04-24
BUG/memory usage: OpamConsole.log and OpamFile.Stats


hannesm:
Dear everyone,

thanks for your time to read this bug report.

So, I have an application (https://github.com/hannesm/maintenance-intent-filter) using opam as a library - and discovered there are some heavy memory usages.

The root cause are two things:
- (a) https://github.com/ocaml/opam/commit/84a3d28ebe374abdb7f03cf068850ef4068f9d98 (part of https://github.com/ocaml/opam/pull/4316) which "queues up OpamConsole messages if OpamCoreConfig.set has not been called". Now, in my application there's nothing calling OpamCoreConfig.set -- so these messages are queued up forever.
- (b) OpamFile.Stats.read_files (a list of strings) getting extended in OpamFile.MakeIO().read_opt - and never gets cleared (nor uniquified)

I was wondering whether for (a) a bounded queue could be used? And for (b), could that be only added if r.print_stats is true -- otherwise these mutable lists are never used. Or could that code be removed entirely, if not used?

FWIW, in Stats.print there's an error:
```OCaml
  let print () =
    let aux kind = function
      | [] -> ()
      | l  ->
        OpamConsole.msg "%d files %s:\n  %s\n"
          (List.length !read_files) kind (String.concat "\n  " l)
    in
    aux "read" !read_files;
    aux "write" !write_files
```

The `List.length` uses always the `!read_files` - while it should use `l`.

If you can decide on acceptable ways forward for both (a) and (b), that would be fantastic.

PR #6485, opened 2025-04-24
Fix a memory leak by removing OPAMSTATS

hannesm:
especially in long-running applications that use opam as a library reading and/or writing lots of opam files, this lead to a huge amount of memory usage for no obvious gain.

Partially addresses #6484

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -13,6 +13,7 @@ users)
 ## Version
 
 ## Global CLI
+  * Remove handling of the `OPAMSTATS` environment variable [#6485 @hannesm]
 
 ## Plugins
 
@@ -94,6 +95,7 @@ users)
 ## Shell
 
 ## Internal
+  * Fix a memory leak happening when running large numbers of commands or opening large number of opam files [#6485 @hannesm - fix #6484]
 
 ## Internal: Unix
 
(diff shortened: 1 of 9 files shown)


hannesm:
update: there was the environment variable OPAMSTATS, but I've not found anything on github that set it. of course, your backwards compatibility may differ -- but I find this specific form of statistics not very useful.

kit-ty-kate:
Personally, i'm fine with removing `OPAMSTATS` as it seems redundant with `--debug-level=3`/`OPAMDEBUG`, which already shows when files are read/written and which commands have been ran, with more useful information (the context where these things happen). LGTM

rjbou (approved):
I've added changelog API and mark `OPAMSTATS` env variable as removed instead of removing it completely.

rjbou:
`OPAMSTATS` is indeed a legacy from initial opam implementation (0.9.0, cf e374647d26fb4b654e47de5497a2cbeedd6b7bd0). As already noted, now debug information gives more information that those available via `OPAMSTATS`.
Discussed in dev meeting, it is indeed good to remove, especially given the memory usage and that now with `patch` library we will read much more files, this number will increase.

rjbou:
Thank your for the PR @hannesm!

kit-ty-kate (approved):

PR #6487, opened 2025-04-24
Do not keep log messages before initialization if the code is ran through a library

kit-ty-kate:
Alternative to https://github.com/ocaml/opam/pull/6486

### DIFF
--- a/src/core/opamCoreConfig.ml
+++ b/src/core/opamCoreConfig.ml
@@ -67,6 +67,7 @@ type t = {
   cygbin: string option;
   git_location: string option;
   set: bool;
+  in_opam: bool;
 }
 
 type 'a options_fun =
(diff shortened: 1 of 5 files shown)


hannesm:
I'm fine with this as well (and closing #6486). I'm a bit nervous adding even more mutable state to the codebase, but as said, it looks fine :)

rjbou (approved):

src/core/opamCoreConfig.ml
```diff
@@ -212,3 +215,5 @@ let answer_is_yes () =
   | _ -> false
 
 let developer = not (String.equal OpamCoreConfigDeveloper.value "")
+
+let set_in_opam () = r := {default with in_opam = true}
```
kit-ty-kate: To add upon rebase
```suggestion
let set_in_opam () = r := {!r with in_opam = true}
```

Issue #6461, opened 2025-04-05
Move the depexts availability calculation at opam update time


kit-ty-kate:
Currently depexts availability is calculated on demand on the platforms where the command to do it, do it in a reasonable amount of time (e.g. see #6426 #4548 and #5372).

Proposal: to allow most (all?) supported platforms to finally get on the same level in terms of depexts feature, we could instead move the availability calculation during `opam update`, during repository cache reload to be more precise. Doing it during cache reload instead of unconditionally during `opam update` would allow users to actually experience #6283 whenever we merge this.

The reasoning behind this is that the set of available system packages don't really change all that often (new packages are slowly added and removed) and such change would be tied to a change in the repository if used anyway. Also use of depexts in vcs/local pins is extremely rare.

The installed depexts calculation is usually much faster and easier to do so this can stay in the same place.

Doing so should allow all platforms to finally get on par with each other, in particular regarding the virtual packages support (see https://github.com/ocaml/opam/issues/4759)

PR #6489, opened 2025-04-25
Do depexts status check at opam update time

arozovyk:
This is a Draft for #6461 

- Added `sys_pkg_statues` to repository state and cache
- Removed `Lazy.t` from `switch_state.sys_packages` 
- Added a lookup into repository state to be used on switch state load 

This results in a significant speedup by avoiding calling `OpamSysInteract.packages_status` each time:  
```py
00:02.330  STATE                  depexts loaded in 1.865s
```

For instance `./opam install conf-clang-format.1` on my machine seems to go from 
```py
 5.06s user 0.38s system 83% cpu 6.495 total
 5.05s user 0.41s system 67% cpu 8.028 total
 5.05s user 0.39s system 87% cpu 6.203 total
```
to 
```py 
 4.08s user 0.15s system 84% cpu 4.987 total
 4.06s user 0.15s system 81% cpu 5.149 total
 4.09s user 0.15s system 89% cpu 4.736 total
```
In both cases (sys call and rt retrieval) the depext availability for the package `clang-format` is checked.

This breaks `depexts`and `pin` tests,  as the mechanism of this PR is not really compatible with the tests as they are (the use of `N0REP0`?)

Will be exploring it further.

### DIFF
--- a/src/client/opamSolution.ml
+++ b/src/client/opamSolution.ml
@@ -379,7 +379,7 @@ let parallel_apply t
       (* Turns out these depexts weren't needed after all. Remember that and
          make the bypass permanent. *)
       try
-        (OpamPackage.Map.find nv (Lazy.force !t_ref.sys_packages)).s_available
+        (OpamPackage.Map.find nv (!t_ref.sys_packages)).s_available
       with Not_found -> OpamSysPkg.Set.empty
     in
     let bypass = OpamSysPkg.Set.union missing_depexts !bypass_ref in
@@ -1152,7 +1152,7 @@ let get_depexts ?(force=false) ?(recover=false) t
       if recover then
         OpamSwitchState.depexts_status_of_packages t pkg_to_install
       else
-        let base = Lazy.force t.sys_packages in
+        let base = t.sys_packages in
         (* workaround: st.sys_packages is not always updated with added
            packages *)
         let more_pkgs =
@@ -1337,9 +1337,9 @@ let install_depexts ?(force_depext=false) ?(confirm=true) t
               sys_map
           | None -> sys_map)
         pkg_to_install
-        (Lazy.force t.sys_packages)
+        (t.sys_packages)
     in
-    { t with sys_packages = lazy sys_packages }
+    { t with sys_packages = sys_packages }
   in
   let confirm =
     confirm && not (OpamSysInteract.Cygwin.is_internal t.switch_global.config)
--- a/src/state/opamRepositoryState.ml
+++ b/src/state/opamRepositoryState.ml
@@ -20,6 +20,7 @@ module Cache = struct
   type t = {
     cached_repofiles: (repository_name * OpamFile.Repo.t) list;
     cached_opams: (repository_name * OpamFile.OPAM.t OpamPackage.Map.t) list;
+    cached_sys_pkg_statues: (repository_name * sys_pkg_status) list;
   }
 
   module C = OpamCached.Make (struct
@@ -54,6 +55,8 @@ module Cache = struct
         cached_opams =
           OpamRepositoryName.Map.bindings
             (filter_out_nourl rt.repo_opams);
+        cached_sys_pkg_statues = OpamRepositoryName.Map.bindings
+        (filter_out_nourl rt.sys_pkg_statues);
       }
 
   let file rt =
@@ -72,7 +75,8 @@ module Cache = struct
     | Some cache ->
       Some
         (OpamRepositoryName.Map.of_list cache.cached_repofiles,
-         OpamRepositoryName.Map.of_list cache.cached_opams)
+         OpamRepositoryName.Map.of_list cache.cached_opams,
+         OpamRepositoryName.Map.of_list cache.cached_sys_pkg_statues)
     | None -> None
 
 end
@@ -122,6 +126,20 @@ let load_repo repo repo_root =
     (OpamRepositoryName.to_string repo.repo_name)
     (t ());
   repo_def, opams
+
+let get_repo_depexts opams gt =
+  let env = OpamPackageVar.resolve_global gt in
+  OpamPackage.Map.fold (fun _ opam acc ->
+      let depexts =
+        List.fold_left (fun depexts (names, filter) ->
+            if OpamFilter.eval_to_bool ~default:false env filter then
+              OpamSysPkg.Set.Op.(names ++ depexts)
+            else
+              depexts)
+          OpamSysPkg.Set.empty (OpamFile.OPAM.depexts opam) 
+      in
+      OpamSysPkg.Set.Op.(depexts ++ acc)) 
+    opams OpamSysPkg.Set.empty
 
 (* Cleaning directories follows the repo path pattern:
    TMPDIR/opam-tmp-dir/repo-dir, defined in [load]. *)
@@ -193,7 +211,7 @@ let load lock_kind gt =
         ) in
         Hashtbl.add repos_tmp name tmp
     ) repositories;
-  let make_rt repos_definitions opams =
+  let make_rt repos_definitions opams sys_pkg_statues =
     let rt = {
       repos_global = (gt :> unlocked global_state);
       repos_lock = lock;
@@ -201,28 +219,36 @@ let load lock_kind gt =
       repositories;
       repos_definitions;
       repo_opams = opams;
+      sys_pkg_statues
     } in
     OpamStd.Sys.at_exit (fun () -> cleanup rt);
     rt
   in
   match Cache.load gt.root with
-  | Some (repofiles, opams) ->
+  | Some (repofiles, opams, sys_pkg_statues) ->
     log "Cache found";
-    make_rt repofiles opams
+    make_rt repofiles opams sys_pkg_statues
   | None ->
     log "No cache found";
     OpamFilename.with_flock_upgrade `Lock_read lock @@ fun _ ->
-    let repofiles, opams =
-      OpamRepositoryName.Map.fold (fun name url (defs, opams) ->
+    let repofiles, opams, sys_pkg_statues =
+      OpamRepositoryName.Map.fold (fun name url (defs, opams, sys_pkg_statues) ->
           let repo = mk_repo name url in
           let repo_def, repo_opams =
             load_repo repo (get_root_raw gt.root repos_tmp name)
           in
+          let repo_depexts = get_repo_depexts repo_opams gt in
+          let repo_sys_pkg_status =
+            OpamSysInteract.packages_status gt.config repo_depexts
+          in
           OpamRepositoryName.Map.add name repo_def defs,
-          OpamRepositoryName.Map.add name repo_opams opams)
-        repos_map (OpamRepositoryName.Map.empty, OpamRepositoryName.Map.empty)
+          OpamRepositoryName.Map.add name repo_opams opams,
+          OpamRepositoryName.Map.add name repo_sys_pkg_status sys_pkg_statues
+        )
+        repos_map (OpamRepositoryName.Map.empty, OpamRepositoryName.Map.empty,
+                   OpamRepositoryName.Map.empty)
     in
-    let rt = make_rt repofiles opams in
+    let rt = make_rt repofiles opams sys_pkg_statues in
     Cache.save_new rt;
     rt
 
--- a/src/state/opamRepositoryState.ml
+++ b/src/state/opamRepositoryState.ml
@@ -251,8 +251,7 @@ let load lock_kind gt =
           in
           OpamRepositoryName.Map.add name repo_def defs,
           OpamRepositoryName.Map.add name repo_opams opams,
-          OpamRepositoryName.Map.add name repo_sys_pkg_status sys_pkg_statues
-        )
+          OpamRepositoryName.Map.add name repo_sys_pkg_status sys_pkg_statues)
         repos_map (OpamRepositoryName.Map.empty, OpamRepositoryName.Map.empty,
                    OpamRepositoryName.Map.empty)
     in
--- a/src/state/opamRepositoryState.ml
+++ b/src/state/opamRepositoryState.ml
@@ -20,7 +20,7 @@ module Cache = struct
   type t = {
     cached_repofiles: (repository_name * OpamFile.Repo.t) list;
     cached_opams: (repository_name * OpamFile.OPAM.t OpamPackage.Map.t) list;
-    cached_sys_pkg_statues: (repository_name * sys_pkg_status) list;
+    cached_sys_pkg_statues: (repository_name * sys_pkg_status OpamPackage.Map.t) list;
   }
 
   module C = OpamCached.Make (struct
@@ -129,7 +129,7 @@ let load_repo repo repo_root =
 
 let get_repo_depexts opams gt =
   let env = OpamPackageVar.resolve_global gt in
-  OpamPackage.Map.fold (fun _ opam acc ->
+  OpamPackage.Map.fold (fun package opam acc ->
       let depexts =
         List.fold_left (fun depexts (names, filter) ->
             if OpamFilter.eval_to_bool ~default:false env filter then
@@ -138,8 +138,8 @@ let get_repo_depexts opams gt =
               depexts)
           OpamSysPkg.Set.empty (OpamFile.OPAM.depexts opam) 
       in
-      OpamSysPkg.Set.Op.(depexts ++ acc)) 
-    opams OpamSysPkg.Set.empty
+      OpamPackage.Map.add package depexts acc ) 
+    opams OpamPackage.Map.empty
 
 (* Cleaning directories follows the repo path pattern:
    TMPDIR/opam-tmp-dir/repo-dir, defined in [load]. *)
@@ -239,7 +239,9 @@ let load lock_kind gt =
           in
           let repo_depexts = get_repo_depexts repo_opams gt in
           let repo_sys_pkg_status =
-            OpamSysInteract.packages_status gt.config repo_depexts
+            OpamPackage.Map.map (fun sys_packages -> 
+                OpamSysInteract.packages_status gt.config sys_packages)
+              repo_depexts
           in
           OpamRepositoryName.Map.add name repo_def defs,
           OpamRepositoryName.Map.add name repo_opams opams,
--- a/src/state/opamRepositoryState.ml
+++ b/src/state/opamRepositoryState.ml
@@ -129,17 +129,19 @@ let load_repo repo repo_root =
 
 let get_repo_depexts opams gt =
   let env = OpamPackageVar.resolve_global gt in
-  OpamPackage.Map.fold (fun package opam acc ->
+  OpamPackage.Map.fold (fun package opam (s,m) ->
+      let open OpamSysPkg.Set.Op in 
       let depexts =
-        List.fold_left (fun depexts (names, filter) ->
+        List.fold_left (fun (depexts) (names, filter) ->
             if OpamFilter.eval_to_bool ~default:false env filter then
-              OpamSysPkg.Set.Op.(names ++ depexts)
+              (names ++ depexts)
             else
               depexts)
           OpamSysPkg.Set.empty (OpamFile.OPAM.depexts opam) 
       in
-      OpamPackage.Map.add package depexts acc ) 
-    opams OpamPackage.Map.empty
+      (depexts ++ s),
+      OpamPackage.Map.add package depexts m ) 
+    opams (OpamSysPkg.Set.empty, OpamPackage.Map.empty)
 
 (* Cleaning directories follows the repo path pattern:
    TMPDIR/opam-tmp-dir/repo-dir, defined in [load]. *)
@@ -237,11 +239,15 @@ let load lock_kind gt =
           let repo_def, repo_opams =
             load_repo repo (get_root_raw gt.root repos_tmp name)
           in
-          let repo_depexts = get_repo_depexts repo_opams gt in
+          let repo_depexts, pkg_to_depext = get_repo_depexts repo_opams gt in
+          let status = OpamSysInteract.packages_status gt.config repo_depexts in 
           let repo_sys_pkg_status =
-            OpamPackage.Map.map (fun sys_packages -> 
-                OpamSysInteract.packages_status gt.config sys_packages)
-              repo_depexts
+            OpamPackage.Map.map (fun set ->
+                OpamSysPkg.Set.Op.
+                  { OpamSysPkg.
+                    s_available = set %% status.s_available;
+                    s_not_found = set %% status.s_not_found })
+              pkg_to_depext
           in
           OpamRepositoryName.Map.add name repo_def defs,
           OpamRepositoryName.Map.add name repo_opams opams,
--- a/src/state/opamStateTypes.mli
+++ b/src/state/opamStateTypes.mli
@@ -90,6 +90,9 @@ type +'lock repos_state = {
   repo_opams: OpamFile.OPAM.t package_map repository_name_map;
   (** All opam files that can be found in the configured repositories *)
 
+  sys_pkg_statues: sys_pkg_status repository_name_map;
+  (** System package statuses of all depexts mentionned in opam files in a given repository *)
+
   repos_tmp: (OpamRepositoryName.t, OpamFilename.Dir.t Lazy.t) Hashtbl.t;
   (** Temporary directories containing the uncompressed contents of the
       repositories *)
--- a/src/state/opamStateTypes.mli
+++ b/src/state/opamStateTypes.mli
@@ -138,7 +138,7 @@ type +'lock switch_state = {
   packages: package_set;
   (** The set of all known packages *)
 
-  sys_packages: sys_pkg_status package_map Lazy.t;
+  sys_packages: sys_pkg_status package_map;
   (** Map of package and their system dependencies packages status. Only
       initialised for otherwise available packages *)
 
--- a/src/state/opamStateTypes.mli
+++ b/src/state/opamStateTypes.mli
@@ -90,7 +90,7 @@ type +'lock repos_state = {
   repo_opams: OpamFile.OPAM.t package_map repository_name_map;
   (** All opam files that can be found in the configured repositories *)
 
-  sys_pkg_statues: sys_pkg_status repository_name_map;
+  sys_pkg_statues: sys_pkg_status package_map repository_name_map;
   (** System package statuses of all depexts mentionned in opam files in a given repository *)
 
   repos_tmp: (OpamRepositoryName.t, OpamFilename.Dir.t Lazy.t) Hashtbl.t;
--- a/src/state/opamSwitchState.ml
+++ b/src/state/opamSwitchState.ml
@@ -565,16 +565,16 @@ let load lock_kind gt rt switch =
     || OpamStateConfig.(!r.no_depexts) then
       OpamPackage.Map.empty
     else (
-(*       let available = (Lazy.force available_packages) in 
-      depexts_status_from_repo available gt.config switch_config rt  *)
-      depexts_status_of_packages_raw gt.config switch_config
+       let available = (Lazy.force available_packages) in 
+      depexts_status_from_repo available gt.config switch_config rt  
+    (*   depexts_status_of_packages_raw gt.config switch_config
           ~env:gt.global_variables
           (Lazy.force available_packages)
           ~depexts:(fun package ->
              let env =
                OpamPackageVar.resolve_switch_raw ~package gt switch switch_config
              in
-             depexts_raw ~env package opams) 
+             depexts_raw ~env package opams)  *)
     )
   in
   let available_packages =
--- a/src/state/opamSwitchState.ml
+++ b/src/state/opamSwitchState.ml
@@ -178,8 +178,6 @@ module Installed_cache = OpamCached.Make(struct
     let name = "installed"
   end)
 
-
-
 let depexts_status_from_repo packages global_config switch_config rt =
   let open OpamSysPkg.Set.Op in
   let bypass =
@@ -187,11 +185,6 @@ let depexts_status_from_repo packages global_config switch_config rt =
     switch_config.OpamFile.Switch_config.depext_bypass
   in
   let repos = repos_list_raw rt switch_config in
-  let initial_map =
-    OpamPackage.Set.fold (fun pkg acc ->
-        OpamPackage.Map.add pkg OpamSysPkg.status_empty acc
-      ) packages OpamPackage.Map.empty
-  in
   let check_status status = 
     if OpamStateConfig.(!r.no_depexts) then
       (* Mark all as available. This is necessary to store the exceptions
@@ -204,27 +197,22 @@ let depexts_status_from_repo packages global_config switch_config rt =
     else
       status
   in 
-  List.fold_left (fun acc_map repo_name ->
+  List.fold_left (fun pkg_to_sys repo_name ->
       match OpamRepositoryName.Map.find_opt repo_name rt.sys_pkg_statues with
-      | None -> acc_map
-      | Some pkg_to_status ->
+      | None -> pkg_to_sys
+      | Some repo_pkg_to_status ->
         OpamPackage.Map.fold (fun pkg sys_status acc ->
             if OpamPackage.Set.mem pkg packages then
-              let current = OpamPackage.Map.find pkg acc in
-              let merged = OpamSysPkg.{
-                  s_available = (sys_status.s_available ++ current.s_available -- bypass);
-                  s_not_found = (sys_status.s_not_found ++ current.s_not_found -- bypass);
+              let sys_pkg_status = OpamSysPkg.{
+                  s_available = (sys_status.s_available -- bypass);
+                  s_not_found = (sys_status.s_not_found -- bypass);
                 } in
-              let status_checked = check_status merged in 
+              let status_checked = check_status sys_pkg_status in 
               OpamPackage.Map.add pkg status_checked acc
             else
               acc
-          ) pkg_to_status acc_map
-    ) initial_map repos
-
-  
-     
-
+          ) repo_pkg_to_status pkg_to_sys
+    ) OpamPackage.Map.empty repos
 
 let depexts_status_of_packages_raw
     ~depexts ?env global_config switch_config packages =
@@ -566,7 +554,7 @@ let load lock_kind gt rt switch =
       OpamPackage.Map.empty
     else (
        let available = (Lazy.force available_packages) in 
-      depexts_status_from_repo available gt.config switch_config rt  
+      depexts_status_from_repo available gt.config switch_config rt
     (*   depexts_status_of_packages_raw gt.config switch_config
           ~env:gt.global_variables
           (Lazy.force available_packages)
--- a/src/state/opamSwitchState.ml
+++ b/src/state/opamSwitchState.ml
@@ -515,8 +515,8 @@ let load lock_kind gt rt switch =
   let sys_packages =
     if not (OpamFile.Config.depext gt.config)
     || OpamStateConfig.(!r.no_depexts) then
-      lazy OpamPackage.Map.empty
-    else lazy (
+       OpamPackage.Map.empty
+    else (
       depexts_status_of_packages_raw gt.config switch_config
         ~env:gt.global_variables
         (Lazy.force available_packages)
@@ -530,7 +530,7 @@ let load lock_kind gt rt switch =
   let available_packages =
     if not (OpamFile.Config.depext gt.config) then available_packages
     else lazy (
-      let sys_packages = Lazy.force sys_packages in
+      let sys_packages =  sys_packages in
       OpamPackage.Set.filter (fun nv ->
           depexts_unavailable_raw sys_packages nv = None)
         (Lazy.force available_packages)
@@ -542,7 +542,7 @@ let load lock_kind gt rt switch =
           OpamPackage.Set.mem pkg installed
           && not (OpamSysPkg.Set.is_empty spkg.OpamSysPkg.s_available
                   && OpamSysPkg.Set.is_empty spkg.OpamSysPkg.s_not_found))
-        (Lazy.force sys_packages)
+        sys_packages
     in
     if OpamSysInteract.stateless_install ~env:gt.global_variables ()
        || OpamPackage.Map.is_empty sys_packages then
@@ -656,7 +656,7 @@ let load_virtual ?repos_list ?(avail_default=true) gt rt =
     opams;
     conf_files = OpamPackage.Name.Map.empty;
     packages;
-    sys_packages = lazy OpamPackage.Map.empty;
+    sys_packages = OpamPackage.Map.empty;
     available_packages;
     reinstall = lazy OpamPackage.Set.empty;
     invalidated = lazy (OpamPackage.Set.empty);
@@ -781,7 +781,7 @@ let depexts_status_of_packages st set =
     ~env:st.switch_global.global_variables ~depexts:(depexts st)
 
 let depexts_unavailable st nv =
-  depexts_unavailable_raw (Lazy.force st.sys_packages) nv
+  depexts_unavailable_raw st.sys_packages nv
 
 let dev_packages st =
   OpamPackage.Set.filter (is_dev_package st)
@@ -1029,7 +1029,7 @@ let universe st
           if OpamSysPkg.Set.is_empty status.OpamSysPkg.s_available
           then acc
           else OpamPackage.Set.add nv acc)
-        (Lazy.force st.sys_packages)
+        st.sys_packages
         OpamPackage.Set.empty)
   in
   let avoid_versions =
@@ -1264,16 +1264,16 @@ let update_pin nv opam st =
   if not (OpamFile.Config.depext st.switch_global.config)
   || OpamSysPkg.Set.is_empty (depexts st nv)
   then st else
-  let sys_packages = lazy (
-    OpamPackage.Map.union (fun _ n -> n)
-      (Lazy.force st.sys_packages)
-      (depexts_status_of_packages st (OpamPackage.Set.singleton nv))
-  ) in
-  let available_packages = lazy (
-    OpamPackage.Set.filter (fun nv -> depexts_unavailable st nv = None)
-      (Lazy.force st.available_packages)
-  ) in
-  { st with sys_packages; available_packages }
+    let sys_packages = (
+      OpamPackage.Map.union (fun _ n -> n)
+        st.sys_packages
+        (depexts_status_of_packages st (OpamPackage.Set.singleton nv))
+    ) in
+    let available_packages = lazy (
+      OpamPackage.Set.filter (fun nv -> depexts_unavailable st nv = None)
+        (Lazy.force st.available_packages)
+    ) in
+    { st with sys_packages; available_packages }
 
 let do_backup lock st = match lock with
   | `Lock_write ->
--- a/src/state/opamSwitchState.ml
+++ b/src/state/opamSwitchState.ml
@@ -178,6 +178,54 @@ module Installed_cache = OpamCached.Make(struct
     let name = "installed"
   end)
 
+
+
+let depexts_status_from_repo packages global_config switch_config rt =
+  let open OpamSysPkg.Set.Op in
+  let bypass =
+    OpamFile.Config.depext_bypass global_config ++
+    switch_config.OpamFile.Switch_config.depext_bypass
+  in
+  let repos = repos_list_raw rt switch_config in
+  let initial_map =
+    OpamPackage.Set.fold (fun pkg acc ->
+        OpamPackage.Map.add pkg OpamSysPkg.status_empty acc
+      ) packages OpamPackage.Map.empty
+  in
+  let check_status status = 
+    if OpamStateConfig.(!r.no_depexts) then
+      (* Mark all as available. This is necessary to store the exceptions
+         afterwards *)
+      OpamSysPkg.{ status_empty with
+                   s_available = status.s_available ++ status.s_not_found; }
+    else if OpamFile.Config.depext_cannot_install global_config then
+      { OpamSysPkg.status_empty with
+        s_not_found = status.s_available ++ status.s_not_found; }
+    else
+      status
+  in 
+  List.fold_left (fun acc_map repo_name ->
+      match OpamRepositoryName.Map.find_opt repo_name rt.sys_pkg_statues with
+      | None -> acc_map
+      | Some pkg_to_status ->
+        OpamPackage.Map.fold (fun pkg sys_status acc ->
+            if OpamPackage.Set.mem pkg packages then
+              let current = OpamPackage.Map.find pkg acc in
+              let merged = OpamSysPkg.{
+                  s_available = (sys_status.s_available ++ current.s_available -- bypass);
+                  s_not_found = (sys_status.s_not_found ++ current.s_not_found -- bypass);
+                } in
+              let status_checked = check_status merged in 
+              OpamPackage.Map.add pkg status_checked acc
+            else
+              acc
+          ) pkg_to_status acc_map
+    ) initial_map repos
+
+  
+     
+
+
 let depexts_status_of_packages_raw
     ~depexts ?env global_config switch_config packages =
   if OpamPackage.Set.is_empty packages then OpamPackage.Map.empty else
@@ -515,22 +563,24 @@ let load lock_kind gt rt switch =
   let sys_packages =
     if not (OpamFile.Config.depext gt.config)
     || OpamStateConfig.(!r.no_depexts) then
-       OpamPackage.Map.empty
+      OpamPackage.Map.empty
     else (
+(*       let available = (Lazy.force available_packages) in 
+      depexts_status_from_repo available gt.config switch_config rt  *)
       depexts_status_of_packages_raw gt.config switch_config
-        ~env:gt.global_variables
-        (Lazy.force available_packages)
-        ~depexts:(fun package ->
-            let env =
-              OpamPackageVar.resolve_switch_raw ~package gt switch switch_config
-            in
-            depexts_raw ~env package opams)
+          ~env:gt.global_variables
+          (Lazy.force available_packages)
+          ~depexts:(fun package ->
+             let env =
+               OpamPackageVar.resolve_switch_raw ~package gt switch switch_config
+             in
+             depexts_raw ~env package opams) 
     )
   in
   let available_packages =
     if not (OpamFile.Config.depext gt.config) then available_packages
     else lazy (
-      let sys_packages =  sys_packages in
+      let sys_packages = sys_packages in
       OpamPackage.Set.filter (fun nv ->
           depexts_unavailable_raw sys_packages nv = None)
         (Lazy.force available_packages)
--- a/tests/reftests/run.ml
+++ b/tests/reftests/run.ml
@@ -1039,9 +1039,9 @@ let run_test ?(vars=[]) ~opam t =
               in
               match cache with
               | None -> print_string "No cache\n"
-              | Some (_, cache) when OpamRepositoryName.Map.is_empty cache ->
+              | Some (_, cache, _) when OpamRepositoryName.Map.is_empty cache ->
                 print_string "Empty cache\n"
-              | Some (_, cache) ->
+              | Some (_, cache, _) ->
                 let cache =
                   if OpamPackage.Name.Map.is_empty nvs then cache else
                     OpamRepositoryName.Map.map
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,8 +18,7 @@ on:
       - 'shell/'
   push:
     branches:
-      - 'master'
-      - '2.**'
+      - '*'
 
 env:
   OPAMBSVERSION: 2.1.0
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -18,7 +18,8 @@ on:
       - 'shell/'
   push:
     branches:
-      - '*'
+      - 'master'
+      - '2.**'
 
 env:
   OPAMBSVERSION: 2.1.0
--- a/src/client/opamAdminCommand.ml
+++ b/src/client/opamAdminCommand.ml
@@ -1078,6 +1078,7 @@ let get_virtual_switch_state repo_root env =
     repos_definitions = singl repo_def;
     repo_opams = singl opams;
     repos_tmp;
+    sys_pkg_statues = singl (OpamSysPkg.status_empty);
   } in
   let gt =
     {gt with global_variables =
--- a/src/client/opamAdminCommand.ml
+++ b/src/client/opamAdminCommand.ml
@@ -1078,7 +1078,7 @@ let get_virtual_switch_state repo_root env =
     repos_definitions = singl repo_def;
     repo_opams = singl opams;
     repos_tmp;
-    sys_pkg_statues = singl (OpamSysPkg.status_empty);
+    sys_pkg_statues = singl (OpamPackage.Map.empty); (* TODO: should it be computed here ? *)
   } in
   let gt =
     {gt with global_variables =
--- a/src/state/opamRepositoryState.mli
+++ b/src/state/opamRepositoryState.mli
@@ -22,7 +22,8 @@ module Cache: sig
   val load:
     dirname ->
     (OpamFile.Repo.t repository_name_map *
-     OpamFile.OPAM.t package_map repository_name_map)
+     OpamFile.OPAM.t package_map repository_name_map *
+     sys_pkg_status repository_name_map)
       option
   val remove: unit -> unit
 end
--- a/src/state/opamRepositoryState.mli
+++ b/src/state/opamRepositoryState.mli
@@ -23,7 +23,7 @@ module Cache: sig
     dirname ->
     (OpamFile.Repo.t repository_name_map *
      OpamFile.OPAM.t package_map repository_name_map *
-     sys_pkg_status repository_name_map)
+     sys_pkg_status package_map repository_name_map)
       option
   val remove: unit -> unit
 end


src/state/opamStateTypes.mli
```diff
@@ -90,6 +90,9 @@ type +'lock repos_state = {
   repo_opams: OpamFile.OPAM.t package_map repository_name_map;
   (** All opam files that can be found in the configured repositories *)
 
+  sys_pkg_statues: sys_pkg_status package_map repository_name_map;
```
arozovyk, line 93:
```suggestion
  repo_sys_pkg_statues: sys_pkg_status package_map repository_name_map;

File: master_changes.md [hunk 2]
```diff
@@ -116,10 +116,12 @@ users)
 
 # API updates
 ## opam-client
+  * Remove the heuristic of recomputing depexts of additional (pinned) packages in `OpamSolution` and move the logic to `OpamClient.install_t` and `OpamAuxCommand.autopin` [#6461 @arozovyk]
```
rjbou: Do not belong to this commit

src/state/opamSwitchState.ml
```diff
@@ -1275,6 +1275,15 @@ let update_pin nv opam st =
   ) in
   { st with sys_packages; available_packages }
 
+let update_sys_packages pkgs st = 
+  if OpamPackage.Set.is_empty pkgs then
+    st 
+  else
+    {st with sys_packages =
+               lazy ((OpamPackage.Map.union (fun x _ -> x)
```
rjbou:
```
               lazy ((OpamPackage.Map.union (fun x _ -> x)
```
Why you take the first element, the one that is already in sys_package instead of updating with the new status ?

master_changes.md
```diff
@@ -105,8 +105,10 @@ users)
 
 ## Reftests
 ### Tests
+  * Use `opam-set-os` in reftests following the depexts update.
 
 ### Engine
+  * Add `opam-set-os` command that combines setting global `os-family` variable followed by a (silent) `opam update` [#6461 @arozovyk]
```
```suggestion
  * Add `opam-set-os` command that combines setting global `os-family` variable followed by a (silent) `opam update` [#6489 @arozovyk]
```

master_changes.md
```diff
@@ -122,6 +122,8 @@ users)
 
 ## opam-state
   * OpamSwitchState: add `update_sys_packages` function to update depexts status of a set of packages. [#6461 @arozovyk]
+  * Add a function to update system dependencies status of additional packages in `OpamSwitchState` [#6461 @arozovyk]
+  * Split depexts status function in `OpamSysInteract` for available and installed to be computed separately [#6461 @arozovyk]
```
```suggestion
  * `OpamSysInteract`: add `available_status` and `installed_status` for available and installed to be computed separately [#6489 @arozovyk]

File: src/state/opamSysInteract.ml [hunk 6]
```diff
@@ -484,21 +484,13 @@ let yum_cmd = lazy begin
     raise (OpamSystem.Command_not_found "yum or dnf")
 end
 
-let packages_status ?(env=OpamVariable.Map.empty) config packages =
+let installed_packages ?(env=OpamVariable.Map.empty) config packages =
   let (+++) pkg set = OpamSysPkg.Set.add (OpamSysPkg.of_string pkg) set in
   (* Some package managers don't permit to request on available packages. In
      this case, we consider all non installed packages as [available]. *)
```
rjbou: not needed comment in this function

src/state/opamSysInteract.ml
```diff
@@ -556,7 +548,7 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
           | Some ps -> OpamSysPkg.Set.union acc ps)
         sys_installed sys_installed
     in
-    compute_sets sys_installed ~sys_available
+    get_relevant sys_installed
```
```suggestion
    get_relevant sys_installed

File: src/state/opamSysInteract.ml [hunk 8]
```diff
@@ -538,7 +530,7 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
       ) OpamSysPkg.Set.empty l
   in
   let compute_sets_with_virtual get_avail_w_virtuals get_installed  =
```
rjbou: `get_avail_w_virtuals` do not need to compute available in this function

src/state/opamSysInteract.ml
```diff
@@ -654,7 +646,7 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
        >    https://dl-cdn.alpinelinux.org/alpine/edge/main
        >    @edge https://dl-cdn.alpinelinux.org/alpine/edge/main
     *)
-    let sys_installed, sys_available =
+    let sys_installed, _sys_available =
```
```suggestion
    let sys_installed =

File: src/state/opamSysInteract.ml [hunk 10]
```diff
@@ -693,15 +685,15 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
...
-            in
-            pkg, installed, add_pkg pkg repo installed instavail)
+              let repo =
+                try Some Re.(Group.get (exec repo_name l) 1)
+                with Not_found -> None
+              in
+              pkg, installed, add_pkg pkg repo installed instavail)
         ("", false, OpamSysPkg.Set.(empty, empty))
       |> (fun (_,_, instavail) -> instavail)
     in
```
rjbou, line 772: there is no need to compute available set here

src/state/opamSysInteract.ml
```diff
@@ -817,20 +809,14 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
...
       | `set pkgs -> pkgs %% packages
     in
-    let sys_available =
-      match test.available with
-      | `all -> packages
-      | `none -> OpamSysPkg.Set.empty
-      | `set pkgs -> pkgs %% packages
-    in
-    compute_sets ~sys_available sys_installed
+    get_relevant  sys_installed
```
```suggestion
    get_relevant sys_installed

File: src/state/opamSysInteract.ml [hunk 12]
```diff
@@ -928,6 +914,267 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
...
+  let with_regexp_sgl re_pkg =
+    List.fold_left (fun pkgs l ->
+        try
+          Re.(Group.get (exec re_pkg l) 1) +++ pkgs
+        with Not_found -> pkgs) OpamSysPkg.Set.empty
+  in
+  let compute_sets_with_virtual get_avail_w_virtuals  =
+    let sys_available, _ = get_avail_w_virtuals () in
+    get_relevant sys_available
+  in
```
rjbou: I don't know if it worth keeping this function

rjbou: `get_avail_w_virtuals` should only return available package in this function

src/state/opamSysInteract.ml
```diff
@@ -928,6 +914,267 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
...
+       >  8.2.2852-r0:
+       >    @edge https://dl-cdn.alpinelinux.org/alpine/edge/main
+       >hwids-udev policy:
+       >  20201207-r0:
+       >    https://dl-cdn.alpinelinux.org/alpine/v3.13/main
+       >    @edge https://dl-cdn.alpinelinux.org/alpine/v3.13/main
+       >    https://dl-cdn.alpinelinux.org/alpine/edge/main
+       >    @edge https://dl-cdn.alpinelinux.org/alpine/edge/main
+    *)
+    let _sys_installed, sys_available =
```
rjbou: Same here, no need to compute installed

src/state/opamSysInteract.ml
```diff
@@ -967,33 +1214,41 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
...
+    Suppose_available
+  | Openbsd -> Suppose_available
+  
+let packages_status ?(env=OpamVariable.Map.empty) ?sys_available config syspkg_set =
+  log "calling new depexts status";
+  let open OpamSysPkg.Set.Op in
+  let installed = installed_packages ~env config syspkg_set in
+  let sys_available = 
+    match sys_available with 
+      Some sys_available -> sys_available
```
```suggestion
      | Some sys_available -> sys_available

File: src/state/opamSysInteract.ml [hunk 15]
```diff
@@ -967,33 +1214,41 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
...
+       Saying no packages are installed results in a warning that
+       conf packages depend on a 'system package that can no longer
+       be found.' But omitting them will mean that they won't be
+       added to the Nix derivation.
+    *)
+    Suppose_available
+  | Openbsd -> Suppose_available
+  
+let packages_status ?(env=OpamVariable.Map.empty) ?sys_available config syspkg_set =
+  log "calling new depexts status";
```
rjbou: If we keep it, it need some rephrasing

src/state/opamSysInteract.ml
```diff
@@ -1280,17 +1535,18 @@ let update ?(env=OpamVariable.Map.empty) config =
...
        This is typically done by installing the 'epel-release' package. \
-       Please see https://fedoraproject.org/wiki/EPEL for more information"
+       Please see https://fedoraproject.org/wiki/EPEL for more information" 
+    in 
+    match available_packages ~env config epel_release_pkg with 
+    | Suppose_available -> Some msg
+    | Available av -> 
+      if OpamSysPkg.Set.equal av epel_release_pkg 
+      then Some msg
+      else None
```
```suggestion
      if OpamSysPkg.Set.equal av epel_release_pkg then
        Some msg
      else None

File: src/state/opamSysInteract.mli [hunk 17]
```diff
@@ -16,10 +16,21 @@ open OpamStateTypes
        repositories, but not installed
      * not found set: packages not found on the defined repositories
    [env] is used to determine host specification.
+   [sys_available] is used to avoid a syscall when possible.
```
rjbou: Not very clear how it avoids a syscall and how should I use that argument to avoid that syscall.

src/state/opamSysInteract.mli
```diff
@@ -16,10 +16,21 @@ open OpamStateTypes
...
+  OpamFile.Config.t -> OpamSysPkg.Set.t -> OpamSysPkg.status
+
+(* Returns available system packages *)
+val available_packages:
+  ?env:gt_variables -> OpamFile.Config.t -> OpamSysPkg.Set.t ->
+  OpamSysPkg.available
+
+(* Returns installed system packages *)
+val installed_packages:
   ?env:gt_variables -> OpamFile.Config.t -> OpamSysPkg.Set.t ->
```
rjbou: These 2 functions should be defined before `package_status` as they will be used by it. A complete documentation is also needed for the first one defined. You can use `package_status` one, and in the second & third you can refer to the first.

src/state/opamRepositoryState.ml
```diff
@@ -193,36 +219,45 @@ let load lock_kind gt =
...
+  | Some (repofiles, opams, sys_available_pkgs) ->
     log "Cache found";
-    make_rt repofiles opams
+    make_rt repofiles opams sys_available_pkgs
   | None ->
     log "No cache found";
     OpamFilename.with_flock_upgrade `Lock_read lock @@ fun _ ->
-    let repofiles, opams =
-      OpamRepositoryName.Map.fold (fun name url (defs, opams) ->
+    let repofiles, opams , repos_sys_available_pkgs =
```
```suggestion
    let repofiles, opams, repos_sys_available_pkgs =

File: src/state/opamUpdate.ml [hunk 20]
```diff
@@ -186,14 +186,49 @@ let repositories rt repos =
...
+        let repo_depexts = 
+          OpamRepositoryState.get_repo_declared_depexts opams gt
+        in
+        OpamSysInteract.available_packages ~env:gt.global_variables 
+          gt.config repo_depexts 
+      ) rt.repo_opams
+  in
+  let write_config_cache rt = 
+    OpamRepositoryState.write_config rt;
+    OpamRepositoryState.Cache.save rt;
```
rjbou:
```suggestion
    OpamRepositoryState.Cache.save rt

File: src/state/opamUpdate.ml [hunk 21]
```diff
@@ -186,14 +186,49 @@ let repositories rt repos =
...
+  in
+  let depexts_available_equal old new_ = 
+    OpamRepositoryName.Map.equal
+      (fun o n ->
+         match o, n with 
+           OpamSysPkg.Suppose_available, _ -> true 
+         |_, OpamSysPkg.Suppose_available -> true
+         | OpamSysPkg.Available os, OpamSysPkg.Available ns -> 
+           OpamSysPkg.Set.equal os ns)
```
rjbou: it may worth to have the `equal` in `OpamSysPkg`

src/state/opamStateTypes.mli
```diff
@@ -90,6 +90,9 @@ type +'lock repos_state = {
   repo_opams: OpamFile.OPAM.t package_map repository_name_map;
   (** All opam files that can be found in the configured repositories *)
 
+  repos_sys_available_pkgs : OpamSysPkg.available repository_name_map;
```
rjbou: Do you remember why we went for a reponame map instead of a `OpamSysPkg.available` that contains all information aggregated?

src/state/opamSwitchState.ml
```diff
@@ -196,8 +222,15 @@ let depexts_status_of_packages_raw
     switch_config.OpamFile.Switch_config.depext_bypass
   in
   let syspkg_set = syspkg_set -- bypass in
+  let status syspkg_set = 
+    if recompute_available then 
+      OpamSysInteract.packages_status ?env global_config syspkg_set
+    else
```
rjbou: It worth take as argument the depexts of pinned packages, like that we recompute only for those ones.

src/state/opamSwitchState.ml
```diff
@@ -1267,22 +1304,38 @@ let update_pin nv opam st =
...
   ) in
   let available_packages = lazy (
     OpamPackage.Set.filter (fun nv -> depexts_unavailable st nv = None)
       (Lazy.force st.available_packages)
   ) in
   { st with sys_packages; available_packages }
 
-let update_sys_packages pkgs st = 
-  if OpamPackage.Set.is_empty pkgs then
+let update_sys_packages packages st =
```
rjbou:
```suggestion
let update_sys_packages packages st =
```

rjbou: Also, why change the name of the argument ?

src/state/opamSwitchState.ml
```diff
@@ -1267,22 +1304,38 @@ let update_pin nv opam st =
...
   let available_packages = lazy (
     OpamPackage.Set.filter (fun nv -> depexts_unavailable st nv = None)
       (Lazy.force st.available_packages)
   ) in
   { st with sys_packages; available_packages }
 
-let update_sys_packages pkgs st = 
-  if OpamPackage.Set.is_empty pkgs then
+let update_sys_packages packages st = 
+  let depexts_s = OpamPackage.Set.fold
```
rjbou:
```suggestion
  let depexts_s =
    OpamPackage.Set.fold

File: src/client/opamSolution.ml [hunk 26]
```diff
@@ -1263,8 +1256,18 @@ let install_sys_packages_t ~propagate_st ~map_sysmap ~confirm env config
       check_again t sys_packages
   and check_again t sys_packages =
     let open OpamSysPkg.Set.Op in
+    let sys_available =
+      match propagate_st t with
```
rjbou: The comment at the beginning of the function need to be updated as `propagate_st` is used for something else now too.

rjbou: The documentation of `OpamSolution.install_sys_packages` too.

src/state/opamSysInteract.ml
```diff
@@ -967,33 +1214,41 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
...
+  let open OpamSysPkg.Set.Op in
+  let installed = installed_packages ~env config syspkg_set in
+  let sys_available = 
+    match sys_available with 
+      Some sys_available -> sys_available
+    | None ->
+      available_packages ~env config syspkg_set 
+  in 
+  match sys_available with
+    OpamSysPkg.Available sys_pkgs ->
```
```suggestion
    | OpamSysPkg.Available sys_pkgs ->

File: src/state/opamRepositoryState.ml [hunk 28]
```diff
@@ -154,6 +158,28 @@ let get_root rt name =
 let get_repo_root rt repo =
   get_root_raw rt.repos_global.root rt.repos_tmp repo.repo_name
 
+let get_repo_available_depexts rt =
+  OpamRepositoryName.Map.fold (fun _ ra acc ->  
+      match ra with 
+        OpamSysPkg.Available sys_pkgs -> OpamSysPkg.Set.Op.(sys_pkgs ++ acc)
+      | OpamSysPkg.Suppose_available -> acc) 
+    rt.repos_sys_available_pkgs OpamSysPkg.Set.empty
```
rjbou: it is not possible to differentiate between `Suppose_available` and `Available set` where all sets are empty. The result is the same.

src/state/opamStateTypes.mli
rjbou: In the map, for each repository we store the information of the kind of avilability `Available | Suppose_available`. This information is system dependent, so it is "repeated".

tests/reftests/run.ml
```diff
@@ -284,6 +284,7 @@ type command =
              output: string option;
              unordered: bool;
              sort: bool;}
+  | Set_os of string
```
rjbou: This change need to be in its own commit:
* first commit that adds `opam-set-os` and use in depext
* second commit that adds the update to the command with the commi that changes the behaviour in opam

tests/reftests/run.ml
```diff
@@ -482,6 +483,14 @@ module Parse = struct
     match cmd with
     | Some "opam-cat" ->
         Opamfile { files = args; filter = rewr; }
+    | Some "opam-set-os" ->
+      (match args with 
+         os :: []-> Set_os os
```
rjbou:
```suggestion
        | os :: []-> Set_os os

File: tests/reftests/run.ml [hunk 31]
```diff
@@ -482,6 +483,14 @@ module Parse = struct
     match cmd with
     | Some "opam-cat" ->
         Opamfile { files = args; filter = rewr; }
+    | Some "opam-set-os" ->
+      (match args with 
+         os :: []-> Set_os os 
+       | _ -> failwith
+                (Printf.sprintf
+                   "Bad usage of opam-set-os %s.\n\
+                    expecting 'opam-set-os [os-name]"
```
rjbou:
```suggestion
                    expecting 'opam-set-os [depext-test-specification]"

File: tests/reftests/run.ml [hunk 32]
```diff
@@ -482,6 +483,14 @@ module Parse = struct
     match cmd with
     | Some "opam-cat" ->
         Opamfile { files = args; filter = rewr; }
+    | Some "opam-set-os" ->
```
rjbou: Update also the readme to add the command

tests/reftests/run.ml
```diff
@@ -1093,9 +1102,15 @@ let run_test ?(vars=[]) ~opam t =
...
-          | None -> vars
-          | Some v -> (v, r) :: List.filter (fun (w,_) -> v <> w) vars)
+          (match output with
+            | None -> vars
+            | Some v -> (v, r) :: List.filter (fun (w,_) -> v <> w) vars)
+        | Set_os os ->
+          ignore @@ run_cmd ~opam ~dir ~vars
+            "opam" ["var"; "--global"; Printf.sprintf "os-family=%s" os];
+          ignore @@ run_cmd ~opam ~dir ~vars ~silent:true
+            "opam" ["update"; "default"];
```
rjbou: Why only default ?

tests/reftests/run.ml
rjbou: And the test of reftest

tests/reftests/depexts.test
```diff
@@ -510,17 +510,18 @@ opam-version: "2.0"
...
-
-<><> Installing new switch packages <><><><><><><><><><><><><><><><><><><><><><>
-Switch invariant: ["comp"]
-[NOTE] External dependency handling not supported for OS family 'unknown'.
-       You can disable this check using 'opam option --global depext=false'
-
-<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
--> installed comp.1
-Done.
+Fatal error: External dependency handling not supported for OS family 'unknown'.
```
rjbou: error to investigate

rjbou (commented):
This is a first look, code only. I need to test and play with it to finish the review.

src/state/opamSwitchState.ml
arozovyk: If you remember, It did matter at some point which element to take but it was due to a buggy behaviour of the system overall that led to the fact that it had an importance. 
Here, if the package is present in `sys_package` map, means that its status have already been computed (with `depexts_status_of_packages` by the way) and the status value is the same as from `depexts_status_of_packages` called here L1286.

src/state/opamSysInteract.ml
arozovyk: Indeed, `sys_provides` is used for `sys_installed` though. I refactored a bit, but we know that this part will move quite a bit in the following PR

rjbou: The query commands will change, true, but if we don't introduce dead code it's better.

arozovyk: Refactored the functions to return only what they are supposed to return.

tests/reftests/run.ml
arozovyk: What do you mean by second point?

src/state/opamStateTypes.mli
arozovyk: Changed the type as we only need the `OpamSysPkg.available` type

src/state/opamSwitchState.ml
rjbou: This function is used in two cases : deps only newly created virtual packages and newly pinned packages. It is not an issue for the first case, but for the second case we want to retrieve the last information, no?
If this function is only meant to add new packages, the documentation and the implementation should be updated.

src/format/opamSysPkg.ml
```diff
@@ -62,6 +62,16 @@ let status_empty =
...
 
+(* System package availability *)
+type available = Available of Set.t | Suppose_available
+
+let available_equal a b =
+  match a, b with
+  | Suppose_available, Suppose_available -> true
+  | Available os, Available ns -> Set.equal os ns
+  | _ -> OpamConsole.error_and_exit `Internal_error
+           "Equality check failed due to depexts status missmatch"
```
rjbou: This is a strong requirement for an equality function, it should be noted in the documentation

src/state/opamSysInteract.ml
```diff
@@ -490,21 +490,11 @@ let yum_cmd = lazy begin
...
-    let installed = packages %% sys_installed in
-    match sys_available with
-    | Some sys_available ->
-      let s_available = (packages -- installed) %% sys_available in
-      let s_not_found = packages -- installed -- s_available in
-      { OpamSysPkg.s_available; s_not_found }
-    | None ->
-      let s_available = packages -- installed in
-      { OpamSysPkg.status_empty with s_available }
+  let get_relevant sys_installed =
```
rjbou, line 566: Is there a reason to change the name ? especially that it is not consistent with `compute_sets_with_virtual`

src/state/opamSysInteract.ml
```diff
@@ -607,14 +597,14 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
               let provides =
                 match latest with
                 | Some p -> OpamSysPkg.Map.add p ps provides
-                | None -> provides (* Bad pacman output ?? *)
+                | None -> provides (* FIXME: Bad pacman output ?? *)
```
rjbou: what is the resaon of the fixme ?

src/state/opamRepositoryState.ml
```diff
@@ -154,6 +157,21 @@ let get_root rt name =
 let get_repo_root rt repo =
   get_root_raw rt.repos_global.root rt.repos_tmp repo.repo_name
 
+let get_declared_depexts opams gt =
```
rjbou: Why not use `opamSwitchState.depexts_raw` ?

src/state/opamSwitchState.ml
```diff
@@ -280,14 +309,14 @@ let load lock_kind gt rt switch =
         sel_pinned = pinned; sel_compiler = compiler_packages; } =
     load_selections ~lock_kind gt switch
   in
-  let pinned, pinned_opams =
-    OpamPackage.Set.fold (fun nv (pinned,opams) ->
+  let pinned, pinned_opams, depexpts_present_pins =
```
rjbou:
```
  let pinned, pinned_opams, depexts_present_pins =

File: src/state/opamSwitchState.ml [hunk 40]
```diff
@@ -517,8 +547,9 @@ let load lock_kind gt rt switch =
     || OpamStateConfig.(!r.no_depexts) then
       lazy OpamPackage.Map.empty
     else lazy (
-      depexts_status_of_packages_raw gt.config switch_config
-        ~env:gt.global_variables
+      depexts_status_of_packages_raw
+        ~recompute_available:depexpts_present_pins rt.repos_sys_available_pkgs
```
rjbou: If there is no depext in the newly pinned package, do we want to update `sys_packages` ?

src/state/opamSwitchState.ml
```diff
@@ -1269,22 +1298,39 @@ let update_pin nv opam st =
...
   ) in
   let available_packages = lazy (
     OpamPackage.Set.filter (fun nv -> depexts_unavailable st nv = None)
       (Lazy.force st.available_packages)
   ) in
   { st with sys_packages; available_packages }
 
-let update_sys_packages pkgs st =
-  if OpamPackage.Set.is_empty pkgs then
+let update_sys_packages packages st =
```
rjbou:
```suggestion
let update_sys_packages pkgs st =

File: src/client/opamClient.ml [hunk 42]
```diff
@@ -2203,7 +2203,9 @@ let install_t t ?ask ?(ignore_conflicts=false) ?(depext_only=false)
           nvs (t, deps_of_packages))
       dname_map (t, OpamPackage.Set.empty)
   in
-  let t = OpamSwitchState.update_sys_packages deps_of_packages t in
+  let t =
+    OpamSwitchState.update_sys_packages deps_of_packages t
+  in
```
rjbou, line 2209: it does not belong to this commit

src/state/opamSwitchState.ml
arozovyk: You're right, I intuitevely assumed opam file would be involved when comparing packages, it's not. So we should absolutely recompute the value for newly pinned packages. Changing it

src/state/opamSwitchState.ml
arozovyk: As far as I understand it would be redundant, since `syspkg_set` is already the set of depexts of pinned packages

tests/reftests/depexts.test
arozovyk: As for the your message in d2c62e644888aa2ce4e94d833c73883380db3106

> The fix is not good as in the case where there is no depexts possible,
the previous version was returning an empty map and not supposing the
all packages are available. We may want to add another variant to `available_packages` to handle the no depext case ?

The current behaviour on unsupported OS failure is to return the empty map of depext statuses, which results in ignoring the depexts when trying to install packages.

Current PR [doesn't change this behaviour](https://github.com/arozovyk/opam/blob/28b4831943833f22ebc9cd68f9666d08c4c3953e/src/state/opamSwitchState.ml#L248) because even if we are using the rt level dependecies, we will still get the OS failure while polling for installed sys packages. 

I updated the OS failure handling in `OpamUpdate`, so the rt level depext availiabily fallback is `Suppose_available` for now (since it's just a placeholder that doesn't affect installs nor pins on unsupported OS)

src/state/opamSysInteract.ml
arozovyk: `compute_sets` is more broad name that matched previous logic (relevant packages + status computation), The rename is from the simplification of the function to just the first part (status is computed elsewhere now). 
I renamed `compute_sets_with_virtual` to `compute_installed_with_virtual` to be more specific and consistent

src/state/opamRepositoryState.ml
arozovyk: Moved both versions of the function to `OpamFileTools`

src/state/opamSwitchState.ml
arozovyk: Yes, because it's the initial depexts status computation/resolution on switch state load (also used for install). The `recompute_available` just serves to indicate that we have to also recompute the availability of system packages at repo level. 
Also what we have here is all currently pinned packages on this switch, not just newly pinned.

src/state/opamSysInteract.ml
arozovyk: I will have to recheck my tests on Altlinux

src/state/opamSwitchState.ml
```diff
@@ -259,7 +248,8 @@ let depexts_unavailable_raw sys_packages nv =
 
 let depexts st nv =
   let env v = OpamPackageVar.resolve_switch ~package:nv st v in
- depexts_raw ~env nv st.opams
+  let package_opam = OpamPackage.Map.find nv st.opams in
+ OpamFileTools.extract_depexts ~env package_opam
```
rjbou:
```suggestion
    OpamFileTools.extract_depexts ~env package_opam

File: src/state/opamFileTools.mli [hunk 44]
```diff
@@ -117,3 +117,10 @@ val dep_formula_to_string: formula -> string
 (** Sort opam fields: author, tags, depexts, depends, depopts, conflicts,
     pin_depends, extra_files, extra_sources *)
 val sort_opam: OpamFile.OPAM.t -> OpamFile.OPAM.t
+
+(** Extract depexts from a single opam file *)
+val extract_depexts : env:OpamFilter.env -> OpamFile.OPAM.t -> OpamSysPkg.Set.t
+
```
rjbou: Opinionated: there is no need for the 2 functions, `extract_depexts` can take a map and if there is a single element a singleton can be created. (there is few usages)

src/state/opamFileTools.mli
```diff
@@ -117,3 +117,10 @@ val dep_formula_to_string: formula -> string
 (** Sort opam fields: author, tags, depexts, depends, depopts, conflicts,
     pin_depends, extra_files, extra_sources *)
 val sort_opam: OpamFile.OPAM.t -> OpamFile.OPAM.t
+
+(** Extract depexts from a single opam file *)
+val extract_depexts : env:OpamFilter.env -> OpamFile.OPAM.t -> OpamSysPkg.Set.t
```
rjbou: `get_depexts`?

src/state/opamSwitchState.ml
```diff
@@ -280,14 +299,14 @@ let load lock_kind gt rt switch =
...
-    OpamPackage.Set.fold (fun nv (pinned,opams) ->
+  let pinned, pinned_opams, depexts_present_pins =
+    OpamPackage.Set.fold (fun nv (pinned,opams,depexpts_present) ->
         let overlay_dir =
           OpamPath.Switch.Overlay.package gt.root switch nv.name
         in
         match OpamFileTools.read_opam overlay_dir with
         | None -> (* No overlay => just pinned to a version *)
-          OpamPackage.Set.add nv pinned, opams
+          OpamPackage.Set.add nv pinned, opams, depexpts_present
```
rjbou:
```suggestion
          OpamPackage.Set.add nv pinned, opams, depexts_present
```

src/state/opamSwitchState.ml
```diff
@@ -304,9 +323,10 @@ let load lock_kind gt rt switch =
           let nv = OpamPackage.create nv.name version in
           let o = OpamFile.OPAM.with_version version o in
           OpamPackage.Set.add nv pinned,
-          OpamPackage.Map.add nv o opams
+          OpamPackage.Map.add nv o opams,
+          (OpamFile.OPAM.depexts o) <> [] || depexpts_present
```
rjbou: Here, you only check that the field is not empty. The field can be non empty, but resolved to the current setup the list is empty. It worth using `depexts`

src/state/opamSwitchState.ml
```diff
@@ -162,24 +162,31 @@ let infer_switch_invariant st =
...
+    { OpamSysPkg.s_available; s_not_found }
+  | OpamSysPkg.Suppose_available ->
+    (* Some package managers don't permit to request on available packages.
+       In this case, we consider all non installed packages as [available]. *)
+    let s_available = syspkg_set -- installed in
+    { OpamSysPkg.status_empty with s_available }
+
 let depexts_status_of_packages_raw
-    ~depexts ?env global_config switch_config packages =
+    repos_sys_available_pkgs ~depexts ?(recompute_available=false) ?env
```
rjbou:
```suggestion
    available_syspkgs ~depexts ?(recompute_available=false) ?env
```

src/state/opamSwitchState.ml
rjbou: It's in the load, I thought it was for new pins. On switch state loading, there is no need to have a specific case for pin. `depexts_status_of_packages_raw` is called with all available packages, that already contains pinned ones : 
```ocaml
  let available_packages =
    lazy (compute_available_and_pinned_packages gt switch switch_config
            ~pinned ~opams)
  in

File: src/state/opamRepositoryState.ml [hunk 49]
```diff
@@ -193,36 +196,56 @@ let load lock_kind gt =
...
+          repo_depexts)
+        repos_map (OpamRepositoryName.Map.empty,
+                   OpamRepositoryName.Map.empty,
+                   OpamSysPkg.Set.empty)
+    in
+    let repos_sys_available_pkgs =
+    try
+      OpamSysInteract.available_packages ~env:gt.global_variables
+        gt.config repo_depexts
+    with Failure msg ->
```
rjbou: To check, but I think that windows failure comes from the fact that even if opam is configured with no depexts, this is called

src/state/opamRepositoryState.ml
```diff
@@ -193,36 +196,56 @@ let load lock_kind gt =
...
+      OpamConsole.note "%s\nYou can disable this check using 'opam \
+                        option --global depext=false'"
+        msg;
+      Suppose_available
```
rjbou: For consistency, it should be an empty map

master_changes.md
```diff
@@ -114,6 +114,11 @@ users)
 ### Engine
  * Fix gcc < 14.3 bug on mingw i686 [#6624 @kit-ty-kate]
   * Fix support for removing local link directories [#6450 @kit-ty-kate]
+  * Use `opam-set-os` in reftests following the depexts update.
+
+### Engine
+  * `gawk` was re-added to the base fedora images [#6473 @kit-ty-kate]
```
rjbou: for last rebase : clean  changes

src/state/opamSwitchState.ml
arozovyk: My `ocp-indent` doesn't agree... 👀

src/state/opamSwitchState.ml
arozovyk: If I remove `~recompute_available:depexts_present_pins` it breaks the autopin mechanism. After some digging, it seems to me that the reason is the fact that `OpamSwitchState.load` is called _after_ the autopin that correctly calls `update_pin` updating the depext info. This info is lost during the switch state load, and I don't see how can I recover it without some major refactoring.

For now I propose an optimisation, recompute the depexts status on load only if they are present in pins and not already in the repository level cache : d1941d03397fec259c40973a349272278247ee58

arozovyk: Just to note the discussed tradeoff here: when a switch contains pins that have external dependencies, we will poll system packages availability on every switch load (`install`, `switch create .` etc) which is the current behaviour (oppisite to this PR). This is a tradoff that we accept for now as it should not affect the majority of users.

src/state/opamRepositoryState.ml
arozovyk: Added a `No_depexts` variant instead

src/state/opamSwitchState.ml
arozovyk: This is done

src/format/opamSysPkg.mli
```diff
@@ -28,6 +28,19 @@ val status_empty: status
 
 val string_of_status: status -> string
 
+val combine_status : status -> status -> status
```
rjbou:
```suggestion
(* Merge status sets *)
val combine_status : status -> status -> status
```

src/format/opamSysPkg.ml
```diff
@@ -67,6 +67,37 @@ let string_of_status sp =
...
+
+let combine_available a a' =
+  match a, a' with
+  | Available s, Available s' -> Available (Set.union s s')
+  | Suppose_available, Suppose_available -> Suppose_available
+  | Available s, Suppose_available | Suppose_available, Available s ->
+    Available s
+  | No_depexts, _  |  _, No_depexts ->
+    OpamConsole.error_and_exit `Bad_arguments
+      "Can't combine No_depexts"
```
rjbou:
```suggestion
  | No_depexts, _  |  _, No_depexts -> No_depexts
```

src/format/opamSysPkg.ml
```diff
@@ -67,6 +67,37 @@ let string_of_status sp =
...
+let string_of_available = function
+| Available av -> Set.to_string av
+| Suppose_available -> "Suppose available"
+| No_depexts -> "No depexts"
+
+let check_available_equal a b =
+  match a, b with
+  | Suppose_available, Suppose_available -> true
+  | Available os, Available ns -> Set.equal os ns
+  | _ -> false
```
rjbou:
```suggestion
  | Available os, Available ns -> Set.equal os ns
  | Suppose_available, Suppose_available
  | No_depexts, No_depexts -> true
  | Available _, _
  | Suppose_available, _
  | No_depexts, _ -> false
```

src/format/opamSysPkg.mli
```diff
@@ -28,6 +28,19 @@ val status_empty: status
...
+
+(** System package availability *)
+type available = Available of Set.t | Suppose_available | No_depexts
```
rjbou:
```suggestion
type available =
  (* Set of available system packages )
  | Available of Set.t
  (* In this system, all packages are considered available *)
  | Suppose_available
  (* Depext system disabled *)
  | No_depexts
```

rjbou: We may consider changing the name of the type to `availability`, wdyt ?

src/format/opamSysPkg.mli
```diff
@@ -28,6 +28,19 @@ val status_empty: status
...
+
+val string_of_available : available -> string
+
+val combine_available : available -> available -> available
```
rjbou:
```suggestion
(* Combine two available. If it is not the same variant, the result is 
   the smallest one. *)
val combine_available : available -> available -> available
```

src/state/opamRepositoryState.ml
```diff
@@ -195,36 +198,56 @@ let load lock_kind gt =
...
     OpamFilename.with_flock_upgrade `Lock_read lock @@ fun _ ->
-    let repofiles, opams =
-      OpamRepositoryName.Map.fold (fun name url (defs, opams) ->
+    let repofiles, opams, repo_depexts =
+      OpamRepositoryName.Map.fold (fun name url (defs, opams, all_depexts) ->
           let repo = mk_repo name url in
           let repo_def, repo_opams =
             load_repo repo (get_root_raw gt.root repos_tmp name)
           in
+          let repo_depexts = OpamFileTools.get_depexts repo_opams
```
```suggestion
          let repo_depexts =
            OpamFileTools.get_depexts repo_opams
```

src/state/opamRepositoryState.ml
```diff
@@ -195,36 +198,56 @@ let load lock_kind gt =
...
+                   OpamRepositoryName.Map.empty,
+                   OpamSysPkg.Set.empty)
+    in
+    let repos_sys_available_pkgs =
+      try
+        OpamSysInteract.available_packages ~env:gt.global_variables
+          gt.config repo_depexts
+      with Failure msg ->
+        OpamConsole.note "%s\nYou can disable this check using 'opam \
+                          option --global depext=false'"
```
```suggestion
      with Failure msg ->
        OpamConsole.note
          "%s\nYou can disable this check using \
           'opam option --global depext=false'"
```

src/state/opamSwitchState.ml
```diff
@@ -196,8 +205,25 @@ let depexts_status_of_packages_raw
...
   let syspkg_set = syspkg_set -- bypass in
+  let status syspkg_set =
+    let status =
+      resolve_depext_availability ?env global_config available_syspkgs
+        syspkg_set
+    in
+    if OpamPackage.Set.is_empty recompute_available then
+      status
+    else
+      let pinned =
```
rjbou:
```suggestion
      let pinned_status =
```
or
```suggestion
      let status_of_pinned =
```

src/state/opamSysInteract.mli
```diff
@@ -10,16 +10,40 @@
 
 open OpamStateTypes
 
+
```
rjbou, line 32:
```suggestion
```

src/state/opamSysInteract.mli
```diff
@@ -10,16 +10,40 @@
...
+    (e.g., on systems where querying is unsupported or skipped).
+
+    [env] is used to determine host specification.
+    [config] is used to determine Windows depext installation. *)
+val available_packages:
+  ?env:gt_variables -> OpamFile.Config.t -> OpamSysPkg.Set.t ->
+  OpamSysPkg.available
+
+(* Returns the subset of the given system [pkgs]
+   that are currently installed on the system. *)
```
rjbou: Maybe more verbose ?

src/state/opamUpdate.ml
```diff
@@ -186,14 +186,47 @@ let repositories rt repos =
       ~dry_run:OpamStateConfig.(!r.dryrun)
       repos
   in
+  let get_sys_available rt =
+    let opams = OpamRepositoryName.Map.fold (fun _ opams acc ->
+        OpamPackage.Map.union (fun _ x -> x) acc opams
+      ) rt.repo_opams (OpamPackage.Map.empty)
```
rjbou:
```suggestion
        OpamPackage.Map.union (fun _ x -> x) acc opams)
         rt.repo_opams (OpamPackage.Map.empty)
```

src/format/opamSysPkg.mli
arozovyk: renamed to `availability_mode` since it's about how we will act given certain type of availability

master_changes.md
```diff
@@ -98,6 +99,10 @@ users)
 
 ## Client
   * [NEW] Fetch shared archive sources without checksums [#6627 @psafont - fix #5638]
+  * `OpamSolution` Remove the heuristic of recomputing depexts of additional (pinned) packages. [#6489 @arozovyk fix #6489]
+  * `OpamClient` update the system package status check for dependencies during `opam install --deps-only`, including support for pinned packages; also update this in `OpamAuxCommands.autopin` [#6489 @arozovyk fix #6461]
+  * `OpamSolution.install_sys_packages_t` check for availability of system packages in `repo_state` before installing depexts [#6489 @arozovyk fix #6489]
+  * `OpamSolution.get_depexts` remove no longer needed `recover` option that was used with `--depext-only` option  [#6489 @arozovyk fix #6489]
```
kit-ty-kate: I'm not exactly sure why these are here. The `Client` section is meant for changes in the way the client works (high level changes, something that users can see), but these seem to be more meant for the `opam-client` API section at the bottom of the file.

However their description is a bit confusing to me as they mix very low level talk with slightly higher-level description that i personally fail to understand.

Could we get things split between API changes and higher-level description here?

src/state/opamFileTools.mli
```diff
@@ -117,3 +117,7 @@ val dep_formula_to_string: formula -> string
 (** Sort opam fields: author, tags, depexts, depends, depopts, conflicts,
     pin_depends, extra_files, extra_sources *)
 val sort_opam: OpamFile.OPAM.t -> OpamFile.OPAM.t
+
+(** Get depexts from all the opam files associated to packages *)
+val get_depexts : env:OpamFilter.env -> OpamFile.OPAM.t OpamPackage.Map.t ->
```
kit-ty-kate: i'm not too much of a fan of the `OpamFile.OPAM.t OpamPackage.Map.t ` parameter of this function.
If this function is meant to be used with the set of all accessible opam packages during the update like is done in `OpamUpdate`, then we instead want something like `OpamFile.OPAM.Set.t` since having a map indexed on packages can hide packages that are available in some switch but not others (e.g. you have two repository one is the default opam.ocaml.org, and the other some fork of it with overlapping packages, then you want a merge of both repositories in such a way that every depexts are listed, instead of only the depexts of packages of one of the two repository)

kit-ty-kate: Note that in practice it's fine, i've fixed the issue i'm describing in `OpamUpdate` directly, and really having `OpamFile.OPAM.Set` seems too complicated, but it's something not ideal that i'm just talking about out loud.

No need to change anything as far as i'm concerned.

src/state/opamSysInteract.ml
```diff
@@ -934,6 +920,251 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
...
+    (* From #5372 - modified: no limit to the query length and the header is not present *)
+    let sys_available =
+      let query = Printf.sprintf "/%s/" (names_re ()) in
+      log "Running homebrew query %s" query ;
+      run_query_command "brew" ["search"; "--formula"; query]
+      |> List.fold_left (fun acc x ->
+          OpamSysPkg.Set.add (OpamSysPkg.of_string x) acc)
+        OpamSysPkg.Set.empty
+    in
+    OpamSysPkg.Available (get_relevant sys_available)
```
kit-ty-kate:
```suggestion
  | Homebrew -> OpamSysPkg.Suppose_available
```
This should be in a separate PR

kit-ty-kate (commented):
I did a first quick pass and it looks great! The different commits' purpose are very clear now. I'll do a final pass and some live tests and performance evaluation on each platforms on Monday.

In the meantime i've pushed some typo fixes and other minor changes.

Could you also open a separate queued PR for the homebrew availability support that was sneakily added? This should be evaluated separately as opam currently doesn't support it and no notes or changelog item were added.

src/format/opamSysPkg.ml
```diff
@@ -67,6 +67,41 @@ let string_of_status sp =
...
+  Set.Op.{
+    s_available = st.s_available ++ st'.s_available;
+    s_not_found = st.s_available ++ st'.s_not_found;
+  }
+
+(* System package availability *)
+type availability_mode =
+  | Available of Set.t
+  | Suppose_available
+  | No_depexts
```
kit-ty-kate: Looking at the only places where this value is created (only when the checks raised an exception), i feel like this case is redundant with `Suppose_available` and doesn't add any value but confusion to the variant
```suggestion
```

arozovyk: I'll let @rjbou make the final call about this :)

src/state/opamSysInteract.ml
arozovyk: Indeed, my bad. 
Currently I don't have a clear explanation of why I had to change the code of https://github.com/ocaml/opam/pull/5372 in order for it to function with my homebrew, but I'll investigate it thoroughly before opening a PR

src/state/opamSysInteract.ml
```diff
@@ -679,35 +669,37 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
...
           | Some (pkg, _repo) -> pkg
           | None -> s)
       |> (fun l -> run_query_command "apk" ("policy"::l))
-      |> List.fold_left (fun (pkg, installed, instavail) l ->
+      |> List.fold_left (fun (pkg, installed, inst) l ->
           try (* package name *)
-            Re.(Group.get (exec pkg_name l) 1), false, instavail
+            Re.(Group.get (exec pkg_name l) 1), false, inst
           with Not_found ->
             if l.[2] <> ' ' then (* only version field is after two spaces *)
```
kit-ty-kate:
```suggestion
            if not (OpamCompat.String.starts_with ~prefix:"  " l) then (* only version field is after two spaces *)
```

src/state/opamSysInteract.ml
```diff
@@ -934,6 +919,241 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
...
+          match OpamStd.String.cut_at s '@' with
+          | Some (pkg, _repo) -> pkg
+          | None -> s)
+      |> (fun l -> run_query_command "apk" ("policy"::l))
+      |> List.fold_left (fun (pkg, avail) l ->
+          try
+            (* package name line *)
+            (Re.Group.get (Re.exec pkg_name l) 1, avail)
+          with Not_found ->
+            if OpamCompat.String.starts_with ~prefix:"  " l then
```
kit-ty-kate: Oops my bad (boolean flip)
```suggestion
            if not (OpamCompat.String.starts_with ~prefix:"  " l) then
```

kit-ty-kate (approved):
lgtm

tests/reftests/depexts.test
```diff
@@ -524,3 +524,136 @@ Switch invariant: ["comp"]
...
+### : <><> Creating initial switch 'comp' (invariant ["comp"]) ><><><><><><><><><><><>
+### : 
+### : <><> Installing new switch packages <><><><><><><><><><><><><><><><><><><><><><>
+### : Switch invariant: ["comp"]
+### : [NOTE] External dependency handling not supported for OS family 'unknown'.
+### :        You can disable this check using 'opam option --global depext=false'
+### : 
+### : <><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+### : -> installed comp.1
+### : Done.
```
rjbou: What is that comment ?

tests/reftests/depexts.test
```diff
@@ -524,3 +524,136 @@ Switch invariant: ["comp"]
...
+       You can disable this check using 'opam option --global depext=false'
+The following actions will be performed:
+=== upgrade 1 package
+  - upgrade unkn 1 to 2
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> removed   unkn.1
+-> installed unkn.2
+Done.
+### :VII: disabled depexts
```
rjbou, line 558:
```suggestion
### :VII: Disabled depexts

File: src/state/opamSysInteract.ml [hunk 71]
```diff
@@ -974,33 +1194,43 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
...
+       be found.' But omitting them will mean that they won't be
+       added to the Nix derivation.
+    *)
+    OpamSysPkg.Suppose_available
+  | Openbsd -> OpamSysPkg.Suppose_available
+
+let packages_status ?(env=OpamVariable.Map.empty) ?sys_available config syspkg_set =
+  let open OpamSysPkg.Set.Op in
+  log "computing depexts status for %a system packages (%s availability)"
+    (fun fmt x -> Format.pp_print_int fmt (OpamSysPkg.Set.cardinal x)) syspkg_set
```
rjbou: slog ?

master_changes.md
```diff
@@ -207,13 +207,16 @@ users)
...
+  * `OpamSysInteract`: add `available_packages` and `installed_packages` to be computed separately, redefine `packages_status` accordingly [#6489 @arozovyk]
 
 ## opam-solver
 
 ## opam-format
   * `OpamFormula.equal_relop`: was added [#6644 @kit-ty-kate]
   * `OpamTypesBase.{action,pkg_flag,simple_arg,arg,filter,command}_equal`: were added [#6644 @kit-ty-kate]
   * `OpamVariable.variable_contents_equal`: was added [#6644 @kit-ty-kate]
+  * `OpamSysPkg`: add `availability_mode` type to indicate the availability of system packages on a given system. [#6489 @arozovyk]
+  * `OpamSysPkg`: add `combine_status` `string_of_availability_mode` `combine_availability_mode` `check_availability_mode_equal` functions. [#6489 @arozovyk]
```
rjbou: these functions are added in this commit but not used in it, it's better to extract them to their own commit (or in the commit where they are used)

src/format/opamSysPkg.mli
```diff
@@ -28,6 +28,27 @@ val status_empty: status
...
+  | Suppose_available (** In this system, all packages are considered
+                          available *)
+  | No_depexts (** Depext system disabled *)
+
+val string_of_availability_mode : availability_mode -> string
+
+(** Combine two availability_modes. If it is not the same variant, the result is
+    the smallest one. *)
+val combine_availability_mode : availability_mode -> availability_mode ->
+  availability_mode
```
rjbou:
```suggestion
val combine_availability_mode :
  availability_mode -> availability_mode -> availability_mode

File: src/state/opamFileTools.mli [hunk 74]
```diff
@@ -117,3 +117,7 @@ val dep_formula_to_string: formula -> string
...
+  OpamSysPkg.Set.t
```
rjbou:
```suggestion
val get_depexts :
  env:OpamFilter.env -> OpamFile.OPAM.t OpamPackage.Map.t ->

File: master_changes.md [hunk 75]
```diff
@@ -157,6 +157,7 @@ users)
   * Add complete tests for the `pin-depends` feature [#6611 @rjbou]
   * Add more tests for depexts behaviour with unknown family types [#6489 @arozovyk]
   * Use the new `opam-set-os` command when applicable [#6489 @arozovyk]
+  * Add depexts tests with debug section that demostrate system availability polling [#6489 @arozovyk]
```
rjbou, line 148:
```suggestion
* Add depexts tests with debug section that demonstrate system availability polling [#6489 @arozovyk]

File: tests/reftests/depexts.test [hunk 76]
```diff
@@ -661,3 +661,21 @@ The following actions will be performed:
 -> installed other.dev
 -> installed foo.2.0
 Done.
+### opam option depext=true
+Set to 'true' the field depext in global configuration
+### :VIII: Debug output for depexts loading
```
rjbou, line 646:
```suggestion
### :VIII: Depexts polling

File: tests/reftests/reftests.test [hunk 77]
```diff
@@ -4,7 +4,10 @@ N0REP0
 ### ###################
 ### :I:1: opam root
 ### test -d OPAM
-### ls OPAM
+### : log/ may exist on Linux where depexts availability is queried during repo load,
+### : but not on macOS/Windows where we return Suppose_available without system calls
+### : so the log/ directory is not created.
```
rjbou, line 9:
```suggestion
### : but not on some tested OSes (macport, Windows) where we suppose that all 
### : packages are available without system call so the 'log/' directory is not
### : created.

File: src/state/opamUpdate.ml [hunk 78]
```diff
@@ -196,7 +196,10 @@ let get_sys_available rt =
       OpamSysPkg.No_depexts
 
 let update_sys_available_cache ?(force=false) rt =
+  if OpamConsole.disp_status_line () then
+    OpamConsole.status_line "Loading system package availability...";
```
rjbou, line 167: to squash with previous commit

src/state/opamUpdate.ml
```diff
@@ -162,6 +162,52 @@ let repository rt repo =
...
+    try
+      OpamSysInteract.available_packages ~env:rt.repos_global.global_variables
+        rt.repos_global.config repo_depexts
+    with Failure msg ->
+      OpamConsole.note "%s\nYou can disable this check using 'opam \
+                        option --global depext=false'"
+        msg;
+      OpamSysPkg.No_depexts
+
+let update_sys_available_cache ?(force=false) rt =
```
rjbou, line 165: This function shouldn't write repo config & update the cache. If we add something else in `rt`, we will need to extract that in the future. Plus, from `OpamCommands.update`, we want to write the cache if changed, not repo config.

src/state/opamUpdate.ml
```diff
@@ -189,59 +235,14 @@ let repositories rt repos =
       ~dry_run:OpamStateConfig.(!r.dryrun)
       repos
   in
-  let get_sys_available rt =
```
rjbou: This part was added in one of the first commits. Instead of moving piece of codes, it worth having this externalised functions from the beginning at its own place.

src/state/opamSysInteract.ml
```diff
@@ -974,33 +1194,43 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
...
+    (if Option.is_some sys_available then "using provided" else "querying system");
```
```suggestion
  log "Computing depexts status for %a system packages (%s availability)"
    (fun fmt x -> Format.pp_print_int fmt (OpamSysPkg.Set.cardinal x)) syspkg_set
    if (Option.is_some sys_available then "using stored" else "querying system");
```

src/format/opamSysPkg.mli
rjbou: this function is not used

tests/reftests/depexts.test
arozovyk: oops, (a copy from above), removed

src/state/opamUpdate.ml
```diff
@@ -162,6 +162,55 @@ let repository rt repo =
...
+            else
+              OpamFile.OPAM.with_depexts (depexts_x @ depexts_y) x)
+          acc opams)
+      rt.repo_opams OpamPackage.Map.empty
+  in
+  let repo_depexts =
+    OpamFileTools.get_depexts merged_depexts_opams
+      ~env:(OpamPackageVar.resolve_global rt.repos_global)
+  in
+  if not (OpamFile.Config.depext rt.repos_global.config) then
```
rjbou: Does these  `if not (OpamFile.Config.depext rt.repos_global.config) then` handle setting no depexts via cli or environnement variable ?

src/state/opamUpdate.ml
```diff
@@ -162,6 +162,52 @@ let repository rt repo =
...
```
rjbou: This if should be at the  beginning, no need to compute other values if we return no depexts

src/state/opamUpdate.ml
arozovyk: Via cli https://github.com/ocaml/opam/pull/6489/files#diff-6b542a62b86285344c7cdec7b0403775193ced334b88b2c8f1bb1bd0d6044e43R98

src/state/opamUpdate.ml
arozovyk: Done

rjbou:
Discussed on dev meeting: we need to store the os-family in the repository state cache in order to be able to invalidate it if the distribution/os/os-family changes (via config, environment variable, `opam var`, etc.).
This should make reftest's command `opam-set-os` no longer do an update. We keep #6741 as is (no revert), it's now a mainly a shortcut.

src/client/opamClient.ml
```diff
@@ -1520,7 +1520,9 @@ let determine_windows_configuration ?cygwin_setup ?git_location
         `Internal, pkgs
       | (`Root _ | `Path _) as mechanism ->
         let cygwin_packages =
-          if cygwin_is_tweakable && not OpamStateConfig.(!r.no_depexts) then
+          if cygwin_is_tweakable
+          && (not OpamStateConfig.(!r.no_depexts)
+              || OpamFile.Config.depext config) then
```
kit-ty-kate: Wouldn't we want this?
```suggestion
          && not (OpamStateConfig.no_depext gt) then
```

src/core/opamSystem.mli
```diff
@@ -56,6 +56,13 @@ val verbose_for_base_commands: unit -> bool
 
 (** {2 Filesystem management} *)
 
+(** Check if path is read only.
+    If [path] is a directory, it creates a file in that directory, otherwise,
+    it uses given filename. Then tries to create the file with write
+    permissions, and catch the read-only error ({Unix.EROFS}).
+    Error if the path doesn't exist. *)
+val is_read_only: string -> bool
```
kit-ty-kate:
```suggestion
val is_dir_read_only: string -> bool
```
otherwise it's a bit hard to understand what this function does by its name

src/core/opamSystem.ml
```diff
@@ -188,6 +188,21 @@ let remove_dir dir =
...
 
+let is_read_only path =
+  try
+    let f =
+      if is_reg_dir path then
+        path ^ Filename.dir_sep ^ temp_basename "ro-test"
+      else path
+    in
+    let flags = [Unix.O_WRONLY; Unix.O_CREAT; Unix.O_APPEND; Unix.O_SHARE_DELETE] in
+    let fd = Unix.openfile f flags 0o644 in
```
kit-ty-kate: what about
```suggestion
    Unix.access f [W_OK];
```
instead?

src/state/opamGlobalState.ml
```diff
@@ -229,3 +229,7 @@ let fix_switch_list gt =
       write gt, gt
     with OpamSystem.Locked -> gt
   else gt
+
+let is_root_read_only =
+  if Sys.win32 then fun _ -> false else
+    fun gt -> OpamFilename.is_dir_read_only gt.root
```
kit-ty-kate: I think it's best to handle Windows even if there isn't any sandbox yet. This way there aren't any weird obscure code to think about changing whenever we do add sandbox support to Windows
```suggestion
let is_root_read_only gt =
  OpamFilename.is_dir_read_only gt.root
```

tests/reftests/legacy-git.test
```diff
@@ -846,18 +846,18 @@ ocaml.system
 The following actions will be performed:
```
kit-ty-kate, line 846: i feels to me that the changes from legacy-git and legacy-local are simply output regenerations of the ` | unordered` tests. It would be nice to avoid the unnecessary diff in the final state of this PR

tests/reftests/no-depexts-sandboxed.unix.test
```diff
@@ -0,0 +1,146 @@
...
+### :I: check in-commands opam calls & depexts
+### <pkg:foo.1>
+opam-version: "2.0"
+build: [ "sh" "build.sh" ]
+### <pkg:bar.1>
+opam-version: "2.0"
+depexts: [ "xbar" ]
+### <build.sh>
+
+OPAM=$(cygpath ${OPAM} 2>/dev/null || echo ${OPAM})
```
kit-ty-kate:
```suggestion
OPAM=$(cygpath "${OPAM}" 2>/dev/null || echo "${OPAM}")
```

tests/reftests/no-depexts-sandboxed.unix.test
```diff
@@ -0,0 +1,146 @@
...
+echo ---
+echo +++ BIN
+${OPAM} var bin
```
kit-ty-kate:
```suggestion
"${OPAM}" var bin
```

tests/reftests/no-depexts-sandboxed.unix.test
```diff
@@ -0,0 +1,146 @@
...
+echo +++ BAR/lib
+${OPAM} var bar:lib
```
kit-ty-kate:
```suggestion
"${OPAM}" var bar:lib
```

src/client/opamClient.ml
rjbou: We don't have `gt` at this stage. But i've added (in the incoming version of the PR) a `OpamStateConfig.no_depexts_raw` that can be used only with a config file.

src/core/opamSystem.mli
rjbou: I intentionally kept it no dir specific (takes dir and file), but we can have it work only with dirs.

tests/reftests/legacy-git.test
rjbou: yes!

src/core/opamSystem.ml
rjbou: go for it!

tests/reftests/depexts.test
```diff
@@ -524,3 +526,201 @@ Switch invariant: ["comp"]
...
+### : <><> Creating initial switch 'comp' (invariant ["comp"]) ><><><><><><><><><><><>
+### : 
+### : <><> Installing new switch packages <><><><><><><><><><><><><><><><><><><><><><>
+### : Switch invariant: ["comp"]
+### : [NOTE] External dependency handling not supported for OS family 'unknown'.
+### :        You can disable this check using 'opam option --global depext=false'
+### : 
+### : <><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+### : -> installed comp.1
+### : Done.
```
kit-ty-kate: why add comments like this?

src/state/opamSysInteract.ml
```diff
@@ -213,6 +191,93 @@ let family ~env () =
...
+    let equal_elem q q' =
+      match q, q' with
+      | `all, `all
+      | `none, `none
+        -> true
+      | `set s, `set s' ->
+        OpamSysPkg.Set.equal s s'
+      | (`all | `none | `set _), _ -> false
+    in
+    t.osd_install && t'.osd_install
```
kit-ty-kate, line 233:
```suggestion
    Bool.equal t.osd_install t'.osd_install
```

tests/reftests/depext-only.test
```diff
@@ -87,11 +87,22 @@ The following system packages will first need to be installed:
...
 Try 'opam install --help' or 'opam --help' for more information.
 # Return code 2 #
+### opam install foo --depext-only --no-depexts
+opam: --depext-only and --assume-depexts can't be used together
+Usage: opam install [OPTION]… [PACKAGES]…
+Try 'opam install --help' or 'opam --help' for more information.
+# Return code 2 #
+### opam option depext=false
+Set to 'false' the field depext in global configuration
+### opam install foo --depext-only
```
kit-ty-kate: I'm not sure i agree this shouldn't do anything. I can see a case for disabling depext by default but selectively installing them when needed.

kit-ty-kate: If it's too complicated to do with the new system though, we should at least display an error message saying to set `opam option depext=true` and exit <> 0, but doing it automatically for the duration of the command would be best i think.

tests/reftests/depexts.test
```diff
@@ -524,3 +526,201 @@ Switch invariant: ["comp"]
...
+[NOTE] Package other does not exist in opam repositories registered in the current switch.
+other is now pinned to file://${BASEDIR}/other (version dev)
+The following actions will be performed:
+=== install 1 package
+  - install other dev (pinned)
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> retrieved other.dev  (file://${BASEDIR}/other)
+-> installed other.dev
+[NOTE] Requirement for system package other-depext overridden in this switch. Use `opam option 'depext-bypass-="other-depext"'' to revert.
```
kit-ty-kate:
```suggestion
[NOTE] Requirement for system package other-depext overridden in this switch. Use `opam option 'depext-bypass-="other-depext"'` to revert.
```

tests/reftests/depext-only.test
rjbou: it's easily doable with
```diff
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -1875,6 +1875,11 @@ let install cli =
                  else "no-depexts"))
     else
     OpamGlobalState.with_ `Lock_none @@ fun gt ->
+    if depext_only && OpamStateConfig.no_depexts gt then
+      `Error (true,
+              Printf.sprintf "Depexts mechanism is disabled, \
+                              --depext-only can't be used.")
+    else
     OpamSwitchState.with_ `Lock_write gt @@ fun st ->
     let pure_atoms =
       List.filter_map (function `Atom a -> Some a | _ -> None)

File: tests/reftests/depexts.test (hunk 92 above)
Comment (rjbou) [reply tests/reftests/depexts.test]: good question, @arozovyk ?

Comment (arozovyk) [inline tests/reftests/depexts.test]: I kept it for reference at some point while enabling/disabling the main behaviour of polling depexts on update/init

It's a developement artifact at this point and should be removed 

```suggestion
```

tests/reftests/depext-only.test
```diff
@@ -87,11 +87,34 @@ The following system packages will first need to be installed:
 + echo "sys-foo"
 - sys-foo
 ### :I:5: No depext
+### opam switch create no-depext --empty
 ### OPAMNODEPEXTS=1 opam install foo --depext-only
```
kit-ty-kate:
```suggestion
### OPAMNODEPEXTS=1 opam install foo --depext-only | sed-cmd echo
```

Issue #6448, opened 2025-04-01
opam hangs when updating repositories


mtelvers:
`opam-repo-ci` uses the Docker base images. The Docker base images build `opam` from the `master` branch each week. This week, we are seeing many job timeouts.  The jobs hang at the point where they run `opam init --reinit --ni`.

```
 => [ 4/11] RUN opam init --reinit -ni                                                                                                                                                  
 => => # You may want to back it up before going further.                                                                                                                                      
 => => # Continue? [Y/n] y                                                                                                                                                                     
 => => # [NOTE] The 'jobs' option was reset, its value was 39 and its new value will vary according to the current number of cores on your machine. You can restore the fixed value using:     
 => => #            opam option jobs=39 --global                                                                                                                                               
 => => # Format upgrade done.                                                                                                                                                                  
 => => # <><> Updating repositories ><><><><><><><><><><><><><><><><><><><><><><><><><><>                                                                                                      
```

Typically, this step takes < 2 minutes, but now takes > 2 hours.

Steps to reproduce.  Paste the following into the shell.  Change `opam-2.3` to `opam-dev` to see the difference.  On my machine, with `opam-2.3` the build was completed in ~4 minutes.  With `opam-dev` it has been running for nearly two hours

```
cd $(mktemp -d)
git clone --recursive "https://github.com/ocaml/opam-repository.git" && cd "opam-repository" && git fetch origin "refs/pull/27696/head" && git reset --hard 46b8cc5a
git fetch origin master
git merge --no-edit 4d8fa0fb8fce3b6c8b06f29ebcfa844c292d4f3e
cat > ../Dockerfile <<'END-OF-DOCKERFILE'
FROM ocaml/opam:debian-12-ocaml-4.09@sha256:13bd7f0979922adb13049eecc387d65d7846a3058f7dd6509738933e88bc8d4a
USER 1000:1000
WORKDIR /home/opam
RUN sudo ln -f /usr/bin/opam-2.3 /usr/bin/opam
RUN opam init --reinit -ni
RUN opam option solver=builtin-0install && opam config report
ENV OPAMDOWNLOADJOBS="1"
ENV OPAMERRLOGLEN="0"
ENV OPAMPRECISETRACKING="1"
ENV CI="true"
ENV OPAM_REPO_CI="true"
RUN rm -rf opam-repository/
COPY --chown=1000:1000 . opam-repository/
RUN opam repository set-url --strict default opam-repository/
RUN opam update --depexts || true
[...]

PR #6490, opened 2025-04-25
Fix patch/diff: add support for empty files, removing empty directories, large patches and move to a new directory

kit-ty-kate:
Requires https://github.com/hannesm/patch/pull/23
Requires https://github.com/hannesm/patch/pull/26
Fixes https://github.com/ocaml/opam/issues/6448


src/core/opamSystem.ml
```diff
@@ -130,6 +130,17 @@ let get_files dirname =
...
+  let dir = Unix.opendir dirname in
+  Fun.protect ~finally:(fun () -> Unix.closedir dir) @@ fun () ->
+  let rec loop () =
+    match Unix.readdir dir with
+    | "." | ".." -> loop ()
+    | _ -> false
+    | exception End_of_file -> true
+  in
+  loop ()
+
```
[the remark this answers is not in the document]
kit-ty-kate: ah i didn't notice this function. However reading its definition i'm worried as to why it would ignore failures and even return that an unavailable directory is not empty

src/core/opamSystem.ml
```diff
@@ -185,6 +196,12 @@ let remove_dir dir =
       remove_file_t ~with_log:true dir
   end
 
+let rec remove_empty_dir_recursively dir =
+  if directory_is_empty dir then begin
+    Unix.rmdir dir;
+    remove_empty_dir_recursively (Filename.dirname dir)
+  end
```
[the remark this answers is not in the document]
kit-ty-kate: sounds reasonable, will do.

src/core/opamSystem.ml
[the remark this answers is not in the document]
rjbou: It's usage needs that for simplicity. It can be updated though and its documentation added.

src_ext/Makefile.sources
```diff
@@ -60,5 +60,5 @@ MD5_swhid_core = 77d88d4b1d96261c866f140c64d89af8
 URL_menhir = https://gitlab.inria.fr/fpottier/menhir/-/archive/20240715/archive.tar.gz
 MD5_menhir = d39a8943fe1be28199e5ec1f4133504c
 
-URL_patch = https://github.com/hannesm/patch/releases/download/v3.0.0-alpha1/patch-3.0.0-alpha1.tar.gz
-MD5_patch = 03aa87f8500c9caf4a73b2299c19b514
+URL_patch = https://github.com/kit-ty-kate/patch/archive/empty-delete-create.tar.gz
+MD5_patch = 97188452f23aaed44e21b022886ce973
```
kit-ty-kate, line 64:
```suggestion
URL_patch = https://github.com/hannesm/patch/releases/download/v3.0.0-alpha2/patch-3.0.0-alpha2.tar.gz
MD5_patch = 7f11023c7231b916cfe3dd28ff6ce948
```

src/core/opamSystem.ml
```diff
@@ -389,8 +389,25 @@ let dirs dir =
   directories_with_links dir
 
 let dir_is_empty dir =
-  try in_dir dir (fun () -> Sys.readdir (Sys.getcwd ()) = [||])
-  with File_not_found _ -> false
+  try
+    let dir = Unix.opendir dir in
+    Fun.protect ~finally:(fun () -> Unix.closedir dir) @@ fun () ->
+    let rec loop () =
+      match Unix.readdir dir with
```
rjbou, line 396: Is there a reason to use `Unix.readdir` isntead of the previous version with Sys.readdir ? is it to retrieve all errors ?

src/core/opamSystem.mli
```diff
@@ -174,7 +176,7 @@ val rec_dirs: string -> string list
 (** Return the list of directories in the current directory. *)
 val dirs: string -> string list
 
-val dir_is_empty: string -> bool
+val dir_is_empty: string -> bool option
```
rjbou, line 179: It worth having some doc
```suggestion
(* Return [Some true] if the directory exists and is empty,
   [Some false] if non empty, [None] if the directory doesn't exist,
   or is unreachable. *)
val dir_is_empty: string -> bool option
```

src/core/opamSystem.mli
```diff
@@ -141,6 +141,8 @@ val remove_file: string -> unit
     directory (not for symlinks or other files). *)
 val remove_dir: string -> unit
 
+val rmdir_cleanup : string -> unit
```
rjbou, line 144:
```suggestion
(** Removes an empty directory, as well as any empty leading path components.
    Must be called only on a directory that is known to not have empty parents,
    only internal opam directory (and not tmp dir). *)
val rmdir_cleanup : string -> unit
```
Copied from Filename

src/core/opamSystem.ml
kit-ty-kate: Yes, it's much faster. The previous version used 2 system calls + a combination of several system calls + the allocation of an array + a loop to set the elements of the array.

The `Unix.readdir` version simply do 3 system calls: 1 opendir, 1 readdir, 1 closedir

rjbou: ok! :+1:

Issue #5814, opened 2024-01-31
opam recompiles unchanged package on upgrade


Khady:
We have been seeing a weird behavior of opam, both in CI and on dev machines, where a package that hasn't seen a change in months gets recompiled once in a while by opam. It is hard to investigate, we don't see much in opam logs, even when increasing the verbosity. And we can't systematically reproduce.

The message we get looks like this. Here it is on `opam upgrade --fixup` but I believe that it happens on basic `opam upgrade` too.

```
[06:14:31 #] opam upgrade --fixup
The following actions will be performed:
  ↻ recompile ocurl                           20231103+git
          [upstream or system changes]
```

The package definition is like this

```
opam-version: "2.0"
name: "ocurl"
maintainer: "ygrek@autistici.org"
homepage: "https://ygrek.org/p/ocurl"
license: "MIT"
authors: [ "Lars Nilsson" "ygrek" ]
dev-repo: "git+https://github.com/ygrek/ocurl.git"
bug-reports: "https://github.com/ygrek/ocurl/issues"
build: [
]
depends: [
  "curl" {= version}
]
synopsis: "Transitional dummy package: ocurl -> curl"
description: "Use `opam install curl`"
```

Our repository is a local repo (with the `file:///` prefix).

It happens at least with opam 2.2.0~alpha3 and 2.1.2.

This is one of the reasons behind our request for #5776.

kit-ty-kate:
How often does this happen? Is it linked to a call to `opam update`?

Khady:
We see it every day. But we runs thousands of builds per day. So it might be something uncommon which appears at scale. I don’t have the exact frequency at which it is triggered.

yes we always run `opam update` before to run those `opam upgrade`. I’m not aware of `opam upgrade` having this behaviour without an update before hand. That could be the case, but it’s not something that we do.

kit-ty-kate:
> and on dev machines

Is this on a bare metal file system or in docker? There is a known issue with timestamps in docker and it might be that

Khady:
Bare file system

kit-ty-kate:
We managed to debug this. The culprit was the `name` field which is not defined in the `<switch>/.opam-switch/packages/ocurl.20231103+git` but is defined in the original opam file, thus the `OpamFile.OPAM.effectively_equal` function detects it is different.

I'll have a look at fixing that soon.

kit-ty-kate:
[...]

PR #6505, opened 2025-05-03
Fix erroneous recompilations on files with empty but present build/runtest/install/remove field

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/5814
See description in https://github.com/ocaml/opam/issues/5814


tests/reftests/upgrade.test
```diff
@@ -510,3 +510,39 @@ The following actions would be performed:
...
 === upgrade 1 package
   - upgrade   foo 3 to 4
+### :::::::::::::::::::::::
+### : Make sure packages with empty-but-present build/run-test/install/remove
+### : don't recompile unnecessarily
+### :::::::::::::::::::::::
+### opam switch create no-recompile --empty
+### <pkg:ocaml.5>
+opam-version: "2.0"
+### :WARN: Do NOT use "<pin:...>" for the following line
```
kit-ty-kate, line 520: See https://github.com/ocaml/opam-file-format/issues/63

PR #6515, opened 2025-05-10
A few performance improvements

dra27:
Our version comparison function is very hot. While looking around it in the context of #4245, I picked up on a few easy wins. These are some basic results from hyperfine on my system for `opam show dune`, but they must be taken with an appropriate amount of salt:

```
OCaml 5.4.0
Base:    902.6 ms ±  32.7 ms
ramp_up: 553.4 ms ±  48.5 ms
closure: 539.6 ms ±  26.9 ms
files:   378.0 ms ±  14.5 ms

OCaml 5.3.0
Base:    913.2 ms ±  46.4 ms
ramp_up: 878.5 ms ±  36.5 ms
closure: 866.5 ms ±  34.7 ms
files:   756.4 ms ±  38.7 ms

OCaml 4.14.2
Base:    1.090 s ±  0.030 s
ramp_up: 1.114 s ±  0.053 s
closure: 1.108 s ±  0.059 s
files:   842.2 ms ±  26.9 ms
```

The interesting part is that on OCaml 5.4, `opam show` spends almost its entire time unmarshalling the repository state cache. However, these changes shouldn't be assessed in terms of performance _improvement_ because what's happening on all versions of OCaml is a reduction in GC pressure (i.e. if opam were doing a bit more work, the GC would almost certainly run again and the time reduction wouldn't be as stark). The three changes are:

- Use OCaml 5.4's `Gc.ramp_up` for the unmarshalling call in `OpamCached`. This is logically sensible, as the caches are never collected. The speedup comes from the fact that the major GC doesn't actually have to run any more in the `opam show` command, and the 5.3.0 and 4.14.2 results for that commit are well within the noise of the microbenchmark.
- `OpamStd.String.compare_case` added in #4328 contained a small mistake - the closure allocation for the `aux` function shows up in flame graphs. The derivation of sets of packages from the map of opams in OpamRepositoryState (which is also done in OpamSwitchState) is mostly dominated by `OpamVersionCompare`, but eliminating the closure in `compare_case` has a benefit (as it happens, I also got a small benefit from a Hacker's Delight-style optimisation noting that opam package names can't contain certain characters, but as the function is also used for system package names, I decided not to commit that one!). The diff looks noisy without space-change being ignored - all I've done is move the `aux` function out of `compare_case`.
- ~Finally, I noticed with `opam show dune` that the switch selections of the switches are read in proportion to the number of versions being displayed. That creates both unnecessary I/O and GC pressure.~ extracted in #6818

I think all three of these changes are worth putting in, but I'd countenance against hyping the performance improvement on `opam show` because it won't always show up that way!

### DIFF
--- a/src/core/opamVersionCompare.ml
+++ b/src/core/opamVersionCompare.ml
@@ -25,18 +25,6 @@ let rec skip_while_from i f w m =
   if (i : int) = m then i
   else if f w.[i] then skip_while_from (i + 1) f w m else i
 
-(* splits a version into (prefix,revision). The revision starts on the
- * right-most occurrence of '-', or is empty in case the version does
- * not contain '-'.  *)
-let extract_revision x =
-  try
-    let di = String.rindex x '-' in
-    let before = String.sub x 0 di in
-    let after = String.sub x (di+1) (String.length x - di -1) in
-    (before,after)
-  with
-    | Not_found -> (x,"")
-
 (* character comparison uses a modified character ordering: '~' first,
    then letters, then anything else *)
 let compare_chars c1 c2 = match c1 with
@@ -59,12 +47,7 @@ let skip_zeros x xi xl = skip_while_from xi (fun c -> c = '0') x xl
 (* compare versions chunks, that is parts of version strings that are
  * epoch, upstream version, or revisision. Alternates string comparison
  * and numerical comparison.  *)
-let compare_chunks x y =
-  (* x and y may be empty *)
-  let xl = String.length x
-  and yl = String.length y
-  in
-  let rec loop_lexical xi yi =
+  let rec loop_lexical xl yl x y xi yi =
     assert ((xi : int) <= xl && (yi : int) <= yl);
     match ((xi : int) = xl, (yi : int) = yl) with (* which of x and y is exhausted? *)
       | true,true -> 0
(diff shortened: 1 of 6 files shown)


src/core/opamVersionCompare.ml
```diff
@@ -106,38 +89,43 @@ let compare_chunks x y =
...
-    and (u2,r2) = extract_revision y in
-    let u_comp = compare_chunks u1 u2 in
+    let lx = String.length x in
+    let ly = String.length y in
+    let rx =
+      try String.rindex x '-'
+      with Not_found -> lx in
+    let ry =
+      try String.rindex y '-'
+      with Not_found -> ly in
```
kit-ty-kate, line 125: Per this discussion https://github.com/ocaml/opam/pull/6576, i'm wondering if reimplementing `String.rindex` to something that returns `-1` instead of the exception might be a little faster given the apparent (small but there) cost of exceptions, given the hotness of this code path.

kit-ty-kate: I tried it and found no improvement at all

kit-ty-kate (approved):

src/core/opamVersionCompare.ml
```diff
@@ -106,38 +89,46 @@ let compare_chunks x y =
...
+    let u_comp = loop_lexical rx ry x y 0 0 in
+    if u_comp <> 0 then
+      normalize_comp_result u_comp
+    else
+      normalize_comp_result
+        (loop_lexical lx ly x y rx ry)
```
dra27: @kit-ty-kate You made this change when you rebased it:
```suggestion
        (loop_lexical lx ly x y
           (OpamCompat.Int.min (rx + 1) lx) (OpamCompat.Int.min (ry + 1) ly))
```
How come? It's not necessary for correctness, I don't think? (it's slower, albeit only very slightly)

dra27: Oh dear me - brain switched on this time! Faster without this change, and unfortunately wronger, of course...

rjbou (approved):
I've rebased the PR, extracted `opam show` improvement in its own PR (to keep this PR for core changes), and split commits that concerns functions additions.
LGTM!

Issue #6501, opened 2025-04-30
opam install . --deps-only shows "nothing to do" when a package of the same name and version is already installed but not pinned


kit-ty-kate:
This is yet another bug similar to https://github.com/ocaml/opam/issues/5567 and all the similar issues, and was supposed to be fixed by https://github.com/ocaml/opam/pull/6209 as part of `2.4.0~alpha1`:
However it seems that there are still some edge-cases and we forgot to test for them:
```
$ opam install -t --deps .
$ cat *.opam
opam-version: "2.0"
version: "3.0.0~alpha1"
maintainer: "Hannes Mehnert <hannes@mehnert.org>"
authors: [
  "Hannes Mehnert <hannes@mehnert.org>"
  "Kate <kit-ty-kate@outlook.com>"
]
homepage: "https://github.com/hannesm/patch"
doc: "https://hannesm.github.io/patch/"
dev-repo: "git+https://github.com/hannesm/patch.git"
bug-reports: "https://github.com/hannesm/patch/issues"
license: "ISC"

depends: [
  "ocaml" {>= "4.08"}
  "dune" {>= "3.0"}
  "alcotest" {with-test & >= "1.7.0"}
  "crowbar" {with-test}
]
build: [
  ["dune" "subst"] {dev}
  ["dune" "build" "-p" name "-j" jobs]
  ["dune" "runtest" "-p" name "-j" jobs] {with-test}
]
synopsis: "Patch library purely in OCaml"
description: """
This is a library which parses unified diff and git diff output, and can
apply a patch in memory.
"""
x-maintenance-intent: [ "(latest)" ]
$ opam install alcotest
The following actions will be performed:
=== recompile 2 packages
  ↻ lambda-term        3.3.2  [uses logs]
  ↻ logs               0.8.0  [uses fmt]
=== install 4 packages
  ∗ alcotest           1.9.0
  ∗ astring            0.8.5  [required by alcotest]
  ∗ fmt                0.10.0 [required by alcotest]
  ∗ ocaml-syntax-shims 1.0.0  [required by alcotest]

Proceed with ↻ 2 recompilations and ∗ 4 installations? [Y/n] n
$ opam list | grep patch
patch                3.0.0~alpha1 Patch library purely in OCaml
$ vim patch.opam
$ opam install -t --deps .
The following actions will be performed:
=== remove 1 package
  ⊘ patch              3.0.0~alpha1 [no longer available]
=== recompile 2 packages
[...]

PR #6522, opened 2025-05-23
Fix simulated pinned packages reinstall trigger

rjbou:
partial fix of #6501

### DIFF
--- a/src/client/opamAuxCommands.ml
+++ b/src/client/opamAuxCommands.ml
@@ -416,25 +416,14 @@ let simulate_local_pinnings ?quiet ?(for_view=false) st to_pin =
     );
     reinstall = lazy (
       let open OpamPackage.Set.Op in
-      let installed_pinned = st.pinned %% st.installed in
-      OpamPackage.Set.fold (fun pinned_pkg reinstall ->
-          match
-            OpamPackage.Set.find_opt
-              (fun pkg ->
-                 OpamPackage.Name.equal
-                   (OpamPackage.name pinned_pkg)
-                   (OpamPackage.name pkg))
-              local_packages
-          with
-          | None -> reinstall
-          | Some local_pkg ->
-            let old_opam = OpamPackage.Map.find pinned_pkg st.installed_opams in
-            let new_opam = OpamPackage.Map.find local_pkg local_opams in
-            if OpamFile.OPAM.effectively_equal old_opam new_opam then
-              reinstall
-            else
-              OpamPackage.Set.add local_pkg
-                (OpamPackage.Set.remove pinned_pkg reinstall))
+      let installed_pinned = local_packages %% st.installed in
+      OpamPackage.Set.fold (fun pkg reinstall ->
+          let old_opam = OpamPackage.Map.find pkg st.installed_opams in
+          let new_opam = OpamPackage.Map.find pkg local_opams in
+          if OpamFile.OPAM.effectively_equal old_opam new_opam then
+            reinstall
+          else
+            OpamPackage.Set.add pkg reinstall)
         installed_pinned (Lazy.force st.reinstall)
     );
     pinned;
--- a/tests/reftests/simulated-pin.test
+++ b/tests/reftests/simulated-pin.test
@@ -59,6 +59,7 @@ Done.
 opam-version: "2.0"
 depends: "dep-reinst" { = "1" }
 ### opam install ./reinst --deps-only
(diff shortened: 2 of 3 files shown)


src/client/opamAuxCommands.ml
```diff
@@ -407,25 +407,15 @@ let simulate_local_pinnings ?quiet ?(for_view=false) st to_pin =
...
-                (OpamPackage.Set.remove pinned_pkg reinstall))
+      let installed_pinned = local_packages %% st.installed in
+      OpamPackage.Set.fold (fun pkg reinstall ->
+          let old_opam = OpamPackage.Map.find pkg st.installed_opams in
+          let new_opam = OpamPackage.Map.find pkg local_opams in
+          if OpamFile.OPAM.effectively_equal old_opam new_opam then
+            reinstall
+          else
+            OpamPackage.Set.add pkg
+              (OpamPackage.Set.remove pkg reinstall))
```
kit-ty-kate, line 418:
```suggestion
            OpamPackage.Set.add pkg reinstall)
```

tests/reftests/pin.test
```diff
@@ -1427,3 +1427,258 @@ Pin and install them? [Y/n] y
...
+opam-version: "2.0"
+depends: "dep-reinst" { = "2" }
+### opam install ./reinst --deps-only
+[reinst.1] synchronised (file://${BASEDIR}/reinst)
+[reinst] Installing new package description from upstream file://${BASEDIR}/reinst
+The following actions will be performed:
+=== recompile 1 package
+  - recompile reinst     1 (pinned) [uses dep-reinst]
+=== upgrade 1 package
+  - upgrade   dep-reinst 1 to 2     [required by reinst]
```
kit-ty-kate, line 1503: This should be rejected because the dependency constraints from the pinned package and the requested deps are conflicting.

tests/reftests/pin.test
```diff
@@ -1427,3 +1427,258 @@ Pin and install them? [Y/n] y
...
+-> installed dep-reinst.1
+-> retrieved reinst.1  (file://${BASEDIR}/reinst)
+-> installed reinst.1
+Done.
+### <pin:reinst/reinst.opam>
+opam-version: "2.0"
+depends: "dep-reinst" { = "2" }
+### opam install ./reinst --deps-only
+[reinst.1] synchronised (file://${BASEDIR}/reinst)
+[reinst] Installing new package description from upstream file://${BASEDIR}/reinst
```
kit-ty-kate, line 1498: Similarly, this should not happen as we are trying to install only the dependency of that package, not the package itself.

kit-ty-kate (changes requested):

tests/reftests/simulated-pin.test
```diff
@@ -0,0 +1,274 @@
...
+opam-version: "2.0"
+depends: "dep-reinst" { = "2" }
+### opam install ./reinst --deps-only
+[reinst.1] synchronised (file://${BASEDIR}/reinst)
+[reinst] Installing new package description from upstream file://${BASEDIR}/reinst
+The following actions will be performed:
+=== recompile 1 package
+  - recompile reinst     1 (pinned) [uses dep-reinst]
+=== upgrade 1 package
+  - upgrade   dep-reinst 1 to 2     [required by reinst]
```
rjbou: cf https://github.com/ocaml/opam/pull/6522#discussion_r2108984515
>  This should be rejected because the dependency constraints from the pinned package and the requested deps are conflicting.

Issue #6529 & fixed #6530

tests/reftests/simulated-pin.test
```diff
@@ -0,0 +1,274 @@
...
+-> installed dep-reinst.1
+-> retrieved reinst.1  (file://${BASEDIR}/reinst)
+-> installed reinst.1
+Done.
+### <pin:reinst/reinst.opam>
+opam-version: "2.0"
+depends: "dep-reinst" { = "2" }
+### opam install ./reinst --deps-only
+[reinst.1] synchronised (file://${BASEDIR}/reinst)
+[reinst] Installing new package description from upstream file://${BASEDIR}/reinst
```
rjbou: cf comment https://github.com/ocaml/opam/pull/6522#discussion_r2108985491
> Similarly, this should not happen as we are trying to install only the dependency of that package, not the package itself.

Issue #6532 & fixed in #6533

kit-ty-kate (approved):

PR #6523, opened 2025-05-23
cygwin: embed setup.exe to use it as a fallback as a last resort

### LINKED ISSUE #6498: `setup.exe`: embed in Windows binary
`setup.exe` is needed for some Cygwin related actions (install of the internal Cygwin, install Cygwin packages). Sometimes, its download is not possible because the website have some reachability issues by its own or because of too many downloads of the exe (and we don't store mirrors).

Discussed in dev meeting, @kit-ty-kate proposed to embed the `setup.exe` into the Windows binary. It only add 2M and resolve most network issue related to that binary (the binary itself and its sha512).

rjbou:
fix #6498


kit-ty-kate:
Superseded by https://github.com/ocaml/opam/pull/6526

PR #6526, opened 2025-06-02
Embed cygwin's setup.exe on windows as a fallback when cygwin.com is unavailable

kit-ty-kate:
Supersedes #6523 
Fixes https://github.com/ocaml/opam/issues/6498


configure.ac
```diff
@@ -89,6 +89,14 @@ AC_ARG_WITH([dune],
...
   AS_HELP_STRING([--with-vendored-deps],[Use vendored dependencies instead of any available in the compiler installation]))
 
+AC_ARG_WITH([embed-cygwin-setup],
+  AS_HELP_STRING([--embed-cygwin-setup],["Embed Cygwin's setup.exe into the executable"]))
+
+AS_IF([test "x${with_embed_cygwin_setup}" = "xyes"],[
+  curl --silent -Lo cygwin-setup.exe https://cygwin.com/setup-x86_64.exe
+  ocaml shell/embed.ml cygwin-setup.exe > src/state/opamEmbedCygwinSetup.ml
+])
+
```
dra27: embedded rather than embed? So `--with-embedded-cygwin-setup` (or it could just be `--with-cygwin-setup`?):
```suggestion
AC_ARG_WITH([embedded-cygwin-setup],
  AS_HELP_STRING([--with-embedded-cygwin-setup],["Embed Cygwin's setup.exe into the executable"]))

AS_IF([test "x${with_embedded_cygwin_setup}" = "xyes"],[
  curl --silent -Lo cygwin-setup.exe https://cygwin.com/setup-x86_64.exe
  ocaml shell/embed.ml cygwin-setup.exe > src/state/opamEmbedCygwinSetup.ml
])

```

In passing, to avoid compulsory network access in `configure` when using this, it'd be tempting to support `--with-embedded-cygwin-setup` meaning to download and `--with-embedded-cygwin-setup=setup-x86_64.exe` to use a local copy

shell/embed.ml
```diff
@@ -0,0 +1,22 @@
+let add_chan buf chan =
```
rjbou:
```
(* General module to embed content.
   Usage: 'ocaml shell/embed.ml > opamEmbedModule.ml'
*)
let add_chan buf chan =

File: configure.ac [hunk 3]
```diff
@@ -89,6 +89,14 @@ AC_ARG_WITH([dune],
 AC_ARG_WITH([vendored-deps],
   AS_HELP_STRING([--with-vendored-deps],[Use vendored dependencies instead of any available in the compiler installation]))
 
+AC_ARG_WITH([embed-cygwin-setup],
```
rjbou: Do we want to make this option available only on Windows system ?

configure.ac
```diff
@@ -89,6 +89,14 @@ AC_ARG_WITH([dune],
...
+  AS_HELP_STRING([--embed-cygwin-setup],["Embed Cygwin's setup.exe into the executable"]))
+
+AS_IF([test "x${with_embed_cygwin_setup}" = "xyes"],[
+  curl --silent -Lo cygwin-setup.exe https://cygwin.com/setup-x86_64.exe
```
rjbou: Do we want to check the hash ?

shell/embed.ml
```diff
@@ -0,0 +1,22 @@
...
+    let bs = Buffer.create 3_000_000 in
+    add_chan bs ic;
+    close_in ic;
+    Buffer.contents bs
+  in
+  Printf.printf
+    {|(* THIS FILE IS AUTOMATICALLY GENERATED, EDIT configure.ac & shell/embed.ml INSTEAD *)
+let content = Some "%s"
+|}
+    (String.escaped content_setup)
```
rjbou, line 17: We loose some compression by not using base64 (2.1M vs 1.5M)

kit-ty-kate: That's not what i'm observing. Base64 doesn't compress; on the contrary as far as i know it makes the data bigger due to its encoding.
Test done encoding my local `/usr/bin/cat` ELF binary:
- base64: 1.87MB
- String.escaped: 1.77MB

dra27: Erm, apples and oranges are being compared here - the base64 version was writing an actual base64 string into the _executable_ - so the setup executable will be 33% larger. This version is writing the OCaml string constant to the _source file_ ... so the data in the executable will be a verbatim copy of the original executable (plus a tiny bit of overhead from the marshal header)

rjbou: Indeed, I tested the `base64` tool and diffed the generated file, not the library directly in the code.

src/state/opamEmbedCygwinSetup.mli
```diff
@@ -0,0 +1,11 @@
...
+(*                                                                        *)
+(*    Copyright 2025 Kate Deplaix                                         *)
+(*                                                                        *)
+(*  All rights reserved. This file is distributed under the terms of the  *)
+(*  GNU Lesser General Public License version 2.1, with the special       *)
+(*  exception on linking described in the file LICENSE.                   *)
+(*                                                                        *)
+(**************************************************************************)
+
+val content : string option
```
rjbou:
```suggestion
(* Cygwin setup-x86_64.exe content if embed in opam *)
val content : string option
```

kit-ty-kate: done

configure.ac
```diff
@@ -89,6 +89,25 @@ AC_ARG_WITH([dune],
...
   AS_HELP_STRING([--with-vendored-deps],[Use vendored dependencies instead of any available in the compiler installation]))
 
+AC_ARG_WITH([cygwin-setup],
+  AS_HELP_STRING([--with-cygwin-setup],["Embed Cygwin's setup.exe into the executable"]),
+  [AS_IF([test ${WIN32} -eq 0],[AC_MSG_ERROR([--with-cygwin-setup is only supported on Windows])])])
+
+AS_IF([test "x$with_cygwin_setup" != "x"],[
+  AS_CASE([$with_cygwin_setup],
+    [yes], [
+      curl --silent -LO https://cygwin.com/sha512.sum
```
dra27: The file includes checksums for the 32-bit setup program which causes this:
```
shasum: setup-x86.exe: No such file or directory
setup-x86.exe: FAILED open or read
setup-x86_64.exe: OK
shasum: WARNING: 1 listed file could not be read
```
Possibly just:
```suggestion
      curl --silent -L https://cygwin.com/sha512.sum | grep -F x86_64 > sha512.sum
```
which then has:
```
checking OCaml Sys.os_type... Win32
setup-x86_64.exe: OK
checking for gawk... gawk
```

configure.ac
```diff
@@ -89,6 +89,25 @@ AC_ARG_WITH([dune],
...
+      curl --silent -LO https://cygwin.com/setup-x86_64.exe
+      shasum -c sha512.sum
+      CYGWIN_SETUP=setup-x86_64.exe
+    ],
+    [no], [],
+    [CYGWIN_SETUP=$with_cygwin_setup])
```
dra27: Means that a Cygwin path can be used to specify the file (e.g. `$PWD/cache`, etc.)
```suggestion
    [CYGWIN_SETUP="$(cygpath -m "$with_cygwin_setup")"])
```

src/state/dune
```diff
@@ -15,3 +15,9 @@
   (targets opamScript.ml)
   (deps    ../../shell/crunch.ml (glob_files shellscripts/*.*sh))
   (action  (with-stdout-to %{targets} (run ocaml %{deps}))))
+
+; Embed Cygwin mechanism (done in the configure script on Windows if requested)
```
dra27: Two completely optional observations, just having spotted it: `Embedded`, rather than `Embed`? It could be processed in `crunch.ml`, but it's not like adding a module particularly matters?

src/state/opamSysInteract.ml
```diff
@@ -278,11 +278,14 @@ module Cygwin = struct
...
-           OpamConsole.warning "%s failed to update" setupexe;
-           Done ()
-         end else
-           Printexc.raise_with_backtrace exn backtrace
+         if dst_exists then
+           OpamConsole.warning "%s failed to update" setupexe
+         else
+           begin match OpamEmbedCygwinSetup.content with
+           | None -> Printexc.raise_with_backtrace exn backtrace
+           | Some setup_exe_content -> OpamFilename.write dst setup_exe_content
```
dra27: I think it ought to at least display a log entry that it wrote from an internal copy, or even better a warning, e.g.
```suggestion
           | Some setup_exe_content ->
             OpamConsole.warning "Cygwin setup Failed to download - using embedded copy instead";
             OpamFilename.write dst setup_exe_content
```

src/state/dune
kit-ty-kate: `crunch.ml` does something slightly different with regards to `\n` so i'd rather not overcomplicate this PR. We can definitely look into merging the two later though if you want. If you're fine with that i'll open a separate issue to fix later.

kit-ty-kate:
@dra27 i believe that's everything included

dra27:
There was a typo in the change that meant setup was no longer downloaded - I spotted the "unclean" files while checking that and, as I'm on a train, also happened to spot that it ignores errors if the download fails, which is the backdrop for the other two commits!

PR #6528, opened 2025-06-04
Forbid 'opam install --deps-only' to ever installing/upgrading one of the requested packages

kit-ty-kate:
Alternative to https://github.com/ocaml/opam/pull/6520
Partially fixes https://github.com/ocaml/opam/issues/6501

In my opinion i think it makes very little sense for `--deps-only` to install packages listed in the argument so i think it makes sense to simply forbid it altogether.
As a nice bonus this fixes the issue in https://github.com/ocaml/opam/issues/6501 where the packages with the same name already installed, are removed.


kit-ty-kate:
Superseded by https://github.com/ocaml/opam/pull/6520

PR #6530, opened 2025-06-04
`opam install . --deps-only` comply with the real constraints of packages

### LINKED ISSUE #6529: `opam install . --deps-only` do not comply with packages constraints
Example (hand written)
```
$ opam install pkg
$ opam show pkg -f depends
"foo" { = "1" }
$ grep depends pkg/opam
depends: "foo" { = "2" }
$ opam install ./pkg --deps-only
=== recompile 1 package
  - recompile pkg     1 (pinned) [uses foo]
=== upgrade 1 package
  - upgrade   foo 1 to 2     [required by pkg]
[...]
```
Here, `pkg` should be removed as its constraint `foo=1` is no more satisfied.

rjbou:
When a package `pkg` is already installed and `opam install ./pkg --deps` is called, if there is a conflict between installed `pkg` dependencies and local `pkg` declaration, the conflict is not seen and the already installed `pkg` is kept.

For the code itself, the universe given to the solver was retrieving information from the simulated switch state, so the simulated pinned package information, not the good opam definition. We keep then the old information in the switch state, like that we can rearrange the package set given to the solver when creating the universe.

Fix #6529

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -36,6 +36,7 @@ users)
 ## Config
 
 ## Pin
+  * [BUG] When using `--deps-only`, no longer take into account the simulated pin information. This is hit when a package `pkg` is already installed and `opam install ./pkg --deps` is called, if there is a conflict between installed `pkg` dependencies and local `pkg` declaration, the conflict is not seen and the already installed `pkg` is kept. [#6530 @rjbou - fix #6529]
 
 ## List
 
--- a/master_changes.md
+++ b/master_changes.md
@@ -110,6 +110,7 @@ users)
 
 ## Reftests
 ### Tests
+  * Add some related pin tests: fetching, reinstall trigger & simulated pin (deps-only) [#6530 @rjbou]
 
 ### Engine
 
--- a/src/client/opamAuxCommands.ml
+++ b/src/client/opamAuxCommands.ml
@@ -390,6 +390,15 @@ let simulate_local_pinnings ?quiet ?(for_view=false) st to_pin =
       ++ local_packages
     else st.pinned
   in
+  let overwrote_opams =
+    if for_view then OpamPackage.Map.empty else
+      OpamPackage.Map.filter_map (fun nv opam ->
+          if OpamPackage.Set.mem nv local_packages then
+             Some (OpamPackage.Set.mem nv st.pinned, opam)
+           else
+             None)
+        st.opams
+  in
   let st = {
     st with
     opams =
@@ -429,6 +438,7 @@ let simulate_local_pinnings ?quiet ?(for_view=false) st to_pin =
         installed_pinned (Lazy.force st.reinstall)
     );
     pinned;
+    overwrote_opams;
   } in
   st, local_packages
 
--- a/src/state/opamStateTypes.mli
+++ b/src/state/opamStateTypes.mli
@@ -166,6 +166,10 @@ type +'lock switch_state = {
       of removed system dependencies. Only packages which are unavailable end up
       in this set, they are otherwise put in {!field:reinstall}. *)
 
+  overwrote_opams: (bool * OpamFile.OPAM.t) package_map;
+  (** In case of simulated pins, keep the old information of opam files. The
+      boolean is set to true if the package was previously pinned. *)
+
   (* Missing: a cache for
      - switch-global and package variables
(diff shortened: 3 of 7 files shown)


master_changes.md
```diff
@@ -36,6 +36,7 @@ users)
 ## Config
 
 ## Pin
+  * [BUG] When using `--deps-only`, no longer take into account the simulated pin information. This is hit when a package `pkg` is already installed and `opam install ./pkg --deps` is called, if there is a conflict between installed `pkg` dependencies and local `pkg` declaration, the conflict is not seen and the already installed `pkg` is kept. [#XXX @rjbou]
```
```suggestion
  * [BUG] When using `--deps-only`, no longer take into account the simulated pin information. This is hit when a package `pkg` is already installed and `opam install ./pkg --deps` is called, if there is a conflict between installed `pkg` dependencies and local `pkg` declaration, the conflict is not seen and the already installed `pkg` is kept. [#6530 @rjbou - fix #6529]
```

master_changes.md
```diff
@@ -110,6 +111,7 @@ users)
 
 ## Reftests
 ### Tests
+  * Add some related pin tests: fetching, reinstall trigger & simulated pin (deps-only) [#XXX @rjbou]
```
```suggestion
  * Add some related pin tests: fetching, reinstall trigger & simulated pin (deps-only) [#6530 @rjbou]
```

src/state/opamStateTypes.mli
```diff
@@ -166,6 +166,10 @@ type +'lock switch_state = {
       of removed system dependencies. Only packages which are unavailable end up
       in this set, they are otherwise put in {!field:reinstall}. *)
 
+  overwrote_opams: (bool * OpamFile.OPAM.t) package_map;
```
rjbou, line 169: The last commit can be removed if we merge #6531 before. To be squashed otherwise

src/state/opamSwitchState.ml
```diff
@@ -973,6 +975,19 @@ let universe st
...
+  let overwrote_opams, overwrote_unpinned =
+    OpamPackage.Map.fold (fun nv (pinned, opam) (opams, unpinneds) ->
+        OpamPackage.Map.add nv opam opams,
+        if pinned then unpinneds else OpamPackage.Set.add nv unpinneds)
+      st.overwrote_opams OpamPackage.(Map.empty, Set.empty)
+  in
+  let opams =
+    OpamPackage.Map.union (fun _ o -> o) st.opams overwrote_opams
+  in
+  let u_pinned = OpamPinned.packages st -- overwrote_unpinned in
```
rjbou: This is not the most optimal, it can be directly computed during the fold, but i found it more clear that way

The other way:
```suggestion
  let opams, u_pinned =
    OpamPackage.Map.fold (fun nv (was_pinned, opam) (opams, pinned) ->
        OpamPackage.Map.add nv opam opams,
        if was_pinned then pinned else OpamPackage.Set.remove nv pinned)
      st.overwrote_opams (st.opams, OpamPinned.packages st)
  in

File: src/client/opamAuxCommands.ml [hunk 5]
```diff
@@ -390,6 +390,14 @@ let simulate_local_pinnings ?quiet ?(for_view=false) st to_pin =
       ++ local_packages
     else st.pinned
   in
+  let overwrote_opams =
+    if for_view then OpamPackage.Map.empty else
+      OpamPackage.Map.filter (fun nv _ ->
+          OpamPackage.Set.mem nv local_packages)
+        st.opams
+      |> OpamPackage.Map.mapi (fun nv opam ->
+          OpamPackage.Set.mem nv st.pinned, opam)
```
kit-ty-kate: This can be done faster in one pass (if we don't merge https://github.com/ocaml/opam/pull/6531 before ofc):
```suggestion
      OpamPackage.Map.filter_map (fun nv opam ->
          if OpamPackage.Set.mem nv local_packages then
             Some (OpamPackage.Set.mem nv st.pinned, opam)
           else
             None)
        st.opams
```

kit-ty-kate (approved):
Tested beautifully together with https://github.com/ocaml/opam/pull/6522, https://github.com/ocaml/opam/pull/6520, https://github.com/ocaml/opam/pull/6533

PR #6531, opened 2025-06-04
Clean cudf attributes

rjbou:
While looking at #XXX, I found out that there is several attributes given to the solver that are not needed : pinned status and is the package is installed as root. None of the solvers uses that attribute. Digging in the history, I guess that it may have been used before to handle pinned or roots via the solver. These addition date back to the very beginning of opam (#751 for pinned, and ce584b5d51894d11a9d185bb00bfd28fbc8ea34d -no pr- for installed roots).

I checked locally with opam lib rev deps, nothing broke.

TODO
* [ ] add a cudf test to highlight the changes
* [x] the may be queued on #6530, the last two commit are here to revert some changes. If this PR is merged before, they can be removed.
* [ ] update API changelog

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -36,6 +36,7 @@ users)
 ## Config
 
 ## Pin
+  * [BUG] When using `--deps-only`, no longer take into account the simulated pin information. This is hit when a package `pkg` is already installed and `opam install ./pkg --deps` is called, if there is a conflict between installed `pkg` dependencies and local `pkg` declaration, the conflict is not seen and the already installed `pkg` is kept. [#XXX @rjbou]
 
 ## List
 
--- a/master_changes.md
+++ b/master_changes.md
@@ -99,7 +99,7 @@ users)
 ## Shell
 
 ## Internal
-  * Cudf: remove unused attribute pinned [#XXX @rjbou]
+  * Cudf: remove unused attribute pinned and installed-roots [#XXX @rjbou]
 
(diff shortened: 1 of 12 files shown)


master_changes.md
```diff
@@ -36,6 +36,7 @@ users)
 ## Config
 
 ## Pin
+  * [BUG] When using `--deps-only`, no longer take into account the simulated pin information. This is hit when a package `pkg` is already installed and `opam install ./pkg --deps` is called, if there is a conflict between installed `pkg` dependencies and local `pkg` declaration, the conflict is not seen and the already installed `pkg` is kept. [#XXX @rjbou]
```
rjbou, line 39:
```suggestion
  * [BUG] When using `--deps-only`, no longer take into account the simulated pin information. This is hit when a package `pkg` is already installed and `opam install ./pkg --deps` is called, if there is a conflict between installed `pkg` dependencies and local `pkg` declaration, the conflict is not seen and the already installed `pkg` is kept. [#6531 @rjbou]
```

master_changes.md
```diff
@@ -99,6 +100,7 @@ users)
 ## Shell
 
 ## Internal
+  * Cudf: remove unused attribute pinned and installed-roots [#XXX @rjbou]
```
rjbou, line 103:
```suggestion
  * Cudf: remove unused attribute pinned and installed-roots [#6531 @rjbou]
```

PR #6533, opened 2025-06-04
Fix `opam install . --deps` updating internal sources and stored opam file

### LINKED ISSUE #6532: `opam install . --deps-only` updates instannly source and opam file
If a package `foo` is already pinned, `opam install ./foo --deps-only` fetches and write the new sources, and stores the new opam file. The simulated pin done for `deps-only` should be a no-op regarding opam internals on disk.

rjbou:
During the pinning process, sometimes we don't want to completely repin (in case of an edit for example), so a fetch is done. But it was done even if the pin is simulated.
I've also added some tests to check those fetch cases (from #4542 & #2893's https://github.com/ocaml/opam/pull/2893/commits/02d1ea9bb604d680fd6910848f11c8a49583afb1)

Fix #6532

TODO
* [x] update API changelog ?

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -37,6 +37,7 @@ users)
 
 ## Pin
   * [BUG] When using `--deps-only`, no longer take into account the simulated pin information. This is hit when a package `pkg` is already installed and `opam install ./pkg --deps` is called, if there is a conflict between installed `pkg` dependencies and local `pkg` declaration, the conflict is not seen and the already installed `pkg` is kept. [#6530 @rjbou - fix #6529]
+  * [BUG] In case of simulated pin (deps-only, depext-only), no longer fetch and write the sources [#6533 @rjbou - fix #6532]
 
 ## List
 
--- a/tests/reftests/autopin.test
+++ b/tests/reftests/autopin.test
@@ -187,8 +187,6 @@ The following actions would be performed:
 The following system packages will first need to be installed:
     some-depext
 ### opam install ./foo --deps-only | sed-cmd echo
-[foo-core.2.0] synchronised (file://${BASEDIR}/foo)
-[foo-format.2.0] synchronised (file://${BASEDIR}/foo)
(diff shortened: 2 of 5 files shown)


master_changes.md
```diff
@@ -36,6 +36,7 @@ users)
 ## Config
 
 ## Pin
+  * [BUG] In case of simulated pin (deps-only, depext-only), no longer fetch and write the sources [#XXX @rjbou]
```
```suggestion
  * [BUG] In case of simulated pin (deps-only, depext-only), no longer fetch and write the sources [#6533 @rjbou - fix #6532]
```

master_changes.md
```diff
@@ -110,6 +111,9 @@ users)
 
 ## Reftests
 ### Tests
+  * Add some related pin tests: fetching, reinstall trigger & simulated pin (deps-only) [#XXX @rjbou]
+  * Add working dir test case to check the case where a requested working-dir package is already pinned with another url [#XXX @rjbou]
+  * Add a pin edit test case to check that the additional information is not lost in case of repin to the same url [#XXX @rjbou]
```
```suggestion
  * Add some related pin tests: fetching, reinstall trigger & simulated pin (deps-only) [#6533 @rjbou]
  * Add working dir test case to check the case where a requested working-dir package is already pinned with another url [#6533 @rjbou]
  * Add a pin edit test case to check that the additional information is not lost in case of repin to the same url [#6533 @rjbou]
```

kit-ty-kate (approved):

tests/reftests/autopin.test
```diff
@@ -247,14 +245,10 @@ The following system packages will first need to be installed:
 [WARNING] Opam package foo-format.2.0 depends on the following system package that can no longer be found: some-depext
 The following actions will be performed:
 === remove 1 package
-  - remove    foo-format 2.0 (pinned)
-=== recompile 1 package
-  - recompile foo-core   2.0 (pinned) [uses foo-format]
```
rjbou, line 252: note: This recompile is a consequence of the previous syncing, that marked `foo-core` to reinstall.

Issue #6543, opened 2025-06-12
opam tree <local dir/file> doesn't take into account the dependencies of the packages requested if a package of the same name is already installed


kit-ty-kate:
Currently with opam 2.3.0:
```
### : make sure the dependencies are taken from the correct opam description
### opam switch create opam-description-1 --empty
### <pkg:miou.0.3.1>
opam-version: "2.0"
### <pkg:miou.0.3.0>
opam-version: "2.0"
### <pin:test/opam>
opam-version: "2.0"
depends: ["miou"]
### opam pin add test.custom ./test

Package test does not exist, create as a NEW package? [y/n] y
[test.custom] synchronised (file://${BASEDIR}/test)
test is now pinned to file://${BASEDIR}/test (version custom)

The following actions will be performed:
=== install 2 packages
  - install miou 0.3.1           [required by test]
  - install test custom (pinned)

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
-> installed miou.0.3.1
-> retrieved test.custom  (file://${BASEDIR}/test)
-> installed test.custom
Done.
### <pin:test/opam>
opam-version: "2.0"
depends: ["miou" {= "0.3.0"}]
### opam tree ./test
test.custom
'-- miou.0.3.1
### : same but with non-pinned packages
### opam switch create opam-description-2 --empty
### <pkg:miou.0.3.1>
opam-version: "2.0"
### <pkg:miou.0.3.0>
opam-version: "2.0"
### <pkg:test.1>
opam-version: "2.0"
depends: ["miou"]
### opam install test
The following actions will be performed:
=== install 2 packages
  - install miou 0.3.1 [required by test]
  - install test 1

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
-> installed miou.0.3.1
-> installed test.1
Done.
### <pin:test/opam>
opam-version: "2.0"
build: "false"
depends: ["miou" {= "0.3.0"}]
### opam tree ./test
test.1
```
With the current master (708d62ec2b020fb1f176f7f8184f93bd1259496d) we get a similar result, except the last line becomes:
```
test.1
'-- miou.0.3.1
```
which also shouldn't happen.

The expected result for both calls to `opam tree ./test` are of course:
```
test.dev
'-- miou.0.3.0
```

PR #6544, opened 2025-06-12
Fix opam install --deps-only using the opam description of the wrong package in some cases

### LINKED ISSUE #6535: opam install --deps-only uses the wrong opam definition when building a package
Similarly to https://github.com/ocaml/opam/issues/6529, the build steps also use the wrong opam definition.
Using #6530 + #6533 + #6522 + #6520:
```
$ cat opam
opam-version: "2.0"
name: "test"
depends: ["miou"]
$ opam install .
[test] Installing new package description from upstream file:///tmp/tmp
=== install 2 packages
  ∗ miou 0.3.1 [required by test]
  ∗ test dev

Proceed with ∗ 1 installation? [Y/n] y

⬇ retrieved miou.0.3.1  (cached)
∗ installed miou.0.3.1
∗ installed test.dev
Done.

$ cat opam
opam-version: "2.0"
name: "test"
build: "false"
depends: ["miou" {= "0.3.0"}]
$ opam install --deps .
=== downgrade 1 package
  ↘ miou 0.3.1 to 0.3.0 [required by test]
=== recompile 1 package
  ↻ test dev (pinned)   [uses miou]

Proceed with ↘ 1 downgrade and ↻ 1 recompilation? [Y/n] y

⬇ retrieved miou.0.3.0  (cached)
⬇ retrieved test.dev  (file:///tmp/tmp)
⊘ removed   test.dev
⊘ removed   miou.0.3.1
∗ installed miou.0.3.0
[ERROR] The compilation of test.dev failed at "false".

#=== ERROR while compiling test.dev ===========================================#
# context     2.4.0~alpha3~dev | linux/arm64 | kit-ty-kate-platform.1 ocaml-base-compiler.5.4.0~alpha1 | pinned(file:///tmp/tmp)
# path        ~/.opam/5.4/.opam-switch/build/test.dev
# command     ~/.opam/opam-init/hooks/sandbox.sh build false
# exit-code   1
# env-file    ~/.opam/log/test-21406-cf6c92.env
# output-file ~/.opam/log/test-21406-cf6c92.out
```

kit-ty-kate:
Fixes #6535

https://github.com/ocaml/opam/pull/6530 fixed the issue in the solver but created a discrepancy between what the solver gives displays and what is actually going to be executed.

To fix this we simply move the bit that re-add the non-simulated opam definitions, up before any solver/actions, and after any code specific to `--deps-only` where we actually need those information.

After careful review of the different usages of the autopin system, i think most other commands do not require fixes, at the exception of:
- the already broken `opam tree`: https://github.com/ocaml/opam/issues/6543
- possibly `opam switch create --deps-only` but i haven't managed to produce a test-case that show any issue yet. In any case i suspect this would be a very niche corner case if there is an issue there.

### DIFF
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -2276,6 +2276,23 @@ let install_t t ?ask ?(ignore_conflicts=false) ?(depext_only=false)
     OpamSwitchAction.write_selections t
   );
 
+  (* We remove the simulated state for deps-only in order to be able to do
+     actions (solve, apply) with the non simulated opam file. The simulated
+     meaningful information is already "stored" in the 'deps-of-xxx' packages
+     opam files. *)
+  let t =
+    if deps_only then
+      let opams, pinned =
+        OpamPackage.Map.fold (fun nv (was_pinned, opam) (opams, pinned) ->
+            OpamPackage.Map.add nv opam opams,
+            if was_pinned then pinned else OpamPackage.Set.remove nv pinned)
+          t.overwrote_opams (t.opams, OpamPinned.packages t)
+      in
+      {t with opams; pinned; overwrote_opams = OpamPackage.Map.empty}
+    else
+      t
+  in
+
   OpamSolution.check_availability t available_packages atoms;
 
   if pkg_new = [] && OpamPackage.Set.is_empty pkg_reinstall &&
--- a/master_changes.md
+++ b/master_changes.md
@@ -24,6 +24,7 @@ users)
 ## Actions
 
 ## Install
+  * [BUG] Fix `opam install --deps-only` using the opam description of the wrong package in some cases [#6544 @kit-ty-kate - fix #6535]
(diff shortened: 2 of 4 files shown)


src/client/opamClient.ml
```diff
@@ -2276,6 +2276,19 @@ let install_t t ?ask ?(ignore_conflicts=false) ?(depext_only=false)
     OpamSwitchAction.write_selections t
   );
 
+  let t =
```
rjbou, line 2283:
```suggestion
  (* We remove the simulated state for deps-only in order to be able to do
     actions (solve, apply) with the non simulated opam file. The simulated one
     meaningful information is already "stored" in the 'deps-of-xxx' packages
     opam files. *)
  let t =
```

kit-ty-kate: applied

PR #6539, opened 2025-06-10
reftest: add test for opam repo ci reinstall workflow

rjbou:
Test for https://github.com/ocurrent/opam-repo-ci/issues/441
queued on
* [x] #6549

PR #6550, opened 2025-06-16
No longer remove source dir on reinstall

### LINKED ISSUE #6554: Reinstalling package sometimes refteches sources
When a reinstall of a package is done, the source directory is sometimes emptied and a new fetch of the archive is done, instead of reusing the same sources.

Noticed in https://github.com/ocurrent/opam-repo-ci/issues/441 & #6539. Introduced/highlighted with #6436.

rjbou:
First noticed on #6539.
In #6436, removing build dir changes made the call to the function to clean packages artefacts no more called only for non pinned packages. This function was originally called when calling `opam clean`, it is no more the case, it is safe to no longer remove sources directory of non pinned package in it.
Fix #6554
Queued on 
* [x] #6539 for test
* [x] #6549 for reftest engine updates needed


master_changes.md
```diff
@@ -24,6 +24,7 @@ users)
 ## Actions
 
 ## Install
+  * [BUG] Fix sources directory removed on reinstall actions: no longer remove sources directory if the package is not pinned when cleaning after a removal action [#6550 @rjbou - fix #6551]
```
rjbou, line 29:
```suggestion
  * [BUG] Fix sources directory removed on reinstall actions: no longer remove sources directory if the package is not pinned when cleaning after a removal action [#6550 @rjbou - fix #6554]
```

tests/reftests/dune.inc
```diff
@@ -1,4 +1,25 @@
 
+(rule
+ (alias reftest-action-disk)
```
rjbou, line 3: If you find a better name for the test

kit-ty-kate: i think it's fine

kit-ty-kate (approved):

PR #6571, opened 2025-06-25
on `opam switch create <version>`, use avoid-version compiler if it is the only one available

### LINKED ISSUE #6563: opam switch create 5.4.0~alpha1 doesn't work anymore
https://github.com/ocaml/opam/pull/6494 removed compiler packages from the invariant to eliminate non-explicit use of `ocaml-system`, but doing so we forgot about the cases such as the ones advertised in [the compiler announcements](https://discuss.ocaml.org/t/first-alpha-release-of-ocaml-5-4-0/16722) where the only available package is tagged with `avoid-version`.

```
$ opam switch create 5.4 5.4.0~alpha1
[ERROR] No compiler matching `5.4.0~alpha1' found, use `opam switch list-available' to see what is available, or use `--packages' to select packages explicitly.
```

I think the wanted fix for this is to only remove the avoid-version packages from the formula when at least one non-avoid-version package with that version is available. If there are only avoid-version packages available then we could either:
- add them all to the formula
- or error with a good error message asking to be more precise if there are more than one

rjbou:
fix #6563

### DIFF
--- a/src/client/opamSwitchCommand.ml
+++ b/src/client/opamSwitchCommand.ml
@@ -746,14 +746,17 @@ let guess_compiler_invariant ?repos rt strings =
   in
   let opams = OpamRepositoryState.build_index rt repos  in
   let packages = OpamPackage.keys opams in
-  let compiler_packages =
-    OpamPackage.Map.filter
-      (fun _ opam ->
-         OpamFile.OPAM.has_flag Pkgflag_Compiler opam &&
-         not (OpamFile.OPAM.has_flag Pkgflag_AvoidVersion opam) &&
-         not (OpamFile.OPAM.has_flag Pkgflag_Deprecated opam))
-      opams
-    |> OpamPackage.keys
+  let compiler_packages, removed_compiler_packages =
+    OpamPackage.Map.fold (fun nv opam (comp, avoid) ->
+        if OpamFile.OPAM.has_flag Pkgflag_Compiler opam then
+          if OpamFile.OPAM.has_flag Pkgflag_AvoidVersion opam
+          || OpamFile.OPAM.has_flag Pkgflag_Deprecated opam then
+            comp, OpamPackage.Set.add nv avoid
+          else
+            OpamPackage.Set.add nv comp, avoid
+        else
+          comp, avoid)
+      opams OpamPackage.Set.(empty, empty)
   in
   let invariant =
     List.fold_left (fun acc str ->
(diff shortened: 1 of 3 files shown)


src/client/opamSwitchCommand.ml
```diff
@@ -766,13 +769,23 @@ let guess_compiler_invariant ?repos rt strings =
         with Failure _ | Not_found ->
         try
           let v = OpamPackage.Version.of_string str in
+          let filter = OpamPackage.Set.filter (fun nv -> nv.version = v) in
```
kit-ty-kate:
```suggestion
          let filter = OpamPackage.Set.filter (fun nv -> OpamPackage.Version.equal nv.version v) in
```

tests/reftests/switch-creation.test
```diff
@@ -156,3 +156,57 @@ Switch invariant: ["ocaml-base-compiler" {= "4.14.0"}]
 <><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
 -> installed ocaml-base-compiler.4.14.0
 Done.
+### :c: switch create <version> should keep avoid version packages if only
+### ::: avoid version compiler packages are compliant
+### rm -rf REPO/packages
+### <pkg:ocaml-base-compiler.5.4.0~alpha1>
+opam-version: "2.0"
+flags: [compiler avoid-version]
+### <pkg:ocaml-variant.5.4.0~alpha1>
```
kit-ty-kate: In opam-repository that package has version `5.4.0~alpha1+options`. I think the tests should at least have the case where only one avoid-version package is available (instead of two)

kit-ty-kate (commented):
lgtm otherwise

src/client/opamSwitchCommand.ml
```diff
@@ -766,13 +769,26 @@ let guess_compiler_invariant ?repos rt strings =
...
           let candidates =
-            OpamPackage.Set.filter (fun nv -> nv.version = v)
-              compiler_packages
+            let candidates = filter compiler_packages in
+            if OpamPackage.Set.is_empty candidates then
+              let avoided_candidates = filter removed_compiler_packages in
+              if OpamPackage.Set.is_empty avoided_candidates then
+                raise Not_found
+              else
+                (OpamConsole.note
```
kit-ty-kate: Thinking about it, i feel like the note doesn't make much sense when only one version is available. It's like scolding users for being explicit. Adding a reftest for this would be nice too.
```suggestion
                (if not (OpamPackage.Set.is_singleton avoided_candidates) then
                   OpamConsole.note
```

tests/reftests/switch-creation.test
```diff
@@ -156,3 +156,45 @@ Switch invariant: ["ocaml-base-compiler" {= "4.14.0"}]
...
+### opam switch create --solver=builtin-0install 5.4.0~alpha1install 5.4.0~alpha1
+[NOTE] Compilers found for version 5.4.0~alpha1 are all flagged with 'avoid-version'.
+       It is better to use a more explicit command like 'opam switch create <switch-name> <package-name.version>'
+
+<><> Installing new switch packages <><><><><><><><><><><><><><><><><><><><><><>
+Switch invariant: ["ocaml-base-compiler" {= "5.4.0~alpha1"}]
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> installed ocaml-base-compiler.5.4.0~alpha1
+Done.
```
kit-ty-kate: I don't think these test make much sense given the switch invariant contains only one package
```suggestion
```

src/client/opamSwitchCommand.ml
```diff
@@ -766,13 +769,26 @@ let guess_compiler_invariant ?repos rt strings =
...
+                   "Compilers found for version %s are all flagged with \
+                    'avoid-version'.\nIt is better to use a more explicit \
+                    command like \
+                    'opam switch create <switch-name> <package-name.version>'"
+                   str;
```
```suggestion
                   "Compiler%s found for version %s %s flagged with \
                    'avoid-version'.\nIt is better to use a more explicit \
                    command like \
                    'opam switch create <switch-name> <package-name.version>'"
                    (if OpamPackage.Set.is_singleton avoided_candidates then "" else "s")
                    str
                    (if OpamPackage.Set.is_singleton avoided_candidates then "is" else "ar all")
```

src/client/opamSwitchCommand.ml
rjbou: It just needs a rephrase to handle singleton. OR you are thinking about something else where there is only 1 package ?

kit-ty-kate: No what i meant is i don't think we should show a note at all when there is only one possible version

kit-ty-kate (approved):

kit-ty-kate:
The segfault in the Cold-Linux action in CI is unrelated (probably a github action base image bug once again). To be sure i've launched a PR with the same content on my fork and it worked just fine: https://github.com/kit-ty-kate/opam/pull/14

Thanks!

PR #6576, opened 2025-07-01
Speedup OpamStd.{Set,Map}.is_singleton

kit-ty-kate:
Idea from @gasche in https://github.com/ocaml/ocaml/pull/14119#issuecomment-3022214481

### DIFF
--- a/src/core/opamStd.ml
+++ b/src/core/opamStd.ml
@@ -244,8 +244,9 @@ module Set = struct
       !r
 
     let is_singleton s =
-      not (is_empty s) &&
-      min_elt s == max_elt s
+      let exception Break in
+      try S.fold (fun _ b -> if b then raise_notrace Break else true) s false
+      with Break -> false
 
     let choose_one s =
       if is_empty s then raise Not_found
@@ -379,8 +380,9 @@ module Map = struct
         m1 m2
 
     let is_singleton s =
-      not (is_empty s) &&
-      fst (min_binding s) == fst (max_binding s)
+      let exception Break in
+      try M.fold (fun _ _ b -> if b then raise_notrace Break else true) s false
+      with Break -> false
 
     let to_string string_of_value m =
       if M.cardinal m > max_print then
--- a/master_changes.md
+++ b/master_changes.md
@@ -133,3 +133,4 @@ users)
 ## opam-format
 
 ## opam-core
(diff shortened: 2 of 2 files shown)


src/core/opamStd.ml
```diff
@@ -244,8 +244,16 @@ module Set = struct
...
+      let had_one_iteration = ref false in
+      try
+        S.iter (fun _ ->
+            if !had_one_iteration then
+              raise Not_found
+            else
+              had_one_iteration := true)
+          s;
+        !had_one_iteration
+      with Not_found -> false
```
dbuenzli: What about: 

```ocaml
let is_singleton set =
  let count_to_one _ count = if count > 1 then raise Exit else count + 1 in
  try Int.equal 1 (S.fold count_to_one set 0) with Exit -> false
```

(In turn that would be easy to generalize to `has_cardinal : int -> Set.t -> bool`).

dbuenzli:
> (eg. javascript; but why would one ever run `opam` inside the browser, right?).

Never say never :–) More seriously note that some hosting providers nowadays provide managed  `node.js` hosting (in the same way they used to provide managed `php` hosting) so compiling to JavaScript has other uses.

src/core/opamStd.ml
kit-ty-kate: thanks for the suggestion, i've pushed a similar one (with local exception and booleans instead)

src/core/opamStd.ml
```diff
@@ -379,8 +380,9 @@ module Map = struct
         m1 m2
 
     let is_singleton s =
-      not (is_empty s) &&
-      fst (min_binding s) == fst (max_binding s)
+      let exception Break in
+      try M.fold (fun _ _ b -> if b then raise Break else true) s false
```
dbuenzli: raise_notrace ?

kit-ty-kate: ah i forgot this function existed. thx

gasche:
I still think that you should benchmark because I suspect that this may be slower in practice, due to the (local) exception.

kit-ty-kate:
> I still think that you should benchmark because I suspect that this may be slower in practice, due to the (local) exception.

mmh ok fair enough, most cases that i tested (up to sets of cardinal 10_000_000) the difference is insignificant, whereas https://github.com/ocaml/ocaml/pull/14118 is of course faster.

PR #6579, opened 2025-07-03
GHA debug Cold-Linux

kit-ty-kate:


kit-ty-kate:
It magically doesn't get oom-killed now. I guess something is wrong with the GHA machines

PR #6592, opened 2025-07-10
Update Windows bootstrap dependencies

MisterDA:
I've noticed that some of the dependencies could be updated. The updates could help when bootstrapping or building opam under Windows.

- FlexDLL 0.44 (update from 0.43) brings fixes for linking with long filenames, linking under the multicore runtime, and fixes for cross-compilation. See the [release notes](https://github.com/ocaml/flexdll/releases/tag/0.44) and [release announcement](https://discuss.ocaml.org/t/flexdll-0-44-released/16614). FlexDLL 0.44 has already been published to the opam-repository and is used by all up-to-date switches on Windows.
- msvs-detect 0.7.0 (update from 0.5.0) brings support for the latest Visual Studio, and various fixes. See [release notes](https://github.com/metastack/msvs-tools/releases).
- mingw-w64 13.0.0 (update from 12.0.0, CI only) has various improvements, but this shouldn't affect us too much ([changelog](https://www.mingw-w64.org/changelog/)).

I'm a bit dissatisfied to "dump" this PR on the maintainers this late in the release process, without being able to check if the CI follows through, but I don't have much more time available.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -99,6 +99,7 @@ users)
   * Add missing constraints to avoid cmdliner 2.0.0 [#6707 @kit-ty-kate]
   * Add patch library dependency to opam-state [#6614 @arozovyk]
   * Update the bootstrap compiler's flexdll to 0.44 [#6592 @MisterDA]
+  * Update the `msvs-detect` script used on Windows during compiler bootstrap to 0.7.0 [#6592 @MisterDA]
 
 ## Infrastructure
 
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
Turns out all of these dependencies simply do not work with OCaml 4.14 so i have to revert this PR. I'm not too sure why it works in CI but not during the release process though. We can revisit it once we upgrade to OCaml 5 (hopefully 5.5 or whenever the GC pacing issue is taken care of)

MisterDA:
> Turns out all of these dependencies simply do not work with OCaml 4.14 so i have to revert this PR. I'm not too sure why it works in CI but not during the release process though.

Sorry, but that's simply wrong. The dependencies work well with OCaml 4.14. It's the bootstrap process that passes opam's version to FlexDLL instead of FlexDLL's version.

Issue #6610, opened 2025-07-20
[2.4 regression] opam install --deps-only does not handle pin dependencies


kit-ty-kate:
https://github.com/ocaml/opam/pull/6375 seems to have broken the following case:
```
$ cat opam
opam-version: "2.0"
depends: "spdx_licenses"
pin-depends: ["spdx_licenses.dev" "git+https://github.com/kit-ty-kate/spdx_licenses.git"]
$ opam install --deps .
[...]
Nothing to do.
```
We would expect the pin-depends to be pinned and the new pinned package to be installed.

Encountered in [dune's CI](https://github.com/ocaml/dune/actions/runs/16395597419/job/46327451684#step:11:17) by @anmonteiro

kit-ty-kate:
Fixed in 2.4.1 but moving the milestone to 2.5.0~alpha1 since the PR on master isn't merged yet.

PR #6611, opened 2025-07-20
Fix pinned depends being ignored on simulated pins

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/6610

Weirdly enough, since this is reapplying https://github.com/ocaml/opam/pull/6375, the test added in https://github.com/ocaml/opam/pull/6380 should show the previous behaviour of double-pins, but we might have fixed it again another way.


rjbou (approved):
lgtm!
I've used your tests added in autopin to write `pin-depends` and `pin-depends-ignore` pin-depends (as the name doesn't let it know :upside_down_face:) specific tests. It's in separate commits (and no changelog, i've also added `proceed-actions` autoanswer), feel free to squash/reorder commits if it's ok for you.
About the last commit, if you can add also some textual information, we don't always have issue numbers in mind :)

master_changes.md
```diff
@@ -120,6 +120,7 @@ users)
   * Replace every polymorphic uses of `List.mem` by a version that doesn't use `Repr.equal` [#6644 @kit-ty-kate]
   * Simplify the `src_ext/update-sources.sh` script [#6701 @kit-ty-kate]
   * Homogeneise verbose command output between sandboxed and non sandboxed one [#6675 @rjbou]
+  * Add the `install-pin-depends`, `ignore-pin-depends` and `switch-clean-up` named questions [#6611 @kit-ty-kate]
```
rjbou: i've added `proceed-actions` too

PR #6612, opened 2025-07-20
[2.4] Fix pinned depends being ignored on simulated pins

kit-ty-kate:
Backports #6611 on the 2.4 branch.


tests/reftests/autopin.test
```diff
@@ -311,3 +311,84 @@ opam believes some required external dependencies are missing. opam can:
 + echo "some-depext"
 - some-depext
 Nothing to do.
+### : Make sure pin-depends are installed when using "opam install --deps-only"
```
rjbou, line 314: just a note, not for 2.4 PR: in the master PR, this test should be moved on its own pin-depends reftest

Issue #6599, opened 2025-07-15
Move CRLF preprocessing of patch files to the patch library


kit-ty-kate:
`OpamSystem.translate_patch` ensures the CRLF and LF files play nice with each other on different systems. Currently a second patch parser for this purpose is located in the `translate_patch` function but it would be nice to integrate this into the patch library itself as a builtin feature, which would allow us to remove some code and be way more efficient.

PR #5146, opened 2022-05-19
Repository update with no changes: cache part

rjbou:
Sequel of #5043 : if the is no changes in all updated repositories, do not write the unmodified repos-config & cache.

PR #6614, opened 2025-07-22
`opam update`:  load only changed opam files 

### LINKED ISSUE #5824: `opam update` completely reloads the repository
This is most noticeable on Windows, where the file system I/O between opam / OCaml is effectively slower, but the problem can be seen in less-cached environments - e.g. in a Docker container on Linux.

For a git remote, opam is able to short-circuit reloading the entire repository when it detects that there are no changes (this is a good reason for using git remotes, as it's a faster check than for the https remote). However, if there are _any_ changes, the entire repository is reloaded.

It seems we could be fairly easily doing something a little better, and possibly also for the https remote on update. The update is actually applied by means of a patch (even for git remotes), which means that opam _always_ knows exactly which files have changed.

Related to @kit-ty-kate's work on #5741 (and also #5648).

arozovyk:
Queued on:
- https://github.com/ocaml/opam/pull/6715

This PR leverages the fact that we already compute a diff/patch during update to determine exactly which files have changed and it what way. Instead of reloading the entire repository we:

- Propagate `Patch.operation` info from `OpamFilename.patch`   
- Add incremental loading via `OpamRepositoryState.load_opams_incremental`
- Maintain existing opam files that haven't changed, avoiding unnecessary I/O

On Windows the `OpamRepositoryState.load_opams_from_dir` in `OpamUpdate.repository` currently takes ~10s. This PR brings it down to ~0.01s (the time it takes to load the opam files of a usual repository change). 
On unix, it is a bit less dramatic, but still going from ~3.5s to 1ms

Current benchmarks ( added in #6681 )  :
update_benchmarks
So this is naturally very noticable for `opam update` processing small diffs. 

closes #5824

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -52,6 +52,7 @@ users)
 
 ## Update / Upgrade
   * Fix the false-positive mismatch debug warning during `opam update` when faced with nested extra-files on Windows [#6715 @kit-ty-kate]
+  * Implement incremental opam file loading to process only changed files during repository updates and repository state loading [#6614 @arozovyk - fix #5824]
 
 ## Tree
 
@@ -88,6 +89,7 @@ users)
   * Remove `seq` from the list of packages to download-if-missing as it is no longer a dependency of `re` [#6700 @kit-ty-kate]
   * `./configure --enable-static` is now supported on OpenBSD [#6705 @flumf]
   * Add missing constraints to avoid cmdliner 2.0.0 [#6707 @kit-ty-kate]
+  * Add patch library dependency to opam-state [#6614 @arozovyk]
 
 ## Infrastructure
 
@@ -196,6 +198,7 @@ users)
 
 ## opam-state
   * `OpamSwitchState.files`: was removed [#6662 @kit-ty-kate]
+  * `OpamRepositoryState` add `load_opams_from_diff` to update package definitions based on file change operations (diff) [#6614 @arozovyk]
 
 ## opam-solver
 
@@ -232,4 +235,4 @@ users)
   * `OpamFilename`: add `parse_patch` that preprocesses and parses a patch file. [#6614 @arozovyk]
   * `OpamSystem`: add lower-level `parse_patch` that preprocesses and parses a patch file. [#6614 @arozovyk]
   * `OpamFilename.patch`: use variants to make the input either `Filename.t` or reuse `Patch.diffs` directly. Remove the `?preprocess` argument since the preprocess logic is moved to the `OpamFilename.parse_patch` function that is called only in `OpamVCS` (mirroring the previous logic). [#6614 @arozovyk]
-  * `OpamSystem.patch` change the signature to work directly with `Patch.diffs` (implementation is now the previously `internal_patch` function), parsing is now done separately. [#6614 @arozovyk]
+  * `OpamSystem.patch`: change the signature to work directly with `Patch.diffs` (implementation is now the previously `internal_patch` function), parsing is now done separately. [#6614 @arozovyk]
--- a/src/state/dune
+++ b/src/state/dune
@@ -1,7 +1,7 @@
 (library
   (name        opam_state)
   (public_name opam-state)
-  (libraries   opam-repository re spdx_licenses)
+  (libraries   opam-repository re spdx_licenses patch)
   (synopsis    "OCaml Package Manager instance management library")
   (modules_without_implementation OpamStateTypes)
   (modules     :standard)
--- a/src/state/opamRepositoryState.ml
+++ b/src/state/opamRepositoryState.ml
@@ -76,6 +76,35 @@ module Cache = struct
 
 end
 
+let get_root_raw root repos_tmp name =
+  match Hashtbl.find repos_tmp name with
+  | lazy repo_root -> repo_root
+  | exception Not_found -> OpamRepositoryPath.root root name
+
+let get_root rt name =
+  get_root_raw rt.repos_global.root rt.repos_tmp name
+
+let get_repo_root rt repo =
+  get_root_raw rt.repos_global.root rt.repos_tmp repo.repo_name
+
+let read_package_opam ~repo_name ~repo_root package_dir =
+  match OpamFileTools.read_repo_opam ~repo_name ~repo_root package_dir with
+  | Some opam ->
+    (try
+       let nv =
+         OpamPackage.of_string
+           (OpamFilename.Base.to_string (OpamFilename.basename_dir package_dir))
+       in
+       Some (nv, opam)
+     with Failure _ ->
+       log "ERR: directory name not a valid package: ignored %s"
+         (OpamFilename.to_string OpamFilename.Op.(package_dir // "opam"));
+       None)
+  | None ->
+    log "ERR: Could not load %s, ignored"
+      (OpamFilename.to_string OpamFilename.Op.(package_dir // "opam"));
+    None
+
 let load_opams_from_dir repo_name repo_root =
   if OpamConsole.disp_status_line () || OpamConsole.verbose () then
     OpamConsole.status_line "Processing: [%s: loading data]"
@@ -85,22 +114,9 @@ let load_opams_from_dir repo_name repo_root =
     if OpamFilename.exists_dir dir then
       let fnames = Sys.readdir (OpamFilename.Dir.to_string dir) in
       if Array.exists (fun f -> f = "opam") fnames then
-        match OpamFileTools.read_repo_opam ~repo_name ~repo_root dir with
-        | Some opam ->
-          (try
-             let nv =
-               OpamPackage.of_string
-                 OpamFilename.(Base.to_string (basename_dir dir))
-             in
-             OpamPackage.Map.add nv opam r
-           with Failure _ ->
-             log "ERR: directory name not a valid package: ignored %s"
-               OpamFilename.(to_string Op.(dir // "opam"));
-             r)
-        | None ->
-          log "ERR: Could not load %s, ignored"
-            OpamFilename.(to_string Op.(dir // "opam"));
-          r
+        match read_package_opam ~repo_name ~repo_root dir with
+        | Some (nv, opam) -> OpamPackage.Map.add nv opam r
+        | None -> r
       else
         Array.fold_left (fun r name -> aux r OpamFilename.Op.(dir / name))
           r fnames
@@ -108,6 +124,69 @@ let load_opams_from_dir repo_name repo_root =
   in
   Fun.protect
     (fun () -> aux OpamPackage.Map.empty (OpamRepositoryPath.packages_dir repo_root))
+    ~finally:OpamConsole.clear_status
+
+let load_opams_from_diff repo diffs rt =
+  if OpamConsole.disp_status_line () || OpamConsole.verbose () then
+    OpamConsole.status_line "Processing: [%s: loading data]"
+      (OpamConsole.colorise `blue (OpamRepositoryName.to_string repo.repo_name));
+  let existing_opams =
+    OpamRepositoryName.Map.find repo.repo_name rt.repo_opams
+  in
+  let repo_root = get_repo_root rt repo in
+  let process_file (opams, processed_dirs) file ~is_removal =
+    let pkg_dir =
+      let file = OpamFilename.raw file in
+      let dirname = OpamFilename.dirname file in
+      let basename = OpamFilename.basename_dir dirname in
+      let full_path =
+        Filename.concat
+          (OpamFilename.Dir.to_string repo_root)
+          (OpamFilename.Dir.to_string dirname)
+      in
+      if OpamFilename.Base.to_string basename = "files" then
+        OpamFilename.Dir.of_string (Filename.dirname full_path)
+      else
+        OpamFilename.Dir.of_string full_path
+    in
+    if OpamFilename.Dir.Set.mem pkg_dir processed_dirs then
+      opams, processed_dirs
+    else
+      let processed_dirs = OpamFilename.Dir.Set.add pkg_dir processed_dirs in
+      match read_package_opam ~repo_name:repo.repo_name ~repo_root pkg_dir with
+      | Some (nv, opam) -> OpamPackage.Map.add nv opam opams, processed_dirs
+      | None ->
+        if is_removal then
+          match OpamPackage.of_dirname pkg_dir with
+          | None ->
+            log "ERR: directory name not a valid package: ignored %s"
+              (OpamFilename.Dir.to_string pkg_dir);
+            opams, processed_dirs
+          | Some nv ->
+            OpamPackage.Map.remove nv opams, processed_dirs
+        else
+          opams, processed_dirs
+  in
+  let remove_file file acc = process_file acc file ~is_removal:true in
+  let add_file file acc = process_file acc file ~is_removal:false in
+  let process_operation acc = function
+    | Patch.Edit (old_file, new_file) ->
+      if String.equal old_file new_file
+      then
+        add_file new_file acc
+      else
+        remove_file old_file acc |> add_file new_file
+    | Patch.Delete file -> remove_file file acc
+    | Patch.Create file -> add_file file acc
+    | Patch.Git_ext (file1, file2, git_ext) ->
+      match git_ext with
+      | Patch.Rename_only (_, _) -> remove_file file1 acc |> add_file file2
+      | Patch.Delete_only -> remove_file file1 acc
+      | Patch.Create_only -> add_file file2 acc
+  in
+  Fun.protect
+    (fun () -> List.fold_left process_operation
+        (existing_opams, OpamFilename.Dir.Set.empty) diffs |> fst)
     ~finally:OpamConsole.clear_status
 
 let load_repo repo repo_root =
@@ -141,17 +220,6 @@ let remove_from_repos_tmp rt name =
 let cleanup rt =
   Hashtbl.iter (fun _ tmp_dir -> clean_repo_tmp tmp_dir) rt.repos_tmp;
   Hashtbl.clear rt.repos_tmp
-
-let get_root_raw root repos_tmp name =
-  match Hashtbl.find repos_tmp name with
-  | lazy repo_root -> repo_root
-  | exception Not_found -> OpamRepositoryPath.root root name
-
-let get_root rt name =
-  get_root_raw rt.repos_global.root rt.repos_tmp name
-
-let get_repo_root rt repo =
-  get_root_raw rt.repos_global.root rt.repos_tmp repo.repo_name
 
 let load lock_kind gt =
   log "LOAD-REPOSITORY-STATE %@ %a" (slog OpamFilename.Dir.to_string) gt.root;
--- a/src/state/opamRepositoryState.mli
+++ b/src/state/opamRepositoryState.mli
@@ -51,6 +51,21 @@ val get_repo: 'a repos_state -> repository_name -> repository
 
 val load_opams_from_dir: repository_name -> dirname -> OpamFile.OPAM.t OpamPackage.Map.t
 
+(** [load_opams_from_diff repo diffs rt] incrementally
+    updates package definitions by processing only changed files.
+
+    Starting from existing packages in [rt.repo_opams], it:
+    - Adds new packages
+    - Removes packages whose opam files were deleted
+    - Updates packages from changed opam files that exist
+
+      @param diffs List of [Patch.operation] representing creations, deletions,
+    and modifications of regular files
+*)
+val load_opams_from_diff:
+  repository -> Patch.operation list -> 'a repos_state
+  -> OpamFile.OPAM.t package_map
+
 (** Load all the metadata within the local mirror of the given repository,
     without cache *)
 val load_repo:
--- a/src/state/opamUpdate.ml
+++ b/src/state/opamUpdate.ml
@@ -99,7 +99,7 @@ let repository rt repo =
   | `No_changes ->
     log "Repository did not change: nothing to do.";
     Done None
-  | `Changes _ ->
+  | `Changes diffs ->
     log "Repository has new changes";
     let repo_file = OpamFile.Repo.safe_read repo_file_path in
     let repo_file = OpamFile.Repo.with_root_url repo.repo_url repo_file in
@@ -134,7 +134,10 @@ let repository rt repo =
         (Printexc.to_string e)
     | None ->
       let opams =
-        OpamRepositoryState.load_opams_from_dir repo.repo_name repo_root
+        match diffs with
+        | [] ->
+          OpamRepositoryState.load_opams_from_dir repo.repo_name repo_root
+        | diffs -> OpamRepositoryState.load_opams_from_diff repo diffs rt
       in
       let local_dir = OpamRepositoryPath.root gt.root repo.repo_name in
       if OpamRepositoryConfig.(!r.repo_tarring) then
--- a/tests/reftests/update.test
+++ b/tests/reftests/update.test
@@ -395,7 +395,6 @@ EOF
 
 <><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
 [git-test] synchronised from git+file://${BASEDIR}/GITREPO
-opam-file                       Missing expected extra files qux.installl at ${BASEDIR}/OPAM/repo/git-test/packages/qux/qux.1/files
 Now run 'opam upgrade' to apply any package updates.
 ### opam list --all --repo=git-test -s
 new-pkg
@@ -411,7 +410,6 @@ rm 'packages/prefix-pkg-b/pkg-b.1/opam'
 
 <><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
 [git-test] synchronised from git+file://${BASEDIR}/GITREPO
-opam-file                       Missing expected extra files qux.installl at ${BASEDIR}/OPAM/repo/git-test/packages/qux/qux.1/files
 Now run 'opam upgrade' to apply any package updates.
 ### opam list --all --repo=git-test -s
 new-pkg
@@ -426,7 +424,6 @@ rm 'packages/pkg-a/pkg-a.1/opam'
 
 <><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
 [git-test] synchronised from git+file://${BASEDIR}/GITREPO
-opam-file                       Missing expected extra files qux.installl at ${BASEDIR}/OPAM/repo/git-test/packages/qux/qux.1/files
 Now run 'opam upgrade' to apply any package updates.
 ### opam list --all --repo=git-test -s
 new-pkg
@@ -440,7 +437,6 @@ renamed-pkg
 
 <><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
 [git-test] synchronised from git+file://${BASEDIR}/GITREPO
-opam-file                       Missing expected extra files qux.installl at ${BASEDIR}/OPAM/repo/git-test/packages/qux/qux.1/files
 Now run 'opam upgrade' to apply any package updates.
 ### opam list --all --repo=git-test -s
 new-pkg
@@ -456,7 +452,6 @@ rm 'packages/prefix-pkg-c/pkg-c.1/files/k/1'
 <><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
 [git-test] synchronised from git+file://${BASEDIR}/GITREPO
 opam-file                       Missing expected extra files 1, k/1 at ${BASEDIR}/OPAM/repo/git-test/packages/prefix-pkg-c/pkg-c.1/files
-opam-file                       Missing expected extra files qux.installl at ${BASEDIR}/OPAM/repo/git-test/packages/qux/qux.1/files
 ### : ^ opam was updated so the missing extra files are detected
 ### opam list --all --repo=git-test -s
 new-pkg
@@ -471,8 +466,6 @@ renamed-pkg
 
 <><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
 [git-test] synchronised from git+file://${BASEDIR}/GITREPO
-opam-file                       Missing expected extra files 1, k/1 at ${BASEDIR}/OPAM/repo/git-test/packages/prefix-pkg-c/pkg-c.1/files
-opam-file                       Missing expected extra files qux.installl at ${BASEDIR}/OPAM/repo/git-test/packages/qux/qux.1/files
 Now run 'opam upgrade' to apply any package updates.
 ### opam list --all --repo=git-test -s
 new-pkg
@@ -487,8 +480,6 @@ renamed-pkg
 
 <><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
 [git-test] synchronised from git+file://${BASEDIR}/GITREPO
-opam-file                       Missing expected extra files 1, k/1 at ${BASEDIR}/OPAM/repo/git-test/packages/prefix-pkg-c/pkg-c.1/files
-opam-file                       Missing expected extra files qux.installl at ${BASEDIR}/OPAM/repo/git-test/packages/qux/qux.1/files
 Now run 'opam upgrade' to apply any package updates.
 ### : Verify pkg-d was removed from repository state.
 ### : This works correctly even if the first processed
@@ -528,9 +519,7 @@ FILE(config)                    Read ${BASEDIR}/OPAM/config in 0.000s
 REPO_BACKEND                    diff: ${BASEDIR}/OPAM/repo/{incremental,incremental.new}
 REPO_BACKEND                    Internal diff (non-empty, 1 changed files) done in 0.00s.
 [incremental] synchronised from file://${BASEDIR}/INCR_REPO
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/incremental/packages/beta/beta.1/opam in 0.000s
 FILE(opam)                      Read ${BASEDIR}/OPAM/repo/incremental/packages/gamma/gamma.1/opam in 0.000s
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/incremental/packages/alpha/alpha.1/opam in 0.000s
 ### opam show gamma.1 --field synopsis
 Gamma package - UPDATED
 ### :III:2: Add a new package - update should process only 1 new file
@@ -544,10 +533,7 @@ FILE(config)                    Read ${BASEDIR}/OPAM/config in 0.000s
 REPO_BACKEND                    diff: ${BASEDIR}/OPAM/repo/{incremental,incremental.new}
 REPO_BACKEND                    Internal diff (non-empty, 1 changed files) done in 0.00s.
 [incremental] synchronised from file://${BASEDIR}/INCR_REPO
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/incremental/packages/beta/beta.1/opam in 0.000s
 FILE(opam)                      Read ${BASEDIR}/OPAM/repo/incremental/packages/delta/delta.1/opam in 0.000s
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/incremental/packages/gamma/gamma.1/opam in 0.000s
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/incremental/packages/alpha/alpha.1/opam in 0.000s
 Now run 'opam upgrade' to apply any package updates.
 ### opam list --all -s --repo=incremental
 alpha
@@ -563,9 +549,6 @@ FILE(config)                    Read ${BASEDIR}/OPAM/config in 0.000s
 REPO_BACKEND                    diff: ${BASEDIR}/OPAM/repo/{incremental,incremental.new}
 REPO_BACKEND                    Internal diff (non-empty, 1 changed files) done in 0.00s.
 [incremental] synchronised from file://${BASEDIR}/INCR_REPO
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/incremental/packages/delta/delta.1/opam in 0.000s
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/incremental/packages/gamma/gamma.1/opam in 0.000s
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/incremental/packages/alpha/alpha.1/opam in 0.000s
 Now run 'opam upgrade' to apply any package updates.
 ### opam list --all -s --repo=incremental
 alpha
@@ -587,7 +570,6 @@ FILE(config)                    Read ${BASEDIR}/OPAM/config in 0.000s
 REPO_BACKEND                    diff: ${BASEDIR}/OPAM/repo/{incremental,incremental.new}
 REPO_BACKEND                    Internal diff (non-empty, 4 changed files) done in 0.00s.
 [incremental] synchronised from file://${BASEDIR}/INCR_REPO
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/incremental/packages/gamma/gamma.1/opam in 0.000s
 FILE(opam)                      Read ${BASEDIR}/OPAM/repo/incremental/packages/zeta/zeta.1/opam in 0.000s
 FILE(opam)                      Read ${BASEDIR}/OPAM/repo/incremental/packages/epsilon/epsilon.1/opam in 0.000s
 FILE(opam)                      Read ${BASEDIR}/OPAM/repo/incremental/packages/alpha/alpha.1/opam in 0.000s
--- a/opam-state.opam
+++ b/opam-state.opam
@@ -32,5 +32,6 @@ depends: [
   "opam-repository" {= version}
   "re" {>= "1.9.0"}
   "spdx_licenses" {>= "1.0.0"}
+  "patch" {>= "3.0.0"}
   "dune" {>= "2.8.0"}
 ]
--- a/tests/reftests/extrafile.test
+++ b/tests/reftests/extrafile.test
@@ -689,14 +689,7 @@ cp p.patch $pkgpath/files/dir/
 
 <><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
 [default] synchronised from file://${BASEDIR}/REPO
-opam-file                       Mismatching extra-files at ${BASEDIR}/OPAM/repo/default/packages/bad-md5/bad-md5.1: wrong checksum (1)
-opam-file                       Missing expected extra files /etc/passwdd at ${BASEDIR}/OPAM/repo/default/packages/escape-absolute/escape-absolute.1/files
-opam-file                       Mismatching extra-files at ${BASEDIR}/OPAM/repo/default/packages/escape-good-md5/escape-good-md5.1: missing from 'files' directory (1)
-opam-file                       Mismatching extra-files at ${BASEDIR}/OPAM/repo/default/packages/good-md5-bad-sha256/good-md5-bad-sha256.1: missing from 'files' directory (1)
 opam-file                       Mismatching extra-files at ${BASEDIR}/OPAM/repo/default/packages/good-md5-good-sha256/good-md5-good-sha256.1: missing from 'files' directory (1)
-opam-file                       Missing extra-files field for p.patch for no-checksum.1, ignoring them.
-opam-file                       Missing extra-files field for p.patch for not-mentioned.1, ignoring them.
-opam-file                       Missing expected extra files p.patch at ${BASEDIR}/OPAM/repo/default/packages/not-present/not-present.1/files
 Now run 'opam upgrade' to apply any package updates.
 ### opam install good-nested-md5.1
 The following actions will be performed:
--- a/tests/reftests/repository.test
+++ b/tests/reftests/repository.test
@@ -123,10 +123,6 @@ FILE(opam)                      Read ${BASEDIR}/OPAM/tarring/.opam-switch/packag
 [tarred] synchronised from file://${BASEDIR}/REPO
 FILE(repo)                      Read ${BASEDIR}/OPAM/repo/tarred/repo in 0.000s
 FILE(opam)                      Read ${BASEDIR}/OPAM/repo/tarred/packages/foo/foo.4/opam in 0.000s
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/tarred/packages/foo/foo.3/opam in 0.000s
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/tarred/packages/foo/foo.2/opam in 0.000s
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/tarred/packages/foo/foo.1/opam in 0.000s
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/tarred/packages/foo/foo.5/opam in 0.000s
 FILE(repos-config)              Wrote ${BASEDIR}/OPAM/repo/repos-config atomically in 0.000s
 CACHE(repository)               Writing the repository cache to ${BASEDIR}/OPAM/repo/state-magicv.cache ...
 CACHE(repository)               ${BASEDIR}/OPAM/repo/state-magicv.cache written in 0.000s
@@ -208,7 +204,6 @@ CACHE(repository)               Loaded ${BASEDIR}/OPAM/repo/state-magicv.cache i
 [repo2] synchronised from file://${BASEDIR}/REPO2
 FILE(repo)                      Read ${BASEDIR}/OPAM/repo/repo2/repo in 0.000s
 FILE(opam)                      Read ${BASEDIR}/OPAM/repo/repo2/packages/bar/bar.2/opam in 0.000s
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/repo2/packages/bar/bar.1/opam in 0.000s
 [tarred] no changes from file://${BASEDIR}/REPO
 FILE(repos-config)              Wrote ${BASEDIR}/OPAM/repo/repos-config atomically in 0.000s
 CACHE(repository)               Writing the repository cache to ${BASEDIR}/OPAM/repo/state-magicv.cache ...
@@ -245,17 +240,10 @@ FILE(opam)                      Read ${BASEDIR}/OPAM/tarring/.opam-switch/packag
 <><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
 [repo2] synchronised from file://${BASEDIR}/REPO2
 FILE(repo)                      Read ${BASEDIR}/OPAM/repo/repo2/repo in 0.000s
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/repo2/packages/bar/bar.2/opam in 0.000s
 FILE(opam)                      Read ${BASEDIR}/OPAM/repo/repo2/packages/bar/bar.3/opam in 0.000s
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/repo2/packages/bar/bar.1/opam in 0.000s
 [tarred] synchronised from file://${BASEDIR}/REPO
 FILE(repo)                      Read ${BASEDIR}/OPAM/repo/tarred/repo in 0.000s
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/tarred/packages/foo/foo.4/opam in 0.000s
 FILE(opam)                      Read ${BASEDIR}/OPAM/repo/tarred/packages/foo/foo.6/opam in 0.000s
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/tarred/packages/foo/foo.3/opam in 0.000s
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/tarred/packages/foo/foo.2/opam in 0.000s
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/tarred/packages/foo/foo.1/opam in 0.000s
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/tarred/packages/foo/foo.5/opam in 0.000s
 FILE(repos-config)              Wrote ${BASEDIR}/OPAM/repo/repos-config atomically in 0.000s
 CACHE(repository)               Writing the repository cache to ${BASEDIR}/OPAM/repo/state-magicv.cache ...
 CACHE(repository)               ${BASEDIR}/OPAM/repo/state-magicv.cache written in 0.000s
@@ -684,7 +672,6 @@ opam-version: "2.0"
 opam-file                       opam-version "5.0" unsupported on ${BASEDIR}/OPAM/repo/repo/packages/bad-opam-version/bad-opam-version.2/opam. Added as dummy unavailable package.
 opam-file                       opam-version "5.0" unsupported on ${BASEDIR}/OPAM/repo/repo/packages/bad-opam-version/bad-opam-version.3/opam. Added as dummy unavailable package.
 [versions] synchronised from file://${BASEDIR}/VRS
-opam-file                       opam-version "5.0" unsupported on ${BASEDIR}/OPAM/repo/versions/packages/bad-opam-version/bad-opam-version.1/opam. Added as dummy unavailable package.
 Now run 'opam upgrade' to apply any package updates.
 ### opam list -A --all-versions -s
 a.1


src/repository/opamRepositoryBackend.ml
```diff
@@ -171,8 +171,20 @@ let get_diff parent_dir dir1 dir2 =
...
     None
   | diffs ->
-    log "Internal diff (non-empty) done in %.2fs." (chrono ());
+    let changed_files =
+      List.map (fun diff ->
+          let file = match diff.Patch.operation with
+            | Patch.Edit (file, _) -> file
+            | Patch.Delete file -> file
+            | Patch.Create file -> file
+            | Patch.Git_ext (file, _, _) -> file
```
kit-ty-kate: i think you want to keep the whole `diff.Patch.operation` around. Otherwise you won't know when a file has been moved

kit-ty-kate: i think this also warrants a test. I don't think we have a test for files being moved around, empty files being deleted and empty files being created in a git repository.

src/state/opamRepositoryState.ml
```diff
@@ -111,6 +111,41 @@ let load_opams_from_dir repo_name repo_root =
...
+  List.fold_left (fun acc file ->
+      let packages_dir = OpamRepositoryPath.packages_dir repo_root in
+      let file_dir = OpamFilename.dirname file in
+      if OpamFilename.dir_starts_with packages_dir file_dir &&
+         OpamFilename.check_suffix file "opam" then
+        let package_dir = OpamFilename.dirname file in
+        (try
+           let nv =
+             OpamPackage.of_string
+               OpamFilename.(Base.to_string (basename_dir package_dir))
```
kit-ty-kate: I'd rather avoid new `open`s
```suggestion
               (OpamFilename.Base.to_string (OpamFilename.basename_dir package_dir))
```

src/state/opamRepositoryState.ml
```diff
@@ -111,6 +111,41 @@ let load_opams_from_dir repo_name repo_root =
     (fun () -> aux OpamPackage.Map.empty (OpamRepositoryPath.packages_dir repo_root))
     ~finally:OpamConsole.clear_status
 
+let load_opams_incremental repo_name repo_root changed_files rt =
+  let existing_opams =
+    try OpamRepositoryName.Map.find repo_name rt.repo_opams
+    with Not_found -> OpamPackage.Map.empty
+  in
+  List.fold_left (fun acc file ->
+      let packages_dir = OpamRepositoryPath.packages_dir repo_root in
```
kit-ty-kate: this line does not depend on the parameters of the fold so it can be moved outside of it to avoid repeating it for nothing

src/repository/opamRepositoryBackend.ml
```diff
@@ -15,7 +15,7 @@ let slog = OpamConsole.slog
 
 type update =
   | Update_full of dirname
-  | Update_patch of filename
+  | Update_patch of (filename * Patch.t list)
```
kit-ty-kate, line 18: `Patch.t` can sometimes get quite large in memory. I think it'd be best to only use the list of operation as we're gonna use more memory during the load
```suggestion
  | Update_patch of (filename * Patch.operation list)
```

src/state/opamRepositoryState.ml
```diff
@@ -111,6 +111,106 @@ let load_opams_from_dir repo_name repo_root =
...
+      Option.map (fun f -> Patch.Delete f) (strip_repo_prefix file)
+    | Patch.Create file ->
+      Option.map (fun f -> Patch.Create f) (strip_repo_prefix file)
+    | Patch.Git_ext (file1, file2, git_ext) ->
+      Option.bind (strip_repo_prefix file1)
+        (fun f1 -> Option.bind (strip_repo_prefix file2)
+            (fun f2 -> Some (Patch.Git_ext (f1, f2, git_ext))))
+  in
+  let process_file acc file ~is_removal =
+    if not (String.ends_with file ~suffix:"opam") then
```
kit-ty-kate:
```suggestion
    if not (String.ends_with file ~suffix:"/opam") then
```

src/state/opamRepositoryState.ml
```diff
@@ -111,6 +111,106 @@ let load_opams_from_dir repo_name repo_root =
...
+  in
+  let packages_dir = OpamRepositoryPath.packages_dir repo_root in
+  let strip_repo_prefix file =
+    match OpamStd.String.cut_at file '/' with
+    | Some (repo_part, path_rest) ->
+      let repo_part =
+        if String.ends_with ~suffix:".new" repo_part then
+          String.sub repo_part 0 (String.length repo_part - 4)
+        else
+          repo_part
```
kit-ty-kate:
```suggestion
        OpamStd.String.remove_suffix ~suffix:".new" repo_part
```

src/repository/opamRepositoryBackend.ml
arozovyk: I kept the `diff.Patch.operation` instead of a file. It seems to be working fine for HTTP/local backends as you can see in reftest i added.

As for git repository test, iI tested it here  https://github.com/ocaml/opam/commit/071a04d46b5eb2e6e122a4125c5986d96c6bb3cd 
There are some things to be addresed: first we have to parse the patch file produced by git diff (`OpamGit.VCS.diff`). I also ran into the fact that only `Git_ext.Rename_only` is properly recognized, while git creation and deletion fallback to `Patch.Create` / `Patch.Delete` instead of `Git_ext.Delete_only` / `Git_ext.Create_only` which is fine since it's the same action in the end, but still quite confusing. (the reason seems to be `patch` ignoring the git headers when `---`/ `+++` are present)

src/state/opamRepositoryState.ml
```diff
@@ -111,6 +111,106 @@ let load_opams_from_dir repo_name repo_root =
...
+      in
+      if
+        String.equal (OpamRepositoryName.to_string repo_name) repo_part
```
kit-ty-kate: i don't think this is true for git repositories. In any case i don't think the whole check is relevant here. Simply removing the part before the first `/` should be sufficient

src/state/opamRepositoryState.ml
```diff
@@ -111,6 +111,106 @@ let load_opams_from_dir repo_name repo_root =
...
+    | None ->
+      log "ERR: Malformed file path (no '/'): %s, ignored." file;
+      None
+  in
+  let filter_and_strip p =
+    match p.Patch.operation with
+    | Patch.Edit (old_file, new_file) ->
+      Option.bind (strip_repo_prefix old_file)
+        (fun o -> Option.bind (strip_repo_prefix new_file)
+            (fun n -> Some (Patch.Edit (o, n))))
```
kit-ty-kate: you won't need it anymore with the above comment but note that `OpamStd.Option.Op` exists for patterns like that

src/repository/opamRepositoryBackend.ml
kit-ty-kate: > first we have to parse the patch file produced by git diff

this should already be done. The parse is done in `OpamRepository.apply_repo_update` for all backends, it would be nice to also avoid duplication with data we already have in the case of the HTTP/local backend.

arozovyk: > empty files being deleted and empty files being created

Not sure I understand right. What would be the expected behaviour ?  New empty opam files will fail on read, thus will be removed. As for deletion - is there a situation where we can end up with an empty opam file in the repo?

src/state/opamRepositoryState.ml
kit-ty-kate: actually this probably should match against the whole `packages/<valid-pkg-name>/<valid-pkg-name>.<valid-version>/opam` pattern

src/repository/opamRepositoryBackend.ml
arozovyk: > The parse is done in done in OpamRepository.apply_repo_update... avoid duplication in the case of the HTTP/local backend.

In `OpamRepositoryBackend` the patch file is created after the we compute `Patch.diff` for each file (diffs that I currently reuse). And for VCS it's the other way around (`Patch.parse` produces the diffs). So I feel like I should reuse the the info from `OpamRepository.apply_repo_update` (by propagating it from `OpamSystem.patch`) and keep the HTTP/local backends as is.

arozovyk: > avoid duplication

Hmm, looked at it a bit more. I think I understand better what you mean for HTTP/local, we currently compute it `fetch_repo_update` then parse it again in `apply_repo_update`

arozovyk:
Updated simplfying a bunch of things. 
We can't avoid parsing for diffs in case of Http and local since we have to first preprocess it, so I ended up reusing the result of `Patch.parse` from `OpamSystem.patch` in general for all backends.

src/repository/opamRepositoryBackend.ml
```diff
@@ -16,6 +16,7 @@ let slog = OpamConsole.slog
 type update =
   | Update_full of dirname
   | Update_patch of filename
+  | Update_diffs of (filename * Patch.t list)
```
kit-ty-kate: Why not simply merge the two and call `Patch.parse` directly in OpamVCS? That's more what i had in mind
```suggestion
  | Update_patch of (filename * Patch.t list)
```

src/client/opamAction.ml
```diff
@@ -367,18 +367,19 @@ let prepare_package_build env opam nv dir =
...
     in
     let rec aux = function
       | [] -> []
       | (patchname,filter)::rest ->
         if OpamFilter.opt_eval_to_bool env filter then
           (print_apply patchname;
            match patch patchname with
-           | None -> aux rest
-           | Some err -> (patchname, err) :: aux rest)
+           | _ when dryrun -> aux rest
```
kit-ty-kate: This changes the behaviour of `--dry-run` slightly, from not applying the patch at all, to applying it but ignoring its result. I think we should keep the previous behaviour for now. If you want to change it i think it should be in a separate PR

src/repository/opamRepositoryBackend.ml
```diff
@@ -106,6 +107,24 @@ let get_files_for_diff parent_dir dir1 dir2 =
...
+let strip_repo_suffix d =
+  let rm_prefix f =
+    match (OpamStd.String.cut_at f '/') with
+    | None ->
+      log "Warning: failed to remove prefix of %s" f;
+      f
+    | Some (_, r) -> r
+  in
+  let operation=
+    match Patch.(d.operation) with
```
kit-ty-kate: you can avoid the `open` like that:
```suggestion
    match d.Patch.operation with
```

src/core/opamSystem.ml
```diff
@@ -1669,25 +1669,27 @@ let internal_patch ~allow_unclean ~patch_filename ~dir diffs =
...
-  if not (Sys.file_exists p) then
-    (OpamConsole.error "Patch file %S not found." p;
-     raise Not_found);
-  let p' =
-    if preprocess then
+let patch ~allow_unclean ~dir patch_source =
+  match patch_source with
+  | `Patch_diffs diffs ->
+    internal_patch ~allow_unclean ~patch_filename:dir ~dir diffs;
+    `Patched (List.map (fun d -> Patch.(d.operation)) diffs)
```
kit-ty-kate: Same here
```suggestion
    `Patched (List.map (fun d -> d.Patch.operation) diffs)
```

src/repository/opamRepository.ml
```diff
@@ -565,10 +569,10 @@ let update repo repo_root =
   | Update_err e -> raise e
   | Update_empty ->
     log "update empty, no validation performed";
-    apply_repo_update repo repo_root Update_empty @@+ fun () ->
+    apply_repo_update repo repo_root Update_empty @@+ fun _ ->
```
kit-ty-kate:
```suggestion
    apply_repo_update repo repo_root Update_empty @@+ fun diffs ->
    assert (diffs = []);
```

src/repository/opamRepositoryBackend.ml
```diff
@@ -171,8 +190,9 @@ let get_diff parent_dir dir1 dir2 =
     log "Internal diff (empty) done in %.2fs." (chrono ());
     None
   | diffs ->
-    log "Internal diff (non-empty) done in %.2fs." (chrono ());
+    log "Internal diff (non-empty, changed files %d) done in %.2fs."
+      (List.length diffs) (chrono ());
```
kit-ty-kate: In 99% of cases (for people that do not use `--debug`) the `List.length` will run across the entire list for nothing. To avoid that `%a` should be used instead.
```suggestion
    log "Internal diff (non-empty, changed files %a) done in %.2fs."
      (fun fmt l -> string_of_int (List.length l)) diffs (chrono ());
```

src/state/opamRepositoryState.ml
```diff
@@ -111,6 +111,53 @@ let load_opams_from_dir repo_name repo_root =
...
+    | Patch.Delete file -> remove_file file acc
+    | Patch.Create file -> add_file file acc
+    | Patch.Git_ext (file1, file2, git_ext) ->
+      match git_ext with
+      | Patch.Rename_only (_, _) -> remove_file file1 acc |> add_file file2
+      | Patch.Delete_only -> remove_file file1 acc
+      | Patch.Create_only -> add_file file2 acc
+  in
+  List.fold_left (fun acc o -> process_operation acc o)
+    existing_opams diffs
```
kit-ty-kate:
```suggestion
  List.fold_left process_operation existing_opams diffs
```

src/repository/opamRepositoryBackend.ml
```diff
@@ -106,6 +107,24 @@ let get_files_for_diff parent_dir dir1 dir2 =
     in
     aux [] files1 files2
 
+(* Needed to be coherent with vcs patches that do not have repo name suffix *)
+let strip_repo_suffix d =
+  let rm_prefix f =
+    match (OpamStd.String.cut_at f '/') with
```
kit-ty-kate:
```suggestion
    match OpamStd.String.cut_at f '/' with
```

src/repository/opamRepositoryBackend.ml
```diff
@@ -106,6 +107,24 @@ let get_files_for_diff parent_dir dir1 dir2 =
...
+    | None ->
+      log "Warning: failed to remove prefix of %s" f;
+      f
+    | Some (_, r) -> r
+  in
+  let operation=
```
kit-ty-kate:
```suggestion
  let operation =
```

src/state/opamRepositoryState.ml
```diff
@@ -111,6 +111,53 @@ let load_opams_from_dir repo_name repo_root =
...
+  let process_file acc file ~is_removal =
+    match OpamPackage.of_filename (OpamFilename.of_string file) with
+    | None ->
+      log "ERR: directory name not a valid package: ignored %s" file;
+      acc
+    | Some nv ->
+      if is_removal then
+        OpamPackage.Map.remove nv acc
+      else
+        let package_dir = OpamFilename.dirname (OpamFilename.Op.(repo_root // file)) in
```
kit-ty-kate:
```suggestion
        let package_dir = OpamFilename.dirname OpamFilename.Op.(repo_root // file) in
```

src/core/opamFilename.mli
```diff
@@ -276,13 +276,19 @@ val remove_prefix_dir: Dir.t -> Dir.t -> string
...
+    Returns [`Patched operations] on success with the list of applied patch
+    operations, or [`Exception exn] on error.
 
     @param allow_unclean decides if applying a patch on a directory which
     differs slightly from the one described in the patch file is allowed.
     Allowing unclean applications imitates the default behaviour of GNU Patch. *)
-val patch: ?preprocess:bool -> allow_unclean:bool -> t -> Dir.t -> exn option
+val patch: allow_unclean:bool ->
+  [`Patch_file of string | `Patch_diffs of Patch.t list ] -> Dir.t ->
+  [`Exception of exn | `Patched of Patch.operation list ]
```
kit-ty-kate: The shape of this type looks very much like `result`. Is there a reason to not simply use it?
```suggestion
  (Patch.operation list, exn) result
```

src/core/opamFilename.mli
```diff
@@ -276,13 +276,19 @@ val remove_prefix_dir: Dir.t -> Dir.t -> string
...
+
+    Returns [`Patched operations] on success with the list of applied patch
+    operations, or [`Exception exn] on error.
 
     @param allow_unclean decides if applying a patch on a directory which
     differs slightly from the one described in the patch file is allowed.
     Allowing unclean applications imitates the default behaviour of GNU Patch. *)
-val patch: ?preprocess:bool -> allow_unclean:bool -> t -> Dir.t -> exn option
+val patch: allow_unclean:bool ->
+  [`Patch_file of string | `Patch_diffs of Patch.t list ] -> Dir.t ->
```
kit-ty-kate: Once `Update_patch` and `Update_diffs` merged, i believe these can be merged the same way as well

src/state/opamRepositoryState.ml
```diff
@@ -111,6 +111,53 @@ let load_opams_from_dir repo_name repo_root =
...
+          log "ERR: Could not read opam file %s, ignored" file;
+          acc
+  in
+  let remove_file file acc = process_file acc file ~is_removal:true in
+  let add_file file acc = process_file acc file ~is_removal:false in
+  let process_operation acc = function
+    | Patch.Edit (old_file, new_file) ->
+      if OpamFilename.Dir.equal
+          (OpamFilename.Dir.of_string old_file)
+          (OpamFilename.Dir.of_string new_file)
```
kit-ty-kate: This seems wrong. These files should never be directories

src/state/opamRepositoryState.ml
```diff
@@ -111,6 +111,53 @@ let load_opams_from_dir repo_name repo_root =
     (fun () -> aux OpamPackage.Map.empty (OpamRepositoryPath.packages_dir repo_root))
     ~finally:OpamConsole.clear_status
 
+let load_opams_incremental repo_name repo_root diffs rt =
+  let existing_opams =
+    match OpamRepositoryName.Map.find_opt repo_name rt.repo_opams with
+    | Some opams -> opams
+    | None -> OpamPackage.Map.empty
+  in
+  let process_file acc file ~is_removal =
```
kit-ty-kate: I'm not sure if it's done elsewhere or something, but this function seems to assume every files that can change are opam files. What about changes from their `files` directory or other non-opam file related changes?

tests/reftests/repository-update.test
```diff
@@ -0,0 +1,181 @@
+N0REP0
```
kit-ty-kate: These tests should be added all at once in the first commit of the PR, ran to unsure the initial state is correct, then ran again for each new commits to show new behaviours

kit-ty-kate (commented):
First pass of review. I haven't looked at everything yet but i think that's enough for now

src/state/opamRepositoryState.ml
arozovyk: The function purpose is to load opam files to update their definitions in `rt`, all other files are ignored as it was already the case with original `load_opams_from_dir` function. Any other extra files are read by `OpamFileTools.read_repo_opam` and stored through metadata

src/core/opamFilename.mli
arozovyk: In `OpamAction.prepare_package_build` we use it with `'Patch_file`to apply patches trough a file (that has to be parsed), otherwise we use `'Patch_diffs` and avoid parsing completely for Local and HTTP, VCS patch file parsing being done in `OpamVCS` as you suggested

src/repository/opamRepositoryBackend.ml
arozovyk: 35c43fb43f7163df21d54d513185ca875a9b8722

tests/lib/patchDiff.ml
```diff
@@ -372,25 +372,26 @@ let diff_patch dir setup =
...
       | None -> print "No diff\n"; None
-      | some -> some
+      | Some (f,_) -> Some f
   in
   match diff with
   | None -> ()
   | Some diff ->
     if git then make_git_repo dir;
-    print "%s\n" (OpamFilename.read diff);
+    print "%s\n" (OpamFilename.to_string diff);
```
kit-ty-kate: i believe this change, breaks the test (file content vs filename)

src/repository/opamRepositoryBackend.ml
```diff
@@ -171,8 +189,9 @@ let get_diff parent_dir dir1 dir2 =
     log "Internal diff (empty) done in %.2fs." (chrono ());
     None
   | diffs ->
-    log "Internal diff (non-empty) done in %.2fs." (chrono ());
+    log "Internal diff (non-empty, changed files %a) done in %.2fs."
```
kit-ty-kate:
```suggestion
    log "Internal diff (non-empty, %a changed files) done in %.2fs."
```

src/repository/opamRepositoryBackend.mli
```diff
@@ -18,9 +18,9 @@ type update =
   | Update_full of dirname
   (** No previous known state, the full contents have been put in the given
       temporary directory *)
-  | Update_patch of filename
-  (** The given patch file corresponds to the update, i.e. applying it to the
-      local repository with 'patch -p1' would get it to the upstream state *)
+  | Update_patch of (filename * Patch.t list)
+  (** The list of file-level operation to apply to a local repository in order
```
kit-ty-kate:
```suggestion
  (** A patch file and its list of file-level operation to apply to a local repository in order
```

src/repository/opamRepository.ml
```diff
@@ -524,38 +524,37 @@ let apply_repo_update repo repo_root = function
...
     OpamConsole.msg "[%s] synchronised from %s\n"
       (OpamConsole.colorise `green
          (OpamRepositoryName.to_string repo.repo_name))
       (OpamUrl.to_string repo.repo_url);
     log "%a: applying patch update at %a"
       (slog OpamRepositoryName.to_string) repo.repo_name
       (slog OpamFilename.to_string) f;
-    let preprocess =
-      match repo.repo_url.OpamUrl.backend with
-      | `http | `rsync -> false
```
kit-ty-kate, line 538: i don't think we should start running the the preprocessing for local and http repositories. Could you bring the `?preprocess` argument back?

src/core/opamFilename.mli
kit-ty-kate: If this signature is to be kept, i think it should at least convert the `Patch_file of string` to `Patch_file of OpamFilename.t`

tests/reftests/repository-update.test
kit-ty-kate: we already have a reftest called `update.test` i think it should probably go there instead

tests/reftests/repository-update.test
```diff
@@ -0,0 +1,181 @@
...
+### git -C GITREPO add -A
+### git -C GITREPO commit -qm "initial commit"
+### opam repository add git-test --kind=git git+file://${BASEDIR}/GITREPO --this-switch
+[git-test] Initialised
+### opam list --all --all-versions --repo=git-test -s
+delete-pkg.1
+rename-pkg.1
+### : Git rename of opam file
+### mkdir -p GITREPO/packages/rename-pkg/renamed-pkg.1
+### git -C GITREPO mv packages/rename-pkg/rename-pkg.1/opam packages/rename-pkg/renamed-pkg.1/opam
```
kit-ty-kate: for consistency
```suggestion
### git -C GITREPO mv packages/rename-pkg/rename-pkg.1/opam packages/renamed-pkg/renamed-pkg.1/opam
```

src/repository/opamRepository.ml
kit-ty-kate: Maybe this PR depends on the completion of https://github.com/ocaml/opam/issues/6599 first

src/state/opamUpdate.ml
```diff
@@ -67,7 +67,7 @@ let repository rt repo =
     in
     OpamProcess.Job.with_text text @@
     OpamRepository.update r repo_root @@+ fun has_changes ->
-    let has_changes = if redirect then `Changes else has_changes in
+    let has_changes = if redirect then `Changes [] else has_changes in
```
kit-ty-kate, line 70: That seems unnecessary. @rjbou do you know why this line was added in https://github.com/ocaml/opam/pull/5146 ?

tests/reftests/repository-update.test
```diff
@@ -0,0 +1,181 @@
...
+Now run 'opam upgrade' to apply any package updates.
+### opam list --all --all-versions -s
+baz.1
+foo.1
+newpkg.1
+### : File edit
+### <REPO/packages/foo/foo.1/opam>
+opam-version: "2.0"
+synopsis: "Updated package"
+### opam update default
```
kit-ty-kate: it would be nice if the new incremental behaviour was shown in the test. We could select the right OPAMDEBUGSECTION to show that only a limited number of files are being read

src/core/opamFilename.mli
arozovyk: changed in 3567e8ae2999193aafd208be15e24f683ed8d522
Using `OpamFilename.t` in `OpamSystem` creates a cycle. Moved some logic and did some refactoring instead.

kit-ty-kate: > Using `OpamFilename.t` in `OpamSystem` creates a cycle

of course, if you look at the rest of the module, one of its goal is to abstract `OpamSystem` functions on top of `OpamFilename.t` and `OpamFilename.Dir.t`. The way you did it goes in the same direction and was indeed what i meant. Thanks!

src/repository/opamRepository.ml
arozovyk: We are not, this is naturally done by the fact that `OpamSystem.parse_patch_file` is only called in `OpamVCS` (in `OpamAction.prepare_package_build` as well, but that's a different story), for local and http repositories we reuse the diffs directly.

src/core/opamSystem.ml
```diff
@@ -1597,18 +1597,21 @@ let translate_patch ~dir orig corrected =
 
 exception Internal_patch_error of string
 
-let internal_patch ~allow_unclean ~patch_filename ~dir diffs =
+let patch ~allow_unclean ?patch_filename ~dir diffs =
   let internal_patch_error fmt =
     Printf.ksprintf (fun str -> raise (Internal_patch_error str)) fmt
   in
+  let patch_info_path =
+    OpamStd.Option.default ("in directory"^dir) patch_filename
```
arozovyk:
```suggestion
    OpamStd.Option.default ("in directory "^dir) patch_filename
```

master_changes.md
arozovyk, line 1: To be updated on rebase

tests/reftests/repository-update.test
arozovyk: c656c890c283a8cfd11316e16b08cb6b861280b1

tests/lib/patchDiff.ml
```diff
@@ -382,15 +382,14 @@ let diff_patch dir setup =
     let apply ~git diff =
       let git = if git then "GIT " else "" in
       let result =
-        OpamFilename.patch ~allow_unclean:false diff
-          (dir / first)
+        OpamFilename.patch ~allow_unclean:false(`Patch_file diff) (dir / first)
```
kit-ty-kate: \> 80 characters + minor missing space
```suggestion
        OpamFilename.patch ~allow_unclean:false (`Patch_file diff)
          (dir / first)
```

src/repository/opamVCS.ml
```diff
@@ -52,7 +52,13 @@ module Make (VCS: VCS) = struct
...
       @@| function
       | None -> OpamRepositoryBackend.Update_empty
-      | Some patch -> OpamRepositoryBackend.Update_patch patch
+      | Some patch_file ->
+        let p = OpamFilename.to_string patch_file in
+        try
+          let diffs = OpamSystem.parse_patch_file
+              ~dir:(OpamFilename.Dir.to_string repo_root) p in
+          OpamRepositoryBackend.Update_patch (patch_file, diffs)
+        with exn -> Update_err exn
```
kit-ty-kate, line 61: If we had `OpamFilename.parse_patch_file` for a more `OpamFilename` friendly API we could simplify to:
```suggestion
      | Some patch_file ->
        match OpamFilename.parse_patch_file ~dir:repo_root patch_file with
        | diffs -> OpamRepositoryBackend.Update_patch (patch_file, diffs)
        | exception exn -> Update_err exn
```

src/repository/opamRepositoryBackend.ml
```diff
@@ -106,6 +106,24 @@ let get_files_for_diff parent_dir dir1 dir2 =
...
+      log "Warning: failed to remove prefix of %s" f;
+      f
+    | Some (_, r) -> r
+  in
+  let operation =
+    match d.Patch.operation with
+    | Patch.Create f -> Patch.Create (rm_prefix f)
+    | Patch.Delete f -> Patch.Delete (rm_prefix f)
+    | Patch.Edit (f1, f2) -> Patch.Edit (rm_prefix f1, rm_prefix f2)
+    | Patch.Git_ext (f1, f2, ext) -> Patch.Git_ext (rm_prefix f1, rm_prefix f2, ext)
```
kit-ty-kate: \> 80 columns
```suggestion
    | Patch.Git_ext (f1, f2, ext) ->
      Patch.Git_ext (rm_prefix f1, rm_prefix f2, ext)
```

tests/reftests/update.test
```diff
@@ -5,5 +5,248 @@ opam-version: "2.0"
 
 <><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
 [default] synchronised from file://${BASEDIR}/REPO
-Processing: [default: loading data]
 Now run 'opam upgrade' to apply any package updates.
+### <REPO/repo>
```
kit-ty-kate, line 57:
```suggestion
### : Test incrementality 
### rm -rf REPO
### <REPO/repo>
```

kit-ty-kate (commented):
lgtm overall, however it's a hard PR to review without its final clean commit structure so i'd rather wait for whenever this is finished.

I also think this PR should be queued on a PR adding a benchmark test for opam update as we currently don't have any. It would ideal to have a test for three kinds of cases:
- no diff
- small diff (just a few files)
- large diff (e.g. from [ocaml/opam-repository#2025-01-before-archiving-phase1](https://github.com/ocaml/opam-repository/releases/tag/2025-01-before-archiving-phase1) to now)

with both local path and local git repositories

src/state/opamUpdate.ml
arozovyk: I think it's needed to update the repo config on redirection even though there are no actual changes.

master_changes.md
```diff
@@ -48,6 +48,7 @@ users)
 ## Var/Option
 
 ## Update / Upgrade
+  * Implement incremental opam file loading to process only changed files during repository updates [#6614 @arozovyk]
```
```suggestion
  * Implement incremental opam file loading to process only changed files during repository updates [#6614 @arozovyk - fix #5824]
```

src/client/opamRepositoryCommand.ml
```diff
@@ -271,6 +271,8 @@ let update_with_auto_upgrade rt repo_names =
              OpamFile.Repo.with_root_url r.repo_url
            in
            let opams =
+             (* TODO: propagate changed files from opam update to load opams
+                incrementally *)
```
kit-ty-kate: Could you open a ticket for that? If there is a TODO in the code it's unlikely we'll see it and fix it so it's better as its own ticket

src/repository/opamRepositoryBackend.ml
```diff
@@ -106,6 +106,25 @@ let get_files_for_diff parent_dir dir1 dir2 =
     in
     aux [] files1 files2
 
+(* Needed to be coherent with vcs patches that do not have repo name suffix *)
+let strip_repo_suffix d =
```
kit-ty-kate: this is probably something that might be useful as part of the `patch` library. Could you open a ticket to https://github.com/hannesm/patch/issues to keep track of it?

src/core/opamSystem.mli
```diff
@@ -372,6 +372,11 @@ val get_eol_encoding : string -> bool option
 *)
 val translate_patch: dir:string -> string -> string -> unit
 
+(** [parse_patch_file ~dir patch_file] processes and parses a patch file.
+    Returns the parsed patch diffs or raises an exception if the patch file
+    doesn't exist or can't be parsed. *)
```
rjbou, line 378: it worth explicit the exceptions

src/repository/opamRepositoryBackend.ml
```diff
@@ -106,6 +106,25 @@ let get_files_for_diff parent_dir dir1 dir2 =
...
+  let rm_prefix f =
+    match OpamStd.String.cut_at f '/' with
```
rjbou, line 113: Are we sure that it is always `/` ? Maybe it worth using a prefix path removal function ?

src/repository/opamRepositoryBackend.ml
```diff
@@ -106,6 +106,25 @@ let get_files_for_diff parent_dir dir1 dir2 =
     in
     aux [] files1 files2
 
+(* Needed to be coherent with vcs patches that do not have repo name suffix *)
```
rjbou: In the comment you can also add more information about what does the function

src/repository/opamRepositoryBackend.ml
```diff
@@ -106,6 +106,25 @@ let get_files_for_diff parent_dir dir1 dir2 =
...
+let strip_repo_suffix d =
+  let rm_prefix f =
+    match OpamStd.String.cut_at f '/' with
+    | None ->
+      log "Warning: failed to remove prefix of %s" f;
```
rjbou: Maybe give more information on the context
```suggestion
  log "internal diff: failed to remove prefix of %s" f;
```
I don't know if it worth having it at log level 1, @kit-ty-kate ?

src/repository/opamRepositoryBackend.ml
rjbou:
```suggestion
let strip_repo_suffix patch =

File: src/repository/opamRepositoryBackend.ml [hunk 43]
```diff
@@ -171,8 +190,9 @@ let get_diff parent_dir dir1 dir2 =
     log "Internal diff (empty) done in %.2fs." (chrono ());
     None
   | diffs ->
-    log "Internal diff (non-empty) done in %.2fs." (chrono ());
+    log "Internal diff (non-empty, %a changed files) done in %.2fs."
+      (fun fmt l -> Format.pp_print_int fmt (List.length l)) diffs (chrono ());
```
rjbou: You can use `OpamConsole.slog` i think here

src/repository/opamRepositoryBackend.mli
```diff
@@ -18,9 +18,9 @@ type update =
   | Update_full of dirname
   (** No previous known state, the full contents have been put in the given
       temporary directory *)
-  | Update_patch of filename
-  (** The given patch file corresponds to the update, i.e. applying it to the
-      local repository with 'patch -p1' would get it to the upstream state *)
+  | Update_patch of (filename * Patch.t list)
+  (** A patch file and its and its list of file-level operation to apply to a
```
rjbou:
```suggestion
(** A patch file and its list of file-level operation to apply to a

File: src/repository/opamRepositoryBackend.mli [hunk 45]
```diff
@@ -18,9 +18,9 @@ type update =
...
+      local repository in order to get it to the upstream state *)
```
rjbou: Is there a reason to remove the `patch -p1` information ?

src/state/opamRepositoryState.mli
```diff
@@ -51,6 +51,18 @@ val get_repo: 'a repos_state -> repository_name -> repository
 
 val load_opams_from_dir: repository_name -> dirname -> OpamFile.OPAM.t OpamPackage.Map.t
 
+(** [load_opams_incremental repo_name repo_root diffs rt] incrementally
+      updates package definitions by processing only changed files.
+
+      Starting from existing packages in [rt.repo_opams], it:
+      - Adds/updates packages from changed opam files that exist
+      - Removes packages whose opam files were deleted
```
rjbou:
```suggestion
      - Adds new packages
      - Updates packages from changed opam files that exist
      - Removes packages whose opam files were deleted

File: src/state/opamRepositoryState.mli [hunk 47]
```diff
@@ -51,6 +51,18 @@ val get_repo: 'a repos_state -> repository_name -> repository
...
+
+      @param diffs List of [Patch.operation] representing creations, deletions, and
+      modifications of regular files *)
```
rjbou:
```suggestion
      modifications of regular files
*)

File: src/state/opamRepositoryState.mli [hunk 48]
```diff
@@ -51,6 +51,18 @@ val get_repo: 'a repos_state -> repository_name -> repository
...
+val load_opams_incremental: repository_name -> dirname -> Patch.operation list ->
+  'a repos_state -> OpamFile.OPAM.t package_map
```
rjbou:
```suggestion
val load_opams_incremental:
  repository_name -> dirname -> Patch.operation list ->'a repos_state
  -> OpamFile.OPAM.t package_map

File: src/state/opamRepositoryState.ml [hunk 49]
```diff
@@ -110,6 +110,52 @@ let load_opams_from_dir repo_name repo_root =
     (fun () -> aux OpamPackage.Map.empty (OpamRepositoryPath.packages_dir repo_root))
     ~finally:OpamConsole.clear_status
 
+let load_opams_incremental repo_name repo_root diffs rt =
+  let existing_opams =
+    match OpamRepositoryName.Map.find_opt repo_name rt.repo_opams with
+    | Some opams -> opams
+    | None -> OpamPackage.Map.empty
```
rjbou: Is it possible that we load opam files incrementally for an unregistered repository ?

src/state/opamRepositoryState.ml
```diff
@@ -110,6 +110,52 @@ let load_opams_from_dir repo_name repo_root =
     (fun () -> aux OpamPackage.Map.empty (OpamRepositoryPath.packages_dir repo_root))
     ~finally:OpamConsole.clear_status
 
+let load_opams_incremental repo_name repo_root diffs rt =
+  let existing_opams =
```
rjbou, line 133:
```suggestion
  if OpamConsole.disp_status_line () || OpamConsole.verbose () then
    OpamConsole.status_line "Processing: [%s: loading data]"
      (OpamConsole.colorise `blue (OpamRepositoryName.to_string repo_name));
  let existing_opams =
```
or another text to differentiate from `load_opams_from_dir`
with at the end of the function a cleatr status
```
  Fun.protect
    (fun () -> 
      List.fold_left process_operation existing_opams diffs)
    ~finally:OpamConsole.clear_status

File: src/state/opamRepositoryState.ml [hunk 51]
```diff
@@ -110,6 +110,52 @@ let load_opams_from_dir repo_name repo_root =
...
+    match OpamRepositoryName.Map.find_opt repo_name rt.repo_opams with
+    | Some opams -> opams
+    | None -> OpamPackage.Map.empty
+  in
+  let process_file acc file ~is_removal =
```
```suggestion
  let process_file opams file ~is_removal =

File: src/state/opamRepositoryState.ml [hunk 52]
```diff
@@ -110,6 +110,52 @@ let load_opams_from_dir repo_name repo_root =
...
+        in
+        match
+          OpamFileTools.read_repo_opam ~repo_name ~repo_root package_dir
+        with
+        | Some opam ->
+          OpamPackage.Map.add nv opam acc
+        | None ->
+          log "ERR: Could not read opam file %s, ignored" file;
+          acc
+  in
```
rjbou: It worth factorising this function with the content of `load_opams_from_dir`

src/state/dune
```diff
@@ -1,7 +1,7 @@
 (library
   (name        opam_state)
   (public_name opam-state)
-  (libraries   opam-repository re spdx_licenses)
+  (libraries   opam-repository re spdx_licenses patch)
```
rjbou, line 4: missing opam file update ?

src/state/opamUpdate.ml
```diff
@@ -134,7 +134,12 @@ let repository rt repo =
         (Printexc.to_string e)
     | None ->
       let opams =
-        OpamRepositoryState.load_opams_from_dir repo.repo_name repo_root
+        match diffs with
+        | [] ->
+          OpamRepositoryState.load_opams_from_dir repo.repo_name repo_root
+        | files ->
+          OpamRepositoryState.load_opams_incremental repo.repo_name repo_root
+            files rt
```
rjbou: You have as argument rt and rt internals (repo_root can be retrieved from rt, repo name from repo & rt). Either it's a raw function that takes reponame, reporoot and opams, or it's a function that uses rt and take repo & rt.

src/state/opamUpdate.ml
```diff
@@ -134,7 +134,12 @@ let repository rt repo =
         (Printexc.to_string e)
     | None ->
       let opams =
-        OpamRepositoryState.load_opams_from_dir repo.repo_name repo_root
+        match diffs with
+        | [] ->
+          OpamRepositoryState.load_opams_from_dir repo.repo_name repo_root
+        | files ->
```
rjbou: Is it files or diffs ?

master_changes.md
```diff
@@ -162,9 +166,13 @@ users)
 
 ## opam-repository
   * `OpamLocal.rsync_*`: Change the return type from `OpamFilename.*` to `unit` [#6658 @kit-ty-kate]
+  * `OpamRepository.update` changed the `'Changes` return type to include `Patch.operation list` of changes. [#6614 @arozovyk]
+  * `OpamRepositoryBackend.update` type : include the `Patch.t list` in `Update_patch` variant [#6614 @arozovyk]
+  * `OpamRepositoryBackend.get_diff` include `Patch.t list` in the return type (along with `filename`) [#6614 @arozovyk]
```
```suggestion
  * `OpamRepository.update`: changed the `'Changes` return type to include `Patch.operation list` of changes. [#6614 @arozovyk]
  * `OpamRepositoryBackend.update` type : include the `Patch.t list` in `Update_patch` variant [#6614 @arozovyk]
  * `OpamRepositoryBackend.get_diff`: include `Patch.t list` in the return type (along with `filename`) [#6614 @arozovyk]
```

master_changes.md
```diff
@@ -162,9 +166,13 @@ users)
...
 
 ## opam-state
   * `OpamSwitchState.files`: was removed [#6662 @kit-ty-kate]
+  * Add `OpamRepositoryState.load_opams_incremental` to update package definitions based on file change operations [#6614 @arozovyk]
```
```suggestion
  * `OpamRepositoryState`: add `load_opams_incremental` to update package definitions based on file change operations (diff) [#6614 @arozovyk]
```

master_changes.md
```diff
@@ -190,3 +198,7 @@ users)
   * `OpamStd.String.contains_char`: was removed. Use `Stdlib.String.contains` instead. [#6442 @kit-ty-kate]
   * `OpamStd.String.map`: was removed. Use `Stdlib.String.map` instead. [#6442 @kit-ty-kate]
   * `OpamStd.String.{starts_with,ends_with,for_all,fold_left}`: were moved to `OpamCompat.String` [#6442 @kit-ty-kate]
+  * Add `OpamFilename.parse_patch_file` that preprocesses and parses a patch file. [#6614 @arozovyk]
+  * Add lower-level `OpamSystem.parse_patch_file` that preprocesses and parses a patch file. [#6614 @arozovyk]
+  * `OpamFilename.patch` use variants to make the input either `Filename.t` or reuse `Patch.diffs` directly. Remove the `?preprocess` argument since the preprocess logic is moved to the `OpamFilename.parse_patch_file` function that is called only in `OpamVCS` (mirroring the previous logic). [#6614 @arozovyk]
+  * `OpamSystem.patch` change the signaure to work directly with `Patch.diffs` (implementation is now the previouly `internal_patch` function), parsing is now done separetly. [#6614 @arozovyk]
```
```suggestion
  * `OpamFilename`: add `parse_patch_file` that preprocesses and parses a patch file. [#6614 @arozovyk]
  * `OpamSystem`: add lower-level `parse_patch_file` that preprocesses and parses a patch file. [#6614 @arozovyk]
  * `OpamFilename.patch`: use variants to make the input either `Filename.t` or reuse `Patch.diffs` directly. Remove the `?preprocess` argument since the preprocess logic is moved to the `OpamFilename.parse_patch_file` function that is called only in `OpamVCS` (mirroring the previous logic). [#6614 @arozovyk]
  * `OpamSystem.patch` change the signaure to work directly with `Patch.diffs` (implementation is now the previouly `internal_patch` function), parsing is now done separetly. [#6614 @arozovyk]
```

src/core/opamFilename.mli
```diff
@@ -374,3 +377,8 @@ end
 
 (** Convert a filename to an attribute, relatively to a root *)
 val to_attribute: Dir.t -> t -> Attribute.t
+
+(** [parse_patch_file ~dir patch_file] processes and parses a patch file.
+    Returns the parsed patch diffs or raises an exception if the patch file
+    doesn't exist or can't be parsed. *)
+val parse_patch_file : dir:Dir.t -> t -> Patch.t list
```
rjbou: to move near `patch`

src/core/opamSystem.mli
```diff
@@ -372,6 +372,11 @@ val get_eol_encoding : string -> bool option
...
+val parse_patch_file: dir:string -> string -> Patch.t list
```
rjbou:
```suggestion
val parse_patch: dir:string -> file:string -> Patch.t list
```
suggestion

src/core/opamFilename.mli
```suggestion
val parse_patch : dir:Dir.t -> t -> Patch.t list
```
suggestion, the argument is a filename

src/core/opamSystem.mli
```diff
@@ -348,14 +348,14 @@ val get_lock_fd: lock -> Unix.file_descr
...
-    apply.
+(** [patch ~allow_unclean ?patch_filename ~dir diffs] applies a patch to
+    directory [dir].
 
     @param allow_unclean decides if applying a patch on a directory which
     differs slightly from the one described in the patch file is allowed.
     Allowing unclean applications imitates the default behaviour of GNU Patch. *)
-val patch: ?preprocess:bool -> allow_unclean:bool -> dir:string -> string -> exn option
+val patch: allow_unclean:bool -> ?patch_filename:string -> dir:string ->
+  Patch.t list -> unit
```
rjbou:
```suggestion
val patch:
  allow_unclean:bool -> ?patch_filename:string -> dir:string
  -> Patch.t list -> unit
```

src/repository/opamRepository.ml
```diff
@@ -565,7 +564,8 @@ let update repo repo_root =
   | Update_err e -> raise e
   | Update_empty ->
     log "update empty, no validation performed";
-    apply_repo_update repo repo_root Update_empty @@+ fun () ->
+    apply_repo_update repo repo_root Update_empty @@+ fun diffs ->
+    assert (diffs = []);
```
rjbou, line 568: I don't know if it worth having a internal error that is more verbose

src/repository/opamRepository.mli
```diff
@@ -23,10 +23,12 @@ val packages_with_prefixes: dirname -> string option package_map
...
 
 (** Update {i $opam/repo/$repo}. Returns [`No_changes] if the update did not
-    bring any changes, and [`Changes] otherwise.
+    bring any changes, and [`Changes of Patch.operation list]
+    representing creations, deletions, and modifications of regular files.
 
     @raise Failure in case the update couldn't be achieved. *)
-val update: repository -> dirname -> [`Changes | `No_changes] OpamProcess.job
+val update: repository -> dirname -> [`Changes of Patch.operation list
+                                     |`No_changes] OpamProcess.job
```
rjbou:
```suggestion
val update:
  repository -> dirname
  -> [`Changes of Patch.operation list |`No_changes] OpamProcess.job
```

src/repository/opamRepositoryBackend.ml
rjbou:
```suggestion
(* Needed to be coherent with vcs patches that do not have repo name suffix ".new" *)
```
?

src/state/opamRepositoryState.mli
```diff
@@ -51,6 +51,18 @@ val get_repo: 'a repos_state -> repository_name -> repository
...
+(** [load_opams_incremental repo_name repo_root diffs rt] incrementally
+      updates package definitions by processing only changed files.
+
+      Starting from existing packages in [rt.repo_opams], it:
+      - Adds/updates packages from changed opam files that exist
+      - Removes packages whose opam files were deleted
+
+      @param diffs List of [Patch.operation] representing creations, deletions, and
+      modifications of regular files *)
+val load_opams_incremental: repository_name -> dirname -> Patch.operation list ->
```
rjbou:
```suggestion
val load_opams_from_diff: repository_name -> dirname -> Patch.operation list ->
```
?
To mimic `load_opams_from_dir`

src/state/opamUpdate.ml
rjbou: I don't remember, I need to investigate for that

src/repository/opamRepositoryBackend.ml
arozovyk: `OpamFilename.remove_prefix` ? Not sure this works since I have to specify a prefix that I would still have to extract from the string. Or do you have something else in mind?

src/repository/opamRepositoryBackend.ml
kit-ty-kate: i think it's fine to have at level 1. We could even think about raising a proper warning if this happens given it would mean something went extremely wrong while parsing the patch file

src/repository/opamRepositoryBackend.ml
kit-ty-kate: it is always `/` (it's part of the patch definition). `remove_prefix` shouldn't be used here.

src/state/opamRepositoryState.ml
arozovyk: Should we fail here instead?

src/state/opamUpdate.ml
arozovyk: changed to 
```ocaml
val load_opams_from_diff:
  repository -> Patch.operation list -> 'a repos_state
  -> OpamFile.OPAM.t package_map
  ```
  Had to move up the `get_root` functions for this one

File: src/state/opamRepositoryState.ml (hunk 49 above)
Comment (kit-ty-kate) [reply src/state/opamRepositoryState.ml]: yes i'd say let it fail with not_found (just use `OpamRepositoryName.Map.find` instead of `OpamRepositoryName.Map.find_opt`)

File: src/repository/opamRepositoryBackend.ml (hunk 38 above)
Comment (kit-ty-kate) [reply src/repository/opamRepositoryBackend.ml]: Done in https://github.com/hannesm/patch/issues/34

File: src/repository/opamRepository.ml (hunk 62 above)
Comment (kit-ty-kate) [reply src/repository/opamRepository.ml:568]: i don't think so, this is something that pragmatically will not happen, we just don't have the GADTs to encode the invariant

File: src/state/opamRepositoryState.ml [hunk 65]
```diff
@@ -110,6 +121,57 @@ let load_opams_from_dir repo_name repo_root =
     (fun () -> aux OpamPackage.Map.empty (OpamRepositoryPath.packages_dir repo_root))
     ~finally:OpamConsole.clear_status
 
+let load_opams_from_diff repo diffs rt =
+  if OpamConsole.disp_status_line () || OpamConsole.verbose () then
+    OpamConsole.status_line "Processing: [%s: loading changed data]"
```
kit-ty-kate: i think this has a bag ring to it and also confusing for users, i'd rather keep the previous vague "loading data"

src/state/opamUpdate.ml
kit-ty-kate: For what purpose was this function changed?

arozovyk: Because the suggestion 
>  it's a function that uses rt and take repo & rt

Made sense to me. You don't agree?

src/state/opamRepositoryState.ml
arozovyk: Reverted.

src/state/opamRepositoryState.ml
arozovyk: https://github.com/ocaml/opam/pull/6614/commits/513e5468900d02c5ece0d005cff686a9ab1b9175#diff-32b2e29b19f214e4ace672d2361c5510d7a143b50930824b23c53d40f22c43ddR90

src/state/opamRepositoryState.ml
```diff
@@ -76,6 +76,35 @@ module Cache = struct
...
+let get_repo_root rt repo =
+  get_root_raw rt.repos_global.root rt.repos_tmp repo.repo_name
+
+let read_package_opam ~repo_name ~repo_root package_dir =
+  match OpamFileTools.read_repo_opam ~repo_name ~repo_root package_dir with
+  | Some opam ->
+    (try
+       let nv =
+         OpamPackage.of_string
+           OpamFilename.(Base.to_string (basename_dir package_dir))
```
kit-ty-kate:
```suggestion
           OpamFilename.Base.to_string (OpamFilename.basename_dir package_dir)
```

src/state/opamRepositoryState.ml
```diff
@@ -76,6 +76,35 @@ module Cache = struct
...
+       in
+       Some (nv, opam)
+     with Failure _ ->
+       log "ERR: directory name not a valid package: ignored %s"
+         OpamFilename.(to_string Op.(package_dir // "opam"));
```
kit-ty-kate:
```suggestion
         OpamFilename.to_string OpamFilename.Op.(package_dir // "opam");
```

src/state/opamRepositoryState.ml
```diff
@@ -76,6 +76,35 @@ module Cache = struct
...
+       None)
+  | None ->
+    log "ERR: Could not load %s, ignored"
+      OpamFilename.(to_string Op.(package_dir // "opam"));
```
kit-ty-kate:
```suggestion
      OpamFilename.to_string OpamFilename.Op.(package_dir // "opam");
```

src/state/opamUpdate.ml
kit-ty-kate: I don't really care all that much, it's just that i didn't think that was what @rjbou meant in her comment so i was surprised

kit-ty-kate (approved):
All good to go. https://github.com/ocaml/opam/pull/6715 should be merged first as it fixes the CI failure on Windows

master_changes.md
```diff
@@ -142,6 +144,10 @@ users)
   * Add a test showing the behaviour of `opam var <pkg>:opamfile` [#6659 @kit-ty-kate]
   * Add a test to show homogeneity of outputs on verbose between sandboxed and non sandboxed commands (with `-vv`) [#6675 @rjbou]
   * Update `sed-cmd` reftest reftest [#6675 @rjbou]
+  * Add opam file loading tests to `update.test` to demonstrate current behaviour of loading full repository instead of only changed files. [#6614 @arozovyk]
+  * Add extra file loading tests to `update.test` [#6614 @arozovyk]
+  * Update `action-disk`, `repository` and `update` reftests to reflect incremental loading behavior - fewer opam file reads logged during repository updates [#6614 @arozovyk]
+  * Add specific extra file tests to  `update.test` in order to outline the edge cases of removing opams and extra files that have uncommon prefixes in their path [#6614 @arozovyk]
```
```suggestion
  * Add opam file loading tests to `update.test` to demonstrate current behaviour of loading full repository instead of only changed files. [#6614 @arozovyk @rjbou @kit-ty-kate]
```

tests/reftests/update.test
```diff
@@ -7,3 +8,597 @@ opam-version: "2.0"
...
+
+<><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
+REPO_BACKEND                    diff: ${BASEDIR}/OPAM/repo/{default,default.new}
+REPO_BACKEND                    Internal diff (non-empty, 2 changed files) done in 0.00s.
+[default] synchronised from file://${BASEDIR}/REPO
+FILE(config)                    Read ${BASEDIR}/OPAM/config in 0.000s
+FILE(opam)                      Read ${BASEDIR}/OPAM/repo/default/packages/qux/qux.1/opam in 0.000s
+opam-file                       Mismatching extra-files at ${BASEDIR}/OPAM/repo/default/packages/qux/qux.1: wrong checksum (1)
+Now run 'opam upgrade' to apply any package updates.
+### # ERROR ?
```
rjbou:
```suggestion
```

tests/reftests/update.test
```diff
@@ -7,3 +8,597 @@ opam-version: "2.0"
...
+opam-version: "2.0"
+### sh add-files.sh sha REPO install qux.1
+### rm REPO/packages/qux/qux.1/files/qux.install
+### OPAMDEBUGSECTIONS="opam-file" OPAMDEBUG=-3 opam update default | unordered
+
+<><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
+[default] synchronised from file://${BASEDIR}/REPO
+FILE(config)                    Read ${BASEDIR}/OPAM/config in 0.000s
+opam-file                       Missing expected extra files qux.install at ${BASEDIR}/OPAM/repo/default/packages/qux/qux.1/files
+### # ERROR ?
```
rjbou:
```suggestion
```

rjbou:
Updated & rebased

kit-ty-kate:
Thanks a lot!

Issue #6349, opened 2025-01-06
produce a reproducible index.tar.gz


hannesm:
Dear Madam or Sir,

in the light of https://github.com/ocaml/opam/issues/5553, and from a security perspective, wouldn't it be great if we could rely on "taking a specific git commit from the opam-repository, we'll be able to reproduce the index.tar.gz served to clients".

This would improve the security since at the moment, it is tough for a client to check that the index.tar.gz contains all the opam files as advertised.

Now, the index.tar.gz is generated as far as I can tell by `repository/opamHTTP.ml`:

```OCaml
let make_index_tar_gz repo_root =
  OpamFilename.in_dir repo_root (fun () ->
    let to_include = [ "version"; "packages"; "repo" ] in
    match List.filter Sys.file_exists to_include with
    | [] -> ()
    | d  -> OpamSystem.command ("tar" :: "czhf" :: "index.tar.gz" :: "--exclude=.git*" :: d)
  )
```

When doing releases to the opam-repository, we've faced this issue several times, and gladly b0 (previous topkg) provide a minimal tar implementation to do that: https://github.com/b0-system/b0/blob/master/src/std/b0_tar.ml

Using that approach would immediately lead to a reproducible tar archive. The timestamp of all files can easily be set to the last git commit.

Is that something that would be accepted as a PR (in case you don't have the resources to work on it)?

PR #6625, opened 2025-07-31
Improve performance of opam update/init by changing the structure of the internal http opam repositories (use the tar.gz as-is)

### LINKED ISSUE #3050: idea for accelerating 'opam update'
Hello,
I have a prototype there: https://github.com/UnixJunkie/opmer
Be careful, it is not yet super fast and it might very well contain bugs.

Here is a usage example with two opam-repository checkouts:

    opmer -v --hash $HOME/old/opam-repository
    opmer -v --hash $HOME/new/opam-repository
    opmer --diff $HOME/old/opam-repository:$HOME/new/opam-repository -o diff.log

### LINKED ISSUE #5346: `opamfile` variable refers to missing file
Prior to 2.1.3:

```
dra@thor:~$ opam var ocaml:opamfile
/tmp/opam-18607-5b5b85/default/packages/ocaml/ocaml.5.0.0/opam
```

which was clearly an error. I couldn't find which fix is responsible in 2.1.3, but in 2.1.3 I get:

```
dra@thor:~$ opam var ocaml:opamfile
/home/dra/.opam/repo/default/packages/ocaml/ocaml.5.0.0/opam
```

but this is only slightly better, as the repository is in tar'd form, so the file doesn't actually exist.

I think it would be better if the final state of the variable pointed instead to `/home/dra/.opam/5.0-beta/.opam-switch/packages/ocaml.5.0.0/opam` (I can't remember if that's technically valid _during_ the build?)

### LINKED ISSUE #5484: opam update, Could not update repository "default": Failed to extract archive
Hi, an 'opam update' gives an error 2. Any ideas on how to investigate this?
```
teg@hilbert:~$ opam update

[ERROR] Could not update repository "default": Failed to extract archive /tmp/opam-17061-2b9851/index.tar.gz:
        "/usr/bin/tar xfz /tmp/opam-17061-2b9851/index.tar.gz -C /tmp/opam-17061-edec5a/default.new" exited with
        code 2

teg@hilbert:~$ opam update --debug-level=10
00:00.036  FILE(config)           Read ~/.opam/config in 0.000s
00:00.036  CLI                    Parsing CLI version 2.1
00:00.036  GSTATE                 LOAD-GLOBAL-STATE @ /home/teg/.opam
00:00.037  SYSTEM                 LOCK /home/teg/.opam/lock (none => read)
00:00.037  SYSTEM                 LOCK /home/teg/.opam/config.lock (none => write)
00:00.037  FILE(config)           Read ~/.opam/config in 0.000s
00:00.037  CLIENT                 UPDATE 
00:00.037  RSTATE                 LOAD-REPOSITORY-STATE @ /home/teg/.opam
00:00.037  FILE(repos-config)     Read ~/.opam/repo/repos-config in 0.000s
00:00.038  SYSTEM                 LOCK /home/teg/.opam/repo/state-11ABC690.cache (none => read)
00:00.164  CACHE(repository)      Loaded /home/teg/.opam/repo/state-11ABC690.cache in 0.126s
00:00.164  SYSTEM                 LOCK /home/teg/.opam/repo/state-11ABC690.cache (read => none)
00:00.164  RSTATE                 Cache found
00:00.164  STATE                  LOAD-SWITCH-STATE @ 5.0.0
00:00.164  FILE(switch-config)    Read ~/.opam/5.0.0/.opam-switch/switch-config in 0.000s
00:00.164  FILE(switch-state)     Read ~/.opam/5.0.0/.opam-switch/switch-state in 0.000s
00:00.164  SYSTEM                 LOCK /home/teg/.opam/5.0.0/.opam-switch/packages/cache (none => read)
00:00.164  CACHE(installed)       Loaded /home/teg/.opam/5.0.0/.opam-switch/packages/cache in 0.000s
00:00.164  SYSTEM                 LOCK /home/teg/.opam/5.0.0/.opam-switch/packages/cache (read => none)
00:00.699  FILE(.config)          Read ~/.opam/5.0.0/.opam-switch/config/ocaml.config in 0.000s
00:00.699  STATE                  Switch state loaded in 0.535s
00:00.699  SYSTEM                 LOCK /home/teg/.opam/repo/lock (none => write)
00:00.699  SYSTEM                 LOCK /home/teg/.opam/repo/lock (none => write)

00:00.703  PARALLEL               Iterate over 1 task(s) with 3 process(es)
00:00.703  PARALLEL               Starting job 0 (worker 1/3): 0
00:00.703  SYSTEM                 mkdir /tmp/opam-17122-e3a56e
00:06.658  REPOSITORY             update default from https://opam.ocaml.org
00:06.658  CURL                   pull-repo-update
00:06.658  SYSTEM                 mkdir /tmp/opam-17122-e3a56e/default.new
00:06.658  SYSTEM                 mkdir /tmp/opam-17122-46fd0e
00:06.659  PARALLEL               Next task in job 0: /usr/bin/curl --write-out %{http_code}\n --retry 3 --retry-delay 2 --user-agent opam/2.1.2 -L -o /tmp/opam-17122-46fd0e/index.tar.gz.part -- https://opam.ocaml.org/index.tar.gz
Processing  1/1: [default: http]
00:14.034  PARALLEL               Collected task for job 0 (ret:0)
00:14.040  SYSTEM                 [log-17122-f45b9b] (in 0.005s) mv /tmp/opam-17122-46fd0e/index.tar.gz.part /tmp/opam-17122-46fd0e/index.tar.gz
00:14.040  PARALLEL               Next task in job 0: /usr/bin/tar xfz /tmp/opam-17122-46fd0e/index.tar.gz -C /tmp/opam-17122-e3a56e/default.new
00:19.604  PARALLEL               Collected task for job 0 (ret:2)
00:19.614  SYSTEM                 rmdir /tmp/opam-17122-46fd0e
00:19.619  SYSTEM                 rmdir /tmp/opam-17122-e3a56e/default.new
[ERROR] Could not update repository "default": Failed to extract archive /tmp/opam-17122-46fd0e/index.tar.gz:
        "/usr/bin/tar xfz /tmp/opam-17122-46fd0e/index.tar.gz -C /tmp/opam-17122-e3a56e/default.new" exited with
        code 2
00:21.366  PARALLEL               Job 0 finished
00:21.366  FILE(repos-config)     Wrote /home/teg/.opam/repo/repos-config in 0.000s
00:21.366  SYSTEM                 rm /home/teg/.opam/repo/state-11ABC690.cache
00:21.373  SYSTEM                 LOCK /home/teg/.opam/repo/state-11ABC690.cache (none => write)
00:21.373  CACHE(repository)      Writing the repository cache to ~/.opam/repo/state-11ABC690.cache ...
00:22.029  CACHE(repository)      ~/.opam/repo/state-11ABC690.cache written in 0.656s
00:22.029  SYSTEM                 LOCK /home/teg/.opam/repo/state-11ABC690.cache (write => none)
00:22.029  SYSTEM                 LOCK /home/teg/.opam/repo/lock (write => none)
00:22.108  SYSTEM                 LOCK /home/teg/.opam/5.0.0/.opam-switch/lock (none => none)
00:22.108  SYSTEM                 rmdir /tmp/opam-17122-e3a56e/default
00:25.562  SYSTEM                 rmdir /tmp/opam-17122-e3a56e
00:25.566  SYSTEM                 LOCK /home/teg/.opam/repo/lock (none => none)
00:25.566  SYSTEM                 LOCK /home/teg/.opam/config.lock (write => none)
```

The man pages for tar (GNU tar 1.34) gives this for return code 2
```
2      Fatal error.  This means that some fatal, unrecoverable error occurred.
```

The file system have a lot of room:
```
teg@hilbert:~$ df -H /tmp
Filesystem      Size  Used Avail Use% Mounted on
/dev/sda5       2.0G   97M  1.8G   6% /tmp
```

```
teg@hilbert:~$ opam config report
# opam config report
# opam-version         2.1.2 
# self-upgrade         no
# system               arch=x86_64 os=linux os-distribution=debian os-version=12
# jobs                 3
# repositories         1 (http), 1 (version-controlled) (default repo at ed47929e)
# pinned               0
# current-switch       5.0.0
# ocaml:native         true
# ocaml:native-tools   true
# ocaml:native-dynlink true
# ocaml:stubsdir       /home/teg/.opam/5.0.0/lib/ocaml/stublibs:/home/teg/.opam/5.0.0/lib/ocaml
# ocaml:preinstalled   false
# ocaml:compiler       5.0.0
```

### LINKED ISSUE #5559: `opam update` very slow on MacOS
Hi,

I'm on MacOS, and updating takes far longer than on an Ubuntu machine I have access too (~15min vs ~20s):

```
$ time opam update -R default --switch=default

<><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><>  🐫 
[default] no changes from https://opam.ocaml.org
opam update -R default --switch=default  28.91s user 303.24s system 33% cpu 16:28.17 total
```

It seems to me very long. I'd appreciate if anybody could point me towards updating faster.

The network should not be too much of a problem, network speedtest reports ~100Mbps in average download speed.

(This has been a problem for quite a long time, but I did not take the time to raise an issue)

Here is the result of `opam config report --switch=default` (modified to use $HOME).
```
# opam config report
# opam-version         2.1.5 
# self-upgrade         no
# system               arch=x86_64 os=macos os-distribution=homebrew os-version=13.4
# jobs                 7
# repositories         1 (http), 1 (version-controlled) (default repo at 93bb2e67)
# pinned               4 (git)
# current-switch       default
# ocaml:native         true
# ocaml:native-tools   true
# ocaml:native-dynlink true
# ocaml:stubsdir       $HOME/.opam/default/lib/ocaml/stublibs:$HOME/.opam/default/lib/ocaml
# ocaml:preinstalled   false
# ocaml:compiler       5.0.0
```

Thank you
Hadrien

### LINKED ISSUE #5648: Allow to use a "compressed" (one file) repository format for performance and sustainability purpose

### LINKED ISSUE #5741: opam.ocaml.org repository takes 10+ minutes to download on Windows

### LINKED ISSUE #6974: Add a test showing the 2.1/OPAMREPOTARRING=1 to 2.6 upgrade
Related to https://github.com/ocaml/opam/pull/6949

It would be nice to have that on top of adding a 2.5 to 2.6 upgrade before the release

kit-ty-kate:
Fixes #5741
Fixes https://github.com/ocaml/opam/issues/5648
Fixes https://github.com/ocaml/opam/issues/5484
~~Fixes https://github.com/ocaml/opam/issues/5346~~
Fixes https://github.com/ocaml/opam/issues/5559
Fixes https://github.com/ocaml/opam/issues/3050
Fixes #6974 
cc @hannesm to check if it works for conex (2.1 worked with tar.gz files already so i'm not too scared about breakages)

### Reasoning

#5741 shows that assumptions that hold true "most of the time" on some unix platforms such as "it is ok to scan a large tree structure of files and directories", don't hold true on other platforms. Systems such as Windows, network filesystems, busy shared servers whose disk is being constantly used, harddrives, … suffer from this.

In opam we can have 3 types of repositories:
1. HTTP (the default): where we download a tar.gz
2. VCS (aka. mostly git these days): where we use the vcs command line to get the files
3. local/ssh: where we either use our own copy primitives or use rsync

Out of the three, the most critical for first time users is the first one. It is also the one that suffer the most from these issues as currently we:
1. untar it
2. diff with the current repository (used for conex and for https://github.com/ocaml/opam/pull/6614)
3. patch the changed files
4. remove the directory
5. rescan the whole repository

VCS do not have step 4. Steps 1, 2 and 3 are builtin and heavily optimized. Is left only step 5 which should be improved by https://github.com/ocaml/opam/pull/6614 and for which we can improve further later by using `git cat-file` or even parse PACK files using `ocaml-git`.

Local/ssh repositories are the ones left a bit with very few things we can do about them. https://github.com/ocaml/opam/pull/5966 should help, but beyond, maybe we might want to require that people use git even for local repositories.

For HTTP though, the untarring (which takes 1+ minute) is the main issue. Thus this here PR.

### Design decisions

Instead of untarring we simply use the tar.gz as-is and use `ocaml-tar` to read it on the fly.
The new update steps are:
1. diff the two tar.gz
2. remove the old tar.gz
3. move the new tar.gz in its place
4. scan what has changed (required https://github.com/ocaml/opam/pull/6614)

Given the ubiquity of the use of `OpamFilename.Dir.t` to mean both any random directory and a repository directory, i chose to first abstract over it in a new `OpamRepositoryRoot` module, and work with the help of the type checker from there. Its interface help see what are the actions that opam does on repositories. While i'd rather keep them, the `Tar` and `Dir` submodules can be removed when everything is done.

The `REPOSITORYTARRING` environment variable is removed by this work, given the repositories are tarred already.

I had simplify `opam var pkg:opamfile` for this work. Previously it would point to the file in the repository. However this isn't what it's supposed to be doing. Instead it should point to the `<switch>/.opam-switch/packages/` directory which actually reflects the opam file that was used to installed. Otherwise the opam file can change between before and after the user has called `opam update` etc.

### TODO

There are a number of `assert false (* TODO *)` in this draft PR. Those are to be fixed before undrafting but i felt reasonably confident with the rest of them to open this draft PR in this state to put more eyes on this work and to increase my self-motivation.

- The main thing to do is to do the diff function between two tar files and between a directory and a tar.
- The other thing is to fill `OpamRepositoryState.get_repo_files`: a function which extracts a limited number of files from the tar.gz to a new cache directory.

Some of these changes should probably be extracted to separate PRs but let's do that at the end when we have something that actually works.

While early form of this work started a year and a half ago, i believe the crust left over from that time should be minimal, after 6 different branches. The final rebase and split into smaller PRs shouldn't be too painful.

### Future work

In the future we can use `ocaml-tar` that we now depend on to replace some of the uses of the `tar` command. This should allow us to have better behaviours with things like symlinks on windows or even add new features such as excluding some directories (see https://github.com/ocaml/ocaml/pull/14152).

As mentioned above we can also improve local and git/vcs repositories with or without `ocaml-git`.

---
Queued PRs
* [x] ~~#6871~~
* [x] #6913
* [x] #6914 
* [x] #6915 
* [x] #6917 
* [x] #6680
* [x] #6934
* [x] #6945
* [x] #6949
* [x] #6941
* [x] #6939
* [x] #6970
* [x] #6971
* [x] #6998 
* [x] #7002

A run of reftests with `OPAMREPOSITORYTARRING` enabled for all tests can be found on top of [rjbou/repotar-reftest-repotarring-enabled](https://github.com/rjbou/opam/tree/repotar-reftest-repotarring-enabled)

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -93,6 +93,7 @@ users)
   * Fix switch and repo format upgrade on Windows. A block occurred because the global lock fd was reopened instead of using the one already opened.  [#6839 @rjbou]
   * Add opam root format upgrade conditional mechanism for hard upgrades [#6949 @rjbou]
   * Stop the opam 1.2 to 2.0 repository upgrade process from downloading packages without checksums to add a non-trusted md5 [#6978 @kit-ty-kate]
+  * Add conditional upgrade to opam root 2.6. It is triggered by an already present repository archive (from `OPAMREPOSITORYTARRING` usage, or 2.1 opamroot) [#6625 @rjbou]
 
 ## Sandbox
   * Allow the macOS sandbox to write in the `/var/folders/` and `/var/db/mds/` directories as it is required by some of macOS core tools [#4797 @kit-ty-kate - fix #4389 #6460]
@@ -197,6 +198,7 @@ users)
   * Add in `repository-http` a test case for switching from directory to archive format, automatically [#6625 @rjbou]
   * Add in `repository` test cases for switching automatically from directory to archive format & vice versa [#6625 @rjbou]
   * Add in `repository` test cases for upgrade opam root from 2.5 with repo tarring or 2.1 to 2.6, with `OPAMREPOSITORYTARRING` enabled (trigger upgrade) [#6625 @rjbou]
+  * Add 2.6 root test cases in opamroot-versions [#6625 @rjbou]
 
 ### Engine
   * Add `http-server` to launch a minimal http server [#6939 @rjbou]
--- a/master_changes.md
+++ b/master_changes.md
@@ -332,4 +332,5 @@ users)
   * `OpamFilename.Unix.Dir`: add `basename` and `dirname` [#6625 @rjbou]
   * `OpamFilename.Unix`: add `remove_prefix` [#6625 @rjbou]
   * `OpamFilename.Unix`: add `to_relative_canonical` [#6625 @kit-ty-kate @rjbou]
+  * `OpamFilename.Unix`: add `root_dir` [#6625 @rjbou]
   * `OpamFilename.Unix.{to_dir,to_filename}`: now translate `/` into filesystem directory separator [#6625 @rjbou]
--- a/master_changes.md
+++ b/master_changes.md
@@ -195,6 +195,8 @@ users)
   * Add a test ensuring installing files through a .install file can't escape the opam switch (CVE-2026-57825) [#7005 @NathanReb]
   * Add a `opam repo set-url` case in repository-http [#6625 @rjbou]
   * Add in `repository-http` a test case for switching from directory to archive format, automatically [#6625 @rjbou]
+  * Add in `repository` test cases for switching automatically from directory to archive format & vice versa [#6625 @rjbou]
+  * Add in `repository` test cases for upgrade opam root from 2.5 with repo tarring or 2.1 to 2.6, with `OPAMREPOSITORYTARRING` enabled (trigger upgrade) [#6625 @rjbou]
 
 ### Engine
   * Add `http-server` to launch a minimal http server [#6939 @rjbou]
--- a/master_changes.md
+++ b/master_changes.md
@@ -258,6 +258,7 @@ users)
   * `OpamTar.create`: add `?except_vcs` argument exclude VCS files from archive creation [#6625 @rjbou]
   * `OpamTar`: when an archive is opened, the first step is to check and normalise canonical paths (no `/../`, remove `./`, etc.) [#6625 @rjbou]
   * `OpamRepositoryRoot`: add `remove_prefix` and `remove_prefix_dir` [#6625 @rjbou]
+  * `OpamRepositoryRoot`: add `read_file` that reads an `OpamFile.t` using its pp from the archive [#6625 @rjbou]
 
 ## opam-state
   * `OpamStateConfig.t`: replace `no_depexts` fields that contains disabling informations by `depexts` field that returns if the depexts mechanism is enabled. This field is automatically update by global config value in `OpamStateConfig.load_defaults` [#6489 @rjbou]
--- a/master_changes.md
+++ b/master_changes.md
@@ -253,6 +253,7 @@ users)
   * `OpamRepositoryRoot` was added [#6680 @kit-ty-kate @rjbou]
   * `OpamTar`: add module to manipulate tar gz archive. It handles only files, not directories [#6945 @kit-ty-kate @rjbou]
   * `OpamRepositoryCommand.update_with_auto_upgrade`, `OpamUpdate.repository`: no longer call an external process to create an archive [#6945 @kit-ty-kate]
+  * `OpamTar`: add `patch` function to patch files in an tar gz archive [#6625 @rjbou]
 
 ## opam-state
   * `OpamStateConfig.t`: replace `no_depexts` fields that contains disabling informations by `depexts` field that returns if the depexts mechanism is enabled. This field is automatically update by global config value in `OpamStateConfig.load_defaults` [#6489 @rjbou]
@@ -336,3 +337,5 @@ users)
   * `OpamFilename.Unix`: add `to_relative_canonical` [#6625 @kit-ty-kate @rjbou]
   * `OpamFilename.Unix`: add `root_dir` [#6625 @rjbou]
   * `OpamFilename.Unix.{to_dir,to_filename}`: now translate `/` into filesystem directory separator [#6625 @rjbou]
+  * `OpamPatch.patch`: no longer patch a file on disk, but take as argument a filesystem abstraction `FS_ABSTR` that delivers the needed functions [#6625 @rjbou]
+  * `OpamPatch.parse_patch`: no longer take `~dir` the directory to translate the patch in as argument, it now takes `~translate` argument that is a string option (directory option), if we want to perform a translation in that directory [#6625 @rjbou]
--- a/master_changes.md
+++ b/master_changes.md
@@ -254,6 +254,7 @@ users)
   * `OpamTar`: add module to manipulate tar gz archive. It handles only files, not directories [#6945 @kit-ty-kate @rjbou]
   * `OpamRepositoryCommand.update_with_auto_upgrade`, `OpamUpdate.repository`: no longer call an external process to create an archive [#6945 @kit-ty-kate]
   * `OpamTar`: add `patch` function to patch files in an tar gz archive [#6625 @rjbou]
+  * `OpamTar.create`: add `?flat` argument to do not integrate the target root directory in the archive [#6625 @rjbou]
 
 ## opam-state
   * `OpamStateConfig.t`: replace `no_depexts` fields that contains disabling informations by `depexts` field that returns if the depexts mechanism is enabled. This field is automatically update by global config value in `OpamStateConfig.load_defaults` [#6489 @rjbou]
@@ -277,7 +278,6 @@ users)
   * `OpamRepositoryState`: add `syspkgs_available` that returns the stored depext availability status in repository state [#6489 @rjbou]
   * `OpamSysInteract`: add `available_packages_and_family` that returns availability status and the os family [#6489 @rjbou]
   * `OpamRepositoryState.load_opams_from_dir`: now sorts files and directories read from disk before processing them [#6941 @rjbou]
-
 
 ## opam-solver
 
--- a/master_changes.md
+++ b/master_changes.md
@@ -194,6 +194,7 @@ users)
   * Add a tests for the several layouts of packages in a repository [#6941 @rjbou]
   * Add a test ensuring installing files through a .install file can't escape the opam switch (CVE-2026-57825) [#7005 @NathanReb]
   * Add a `opam repo set-url` case in repository-http [#6625 @rjbou]
+  * Add in `repository-http` a test case for switching from directory to archive format, automatically [#6625 @rjbou]
 
 ### Engine
   * Add `http-server` to launch a minimal http server [#6939 @rjbou]
--- a/master_changes.md
+++ b/master_changes.md
@@ -295,6 +295,7 @@ users)
   * `OpamRepositoryPath.{root,repo,packages_dir,packages,opam,files,descr,url}: have been moved to a new `OpamRepositoryPath.Make` functor [#6680 @rjbou @kit-ty-kate]
   * `OpamFilter.expand_interpolations_in_file`: changed argument type from `basename` to `filename` [#6910 @NathanReb]
   * `OpamFile.OPAM.print_errors`: now prints the repository if the informations is available [#6971 @rjbou]
+  * `OpamFile.*.read_from_string`: add optional `?loc` string argument to propagate location information when available (path on disk, archive, etc.), and add logging with level 3 that displays it [#6625 @rjbou]
 
 ## opam-core
   * `OpamCmdliner` was added. It is accessible through a new `opam-core.cmdliner` sub-library [#6755 @kit-ty-kate]
--- a/master_changes.md
+++ b/master_changes.md
@@ -71,6 +71,7 @@ users)
 ## Repository
   * No longer call tar tool to create archives, use tar library instead [#6945 @kit-ty-kate]
   * [BUG] Do not fail on directories named `opam` when scanning the `packages` directory of a repository during `opam repo add` or `opam init` (worked on subsequent `opam update`) [#6941 @kit-ty-kate @rjbou]
+  * Speedup repository operations on certain file-systems (e.g. NTFS on Windows or IO constrained machines) by changing its storage in the opam root from plain directory to archive, for HTTP repositories, or non-VCS one if `OPAMREPOSITORYTARRING` is enabled [#6625 @rjbou @kit-ty-kate @arozovyk - fix #5346 #5741 #5648 #5484 #5559 #3050 #6974]
 
 ## Lock
   * [BUG] Fix `undefined variable` error when a lock file filter contains an undefined variables: fail gracefully with strict mode, continue and default the variable to false on normal mode [#6947 @rjbou - fix #6946]
@@ -260,6 +261,8 @@ users)
   * `OpamTar`: when an archive is opened, the first step is to check and normalise canonical paths (no `/../`, remove `./`, etc.) [#6625 @rjbou]
   * `OpamRepositoryRoot`: add `remove_prefix` and `remove_prefix_dir` [#6625 @rjbou]
   * `OpamRepositoryRoot`: add `read_file` that reads an `OpamFile.t` using its pp from the archive [#6625 @rjbou]
+  * `OpamRepositoryBackend.get_diff`: now computes the diff between two repository roots (dir, archive), instead of only dirs [#6625 @rjbou]
+  * `OpamRepositoryRoot`: add `Tgz` module for tar gz archive repository root support [#6625 @rjbou @kit-ty-kate]
 
 ## opam-state
   * `OpamStateConfig.t`: replace `no_depexts` fields that contains disabling informations by `depexts` field that returns if the depexts mechanism is enabled. This field is automatically update by global config value in `OpamStateConfig.load_defaults` [#6489 @rjbou]
@@ -283,6 +286,10 @@ users)
   * `OpamRepositoryState`: add `syspkgs_available` that returns the stored depext availability status in repository state [#6489 @rjbou]
   * `OpamSysInteract`: add `available_packages_and_family` that returns availability status and the os family [#6489 @rjbou]
   * `OpamRepositoryState.load_opams_from_dir`: now sorts files and directories read from disk before processing them [#6941 @rjbou]
+  * `OpamFileTools`: add new `lint_repo_package` that lints a file from a repository root [#6625 @rjbou]
+  * `OpamRepositoryState`: add `load_opams_from_tgz` [#6625 @rjbou]
+  * `OpamRepositoryState`: add `load_opams` that operates from a repository root [#6625 @rjbou]
+  * `OpamStateTypes.repos_state`: remove `repos_tmp` field [#6625 @kit-ty-kate @rjbou]
 
 ## opam-solver
 
@@ -303,6 +310,7 @@ users)
   * `OpamFile.OPAM.print_errors`: now prints the repository if the informations is available [#6971 @rjbou]
   * `OpamFile.*.read_from_string`: add optional `?loc` string argument to propagate location information when available (path on disk, archive, etc.), and add logging with level 3 that displays it [#6625 @rjbou]
   * `OpamFile.*`: add `safe_read_from_string` [#6625 @rjbou]
+  * `OpamRepositoryPath.tar`: renamed to `repo_tarring`, deprecated, no longer used [#6625 @rjbou]
 
 ## opam-core
   * `OpamCmdliner` was added. It is accessible through a new `opam-core.cmdliner` sub-library [#6755 @kit-ty-kate]
--- a/master_changes.md
+++ b/master_changes.md
@@ -257,6 +257,7 @@ users)
   * `OpamTar.create`: add `?flat` argument to do not integrate the target root directory in the archive [#6625 @rjbou]
   * `OpamTar.create`: add `?except_vcs` argument exclude VCS files from archive creation [#6625 @rjbou]
   * `OpamTar`: when an archive is opened, the first step is to check and normalise canonical paths (no `/../`, remove `./`, etc.) [#6625 @rjbou]
+  * `OpamRepositoryRoot`: add `remove_prefix` and `remove_prefix_dir` [#6625 @rjbou]
 
 ## opam-state
   * `OpamStateConfig.t`: replace `no_depexts` fields that contains disabling informations by `depexts` field that returns if the depexts mechanism is enabled. This field is automatically update by global config value in `OpamStateConfig.load_defaults` [#6489 @rjbou]
--- a/master_changes.md
+++ b/master_changes.md
@@ -256,6 +256,7 @@ users)
   * `OpamTar`: add `patch` function to patch files in an tar gz archive [#6625 @rjbou]
   * `OpamTar.create`: add `?flat` argument to do not integrate the target root directory in the archive [#6625 @rjbou]
   * `OpamTar.create`: add `?except_vcs` argument exclude VCS files from archive creation [#6625 @rjbou]
+  * `OpamTar`: when an archive is opened, the first step is to check and normalise canonical paths (no `/../`, remove `./`, etc.) [#6625 @rjbou]
 
 ## opam-state
   * `OpamStateConfig.t`: replace `no_depexts` fields that contains disabling informations by `depexts` field that returns if the depexts mechanism is enabled. This field is automatically update by global config value in `OpamStateConfig.load_defaults` [#6489 @rjbou]
--- a/master_changes.md
+++ b/master_changes.md
@@ -332,3 +332,4 @@ users)
   * `OpamFilename.Unix.Dir`: add `basename` and `dirname` [#6625 @rjbou]
   * `OpamFilename.Unix`: add `remove_prefix` [#6625 @rjbou]
   * `OpamFilename.Unix`: add `to_relative_canonical` [#6625 @kit-ty-kate @rjbou]
+  * `OpamFilename.Unix.{to_dir,to_filename}`: now translate `/` into filesystem directory separator [#6625 @rjbou]
--- a/master_changes.md
+++ b/master_changes.md
@@ -196,6 +196,7 @@ users)
 
 ### Engine
   * Add `http-server` to launch a minimal http server [#6939 @rjbou]
+  * Add to automatic path subsitutions lines that contains `packages` [#6625 @rjbou]
 
 ## Github Actions
   * Add OCaml 5.4 to the test matrix [#6732 @kit-ty-kate]
--- a/master_changes.md
+++ b/master_changes.md
@@ -255,6 +255,7 @@ users)
   * `OpamRepositoryCommand.update_with_auto_upgrade`, `OpamUpdate.repository`: no longer call an external process to create an archive [#6945 @kit-ty-kate]
   * `OpamTar`: add `patch` function to patch files in an tar gz archive [#6625 @rjbou]
   * `OpamTar.create`: add `?flat` argument to do not integrate the target root directory in the archive [#6625 @rjbou]
+  * `OpamTar.create`: add `?except_vcs` argument exclude VCS files from archive creation [#6625 @rjbou]
 
 ## opam-state
   * `OpamStateConfig.t`: replace `no_depexts` fields that contains disabling informations by `depexts` field that returns if the depexts mechanism is enabled. This field is automatically update by global config value in `OpamStateConfig.load_defaults` [#6489 @rjbou]
--- a/master_changes.md
+++ b/master_changes.md
@@ -296,6 +296,7 @@ users)
   * `OpamFilter.expand_interpolations_in_file`: changed argument type from `basename` to `filename` [#6910 @NathanReb]
   * `OpamFile.OPAM.print_errors`: now prints the repository if the informations is available [#6971 @rjbou]
   * `OpamFile.*.read_from_string`: add optional `?loc` string argument to propagate location information when available (path on disk, archive, etc.), and add logging with level 3 that displays it [#6625 @rjbou]
+  * `OpamFile.*`: add `safe_read_from_string` [#6625 @rjbou]
 
 ## opam-core
   * `OpamCmdliner` was added. It is accessible through a new `opam-core.cmdliner` sub-library [#6755 @kit-ty-kate]
--- a/master_changes.md
+++ b/master_changes.md
@@ -331,3 +331,4 @@ users)
   * `OpamFilename.Unix`: add `basename` and `dirname` [#6625 @rjbou]
   * `OpamFilename.Unix.Dir`: add `basename` and `dirname` [#6625 @rjbou]
   * `OpamFilename.Unix`: add `remove_prefix` [#6625 @rjbou]
+  * `OpamFilename.Unix`: add `to_relative_canonical` [#6625 @kit-ty-kate @rjbou]
--- a/master_changes.md
+++ b/master_changes.md
@@ -329,3 +329,4 @@ users)
   * `OpamSystem`: add `open_in` and `open_in_bin` that checks is the file is a directory and raise `Sys_error` in that case [#6941 @rjbou]
   * `OpamFilename.{open_in,open_in_bin}`: now errors when the filename is a directory, in all platforms [#6941 @rjbou]
   * `OpamFilename.Unix`: add `basename` and `dirname` [#6625 @rjbou]
+  * `OpamFilename.Unix.Dir`: add `basename` and `dirname` [#6625 @rjbou]
--- a/master_changes.md
+++ b/master_changes.md
@@ -330,3 +330,4 @@ users)
   * `OpamFilename.{open_in,open_in_bin}`: now errors when the filename is a directory, in all platforms [#6941 @rjbou]
   * `OpamFilename.Unix`: add `basename` and `dirname` [#6625 @rjbou]
   * `OpamFilename.Unix.Dir`: add `basename` and `dirname` [#6625 @rjbou]
+  * `OpamFilename.Unix`: add `remove_prefix` [#6625 @rjbou]
--- a/master_changes.md
+++ b/master_changes.md
@@ -326,3 +326,6 @@ users)
   * `OpamFilename.{,dir_}starts_with`: Fix a bug where `foo/bar` would be considered a prefix of `foo/bar-baz` [#6953 @NathanReb - fix #6948]
   * `OpamFilename.{,dir_}starts_with`: `/` and `\` are now equivalent on Windows [#6953 @NathanReb]
   * `OpamFilename.starts_with`: `starts_with "a/b" "a/b"` no longer returns `true` [#6953 @NathanReb]
+  * `OpamSystem`: add `open_in` and `open_in_bin` that checks is the file is a directory and raise `Sys_error` in that case [#6941 @rjbou]
+  * `OpamFilename.{open_in,open_in_bin}`: now errors when the filename is a directory, in all platforms [#6941 @rjbou]
+  * `OpamFilename.Unix`: add `basename` and `dirname` [#6625 @rjbou]
--- a/master_changes.md
+++ b/master_changes.md
@@ -321,7 +321,7 @@ users)
   * `OpamSystem.{command,commands,read_command_output}`: add a `?dir: dirname` optional arg to launch the command in a specific directory [#6910 @NathanReb]
   * `OpamFilename.Unix`: add `starts_with` [#6945 @rjbou]
   * `OpamCompat.Seq`: add `to_dispenser` [#6945 @kit-ty-kate]
-  * `OpamSystem.directories_with_links`, `OpamSystem.rec_files`, `OpamFilename.rec_files`: add optional `?except_vcs` that default to false to exclude VCS directories [#6945 @kit-ty-kate @rjbou]
+  * `OpamSystem.directories_with_links`, `OpamSystem.rec_files`, `OpamFilename.rec_files`: add optional `?except_vcs` that default to false to exclude VCS directories [#6625 @kit-ty-kate @rjbou]
   * `OpamFilename.make_tar_gz{_job}`: rename `make_tar_gz_job` into `make_tar_gz` as it no longer need an external process call [#6945 @kit-ty-kate]
   * `OpamFilename.{,dir_}starts_with`: Fix a bug where `foo/bar` would be considered a prefix of `foo/bar-baz` [#6953 @NathanReb - fix #6948]
   * `OpamFilename.{,dir_}starts_with`: `/` and `\` are now equivalent on Windows [#6953 @NathanReb]
--- a/master_changes.md
+++ b/master_changes.md
@@ -192,6 +192,7 @@ users)
   * Add a test showing the behaviour of `opam repo add` and `opam update` when faced with a repository containing an `opam` directory [#6995 @kit-ty-kate]
   * Add a tests for the several layouts of packages in a repository [#6941 @rjbou]
   * Add a test ensuring installing files through a .install file can't escape the opam switch (CVE-2026-57825) [#7005 @NathanReb]
+  * Add a `opam repo set-url` case in repository-http [#6625 @rjbou]
 
 ### Engine
   * Add `http-server` to launch a minimal http server [#6939 @rjbou]
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -224,10 +224,10 @@ let environment_variables =
       "enables option --no-checksums when available.";
       "REPOSITORYTARRING", cli_from cli2_2,
       (fun b -> REPOSITORYTARRING (env_bool b)),
-      "internally store the repositories as tar.gz files. This can be much \
-       faster on filesystems that don't cope well with scanning large trees \
-       but have good caching in /tmp. However this is slower in the \
-       general case.";
+      "internally store all non-vcs repositories as tar.gz files (by default, \
+       http repositories are stored as archive). This can be much faster on \
+       filesystems that don't cope well with scanning large trees. All \
+       operation are done in-memory, no I/O for scanning the repository.";
       "REQUIRECHECKSUMS", cli_original,
       (fun v -> REQUIRECHECKSUMS (env_bool v)),
       "Enables option `--require-checksums' when available \
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -2524,32 +2524,31 @@ let repository cli =
       in
       OpamGlobalState.with_ `Lock_none @@ fun gt ->
       OpamRepositoryState.with_ `Lock_write gt @@ fun rt ->
-      OpamFilename.with_tmp_dir @@ fun tmp_dir ->
+      OpamFilename.with_tmp_dir @@ fun inn ->
+      let repo_root =
+        OpamRepositoryState.get_repo_root rt
+          (OpamRepositoryState.get_repo rt name)
+      in
+      if not (OpamRepositoryRoot.exists repo_root) then
+        OpamConsole.error_and_exit `Internal_error
+          "Repository not found, consider running 'opam update %s' \
+           to retrieve a consistent state."
+          (OpamRepositoryName.to_string name);
       let rt0 = rt in
-      let backup =
-        let tar = OpamRepositoryPath.tar gt.root name in
-        if OpamFilename.exists tar then
-          (let target = OpamFilename.create tmp_dir (OpamFilename.basename tar) in
-           OpamFilename.copy ~src:tar ~dst:target;
-           fun () -> OpamFilename.copy ~src:target ~dst:tar)
-        else
-          (let dir = OpamRepositoryRoot.Dir.Path.root gt.root name in
-           if not (OpamRepositoryRoot.Dir.exists dir) then
-             OpamConsole.error_and_exit `Internal_error
-               "Repository not found, consider running 'opam update %s' \
-                to retrieve a consistent state."
-               (OpamRepositoryName.to_string name);
-           let target = OpamRepositoryRoot.Dir.backup ~inn:tmp_dir dir in
-           OpamRepositoryRoot.Dir.copy ~src:dir ~dst:target;
-           fun () -> OpamRepositoryRoot.Dir.copy ~src:target ~dst:dir)
+      let backup = OpamRepositoryRoot.backup ~inn repo_root in
+      OpamRepositoryRoot.copy ~src:repo_root ~dst:backup;
+      let restore_backup () =
+        OpamRepositoryRoot.copy ~src:backup ~dst:repo_root
       in
       let rt = OpamRepositoryCommand.set_url rt name url trust_anchors in
       let failed, rt =
         OpamRepositoryCommand.update_with_auto_upgrade rt [name]
       in
       OpamRepositoryState.drop rt;
-      if failed <> [] then
-        (let repo = OpamRepositoryState.get_repo rt0 name in
+      (match failed with
+       | [] -> `Ok ()
+       | _ ->
+         let repo = OpamRepositoryState.get_repo rt0 name in
          OpamConsole.error
            "Fetching repository %s with %s fails, reverting to %s"
            (OpamRepositoryName.to_string name)
@@ -2558,10 +2557,9 @@ let repository cli =
          let rt =
            OpamRepositoryCommand.set_url rt0 name repo.repo_url repo.repo_trust
          in
-         backup ();
+         restore_backup ();
          OpamRepositoryState.drop rt;
-         OpamStd.Sys.exit_because `Sync_error);
-      `Ok ()
+         OpamStd.Sys.exit_because `Sync_error)
     | Some `set_repos, names ->
       let names = List.map OpamRepositoryName.of_string names in
       OpamGlobalState.with_ `Lock_none @@ fun gt ->
@@ -2740,7 +2738,7 @@ let switch cli =
         lists installed switches, with one switch argument, defaults to \
         $(b,set).";
     `P (Printf.sprintf
-         "Switch handles $(i,SWITCH) can be either a plain name, for switches \
+       "Switch handles $(i,SWITCH) can be either a plain name, for switches \
          that will be held inside $(i,~%s.opam), or a directory name, which in \
          that case is the directory where the switch prefix will be installed, as \
          %s. Opam will automatically select a switch by that name found in the \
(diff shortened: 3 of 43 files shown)


hannesm:
Hey, thanks for your work on that. Since you asked me directly

> cc @hannesm to check if it works for conex

> Instead of untarring we simply use the tar.gz as-is and use ocaml-tar to read it on the fly.
> The new update steps are:
> 
>    diff the two tar.gz
>    remove the old tar.gz
>    move the new tar.gz in its place

This should be fine from the design point of view for conex.

Conex will need to interject between step 0 (you downloaded the tarball) and 2 (remove the old tar.gz). Currently, conex requires a diff file on disk, and the old repository as directory. But we can revise that interface, and conex could as well work on two tar.gz (and/or on two directories).

I guess you have a clear understanding of the update process currently, and since you mention the different kinds (http, git, local) -- maybe we should re-think how opam and conex should interact to avoid burden paid by people not using conex, and avoid the burden of duplicating computations in both opam and conex. The latter may need to include conex as a library into opam.

I'm away for the next 10 days (back on August 10th), but am happy to discuss this afterwards - esp.since I plan to revive my work on conex thereafter.

hannesm:
To be more precise, given that
- (a) local repositories (rsync) aren't really worth to verify (under the assumption that whoever has access to the repository can as well install arbitrary packages) [which may be revisited if there's the NFS use case or a shared server]
- (b) VCS (git): we could do the `git fetch` and provide conex with the local repository and the commit that the update should be to
- (c) http: provide old and new tarball

Then conex could do what is needed (compute the set of changed opam files, verify signatures; exit 0 on success); and could even report back the set of changed files to opam (I suspect this is what #6614 depends on) - using a file, or a socket, or if integrated with opam, this will be much simpler (using shared memory).

For opam itself, I guess that #6349 and #5553 will improve a lot of updates already.

kit-ty-kate:
Some update: I've been splitting this PR into several smaller ones to make working on it collaboratively, rebasing it and cleaning it up, easier.

So far:
* #6658, https://github.com/ocaml/opam/pull/6660, https://github.com/ocaml/opam/pull/6661, https://github.com/ocaml/opam/pull/6662 were merged
* https://github.com/ocaml/opam/pull/6659, https://github.com/ocaml/opam/pull/6679, https://github.com/ocaml/opam/pull/6680 are being worked on and will be merged in order

Once these done, i expect this PR to be fairly small, self-contained and easier to review.

src/client/opamCommands.ml
```diff
@@ -2506,22 +2506,20 @@ let repository cli =
...
                 to retrieve a consistent state."
                (OpamRepositoryName.to_string name);
-           let target =
-             OpamFilename.(Op.(tmp_dir / Base.to_string (basename_dir dir)))
-           in
-           OpamFilename.copy_dir ~src:dir ~dst:target;
-           fun () -> OpamFilename.copy_dir ~src:target ~dst:dir)
+           let target = OpamRepositoryRoot.Dir.backup ~tmp_dir dir in
+           OpamRepositoryRoot.Dir.copy ~src:dir ~dst:target;
+           fun () -> OpamRepositoryRoot.Dir.copy ~src:target ~dst:dir)
```
rjbou: Why not use directly `OpamRepositoryRoot.exists` and `OpamRepositoryRoot.copy` ?

src/state/opamUpdate.ml
```diff
@@ -18,11 +18,23 @@ open OpamFilename.Op
...
+    OpamFile.Repo.safe_read (OpamRepositoryPath.repo dir)
+  | OpamRepositoryRoot.Tar tar ->
+    let exception Found of string in
+    try
+      OpamTar.fold_reg_files (fun () filename content ->
+          if filename = "/repo" then
+            raise (Found content);
+        ) () (Unix.openfile (OpamRepositoryRoot.Tar.to_string tar) [Unix.O_RDONLY] 0);
+      OpamFile.Repo.empty
+    with Found content -> OpamFile.Repo.read_from_string content
```
rjbou: This logic is used twice in the code, worth haveing a function that does it

src/state/opamUpdate.ml
```diff
@@ -123,41 +135,35 @@ let repository rt repo =
...
-                OpamRepositoryName.Map.add repo.repo_name opams rt.repo_opams;
-            }
-        ))
+    let local_dir = OpamRepositoryPath.root gt.root repo.repo_name in
+    let opams =
+      match repo_root with
+      | OpamRepositoryRoot.Dir dir ->
+        OpamRepositoryState.load_opams_from_dir repo.repo_name dir
+      | OpamRepositoryRoot.Tar tar ->
+        OpamRepositoryState.load_opams_from_tar_gz repo.repo_name tar
```
rjbou: better have a `OpamRepositoryState.load_opams` that does that match

src/state/opamUpdate.ml
```diff
@@ -123,41 +135,35 @@ let repository rt repo =
...
+    in
+    begin match repo_root with
+    | OpamRepositoryRoot.Dir _ ->
+      if OpamRepositoryRoot.Tar.exists tarred_repo then
+        OpamRepositoryRoot.Tar.remove tarred_repo;
+    | OpamRepositoryRoot.Tar _ ->
+      if OpamRepositoryRoot.Dir.exists local_dir then
+        OpamRepositoryRoot.Dir.remove local_dir;
```
rjbou: Same here : why not use the abstracted logic ?

src/state/opamUpdate.ml
```diff
@@ -123,41 +135,35 @@ let repository rt repo =
...
-              repositories =
-                OpamRepositoryName.Map.add repo.repo_name repo rt.repositories;
-              repos_definitions =
-                OpamRepositoryName.Map.add repo.repo_name repo_file
-                  rt.repos_definitions;
-              repo_opams =
-                OpamRepositoryName.Map.add repo.repo_name opams rt.repo_opams;
-            }
-        ))
+    let local_dir = OpamRepositoryPath.root gt.root repo.repo_name in
```
rjbou: why do you retrieve both ?

src/repository/opamTar.ml
```diff
@@ -0,0 +1,119 @@
...
+        | [] -> Tar.return (Ok None)
+        | x::xs -> r := xs; Tar.return (Ok (Some x))
+    in
+    let t = Tar.out ~level:Ustar entries in
+    let t = Tar_gz.out_gzipped ~level:4 ~mtime:0l Gz.Unix t in
+    let buf = Buffer.create 10_485_760 in
+    to_buffer buf t;
+    let str = Buffer.contents buf in
+    let _ : int = Unix.write_substring fd str 0 (String.length str) in
+    ()
```
kit-ty-kate: @hannesm @dinosaure @reynir is there a way to write this in a cleaner way? I find the `ocaml-tar` API to be quite confusing and most of the examples i could find use `Lwt_stream`.

src/repository/opamTar.ml
```diff
@@ -0,0 +1,119 @@
...
+            in
+            (Some Tar.Header.Ustar, hdr, fun () -> Tar.return (Ok (Some content))) :: acc)
+          t []
+      in
+      let r = ref x in
+      fun () ->
+        match !r with
+        | [] -> Tar.return (Ok None)
+        | x::xs -> r := xs; Tar.return (Ok (Some x))
+    in
```
reynir, line 185: I think this can be `Map.to_seq t |> Seq.map ... |> Seq.to_dispenser` which may be slightly cleaner

kit-ty-kate: Thanks! Yeah it is a bit clearer although it is mostly moving the part dealing with the reference to a separate function (that we have to reimplement for compatibility with OCaml 4.08).

It's obviously fine for us at the moment but i would suggest to improve the ocaml-tar API in the future. Maybe allowing a list or having some sort of buffer would be much more efficient for this kind of use-case i feel like

src/repository/opamRepositoryBackend.ml
```diff
@@ -126,7 +126,20 @@ let strip_repo_suffix patch =
   in
   {patch with operation}
 
-let get_diff parent_dir dir1 dir2 =
```
rjbou: `get_diff`is historically in `OpamRepositoryBackend`, but given the functions added, its use in another PR for pinning stuff, i think it should be moved. Maybe a `core/opamDiff.ml`?

kit-ty-kate: I don't think we should do this. This would add four new (unused in the package itself) dependencies to `opam-core` for no tangible reason. This seems unwarranted for `opam-format` users.

rjbou: [discussed in dev meeting] Not adding 4 dependencies in `opam-core`/`opam-format` is a good argument. We can instead have a specific modul in `opam-repository`. The other call of these functions are against source code repository, so it stays relevant.

src/repository/opamTar.ml
```diff
@@ -0,0 +1,121 @@
...
+      run buf t
+    in
+    let entries =
+      let dispenser =
+        Map.fold (fun path content acc ->
+            let hdr =
+              Tar.Header.make ~file_mode:0 ~mod_time:0L ~user_id:0 ~group_id:0
+                path (Int64.of_int (String.length content))
+            in
+            let data = fun () -> Tar.return (Ok (Some content)) in
```
dinosaure, line 103: Here, the given stream must emit `None` at some points. Something like:
```ocaml
let data =
  let closed = ref false in
  fun () -> match !closed with
    | false -> closed := true; Tar.return (Ok (Some content))
    | true -> Tar.return (Ok None) in
...
```

Should fix you infinite loop.

src/repository/opamTar.ml
dinosaure: To be precise (and this is exactly as you requested), here we could use `Unix.write_string` directly rather than buffering and writing the archive in one go (which would remove the atomicity but reduce memory usage; another approach is to write to `/tmp` and use `Unix.rename`).

The basic idea is to decouple the I/O from the generation of a `tar.gz` file so that the user can decide how to actually write it.

A final suggestion would also be to generate a `string Seq.t` at the end if you prefer.

src/repository/opamRepositoryBackend.ml
rjbou: In end up a single function in last versions of the code.

src/repository/opamTar.ml
```diff
@@ -0,0 +1,235 @@
...
+let create ?(flat=false) tar dir =
+  log "creating archive %s from %s"
+    (OpamFilename.to_string tar)
+    (OpamFilename.Dir.to_string dir);
+  let fd =
+    Unix.openfile (OpamFilename.to_string tar)
+      [Unix.O_CREAT; Unix.O_TRUNC; Unix.O_WRONLY] 0o640
+  in
+  Fun.protect ~finally:(fun () -> Unix.close fd) @@ fun () ->
+  let files = OpamFilename.rec_files dir in
```
rjbou:
```suggestion
  let files = OpamFilename.rec_files ~except_vcs:true dir in
```

src/core/opamPatch.ml
```diff
@@ -322,87 +322,165 @@ let translate_patch ~dir orig corrected =
...
+              C.write (C.ext file ".orig") content patching)
+            patching content
+        in
+        patching, x
       | None -> assert false (* See NOTE above *)
       | exception _ ->
-        Option.iter (OpamSystem.write (file^".orig")) content;
-        OpamSystem.write (file^".rej") (Format.asprintf "%a" Patch.pp diff);
+        (* TAR TODO : write somewhere else ?
+           Option.iter (write (file^".orig")) content;
```
rjbou: add an function in configurator fro writing rejected files : write on disk, no-op on archive

src/core/opamPatch.ml
```diff
@@ -322,87 +322,165 @@ let translate_patch ~dir orig corrected =
   end;
   close_in ch
 
+(* Patch configurator *)
+
+module type PATCH_CONF = sig
```
rjbou: rename on `FS_ABSTR`

src/repository/opamRepositoryRoot.ml
```diff
@@ -56,62 +56,182 @@ module Dir = struct
...
+  let remove = OpamFilename.remove
+  let extract_in = OpamFilename.extract_in
+  let download_as = OpamDownload.download_as
+  let copy = OpamFilename.copy
+  let move = OpamFilename.move
+  let is_symlink = OpamFilename.is_symlink
+
+  let archives : (OpamHash.t, string OpamFilename.Unix.Map.t) Hashtbl.t = Hashtbl.create 8
+  let unload_repo_tars () = Hashtbl.clear archives
+
```
rjbou: extract this memoization to a draft PR, we'll see later if we need it or no

src/client/opamRepositoryCommand.ml
```diff
@@ -251,18 +251,15 @@ let update_with_auto_upgrade rt repo_names =
                 OpamRepositoryState.Cache.remove ());
              OpamConsole.msg "Upgrading repository \"%s\"...\n"
                (OpamRepositoryName.to_string r.repo_name);
-             let repo_root =
-               match OpamRepositoryState.get_repo_root rt r with
-               | OpamRepositoryRoot.Dir x -> x
+             let repo_root = OpamRepositoryState.get_repo_root rt r in
+             OpamRepositoryRoot.on_dir (fun dir ->
+                 OpamAdminRepoUpgrade.do_upgrade (OpamRepositoryRoot.Dir.of_dir dir))
```
rjbou, line 254: Open an issue for updating do_upgrade to perform upgrade on tar directly

src/state/opamPinned.ml
```diff
@@ -310,3 +310,32 @@ let orig_opam_file st name opam =
...
+    | OpamRepositoryRoot.Tar tar ->
+      let dir = OpamFilename.Unix.Dir.of_string rel in
+      let opam_files, locked_files = files (OpamFilename.Unix.Dir.to_dir dir) in
+      (* TAR TODO : use a generic files functions *)
+      let opam_files = List.map OpamFilename.Unix.of_filename opam_files in
+      let locked_files = List.map OpamFilename.Unix.of_filename locked_files in
+      (* TAR TOQUESTION : some comments to check
+         it is better to fold over the full repo only once and the look again
+         on smaller lists instead of looking twice over all repo where there is
+         no locked file *)
```
rjbou: remove comment, and update the function (maybe in the rewrite commit ?) to no longer check for `*.opam` & `*.opam.locked` when it is looking into a repository

src/repository/opamRepositoryRoot.ml
```diff
@@ -56,62 +56,182 @@ module Dir = struct
...
+  let patch ~allow_unclean patch_source tar =
+    let patch_source =
+      match patch_source with
+      | `Patch_file f -> `Patch_file (OpamFilename.to_string f)
+      | `Patch_diffs _ as d -> d
+    in
+    OpamTar.patch ~allow_unclean patch_source tar
+
+  let extract_files cond t =
```
rjbou: s/extract_files/filter_files

src/repository/opamRepositoryRoot.ml
```diff
@@ -56,62 +56,182 @@ module Dir = struct
...
+       really happen irl ? *)
+    let hash = OpamHash.compute ~kind:`SHA256 (OpamFilename.to_string tar) in
+    match Hashtbl.find_opt archives hash with
+    | Some contents ->
+      OpamFilename.Unix.Map.fold (fun filename content acc ->
+          f acc filename content)
+        contents x
+    | None ->
+      let result, map =
+        OpamTar.fold_reg_files (fun (acc, map) file content ->
```
rjbou: When we retrieve an archive from outside, we need to perform a pass to if some paths are canonical:
* `^./` : delete the ./
* `..` at any place : fail
* `/` : fail

Add the check in OpamTar.fold_reg_files
see #6924 canonical function

src/state/opamFileTools.ml
```diff
@@ -1612,6 +1612,13 @@ let read_repo_opam_t ~repo_name ~repo_root dir file content xfs =
   >>| OpamFile.OPAM.with_metadata_dir
     (Some (Some repo_name, rel))
 
+let read_repo_opam_tar ~repo_name ~repo_root dir file content xfs =
+  let file = OpamFilename.Unix.to_filename file in
+  let dir = OpamFilename.Unix.Dir.to_dir dir in
+  let repo_root = OpamRepositoryRoot.Tar repo_root in
+  let xfs = OpamFilename.Unix.Map.map (fun c -> lazy c) xfs in
```
rjbou: Use Lazy.from_val

src/state/opamRepositoryState.ml
```diff
@@ -102,6 +103,30 @@ let get_repo_files rt name dir =
...
+          let content = lazy (
+            (* TAR TODO : veru hackish remove ? *)
+            OpamConsole.log "SYSTEM" ~level:5
+              "read %s from %s"
+              (OpamFilename.Unix.to_string
+                 (OpamFilename.Unix.Dir.of_string
+                    (OpamRepositoryName.to_string name)
+                  // OpamFilename.Unix.to_string filename))
+              (OpamRepositoryRoot.Tar.to_string tar);
+            content)
```
rjbou: remove in final, kept ftm for debug

src/state/opamRepositoryState.ml
```diff
@@ -167,6 +167,56 @@ let read_package_opam_tar ~repo_name ~repo_root package_dir filename content ext
       (OpamFilename.Unix.to_string OpamFilename.Unix.Op.(package_dir // "opam"));
     None
 
+let load_raw_opams_and_aux_from_tar _repo_name tar =
```
rjbou: add comment for the different steps

src/state/opamRepositoryState.ml
```diff
@@ -167,6 +167,56 @@ let read_package_opam_tar ~repo_name ~repo_root package_dir filename content ext
...
+        OpamFilename.Unix.Dir.t
+        * (OpamFilename.Unix.t * string * string OpamFilename.Unix.Map.t)
+    in
+    List.fold_left (fun acc (filename, content) ->
+        try
+          OpamFilename.Unix.Dir.Map.iter (fun dir value ->
+              if OpamFilename.Unix.starts_with dir filename then
+                raise (Found (dir, value))
+            ) acc;
+          (* TAR TOQUESTION add a skipping message ? *)
```
rjbou: remove comment & add newline at the end of the function

src/client/opamListCommand.ml
```diff
@@ -875,6 +875,16 @@ let info st ~fields ~raw ~where ?normalise ?(show_empty=false)
...
+           in
+           if OpamRepositoryRoot.Tar.exists tar &&
+              not (OpamRepositoryRoot.Dir.exists repo_dir) then
+             Printf.sprintf "<%s>%s%s"
+               (OpamRepositoryRoot.Tar.to_string tar)
+               Filename.dir_sep
+               rdir
+           else
            OpamFilename.Dir.to_string
              OpamRepositoryRoot.Dir.Op.(repo_dir / rdir)
```
rjbou: + indent

src/state/opamFormatUpgrade.ml
```diff
@@ -1152,6 +1152,53 @@ let v2_2 = OpamVersion.of_string "2.2"
...
+                in
+                let is_repo =
+                  String.equal (OpamFilename.Unix.to_string rfile)
+                    OpamRepositoryPathName.repo_f
+                in
+                raise (Found (is_package || is_repo)))
+              () file;
+            false
+          with Found is_repo_or_package_dir -> not is_repo_or_package_dir))
+    (OpamFilename.files (root / OpamRepositoryPathName.repo_d))
```
rjbou: fold over repo list instead of files

tests/reftests/opamroot-versions.test
```diff
@@ -3567,3 +3567,305 @@ STATE                           Switch state loaded in 0.000s
 # Name            # Installed # Synopsis
 i-am-package      2           One-line description
 i-am-sys-compiler 2           One-line description
+### rm -rf _opam
```
rjbou, line 3570: add also a test with an archive present to see the hard upgrade

tests/reftests/repository.test
```diff
@@ -967,3 +966,577 @@ Now run 'opam upgrade' to apply any package updates.
...
+opam-version: "2.0"
+opam-root-version: "2.5.0"
+repositories: "old-tarred"
+installed-switches: "empty"
+switch: "empty"
+download-jobs: 3
+eval-variables: [
+  [
+    sys-ocaml-version
+    ["ocamlc" "-vnum"]
```
rjbou:
```suggestion
### head -2 OPAM/config
opam-version: "2.0"
opam-root-version: "2.5.0"
```

tests/reftests/repository.test
```diff
@@ -967,3 +966,577 @@ Now run 'opam upgrade' to apply any package updates.
...
```
rjbou:
```suggestion
### head -2 OPAM/config
opam-version: "2.0"
opam-root-version: "2.5.0"
```

src/client/opamRepositoryCommand.ml
rjbou: opened: #6994

tests/reftests/opamroot-versions.test
rjbou: done in repository test. It too complicated (and duplicates a lot) to add it here

rjbou:
The main comment will need a rewrite, and maybe the main commit message

src/core/opamFilename.mli
```diff
@@ -442,4 +442,8 @@ module Unix : sig
   (** Adds a dot and the given file extension *)
   val add_extension: t -> string -> t
 
+  (** Return [Error err] if path is absolute or contains '..'
+      Return [Ok filename] where trailing '/' are removed and './' at the
+      beginning *)
```
```suggestion
      Return [Ok filename] where trailing '/' and './' at the beginning
      are removed *)

File: src/repository/opamTar.ml [hunk 28]
```diff
@@ -195,12 +205,50 @@ let create tar dir =
...
+  let root_to_string = OpamFilename.to_string
+  let file_to_string = OpamFilename.Unix.to_string
+  let end_slash = Fun.id
+  let get_path ~fail:_ _target =
+    (* check if escapable ? *)
+    OpamFilename.Unix.of_string
+  let on_rejection _ _ _ = ()
+  let ext file ext = OpamFilename.Unix.add_extension file ext
+  let write file content target = Tar.add file content target
+  let exists file = Tar.exists file
```
kit-ty-kate:
```suggestion
  let ext = OpamFilename.Unix.add_extension
  let write = Tar.add
  let exists = Tar.exists
```

src/repository/opamTar.ml
```diff
@@ -195,12 +205,50 @@ let create tar dir =
...
+  let exists_dir _file _target = false
+  let read file = Tar.read file
+  let remove file = Tar.remove file
```
kit-ty-kate:
```suggestion
  let read = Tar.read
  let remove = Tar.remove
```

src/core/opamPatch.mli
```diff
@@ -19,18 +19,95 @@
...
+
+  (* Do we need to translate the patch file for this [root] *)
+  val translate_patch : bool
+
+  val root_to_string : root -> string
+  val file_to_string : file -> string
+
+  (* For directories, we need to add an end slash *)
+  val end_slash : root -> root
+
```
kit-ty-kate: This function only makes sense in get_path
```suggestion
```

src/repository/opamTar.ml
```diff
@@ -204,3 +204,36 @@ let create tar dir =
...
+  module Tar = Inplace
+  type file = OpamFilename.Unix.t
+  type target = Tar.t
+  let root_label = "archive"
+  let translate_patch = false
+  let root_to_string = OpamFilename.to_string
+  let file_to_string = OpamFilename.Unix.to_string
+  let end_slash = Fun.id
+  let get_path ~fail:_ _target =
+    (* check if escapable ? *)
```
kit-ty-kate: yes i think check_canonical should be used here

src/repository/opamTar.ml
```diff
@@ -64,6 +64,16 @@ let run archive =
...
+  let f acc filename content =
+    let filename =
+      match OpamFilename.Unix.check_canonical filename with
+      | Ok filename -> filename
+      | Error err ->
+        raise_error archive "Path '%s' not allowed: %s"
+                    (OpamFilename.Unix.to_string filename) err
+    in
+    f acc filename content
+  in
```
rjbou, line 76: An update in the documentation of the module is needed to highlight that check & "rewrite"

src/core/opamFilename.mli
```diff
@@ -420,9 +427,31 @@ module Unix : sig
...
+
+  (** Return the base name *)
+  val basename: t -> Base.t
+
+  (** Return the directory name *)
+  val dirname: t -> Dir.t
+
+  (** Adds a dot and the given file extension *)
+  val add_extension: t -> string -> t
+
```
kit-ty-kate: This function isn't used anywhere (it's only used was in PatchFS but its result is ignored)
```suggestion
```

src/core/opamPatch.mli
```diff
@@ -19,18 +19,95 @@
...
+  (* Identity unless some checks need to be done. Take a [fail] function. *)
+  val get_path : fail:(unit -> unit) -> root -> string -> file
+
+  (* [file] extension *)
+  val ext : file -> string -> file
+
+  (* Write a [file] *)
+  val write : file -> string -> target -> target
```
kit-ty-kate:
```suggestion
  (* Write a [file] *)
  val write : file -> ?ext:string -> string -> target -> target
```

src/core/opamPatch.ml
```diff
@@ -322,87 +322,159 @@ let translate_patch ~dir orig corrected =
...
-      let content = OpamSystem.read file in
-      let content = patch ~file:file (Some content) diff in
-      OpamSystem.write file content;
-      if file1_exists && file1 <> (file2 : string) then
-        OpamSystem.rmdir_cleanup (Filename.dirname file1)
+      let content = FS.read file patching in
+      let patching, content = patch file (Some content) diff patching in
+      let patching = FS.write file content patching in
+      let patching =
+        if file1_exists && file1 <> (file2 : FS.file) then
```
kit-ty-kate: `FS.file` is abstract. There is no guarantee on the structure of it so we need to use a proper equality function
```suggestion
        if file1_exists && not (FS.eq_file file1 file2) then
```

src/core/opamPatch.ml
```diff
@@ -322,87 +322,159 @@ let translate_patch ~dir orig corrected =
...
+    let fail () =
       internal_patch_error "Patch %S tried to escape its scope."
-        patch_info_path;
-    file
+        patch_info_path
+    in
+    FS.get_path ~fail to_patch file
   in
-  let patch ~file content diff =
+ let patch file content diff patching =
```
kit-ty-kate:
```suggestion
  let patch file content diff patching =
```

kit-ty-kate (commented):
partial review up to c1094b255436285a65b8a64d74becdf80aec4001 (also partially)

src/core/opamFilename.ml
```diff
@@ -450,25 +450,48 @@ let link ?(relative=false) ~target ~link =
...
+  let translate_patch = true
+  let root_to_string root = root
+  let file_to_string file = file
+  let end_slash dir = Filename.concat (OpamSystem.real_path dir) ""
+  let get_path ~fail dir file =
+    let file = OpamSystem.real_path (Filename.concat dir file) in
+    if not (OpamStd.String.is_prefix_of ~from:0 ~full:file dir) then
+      fail ();
+    file
+  let ext file ext = file ^ ext
```
rjbou:
```suggestion
  let ext file ext = file ^ "." ^ ext

File: src/core/opamFilename.mli (hunk 33 above)
Comment (rjbou) [reply src/core/opamFilename.mli]: The result is stored in the archive, which is not the best. Should we drop it, or as rejection, have a `FS.on_accept` that writes only on disk ?

File: src/core/opamFilename.ml (hunk 37 above)
Comment (kit-ty-kate) [reply src/core/opamFilename.ml]: This doesn't match the behaviour of its usage in `OpamPatch.patch_t`

File: src/core/opamFilename.mli (hunk 33 above)
Comment (kit-ty-kate) [reply src/core/opamFilename.mli]: Technically it could be used but in practice `OpamPatch.patch` is never called with `~allow_unclean:true` on a repository so it will never get used

File: src/core/opamFilename.ml [hunk 38]
```diff
@@ -800,18 +816,83 @@ module Unix = struct
...
+            in
+            Ok res
+    in
+    match String.split_on_char dir_sep.[0] filename with
+    | [] -> assert false
+    (* ^/ : fail *)
+    | ""::_ -> Error "begins with '/'"
+    (* ^./ : delete the ./ *)
+    | "."::elements -> check_parent_dir_and_remove_trailing_slashes elements
+    | [_] -> Ok filename
```
kit-ty-kate: this ignores the `".."` case as well as `"./"`

src/core/opamFilename.ml
```diff
@@ -800,18 +816,83 @@ module Unix = struct
...
+    | elements -> check_parent_dir_and_remove_trailing_slashes elements
```
kit-ty-kate: i'd rather replace this with the version from https://github.com/ocaml/opam/pull/6924/changes/3f057a83fa29a3172cab54e1ad991a1412b170b2
```suggestion
  let to_relative_canonical_list x =
    if Filename.is_relative x then
      match String.split_on_char '/' x with
      | [] | ""::_::_ -> assert false
      | [""] -> [current_dir_name]
      | x ->
        List.filter (fun seg ->
            if seg = ".." then
              invalid_arg "'..' is forbidden"
            else
              seg <> "" && not (String.equal current_dir_name seg))
          x
    else
      invalid_arg "not relative"

  let check_relative_canonical filename =
    try Ok (String.concat dir_sep (to_relative_canonical_list filename))
    with Invalid_argument msg -> Error msg
```

src/core/opamFilename.mli
kit-ty-kate: Then i guess adding an `FS.on_unclean_accept` should be fine

src/core/opamPatch.mli
```diff
@@ -19,18 +19,95 @@
...
+
+  (* Function called when a patch application fails
+
+     @param file is the filename on which the patch was rejected
+     @param content is the file content, if it exists
+     @param diff is the patch that failed to apply
+  *)
+  val on_rejection : file -> string option -> Patch.t -> unit
```
kit-ty-kate:
```suggestion
  val on_unclean_rejection : file -> string option -> Patch.t -> unit
```

src/repository/opamTar.ml
```diff
@@ -195,12 +205,50 @@ let create tar dir =
...
+  let write file content target = Tar.add file content target
+  let exists file = Tar.exists file
+  let exists_dir _file _target = false
+  let read file = Tar.read file
+  let remove file = Tar.remove file
+  let remove_dir file target =
+    Tar.remove_dir (OpamFilename.Unix.dirname file) target
+  let same_dirname ~src ~dst =
+    OpamFilename.Unix.dirname src
+    <> (OpamFilename.Unix.dirname dst : OpamFilename.Unix.Dir.t)
```
kit-ty-kate:
```suggestion
    OpamFilename.Unix.Dir.equal
      (OpamFilename.Unix.dirname src)
      (OpamFilename.Unix.dirname dst)
```

src/core/opamFilename.ml
```diff
@@ -450,25 +450,48 @@ let link ?(relative=false) ~target ~link =
...
+    write (file^".rej") (Format.asprintf "%a" Patch.pp diff) ()
+  let exists file _target = Sys.file_exists file
+  let exists_dir file _target =
+    let dir = Filename.dirname file in
+    Sys.file_exists dir && Sys.is_directory dir
+  let read file _target = OpamSystem.read file
+  let remove file _target = OpamSystem.remove_file file
+  let remove_dir file _target = OpamSystem.rmdir_cleanup (Filename.dirname file)
+  let same_dirname ~src ~dst =
+    Filename.dirname src <> (Filename.dirname dst : string)
```
kit-ty-kate:
```suggestion
    Filename.dirname src = (Filename.dirname dst : string)
```

src/core/opamPatch.ml
```diff
@@ -322,87 +322,159 @@ let translate_patch ~dir orig corrected =
...
-      OpamSystem.mv src dst;
-      let dirname_src = Filename.dirname src in
-      if dirname_src <> (Filename.dirname dst : string) then
-        OpamSystem.rmdir_cleanup dirname_src
+      let patching = FS.mv ~src ~dst patching in
+      let patching =
+        if FS.same_dirname ~src ~dst then
+          FS.remove_dir src patching
+        else
+          patching
```
kit-ty-kate:
```suggestion
        if FS.same_dirname ~src ~dst then
          patching
        else
          FS.remove_dir src patching
```

src/repository/opamRepositoryRoot.mli
```diff
@@ -76,12 +115,32 @@ val dirname : t -> dirname
 val basename : t -> basename
 val to_string : t -> string
 
+val remove_prefix: filename -> t -> filename
+val remove_prefix_dir: dirname -> t -> dirname
```
kit-ty-kate: It would otherwise be inconsistent with `OpamFilename` etc
```suggestion
val remove_prefix: t -> filename -> filename
val remove_prefix_dir: t -> dirname -> dirname
```

tests/reftests/run.ml
```diff
@@ -664,6 +664,10 @@ let common_filters ?opam dir =
       str "opam-";
       rep1 (alt [xdigit; char '-'])],
     Sed "${OPAMTMP}";
+    (* We need this sed to ensure that the line containing 'packages' is
+       rewritten with the good dir sep replacements *)
+    seq [ str "packages"; set "/\\" ],
```
kit-ty-kate: why would the rest of the path only contains `/` but only the packages part would contain `\`?

src/client/opamArg.ml
```diff
@@ -224,10 +224,10 @@ let environment_variables =
       "enables option --no-checksums when available.";
       "REPOSITORYTARRING", cli_from cli2_2,
       (fun b -> REPOSITORYTARRING (env_bool b)),
-      "internally store the repositories as tar.gz files. This can be much \
-       faster on filesystems that don't cope well with scanning large trees \
-       but have good caching in /tmp. However this is slower in the \
-       general case.";
+      "internally store all non-vcs repositories as tar.gz files (by default, \
+       http repositories are stored as archive).This can be much faster on \
```
kit-ty-kate:
```suggestion
       http repositories are stored as archive). This can be much faster on \
```

src/client/opamCommands.ml
```diff
@@ -2764,7 +2762,7 @@ let switch cli =
     mk_subdoc ~cli ~defaults:["","list";"SWITCH","set"]
       ~extra_defaults:[
         cli_from cli2_2, "-",
-        "Switches back to the previous non-local switch (similar to $(b,git switch -)). \
+         "Switches back to the previous non-local switch (similar to $(b,git switch -)). \
```
```suggestion
        "Switches back to the previous non-local switch (similar to $(b,git switch -)). \
```

src/format/opamRepositoryPath.mli
```diff
@@ -104,7 +104,8 @@ module Make (I : PATH_REPR) : PATH
 
 (* {2} Other paths *)
 
-val tar: dirname -> repository_name -> filename
+(** obsolete, no longer used *)
+val repo_tarring: dirname -> repository_name -> filename
```
kit-ty-kate: If it's not used, why not simply remove it?
```suggestion
```

src/repository/opamRepositoryRoot.mli
```diff
@@ -76,12 +115,32 @@ val dirname : t -> dirname
...
+
+val is_tgz: t -> bool
+val is_dir: t -> bool
```
kit-ty-kate: is_dir isn't used and is_tgz is only used to get the name so it would make more sense to simply get the name directly from OpamRepositoryRoot
```suggestion
val get_backend_name: t -> string
```

src/repository/opamRepositoryRoot.mli
```diff
@@ -90,3 +149,13 @@ val patch :
 (** Returns a pair [(exists, f)] where [exists] tells whether the
     [repo] file exists in the repository and [f] reads it *)
 val delayed_read_repo : t -> bool * (unit -> OpamFile.Repo.t)
+
+(** TAR TODOC remove directory and archive, if present *)
```
kit-ty-kate:
```suggestion
(** remove directory and archive, if present *)
```

src/state/opamFileTools.ml
```diff
@@ -1321,17 +1365,40 @@ let add_aux_files ?dir ?(files_subdir_hashes=false) opam =
...
+    let url_file : OpamFile.URL_legacy.t OpamFile.t =
       OpamFile.make (dir // "url")
     in
-    let (descr_file: OpamFile.Descr_legacy.t OpamFile.t)  =
+    let descr_file : OpamFile.Descr_legacy.t OpamFile.t =
       OpamFile.make (dir // "descr")
     in
-    let files_dir =
-      OpamFilename.Op.(dir / OpamRepositoryPathName.files_d)
+    let files_dir = OpamFilename.Op.(dir / "files") in
```
kit-ty-kate:
```suggestion
    let files_dir = OpamFilename.Op.(dir / OpamRepositoryPathName.files_d) in
```

tests/reftests/action-disk.test
```diff
@@ -110,8 +110,12 @@ SYSTEM                          mkdir ${BASEDIR}/OPAM/repo/default/packages/main
 SYSTEM                          mkdir ${BASEDIR}/OPAM/repo/default/packages/main-repo/main-repo.2/files
 FILE(repo)                      Read ${BASEDIR}/OPAM/repo/default/repo in 0.000s
 Processing: [default: loading data]
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/default/packages/main-repo/main-repo.2/opam in 0.000s
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/default/packages/main-repo/main-repo.1/opam in 0.000s
+SYSTEM                          read ${BASEDIR}/OPAM/repo/default/packages/main-repo/main-repo.1/opam
+FILE(opam)                      Read packages/main-repo/main-repo.1/opam (out of ${BASEDIR}/OPAM/repo/default) in 0.000s
+SYSTEM                          read ${BASEDIR}/OPAM/repo/default/packages/main-repo/main-repo.1/files/x-file.main-repo.1
```
kit-ty-kate, line 115: why are we suddenly reading the extra-files?

tests/reftests/repository-layout.test
```diff
@@ -147,15 +147,21 @@ RSTATE                          Cache found
...
 UPDATE                          Repository has new changes
 RSTATE                          load repo new-default from dir
-FILE(opam)                      Read ${BASEDIR}/OPAM/repo/new-default/packages/good/good.1/opam in 0.000s
+FILE(opam)                      Read packages/good/good.1/opam (out of ${BASEDIR}/OPAM/repo/new-default) in 0.000s
+### OPAMDEBUGSECTIONS="FILE(opam) opam-file RSTATE UPDATE" OPAMDEBUG=-3 opam repository add new-default ./REPO --this-switch | unordered
+[new-default] no changes from file://${BASEDIR}/REPO
+UPDATE                          Repository did not change: nothing to do.
+FILE(config)                    Read ${BASEDIR}/OPAM/config in 0.000s
+RSTATE                          LOAD-REPOSITORY-STATE @ ${BASEDIR}/OPAM
+RSTATE                          Cache found
```
kit-ty-kate: I believe this is an artefact of the rebase over https://github.com/ocaml/opam/pull/6941
```suggestion
```

tests/reftests/repository.test
```diff
@@ -62,7 +62,6 @@ plain
 repos-config
 tarred.tar.gz
 ### opam install foo.3 -vv | grep '^\+' | sed-cmd test tar
```
kit-ty-kate:
```suggestion
### opam install foo.3 -vv | grep '^\+' | sed-cmd test
```

master_changes.md
```diff
@@ -71,6 +71,7 @@ users)
 ## Repository
   * No longer call tar tool to create archives, use tar library instead [#6945 @kit-ty-kate]
   * [BUG] Do not fail on directories named `opam` when scanning the `packages` directory of a repository during `opam repo add` or `opam init` (worked on subsequent `opam update`) [#6941 @kit-ty-kate @rjbou]
+  * Speedup repository operations on Windows & BSD by changing its storage in the opam root from plain directory to archive, for HTTP repositories, or non-VCS one if `OPAMREPOSITORYTARRING` is enabled [#6625 @rjbou @kit-ty-kate @arozovyk - fix #5741 #5648 #5484 #5559 #3050]
```
```suggestion
  * Speedup repository operations on certain file-systems (e.g. NTFS on Windows or IO constrained machines) by changing its storage in the opam root from plain directory to archive, for HTTP repositories, or non-VCS one if `OPAMREPOSITORYTARRING` is enabled [#6625 @rjbou @kit-ty-kate @arozovyk - fix #5741 #5648 #5484 #5346 #5559 #3050 #6974]
```

kit-ty-kate (commented):
lgtm modulo green CI otherwise

tests/reftests/run.ml
rjbou: The function `str_replace_path` handles the reste of the path

tests/reftests/action-disk.test
rjbou: On loading, there is a checksum checks, extra files were directly read by `Digest.file`, while here, we read them and use  `Digest.string`

rjbou:
Should have resolved all suggestions. I've pushed twice (without and with rebase, to highlight the diff). Test branch is updated too.

src/core/opamFilename.ml
```diff
@@ -462,25 +462,51 @@ let link ?(relative=false) ~target ~link =
...
+    let file = OpamSystem.real_path (Filename.concat dir file) in
+    if not (OpamStd.String.is_prefix_of ~from:0 ~full:file dir) then
+      fail ();
+    file
+  let write file content _target = OpamSystem.write file content
+  let on_unclean_accept file content =
+    Option.iter (fun c -> write (file^".orig") c ()) content
+  let on_unclean_reject file content diff =
+    on_unclean_accept file content;
+    Option.iter (fun c -> write (file^".orig") c ()) content;
```
kit-ty-kate: This line seems duplicated with the new call just above
```suggestion
```

rjbou:
Updated

rjbou (approved):

kit-ty-kate:
Ignoring cygwin CI failure (fixed by #7007) and random mingw failure.

:tada:

PR #6631, opened 2025-08-07
Improve the opam pin man page by being more explicit about which arguments are optional

kit-ty-kate:
Inspired by https://discuss.ocaml.org/t/opam-pin-does-not-have-the-desired-effect/17084/7


dra27 (approved):
LGTM!

PR #6634, opened 2025-08-08
Use a C binding instead of system utilities to get the number of cores of the current machine

### LINKED ISSUE #6632: C binding to fetch the number of processors available
While [testing the Haiku port of OCaml](https://github.com/ocaml/ocaml/pull/14138) i tried to compile opam, which worked flawlessly.

`opam` compiles fine and executes fine, however `opam init` will, by default, check for `getconf`, which isn't builtin at the moment on Haiku (it might be in the future given that [it is part of POSIX](https://pubs.opengroup.org/onlinepubs/9799919799/utilities/getconf.html)). This can be worked around by using `opam init --bypass-checks` but i think it would be nice to replace the platform specific code that makes opam require `getconf` (or `sysctl` on FreeBSD) by some [C code](https://stackoverflow.com/questions/150355/programmatically-find-the-number-of-cores-on-a-machine).
This would be nice to further reduce the number of system utilities opam depends on.

kit-ty-kate:
Fixes #6632 

The only remaining use of `getconf` (`getconf LONG_BIT`) is optional and ignored if missing or failing so we can remove the requirement.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -19,6 +19,7 @@ users)
 ## Plugins
 
 ## Init
+  * Remove `getconf` from the list of required runtime tools, which allows `opam init` to work out-of-the-box on Haiku [#6634 @kit-ty-kate - fix #6632]
 
 ## Config report
 
--- a/src/client/opamInitDefaults.ml
+++ b/src/client/opamInitDefaults.ml
@@ -47,15 +47,12 @@ let dragonflybsd_filter = os_filter "dragonfly"
 let not_bsd_filter =
   FNot (FOr (FOr (openbsd_filter, netbsd_filter),
              FOr (freebsd_filter, dragonflybsd_filter)))
-let win32_filter = os_filter "win32"
 let not_win32_filter =
   FOp (FIdent ([], OpamVariable.of_string "os", None), `Neq, FString "win32")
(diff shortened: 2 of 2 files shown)


src/core/opamSystem.ml
```diff
@@ -857,18 +857,13 @@ let install ?(warning=default_install_warning) ?exec src dst =
...
-      match Sys.os () with
-      | Sys.Win32 -> [Env.get "NUMBER_OF_PROCESSORS"]
-      | Sys.FreeBSD -> read_command_output ~verbose:(verbose_for_base_commands ())
-                         ["sysctl"; "-n"; "hw.ncpu"]
-      | _ -> read_command_output ~verbose:(verbose_for_base_commands ())
-               ["getconf"; "_NPROCESSORS_ONLN"]
-    in
-    int_of_string (List.hd ans)
-  with Not_found | Process_error _ | Failure _ -> 1
+  (* TODO: Use Domain.recommended_domain_count with OCaml >= 5.0 *)
```
dra27: No - definitely don't do that!

rjbou:
```suggestion
```

rjbou (commented):
With a changelog, lgtm!

PR #6638, opened 2025-08-12
Invalid character in package name should explain why

lefessan:


kit-ty-kate (approved):
Thanks! I've added some tests to actually show up the feature and opened a follow up PR to handle non-displayable/UTF-8 characters better in https://github.com/ocaml/opam/pull/6640

Issue #6396, opened 2025-02-25
Invalid character in package name clarification


MisterDA:
```
opam: PACKAGES… arguments: Invalid character in package name
```

Yes, but what's the invalid character in the package name? Can we print it, or its position?
(Is it a hyphen?)

kit-ty-kate:
I'm unable to reproduce the problem:
```
$ opam install test@invalid
opam: PACKAGES… arguments: Invalid character in package name "test@invalid"
```
I'm also unable to locate any `Invalid character in package name` error message without the package name in question in our code base. How did you get that error message?

MisterDA:
Sorry, I removed the package name. I don't have it in my logs anymore… but it would be nice if the invalid character would be highlighted, or a regex shown, if it must validate one. The invalid character is shown when the version is invalid.

I grepped and saw:
```
src/format/opamPackage.ml
31:            (Printf.sprintf "Invalid character '%c' in package version %S" c x))
79:        (Printf.sprintf "Invalid character in package name %S" x)

src/format/opamSwitch.ml
34:    failwith (Printf.sprintf "Invalid character in switch name %S" s);
```

I don't think `%S` should be used, the strings don't need to be converted to the OCaml syntax.

PR #6640, opened 2025-08-13
Handle non-displayable characters when detecting an erroneous package name or version

kit-ty-kate:
Follow up to #6638
~Queued on #6638~
related to #6396

### DIFF
--- a/tests/reftests/show.test
+++ b/tests/reftests/show.test
@@ -457,7 +457,7 @@ Usage: opam show [OPTION]… [PACKAGES]…
 Try 'opam show --help' or 'opam --help' for more information.
 # Return code 2 #
 ### opam show bépo
-opam: PACKAGES… arguments: Invalid character '�' in package name
+opam: PACKAGES… arguments: Invalid character '\195' in package name
       "b\195\169po"
 Usage: opam show [OPTION]… [PACKAGES]…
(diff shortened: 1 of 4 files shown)


tests/reftests/show.test
```diff
@@ -457,8 +457,14 @@ Usage: opam show [OPTION]… [PACKAGES]…
 Try 'opam show --help' or 'opam --help' for more information.
 # Return code 2 #
 ### opam show bépo
-opam: PACKAGES… arguments: Invalid character '�' in package name
+opam: PACKAGES… arguments: Invalid character '\195' in package name
```
rjbou, line 460: Shouldn't these test be on their specific file ?

rjbou (commented):
On the idea, lgtm! Some comment though
* add the test (opam show bépo) before the fix
* changelog
* a more explicit comit message for commit #2: minor simplification in OpamFormula.atom_of_string for ex

kit-ty-kate:
> Some comment though

yeah sorry, i didn't clean up the PR at the time because it was queued on another PR so it needed rebasing anyway and i forgot to do it during the post-merge rebase.

tests/reftests/show.test
kit-ty-kate: with just error cases or something like that? Why not, although i don't feel a strong need for it personally

PR #6641, opened 2025-08-14
[Haiku] Add missing references to Haiku package manager

Sylvain78:
Add missing references to Haiku package manager.

### DIFF
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -123,6 +123,7 @@ type families =
   | Dummy of test_setup
   | Freebsd
   | Gentoo
+  | Haiku
   | Homebrew
   | Macports
   | Msys2
@@ -191,6 +192,7 @@ let family ~env () =
       end
     | "debian" | "ubuntu" -> Debian
     | "gentoo" -> Gentoo
+    | "haiku" -> Haiku
     | "homebrew" -> Homebrew
     | "macports" -> Macports
     | "nixos" -> Nix
@@ -863,6 +865,13 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
       |> package_set_of_pkgpath
     in
     compute_sets sys_installed
+  | Haiku ->
+    let sys_installed =
+      run_query_command "pkgman" ["search"; "%n\n%o"]
+      |> List.map OpamSysPkg.of_string
+      |> OpamSysPkg.Set.of_list
+    in
+    compute_sets sys_installed
   | Homebrew ->
     (* accept 'pkgname' and 'pkgname@version'
        exampe output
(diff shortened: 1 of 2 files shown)


src/state/opamSysInteract.ml
```diff
@@ -1261,6 +1272,7 @@ let update ?(env=OpamVariable.Map.empty) config =
       if test.install then None else Some (`AsUser "false", [])
     | Freebsd -> None
     | Gentoo -> Some (`AsAdmin "emerge", ["--sync"])
+    | Haiku -> Some (`AsAdmin "pkgman", ["update"])
```
kit-ty-kate, line 1275:
```suggestion
    | Haiku -> Some (`AsUser "pkgman", ["full-sync"]) 
```

src/state/opamSysInteract.ml
```diff
@@ -1026,6 +1035,7 @@ let package_manager_name_t ?(env=OpamVariable.Map.empty) config =
       `AsUser "false"
   | Freebsd -> `AsAdmin "pkg"
   | Gentoo -> `AsAdmin "emerge"
+  | Haiku -> `AsAdmin "pkgman"
```
kit-ty-kate, line 1038:
```suggestion
  | Haiku -> `AsUser "pkgman"
```

src/state/opamSysInteract.ml
```diff
@@ -863,6 +865,13 @@ let packages_status ?(env=OpamVariable.Map.empty) config packages =
       |> package_set_of_pkgpath
     in
     compute_sets sys_installed
+  | Haiku ->
+    let sys_installed =
+      run_query_command "pkgman" ["search"; "%n\n%o"]
+      |> List.map OpamSysPkg.of_string
+      |> OpamSysPkg.Set.of_list
```
kit-ty-kate, line 872: This doesn't work at all. I've looked into it and the best we can do seems to be `-a`/`--all` but we need to extract a list from that (extract the second column after the `-----` line). We could even extract the list of installed packages this way by looking at the first column (if it contains `s`, `S`, `h` or `H` then it is installed). Or we could wait for https://github.com/ocaml/opam/pull/6489 and split the two into `pkgman search -a` and `pkgman search -i -a`
```suggestion
      run_query_command "pkgman" ["search"; "-a"]
      |> some_post_processing
```

src/state/opamSysInteract.ml
kit-ty-kate: Haiku doesn't seem to have sudo/doas or any sort of distinction between admin and users as far as the package manager is concerned. It has `su` though so maybe? Do you know where this is documented? I couldn't find if haiku has a notion of superuser/privileged user

PR #6442, opened 2025-03-30
Cleanup OpamStd of duplicated functions that already exist in Stdlib

kit-ty-kate:
A little spring cleaning

PR #6644, opened 2025-08-18
Replace every polymorphic uses of List.mem by a version that doesn't use Repr.equal

kit-ty-kate:
Some clean up.

See https://github.com/ocaml/ocaml/pull/9928 for the underlying reason why `Repr`/`(=)`/`Stdlib.compare`/... should be avoided.

~Queued on top of https://github.com/ocaml/opam/pull/6442~

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -102,6 +102,7 @@ users)
 ## Shell
 
 ## Internal
+  * Replace every polymorphic uses of `List.mem` by a version that doesn't use `Repr.equal` [#6644 @kit-ty-kate]
 
 ## Internal: Unix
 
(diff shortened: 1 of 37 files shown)


master_changes.md
```diff
@@ -167,6 +168,7 @@ users)
   * `OpamCompat.String.{starts_with,ends_with,for_all,fold_left}`: were added [#6442 @kit-ty-kate]
   * `OpamStd.List.fold_left_map`: was moved to `OpamCompat.List.fold_left_map` [#6442 @kit-ty-kate]
   * `OpamStd.List.{cons,find_opt,filter_map}`: were removed. Use `Stdlib.List` instead. [#6442 @kit-ty-kate]
+  * `OpamStd.List.mem`: was added [#6644 @kit-ty-kate]
```
```suggestion
  * `OpamStd.List.mem`: was added, having as argument the equality function [#6644 @kit-ty-kate]
```

master_changes.md
```diff
@@ -167,6 +168,7 @@ users)
   * `OpamCompat.String.{starts_with,ends_with,for_all,fold_left}`: were added [#6442 @kit-ty-kate]
```
rjbou, line 188: Missing equality function that are added

Issue #5968, opened 2024-05-22
opam broken on ubuntu 24.04 (`bwrap: operation not permitted`)


dlesbre:
I've recently upgraded my computer from kubuntu 23.10 to kubuntu 24.04. This lead to errors with opam. Specifically, `opam upgrade` failed at trying to build the first package (`dune.3.15.2`), and `opam switch create 4.14.1` also failed with the same error message:
```
bwrap: loopback: Failed RTM_NEWADDR: Operation not permitted
```

Doing some digging, it seems to be related to ubuntu deciding to restrict unprivileged user namespaces https://ubuntu.com/blog/ubuntu-23-10-restricted-unprivileged-user-namespaces. I found a fix, detailed here: https://etbe.coker.com.au/2024/04/24/ubuntu-24-04-bubblewrap/ which got opam working again.

Config:
```
<include here the output of `opam config report`, if applicable>
# opam config report
# opam-version         2.1.5 
# self-upgrade         no
# system               arch=x86_64 os=linux os-distribution=ubuntu os-version=24.04
# solver               builtin-mccs+glpk
# install-criteria     -removed,-count[avoid-version,changed],-count[version-lag,request],-count[version-lag,changed],-count[missing-depexts,changed],-changed
# upgrade-criteria     -removed,-count[avoid-version,changed],-count[version-lag,solution],-count[missing-depexts,changed],-new
# jobs                 19
# repositories         2 (http) (default repo at 76de4586)
# pinned               1 (rsync)
# current-switch       4.14
# ocaml:native         true
# ocaml:native-tools   true
# ocaml:native-dynlink true
# ocaml:stubsdir       /home/dorian/.opam/4.14/lib/ocaml/stublibs:/home/dorian/.opam/4.14/lib/ocaml
# ocaml:preinstalled   false
# ocaml:compiler       4.14.1
```

cuihtlauac:
Thanks @dlesbre, I've experienced the same issue. Upgrade to 24.04, and then `opam upgrade` failure. Doing as suggested by @etbe worked. I'm copying his suggestion here:
> To resolve that you could upgrade to SE Linux, but the other option is to create a file named `/etc/apparmor.d/bwrap` with the following contents:
```
abi <abi/4.0>,
include <tunables/global>

profile bwrap /usr/bin/bwrap flags=(unconfined) {
  userns,

  # Site-specific additions and overrides. See local/README for details.
  include if exists <local/bwrap>
}
```
[...]

PR #6647, opened 2025-08-22
Add AppArmor support to the install script

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/5968
Was tested with success in https://github.com/ocaml/opam/issues/5968#issuecomment-2848669730


rjbou (approved):
Light approval. This PR was tested successfully by users that had the issue, and i tested it locally successfully too. But I'm not an apparmor expert, it's difficult for me to validate if it is the good level of permission (though it is mimicking official opam one).
Let's merge it and use the 2.5 alpha to supervise and have feedback on it.

PR #6394, opened 2025-02-21
GHA: add a job that test opam libs dependencies compilation

rjbou:
In it based on the same mechanism than opam-rt, it permits to keep them synchronised with the current API
TODO
* [x] add a release step to check CI log for failing projects
* [x] add a mechanism to retrieve automatically opam lib rev deps
* [x] add a mechanism to eliminate pinned packages that rely on old versions of opam
* [x] condition job launch to changes of mli files

This PR also contains an enhancement of changed files step (and it use/conditionning).

PR #6648, opened 2025-08-22
GHA: Ensure every part of the scripts are run with 'set -ue'

kit-ty-kate:
Follow-up to https://github.com/ocaml/opam/pull/6394#discussion_r2294659403 to ensure similar problems don't happen in the future.

`set -o pipefail` could also be added at some point in the future for even more safety, however this one is usually a bit trickier.


.github/scripts/main/main.sh
```diff
@@ -94,25 +92,23 @@ fi
 if [ "$OPAM_TEST" = "1" ]; then
   # test if an upgrade is needed
   set +e
```
rjbou: shouldn't this line be removed ?

kit-ty-kate: oops, good catch. Removed.

rjbou (approved):

PR #6650, opened 2025-08-24
Fix URL to Software Heritage

gahr:
It's currently interpreted as a relative path, and in the rendered version over at https://opam.ocaml.org it points to https://opam.ocaml.org/doc/www.softwareheritage.org.


kit-ty-kate:
Thanks a lot!

PR #6655, opened 2025-08-28
GHA: reverse dependencies job follow-up

arozovyk:
Follow-up to #6394  
- Pin local dev repo of a package that has correctly identified branch, (instead of pinning with `kind git`)
- Filter out packages that are not coinstallable with opam libs.


kit-ty-kate (approved):

kit-ty-kate:
Thanks!

PR #6659, opened 2025-08-28
Fix var opamfile

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/5346


tests/reftests/var-option.test
```diff
@@ -775,4 +775,4 @@ The following actions will be performed:
 -> installed test.1
 Done.
 ### opam var test:opamfile
-${BASEDIR}/OPAM/repo/default/packages/test/test.1/opam
+${BASEDIR}/OPAM/var-opamfile/.opam-switch/packages/test.1/opam
```
rjbou, line 786: As you are specifically testing the opamfile variables, it worth adding a test for all cases : not installed, pinned and not installed, not pinned & installed, present or not in the repository.

kit-ty-kate: i agree. done.

src/state/opamPackageVar.ml
```diff
@@ -308,20 +308,14 @@ let resolve st ?opam:opam_arg ?(local=OpamVariable.Map.empty) v =
...
-          |> string
-        )
+      let nv = get_nv opam in
+      if OpamPackage.Set.mem nv st.installed then
+        OpamPath.Switch.installed_opam st.switch_global.root st.switch nv
+        |> OpamFile.to_string
+        |> string
+        |> Option.some
+      else
+        None
```
rjbou, line 325: To keep the same behaviour, we need to keep the old computation here to be able to retrieve the information during package's actions (build/install)

kit-ty-kate: Our own manual says:
> opamfile: if the package is installed, path of its opam file, from opam internals, otherwise not defined

I'm not sure we should invite users to use `_:opamfile` in `build`/`install`/…

rjbou: It can be used in wrappers. We can check if it is already written for post install wrappers, but it won't resolve the issue for pre build wrappers.

rjbou (approved):
I think the is the good way to do it, especially if we don't have the invariant about `packages/` directory.

I've added a test for internal variables in an opam file, that completes the one in `var-option` that is specific for `opamfile`.

Maybe the PR / changes need to be updated as we change the specification, so in it no longer making it match the old specification.

src/state/opamPackageVar.ml
```diff
@@ -309,20 +309,20 @@ let resolve st ?opam:opam_arg ?(local=OpamVariable.Map.empty) v =
...
-        )
+      let nv = get_nv opam in
+      let ret file = Some (string (OpamFile.to_string file)) in
+      if OpamPackage.Set.mem nv st.installed then
+        ret (OpamPath.Switch.installed_opam st.switch_global.root st.switch nv)
+      else if OpamPackage.Set.mem nv st.pinned then
+        ret (OpamPath.Switch.Overlay.opam st.switch_global.root st.switch (OpamPackage.name nv))
+      else if OpamPackage.Set.mem nv st.packages then
+        let opam = OpamPackage.Map.find nv st.opams in
+        let temp_dir = OpamFilename.mk_tmp_dir () in
```
rjbou, line 320: This temporary file won't be cleaned, but it is the wanted behaviour.

tests/reftests/dune.inc
```diff
@@ -1343,6 +1343,69 @@
...
+ (action
+  (diff opamfile-internal-variables.win32.test opamfile-internal-variables.win32.out)))
+
+(alias
+ (name reftest)
+ (enabled_if (and (= %{os_type} "Win32") (or (<> %{env:TESTALL=1} 0) (= %{env:TESTN0REP0=0} 1))))
+ (deps (alias reftest-opamfile-internal-variables.win32)))
+
+(rule
+ (targets opamfile-internal-variables.win32.out)
```
rjbou: i'm not sure about the name of the test...

tests/reftests/opamfile-internal-variables.test
```diff
@@ -0,0 +1,1024 @@
...
+### opam switch create global-wrappers --empty
+### opam install dep
+The following actions will be performed:
+=== install 1 package
+  - install dep 1
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> installed dep.1
+Done.
+### opam install i-am-a-repo-pkg -vv | "${OPAMVERSION}" -> "++current++" | '[0-9a-z]{64}' -> '+hash+' | $MD5 -> +archive-hash+ | sed-cmd bash tar rsync | etc -> etc
```
rjbou: `| etc -> etc` is a workaround to force installed-files lines to go into path rewriter. I don't know if it worth have the installed-files moved in the windows/unix test to avoid that or have the (more) exact output ?

tests/reftests/dune.inc
kit-ty-kate: I renamed it hooks-variables as it is meant to show the values of variables during command hooks

tests/reftests/opamfile-internal-variables.test
kit-ty-kate: I think it's fine for now. Maybe we can think of something to handle path rewriting better in the future but it'll do for now

src/state/opamPackageVar.ml
rjbou: In fact, there is no need to keep it, as it is only available during opam run.

tests/reftests/dune.inc
rjbou: It is not only a hook only test, it contains a test of the plain texts in opam file.

PR #6661, opened 2025-08-28
Add OpamHash.check_string

kit-ty-kate:


src/core/opamHash.mli
```diff
@@ -43,6 +43,7 @@ val to_path: t -> string list
 val sort : t list -> t list
 
 val check_file: string -> t -> bool
+val check_string: string -> t -> bool
```
rjbou, line 49: It worth to take the opportunity to add some doc (even if it is obvious)

```suggestion
(** Compute the hash of the file and check it *)
val check_file: string -> t -> bool

(** Compute the hash of the string and check it *)
val check_string: string -> t -> bool
```

rjbou (approved):

PR #6662, opened 2025-08-28
Remove the unused OpamSwitchState.files function

kit-ty-kate:


rjbou (approved):
Is there a reason for the remove instead of unused ?

PR #6666, opened 2025-09-01
Gha: Fix duplication logic in revdeps script

arozovyk:
Currently the job ignores every package 
https://github.com/ocaml/opam/actions/runs/17376031670/job/49322707535#step:11:449


kit-ty-kate:
Thanks!

PR #6672, opened 2025-09-04
opam remove --force` no longer run commands in current directory

### LINKED ISSUE #6570: `opam remove --force` launches remove commands in current directory
If you force removal of an uninstalled package, opam launches commands in current directory
```
### opam remove bar --force
FILE(config)                    Read ${BASEDIR}/OPAM/config in 0.000s
[NOTE] Forcing removal of (uninstalled) bar.1
[bar: pwd]
+ ${BASEDIR}/OPAM/opam-init/hooks/sandbox.sh "remove" "pwd"
- ${BASEDIR}
 -> removed   bar.1
```

rjbou:
fix #6570


src/client/opamClient.ml
```diff
@@ -2412,30 +2412,40 @@ let remove_t ?ask ~autoremove ~force ?(formula=OpamFormula.Empty) atoms t =
...
         with Not_found ->
           OpamConsole.error "No package %s found for (forced) removal.\n"
-            (OpamFormula.short_string_of_atom atom)
+            (OpamFormula.short_string_of_atom atom);
+          true
       in
-      List.iter force_remove not_installed
+      List.fold_left (fun nothing_to_do atom ->
+          nothing_to_do && force_remove atom)
+        true not_installed
```
kit-ty-kate, line 2443: This changes the behaviour from "run force_remove for all elements of not_installed" to "[...] until it returns false". Is that expected?

rjbou: absolutely not!
```suggestion
      List.fold_left (fun nothing_to_do atom ->
          force_remove atom && nothing_to_do)
        true not_installed
```

kit-ty-kate: I believe this can be simplified into:
```suggestion
      List.for_all force_remove not_installed
```

rjbou: Wouldn't `for_all` have the same issue than the first implem ?

src/client/opamClient.ml
```diff
@@ -2412,30 +2412,38 @@ let remove_t ?ask ~autoremove ~force ?(formula=OpamFormula.Empty) atoms t =
...
 
-  let nothing_to_do = ref true in
-  let packages, not_installed =
-    get_installed_atoms t atoms in
-  if not_installed <> [] then (
+  let packages, not_installed = get_installed_atoms t atoms in
+
+  let nothing_to_do =
+    if not_installed = [] then true else
     if force then
```
kit-ty-kate:
```suggestion
    if not_installed = [] then
      true
    else if force then
```

src/client/opamClient.ml
kit-ty-kate: oops yes, my bad. Please use your suggestion instead

src/client/opamClient.ml
```diff
@@ -2412,30 +2412,38 @@ let remove_t ?ask ~autoremove ~force ?(formula=OpamFormula.Empty) atoms t =
...
+          OpamPackage.Set.filter (OpamFormula.check atom) t.packages
+        in
         try
           let nv = OpamPackage.max_version candidates (fst atom) in
-          OpamConsole.note "Forcing removal of (uninstalled) %s" (OpamPackage.to_string nv);
+          OpamConsole.note "Forcing removal of (uninstalled) %s"
+            (OpamPackage.to_string nv);
+          let d = OpamPath.Switch.remove t.switch_global.root t.switch nv in
+          OpamFilename.rmdir d;
+          OpamFilename.mkdir d;
```
kit-ty-kate, line 2432:
```suggestion
          OpamFilename.cleandir d;
```

kit-ty-kate, line 2432: I think it would be nice to have a dedicated function in `OpamFilename` that does that. It doesn't have to be optimized (only remove the directory content but not the directory if it exists already), but having it there simplifies things for whenever we will optimize it.

We have `cleandir` already, maybe `cleandir_create` or something like that

kit-ty-kate: Actually ignore what i just said, i think i would prefer to have this logic deeper in the code anyway. I don't think this should be done in `OpamClient` (see https://github.com/ocaml/opam/pull/6672/commits/dae8e6b05e9eef680d9dce32dea52db2828203a7). This way we avoid useless checks that were already done and we ensure the same logic applies everywhere the `OpamAction` function is used

tests/reftests/dot-install.test
```diff
@@ -321,6 +321,8 @@ The following actions will be performed:
   - remove no-specified-dir 1
 
 <><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+SYSTEM                          mkdir ${BASEDIR}/OPAM/rem-dir/.opam-switch/remove
+SYSTEM                          mkdir ${BASEDIR}/OPAM/rem-dir/.opam-switch/remove/no-specified-dir.1
```
kit-ty-kate, line 325: Opened https://github.com/ocaml/opam/issues/6737 to fix later

PR #6673, opened 2025-09-04
loggign: fix log file management for tests

rjbou:


kit-ty-kate (commented):
Negative values support for OPAMDEBUG was added in d59bf8538f70375d25540c38fc86af3b10db0388. This behaviour isn't really documented (mostly used for ourselves anyway) and i personally didn't know about it.

I feel like, instead of using `Stdlib.abs` everywhere but the piece of code that needs it in `OpamConsole`, we should probably make it so that `OpamCoreConfig.debug_level` is always >= 0 and have a separate field to encode the sign. Something called `OpamCoreConfig.repeatable_debug : bool` or something to make it clearer.

Otherwise the rest of the changes are obviously correct.

PR #6679, opened 2025-09-09
OpamFile.OPAM API: No longer assume that internal opam repositories are directories

kit-ty-kate:
Extracted from #6625 
Queued on
* [x] #6659
* [x] #6696

To allow to change the type of repository from a directory to something else, we need some of the internal functions that expects to be dealing with a directory to be rearranged a little.
In this PR, we do so for `OpamFile.OPAM.get_extra_files` which currently takes a function returning a directory (repository root), to return instead the list of files and content that users of the function actually expect.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -29,6 +29,7 @@ users)
 
 ## Install
   * More fine grained error message in case of bad hash or missing extra-files error (and remove raw fatal error) [#6696 @rjbou]
+  * Do not ignore extra-files whose name is invalid and fail early in that case [#6679 @rjbou @kit-ty-kate]
 
 ## Build (package)
 
@@ -38,6 +39,7 @@ users)
   * Show the invalid character when detecting an erroneous package name [#6638 @lefessan - fix #6396]
   * Handle non-displayable characters when detecting an erroneous package name or version [#6640 @kit-ty-kate]
   * Remove duplicated directory separator when displaying some rare filenames [#6703 @rjbou]
+  * The name of missing or erroneous extra-files are not displayed instead of their path in error messages [#6679 @kit-ty-kate @rjbou]
 
 ## Switch
 
@@ -64,6 +66,7 @@ users)
 ## Source
   * Better error message, especially in case of `Failure` [#6696 @rjbou]
   * Raise a warning instead of an error when an item of `extra-files` is missing [#6696 @rjbou]
+  * Do not ignore extra-files whose name is invalid and raise a warning in that case [#6679 @rjbou @kit-ty-kate]
 
 ## Lint
 
@@ -208,11 +211,15 @@ users)
 
 ## opam-state
   * `OpamSwitchState.files`: was removed [#6662 @kit-ty-kate]
+  * `OpamSwitchState.overlay_opam_file`: was added [6679 @rjbou]
   * `OpamRepositoryState` add `load_opams_from_diff` to update package definitions based on file change operations (diff) [#6614 @arozovyk]
+  * `OpamRepositoryState.get_repo_files`: was added [#6679 @kit-ty-kate @rjbou]
 
 ## opam-solver
 
 ## opam-format
+  * `OpamFile.OPAM.get_metadata_dir`: was removed [#6679 @kit-ty-kate]
+  * `OpamFile.OPAM.get_extra_files`: no longer takes a named `repos_roots` argument and instead takes a named `get_repo_files` argument. It also now returns the content of the files instead of their path [#6679 @kit-ty-kate @rjbou]
   * `OpamFormula.equal_relop`: was added [#6644 @kit-ty-kate]
   * `OpamTypesBase.{action,pkg_flag,simple_arg,arg,filter,command}_equal`: were added [#6644 @kit-ty-kate]
   * `OpamVariable.variable_contents_equal`: was added [#6644 @kit-ty-kate]
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -3963,32 +3963,33 @@ let lint cli =
         (OpamGlobalState.with_ `Lock_none @@ fun gt ->
          OpamSwitchState.with_ `Lock_none gt @@ fun st ->
          try
-           let nv = match pkg with
-             | name, Some v -> OpamPackage.create name v
-             | name, None -> OpamSwitchState.get_package st name
+           let name, nv = match pkg with
+             | name, Some v -> name, OpamPackage.create name v
+             | name, None -> name, OpamSwitchState.get_package st name
            in
-           let opam = OpamSwitchState.opam st nv in
-           match OpamPinned.orig_opam_file st (OpamPackage.name nv) opam with
-           | None -> raise Not_found
-           | Some file ->
-             let label =
-               match OpamFile.OPAM.metadata_dir opam with
-               | None -> None
-               | Some (None, abs) ->
-                 let filename =
-                   if OpamFilename.starts_with
-                       (OpamPath.Switch.Overlay.dir gt.root st.switch)
-                       (OpamFilename.of_string abs) then
-                     Printf.sprintf "<pinned>/%s" (OpamPackage.to_string nv)
-                   else abs
+           if OpamSwitchState.is_pinned st name then
+             match OpamSwitchState.overlay_opam_file st name with
+             | Some file ->
+               let label =
+                 Printf.sprintf "<pinned>/%s" (OpamPackage.to_string nv)
+               in
+               [`pkg (file, label)]
+             | None -> raise Not_found
+           else
+             let opam = OpamSwitchState.opam st nv in
+             match OpamPinned.orig_opam_file st (OpamPackage.name nv) opam with
+             | None -> raise Not_found
+             | Some file ->
+               let label =
+                 let reponame =
+                   match OpamFile.OPAM.metadata_dir opam with
+                   | None | Some (None, _) -> "repo"
+                   | Some (Some repo, _) -> OpamRepositoryName.to_string repo
                  in
-                 Some filename
-               | Some (Some repo, _rel) ->
-                 Some (Printf.sprintf "<%s>/%s"
-                         (OpamRepositoryName.to_string repo)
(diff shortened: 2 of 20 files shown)


tests/reftests/dot-install.test
```diff
@@ -210,7 +210,6 @@ Not found: ${BASEDIR}/OPAM/inst/.opam-switch/install/dot.changes
 === install 1 package
   - install lot-of-files ~dev
 SYSTEM                          mkdir ${BASEDIR}/OPAM/inst/.opam-switch/build/lot-of-files.~dev
-SYSTEM                          copy ${BASEDIR}/OPAM/repo/default/packages/lot-of-files/lot-of-files.~dev/files/lot-of-files.install.in -> ${BASEDIR}/OPAM/inst/.opam-switch/build/lot-of-files.~dev/lot-of-files.install.in
```
rjbou, line 213: To investigate...

tests/reftests/extrafile.test
```diff
@@ -344,25 +354,27 @@ Bad hash for   - ${BASEDIR}/OPAM/repo/default/packages/good-md5-bad-sha256/good-
...
 
 ### test -f good-md5-bad-sha256.1/p.patch
 ### opam clean --download-cache
 Clearing cache of downloaded files
 ### :I:5: no checksum
 ### opam lint --package no-checksum
-<default>/no-checksum.1: Errors.
-             error  3: File format error in 'extra-files' at line 11, column 2: expected [file checksum]
-# Return code 1 #
+<default>/no-checksum.1: Passed.
```
rjbou: This change is due to the fact that now we take the opam file from repo state instead of the plain file from opam internal repo copy. Stored opam file in repo state removes the mismatching extra files, so there is no longer the error (extra-files field removed).

rjbou: We may change `--package` behaviour to be used only with pinned package...

src/state/opamRepositoryState.ml
```diff
@@ -153,6 +153,14 @@ let get_root rt name =
 let get_repo_root rt repo =
   get_root_raw rt.repos_global.root rt.repos_tmp repo.repo_name
 
+let get_repo_files rt name dir =
```
```suggestion
(* We need to keep the argument as string to be able to have the 
   non resolved path to concatenate. *)
let get_repo_files rt name dir =
```

src/state/opamPinned.mli
```diff
@@ -57,9 +57,3 @@ val files_in_source_w_target:
     [<name>.opam/opam]). This function doesn't check the project directory name
     itself, or the package name that might be specified within the file. *)
 val name_of_opam_filename: ?locked:string -> dirname -> filename -> name option
-
-(** Finds back the location of the opam file this package definition was loaded
-    from *)
-val orig_opam_file:
```
rjbou, line 63: renamed and moved to switch state to fit its new specification

src/state/opamRepositoryState.ml
rjbou: or use a `basename`

src/core/opamCompat.ml
```diff
@@ -101,6 +101,11 @@ module Lazy = struct
   let map f x =
     lazy (f (Lazy.force x))
 
+  let map_val f x =
```
kit-ty-kate, line 105:
```suggestion
  (** NOTE: OCaml >= 4.13 *)
  let map_val f x =
```

master_changes.md
```diff
@@ -214,3 +214,4 @@ users)
   * `OpamStd.String.map`: was removed. Use `Stdlib.String.map` instead. [#6442 @kit-ty-kate]
   * `OpamStd.String.{starts_with,ends_with,for_all,fold_left}`: were moved to `OpamCompat.String` [#6442 @kit-ty-kate]
   * `OpamFilename.create`: deduplicate the directory separator character when the basename starts with one [#6703 @rjbou]
+  * `OpamCompat.Lazy`: add `map_val` [#6679 @rjbou}
```
```suggestion
  * `OpamCompat.Lazy`: add `map_val` [#6679 @rjbou]
```

src/format/opamFile.ml
```diff
@@ -3553,23 +3553,30 @@ module OPAM = struct
...
-        Some OpamFilename.Op.(repos_roots r / rel)
-
-  let get_extra_files ~repos_roots o =
-    OpamStd.Option.Op.(
-      (get_metadata_dir ~repos_roots o >>= fun mdir ->
-       let files_dir = OpamFilename.Op.(mdir / "files") in
+  let get_extra_files ~get_repo_files o =
+    let open OpamFilename.Op in
+    let open OpamStd.Option.Op in
+    (match metadata_dir o with
```
kit-ty-kate, line 3559:
```suggestion
    match metadata_dir o with
```

src/client/opamSwitchCommand.ml
```diff
@@ -627,18 +627,19 @@ let export rt ?(freeze=false) ?(full=false)
...
-                     OpamHash.check_file (OpamFilename.to_string file) hash then
-                    let value = Base64.encode_string ~pad:false (OpamFilename.read file) in
+              List.fold_left (fun (hmap,err) (base, content, hash) ->
+                  match content with
+                  | Some (lazy content) when
+                      OpamHash.check_string content hash ->
+                    let value = Base64.encode_string ~pad:false content in
                     OpamHash.Map.add hash value hmap, err
-                  else hmap, base::err)
+                  | _ -> hmap, base::err)
```
kit-ty-kate:
```suggestion
                  | None | Some _ -> hmap, base::err)
```

src/state/opamSwitchAction.ml
```diff
@@ -146,17 +146,17 @@ let install_metadata st nv =
...
+  List.iter (fun (rel_path, content, _hash) ->
       let dst =
         OpamFilename.create
           (OpamPath.Switch.installed_opam_files_dir
              st.switch_global.root st.switch nv)
           rel_path
       in
       OpamFilename.mkdir (OpamFilename.dirname dst);
-      OpamFilename.copy ~src:f ~dst)
+      OpamFilename.write dst (Lazy.force (Option.get content)))
```
kit-ty-kate: The option should be matched on instead of using `Option.get` and show a proper error. Using `install_metadata` doesn't imply the files exist (if only with `opam reinstall --forget-pending`)

src/state/opamUpdate.ml
```diff
@@ -348,16 +348,15 @@ let pinned_package st ?version ?(autolock=false) ?(working_dir=false) name =
...
-          if OpamFilename.exists file &&
-             OpamHash.check_file (OpamFilename.to_string file) hash then
-            OpamFilename.copy ~src:file
-              ~dst:(OpamFilename.create files_dir rel_file)
-          else
+      List.iter (fun (rel_file, content, hash) ->
+          match content with
+          | Some (lazy content) when OpamHash.check_string content hash ->
+            OpamFilename.write (OpamFilename.create files_dir rel_file) content
+          | _ ->
```
kit-ty-kate:
```suggestion
          | None | Some _ ->
```

src/client/opamCommands.ml
```diff
@@ -3963,30 +3963,42 @@ let lint cli =
...
-             | name, Some v -> OpamPackage.create name v
-             | name, None -> OpamSwitchState.get_package st name
+           let name, nv = match pkg with
+             | name, Some v -> name, OpamPackage.create name v
+             | name, None -> name, OpamSwitchState.get_package st name
            in
-           let opam = OpamSwitchState.opam st nv in
-           match OpamPinned.orig_opam_file st (OpamPackage.name nv) opam with
-           | None -> raise Not_found
+           match OpamSwitchState.overlay_opam_file st name with
```
kit-ty-kate: Something feels a bit off with the use of `overlay_opam_file` vs. the previous `orig_opam_file`, in here and with its use via `?format_from`. I think we should probably avoid breaking compatibility on this and instead do something like copying the file to a temporary directory, and simply remove the ability for `orig_opam_file` to fetch things from the repository and only fallback to overlay if the original file isn't accessible. It was in `OpamPinned` after all

tests/reftests/dot-install.test
kit-ty-kate: It's simply that `OpamSystem.write` has no `log`

src/client/opamCommands.ml
rjbou: This will keep backward compat only for already pinned packages. Pinned package from repo won't have the format kept as now.

tests/reftests/extrafile.test
rjbou: From dev meeting: we decided to put back the mechanism to retrieve the file from repo root. At first, we wanted to avoid it because extracting a file from a tarred repository is a consuming operation.
For the usages that are done in this PR (`opam lint --package` and `opam pin` with version and / or edit), it is conceivable to have this consuming operation as they are not operation that are done often, and it permit to keep the same behaviour (respectively linting without preprocessing and preserve format of opam files).

tests/reftests/extrafile.test
```diff
@@ -658,7 +660,11 @@ The following actions will be performed:
 - No changes have been performed
 # Return code 31 #
 ### opam source escape-good-md5
-Successfully extracted to ${BASEDIR}/escape-good-md5.1
+[WARNING] Some errors extracting to ${BASEDIR}/escape-good-md5.1:
```
rjbou, line 663: todo

src/state/opamPinned.mli
rjbou: kept, cf https://github.com/ocaml/opam/pull/6679#discussion_r2414644674

tests/reftests/dot-install.test
rjbou: maybe todo: have a look why action-disk test is not affected

tests/reftests/extrafile.test
rjbou: Ok, in fact it resolves an error that we had :+1:

tests/reftests/dot-install.test
rjbou: Added logging for read & write. To extract in its own PR

PR #6680, opened 2025-09-09
[API] Abstract over repository roots to simplify the addition of new repository formats

kit-ty-kate:
Extracted from #6625 
~Queued on #6679~
To be merged at the same time as https://github.com/ocaml-opam/opam-rt/pull/86

As mentioned in #6625, adding abstract types over repository roots allows us to simplify and ensure correctness when adding a new format of repository.

~In its current form, this PR also changes a couple of unrelated thing that i failed to change back when extracting from #6625 (removal of some use and logic associated with `OpamRepositoryConfig.repo_tarring`) so it needs at least a cleanup before it is in a mergable state (on top of being queued on the already substantial #6679)~


src/client/opamCommands.ml
```diff
@@ -2512,16 +2512,14 @@ let repository cli =
...
                 to retrieve a consistent state."
                (OpamRepositoryName.to_string name);
-           let target =
-             OpamFilename.(Op.(tmp_dir / Base.to_string (basename_dir dir)))
-           in
-           OpamFilename.copy_dir ~src:dir ~dst:target;
-           fun () -> OpamFilename.copy_dir ~src:target ~dst:dir)
+           let target = OpamRepositoryRoot.Dir.backup ~tmp_dir dir in
+           OpamRepositoryRoot.Dir.copy ~src:dir ~dst:target;
+           fun () -> OpamRepositoryRoot.Dir.copy ~src:target ~dst:dir)
```
rjbou: Why use `Dir` here instead of the abstraction `OpamRepositoryRoot.t` ?

src/state/opamUpdate.ml
```diff
@@ -18,11 +18,14 @@ open OpamFilename.Op
 let log fmt = OpamConsole.log "UPDATE" fmt
 let slog = OpamConsole.slog
 
+let safe_read_repo_file = function
+  | OpamRepositoryRoot.Dir dir ->
+    OpamFile.Repo.safe_read (OpamRepositoryPath.repo dir)
```
rjbou: better have this function is OpamRepositoryState (cf related comment in 6625)

src/state/opamUpdate.ml
```diff
@@ -123,41 +126,30 @@ let repository rt repo =
...
+    let opams =
+      match repo_root with
+      | OpamRepositoryRoot.Dir dir ->
+        OpamRepositoryState.load_opams_from_dir repo.repo_name dir
+    in
+    begin match repo_root with
+    | OpamRepositoryRoot.Dir _ ->
+      if OpamFilename.exists tarred_repo then
+        OpamFilename.remove tarred_repo;
+    end;
```
rjbou: cf related comment in 6625

src/client/opamCommands.ml
kit-ty-kate: This code is guarded under the `else` branch of `if OpamFilename.exists tar` so in this case we know that we're working with directories

src/state/opamUpdate.ml
kit-ty-kate: This diff doesn't exist in this PR anymore

src/state/opamUpdate.ml
kit-ty-kate: I used the already existing `OpamRepositoryRoot.delayed_read_repo` function instead

src/client/opamCommands.ml
rjbou: Yes, i saw. What i wanted to highlight is that (almost) the same code is used for tar & dir repositories : copy in tmp dir the do the copy and return the function that does the copy back. I was wondering if it is possible to factorise it using the abstraction.
```ocaml
  if OpamFilename.exists tar then
    (let target = OpamFilename.create tmp_dir (OpamFilename.basename tar) in
     OpamFilename.copy ~src:tar ~dst:target;
     fun () -> OpamFilename.copy ~src:target ~dst:tar)
  else
     ( [...]
      let target = OpamRepositoryRoot.Dir.backup ~tmp_dir dir in
      OpamRepositoryRoot.Dir.copy ~src:dir ~dst:target;
      fun () -> OpamRepositoryRoot.Dir.copy ~src:target ~dst:dir)
```

src/state/opamUpdate.ml
```diff
@@ -75,7 +74,8 @@ let repository rt repo =
     else match eval_redirect gt r repo_root with
       | None -> Done (r, has_changes)
       | Some (new_url, f) ->
-        OpamFilename.cleandir repo_root;
+        OpamRepositoryRoot.remove repo_root;
+        OpamRepositoryRoot.make_empty repo_root;
```
rjbou, line 78: Do we need to create an empty directory here ? `OpamHTTP` and `OpamVCS` `fetch_repo_update` does not need the directory to exist, but `OpamLocal` requires it. I think it should be in there that the change should be applied. Plus, it'll make the functions more homogeneous.

rjbou: ~And like that, there is only the need to have `OpamRepositoryRoot.Dir.make_empty` and no `OpamRepositoryRoot.make_empty`~ Seems to be used in 6625 - `OpamHTTP.fetch_repo_update`

src/repository/opamRepositoryRoot.ml
```diff
@@ -0,0 +1,92 @@
...
+    let open OpamFilename.Op in
+    tmp_dir / OpamFilename.Base.to_string (OpamFilename.basename_dir repo_root)
+
+  let cwd = OpamFilename.cwd
+  let in_dir = OpamFilename.in_dir
+  let exists = OpamFilename.exists_dir
+  let remove = OpamFilename.rmdir
+  let move = OpamFilename.move_dir
+  let copy = OpamFilename.copy_dir
+  let copy_except_vcs = OpamFilename.copy_dir_except_vcs
```
rjbou, line 32: Doesn't worth to have only 1 function ? 
```suggestion
  let copy ?(include_vcs=true) =
    if include_vcs then OpamFilename.copy_dir
    else OpamFilename.copy_dir_except_vcs
```

src/repository/opamRepositoryRoot.mli
```diff
@@ -0,0 +1,79 @@
...
+(*  GNU Lesser General Public License version 2.1, with the special       *)
+(*  exception on linking described in the file LICENSE.                   *)
+(*                                                                        *)
+(**************************************************************************)
+
+(** This module abstract the notion of repository root over its concrete
+    implementation (could be a database, a file, a directory, etc.) *)
+
+(** Repository root implemented as a directory *)
+module Dir : sig
```
rjbou, line 18: Some functions are only present in the `Dir` module, and not meant to be abstracted via `OpamRepositoryRoot` module. In that case, they will be used only in the match case `OpamRepositoryRoot.Dir dir`, does it worth to work directly with `OpamFilename` functions for those part instead of having the redirections?
A good example of that is the change in `OpamLocal` lines 160-171 (new file)

src/repository/opamRepositoryRoot.mli
```diff
@@ -0,0 +1,79 @@
...
+
+(** [quarantine repo_root] returns a temporary repository root dedicated
+    to [repo_root]. the returned repository is not created on disk and
+    points to a statically known repository located in the same parent
+    directory as [repo_root]. *)
+val quarantine : t -> t
+
+val remove : t -> unit
+val is_empty : t -> bool option
+val make_empty : t -> unit
```
rjbou, line 74: If not removed
```suggestion
(* Creates the root of the repository *)
val make_empty : t -> unit
```

src/client/opamArg.ml
```diff
@@ -221,16 +221,16 @@ let environment_variables =
        'download-command' value from the main config file.";
       "NOCHECKSUMS", cli_original, (fun v -> NOCHECKSUMS (env_bool v)),
       "enables option --no-checksums when available.";
-      "REQUIRECHECKSUMS", cli_original,
-      (fun v -> REQUIRECHECKSUMS (env_bool v)),
-      "Enables option `--require-checksums' when available \
-       (e.g. for `opam install').";
-      "REPOSITORYTARRING", cli_between cli2_2 cli2_5,
+      "REPOSITORYTARRING", cli_from cli2_2,
```
rjbou, line 225: Is the removal of repository tarring moved to 6625 ?

rjbou:
I've update the PR. The last commit is here as a proposal, introduce in the new `OpamRepositoryRoot` the path construction, like that `OpamRepositoryRoot.x.Path` module gives directly the well constructed path (e.g. absolute for dirs, partial for tar). The purpose is to not have to handle when we call the finction to construct a path if in that case we need an absolute path or a partial one (ftm, there is only theses 2 cases).
This is not strictly needed by #6625, but imo it helps writing the code.

src/client/opamAdminRepoUpgrade.ml
```diff
@@ -502,11 +502,13 @@ let do_upgrade_mirror repo_root base_url =
...
   OpamFile.Repo.write
-    (OpamFile.make OpamFilename.Op.(tmp_mirror_dir // "repo"))
-    repo_20;
-  let dir20 = OpamFilename.Dir.of_string upgradeto_version_string in
-  OpamFilename.rmdir dir20;
-  OpamFilename.move_dir ~src:tmp_mirror_dir ~dst:dir20;
+    (OpamRepositoryRoot.Dir.Path.repo tmp_mirror_dir) repo_20;
+  let dir20 =
+    OpamRepositoryRoot.Dir.of_dir
+      OpamRepositoryRoot.Dir.Op.(repo_root / upgradeto_version_string)
```
kit-ty-kate: this change seems like a separate fix rather than a change related to OpamRepositoryRoot

rjbou: true, and it is erroneous, the good fix would be : 

```suggestion
      OpamRepositoryRoot.Dir.Op.(OpamRepositoryRoot.Dir.basename repo_root / upgradeto_version_string)
```

Should we extract it in another PR or have it in a separate commit in this PR ?

rjbou: oh, i just saw that it was already in a separate commit.

kit-ty-kate: erroneous? I don't think so, the code as it was there was fine. `basename repo_root` would result in a totally different behaviour and bring a potential bug

rjbou (approved):

Issue #6111, opened 2024-07-22
Rationale behind upgrading all dependent packages when pinning a package to the already installed version


MSoegtropIMC:
In Coq Platform CI (which is heavily opam based) I see a change in behaviour after updating to opam 2.2: when I pin a package to the version it already has, opam upgrades all packages depending on this package. See e.g. this CI log:

https://github.com/coq/platform/actions/runs/10037269557/job/27736704911#step:5:721

Before Coq 8.15.2 has been installed already - at the point above it is pinned to 8.15.2 which has the effect you see below this point in the log.

This was not so in opam 2.1.X. I can live with this - I now added `--no-action` to the pin commands (CI is still running but I guess this has the desired effect), but I wonder if this change is intended and if so what the rationale behind it is.

A note: I install many packages in parallel with a single request, both for performance reasons and also to catch version issues early. Some critical packages (like Coq) I need to pin and I do this after the installation with individual pin commands. Is there a better way of doing this?

kit-ty-kate:
ah, you'd want something like `opam install <pkg1> pin:<pkg2> ... <pkgn>` where `pkg2` would be pinned but not the other ones?
(actually looking at the scripts in coq/platform it seems to be exactly what you already do internally)

From a UX perspective i'm not sure i like this, however we could think of supporting `opam pin add <pkg1> <pkg2> ... <pkgn>` similarily to https://github.com/ocaml/opam/pull/5537 but we should think about it a bit more as the syntax for `pin add` is already pretty bloated with regard to the number of options it can take. Maybe we could have a separate subcommand for it. We can think about it, however one thing is that performance-wise `opam pin add --no-action` should be negligible already so i'm not sure whether allowing to pin multiples at a time would be an improvement for you.

kit-ty-kate:
> I was not aware of the `opam install pin:<pkg>` syntax

no no, what i meant is "we could do it but this feature does not exist currently"

kit-ty-kate:
> What is wrong with this besides that it is not so obvious?

this + mixing the pin mechanic with the install command is currently only done for local projects and i feel like allowing it more generally would add a confusing feature in terms of design and would tie more knots between the two mechanics in our codebase internally. Pinning is a tricky feature and can get arbitrarily complicated (e.g. https://github.com/ocaml/opam/issues/3202) and adding such a syntactic sugar feels to be like a trap in terms of feature creep.

PR #6691, opened 2025-09-16
Add a test showing the behaviour of version pins when one of the dependencies isn't up-to-date

kit-ty-kate:
See https://github.com/ocaml/opam/issues/6111

### DIFF
--- a/tests/reftests/pin.test
+++ b/tests/reftests/pin.test
@@ -1623,3 +1623,40 @@ Done.
 file1
 ### opam-cat OPAM/flag-edit/.opam-switch/overlay/tide/opam | grep install
 install: [["mkdir" "%{lib}%/%{name}%"] ["cp" "file1" "%{lib}%/%{name}%/file1"]]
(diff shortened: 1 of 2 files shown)


tests/reftests/pin.test
```diff
@@ -1623,3 +1623,35 @@ Done.
...
 ### opam-cat OPAM/flag-edit/.opam-switch/overlay/tide/opam | grep install
 install: [["mkdir" "%{lib}%/%{name}%"] ["cp" "file1" "%{lib}%/%{name}%/file1"]]
+### :C:h: opam pin doesn't upgrade outdated dependencies
+### opam switch create no-upgrade --empty
+### <pkg:dep.1>
+opam-version: "2.0"
+### <pkg:dep.2>
+opam-version: "2.0"
+### <pkg:pkg.1>
+opam-version: "2.0"
```
rjbou, line 1633: From description, i have the feeling that there is a missing `depends: "dep"`, no ?

kit-ty-kate: oops, indeed

PR #6695, opened 2025-09-17
Clean all created temporary files at exit

rjbou:


src/core/opamFilename.mli
```diff
@@ -105,7 +105,8 @@ val with_tmp_dir: (Dir.t -> 'a) -> 'a
 (** Provide an automatically cleaned up temp directory to a job *)
 val with_tmp_dir_job: (Dir.t -> 'a OpamProcess.job) -> 'a OpamProcess.job
 
-(** Raw function to create a temporary directory. No automatic cleanup *)
+(** Raw function to create a temporary directory.
```
kit-ty-kate, line 108: These mk_* functions are misleading, they do not create anything but merely return a unique name. I think we should change this first

src/core/opamFilename.ml
```diff
@@ -261,6 +261,9 @@ let exists filename =
 let opt_file filename =
   if exists filename then Some filename else None
 
+let mk_tmp_file () =
+  of_string @@ OpamSystem.mk_temp_dir ()
```
kit-ty-kate: creating new files by ""creating a directory"" feels a bit wrong

src/core/opamFilename.mli
rjbou: good point. Especially with https://github.com/ocaml/opam/pull/6659/files#diff-38a34666fa25190dcc42466d8decbd8c6d5bb62952d05b09f0d233c37b3bd4d7R320 we have a case where we don't want to clean the temporary file, we should have both functions.

src/core/opamFilename.ml
rjbou: It is because it is non well named, as you pointed on your previous message.

src/core/opamFilename.mli
rjbou: As discussed, there is indeed no need to have the non cleaned temporary files for `opamfile` variable as we stay in opam.

rjbou: > These mk_* functions are misleading, they do not create anything but merely return a unique name. I think we should change this first

I've opened #6704  for that

src/core/opamSystem.ml
```diff
@@ -89,12 +89,29 @@ let temp_name ?dir ?(prefix="opam") () =
...
   tmpdir / (temp_basename prefix)
 
+let temp_dir_cleaner =
+  let to_clean = ref OpamStd.String.Set.empty in
+  OpamStd.Sys.at_exit
+    (fun () ->
+       OpamStd.String.Set.iter (fun d ->
+           try
+             if Sys.file_exists d then
+               Unix.rmdir d;
```
kit-ty-kate, line 99: The check is redundant with the `try ... with Unix_error _` block
```suggestion
             Unix.rmdir d;
```

PR #6696, opened 2025-09-17
Pretty printing of some error for opam source & build actions

rjbou:
queued on
* #6703


rjbou:
The error came from `OpamFilename.create`. The fix is on #6703

PR #6701, opened 2025-09-18
Simplify the src_ext/update-sources.sh script

kit-ty-kate:
Required during #6700 to update menhir as the URL couldn't be parsed.
While trying to fix that i found the script too complex so i chose to simplify it instead.


dra27:
Go for it - it's only intended as a utility, and if the extra error checking isn't being helpful, better to remove it!

PR #6703, opened 2025-09-19
API: remove duplicated directory separator when creating a filename

rjbou:


src/core/opamFilename.ml
```diff
@@ -159,6 +159,7 @@ let create dirname basename =
   if basename = b2 then
     { dirname; basename }
   else
+    let b1 = OpamStd.String.remove_prefix ~prefix:Filename.dir_sep b1 in
```
kit-ty-kate: This subtly changes the behaviour of
```
OpamFilename.create (OpamFilename.raw_dir "") (OpamFilename.Base.of_string "/etc")
```
so i'd rather keep the change of behaviour to a minimum and leave bigger changes for a future code cleanup of this module
```suggestion
    let b1 =
      if dirname = ""
      then b1
      else OpamStd.String.remove_prefix ~prefix:Filename.dir_sep b1
    in
```

kit-ty-kate:
The error on i686-w64-mingw32/4.14 is a spurious failure. Ignoring.

PR #6705, opened 2025-09-25
Static linking for OpenBSD

### LINKED ISSUE #6241: Release binaries: Statically linked OpenBSD binaries
Following https://github.com/ocaml/opam/issues/6232 and its fix in https://github.com/ocaml/opam/pull/6237, i would like to look into whether it's possible and reasonable to statically link the OpenBSD libc and libstdc++ in our release binaries for OpenBSD to allow one binary to support both the latest OpenBSD release and the previous one (OpenBSD has a 6 months release cycle and supports each release for a year)

flumf:
Building with `./configure --enable-static --with-vendored-deps` on an OpenBSD 7.7 system results in a statically linked `opam` binary. Brief testing of the binary shows no problems, although I've admittedly not strenuously tested it, nor have I tried building with non-vendored dependencies.

Resolves #6241.

### DIFF
--- a/configure
+++ b/configure
@@ -1379,9 +1379,9 @@ Optional Features:
   --enable-developer-mode Enable developer features
 
   --enable-static         Compile the opam binary statically. Only Musl-based
-                          Linux distributions and Windows with MinGW is
-                          currently supported. This is the default on Windows
-                          with MinGW
+                          Linux distributions, Windows with MinGW, and OpenBSD
+                          are currently supported. This is the default on
+                          Windows with MinGW
   --enable-cold-check     Fail on some check necessary for make cold
   --disable-certificate-check
                           Do not check the certificate of opam's dependency
@@ -5874,6 +5874,11 @@ fi
 then :
   platform_dependent_stuff="${platform_dependent_stuff} -cclib -lwindowsapp"
 fi
+   ;; #(
+  *-*-openbsd*) :
+
+    support_static=yes
+    platform_dependent_stuff="-cclib -lc++ -cclib -lc++abi -cclib -static"
    ;; #(
   *) :
      ;;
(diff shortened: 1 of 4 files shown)


kit-ty-kate:
I can confirm, the release script now outputs a static OpenBSD binary. Thanks a lot!

kit-ty-kate (approved):

kit-ty-kate:
@semarie i just thought about it just now but none of the architectures supported by OCaml for its native backend (both in 5.x or 4.x) do use gcc, so this is all good on native platforms.

Per [this blog post i found](https://www.cambus.net/the-state-of-toolchains-in-openbsd/), OpenBSD has LLVM by default on:
* OCaml 5 compatible architectures: amd64, arm64, powerpc64, and riscv64
* OCaml 4 (or the various OCaml 5 forks): armv7, i386, macppc
* OCaml 5 (Loongson's fork): loongson (although this was removed from 7.9 anyway)

While compiling opam in bytecode mode is supported it is a miserable experience using it so i wouldn't see anyone purposefully using `--enable-static` on the remaining niche platforms

semarie:
The main architecture using gcc was sparc64 (but we are currently trying to switching it to clang).
The authoritative list of archs with compiler is [ports/infrastructure/mk/arch-defines.mk](https://cvsweb.openbsd.org/annotate/ports/infrastructure/mk/arch-defines.mk). See `CLANG_ARCHS` (arches where the base compiler is clang) and we are building ocaml native on `OCAML_NATIVE_ARCHS` (aarch64, amd64 and i386 currently).

Thanks

kit-ty-kate:
ah good to know thanks, i couldn't find that list anywhere, i didn't think of looking into ports ^^" (i just looked in src)

Issue #6619, opened 2025-07-23
Release: The opam-full-*.tar.gz generation isn't reproducible


kit-ty-kate:
Calling the release script twice, we end up with different checksums because the tarball is generated from a copy of the files to `release/build`, which change their metadata (date of last modified and creation). It would be nice to be able to reproduce them and make sure they tarball hasn't been tempered with easily, without having to extract it and `diff -rua` them.

The binaries though seem to be reproducible, if given the same software at least (tested with `opam-2.4.1-x86_64-linux`)

kit-ty-kate:
https://www.gnu.org/software/tar/manual/html_node/Reproducibility.html

PR #6706, opened 2025-09-26
Make the release archive reproducible

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/6619

While testing #6705 i encountered some failures due to extra `._*` files in the archive (see https://superuser.com/questions/61185/why-do-i-get-files-like-foo-in-my-tarball-on-os-x) which made dune fail.

Using `--no-mac-metadata` fixes this problem but using `git archive` goes even further and should make the archive reproducible (same archive regardless of username, group, uid, gid, current time, …)

### DIFF
--- a/release/Makefile
+++ b/release/Makefile
@@ -30,7 +30,9 @@ $(OUTDIR)/opam-full-$(VERSION).tar.gz:
 	mkdir -p "$(OUTDIR)"
 	git clone $(GIT_URL) -b $(TAG) "$(OUTDIR)/opam-full-$(VERSION)"
 	$(MAKE) -C "$(OUTDIR)/opam-full-$(VERSION)" OCAML=$(call pathsearch,ocaml) download-ext
-	cd "$(OUTDIR)" && tar cz -f $(notdir $@) --exclude .git --no-mac-metadata --no-xattrs opam-full-$(VERSION)
+	git -C "$(OUTDIR)/opam-full-$(VERSION)" add *
+	git -C "$(OUTDIR)/opam-full-$(VERSION)" commit -m "add vendored files"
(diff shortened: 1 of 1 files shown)


release/Makefile
```diff
@@ -30,7 +30,14 @@ $(OUTDIR)/opam-full-$(VERSION).tar.gz:
...
 	git clone $(GIT_URL) -b $(TAG) "$(OUTDIR)/opam-full-$(VERSION)"
 	$(MAKE) -C "$(OUTDIR)/opam-full-$(VERSION)" OCAML=$(call pathsearch,ocaml) download-ext
-	cd "$(OUTDIR)" && tar cz -f $(notdir $@) --exclude .git opam-full-$(VERSION)
+	( set -euo pipefail && \
+	  cd "$(OUTDIR)/opam-full-$(VERSION)" && \
+	  git archive "$(TAG)" \
+		--prefix "opam-full-$(VERSION)/" \
+		--mtime "$$(git show -s --format=%ct "$(TAG)" | tail -1)" \
+		-o "$(abspath $(OUTDIR)/opam-full-$(VERSION).tar)" \
+		$$(for f in $$(git ls-files -i -o -x '*'); do echo "--add-file=$$f" || exit 1; done) && \
```
dra27, line 39: Doing it directly in make has the benefit that the displayed command explicitly shows what was added (I haven’t run it - I’m 99.9% sure the asterisk doesn’t want quoting when used with `$(shell`!)
```suggestion
		$(addprefix --add-file=,$(shell git ls-files -i -o -x *)) && \
```

kit-ty-kate: As discussed during the dev meeting last week, i've tried `$(shell ...)` before but given the change of directory using `cd` above in the same command string, these commands would have to use `git -C` which complicates the whole thing and it didn't work in my case for both `git show` and `git ls-files` for some reason. For now using `sh` subshells work out a bit better. I think we can improve later if needed

PR #6713, opened 2025-09-29
Order build-env after setenv

### LINKED ISSUE #6574: build-env: doesn't seem to be honoured
I have an [opam file](https://github.com/b0-system/b0/blob/c3a62d6038acc4cb6b55c8769e7a61da13d8b9b2/opam) with:

```
build-env: CAML_LD_LIBRARY_PATH += "_build/src/std"
```

My build fails at the `install:` step with:

```
[…]
# Fatal error: cannot load shared library dllb0_stubs
# Reason: dlopen(dllb0_stubs.so, 0x000A): tried: 'dllb0_stubs.so' (no such file), '/System/Volumes/Preboot/Cryptexes/OSdllb0_stubs.so' (no such file), '/usr/lib/dllb0_stubs.so' (no such file, not in dyld cache), 'dllb0_stubs.so' (no such file)
```

If I check the suggested build environment file, the environment variable doesn't seem to be setup as expected by the `build-env` field:

```
> cat ~/.opam/log/b0-51512-5d52f8.env | grep CAML_LD_LIBRARY_PATH
CAML_LD_LIBRARY_PATH=/Users/dbuenzli/.opam/ocaml.5.3.0+bco/lib/stublibs:/Users/dbuenzli/.opam/ocaml.5.3.0+bco/lib/ocaml/stublibs:/Users/dbuenzli/.opam/ocaml.5.3.0+bco/lib/ocaml
```

(Don't know if this is relevant but this package install uses both an `.install` file and an `install:` section).

```
> opam --version
2.3.0
```

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/6574

I'm not too happy with the code but it should work.

The order between build-end and setenv has been that way since at least opam 2.0: https://github.com/ocaml/opam/commit/be714e9b0ccd0ba9af34421e28425c894648459e

### DIFF
--- a/src/client/opamSolution.ml
+++ b/src/client/opamSolution.ml
@@ -1101,7 +1101,8 @@ let confirmation ?ask requested solution =
   OpamConsole.confirm "\nProceed with %s?" (OpamSolver.string_of_stats stats)
 
 let run_hook_job t name ?(local=[]) ?(allow_stdout=false) w =
-  let shell_env = OpamEnv.get_full ~set_opamroot:true ~set_opamswitch:true ~force_path:true t in
+  (* TODO: ????? *)
+  let shell_env = OpamEnv.get_full ~set_opamroot:true ~set_opamswitch:true ~force_path:true ~build_env:[] t in
   let mk_cmd = function
     | cmd :: args ->
       let text = OpamProcess.make_command_text name ~args cmd in
--- a/src/state/opamEnv.ml
+++ b/src/state/opamEnv.ml
@@ -613,7 +613,7 @@ let env_update_resolved_with_default ?comment var =
   let rewrite = Some (SPF_Resolved (Some (default_sep_fmt_str var))) in
   env_update_resolved ?comment ~rewrite var
 
-let compute_updates ?(force_path=false) st =
+let compute_updates ?(force_path=false) ~build_env st =
   (* Todo: put these back into their packages!
   let perl5 = OpamPackage.Name.of_string "perl5" in
   let add_to_perl5lib =  OpamPath.Switch.lib t.root t.switch t.switch_config perl5 in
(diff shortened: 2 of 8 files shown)


src/state/opamEnv.ml
```diff
@@ -688,18 +688,18 @@ let updates_nix st =
      | Some env -> List.map resolve_separator_and_format env)
   | _ -> []
 
-let updates ~set_opamroot ~set_opamswitch ?force_path st =
+let updates ~set_opamroot ~set_opamswitch ?force_path ~build_env st =
   let common =
     updates_common ~set_opamroot ~set_opamswitch st.switch_global.root st.switch
   in
-  common @ compute_updates ?force_path st @ updates_nix st
+  compute_updates ?force_path ~build_env st @ common @ updates_nix st
```
kit-ty-kate, line 695: common and nix updates should probably be moved in `compute_updates` to centralise the ordering in one function. I also don't think nix env should be able to take priority over every others

src/client/opamSolution.ml
```diff
@@ -1101,7 +1101,8 @@ let confirmation ?ask requested solution =
   OpamConsole.confirm "\nProceed with %s?" (OpamSolver.string_of_stats stats)
 
 let run_hook_job t name ?(local=[]) ?(allow_stdout=false) w =
-  let shell_env = OpamEnv.get_full ~set_opamroot:true ~set_opamswitch:true ~force_path:true t in
+  (* TODO: ????? *)
+  let shell_env = OpamEnv.get_full ~set_opamroot:true ~set_opamswitch:true ~force_path:true ~build_env:[] t in
```
kit-ty-kate, line 1105: should pre-command hooks and the like have build-env? It feels like they should

kit-ty-kate:
Moved to the 2.7 milestone as the PR isn't fully ready yet

PR #6715, opened 2025-10-02
Fix the false-positive mismatch debug warning during "opam update" when faced with nested extra-files on Windows

kit-ty-kate:
Discovered in the tests of https://github.com/ocaml/opam/pull/6614

The extra-files loader on Windows breaks when the file is contained in a directory and thinks there is a mismatch between the files listed in the field and the files detected on disk. The faulty behaviour can't be seen in the tests as i'm running them on Unix but it can be seen in the tests results of https://github.com/ocaml/opam/pull/6614 on Windows.

```diff
diff --git a/_build/default/tests/reftests/update.test b/_build/default/tests/reftests/update.out
index 8f52cdb..cd9a0db 100644
--- a/_build/default/tests/reftests/update.test
+++ b/_build/default/tests/reftests/update.out
@@ -415,6 +415,8 @@ EOF
 
 <><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
 [git-test] synchronised from git+file://${BASEDIR}/GITREPO
+opam-file                       Mismatching extra-files at ${BASEDIR}/OPAM/repo/git-test/packages/pkg-a/pkg-a.1: missing from 'files' directory (1) and missing from opam file (1)
+opam-file                       Mismatching extra-files at ${BASEDIR}/OPAM/repo/git-test/packages/prefix-pkg-b/pkg-b.1: missing from 'files' directory (1) and missing from opam file (1)
 Now run 'opam upgrade' to apply any package updates.
 ### opam list --all --repo=git-test -s
 new-pkg
```

### DIFF
--- a/tests/reftests/extrafile.test
+++ b/tests/reftests/extrafile.test
@@ -661,3 +661,49 @@ The following actions will be performed:
 Successfully extracted to ${BASEDIR}/escape-good-md5.1
 ### test -f escape-good-md5.1/p.patch
 # Return code 1 #
+### ::::::::::::::::::::::::
+### : Nested extra-files
+### ::::::::::::::::::::::::
+### <pkg:good-nested-md5.1>
+opam-version: "2.0"
+maintainer: "nobody"
+authors: "nobody neither"
+homepage: "https://no.bo.dy"
+bug-reports: "https://still.nobo.dy"
+dev-repo: "git+https://no.were"
+license: "MIT"
+synopsis: "Initially empty"
+build: [ "test" "-f" "dir/p.patch" ]
+extra-files: [
+  ["dir/p.patch" "md5=patch-md5"]
+]
+### <add-nested-files.sh>
+set -ue
+pkg=$1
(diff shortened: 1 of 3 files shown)


tests/reftests/extrafile.test
```diff
@@ -661,3 +661,40 @@ The following actions will be performed:
...
+]
+### <add-nested-files.sh>
+set -ue
+pkg=$1
+pkgpath="REPO/packages/${pkg%.*}/$pkg"
+mkdir -p $pkgpath/files/dir
+cp p.patch $pkgpath/files/dir/
+### sh add-nested-files.sh good-nested-md5.1
+### sh update-hashes.sh
+### opam update default
```
rjbou: There is no change in the tests in the last commit (as said because the tests was run on windows). It worth adding a comment that at least describes the windows behaviour that is removed in the second commit, to know what is the diff.

kit-ty-kate: I regenerated the test on Windows and the commit now shows the diff

PR #6720, opened 2025-10-05
Add  `opam repo show <reponame>` support

### LINKED ISSUE #6235: opam repo show <repo-name>
The output of `opam repo list` currently is just UI for reading.
However, for example, if you're trying to make sure that a repository of a certain name is attached to the right url, it becomes a bit more complicated than it should.

`opam repo list --short` doesn't really help as it only shows the names of the repository, not its url or rank, and there is no `opam repo show <repo-name>` to actually show what rank and url said repository has.

The proposal would be to add an `opam repo show <repo-name>` command which would look like this:
```
$ opam repo show default
url: https://opam.ocaml.org
rank: 1
```

This way would at least be relatively easy to get information about every repositories as it currently looks like the unpalatable:
```
opam repo list 2> /dev/null | grep "^ [1-9][0-9]* default " | tr -s ' ' | cut -s '-d ' -f4
```
In the future, the output could also be extended with information like `last-modified` (see https://github.com/ocaml/opam/issues/5553)

ProgramingIsTheFuture:
Related to issue https://github.com/ocaml/opam/issues/6235.

Adding a new command that will allow us to get the URL and Rank of a specific repo.

Also supports `all` (will display the url and rank of a repo for every switch):
```sh
$ opam repo show default --all
```

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -2582,10 +2582,14 @@ let repository cli =
            '--all' to see all configured repositories.";
       OpamRepositoryCommand.list rt ~global ~switches ~short;
       `Ok ()
+    | Some `show_repo, [] ->
+      OpamRepositoryState.with_ `Lock_none gt @@ fun rt ->
+      OpamRepositoryCommand.show_repo rt ~switches None;
+      `Ok ()
     | Some `show_repo, [ repo_name ] ->
       OpamRepositoryState.with_ `Lock_none gt @@ fun rt ->
-      OpamRepositoryCommand.show_repo rt ~switches
-        (OpamRepositoryName.of_string repo_name);
+      let repo_name = if scope = [ `All ] then None else Some (OpamRepositoryName.of_string repo_name) in
+      OpamRepositoryCommand.show_repo rt ~switches repo_name;
       `Ok ()
     | command, params -> bad_subcommand ~cli commands ("repository", command, params)
   in
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -2276,6 +2276,8 @@ let repository cli =
      where they are active.";
     cli_original, "priority", `priority, ["NAME"; "RANK"],
     "Synonym to $(b,add NAME --rank RANK)";
+    cli_original, "show", `show_repo, [],
+    "Show the URL and priority of a repository";
   ] in
   let man = [
     `S Manpage.s_description;
@@ -2579,6 +2581,19 @@ let repository cli =
           "These are the repositories in use by the current switch. Use \
            '--all' to see all configured repositories.";
       OpamRepositoryCommand.list rt ~global ~switches ~short;
+      `Ok ()
+    | Some `show_repo, [ repo_name ] ->
+      OpamRepositoryState.with_ `Lock_none gt @@ fun rt ->
+      let switches =
+        if scope = [] ||
+           List.exists (function
+               | `This_switch | `Current_switch | `Switch _ -> true
+               | _ -> false)
+             scope
+        then switches
+        else all_switches
+      in
+      OpamRepositoryCommand.show_repo rt ~switches (OpamRepositoryName.of_string repo_name);
       `Ok ()
     | command, params -> bad_subcommand ~cli commands ("repository", command, params)
   in
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -2276,7 +2276,7 @@ let repository cli =
      where they are active.";
     cli_original, "priority", `priority, ["NAME"; "RANK"],
     "Synonym to $(b,add NAME --rank RANK)";
-    cli_original, "show", `show_repo, [],
+    cli_from cli2_5, "show", `show_repo, [],
     "Show the URL and priority of a repository";
   ] in
   let man = [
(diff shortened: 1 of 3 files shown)


dra27:
(sorry, I can't get past `..` as meaning the parent directory!)

src/client/opamRepositoryCommand.ml
```diff
@@ -219,6 +219,34 @@ let list_all rt ~short =
...
+          try
+            let r = OpamRepositoryName.Map.find repo rt.repositories in
+            let url =
+              OpamConsole.colorise `underline (OpamUrl.to_string r.repo_url)
+            in
+            OpamConsole.msg "switch: %s\n" (OpamSwitch.to_string sw);
+            OpamConsole.msg "url: %s\n" url;
+            OpamConsole.msg "rank: %d\n" rank;
+          with Not_found ->
+            OpamConsole.msg "%s" (OpamConsole.colorise `red "not found"))
```
kit-ty-kate: I can't find a way to hit this line, i'm not sure it is reachable

src/client/opamCommands.ml
```diff
@@ -2580,6 +2582,20 @@ let repository cli =
            '--all' to see all configured repositories.";
       OpamRepositoryCommand.list rt ~global ~switches ~short;
       `Ok ()
+    | Some `show_repo, [ repo_name ] ->
+      OpamRepositoryState.with_ `Lock_none gt @@ fun rt ->
+      let switches =
+        if scope = [] ||
+           List.exists (function
+               | `This_switch | `Current_switch | `Switch _ -> true
+               | _ -> false)
```
kit-ty-kate: Could you list all the variants explicitly? It help for detecting required changes for future maintenance.
I also think that unsupported options should probably display an error message.

src/client/opamCommands.ml
```diff
@@ -2276,6 +2276,8 @@ let repository cli =
      where they are active.";
     cli_original, "priority", `priority, ["NAME"; "RANK"],
     "Synonym to $(b,add NAME --rank RANK)";
+    cli_original, "show", `show_repo, [],
```
```suggestion
    cli_from cli2_5, "show", `show_repo, [],
```

kit-ty-kate (commented):
Thanks for this PR, i've pushed a commit to apply some of our more common coding style, so that the new code doesn't feel too out of place.

I've tried and it works reasonably well, however i've hit some issues:
```
$ opam repo show --switch=does-not-exist default
switch: does-not-exist
url: https://staging.opam.ocaml.org/
rank: 2
```
with `does-not-exist` being a non-existent switch and `default` being an existent repository.
I don't think this should return that kind of output (i would more expect something like an error).

I also think the new command should have a way to display the details of a repository that's not currently attached to any switches (maybe with `--dont-select`?)

I'm also wondering how useful would `opam repo show -a` be, but we could add it later if needed i suppose.

ProgramingIsTheFuture:
There are some other commands with the same behaviour (`abc` is not defined):

```sh
$ opam repo list --switch=abc
[NOTE] These are the repositories in use by the current switch. Use '--all'
       to see all configured repositories.

<><> Repository configuration for switch abc <><><><><><><><><><><><><><><><>
 1 default https://opam.ocaml.org
```

Reading more the code, I found that the actual flag to select switch is `--on-switch` ([defined here](https://github.com/ocaml/opam/blob/ce734b35a18bc98642f7212a280dc4d6c64c969f/src/client/opamCommands.ml#L2319-L2327)). This will produce the expected behaviour (error in case of non-existing switch).

Because `--switch=<switch>` doesn't populate the ``` `Switch <switch>```:
https://github.com/ocaml/opam/blob/ce734b35a18bc98642f7212a280dc4d6c64c969f/src/client/opamCommands.ml#L2375-L2383

Could you please verify?

src/client/opamRepositoryCommand.ml
```diff
@@ -219,6 +219,31 @@ let list_all rt ~short =
...
+          List.mapi (fun i repo -> (i+1, repo)) |>
+          List.filter (fun (_, repo) -> OpamRepositoryName.equal repo repo_name)
+        in
+        match repos with
+        | [] -> None
+        | repos -> Some (sw, repos)
+      ) switches
+  in
+  List.iter (fun (sw, repos)  ->
+    List.iter (fun (rank, repo) ->
```
kit-ty-kate: There is probably a better way to do this but i think it would be nice to have the different switches visually separated. `opam repo show -a default` is quite hard to look at, at the moment
```suggestion
  let did_print_atleast_once = ref false in
  List.iter (fun (sw, repos)  ->
    List.iter (fun (rank, repo) ->
        if !did_print_atleast_once then
          OpamConsole.msg "\n"
        else
          did_print_atleast_once := true;
```

kit-ty-kate (commented):
some more design questions:
- do we want `opam repo show does-no-exist` to exit 0 with no error message?
- do we want an extra `name:` line, even if a bit useless at the moment, it might be useful later if `opam repo show -a` is added
- do we want to have colors? This might help with readability (for example `opam show <pkg>` has its fields coloured blue)
- what is the intended way to parse the output? Wait for switch, then parse each line in this context until a new line starting with `switch: ` ?

ProgramingIsTheFuture:
Hey @kit-ty-kate, I made the above changes.

> Wait for switch, then parse each line in this context until a new line starting with switch:  ?

Yes, the output should be a `name`, `url` and `rank`.

It now works with `-a` and without `repo_name` (it will list all repos for the current switch) and the error message is better now.

PR #6728, opened 2025-10-08
Make the tests even more resistant to user git configuration

dra27:
I got mildly bitten by reftests failures that were nothing to do with what I was breaking, but in fact my local configuration (or lack of it!). I paused for a bit and tried to tweak things. This might want breaking up further:

- The first commit hardens `opam` itself against the user not having configured `init.defaultBranch`. Mostly doesn't matter, as the message isn't normally displayed. `--initial-branch` I think is Git 2.28, but by expressing it as `-c init.defaultBranch=main` it actually maintains compatibility with even older gits than that.
- The second commit hardens the template generator on Windows if Git's `user.name` and `user.email` are not set. The fallback on Unix is to look at `getpwuid`, which isn't implemented on Windows. As a simple fallback, just using `$USERNAME@$USERDOMAIN` is a pretty good default. This showed up as otherwise it means the template doesn't actually have maintainer or email fields set by default.
- While working on later commits, I broke the `patchDiff.ml` test, at which point it would have been nice to see the messages on stderr... which this next commit adds
- Next is a mechanical hardening of `patchDiff.ml` to explicitly set `user.name` and `user.email` on the repo created
- The next commit arose as a result of doing something else that I haven't kept, but I think we should keep this. While I don't propose renaming any branches, it seems a shame to be explicitly specifying `master` as a default for various tests when Git is trying to tend towards `main` - this commit mechanically updates the tests to use `main` wherever possible. (the commit actually arose from removing the `--initial-branch` completely, but I subsequently reverted that)
- Next up is a similar hardening as for `patchDiff.ml` - all the tests now explicitly set `user.name` and `user.email` where needed

The final commit is the point - when running the reftests, the user's global Git configuration is now completely ignored and replaced with one in the reftests itself. This has two benefits:
1. It's hermetic!
2. We don't need to encourage users either to ignore a test failure or to set `protocol.file.allow=always` (see the change in `git.test` in the last commit)

In a similar vein, the last commit also removes most of the manual configuration in CI, so this gets fully tested as well.

Originally, I used this new `tests/reftests/gitconfig` file to include `core.autorcrlf=false`, `init.defaultBranch=main`, `user.name` and `user.email` automatically, thus _removing_ all of them from the tests. But then I decided that was a bad idea and it was better to have every explicit.

No urgency, although I'd quite like the tests to pass first time next time I'm sat at a VM where I happen never to have run `git init` before...!

### DIFF
--- a/.github/scripts/main/opam-rt.sh
+++ b/.github/scripts/main/opam-rt.sh
@@ -10,7 +10,10 @@ export PATH=~/local/bin:$PATH
 export OPAMKEEPLOGS=1
 
 # TODO: Make opam-rt compatible with non-master initial branches
-git config --global init.defaultBranch master
+sudo git config --system init.defaultBranch master
+# TODO: Make opam-rt compatible with a hermetic git configuration
+sudo git config --system user.name 'OPAM test environment'
+sudo git config --system user.email 'noreply@ocaml.org'
 
 cd $CACHE/opam-rt
 make KINDS="local git" run
--- a/.github/scripts/main/preamble.sh
+++ b/.github/scripts/main/preamble.sh
@@ -3,9 +3,11 @@
 CWD=$PWD
 if [ "$RUNNER_OS" = Windows ]; then
   CACHE="$(cygpath 'D:\Cache')"
+  SUDO=''
 else
   CACHE=~/.cache
   CACHE=$(eval echo $CACHE)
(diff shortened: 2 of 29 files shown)


kit-ty-kate:
related to #7073 and https://github.com/ocaml/opam/pull/6992

Issue #6633, opened 2025-08-08
opam install utop fails: ERROR while compiling topkg.1.1.0 (MinGW x64)


blueglyph:
I'm using opam version 2.4.1 in a Windows 10 x64 environment with MinGW 64. Before trying to install OCaml, I updated with `pacman -Syu`.

After installing opam and doing `opam init` by following the instructions on the [website](https://ocaml.org/install#windows) and fixing the paths for MinGW (#6629), I launched:

```
opam install ocaml-lsp-server odoc ocamlformat utop
```

The installation of utop fails, but I'm not familiar with those scripts (it's the most obscure language installation I've ever seen):

```
504:~ $ opam install utop
The following actions will be performed:
=== install 11 packages
  ∗ lambda-term 3.3.2  [required by utop]
  ∗ logs        0.9.0  [required by utop]
  ∗ lwt_react   1.2.0  [required by utop]
  ∗ mew_vi      0.5.0  [required by lambda-term]
  ∗ react       1.2.2  [required by utop]
  ∗ topkg       1.1.0  [required by logs]
  ∗ utop        2.16.0
  ∗ uucp        16.0.0 [required by zed]
  ∗ uuseg       16.0.0 [required by zed]
  ∗ uutf        1.0.4  [required by zed]
  ∗ zed         3.2.3  [required by utop]

Proceed with ∗ 11 installations? [Y/n]

<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><>  🐫
⬇ retrieved lambda-term.3.3.2  (cached)
⬇ retrieved logs.0.9.0  (cached)
⬇ retrieved lwt_react.1.2.0  (cached)
⬇ retrieved mew_vi.0.5.0  (cached)
⬇ retrieved react.1.2.2  (cached)
⬇ retrieved topkg.1.1.0  (cached)
⬇ retrieved utop.2.16.0  (cached)
⬇ retrieved uuseg.16.0.0  (cached)
⬇ retrieved uucp.16.0.0  (cached)
⬇ retrieved uutf.1.0.4  (cached)
⬇ retrieved zed.3.2.3  (cached)
[ERROR] The compilation of topkg.1.1.0 failed at "ocaml pkg/pkg.ml build --pkg-name topkg --dev-pkg false".

#=== ERROR while compiling topkg.1.1.0 ========================================#
[...]

Issue #6455, opened 2025-04-02
opam init should clear environment variables from previous nuked opam roots


kit-ty-kate:
Nuking one's opam root (aka. `rm -rf ~/.opam`) seems to be fairly common way for users to start from scratch when a problem occurs.
However doing this doesn't actually "start from scratch" as the environment variables set from the previous opam installation, stay behind.

Doing the following from a fresh zsh session (including no opam mentions in `~/.zshrc`):
```
opam init --bare --disable-sandbox -a --shell=zsh
opam switch create problem ocaml-base-compiler.5.3.0
rm -rf ~/.opam
```
will leave behind
* the call to `~/.opam/opam-init/init.sh` in `~/.zshrc`
* the following environment variables:
  ```
  PATH=/root/.opam/problem/bin:$PATH
  OPAMNOENVNOTICE=true
  OPAM_SWITCH_PREFIX=/root/.opam/problem
  CAML_LD_LIBRARY_PATH=/root/.opam/problem/lib/stublibs:/root/.opam/problem/lib/ocaml/stublibs:/root/.opam/problem/lib/ocaml
  OCAMLTOP_INCLUDE_PATH=/root/.opam/problem/lib/toplevel
  OCAML_TOPLEVEL_PATH=/root/.opam/problem/lib/toplevel
  MANPATH=:/root/.opam/problem/man
  OPAM_LAST_ENV=/root/.opam/.last-env/env-483e11fe114374e8feace547c3c170f8-0
  ```

Following up with `opam init -n --disable-sandbox`, the environment now looks like this:
```
PATH=/root/.opam/default/bin:/root/.opam/problem/bin:$PATH
OPAMNOENVNOTICE=true
OPAM_SWITCH_PREFIX=/root/.opam/default
CAML_LD_LIBRARY_PATH=/root/.opam/default/lib/stublibs:/root/.opam/default/lib/ocaml/stublibs:/root/.opam/default/lib/ocaml
OCAMLTOP_INCLUDE_PATH=/root/.opam/default/lib/toplevel:/root/.opam/problem/lib/toplevel
OCAML_TOPLEVEL_PATH=/root/.opam/default/lib/toplevel
MANPATH=:/root/.opam/problem/man:/root/.opam/default/man
OPAM_LAST_ENV=/root/.opam/.last-env/env-291dfaf16ea7e7038eb704056d95c782-0
```
[...]

PR #6729, opened 2025-10-08
Don't source `variables.sh` if `OPAM_SWITCH_PREFIX` is already set

### LINKED ISSUE #4649: global switch remains in PATH when local switch is activated
When I `cd` to a directory containing a local opam switch (a `_opam` directory), my `PATH` looks as follows:
```
/home/r/Dokumente/Unisachen/iris/coq/_opam/bin:/home/r/.opam/system/bin:/home/r/bin:/home/r/.cargo/bin:/home/r/.local/bin:/usr/local/bin:/usr/bin:/bin:/usr/local/games:/usr/games
```
Notice now the global "system" switch is in the path even though that switch is not even active! That should not happen, right?

I have the following in my `~/.profile`:
```
# set up OPAM
opam switch system # reset default switch in case I accidentally changed it
test -r /home/r/.opam/opam-init/init.sh && . /home/r/.opam/opam-init/init.sh > /dev/null 2> /dev/null || true
unset MANPATH # unbreak man
```
And in my `~/.bashrc`:
```
# set up OPAM -- needed to get the right local switch when bash is started
test -r /home/r/.opam/opam-init/init.sh && . /home/r/.opam/opam-init/init.sh > /dev/null 2> /dev/null || true
unset MANPATH # unbreak man
```

```
# opam config report
# opam-version      2.0.7 
# self-upgrade      no
# system            arch=x86_64 os=linux os-distribution=debian os-version=testing
# jobs              8
# repositories      4 (http), 3 (local), 2 (version-controlled) (default repo at ff95b837)
# pinned            4 (rsync)
# current-switch    /home/r/Dokumente/Unisachen/iris/coq
```

dra27:
I believe the underlying cause of all the various reports we have had over the years of the environment being incorrect (most recently referred to in https://github.com/ocaml/opam/issues/6455#issuecomment-3303829921) are because the `variables` script is not idempotent.

You can see this on Linux easily by starting a subshell:

```console
opam@eaca5b2675c4:~$ echo $MANPATH
:/home/opam/.opam/default/man
opam@eaca5b2675c4:~$ bash -l
opam@eaca5b2675c4:~$ echo $MANPATH
:/home/opam/.opam/default/man:/home/opam/.opam/default/man
```

or:
```console
Welcome to fish, the friendly interactive shell
Type help for instructions on how to use fish
opam@0ceec1b79104 ~> echo $MANPATH
 /home/opam/.opam/default/man
opam@0ceec1b79104 ~> fish
Welcome to fish, the friendly interactive shell
Type help for instructions on how to use fish
opam@0ceec1b79104 ~> echo $MANPATH
 /home/opam/.opam/default/man /home/opam/.opam/default/man
opam@0ceec1b79104 ~>                                  
```

What then leads to, for example, seeing multiple switch directories in `bin` is that any subsequent call to `eval $(opam env)` and friends will only revert _one_ of those entries, not all of them.

This can be concretely seen on macOS, where the default shell is `zsh` - if opam has been allowed to update `.zshrc`, and from a terminal one launches VS Code, a terminal opened in VS Code _correctly_ resources `.zshrc` and _incorrectly_ ends up with a broken `$PATH` (you can't see the easy `$MANPATH` example, because we don't set `$MANPATH` on macOS).

I think the fix for this trivially simple - if the environment is already set, then the `variables` should just stop. That means, for example, that if you start a subshell, then you correctly inherit the previous environment, which may even be a different switch. I set it up using this `Dockerfile`, which runs using `bash` as the default shell, but for example executes with `fish` as the shell with `docker run --rm -it <sha> fish`:

```Dockerfile
FROM ocaml/opam:ubuntu-24.04-opam
COPY --chmod=755 <<'EOF' /etc/entrypoint.sh
#!/bin/sh
test $# -gt 0 || set -- bash
exec "$@" -l
EOF
ENTRYPOINT ["/etc/entrypoint.sh"]
RUN rm -rf .opam
RUN sudo apt-get update && sudo apt-get install -y fish zsh csh
RUN cat /etc/zsh/newuser.zshrc.recommended > /home/opam/.zshrc
RUN touch /home/opam/.cshrc
#Add opam from this PR to see the fix...
#ADD opam /usr/local/bin/opam
RUN sudo ln -f /usr/bin/opam-2.4 /usr/bin/opam
RUN opam init -a --bare ./opam-repository && opam switch create --empty default
RUN opam init --reinit --shell=zsh -a
RUN opam init --reinit --shell=csh -a
RUN opam init --reinit --shell=fish -a
```

I've utilised a little trick to allow the init-scripts test on Unix actually to test this. I've also put a test for the behaviour of #6455 at the end, but I haven't fixed that here. I think this fix is worth having in its own right, and is more important than the "I nuked my opam root" fix. However, we should fix that too at some point, because "I nuked my opam root" (which is often a sledgehammer response to a problem) is also "I had multiple terminals open, and I've removed the switch which was activated" (which suffers the same problem - `eval $(opam env)` won't revert the previous changes.

Fixes #4649
Fixes https://github.com/ocaml/opam/issues/6633

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -174,6 +174,7 @@ users)
   * Add a test for `opam remove --force` [#6672 @rjbou]
   * Use the new `opam-set-os` command when applicable [#6741 @arozovyk]
   * Add a test showing that `variables.sh` doesn't re-apply updates if sourced more than once in the same session [#6729 @dra27 @kit-ty-kate]
+  * Add a test for #6455 showing the effect on opam env if the switch whose settings were applied by `variables.sh` is deleted [#6729 @dra27]
 
 ### Engine
   * Fix gcc < 14.3 bug on mingw i686 [#6624 @kit-ty-kate]
--- a/tests/reftests/env-idempotent.test
+++ b/tests/reftests/env-idempotent.test
@@ -28,3 +28,8 @@ User configuration:
 ### # we need the switch for variables file
 ### opam switch create fake --empty --root root
 ### bash bash-lc.sh "bash bash-lc.sh \"'$OPAMBIN' exec --root root --no-switch -- env\"" | grep -v MANPATH | grep /root/fake/ | '=.*' -> ''
+### # Simulate nuking a root entirely
+### opam switch create fake2 --empty --root root
+### rm -rf root/fake2
+### bash bash-lc.sh "'$OPAMBIN' exec --root root --switch=fake -- env" | grep -v MANPATH | grep /root/fake2/ | '=.*' -> ''
+PATH


tests/reftests/init-scripts.unix.test
```diff
@@ -1,12 +1,23 @@
 N0REP0
 ### rm -rf $OPAMROOT
-### opam init -na --bypass-checks --bare REPO --root root
+### <locally-in-sh.sh>
+export HOME="$PWD"
+export SHELL="$(command -v sh)"
```
kit-ty-kate:
```suggestion
set -eu
export HOME=$PWD
export SHELL=$(command -v sh)
```

src/state/opamEnv.ml
```diff
@@ -1022,10 +1026,18 @@ let export_in_shell shell =
...
 let env_hook_script shell =
   Option.map (fun script ->
-      export_in_shell shell ("OPAMNOENVNOTICE", "true", None)
+      export_in_shell shell ~unconditional:true ("OPAMNOENVNOTICE", "true", None)
       ^ script)
     (env_hook_script_base shell)
 
+let abort_if_set var shell =
+  match shell with
+  | SH_zsh | SH_bash | SH_sh -> Printf.sprintf "test -z \"$%s\" || return\n" var
```
kit-ty-kate: Fish also works with the same syntax
```suggestion
  | SH_fish | SH_zsh | SH_bash | SH_sh -> Printf.sprintf "test -z \"$%s\" || return\n" var
```

tests/reftests/init-scripts.unix.test
dra27: I get too used to the shellcheck brigade complaining about SC2086 🫣 I don't mind, but this three-line "script" doesn't _need_ `set -eu`?

src/state/opamEnv.ml
dra27: Aw, I was trying to be all fish-idiomatic, given that we use `; or true` in `init.fish`

kit-ty-kate: both styles seem idiomatic https://fishshell.com/docs/current/language.html#combiners-and-or
but if you prefer using `and`/`or` statment, i'd then rather have the same logic as `sh`/etc. here:
```
test -z \"$%s\"; or return\n
```

As a side note, if we were to do the same thing with zsh and bash, the generated code would also look different, probably more like:
```
[[ $OPAM_SWITCH_PREFIX = "" ]] || return
```

tests/reftests/init-scripts.unix.test
kit-ty-kate: the current script no but i'd rather have it to be more future-proof

tests/reftests/init-scripts.unix.test
```diff
@@ -1,12 +1,23 @@
 N0REP0
 ### rm -rf $OPAMROOT
-### opam init -na --bypass-checks --bare REPO --root root
```
kit-ty-kate, line 3: Thinking about it a bit more, i think it might be nice to duplicate this test to another file to make sure we test both `-na` and `-a`

src/state/opamEnv.ml
dra27: Oo, the `set -eu` made me check that again, and it wasn't quite what was meant. The `sh` version is now technically resistant to `set -u` as well. fish's variable expansion is different - as far as I can tell, fish doesn't have any equivalent to `set -u`, so I've not gone with adding the `set -q OPAM_SWITCH_PREFIX` to test for unset as it doesn't seem to be necessary.

tests/reftests/init-scripts.unix.test
dra27: Good idea - in fact, in splitting it, it should now run that test on Windows as well.

src/state/opamEnv.ml
kit-ty-kate: Yeah fish's equivalent to `set -u` is a long term feature request tracked by https://github.com/fish-shell/fish-shell/issues/4163. Lgtm

PR #6745, opened 2025-10-16
GHA: Lighten macos jobs

rjbou:
* remove upgrade from 1.2 to current version macos job
* enable solvers jobs on macos only if `src/solver` is modified

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -549,8 +549,15 @@ jobs:
     steps:
     - name: Checkout tree
       uses: actions/checkout@v5
+    - name: Get changed files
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -549,6 +549,12 @@ jobs:
     steps:
     - name: Checkout tree
       uses: actions/checkout@v5
+    - name: Get changed files
+      id: files
```
kit-ty-kate, line 553: this id isn't used

rjbou: because I forgot to propagate the condition

PR #6746, opened 2025-10-17
Revert "Update Windows bootstrap dependencies #6592"

kit-ty-kate:
Reverts https://github.com/ocaml/opam/pull/6592
```
ake[3]: Leaving directory '/cygdrive/c/Users/opam/opam-release-2.5.0-alpha1/build/x86_64-windows/ocaml-4.14.2/stdlib'
make -C flexdll MSVC_DETECT=0 OCAML_CONFIG_FILE=../Makefile.config CHAINS=mingw64 ROOTDIR=.. \
  OCAMLRUN='$(ROOTDIR)/boot/ocamlruns.exe' NATDYNLINK=false \
  OCAMLOPT='$(OCAMLRUN) $(ROOTDIR)/boot/ocamlc -use-prims ../runtime/primitives -nostdlib -I ../stdlib' \
  flexlink.exe support
make[3]: Entering directory '/cygdrive/c/Users/opam/opam-release-2.5.0-alpha1/build/x86_64-windows/ocaml-4.14.2/flexdll'
echo 'let version = "2.5.0-alpha1"' > version.ml
rm -f COMPILER-*
echo 'let mingw_prefix = "i686-w64-mingw32-"' >> version.ml
x86_64-w64-mingw32-windres -D FLEXDLL_VS_VERSION_INFO=2,5,0-alpha1,0 -D FLEXDLL_FULL_VERSION=\\\"2.5.0-alpha1.0\\\" -i version.rc -o version_res.o
touch COMPILER-41402
echo 'let mingw64_prefix = "x86_64-w64-mingw32-"' >> version.ml
x86_64-w64-mingw32-gcc -Wall -DMINGW -c -o flexdll_mingw64.o flexdll.c
echo 'let msvc = "cl.exe"' >> version.ml
x86_64-w64-mingw32-windres: version.rc:12: syntax error
make[3]: *** [Makefile:210: version_res.o] Error 1
make[3]: *** Waiting for unfinished jobs....
echo 'let msvc64 = "cl.exe"' >> version.ml
echo 'let cygwin64 = "x86_64-pc-cygwin-gcc"' >> version.ml
echo 'let mingw = "i686-w64-mingw32-gcc"' >> version.ml
echo 'let mingw64 = "x86_64-w64-mingw32-gcc"' >> version.ml
echo 'let gnat = "gcc"' >> version.ml
make[3]: Leaving directory '/cygdrive/c/Users/opam/opam-release-2.5.0-alpha1/build/x86_64-windows/ocaml-4.14.2/flexdll'
make[2]: *** [Makefile:171: coldstart] Error 2
make[2]: Leaving directory '/cygdrive/c/Users/opam/opam-release-2.5.0-alpha1/build/x86_64-windows/ocaml-4.14.2'
make[1]: *** [Makefile:313: world.opt] Error 2
make[1]: Leaving directory '/cygdrive/c/Users/opam/opam-release-2.5.0-alpha1/build/x86_64-windows/ocaml-4.14.2'
make: *** [Makefile:122: build/x86_64-windows.env] Error 2
make: *** [qemu] Error 2
```
cc @MisterDA

### DIFF
--- a/.github/scripts/main/ocaml-cache.sh
+++ b/.github/scripts/main/ocaml-cache.sh
@@ -39,7 +39,7 @@ case "$HOST" in
     PREFIX="$OCAML_LOCAL";;
 esac
 
-FLEXDLL_VERSION=0.44
+FLEXDLL_VERSION=0.43
 MINGW_W64_VERSION=12.0.0
 
 curl -sLO "https://github.com/ocaml/ocaml/archive/refs/tags/${OCAML_VERSION}.tar.gz"
--- a/master_changes.md
+++ b/master_changes.md
@@ -72,6 +72,7 @@ users)
 
 ## Build
   * Revert the `msvs-detect` script used on Windows during compiler bootstrap to 0.5.0 [#6746 @kit-ty-kate - revert #6592]
+  * Revert the bootstrap compiler's flexdll to 0.43 [#6746 @kit-ty-kate - revert #6592]
 
 ## Infrastructure
 
--- a/release/Makefile
+++ b/release/Makefile
@@ -4,7 +4,7 @@ OPAM_VERSION = $(subst -,~,$(VERSION))
 GIT_URL = ..
 
 OCAMLV = 4.14.2
-FLEXDLLV = 0.44
+FLEXDLLV = 0.43
 # currently hardcoded in Dockerfile.in
 OCAML_URL = https://github.com/ocaml/ocaml/archive/refs/tags/$(OCAMLV).tar.gz
 FLEXDLL_URL = https://github.com/ocaml/flexdll/archive/refs/tags/$(FLEXDLLV).tar.gz
--- a/src_ext/Makefile
+++ b/src_ext/Makefile
@@ -11,8 +11,8 @@ PATCH ?= patch
 URL_ocaml = https://github.com/ocaml/ocaml/archive/refs/tags/4.14.2.tar.gz
 MD5_ocaml = b28daefda803b5d5910cecf085b1451c
 
-URL_flexdll = https://github.com/ocaml/flexdll/archive/0.44.tar.gz
-MD5_flexdll = 08731d2a5d6465cf7080a14e43e0752e
+URL_flexdll = https://github.com/ocaml/flexdll/archive/0.43.tar.gz
+MD5_flexdll = 6ce706f6c65b2c5adf5791fac678f090
 
 ifndef FETCH
   ifneq ($(shell command -v curl 2>/dev/null),)


MisterDA:
[`opam/release/Makefile#L2`](https://github.com/ocaml/opam/blob/040310f9200a42df561cbcca2314180c2d825839/release/Makefile#L2) sets the `VERSION` variable to (currently) `2.5.0-alpha1`:
```Makefile
VERSION = $(shell git describe $(TAG))
```

FlexDLL [`Makefile#L8-L10`](https://github.com/ocaml/flexdll/blob/3400287999afcdc737f35c1d0e1447c7d2ae5a83/Makefile#L8-L10) has this:
```Makefile
# Fetch the version number from its source, in flexdll.opam
VERSION = \
  $(eval VERSION := $$(shell sed -ne 's/^version: *"\(.*\)"/\1/p' flexdll.opam))$(VERSION)
```

which means that `VERSION` holds the FlexDLL version, _unless_ it is overridden during Make invocation, by passing it as a command-line parameter (setting the variable in the environment doesn't override its value when processing the Makefile) with `make ... VERSION=XXX`. Unfortunately this is exactly what happens at [L175](https://github.com/ocaml/opam/blob/040310f9200a42df561cbcca2314180c2d825839/release/Makefile#L175) and [L184](https://github.com/ocaml/opam/blob/040310f9200a42df561cbcca2314180c2d825839/release/Makefile#L184):
```Makefile
remote: $(OUTDIR)/opam-full-$(VERSION).tar.gz
	# ...
	$(SSH) "$(REMOTE)" 'sh -c "cd $(REMOTE_DIR) && ulimit -s 8192 && $(REMOTE_MAKE) host TAG=$(TAG) VERSION=$(VERSION) OCAMLV=$(OCAMLV)"'
	

qemu: $(OUTDIR)/opam-full-$(VERSION).tar.gz
	# ...
	$(SSH) -p "$(QEMU_PORT)" $(SSH_USER)@localhost "$(REMOTE_SHELL) -c \"cd $(REMOTE_DIR) && $(ULIMIT) $(REMOTE_MAKE) host JOBS=$(JOBS) TAG=$(TAG) VERSION=$(VERSION) OCAMLV=$(OCAMLV)\""
```
This is how the version got overridden.

I suggest you rename `VERSION` to `OPAMV`. We could namespace variables in FlexDLL Makefile too.

Btw `TAG = master` is a bit misleading, `master` is not a tag in Git sense. Maybe `REF` is more appropriate, or `COMMIT_ISH`. (but I digress and I get the intent)

kit-ty-kate:
Thanks for the debug. The most correct fix would then be to replace `$(MAKE)` by `env MAKEFLAGS= MAKEOVERRIDES=  $(MAKE)` per https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html (although their example don't really work in practice weirdly enough, but in the text they do say that the environment takes precedence)

This way if any of OCaml or flexdll change the name of its variables to one of the one we use, these variables won't be overridden by us.

kit-ty-kate:
Superseded by https://github.com/ocaml/opam/pull/6769

PR #6750, opened 2025-10-17
Remove the unused admin scripts

kit-ty-kate:


dra27:
I can see why the scripts can go, but what’s wrong with `opam-admin.top`?

dra27:
I have used it in the past (e.g. https://github.com/dra27/opam/blob/cb298a00b78399af468217718a507f8824d3449a/admin-scripts/ocaml_windows.ml which produced the mass updates in https://github.com/dra27/opam-repository/commit/23827c1cf0244592569fa7ca9a5444a613f549b4), I’ve pointed various people at it in the past as an alternative to horrendously inefficient and fiddly sed/opam-ed scripts, and as it happens I was expecting to use it this week to prototype a solving analysis on packages - so, yes 🙂

Its `iter_packages` mechanism is a very useful method for doing mass updates to repositories - if we haven’t added another mechanism, why remove it? (I was curious as to whether its build had got in the way or made something awkward)

admin-scripts/ocamlinit
```diff
@@ -1,14 +0,0 @@
-#use "topfind";;
```
dra27, line 1: I don't know if @AltGr still uses this, but perhaps it could just be renamed to `init.ml` rather than deleted?

admin-scripts/Makefile
```diff
@@ -1,18 +0,0 @@
-DEPS = core format repository
```
kit-ty-kate, line 3: TODO: keep the makefile for opam_admin_top, probably

kit-ty-kate: actually no, it's built dune in src/tools

admin-scripts/ocamlinit
rjbou: This ocamlinit was removed from root directory and moved here (cf #5526). I think we should keep, maybe elsewhere (shell ?) because it can be useful.

rjbou (changes requested):
I think ocamlinit should stay, maybe somewhere else (that's the requested change).
For other scripts, I'm in favor or not removing them all. Even if hey are not up-to-date, they can be helpful the day we need them, and it will be the occasion to update them. The one i'm thinking about are : couverture.ml, cudf-debug.ml extract-mini-repository.

PR #6751, opened 2025-10-17
Harden the Makefile's inline shell scripts

kit-ty-kate:
Some of the `src_ext` Makefile in particular, broke during the release of 2.5.0~alpha1 for whatever temporary reason but only showed the error well after the error had past

### DIFF
--- a/admin-scripts/Makefile
+++ b/admin-scripts/Makefile
@@ -1,3 +1,5 @@
+.SHELLFLAGS := -eu$(if $(shell (set -o pipefail 2>/dev/null > /dev/null && echo yes)),o pipefail) $(.SHELLFLAGS)
+
 DEPS = core format repository
 
 INCLUDE = $(patsubst %,-I ../src/%,$(DEPS)) -I ../src/tools
--- a/release/Makefile
+++ b/release/Makefile
@@ -1,3 +1,5 @@
+.SHELLFLAGS := -eu$(if $(shell (set -o pipefail 2>/dev/null > /dev/null && echo yes)),o pipefail) $(.SHELLFLAGS)
+
 TAG = master
 VERSION = $(shell git describe $(TAG))
 OPAM_VERSION = $(subst -,~,$(VERSION))
@@ -31,8 +33,7 @@ $(OUTDIR)/opam-full-$(VERSION).tar.gz:
 	git clone $(GIT_URL) -b $(TAG) "$(OUTDIR)/opam-full-$(VERSION)"
 	unset MAKEFLAGS MAKEOVERRIDES && \
 	  $(MAKE) -C "$(OUTDIR)/opam-full-$(VERSION)" OCAML=$(call pathsearch,ocaml) download-ext
-	( set -euo pipefail && \
-	  cd "$(OUTDIR)/opam-full-$(VERSION)" && \
+	( cd "$(OUTDIR)/opam-full-$(VERSION)" && \
 	  git archive "$(TAG)" \
 		--mtime "$$(git show -s --format=%ct "$(TAG)" | tail -1)" \
 		-o "$(abspath $(OUTDIR)/opam-full-$(VERSION).tar)" \
--- a/src_ext/Makefile
+++ b/src_ext/Makefile
@@ -1,3 +1,5 @@
+.SHELLFLAGS := -eu$(if $(shell (set -o pipefail 2>/dev/null > /dev/null && echo yes)),o pipefail) $(.SHELLFLAGS)
+
 ifneq ($(wildcard ../Makefile.config)$(filter-out archives cache-archives,$(MAKECMDGOALS)),)
 -include ../Makefile.config
 endif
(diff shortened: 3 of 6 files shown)


rjbou:
Errors in CI
```
/bin/sh: 1: set: Illegal option -o pipefail
/bin/sh: 1: set: Illegal option -o pipefail

File: release/Makefile [hunk 1]
```diff
@@ -119,6 +119,7 @@ build/$(HOST).env:
 	cd build/$(HOST)/ocaml-$(OCAMLV) && curl -OL $(FLEXDLL_URL)
 	cd build/$(HOST)/ocaml-$(OCAMLV) && tar xzf $(shell basename "$(FLEXDLL_URL)")
 	cd build/$(HOST)/ocaml-$(OCAMLV) && rmdir flexdll && mv "flexdll-$(FLEXDLLV)" flexdll
+	set -euo pipefail && \
```
rjbou: why the pipefail here as there is no pipe (only &&) ?

release/Makefile
```diff
@@ -134,7 +135,8 @@ build/$(HOST).env:
 host: $(OUTDIR)/opam-full-$(VERSION).tar.gz build/$(HOST).env
 	rm -rf build/opam-full-$(VERSION)
 	cd build && tar xzf ../$<
-	( export \
+	( set -euo pipefail && \
```
rjbou: same for pipefail

release/Makefile
kit-ty-kate: the release makefile is ran on known up-to-date platform (latest macOS, latest BSDs, Cygwin, Alpine Linux) which all implement POSIX.1-2024 which includes `set -o pipefail`

kit-ty-kate: the reason it is used even if there no pipe is for future proofing

rjbou (approved):

admin-scripts/Makefile
```diff
@@ -7,7 +9,7 @@ LIBS = $(patsubst %,../src/opam-%.cma,$(DEPS))
 %: %.ml
 	sed 's/^#.*//' $< >$*-tmp.ml
 	ocamlfind ocamlc -package unix,re.glob,ocamlgraph -linkpkg $(INCLUDE) $(LIBS) ../src/tools/opam_admin_top.ml $*-tmp.ml -o $@
-	rm $*-tmp.ml
+	rm "$*-tmp.ml"
```
dra27: I'm not sure about this, given that `$*` is always a valid part of an OCaml filename, so the result here cannot ever need quoting. If we're really going for the belt-and-braces all possible weird filenames fully-hardened version, it's actually:
```suggestion
	rm $(call QUOTE_SINGLE,$*-tmp.ml)
```
given:
```make
QUOTE_SINGLE = '$(subst ','\'',$(1))'
```
(and it ought to be quoted in the previous two commands too)

src_ext/Makefile
```diff
@@ -52,7 +54,7 @@ MD5CHECK = { \
   sum=`md5sum $(1) 2>/dev/null | sed -e 's/^[^a-f0-9]*\([a-f0-9]*\).*/\1/'`; \
   test -n "$$sum" || sum=`md5 $(1) | sed -e 's/^.* = //'`; \
   { test "$$sum" = "$(2)" $(if $(SILENT),,&& echo '$(1) has the expected MD5.'); } || \
-  { rm -f $(1); \
+  { rm -f "$(1)"; \
```
dra27: Same comment here as one the previous `Makefile` - I think the domain of the values used is such that the quoting doesn't make a difference and, if the domain of the input values changes, this isn't necessarily the correct escaping anyway.

dra27 (commented):
Putting quotes around commands doesn't work in make, because the variable is expanded _before_ the shell call is made (i.e. if `$*` is exactly the string `"` then `echo "$*"` will result in `echo """` being passed to the shell).

I would either do the properly hardened approach with `QUOTE_SINGLE` (which comes from https://github.com/ocaml/ocaml/pull/14246 where it's definitely needed because it's in an `install` target) or, to be honest, do nothing - i.e. _either_ have the correctly hardened escaping version, but which is harder to read, or have the unquoted version, rather than a version which at a glance looks quoted, but is not correct.

make targets cannot ever include a space (now or ever - it's a fundamental limitation), so `$*` is always one word.

kit-ty-kate:
actually after trying it a little bit, i'm not sure about `QUOTE_SINGLE`. While the concept is good, in practice i found it a bit too dangerous to use since Make does not warn on undefined variables and the syntax can be difficult to read, i found myself adding new typos which would be hard to detect upon review.

Maybe if we add `RM_RF = rm -rf '$(subst ','\'',$(1))'     [...]     $(call RM_RF,file)` or shorten the variable name to a single letter or something but idk. I find the `$(call ...)` syntax off-putting, making it harder to read and too easy to use wrongly

dra27 (approved):
Let's do it - I'm also making a mental note to put that `.SHELLFLAGS` at the top of every Makefile from now on, because it's a _much_ better default!

PR #6755, opened 2025-10-26
Vendor cmdliner 1.3.0 as opam-core.cmdliner

### LINKED ISSUE #6425: cmdliner 2.0.0 support

kit-ty-kate:
The module name is renamed to OpamCmdliner to allow end-users of the opam libraries to use upstream cmdliner. The library is now also wrapped instead of unwrapped for the same reason.

The `src/core/cmdliner/vendor.sh` script allows us to show and make sure ourselves of the changes between upstream cmdliner and our vendored version.

Fixes #6425 
Backported in 2.5 via https://github.com/ocaml/opam/pull/6756

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -74,6 +74,7 @@ users)
 ## VCS
 
 ## Build
+  * opam no longer depends on `cmdliner` [#6755 @kit-ty-kate - fix #6425]
 
 ## Infrastructure
 
@@ -133,3 +134,4 @@ users)
 ## opam-format
 
 ## opam-core
+  * `OpamCmdliner` was added. It is accessible through a new `opam-core.cmdliner` sub-library [#6755 @kit-ty-kate]
--- a/shell/vendor_cmdliner.sh
+++ b/shell/vendor_cmdliner.sh
@@ -1,0 +1,23 @@
+#!/bin/sh
+
+set -euo pipefail
+
+cd src/core/cmdliner
+rm -rf *.ml *.mli dune
+
+git clone https://github.com/dbuenzli/cmdliner tmp-vendor
+git -C tmp-vendor switch --detach v1.3.0
+
+mv tmp-vendor/src/*.{ml,mli} .
+rm -rf tmp-vendor
+
+mv cmdliner.ml opamCmdliner.ml
+mv cmdliner.mli opamCmdliner.mli
+rm cmdliner_exit.ml{,i}
+
+cat > dune << EOF
+(library
+ (name opamCmdliner)
+ (public_name opam-core.cmdliner)
+ (flags :standard -w -27-32-35-50))
+EOF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,6 @@ src_ext/*.tbz
 src_ext/0install-solver/
 src_ext/archives/*
 src_ext/base64/
-src_ext/cmdliner/
 src_ext/cppo/
 src_ext/cudf/
 src_ext/dose3/
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ This is all you need for installing and using opam, but if you want to use the
 libraries. It's easier to already have a working opam installation in this case,
 so you can do it as a second step.
 
-* Make sure to have `ocamlfind`, `ocamlgraph`, `cmdliner` >= 1.0.0, `cudf` >= 0.7,
+* Make sure to have `ocamlfind`, `ocamlgraph`, `cudf` >= 0.7,
   `dose3` >= 6.1, `re` >= 1.9.0, `opam-file-format` installed. Or run `opam install
   . --deps-only` if you already have a working instance. Re-run
   `./configure` once done
--- a/configure
+++ b/configure
@@ -634,7 +634,6 @@ OCAML_PKG_dose3_algo
 OCAML_PKG_dose3_common
 OCAML_PKG_cudf
 OCAML_PKG_ocamlgraph
-OCAML_PKG_cmdliner
 OCAML_PKG_base64
 OCAML_PKG_re
 OCAML_PKG_unix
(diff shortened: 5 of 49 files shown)


dbuenzli:
> There are reasonable and simple workarounds that allow you to use cmdliner 2.0.0 directly.

I'm just going to mention them because I'm not sure you got them.

1. You can invoke you [evaluation function](https://erratique.ch/software/cmdliner/doc/Cmdliner/Cmd/index.html#eval) with an `env` function that returns `Some "true"` on `"CMDLINER_LEGACY_PREFIXES"` and behave as `Sys.getenv_opt` otherwise (or whatever you inject if you use this). This will make your CLI behave exactly as it is now except for one case: `Arg.enum` value specification. 
2. If you think that the exception of 1. is too widespread you can do a `Unix.putenv` with the corresponding variable set.

doc/index.html
```diff
@@ -301,7 +301,7 @@ <h1>opam %{OPAMVERSION}% API and libraries documentation</h1>
 <tr><th><a href="opam-client/OpamAdminCommand">opamAdminCommand.ml</a></th>
   <td>All sub-commands of the "opam admin" command</td></tr>
 <tr><th><a href="opam-client/OpamCommands">opamCommands.ml</a></th>
-  <td>Opam CLI commands and their handlers as Cmdliner terms</td></tr>
+  <td>Opam CLI commands and their handlers as OpamCmdliner terms</td></tr>
```
rjbou: I'm not sure about this change. Even if we vendor the library, it remains cmdliner terms.

master_changes.md
```diff
@@ -73,6 +73,7 @@ users)
 ## VCS
 
 ## Build
+  * opam no longer depends on `cmdliner` [#6755 @kit-ty-kate - fix #6425]
```
rjbou, line 77:
```suggestion
  * opam no longer depends on `cmdliner`. It is now vendored to avoid version clash and more fine grained . [#6755 @kit-ty-kate - fix #6425]
```
Proposal to be more precise on the how & why

src/core/cmdliner/cmdliner_exit.ml
```diff
@@ -0,0 +1,4 @@
+(*---------------------------------------------------------------------------
```
rjbou: This file is empty ?

shell/vendor_cmdliner.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
```
rjbou, line 1: Where this file is used ?
It needs more documentation & if it is only for the first vendoring (from what i read), it's better to store it in `shell/`

doc/index.html
kit-ty-kate: In spirit yes, but in practice the types aren't compatible

master_changes.md
kit-ty-kate: I don't think we have to do this here. The blog post is the right place for explanations

shell/vendor_cmdliner.sh
kit-ty-kate: It is used for first vendoring and in the future (or now) to have a quick way of showing what has changed (or make sure nothing has changed) compared to upstream 1.3.0. I've moved it to `shell/`

src/core/cmdliner/cmdliner_exit.ml
kit-ty-kate: removed and removal added to the vendoring script

master_changes.md
rjbou: and release notes. But ok to add it at that moment, we just need to remember

doc/index.html
rjbou: Yes, i know :) It is ore about understandability. People understands `Cmdliner` directly, but not `OpamCmdliner`

kit-ty-kate:
> Note that unless I did not read this well, this will prevent people from linking cmdliner and opam libraries.

dune wraps libraries by default, so modules like `Cmdliner_arg` are in fact named `OpamCmdliner__Cmdliner_arg` but accessible as `Cmdliner_arg` locally, allowing us to avoid renaming every call to the modules. See `(wrapped ...)` in [the dune documentation](https://dune.readthedocs.io/en/stable/reference/dune/library.html)

> I honestly have no idea why you decided to make your life so complicated. There are reasonable and simple workarounds that allow you to use cmdliner 2.0.0 directly.

This is really not complicated, this took me 30mins to make and is quite straightforward. We don't need the new features of 2.0.0, and quite the opposite, upgrading to 1.3.0 minus the deprecation to make it compatible with 2.0.0 is quite a bit more work in itself as [this PR](https://github.com/ocaml/opam/pull/6435) shows.

https://github.com/dbuenzli/cmdliner/issues/200 announced the removal of the `CMDLINER_LEGACY_PREFIXES` workaround in 3.0.0 so we're simply thinking ahead. On a more personal note, i personally think this change is quite user-hostile and this lost my trust to any future change that you might make in cmdliner. I personally think cmdliner 2.0.0 should've been a separate new `cmdliner2` library or something like that.

> Note that 1. + 3. gives you exactly what you asked for in [this comment](https://github.com/dbuenzli/cmdliner/issues/200#issuecomment-2695039794).

I don't think it does. The comment was asking for a tangible OCaml expression to ensure that if the feature suddenly disappears, the build would fail. Instead, with the environment variable, the build would continue to succeed but things would now break at runtime.

Note that vendoring this doesn't mean we can't go back to using upstream cmdliner in the future, as long as we're not adding our own patch on top of it, it will be as easy to remove the vendor as it was to create it. Maybe we will decide to stop supporting prefixes and have a proper alias system for shorter names in the future for opam 3.x but we would at least like our command line interface to remain stable for the remaining of opam 2.x.

doc/modules
```diff
@@ -100,7 +100,7 @@ src
 │   ├── opamGitVersion.mli              (generated) Current git version of OPAM
 │   ├── opamArg.ml                      Command-line argument parsers and helpers
 │   │   [ opam exe ]
-│   └── opamMain.ml                     Main, including cmdliner command handling
+│   └── opamMain.ml                     Main, including opam-core.cmdliner command handling
```
rjbou: Same here
```suggestion
│   └── opamMain.ml                     Main, including cmdliner command handling
```

PR #6757, opened 2025-10-26
[2.4 backport] Vendor cmdliner 1.3.0 as opam-core.cmdliner

kit-ty-kate:
Backports https://github.com/ocaml/opam/pull/6755 to the 2.4 branch

This isn’t per-say necessary for us as it would be released at the same time as 2.5.0~alpha2 and opam-repository users would be able to use that version anyway, but i’m opening this PR so that distribution maintainers have a clean patch that they can use.
cc @jamesjer @rwmjones @gildor478 (although please allow a couple of days for the PR to be reviewed)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,7 +16,6 @@ src_ext/*.tbz
 src_ext/0install-solver/
 src_ext/archives/*
 src_ext/base64/
-src_ext/cmdliner/
 src_ext/cppo/
 src_ext/cudf/
 src_ext/dose3/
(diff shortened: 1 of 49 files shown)


kit-ty-kate:
oops my bad, i meant @glondu. Both of your github handle start with 'g' and i must have made a mistake

kit-ty-kate:
Closing since Fedora is going to use the beta instead per https://github.com/ocaml/opam/issues/6425#issuecomment-3504624230

PR #6760, opened 2025-10-28
`install.sh`: update and fix apparmor section

rjbou:
* Fix apparmor profile remplacement option
* Use `install` instead of `mv`+`chmod`+`chown`
* Clean apparmor temporary file
* Use variables instead of plain paths
* Reword apparmor message when user need to check the profile

* [x] Queued on https://github.com/ocaml/opam/pull/6767

### DIFF
--- a/shell/install.sh
+++ b/shell/install.sh
@@ -835,7 +835,9 @@ EOF
     rm "$TMP_APPARMOR_PROFILE"
     echo "AppArmor profile successfully added."
   else
-    echo "Warning: Please make sure an AppArmor profile exists for opam. See $TMP_APPARMOR_PROFILE"
+    echo "Warning: Please make sure an AppArmor profile exists for locally installed opam."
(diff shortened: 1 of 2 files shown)


shell/install.sh
```diff
@@ -826,13 +830,14 @@ EOF
   fi
 
   if [ "$SKIP_APPARMOR" = 0 ]; then
-    xsudo mv /tmp/opam-local.aa.tmp /etc/apparmor.d/opam-local
-    xsudo chmod 644 /etc/apparmor.d/opam-local
-    xsudo chown root:root /etc/apparmor.d/opam-local
-    xsudo apparmor_parser -a /etc/apparmor.d/opam-local
+    xsudo install -m 644 "$TMP_APPARMOR_PROFILE" "$APPARMOR_PROFILE"
```
kit-ty-kate, line 833: The main reason why i haven't used `install` is that it isn't a POSIX tool but i realize that the script already uses `install -m` so i guess this should be alright

PR #6769, opened 2025-11-03
Clean variables before calling make on different projects to avoid clashes

kit-ty-kate:
This fixes the compilation of the OCaml compiler on Windows during the release process

Backported to 2.5 in https://github.com/ocaml/opam/pull/6770

### DIFF
--- a/release/Makefile
+++ b/release/Makefile
@@ -29,7 +29,8 @@ riscv64-linux: $(OUTDIR)/opam-$(VERSION)-riscv64-linux
 $(OUTDIR)/opam-full-$(VERSION).tar.gz:
 	mkdir -p "$(OUTDIR)"
 	git clone $(GIT_URL) -b $(TAG) "$(OUTDIR)/opam-full-$(VERSION)"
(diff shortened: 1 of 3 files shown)


release/Makefile
```diff
@@ -29,7 +29,8 @@ riscv64-linux: $(OUTDIR)/opam-$(VERSION)-riscv64-linux
 $(OUTDIR)/opam-full-$(VERSION).tar.gz:
 	mkdir -p "$(OUTDIR)"
 	git clone $(GIT_URL) -b $(TAG) "$(OUTDIR)/opam-full-$(VERSION)"
-	$(MAKE) -C "$(OUTDIR)/opam-full-$(VERSION)" OCAML=$(call pathsearch,ocaml) download-ext
+	unset MAKEFLAGS MAKEOVERRIDES && \
+	  $(MAKE) -C "$(OUTDIR)/opam-full-$(VERSION)" OCAML=$(call pathsearch,ocaml) download-ext
```
MisterDA, line 33: Is this missing `-j$(JOBS)`?

kit-ty-kate: the release script isn't called with `-j` so it wouldn't matter and i'd rather not download things in parallel anyway to avoid network failures

PR #6775, opened 2025-11-04
[2.5 backport] Allow the macOS sandbox to write in the /var/folders/ and /var/db/mds/ directories

kit-ty-kate:
Backport of https://github.com/ocaml/opam/pull/4797 to the 2.5 branch

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -67,6 +67,7 @@ users)
 ## Format upgrade
 
 ## Sandbox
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
The Doc-* jobs fail due to the change of target branch after the last push (`BASE_REF_SHA: ${{ github.event.pull_request.base.sha }}` is set before that and gets the wrong sha)

PR #6778, opened 2025-11-04
Document how opam chooses the default compiler during opam init and opam switch create

### LINKED ISSUE #6407: Add documentation about default compiler change at init
From #6307, discussed on dev meeting, we should add a note about the default change and a FAQ entry to explain the change and propose the way have the system compiler default behaviour back.

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/6407

### DIFF
--- a/doc/pages/FAQ.md
+++ b/doc/pages/FAQ.md
@@ -437,3 +437,31 @@ We don't recommend using it in day-to-day use of opam in the shell, because
 you'll be typing more and you won't get to notice exciting new features! If the
 behaviour of a command or option is altered, and you write something which in no
 longer valid, opam will try to tell you what the new command should look like.
+
+---
+
+#### 🐫  How does `opam init` and `opam switch create <version>` know which compiler package to choose?
+
+Since opam 2.4, `opam init` will choose `ocaml-base-compiler` as default
+compiler for the `default` switch, and let the solver choose the latest
+version. Of course this default can always be changed using `--compiler`
+or by skipping the creation of the `default` switch all-together with
+`--bare`.
+
+For example if you would rather use the OCaml compiler that you have
+installed by other means, you can use the `ocaml-system` package as follows:
+`opam init --compiler=ocaml-system`. Be careful though, system-wide compilers
+can be setup in a way that the default opam-repository doesn't expect,
+so failures may occur. Thus this option should only be used if you know
+what you're doing.
+
+`opam switch create <version>` on the other hand, works slightly differently.
+Since the only unknown is: which package is going to be used.
+Opam decides this by looking at all the packages that have been tagged as
+`compiler`, filter which ones have the right version, and then let the solver
+decide. For example calling `opam switch create 5.4 5.4.0` will create a
+switch named `5.4` and look for compiler packages with versions `5.4.0` which
+by default, with [opam-repository](https://github.com/ocaml/opam-repository/),
+will get back with `ocaml-base-compiler.5.4.0` and `ocaml-system.5.4.0`.
+Since 2025, `ocaml-system` is tagged as `avoid-version` so the solver will
+choose `ocaml-base-compiler.5.4.0`.
--- a/master_changes.md
+++ b/master_changes.md
@@ -114,6 +114,7 @@ users)
 ## Github Actions
 
 ## Doc
+  * Document how opam chooses the default compiler during `opam init` and `opam switch create` [#6778 @kit-ty-kate - fix #6407]
 
 ## Security fixes
 


doc/pages/FAQ.md
```diff
@@ -437,3 +437,31 @@ We don't recommend using it in day-to-day use of opam in the shell, because
...
+
+---
+
+#### 🐫  How does `opam init` and `opam switch create <version>` know which compiler package to choose?
+
+Since opam 2.4, `opam init` will choose `ocaml-base-compiler` as default
+compiler for the `default` switch, and let the solver choose the latest
+version. Of course this default can always be changed using `--compiler`
+or by skipping the creation of the `default` switch all-together with
+`--bare`.
```
dra27, line 449:
```suggestion
Since opam 2.4, `opam init` will select `ocaml-base-compiler` as the default
compiler package for the `default` switch, and let the solver choose the latest
version. This can always be changed using `--compiler` or by skipping the
creation of the `default` switch altogether with `--bare`.
```

doc/pages/FAQ.md
```diff
@@ -437,3 +437,31 @@ We don't recommend using it in day-to-day use of opam in the shell, because
...
+
+For example if you would rather use the OCaml compiler that you have
+installed by other means, you can use the `ocaml-system` package as follows:
```
dra27, line 452:
```suggestion
installed outside opam, you can use the `ocaml-system` package as follows:
```
?

doc/pages/FAQ.md
```diff
@@ -437,3 +437,31 @@ We don't recommend using it in day-to-day use of opam in the shell, because
...
+`opam init --compiler=ocaml-system`. Be careful though, system-wide compilers
+can be setup in a way that the default opam-repository doesn't expect,
+so failures may occur. Thus this option should only be used if you know
+what you're doing.
```
dra27, line 456:
```suggestion
For example, if you would rather use the OCaml compiler that you have
installed by other means, you can use the `ocaml-system` package as follows:
`opam init --compiler=ocaml-system`. Be careful, though; system-wide compilers
can be set-up in a way that the default opam-repository doesn't expect,
so failures may occur. Thus this option should only be used if you know
what you're doing.
```

doc/pages/FAQ.md
```diff
@@ -437,3 +437,31 @@ We don't recommend using it in day-to-day use of opam in the shell, because
...
+compiler for the `default` switch, and let the solver choose the latest
+version. Of course this default can always be changed using `--compiler`
+or by skipping the creation of the `default` switch all-together with
+`--bare`.
+
+For example if you would rather use the OCaml compiler that you have
+installed by other means, you can use the `ocaml-system` package as follows:
+`opam init --compiler=ocaml-system`. Be careful though, system-wide compilers
+can be setup in a way that the default opam-repository doesn't expect,
+so failures may occur. Thus this option should only be used if you know
```
dra27, line 455:
```suggestion
so failures may occur. We recommend this option should only be used if you know
```
?

doc/pages/FAQ.md
```diff
@@ -437,3 +437,31 @@ We don't recommend using it in day-to-day use of opam in the shell, because
...
+what you're doing.
+
+`opam switch create <version>` on the other hand, works slightly differently.
+Since the only unknown is: which package is going to be used.
```
dra27, line 459:
```suggestion
`opam switch create <version>` on the other hand, works slightly differently,
since the only unknown is exactly which package is going to be used.
```

doc/pages/FAQ.md
```diff
@@ -437,3 +437,31 @@ We don't recommend using it in day-to-day use of opam in the shell, because
...
+Opam decides this by looking at all the packages that have been tagged as
+`compiler`, filter which ones have the right version, and then let the solver
+decide. For example calling `opam switch create 5.4 5.4.0` will create a
+switch named `5.4` and look for compiler packages with versions `5.4.0` which
+by default, with [opam-repository](https://github.com/ocaml/opam-repository/),
+will get back with `ocaml-base-compiler.5.4.0` and `ocaml-system.5.4.0`.
+Since 2025, `ocaml-system` is tagged as `avoid-version` so the solver will
+choose `ocaml-base-compiler.5.4.0`.
```
dra27, line 467:
```suggestion
Opam decides this by finding all packages which have the given and which
that have been flagged as `compiler` and then presents just these packages
to the solver. For example calling `opam switch create 5.4 5.4.0` will create a
switch named `5.4` and look for compiler packages with versions `5.4.0` which
by default, with [opam-repository](https://github.com/ocaml/opam-repository/),
will be with `ocaml-base-compiler.5.4.0`, `ocaml-system.5.4.0` and
`ocaml-variants.5.4.0+options`. `ocaml-system.5.4.0` is flagged as `avoid-version`,
 so the solver will choose `ocaml-base-compiler.5.4.0`.
```
(although I can't remember for certain why ocaml-base-compiler.5.4.0 is guaranteed to be selected, given that it actually has more dependencies than ocaml-variants.5.4.0+options)

dra27 (commented):
Perhaps the FAQ isn't the correct place to be noting `opam switch create 5.4 ocaml.5.4.0`, but it would go well with the part about `opam init --compiler` (i.e. that you can just specify the precise version package and not rely on any of these heuristics)

PR #6804, opened 2025-11-20
Harden makefile invariants

kit-ty-kate:
Split from https://github.com/ocaml/opam/pull/6751. Will update later


dra27:
Oo, one other comment I didn't make on the original PR - make has had `--warn-undefined-variables` for quite some time, but it's difficult to use because it doesn't error. They have recently added other flags for warning on different things and I think being able to surface that as an error (that _might_ not have actually been released yet, though ... I do occasionally hack on GNU make, and I may remembering something from master that isn't released yet).

However, ocaml's CI has some workarounds to be able to test for undefined make variables and make it error in CI; what we've not easily been able to do is get that into a developer's workflow (i.e. have your build fail while you're working on the codebase). The only time it irritates me is having to deal with with testing optional arguments in macros (e.g. having `$(call macro,arg1,arg2)` but where `macro` supports an optional `$3` - you end up having to wrap it with a check on `$(origin)` for `undefined` to avoid the warning).

PR #6812, opened 2025-11-24
Set environment variables needed for bash, zsh completion in opam env

### LINKED ISSUE #6427: Should $(opam env) or .opam/opam-init/* setup paths for shell completion scripts ?
I'm not sure whether this belongs to the `opam-repository` or here but since this could interact with what `opam` does for shell detection I'm doing it here.

`opam` extends quite a few configuration paths so that stuffs gets looked up in the opam prefix (e.g. `PATH`, `MANPATH`[^1]). I was wondering if perhaps `opam` could do the same for shell completion scripts according to the current shell. That way packages can simply install completion scripts in the prefix where the various shell expect them and they get automatically picked up by the user shell as packages come and go.

It seems at [least me an another person] are interested :-)

What do you think ? 

[least me an another person]: https://discuss.ocaml.org/t/installing-shell-completion-file-in-an-opam-package/13540

[^1]: On macos I don't know how, but invoking `manpath` has it.

WardBrian:
Would close #6427 

This updates  `opam env` to set `XDG_DATA_DIRS` (from the [XDG base directory specification](https://specifications.freedesktop.org/basedir/latest/)) to include the current switch's share root, and `FPATH` to include `%{share}%/zsh/site-functions` when the shell is zsh

The former works with [`bash-completion`](https://github.com/scop/bash-completion), which looks in the XDG_DATA_DIRS entries for `bash-completion/completions` sub-directories, while the later sets up zsh's built-in completion mechanisms to look in this location.

Questions for reviewers --

1. I currently only modify FPATH on zsh, but I think it would be safe to do it generically (the only other shell that seems to assign meaning to the FPATH variable is ksh, which opam doesn't support)
2. Assuming we want to keep FPATH zsh-specific, is it worth threading through a shell variable all the way from the CLI to where we compute these updates, or is it fine to keep the `guess_shell_compat` call?

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -59,6 +59,7 @@ users)
 ## Clean
 
 ## Env
+  * Env now sets `XDG_DATA_DIRS` to include the opam share directory and `FPATH` to include share/zsh/site-functions. Tools which install shell autocomplete scripts in these directories will now be picked up automatically. [#6427 @WardBrian - fix #6812]
 
 ## Opamfile
 
--- a/src/state/opamEnv.ml
+++ b/src/state/opamEnv.ml
@@ -613,7 +613,7 @@ let env_update_resolved_with_default ?comment var =
   let rewrite = Some (SPF_Resolved (Some (default_sep_fmt_str var))) in
   env_update_resolved ?comment ~rewrite var
 
-let compute_updates ?(force_path=false) st =
+let compute_updates ?(force_path=false) ?(shell=None) st =
   (* Todo: put these back into their packages!
   let perl5 = OpamPackage.Name.of_string "perl5" in
   let add_to_perl5lib =  OpamPath.Switch.lib t.root t.switch t.switch_config perl5 in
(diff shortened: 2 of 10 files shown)


WardBrian:
The failing windows reftests are fairly inscrutable to me, and I'm extra confused as to why only some configurations are failing in that matrix

kit-ty-kate:
> The failing windows reftests are fairly inscrutable to me, and I'm extra confused as to why only some configurations are failing in that matrix

all Native configurations of windows fail. The only "windows" configurations that work fine are the ones using the cygwin backend of ocaml, which mains that it simply works like unix does.

The error is:
```diff
diff --git a/_build/default/tests/reftests/env-idempotent.test b/_build/default/tests/reftests/env-idempotent.out
index cb5a13d..5d23c8f 100644
--- a/_build/default/tests/reftests/env-idempotent.test
+++ b/_build/default/tests/reftests/env-idempotent.out
@@ -34,4 +34,3 @@ User configuration:
 ### rm -rf root/fake2
 ### bash bash-lc.sh "'$OPAMBIN' exec --root root --switch=fake -- env" | grep -v MANPATH | grep /root/fake2/ | '=.*' -> ''
 PATH
-XDG_DATA_DIRS
File "tests/reftests/env-idempotent.test", line 1, characters 0-0:
"C:\Program Files\Git\cmd\git.exe" --no-pager diff --no-index --color=always -u --ignore-cr-at-eol _build/default/tests/reftests/env-idempotent.test _build/default/tests/reftests/env-idempotent.out
Command exited with code 1.
diff --git a/_build/default/tests/reftests/env.win32.test b/_build/default/tests/reftests/env.win32.out
index 7dee3d0..511d6ed 100644
--- a/_build/default/tests/reftests/env.win32.test
+++ b/_build/default/tests/reftests/env.win32.out
@@ -811,7 +811,17 @@ Processing  2/3: [all-formulae: sh env | grep RAF_ENV | sort]
 [ERROR] Formula can't be completely resolved : RAF_ENVSET_UNRES ("host" true | "target-quoted" true). Using default ';' 'target'.
 [ERROR] Formula can't be completely resolved : RAF_ENVSET_UNRES ":" true | ";" true. Using default ';' 'target'.
 [ERROR] Formula can't be completely resolved : RAF_ENVSET_UNRES ("host" true | "target-quoted" true). Using default ';' 'target'.
+[ERROR] Formula can't be completely resolved : RAF_ENVSET_UNRES ":" true | ";" true. Using default ';' 'target'.
+[ERROR] Formula can't be completely resolved : RAF_ENVSET_UNRES ("host" true | "target-quoted" true). Using default ';' 'target'.
+[ERROR] Formula can't be completely resolved : RAF_ENVSET_UNRES ":" true | ";" true. Using default ';' 'target'.
+[ERROR] Formula can't be completely resolved : RAF_ENVSET_UNRES ("host" true | "target-quoted" true). Using default ';' 'target'.
+[ERROR] Formula can't be completely resolved : RAF_ENVSET_UNRES ":" true | ";" true. Using default ';' 'target'.
+[ERROR] Formula can't be completely resolved : RAF_ENVSET_UNRES ("host" true | "target-quoted" true). Using default ';' 'target'.
 Done.
+[ERROR] Formula can't be completely resolved : RAF_ENVSET_UNRES ":" true | ";" true. Using default ';' 'target'.
+[ERROR] Formula can't be completely resolved : RAF_ENVSET_UNRES ("host" true | "target-quoted" true). Using default ';' 'target'.
+[ERROR] Formula can't be completely resolved : RAF_ENVSET_UNRES ":" true | ";" true. Using default ';' 'target'.
+[ERROR] Formula can't be completely resolved : RAF_ENVSET_UNRES ("host" true | "target-quoted" true). Using default ';' 'target'.
 ### opam env | sort | grep "RAF_ENV"
 RAF_ENVSET_ATOM='\is\atom'; export RAF_ENVSET_ATOM;
 RAF_ENVSET_DBL='sec/ond;fir/st'; export RAF_ENVSET_DBL;
```

The change in `env.win32.test` is straightforward since you've done the same change in `env.unix.test`. Since you've only ran the testsuite on Unix, only the tests that work on Unix were ran and the ones that only run on Windows were skipped. So to fix that you can either run the testsuite on Windows again or simply trust the CI and apply that patch.

Now for `env-idempotent.test`, i'm not exactly sure, it depends what that test was supposed to show, but it could probably be fixed by adding another ` | grep -v XDG_DATA_DIRS` to the command

WardBrian:
> Now for `env-idempotent.test`, i'm not exactly sure, it depends what that test was supposed to show

I am not entirely sure myself, I tried digging through the git history but also came up empty handed. Since it's already filtering out manpath, an extra filter doesn't seem so bad

PR #6818, opened 2025-11-27
`opam show`: Reduce I/O when multiple versions of a package exist

rjbou:
Extracted from #6515, authored by @dra27
_I noticed with `opam show dune` that the switch selections of the switches are read in proportion to the number of versions being displayed. That creates both unnecessary I/O and GC pressure. [...] I'd countenance against hyping the performance improvement on `opam show` because it won't always show up that way!_

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -168,3 +168,4 @@ users)
   * `OpamCompat.Int.min`: was added [#6515 @kit-ty-kate]
   * `OpamStd.String.compare_case`: is now allocation free [#6515 @dra27]
   * `OpamVersionCompare.{compare,equal}`: are now allocation free [#6515 @dra27]
+  * `OpamCompat.Map.add_to_list`: was added [#6818 @dra27]
--- a/master_changes.md
+++ b/master_changes.md
@@ -42,6 +42,7 @@ users)
 ## List
 
 ## Show
+  * Improve performance of `opam show` by reading switch selection only once instead of once per package-version [#6818 @dra27]
 
 ## Var/Option
 
--- a/master_changes.md
+++ b/master_changes.md
@@ -156,6 +156,7 @@ users)
 
 ## opam-state
   * `OpamRepositoryState.load_opams_from_diff` track added packages to avoid removing version-equivalent packages [#6774 @arozovyk fix #6754]
+  * `OpamGlobalState.all_installed_versions`: was added [#6818 @dra27]
 
 ## opam-solver
 
--- a/master_changes.md
+++ b/master_changes.md
@@ -158,6 +158,7 @@ users)
 ## opam-state
   * `OpamRepositoryState.load_opams_from_diff` track added packages to avoid removing version-equivalent packages [#6774 @arozovyk fix #6754]
   * `OpamGlobalState.all_installed_versions`: was added [#6818 @dra27]
+  * `OpamGlobalState.installed_versions`: was removed [#6818 @dra27]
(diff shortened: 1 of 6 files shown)


master_changes.md
```diff
@@ -42,6 +42,7 @@ users)
 ## List
 
 ## Show
+  * Improve performance of `opam show` by reading switch selection only once instead of once per package-version [#6515 @dra27]
```
```suggestion
  * Improve performance of `opam show` by reading switch selection only once instead of once per package-version [#6818 @dra27]
```

master_changes.md
```diff
@@ -153,6 +154,7 @@ users)
 
 ## opam-state
   * `OpamRepositoryState.load_opams_from_diff` track added packages to avoid removing version-equivalent packages [#6774 @arozovyk fix #6754]
+  * `OpamGlobalState.all_installed_versions`: was added [#6515 @dra27]
```
```suggestion
  * `OpamGlobalState.all_installed_versions`: was added [#6818 @dra27]
```

master_changes.md
```diff
@@ -161,3 +163,4 @@ users)
 ## opam-core
   * `OpamCmdliner` was added. It is accessible through a new `opam-core.cmdliner` sub-library [#6755 @kit-ty-kate]
   * `OpamUrl`: rename and expose `local_path` as `looks_like_local_path` [#5979 @kit-ty-kate]
+  * `OpamCompat.Map.add_to_list`: was added [#6515 @dra27]
```
```suggestion
  * `OpamCompat.Map.add_to_list`: was added [#6818 @dra27]
```

src/state/opamGlobalState.mli
```diff
@@ -41,6 +41,10 @@ val switch_exists: 'a global_state -> switch -> bool
     of switches they are installed in *)
 val installed_versions: 'a global_state -> name -> switch list package_map
```
rjbou: Should we keep `installed_versions`? It is not used in the codebase, and sherlocode doesn't highlight any use

src/state/opamGlobalState.ml
```diff
@@ -167,18 +167,22 @@ let all_installed gt =
       OpamPackage.Set.union acc sel.sel_installed)
     gt  OpamPackage.Set.empty
 
-let installed_versions gt name =
+let installed_versions_t gt ~cond =
```
rjbou: The last commit proposes a factorisation of the 2 functions

src/state/opamGlobalState.ml
```diff
@@ -167,18 +167,22 @@ let all_installed gt =
...
+      OpamPackage.Set.fold (fun nv acc ->
+          if cond nv then
+            OpamPackage.Map.add_to_list nv switch acc
+          else acc)
+        sel.sel_installed acc)
     gt OpamPackage.Map.empty
 
+let installed_versions gt name =
+  installed_versions_t gt
+    ~cond:(fun nv -> OpamPackage.Name.equal name (OpamPackage.name nv))
```
rjbou:
```suggestion
    ~cond:(fun nv -> OpamPackage.Name.equal name nv.name)
```

src/state/opamGlobalState.mli
rjbou: Discussed in dev meeting: This can be removed, especially as it is possible to recompute that information from the new function.

kit-ty-kate (approved):
ocaml-benchmarks shows a speedup between 4 and 21%

PR #6823, opened 2025-12-01
Fix the AppArmor support in the install script

### LINKED ISSUE #6820: [ERROR] Sandboxing is not working on your platform neon:
[ERROR] Sandboxing is not working on your platform neon:
        "~/.opam/opam-init/hooks/sandbox.sh build sh -c echo SUCCESS | tee check-write" exited
        with code 1 "bwrap: loopback: Failed RTM_NEWADDR: Operation not permitted"
Do you want to disable it?  Note that this will result in less secure package builds, so please
ensure that you have some other isolation mechanisms in place (such as running within a
container or virtual machine). [y/N]

kit-ty-kate:
AppArmor currently doesn't support having multiple profiles targetting the same path. AppArmor currently ships with a builtin opam profile targetting /usr/bin/opam so we have to avoid installing our opam-local profile if BINDIR=/usr/bin

Fixes #6820 

---
*Special thanks to the AppArmor maintainer who replied to my questions on IRC.*

### DIFF
--- a/shell/install.sh
+++ b/shell/install.sh
@@ -808,6 +808,7 @@ echo "## opam $VERSION installed to $BINDIR"
 if [ -f /etc/apparmor.d/abi/4.0 ] && [ "$(aa-enabled 2> /dev/null)" = Yes ]; then
   TMP_APPARMOR_PROFILE=/tmp/opam-local.aa.tmp
   APPARMOR_PROFILE=/etc/apparmor.d/opam-local
+  UPSTREAM_APPARMOR_PROFILE=/etc/apparmor.d/opam
   cat << EOF > "$TMP_APPARMOR_PROFILE"
 # This profile allows everything and only exists to give the
 # application a name instead of having the label "unconfined"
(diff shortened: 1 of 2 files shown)


shell/install.sh
```diff
@@ -826,7 +827,10 @@ EOF
   SKIP_APPARMOR=0
   APPARMOR_CREATION_OPTION="-a"
   if [ -e "$APPARMOR_PROFILE" ]; then
-    if diff -q "$TMP_APPARMOR_PROFILE" "$APPARMOR_PROFILE"; then
+    if [ -f "$UPSTREAM_APPARMOR_PROFILE" ] && [ "$(realpath "$BINDIR")" = /usr/bin ]; then
+      APPARMOR_PROFILE=$UPSTREAM_APPARMOR_PROFILE
```
rjbou, line 831: Why changing the value of `APPARMOR_PROFILE`?

kit-ty-kate: because in this case you want people to look at the `opam` profile not `opam-local` (see error message at the end of the section). You want users to make sure that the upstream `opam` profile actually targets the right path

rjbou: in that case shouldn't the message below be updated ? it refers to the tmp file where local opam aa profile is written (`TMP_APPARMOR_PROFILE`).

kit-ty-kate: TMP_APPARMOR_PROFILE also exists in this case, it is created above. I think the message still makes sense with the updated value

rjbou: I've missed the last line. lgtm!

rjbou:
Benchmark failing. Skipping.

PR #6826, opened 2025-12-02
Avoid downgrading root packages as much as possible when using the builtin-0install solver

### LINKED ISSUE #6824: builtin-0install: Avoid removing or downgrading packages on upgrades
This might be doable by ordering the explicitly installed packages (pinned or root) earlier in the list given to opam-0install-cudf.

This would require to **not** merge https://github.com/ocaml/opam/pull/6531

Related to #6586

kit-ty-kate:
Partially fixes #6824 

This helps in cases where `a` needs to downgrade `b` and neither are required by anything else in the current switch.
If another package in the switch depends on `a` then anything can happen to `b`, sadly.

I'm guessing 0install-solver prioritizes the first dependency tree scanned (depth first) so ordering requested packages to emulate a priority system doesn't work in most cases because of that.

TODO:
- [ ] tests
- [ ] extract the extra_pkg names away from OpamCudf to avoid recursion in the build plan between OpamCudf and OpamBuiltin0install
- [ ] figure out a way to switch 0install from depth-first to breadth-first or "depth-first modulo first iteration"

### DIFF
--- a/src/solver/opamBuiltin0install.ml
+++ b/src/solver/opamBuiltin0install.ml
@@ -62,23 +62,51 @@ let restricts (pkgs, constraints) (pkg, c) =
   in
   (pkgs, constraints)
 
+(* From OpamCudf *)
+let s_installed_root = "installed-root"
+
 let create_spec ~drop_installed_packages universe request =
+  let open (struct
+    type installed_pkgs =
+      | Keep of Cudf_types.vpkg
+      | Installed_priority of Cudf_types.vpkg
+      | Installed of Cudf_types.vpkg
+  end) in
+  let installed_pkgs =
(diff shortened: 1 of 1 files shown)


talex5:
> if you know a simple way to make 0install-solver non depth-first to at least try it out

The depth-first search is here:

https://github.com/0install/0install/blob/33275432c8579f519ae273b1921189aa3caec707/src/solver/solver_core.ml#L549

I don't recall any reason why it couldn't be breadth-first, but I haven't thought about it for a long time.

(possibly this code is a bit inefficient because it walks the tree every time it has to make a decision... but it's probably lucky I didn't optimise it as that would make changing it harder)

PR #6827, opened 2025-12-04
[API] Split the deprecated url and descr files implementations to their own submodules

kit-ty-kate:
Per discussion with @rjbou regarding https://github.com/ocaml/opam/pull/6253, we'd like to keep the format upgrade code around if only as a basis for opam 3.0.

The real issue with `url` and `descr` is that they are used even when the opam file being parsed is up-to-date. This is documented but we'd like to move forward and stop doing that and require a manual upgrade to read them.

However currently the code for `url` and `descr` are in their own `URL` and `Descr` submodule, which is used for for the deprecated file formats and as a datatype (including the new features). This PR splits these submodules into two of each to separate the legacy part (dealing with the legacy file formats) and the "normal" submodules which still holds the current datatypes.

~TODO: while doing this change i chose to change the representation of synopsis/description, which probably should be moved to its own PR with its own discussion.~

* Queued on https://github.com/ocaml-opam/opam-publish/pull/194

### DIFF
--- a/doc/pages/Manual.md
+++ b/doc/pages/Manual.md
@@ -1211,10 +1211,11 @@ files.
   defines a long description (one or more paragraphs) for the package. This can
   also be defined as the body of an external [`descr`](#descr) file.
 
-- <a id="opamsection-url">`url "{" <url-file> "}"`</a>:
-  defines the URL where the package source can be obtained. This section has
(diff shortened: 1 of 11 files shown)


rjbou (approved):

rjbou:
related to #6834

PR #6821, opened 2025-11-27
Update the changelog, GHA and documentation after the release of opam 2.5.0

kit-ty-kate:

PR #6832, opened 2025-12-09
GHA: Regenerate the cache when OPAM_TEST_REPO_SHA is changed

kit-ty-kate:
Required for https://github.com/ocaml/opam/pull/6821


dra27 (approved):
LGTM!

PR #6393, opened 2025-02-20
Internal `repos-config` new syntax

rjbou:
It is quite difficult and misleading to repository related information in the `repos-config` file (like in #4933 and #6283).
The current format contain a list with a some optional elements, which complicated the parsing
> `repositories: [ <string> { <URL> } { {<string>}+ } { <int> } ... ]`: lists the configured repository idents, their URLs and trust anchors. The format is similar to [repositories:](https://opam.ocaml.org/doc/Manual.html#opamrcfield-repositories) from opamrc, except that the <URL> itself is optional.

The PR implements a new syntax for this file: each repository is a section with it's own defined labeled fields:
```
repo <string> {
  url: <URL>
  fingerprints:  <string>
  quorum: <int>
}
```
Adding then a new field is quite easy.

This change implies an hard upgrade (no best effort loading if read only) to update the opam root.

fix #6327

TODO:
* [ ] update doc
* [ ] update changes
* [x] update opam root test
* [x] move to light upgrade
* [x] add a comment on hard upgrade that it should not be changed, ideally (done in #6417)
* [x] queued on #6417 
* [x] queued on #6839

PR #6839, opened 2025-12-18
Fix switch & repo format upgrade

rjbou:
Seen on #6393 (https://github.com/ocaml/opam/pull/6393#issuecomment-3607243913). The lock can't be acquired when opam want to write the new repo state on Windows. That was caused by taking a new `OpamSystem.lock` on the lock file instead of reusing the already created one and update its system lock.
Introduced in #5305, related to #6417

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -72,6 +72,7 @@ users)
 ## External dependencies
 
 ## Format upgrade
+  * Fix switch and repo format upgrade on Windows. A block occurred because the global lock fd was reopened instead of using the one already opened.  [#6839 @rjbou]
 
 ## Sandbox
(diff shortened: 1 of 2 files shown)


master_changes.md
```diff
@@ -71,6 +71,7 @@ users)
 ## External dependencies
 
 ## Format upgrade
+  * Fix switch and repo format upgrade on Windows. A block occurred because the global lock fd was reopened instead of using the one already opened.  [€6839 @rjbou]
```
```suggestion
  * Fix switch and repo format upgrade on Windows. A block occurred because the global lock fd was reopened instead of using the one already opened.  [#6839 @rjbou]
```

PR #6841, opened 2025-12-29
add `license` as a possible field for `opam list`

raphael-proust:

### DIFF
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -1761,6 +1761,8 @@ let package_listing cli =
   let columns =
     mk_opt ~cli cli_original ["columns"] "COLUMNS" ~section
       (Printf.sprintf "Select the columns to display among: %s.\n\
+                       The special form $(b,<field>:) (field name followed by \
+                       colon, e.g., $(b,license:)) selects an arbitraty field. \
                        The default is $(b,name) when $(i,--short) is present \
                        and %s otherwise."
          (OpamStd.List.concat_map ", " (fun (_,f) -> Printf.sprintf "$(b,%s)" f)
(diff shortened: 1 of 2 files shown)


src/client/opamListCommand.ml
```diff
@@ -624,6 +627,11 @@ let detail_printer ?prettify ?normalise ?(sort=false) installed st nv =
          then s % [`bold;`yellow] else
            s % [`magenta]
      with Not_found -> "--" % [`cyan])
+  | License ->
+    (try mini_field_printer
+      (OpamStd.List.assoc String.equal "license"
```
kit-ty-kate: using assoc_opt here would be preferable

kit-ty-kate (commented):
Thanks for your contribution. Would you have time to add some tests? (format is documented in https://github.com/ocaml/opam/blob/master/tests/reftests/readme.md) If not we'll do it whenever we have some time

src/client/opamListCommand.ml
raphael-proust: done!

src/client/opamArg.ml
```diff
@@ -1761,6 +1761,8 @@ let package_listing cli =
   let columns =
     mk_opt ~cli cli_original ["columns"] "COLUMNS" ~section
       (Printf.sprintf "Select the columns to display among: %s.\n\
+                       The special form $(b,<field>:) (field name followed by \
+                       colon, e.g., $(b,license:)) selects an arbitraty field. \
```
kit-ty-kate, line 1765:
```suggestion
                       colon, e.g., $(b,license:)) selects an arbitrary field. \
```

PR #6846, opened 2026-01-23
document raw-field syntax more prominently

raphael-proust:
split from #6841 as recommended in https://github.com/ocaml/opam/pull/6841#issuecomment-3729238438

### DIFF
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -1761,6 +1761,8 @@ let package_listing cli =
   let columns =
     mk_opt ~cli cli_original ["columns"] "COLUMNS" ~section
       (Printf.sprintf "Select the columns to display among: %s.\n\
+                       The special form $(b,<field>:) (field name followed by \
+                       colon, e.g., $(b,license:)) selects an arbitrary field. \
                        The default is $(b,name) when $(i,--short) is present \
                        and %s otherwise."
          (OpamStd.List.concat_map ", " (fun (_,f) -> Printf.sprintf "$(b,%s)" f)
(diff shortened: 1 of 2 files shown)


rjbou (changes requested):
Thank you for the PR!
There is a need to have a test for `opam list` format. It would be better to have it in a separate PR, it'll permit to review on its own and have it from the beginning the most complete one.
The documentation part, lgtm!

rjbou:
I understand your point of view, even agree in most cases. Opam cram test suite is (ideally) append/promote only with no removal, to be able to follow the changes over time. So if we merge this test today, we'll need to construct the new test around it (restrictive), or craft a new one and we'll have duplicated tests for opam list. That's why we prefer, when we can, from the beginning, to have the most complete test.

In fact, i wasn't asking you to complete the test for the new PR. Opening one as is would be enough for us to complete it on our own.

rjbou (approved):
Thanks!

src/client/opamArg.ml
```diff
@@ -1761,6 +1761,8 @@ let package_listing cli =
   let columns =
     mk_opt ~cli cli_original ["columns"] "COLUMNS" ~section
       (Printf.sprintf "Select the columns to display among: %s.\n\
+                       The special form $(b,<field>:) (field name followed by \
+                       colon, e.g., $(b,license:)) selects an arbitraty field. \
```
kit-ty-kate:
```suggestion
                       colon, e.g., $(b,license:)) selects an arbitrary field. \
```

kit-ty-kate (approved):
other than that lgtm

kit-ty-kate:
Thanks a lot! Sorry for the delay

Issue #6848, opened 2026-01-26
Opam init won't try to install default ocaml compiler when failed once.


przemmeq0:
**1. TL;DR**
Edit: this post describes only a specific scenario. For a comprehensive approach please follow to "Update" post.
After fresh installation of opam, **if** first run of command `opam init` will fail to create default switch (i.e. no ocaml compiler is build), **and** the cleanup is confirmed: `Switch initialisation failed: clean up? ('n' will leave the switch partially
installed) [Y/n] y` **then** (in my case) every other run of `opam init` will silently exit without creating default switch like so:
```
<><> Required setup - please read <><><><><><><><><><><><><><><><><><><><><><><>

  In normal operation, opam only alters files within ~/.opam.

  However, to best integrate with your system, some environment variables
  should be set. If you allow it to, this initialization step will update
  your bash configuration by adding the following line to ~/.bashrc:

    test -r '/home/userpi/.opam/opam-init/init.sh' && . '/home/userpi/.opam/opam-init/init.sh' > /dev/null 2> /dev/null || true

  Otherwise, every time you want to access your opam installation, you will
  need to run:

    eval $(opam env)

  You can always re-run this setup with 'opam init' later.

Do you want opam to configure bash?
> 1. Yes, update ~/.bashrc
  2. Yes, but don't setup any hooks. You'll have to run eval $(opam env)
     whenever you change your current 'opam switch'
  3. Select a different shell
  4. Specify another config file to update instead
  5. No, I'll remember to run eval $(opam env) when I need opam

[1/2/3/4/5] 1

User configuration:
  ~/.bashrc is already up-to-date.
[NOTE] Make sure that ~/.bashrc is well sourced in your ~/.profile.
```
I believe this issue is platform-independent, and only minor overlook in the installation script. I finally managed to complete successful `opam init` after manually deleting `~/.opam` directory, but it isn't obvious, particularly because of apparent "clean up" that `opam init` promised. While the fix is simple when known I haven't found it noted anywhere in installation tutorial, and in my opinion other possible fixes are harmful (e.g. installing outdated ocaml with system package manager and trying to sync it with opam, or manual creation of a switch by novice user like myself).

**2. Full scenario:**
Disclaimer: the steps are reproduced in good will. At any point it was not my intention to be cynical, or insult anyone. I believe that current behavior might feel like "dead end" for other users beside me, and I'm reporting it is just for "smoother experience" for possible future new users.
[...]

PR #6851, opened 2026-01-27
Show a working opam switch create command when opam init fails to create the default switch

kit-ty-kate:
Partially f i x e s https://github.com/ocaml/opam/issues/6848#issuecomment-3806778559

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -23,6 +23,7 @@ users)
 ## Init
   * ✘ Display an appropriate error message when the file given to `opam init --config` does not exist or is in a VCS. This changes the behaviour for VCS local urls that was previously retrieved. [#5979 @kit-ty-kate - fix #5971]
   * Improve the performance of `opam init --bare` by no longer computing the default compiler in that case [#6957 @tuesdayjz @kit-ty-kate - fix #5815]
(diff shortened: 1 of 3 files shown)


rjbou (approved):
thanks!

tests/reftests/init.test
```diff
@@ -152,6 +156,95 @@ Switch invariant: ["ocaml-system"]
 Done.
 ### opam switch invariant
 ["ocaml-system"]
+### :I:5: failed compiler build
```
rjbou: Shouldn't be IV ?
```
### :IV: Advice message when default switch creation fails

Comment (rjbou) [reply tests/reftests/init.test]: In that case, the package definition can be added just for this test, at the beginning (between :IV: and :IV:1:)

File: master_changes.md [hunk 2]
```diff
@@ -21,6 +21,7 @@ users)
 
 ## Init
   * ✘ Display an appropriate error message when the file given to `opam init --config` does not exist or is in a VCS. This changes the behaviour for VCS local urls that was previously retrieved. [#5979 @kit-ty-kate - fix #5971]
+  * Show a working `opam switch create` command when `opam init` fails to create the default switch [#6851 @kit-ty-kate]
```
```suggestion
  * Show a working `opam switch create` command when `opam init` fails to create the default switch [#6851 @kit-ty-kate - partial fix #6848 ]
```

Issue #6853, opened 2026-01-28
FR: PowerShell completion


WardBrian:
Especially with Windows support growing, this would be a real nice-to-have.

The easiest way to get it might be bumping the vendored cmdliner and applying [the patch mentioned here](https://github.com/ocaml/dune/issues/12515#issuecomment-3802602856) to preserve the prefix behavior `opam` desires. I'd be happy to have a go at this as a PR if it sounds desirable.

Issue #6849, opened 2026-01-27
Internal OpamCmdliner lagging behind, missing fixes in upstream cmdliner


jberdine:
What is the expected future of the internal OpamCmdliner copy of cmdliner? I ask because it is lagging behind upstream cmdliner and missing fixes. For example I am missing the fix for https://github.com/dbuenzli/cmdliner/issues/191 that is in cmdliner but not opam.

rjbou:
Discussed in dev meeting: we'll definitively backport the bug fixes (2.6 milestone). For later, we should be able to vendor 2.1 cmdliner with @dbuenzli  proposed patch. It'll require some work on `opam-client` dependencies to upgrade.

PR #6854, opened 2026-01-30
Source compatibility with newer cmdliner versions

WardBrian:
Related to issues #6424, #6849, #6853

This PR makes the necessary code changes to allow compilation under cmdliner 2.0 **while keeping the existing vendored version of 1.3.0**.

These changes are mostly find-and-replace type things, but they would allow (at some future point):

- Updating the vendor to 2.1.0+
- Progressive enhancement from that point (e.g., replacing of the `Arg.conv'` calls this PR ads with 2.0's `Arg.Conv.make`, which would also allow us to specify completion handlers)

---

To verify, change the `vendor_cmdliner.sh` script to 
```sh
#!/bin/sh

set -euo pipefail

cd src/core/cmdliner
rm -rf *.ml *.mli dune

git clone https://github.com/dbuenzli/cmdliner tmp-vendor
git -C tmp-vendor switch --detach v2.1.0

mv tmp-vendor/src/*.{ml,mli} .
rm -rf tmp-vendor

mv cmdliner.ml opamCmdliner.ml
mv cmdliner.mli opamCmdliner.mli

cat > dune << EOF
(library
 (name opamCmdliner)
 (public_name opam-core.cmdliner)
 (flags :standard -w -27-32-35-50))
EOF
```

run it, and re-build.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -163,6 +163,7 @@ users)
   * Fix a rare potential GC corruption in `OpamStubs.uname` [#6880 @avsm @kit-ty-kate @andrew]
   * Fix a rare potential GC corruption in `OpamStubs.enumRegistry` on Windows [#6882 @kit-ty-kate]
   * Remove uses of `Stdlib.ignore` [#6481 @rjbou @kit-ty-kate]
+  * Source compatibility with newer `cmdliner` versions [#6854 @WardBrian]
 
 ## Internal: Unix
 
--- a/src/client/opamAdminCommand.ml
+++ b/src/client/opamAdminCommand.ml
@@ -1422,12 +1422,15 @@ let help =
     | None       -> `Help (`Pager, None)
     | Some topic ->
       let topics = "topics" :: cmds in
-      let conv, _ = OpamCmdliner.Arg.enum (List.rev_map (fun s -> (s, s)) topics) in
+      let conv =
+        OpamCmdliner.Arg.conv_parser
+          (OpamCmdliner.Arg.enum (List.rev_map (fun s -> (s, s)) topics))
+      in
       match conv topic with
-      | `Error e -> `Error (false, e)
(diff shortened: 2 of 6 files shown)


kit-ty-kate (approved):
Thanks a lot! After some minor reformatting to avoid unrelated changes and simplify the diff it looks great.

Note for the remaining reviewer (@rjbou): the github "Hide whitespace" setting makes the diff straightforward

WardBrian:
Thanks!

I wasn't sure if this kind of thing needed a CHANGES entry -- should I just ignore the CI run for that, or do you have a recommendation for what section I should put it under ("Internal"?)

rjbou (commented):
It worth having operators or small function name defined for `OpamCmdliner.Arg.conv*` in `OpamArgTools` and used everywhere.

kit-ty-kate:
> It worth having operators or small function name defined for `OpamCmdliner.Arg.conv*` in `OpamArgTools` and used everywhere.

what would be the point?

WardBrian:
> It worth having operators or small function name defined for `OpamCmdliner.Arg.conv*` in `OpamArgTools` and used everywhere.

Many of these are replaced in #6868 when completions are added

rjbou:
> > It worth having operators or small function name defined for `OpamCmdliner.Arg.conv*` in `OpamArgTools` and used everywhere.

> what would be the point?

to simplify the code, the use of these function. I find it simpler to write/read an converter operator and the main variable/function call, that a plain text wrapper.

> Many of these are replaced in #6868 when completions are added

Fair enough. I prefer operators/tiny functions name for this kind of usage, but it's a personal opinion, we'll see how it goes.

rjbou (approved):
thanks!

kit-ty-kate:
Ignoring windows CI failure (#7030).
Thanks!

PR #6858, opened 2026-02-12
Correct configure instruction in readme

gridbugs:


README.md
```diff
@@ -23,7 +23,7 @@ Either from an existing opam installation, use
   - GNU make
   - OCaml >= 4.08 (or see [below](#compiling-without-ocaml))
   - A C++ compiler (unless building without a solver, see `./configure --without-mccs`)
-* Run `./configure`. If you don't have the dependencies installed, this will
+* Run `./configure --with-vendored-deps`. If you don't have the dependencies installed, this will
```
kit-ty-kate: Something like this would be a bit more correct
```suggestion
* Run `./configure` if you have all the OCaml dependencies installed.
  If not, run `./configure --with-vendored-deps`, this will
```

kit-ty-kate (approved):

PR #6859, opened 2026-02-13
Update PATH in place in env hook

### LINKED ISSUE #6815: opam `env_hook` is too agressive, and makes the default install of opam behave invasively in the shell environment
## Context

The `env_hook` works like this:

https://github.com/ocaml/opam/blob/b5254e3e161a2313c3d279e57789ca6d2fdf179f/src/state/shellscripts/env_hook.sh#L1-L8

Prepending the `_opam_env_hook` to the [PROMPT_COMMAND](https://tldp.org/HOWTO/Bash-Prompt-HOWTO/x264.html), causes it to be invoked before every new command prompt. `opam env` always *prefixes* its `$HOME/.opam/default/bin` directory to the user's `PATH`. As a result, once users have the `env_hook` tied into their shell's rc file, it become impossible for users to prefix this PATH with any other directories. This happens "behind the backs" of unsuspecting users, since the only way to figure out what is happening is to dig thru a few layers of scripts.

## Problem

I say that the current behavior is "invasive" because (IMO) it infringes on the user's expectation that they can control their path, by aggressively dominating the prefix.

Here's an example of pathological behavior:

In my current switch, I have binary `$OPAM_SWITCH_PREFIX/bin/omd`. I want to define my own binary, so I do

```
$ mkdir mybin
$ cat > mybin/omd <<EOF
> #!/usr/bin/sh
> echo "Orchestral Manoeuvres in the Dark"
> EOF
$ chmod +x mybin/omd 
$ ./mybin/omd 
Orchestral Manoeuvres in the Dark
```

now, within my current shell session, I want to be able to have this binary in my path, so I run

```
$ export PATH=$HOME/mybin:$PATH
```

I should now expect I can invoke my new `omd` binary, and it will take precedence. But instead

```
$ omd --help | head
omd [options] [inputfile1 .. inputfileN] [options]
  -o                  file.html Specify the output file (default is stdout).
  --auto-identifiers  Should identifiers be automatically assigned to headings.
  --version           Display the version of the currently installed omd.
  --                  Consider all remaining arguments as input file names.
  -help               Display this list of options
  --help              Display this list of options
```

Because the `env_hook` has insisted on putting the opam binary dir first in my path:

```
$ echo $PATH
/home/me/.opam/default/bin:/home/me/mybin:<...>
```

This is quite unfortunate behavior, IMO, and it has led me to scrub all remnants of the current opam hooks from my rc files.

## Proposal

In general, I think we should try to find a way where we can maintain the convenience of automatic switch management without clobbering a users path.

AFAIU, this is the essential property we want the automatic switch management to guarantees, w/r/t binaries:

1. When a user first *activates* a switch (explicitly via  or implicitly by changing into a directory that has a local switch), the binaries installed in the switch should shadow any synonymous binaries that may be in the users environment.

The current implementation ensures this property, but it does so thru this much stronger property, which I think we *don't* want:

2. Whenever a user runs a command, the binaries installed in the active switch shadow any synonymous binaries that may be in the users environment.

The most similar tool I am familiar with is `direnv`. It does *not* have this problem. I have taken to using direnv to manage local switches by adding the following files to projects that use local switches:

```
$ cat .envrc 
eval $(opam env)
```

IIUC, `direnv` manages to avoid the problem of (2) by computing the diff between the current environment and environment entailed by it's `.envrc` file, and only updating the environment if the settings entailed by the `.envrc` file are not already in the environment.

In the context of opam switches, I think this would mean `opam env` would only prefix a a switche binary path if the current binary path is not already in the path. If it is already in the path, it should return the env unchanged.

I think this should ensure (1), because the switches directory *will* be prefixed a new switch is activated, but it will *not* be prefixed on command prompts that don't don't introduce a new switch.

gridbugs:
Prior to this change, the default behaviour of `opam init` was to register a shell hook which ran before each command, and added/moved the bin dir from the current opam switch to the beginning of the PATH variable. If a user has a custom bin directory at the beginning of their PATH containing an executable with the same name as an executable in their current opam switch, the executable in the opam switch will take precedence.

This commit changes the behaviour of the shell hook to update the existing opam switch in PATH in place rather than prepending it, so the precedence of the opam switch relative to other dirs in PATH is preserved.

This change doesn't affect the behaviour of running `eval $(opam env)`.

Fixes https://github.com/ocaml/opam/issues/6815

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -151,6 +151,7 @@ users)
 
 ## Shell
   * added return 0 for zsh/bash/sh/fish to prevent leaking error codes [#6930 @ta2005]
+  * Update PATH in place in env hook [#6859 @gridbugs fix #6815]
 
 ## Internal
   * Improve cache-loading performance when using OCaml >= 5.4 by using `Gc.ramp_up` [#6515 @dra27]
--- a/src/state/shellscripts/env_hook.csh
+++ b/src/state/shellscripts/env_hook.csh
@@ -1,1 +1,1 @@
-alias precmd 'eval `opam env --shell=csh --readonly`'
+alias precmd 'eval `opam env --shell=csh --readonly --inplace-path`'
--- a/src/state/shellscripts/env_hook.fish
+++ b/src/state/shellscripts/env_hook.fish
@@ -1,3 +1,3 @@
 function __opam_env_export_eval --on-event fish_prompt
-    eval (opam env --shell=fish --readonly 2> /dev/null)
+    eval (opam env --shell=fish --readonly --inplace-path 2> /dev/null)
 end
--- a/src/state/shellscripts/env_hook.sh
+++ b/src/state/shellscripts/env_hook.sh
@@ -1,6 +1,6 @@
 _opam_env_hook() {
  local previous_exit_status=$?;
- eval $(opam env --shell=bash --readonly 2> /dev/null <&- );
+ eval $(opam env --shell=bash --readonly --inplace-path 2> /dev/null <&- );
  return $previous_exit_status;
 };
 if ! [[ "$PROMPT_COMMAND" =~ _opam_env_hook ]]; then
--- a/src/state/shellscripts/env_hook.zsh
+++ b/src/state/shellscripts/env_hook.zsh
@@ -1,5 +1,5 @@
 _opam_env_hook() {
-    eval $(opam env --shell=zsh --readonly 2> /dev/null <&-);
+    eval $(opam env --shell=zsh --readonly --inplace-path 2> /dev/null <&-);
 }
 typeset -ag precmd_functions;
 if [[ -z ${precmd_functions[(r)_opam_env_hook]} ]]; then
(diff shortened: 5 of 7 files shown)


kit-ty-kate (approved):
I like that idea, thanks for bringing options that i didn't know about.

After testing it live, i think the documentation of the option is somewhat misleading as to its behaviour when adding another switch directory to the PATH manually. The documentation says that it "replaces any pre-existing opam path" but it rather "replaces any previously-modified PATH component by opam".

As noted by @rjbou during our meeting, our testsuite is missing test for this option and it would be nice to have some before the full release (this can be done separately from this PR at a later date).

Once merged, we should remember to display a prominent note so that people know about this change. In particular, if their `.bashrc/.zshrc` file contain further additions to PATH after the `source opam-init.sh` line, they will now be place before opam in the PATH. This can be a breaking change for some people but this is also a "bug fix" in some regard given that it normalises the shape of the PATH in interactive shell mode vs. not interactive mode if someone adds things to their PATH after sourcing opam.

rjbou (approved):

kit-ty-kate:
Ignoring the cygwin failure (#7030).

Thanks a lot!

PR #6862, opened 2026-03-01
specify Windows version for windows.h

Firobe:
As initially reported by @hannesm in https://github.com/tarides/opam-monorepo/issues/421, opam-monorepo fails to build in a MSYS2 environment because it vendors opam (in this case opam 2.3.0, but there has been no relevant change since).

Full logs at: https://github.com/ocaml/opam-repository/actions/runs/22346082152/job/64660957616?pr=29454

It seems `GetCurrentProcessToken` is [only available since Windows 8.1](https://learn.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-getcurrentprocesstoken), but MSYS2 defaults to Windows 7.

This PR sets `_WIN32_WINNT` to the [constant for Windows 8.1](https://learn.microsoft.com/en-us/cpp/porting/modifying-winver-and-win32-winnt?view=msvc-170).

My Windows knowledge is limited and I don't really have ways to test this easily, so don't hesitate to tell me I'm wrong :)

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -84,6 +84,7 @@ users)
 ## VCS
 
 ## Build
+  * Fix Windows build on MSYS2 [#6862 @Firobe]
   * opam no longer depends on `cmdliner` [#6755 @kit-ty-kate - fix #6425]
   * Clean variables before calling make on different projects to avoid clashes [#6769 @kit-ty-kate]
   * Add the upcoming OCaml 5.5 (trunk) support when using dune's dev profile [#6670 @kit-ty-kate]
--- a/src/core/dune
+++ b/src/core/dune
@@ -13,6 +13,7 @@
     (names opam_stubs)
     (flags :standard
            -DUNICODE -D_UNICODE -DCAML_NAME_SPACE
+           -D_WIN32_WINNT=0x602 ; Require Windows 8 / Server 2012 or later
            (:include c-flags.sexp)))
   (c_library_flags (:standard
                    (:include c-libraries.sexp)))


kit-ty-kate:
> but MSYS2 defaults to Windows 7

I'm a bit surprised by this. I've compiled opam on MSYS2 before and it built just fine.
Is that a property of MSYS2 as installed by GitHub (as opposed to as installed by hand)?

Firobe:
> > but MSYS2 defaults to Windows 7
> 
> I'm a bit surprised by this. I've compiled opam on MSYS2 before and it built just fine. Is that a property of MSYS2 as installed by GitHub (as opposed to as installed by hand)?

I'm not quite sure honestly. I'm basing this claim on their [changelog](https://www.msys2.org/docs/windows_support/#changelog) where the latest mention of `_WIN32_WINNT` (2022-12-26) mentions it still defaults to Windows 7 for the MinGW environments.

dra27:
Cross-referencing ocaml/ocaml#14484, please can it be done by adding `-D_WIN32_WINNT=0x602` in `configure.ac` instead of altering the C file:

```diff
diff --git a/configure.ac b/configure.ac
index c6d16872c..04d5ef07c 100644
--- a/configure.ac
+++ b/configure.ac
@@ -215,6 +215,8 @@ AS_IF([ test ${WIN32} -eq 1 -a "$GCC" = "yes"],[
 ])

 AS_IF([ test ${WIN32} -eq 1 ],[
+  # Require Windows 8 / Server 2012 or later
+  CONF_CFLAGS="$CONF_CFLAGS -D_WIN32_WINNT=0x602"
   AS_IF([ test "$GCC" = "yes"], [
     AS_IF([ test "x${CC64}" = "x" ], [
       AS_IF([ test "$ARCH" = "i386" ],
```

> 

> I've compiled opam on MSYS2 before and it built just fine.

Is that definitely the case? I get the same behaviour on my MSYS2 installation (and /mingw64/include/_mingw.h has that Windows 7 default still). There's also the 32/64-bit C compiler and so forth that has to be worked (as I still haven't got round to sorting that out properly \<sigh>...)

dra27:
Actually, that flag could be specified directly on L15 of src/core/dune as well, given that the `-DUNICODE` and `-D_UNICODE` Windows-specific ones are also there already.

Firobe:
> Actually, that flag could be specified directly on L15 of src/core/dune as well, given that the `-DUNICODE` and `-D_UNICODE` Windows-specific ones are also there already.

Done, thanks!

kit-ty-kate:
Thanks! That looks great. This should be part of the next 2.6.0~alpha1 release. Ideally we'd like to add MSYS2 to our CI system to make sure there isn't gonna be a regression for this. It might be a bit tricky so if we don't have time to do it before the release we'll merge this anyway so dw.

kit-ty-kate (approved):
Thanks! Merging as-is because we definitely won't have time to add MSYS2 before 2.6.0\~alpha1 but i've opened a TODO ticket as https://github.com/ocaml/opam/issues/6989

PR #6864, opened 2026-03-05
Reorder the list of actions by increased priority

### LINKED ISSUE #6863: Move removals at the end of the list of changes when installing/upgrading/removing/reinstalling a package
Since https://github.com/ocaml/opam/pull/5045, the list of action is ordered with removals first, then downgrades, reinstalls, upgrades and installs.

I feel like removals (and possibly downgrades) should be higher down on the list to avoid having to scroll up the whole list of action. For example when upgrading, the list of reinstalls will be quite large and hide the list of removals and downgrades.

I propose to reorder this the following way:
- reinstalls
- downgrades
- upgrades
- removals
- installs

kit-ty-kate:
Fixes #6863

### DIFF
--- a/src/client/opamSolution.mli
+++ b/src/client/opamSolution.mli
@@ -14,6 +14,8 @@
 open OpamTypes
 open OpamStateTypes
 
+type action_solution = user_action * OpamSolver.solution
+
 (** Resolve an user request *)
 val resolve:
   'a switch_state ->
@@ -21,7 +23,7 @@ val resolve:
   ?reinstall:package_set ->
   requested:package_set ->
   atom request ->
-  (OpamSolver.solution, OpamCudf.conflict) solver_result
+  (action_solution, OpamCudf.conflict) solver_result
 
 (** Apply a solution returned by the solver. If [ask] is not specified, prompts
     the user whenever the solution isn't obvious from the request. [add_roots]
@@ -40,7 +42,7 @@ val apply:
   ?assume_built:bool ->
   ?download_only:bool ->
   ?force_remove:bool ->
-  OpamSolver.solution ->
+  action_solution ->
   rw switch_state * solution_result
 
 (** Call the solver to get a solution and then call [apply]. If [ask] is not
--- a/src/client/opamSolution.mli
+++ b/src/client/opamSolution.mli
@@ -14,6 +14,8 @@
 open OpamTypes
 open OpamStateTypes
 
+type solution = user_action * OpamSolver.solution
+
 (** Resolve an user request *)
 val resolve:
   'a switch_state ->
@@ -21,7 +23,7 @@ val resolve:
   ?reinstall:package_set ->
   requested:package_set ->
   atom request ->
-  (OpamSolver.solution, OpamCudf.conflict) result
+  (solution, OpamCudf.conflict) result
 
 (** Apply a solution returned by the solver. If [ask] is not specified, prompts
     the user whenever the solution isn't obvious from the request. [add_roots]
@@ -40,7 +42,7 @@ val apply:
   ?assume_built:bool ->
   ?download_only:bool ->
   ?force_remove:bool ->
-  OpamSolver.solution ->
+  solution ->
   rw switch_state * solution_result
 
 (** Call the solver to get a solution and then call [apply]. If [ask] is not
--- a/src/client/opamSwitchCommand.ml
+++ b/src/client/opamSwitchCommand.ml
@@ -219,7 +219,7 @@ let install_compiler
       ~requested:(OpamPackage.packages_of_names t.packages roots)
       (OpamSolver.request ~install:additional_installs ())
   in
-  let solution = match solution with
+  let (_, solution) as action_solution = match solution with
     | Success s -> s
     | Conflicts cs ->
       OpamConsole.error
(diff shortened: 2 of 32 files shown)


src/solver/opamSolver.ml
```diff
@@ -770,10 +770,10 @@ let print_solution ~messages ~append ~requested ~reinstall ~available
         (match acts with [_] -> "package" | _ -> "packages");
       OpamConsole.print_table ~sep:" " stdout (List.map snd acts)
   in
-  print_actions (function `Remove _ -> true | _ -> false);
-  print_actions (function `Change (`Down,_,_) -> true | _ -> false);
   print_actions (function `Reinstall _ -> true | _ -> false);
+  print_actions (function `Change (`Down,_,_) -> true | _ -> false);
```
rjbou: Downgrades should be shown after upgrade (more important to notice)

src/solver/opamSolver.ml
```diff
@@ -770,10 +770,10 @@ let print_solution ~messages ~append ~requested ~reinstall ~available
         (match acts with [_] -> "package" | _ -> "packages");
       OpamConsole.print_table ~sep:" " stdout (List.map snd acts)
   in
-  print_actions (function `Remove _ -> true | _ -> false);
-  print_actions (function `Change (`Down,_,_) -> true | _ -> false);
   print_actions (function `Reinstall _ -> true | _ -> false);
```
rjbou, line 773: It worth adding a comment on how the order is set. Before it was a simple from remove to install, here it is by important change to highlight

rjbou (changes requested):

src/solver/opamSolver.ml
kit-ty-kate: i thought about it a bit more and i think it would be better to consider every single type of user actions (install, upgrade, removal) instead. They should all have different priorities. I've pushed a PoC doing just that. What do you think?

kit-ty-kate:
~TODO: add a test with every actions~

src/solver/opamSolver.ml
rjbou: Discussed in dev meeting: seems good, we can try to see if there is a better/cleaner way to do it.

src/solver/opamSolver.ml
kit-ty-kate: i think the code is self describing here. Adding a comment would only duplicate the code

kit-ty-kate:
Test added and rebased

src/solver/opamSolver.ml
rjbou: The code is self describing on what is printed in which order, not the idea or how it is chosen.
```suggestion
  (* We always print reinstall, and after it we select the order of actions
     given the actions. We want to highlight the important changes by putting
     them at the end if the list. *)
  print_actions (function `Reinstall _ -> true | _ -> false);
```

src/client/opamSwitchCommand.ml
```suggestion
  let action, solution = match solution with
```

src/client/opamSwitchCommand.ml
```diff
@@ -229,7 +229,7 @@ let install_compiler
            (OpamSwitchState.unavailable_reason t) cs);
       OpamStd.Sys.exit_because `No_solution
   in
-  let () = match OpamSolver.stats solution with
+  let () = match OpamSolver.stats (snd solution) with
```
```suggestion
  let () = match OpamSolver.stats solution with
```

src/client/opamSwitchCommand.ml
```diff
@@ -262,15 +262,15 @@ let install_compiler
...
+                             -- (OpamSolver.removed_packages (snd solution))
+                             ++ (OpamSolver.new_packages (snd solution)) }
     in
     { t with compiler_packages = base_comp }
   in
   let skip =
     if deps_only then
       let pkgs =
-        OpamPackage.packages_of_names (OpamSolver.new_packages solution)
+        OpamPackage.packages_of_names (OpamSolver.new_packages (snd solution))
```
```suggestion
        OpamPackage.packages_of_names (OpamSolver.new_packages solution)
```

src/client/opamSwitchCommand.ml
```diff
@@ -262,15 +262,15 @@ let install_compiler
     let base_comp =
       OpamSwitchState.compute_invariant_packages
         { t with installed = t.installed
-                             -- (OpamSolver.removed_packages solution)
-                             ++ (OpamSolver.new_packages solution) }
+                             -- (OpamSolver.removed_packages (snd solution))
+                             ++ (OpamSolver.new_packages (snd solution)) }
```
```suggestion
                             -- (OpamSolver.removed_packages solution)
                             ++ (OpamSolver.new_packages solution) }
```

src/client/opamSwitchCommand.ml
```suggestion
      (action, solution) in
```

src/client/opamSolution.mli
```diff
@@ -14,14 +14,16 @@
 open OpamTypes
 open OpamStateTypes
 
+type solution = user_action * OpamSolver.solution
```
rjbou:
```suggestion
type action_solution = user_action * OpamSolver.solution
```
proposal, the type is not just the solution itself, it contains the information of the action because we need it elsewhere (for printing)

kit-ty-kate:
done

rjbou (approved):
thanks!

kit-ty-kate:
Ignoring the random macOS failure (#7037)

PR #6866, opened 2026-03-12
OpamDarcs: remove Num patches fallback

toastal:
2.10.3 was released in 2016. It’s probably safe to remove the code after a decade of maturity.

There is a lot of room for improvement for Darcs tho—there is a big comment talking about Darcs deficiencies, but I don’t think a lot of it is true with my understanding of Darcs. Context files should be used more broadly, but this merge request is specifically a low-hanging change that purely eliminates code. I would like to look at bigger fixes in the nearer future.

Yes, Darcs is still used & is still an active project.

[^1]

[^1]: Please consider [giving up MS GitHub](https://sfconservancy.org/GiveUpGitHub/) or offering a non-proprietary, non-US-corporate-controlled mirror for this free software project. I wish to delete this Microsoft account in the future, but I need more projects like this to support alternative methods to send patches & contribute.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -82,6 +82,7 @@ users)
   * Allow the macOS sandbox to write in the `/var/folders/` and `/var/db/mds/` directories as it is required by some of macOS core tools [#4797 @kit-ty-kate - fix #4389 #6460]
 
 ## VCS
+  * Darcs no longer fall back to “Num Patches” as “Weak Hash” has been supported since 2016 [#6866 @toastal]
 
 ## Build
   * opam no longer depends on `cmdliner` [#6755 @kit-ty-kate - fix #6425]
--- a/src/repository/opamDarcs.ml
+++ b/src/repository/opamDarcs.ml
@@ -98,7 +98,6 @@ module VCS = struct
   let patch_applied _ _ = Done ()
 
   let revision repo_root =
-    (* 'Weak hash' is only supported from 2.10.3, so provide a fallback *)
     darcs repo_root [ "show"; "repo" ] @@> fun r ->
     OpamSystem.raise_on_process_error r;
     try
@@ -108,16 +107,6 @@ module VCS = struct
            | Some (label, value)
              when OpamStd.String.contains ~sub:"Weak Hash" label ->
              Some (Done (Some value))
-           | _ -> None)
-        r.OpamProcess.r_stdout
-    with Not_found ->
-    try
-      OpamStd.List.find_map
-        (fun s ->
-           match OpamStd.String.rcut_at s ' ' with
-           | Some (label, value)
-             when OpamStd.String.contains ~sub:"Num Patches" label ->
-             Some (Done (Some (Printf.sprintf "darcs-%s" value)))
            | _ -> None)
         r.OpamProcess.r_stdout
     with Not_found ->


rjbou (approved):
Thanks for this PR! We don't use/know well darcs, if you want to improve its opam support, your PRs are very welcome.

toastal:
@rjbou I will slowly take a look at these since I think there was some misunderstandings, but it could be a chunky (or even unfeasible) refactor. The “context file” actually holds the entire stable state for a Darcs repository (no “state” hash like Pijul), but this is, as it says on the tin, a *file* not a short string like a hash.

In 2025–2026, I had done a lot of work with Nix + Darcs in both Nixpkgs’s `fetchDarcs`, `nix-prefetch-darcs`, as well as in building Nixtamal so these models are still at the front of my brain.

toastal:
@rjbou

> ```ocaml
> (* Darcs has no branches, no proper diff, no way to reset, can't return a
>    workdir diff including added/removed files... That makes it hard for
>    handling as a remote, and the following is a bit convoluted. *)
> ```

I think the root of this comment/misunderstanding as that the `module VCS` abstraction is wrong as was designed in a manner with the assumption that all VCSs behave like Git & would be compatible with Git’s model. This explains why the comment has sentiment like everything is being [shoehorned](https://en.wikipedia.org/wiki/Shoehorn#Turn_of_phrase) or that all sort of features are missing/wrong. Darcs is a complete VCS as is (tho it would be nice to see channels (branches) like Pijul). If this scaled to Fossil, Pijul, Bazaar, & so on, it would become clearer where the design issues are.

The “fix” would be non-trivial as VCS kinda needs to be thrown out to clearly abstract over this—I had originally tried a unified design too which I had to throw out (with [this file](https://darcs.toastal.in.th/nixtamal/next/lib/input.ml) being the approach I ended up taking). For instance Darcs:

* has no revision, it uses a context_file (file with a unordered collection of patches to get to the end state)
* uses a weak hash just as a quick assert of the patch state such as the context file applying properly
* just uses new folders instead of channels/branches (this also means that a Darcs repo doesn’t need a server/daemon to host a repository, you just put an HTTP server in front of a Darcs folder very simply)

Additionally with regards to that comment & “convoluted”:

* `darcs revert` can reset
* `darcs diff` works
* `darcs whatsnew …` can show the added/dropped files 

---

This is all an aside compared to just eliminating very old code in this merge request.

kit-ty-kate:
Thanks for this contribution. As you may have seen this part of the code is quite ancient, and as it turns out, fairly untested. Ideally we'd like to add some tests for darcs before merging this. We'll do that if we have some time but regardless this should be fine to merge before 2.6.0~alpha1.

> 1. Please consider [giving up MS GitHub](https://sfconservancy.org/GiveUpGitHub/) or offering a non-proprietary, non-US-corporate-controlled mirror for this free software project. I wish to delete this Microsoft account in the future, but I need more projects like this to support alternative methods to send patches & contribute.

Thanks for bringing this up, i've been thinking about it myself for a while. I've been slowly moving to codeberg for my personal projects but there are a couple more considerations for larger and central projects like opam and it might take more time. I've added that item to discuss at our next dev meeting though so that we can at least start a discussion.

kit-ty-kate:
> Thanks for this contribution. As you may have seen this part of the code is quite ancient, and as it turns out, fairly untested. Ideally we'd like to add some tests for darcs before merging this. We'll do that if we have some time but regardless this should be fine to merge before 2.6.0~alpha1.

We indeed didn't have time to do that so let's merge now and do the tests later (https://github.com/ocaml/opam/issues/7019). Thanks!

> > 1. Please consider [giving up MS GitHub](https://sfconservancy.org/GiveUpGitHub/) or offering a non-proprietary, non-US-corporate-controlled mirror for this free software project. I wish to delete this Microsoft account in the future, but I need more projects like this to support alternative methods to send patches & contribute.
> 
> Thanks for bringing this up, i've been thinking about it myself for a while. I've been slowly moving to codeberg for my personal projects but there are a couple more considerations for larger and central projects like opam and it might take more time. I've added that item to discuss at our next dev meeting though so that we can at least start a discussion.

As mentioned a few days ago on discuss[^1], we've opened a yet-to-be-properly-advertised mirror on [codeberg.org/ocaml/opam](https://codeberg.org/ocaml/opam) as a first step.

[^1]: https://discuss.ocaml.org/t/opam-pin-add-error-i-dont-know-about-package-xxx-passed-through-only-packages/18321/6

PR #6867, opened 2026-03-16
Update cmdliner vendor to 2.1.1

### LINKED ISSUE #684: OPAM does not handle dependencies anymore
git-version: 7669e885b1ce1bea3c46be8cb1a79fabbcf7b712

OPAM just do not install dependencies anymore, I tried fd-send-recv and utop.

### LINKED ISSUE #6849: Internal OpamCmdliner lagging behind, missing fixes in upstream cmdliner

WardBrian:
Builds on #6854

- Updates vendored cmdliner to 2.1.1 (closes #6849)
- Updated the `vendor_cmdliner.sh` script to work off patches stored in the repository, rather than manual renaming/file catting. 
- Patches the legacy prefix behavior to be unconditionally true, as mentioned in https://github.com/ocaml/opam/issues/6849#issuecomment-3802583384
- ~Includes the patch for https://github.com/dbuenzli/cmdliner/issues/262, which will be applicable to Opam's own completion~ (Included in 2.1.1).

To review, I recommend:
- reading the contents of `src/core/cmdliner/patches`
- re-running `vendor_cmdliner.sh` to observe that it re-creates the rest of the PR with no diff.


dbuenzli:
I'd suggest to update to 2.1.1 instead as it has your fix for working completion with legacy prefixes.

WardBrian:
@dbuenzli would you mind adding a v2.1.1 tag to the cmdliner repository?

dbuenzli:
Forgot to push the release commit. Done.

PR #6873, opened 2026-03-27
Nushell support

jneem:
This is a continuation of #6330, adding nushell support.

As in #6330, nushell support for `opam env` works via json (so the command to populate the environment is `load-env (opam env | to json)`).

This also adds shell config support, but doesn't yet setup any hooks.

### DIFF
--- a/src/client/opamConfigCommand.ml
+++ b/src/client/opamConfigCommand.ml
@@ -131,6 +131,14 @@ let print_sexp_env output env =
   aux env;
   output ")\n"
 
+let rec print_nu_env output env =
+  match env with
+    | [] -> ()
+    | (k,v,_)::r -> begin
+      Printf.ksprintf output "\t\"%s\" : \"%s\",\n" k v;
+      print_nu_env output r
+    end
+
 let rec print_fish_env output env =
   let set_arr_cmd ?(modf=fun x -> x) k v =
     let v = modf @@ OpamStd.String.split v ':' in
@@ -183,7 +191,7 @@ let print_without_cr s =
     output_string stdout s;
     flush stdout
 
-let print_eval_env ~csh ~sexp ~fish ~pwsh ~cmd env =
+let print_eval_env ~csh ~sexp ~fish ~pwsh ~cmd ~nu env =
   let env = (env : OpamTypes.env :> (string * string * string option) list) in
   let output_normally = OpamConsole.msg "%s" in
   let never_with_cr =
@@ -202,6 +210,11 @@ let print_eval_env ~csh ~sexp ~fish ~pwsh ~cmd env =
     print_pwsh_env output_normally env
   else if cmd then
     print_cmd_env output_normally env
+  else if nu then begin
+    Printf.ksprintf never_with_cr "{\n";
+    print_nu_env never_with_cr env;
+    Printf.ksprintf never_with_cr "}"
+  end
   else
     print_env never_with_cr env
 
@@ -331,7 +344,7 @@ let ensure_env gt switch =
   ignore (ensure_env_aux gt switch)
 
 let env gt switch ?(set_opamroot=false) ?(set_opamswitch=false)
-    ~csh ~sexp ~fish ~pwsh ~cmd ~inplace_path =
+    ~csh ~sexp ~fish ~pwsh ~cmd ~nu ~inplace_path =
   log "config-env";
   let opamroot_not_current =
     let current = gt.root in
@@ -371,7 +384,7 @@ let env gt switch ?(set_opamroot=false) ?(set_opamswitch=false)
   let env =
     ensure_env_aux ~set_opamroot ~set_opamswitch ~force_path gt switch
   in
-  print_eval_env ~csh ~sexp ~fish ~pwsh ~cmd env
+  print_eval_env ~csh ~sexp ~fish ~pwsh ~cmd ~nu env
 [@@ocaml.warning "-16"]
 
 let subst gt fs =
--- a/src/client/opamConfigCommand.ml
+++ b/src/client/opamConfigCommand.ml
@@ -131,13 +131,9 @@ let print_sexp_env output env =
   aux env;
   output ")\n"
 
-let rec print_nu_env output env =
-  match env with
-    | [] -> ()
-    | (k,v,_)::r -> begin
-      Printf.ksprintf output "\t\"%s\" : \"%s\",\n" k v;
-      print_nu_env output r
-    end
+let print_nu_env output env =
+  let json = `O (List.map (fun (k,v,_) -> (k, `String v)) env) in
+  output @@ OpamJson.to_string json
 
 let rec print_fish_env output env =
   let set_arr_cmd ?(modf=fun x -> x) k v =
@@ -211,9 +207,7 @@ let print_eval_env ~csh ~sexp ~fish ~pwsh ~cmd ~nu env =
   else if cmd then
     print_cmd_env output_normally env
   else if nu then begin
-    Printf.ksprintf never_with_cr "{\n";
     print_nu_env never_with_cr env;
-    Printf.ksprintf never_with_cr "}"
   end
   else
     print_env never_with_cr env
--- a/src/core/opamStd.ml
+++ b/src/core/opamStd.ml
@@ -936,7 +936,7 @@ module OpamSys = struct
     fun () -> Lazy.force os
 
   type powershell_host = Powershell_pwsh | Powershell
-  type shell = SH_sh | SH_bash | SH_zsh | SH_csh | SH_fish
+  type shell = SH_sh | SH_bash | SH_zsh | SH_csh | SH_fish | SH_nu
     | SH_pwsh of powershell_host | SH_cmd
 
   let all_shells =
@@ -944,6 +944,7 @@ module OpamSys = struct
      SH_zsh;
      SH_csh;
      SH_fish;
+     SH_nu;
      SH_pwsh Powershell_pwsh;
      SH_pwsh Powershell;
      SH_cmd]
@@ -958,6 +959,7 @@ module OpamSys = struct
     | "zsh"  -> Some SH_zsh
     | "bash" -> Some SH_bash
     | "fish" -> Some SH_fish
+    | "nu"   -> Some SH_nu
     | "pwsh" -> Some (SH_pwsh Powershell_pwsh)
     | "dash"
     | "sh"   -> Some SH_sh
@@ -1078,6 +1080,8 @@ module OpamSys = struct
     match shell with
     | SH_fish ->
       Some (List.fold_left Filename.concat (home ".config") ["fish"; "config.fish"])
+    | SH_nu ->
+      Some (List.fold_left Filename.concat (home ".config") ["nushell"; "env.nu"])
     | SH_zsh  ->
       let zsh_home f =
         try Filename.concat (Env.get "ZDOTDIR") f
--- a/src/core/opamStd.ml
+++ b/src/core/opamStd.ml
@@ -1081,7 +1081,11 @@ module OpamSys = struct
     | SH_fish ->
       Some (List.fold_left Filename.concat (home ".config") ["fish"; "config.fish"])
     | SH_nu ->
-      Some (List.fold_left Filename.concat (home ".config") ["nushell"; "env.nu"])
+      let config_home =
+        try Env.get "XDG_CONFIG_HOME"
+        with Not_found -> home ".config"
+      in
+      Some (List.fold_left Filename.concat config_home ["nushell"; "autoload"; "opam.nu"])
     | SH_zsh  ->
       let zsh_home f =
         try Filename.concat (Env.get "ZDOTDIR") f
--- a/src/core/opamStd.ml
+++ b/src/core/opamStd.ml
@@ -1081,9 +1081,13 @@ module OpamSys = struct
     | SH_fish ->
       Some (List.fold_left Filename.concat (home ".config") ["fish"; "config.fish"])
     | SH_nu ->
-      let config_home =
-        try Env.get "XDG_CONFIG_HOME"
-        with Not_found -> home ".config"
+      (* nu checks XDG_CONFIG_HOME on all operating systems,
+         and then falls back to the system-dependent "dirs" crate *)
+      let config_home = try Env.get "XDG_CONFIG_HOME"
+        with Not_found -> match os () with
+        | Darwin -> home "Library/Application Support"
+        | Win32 -> (try Env.get "APPDATA" with Not_found -> home ".config")
+        | _ -> home ".config"
       in
       Some (List.fold_left Filename.concat config_home ["nushell"; "autoload"; "opam.nu"])
     | SH_zsh  ->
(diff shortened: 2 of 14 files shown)


src/client/opamConfigCommand.ml
```diff
@@ -202,6 +210,11 @@ let print_eval_env ~csh ~sexp ~fish ~pwsh ~cmd env =
     print_pwsh_env output_normally env
   else if cmd then
     print_cmd_env output_normally env
+  else if nu then begin
+    Printf.ksprintf never_with_cr "{\n";
+    print_nu_env never_with_cr env;
+    Printf.ksprintf never_with_cr "}"
```
kit-ty-kate: the braces should be part of print_nu_env similar to what is done for print_sexp_env

src/client/opamConfigCommand.ml
```diff
@@ -131,6 +131,14 @@ let print_sexp_env output env =
   aux env;
   output ")\n"
 
+let rec print_nu_env output env =
+  match env with
+    | [] -> ()
+    | (k,v,_)::r -> begin
+      Printf.ksprintf output "\t\"%s\" : \"%s\",\n" k v;
```
kit-ty-kate: `OpamJson.to_string` should be used instead to avoid escaping issues etc, as well as factorize code that actually work on the json format

src/core/opamStd.ml
```diff
@@ -1078,6 +1080,12 @@ module OpamSys = struct
     match shell with
     | SH_fish ->
       Some (List.fold_left Filename.concat (home ".config") ["fish"; "config.fish"])
+    | SH_nu ->
+      let config_home =
+        try Env.get "XDG_CONFIG_HOME"
+        with Not_found -> home ".config"
+      in
+      Some (List.fold_left Filename.concat config_home ["nushell"; "autoload"; "opam.nu"])
```
kit-ty-kate: the default config location seems to be a bit more complicated than that. In particular, the default `$HOME/.config/...` is only true on Linux, not macOS or Windows (see https://www.nushell.sh/book/configuration.html#configuration-overview)

src/state/opamEnv.ml
```diff
@@ -879,6 +881,8 @@ let opam_env_invocation ?root ?switch ?(set_opamswitch=false) shell =
       Printf.sprintf " \"--%s=%s\"" argname
     | SH_sh | SH_bash | SH_zsh | SH_csh | SH_fish ->
       Printf.sprintf " '--%s=%s'" argname
+    | SH_nu ->
+      Printf.sprintf " '--%s=%s'" argname
```
kit-ty-kate: this can be merged with the two lines just before

src/state/opamEnv.ml
```diff
@@ -954,21 +959,22 @@ let init_file = function
...
+  | SH_csh | SH_fish | SH_nu -> None
   | SH_pwsh _ | SH_cmd -> None
 
 let env_hook_script_base = function
   | SH_sh | SH_bash -> Some OpamScript.env_hook
   | SH_zsh -> Some OpamScript.env_hook_zsh
   | SH_csh -> Some OpamScript.env_hook_csh
   | SH_fish -> Some OpamScript.env_hook_fish
-  | SH_pwsh _ | SH_cmd -> None
+  | SH_pwsh _ | SH_cmd | SH_nu -> None
```
kit-ty-kate: this seems possible using https://www.nushell.sh/book/hooks.html#pre-execution-hooks

src/state/opamEnv.ml
```diff
@@ -1008,6 +1014,9 @@ let export_in_shell ~unconditional shell =
       Printf.sprintf "%sset -gx %s %s;\n"
         (make_comment comment) k v
   in
+  let nu (k,v,comment) =
+      Printf.sprintf "$env.%s = %s %s\n" k v (make_comment comment)
```
kit-ty-kate, line 1018: there seems to be a problem here. The function returns a function that takes `k * v * comment` in which `v` is expected to be quoted (or not) with the general rules from posix shell.

However nushell seems to require that parameter to be quoted. For example when this function is called by `env_hook_script`, `true` isn't quoted as will simply make nushell error out.

I think the function needs to be changed in a way where the quoting is done inside export in shell instead of by the callers of the function (`string_of_update` and `env_hook_script`). This might require some kind of intermediate datatype though i'm not sure yet.

kit-ty-kate: funny thing:
```
$env.TEST = true
```
would actually work, but instead of setting the environment this is actually a boolean test (like in OCaml)

src/state/opamEnv.ml
```diff
@@ -1044,6 +1054,8 @@ let abort_if_set var shell =
       var var
   | SH_fish ->
     Printf.sprintf "test -z \"$%s\"; or return\n" var
+  | SH_nu ->
+    Printf.sprintf "if ($env.%s? | default \"\") != \"\" { return }\n" var
```
kit-ty-kate, line 1062: Could you point out the documentation for the `$var?` syntax? I'm unable to find and local testing only throw errors out

src/state/opamEnv.ml
```diff
@@ -1062,6 +1074,8 @@ let source root shell f =
   | SH_fish ->
     let fname = unix_transform ~using_backslashes:true () in
     Printf.sprintf "test -r '%s' && source '%s' > /dev/null 2> /dev/null; or true\n" fname fname
+  | SH_nu ->
+    Printf.sprintf "const source_path = if ('%s' | path exists) { '%s' } else { null }\nsource $source_path\n" fname fname
```
kit-ty-kate: it feels like it could be shorten to
```suggestion
    Printf.sprintf "if ('%s' | path exists) { source '%s' }\n" fname fname
```

src/state/opamEnv.ml
```diff
@@ -1097,6 +1111,8 @@ let if_interactive_script shell t e =
     Printf.sprintf "if ( $?prompt ) then\n  %s%sendif\n" t @@ ielse e
   | SH_fish ->
     Printf.sprintf "if status is-interactive\n  %s%send\n" t @@ ielse e
+  | SH_nu ->
+    Printf.sprintf "if $nu.is_interactive {\n  %s%s}\n" t @@ ielse_pwsh e
```
kit-ty-kate: the function name should probably be change and the variable is `is-interactive` not `is_interactive` (per https://www.nushell.sh/book/special_variables.html#nu)
```suggestion
    Printf.sprintf "if $nu.is-interactive {\n  %s%s}\n" t @@ ielse_pwsh_nu e
```

src/state/opamEnv.ml
```diff
@@ -1143,11 +1159,13 @@ let string_of_update st shell updates =
            Printf.sprintf "'%s%c' + \"$env:%s\""
              (OpamStd.Env.escape_powershell string) sep envu_var
          | SH_cmd -> Printf.sprintf "%s%c%%%s%%" string sep envu_var
+         | SH_nu -> Printf.sprintf "$env.%s? | default [] | prepend '%s'" envu_var string
```
kit-ty-kate, line 1167: same question as above for `$var?`

src/state/opamEnv.ml
```diff
@@ -1143,11 +1159,13 @@ let string_of_update st shell updates =
...
          | _ -> Printf.sprintf "'%s':\"$%s\"" string envu_var)
       | EqColon | EqPlus ->
         (match shell with
          | SH_pwsh _ -> Printf.sprintf "\"$env:%s\" + '%c%s'" envu_var sep string
          | SH_cmd -> Printf.sprintf "%%%s%%%c%s" envu_var sep string
+         | SH_nu -> Printf.sprintf "$env.%s? | default [] | append '%s'" envu_var string
```
kit-ty-kate, line 1173: same here

kit-ty-kate:
oops i hit enter a bit too fast. Thanks a lot for your contribution, would you have some spare time to look at the issues highlighted above. I think that once they are all fixed this should probably be in a mergeable state

src/state/opamEnv.ml
jneem: It isn't `$var?` on it's own: it's specific to accessing elements of records, like in `$foo.var?`. Docs here: https://www.nushell.sh/book/environment.html#reading-environment-variables

kit-ty-kate: ah thanks for the pointer

src/state/opamEnv.ml
jneem: It feels like it, but no... for reasons I don't really understand, the `source "dep.nu"` in

```nu
if ("dep.nu" | path exists) { source "dep.nu" }
```

is evaluated unconditionally at parse time, while the "if" condition is delayed to run time. *But*, if you write it as

```nu
const source_path = if ("dep.nu" | path exists) { "dep.nu"}
source $source_path
```

then the "if" condition is promoted to parse time. It does look like I can remove the "else null" part, though.

src/state/opamEnv.ml
jneem: Was the `ielse_pwsh` -> `ielse_pwsh_nu` change intentional? I think they have the same syntax here.

src/state/opamEnv.ml
jneem: I'm a bit confused here, because it seems to be acting like an assignment for me:

```
❯ $env.TEST = true
❯ $env.TEST
true
❯ $env.TEST == true
true
```

jneem:
Thanks for the fast response! I should be able to do another iteration over the weekend.

src/state/opamEnv.ml
kit-ty-kate: yes, but as far as i understand it becomes some sort of internal variable. If you do `env | grep TEST` afterwards the variable doesn't appear (but it does if the argument is quoted)

src/state/opamEnv.ml
kit-ty-kate: yes the function name is misleading if we start using it for both powershell and nushell

src/state/opamEnv.ml
jneem: huh, it shows up for me. Starting from a fresh shell:

```
❯ $env.TEST = true
❯ env | grep TEST
TEST=true
```

This is on nushell 0.111.0 on linux. Could it be a platform difference?

src/state/opamEnv.ml
kit-ty-kate: ah thanks, this is explained in https://www.nushell.sh/book/thinking_in_nu.html#think-of-nushell-as-a-compiled-language

src/state/opamEnv.ml
kit-ty-kate: i'm testing using 0.99.1/linux

kit-ty-kate: regardless of the special case of `true`, what does `$env.TEST = blah` give you? With my version it's a parse error

jneem: For me too. Along with a little help message:

```
  help: the parsing of assignments was changed in
        0.97.0, and this would have previously been
        treated as a string. Alternatively, quote the
        string with single or double quotes to avoid it
        being interpreted as a command name. This
        restriction may be removed in a future release.
```

src/state/opamEnv.ml
jneem: Done

src/core/opamStd.ml
jneem: I took a stab at improving it, but it's untested on macOS and Windows. For background, nushell uses the [dirs](https://crates.io/crates/dirs) crate as a fallback if `XDG_CONFIG_HOME` is unset. It's behaviour is documented [here](https://docs.rs/dirs/latest/dirs/fn.config_dir.html). On windows, it actually uses [win32 bindings](https://github.com/dirs-dev/dirs-sys-rs/blob/8bcd4aa2c35990d57a2cff2953793525fc42709c/src/lib.rs#L150), but some random googling suggested that maybe the [APPDATA](https://github.com/putridparrot/EnvironmentVariableCheatsheet) env var would also work...

src/state/opamEnv.ml
jneem: Not yet done, and I've run out of time for today. I'll re-request a review when it's ready for another look.

jneem: Ok, done now.

src/state/opamEnv.ml
jneem: I had a go at moving the quoting into `export_in_shell`, but I had some trouble because `string_of_update` generates shell code that's tricky to quote. Instead, I just added the quoting to `env_hook_script`

rjbou:
Thanks for the (follow-up) PR. We discussed it in dev meeting: we need to take into account the fact that nushell is prone to change:
- make sure that the added commands are part of the baseline of most of versions
- add a minimal version check when enabled/selected
- add a note that informs users that their specific versions could not be supported, and a hint on how to disable/fallback on another shell.
- update (at least) `tests/reftests/init-scripts.unix.test` with the new nushell files

Otherwise, the current content is quite ok (_quite_ bc it needs to be reviewed).

jneem:
> make sure that the added commands are part of the baseline of most of versions

Do you have any guidance for how to determine "most" here? Nushell maintains a [packaging status table](https://www.nushell.sh/book/installation.html#package-managers), but it isn't completely obvious to me where to draw the line. For example, I often use Debian's packaged version to determine the oldest version that's worth supporting, but apparently Debian doesn't package nushell...

kit-ty-kate:
That's a good question, ideally from that table, i would say "replace debian by \<fairly used distribution with the oldest version of nushell\>", which in this case is Fedora (0.99.1).

Sadly in this case, it probably complicates things a bit too much, because `source null` apparently wasn't a thing before 0.102: https://www.nushell.sh/blog/2025-02-04-nushell_0_102_0.html#source-null-use-null-toc

So if that's not possible, i'd say probably 0.104 (latest nushell version in Alpine since 3.22) is a good starting point.

jneem:
I spent some time trying to add a minimal version check to the autoload script, with the goal of producing an actionable error when invoked with an old `nu`. I didn't succeed, though, because the error is raised during parse time. For example, sourcing the script

```nu
if (version | get major) == 0 and (version | get minor) < 102 {
  return
}

const s = if ("dep.nu" | path exists) { "dep.nu" }
source $s
```

with nu 0.99 gives the error

```
Error: nu::parser::type_mismatch

  × Type mismatch.
   ╭─[/home/<me>/test.nu:6:8]
 5 │ const s = if ("dep.nu" | path exists) { "dep.nu" }
 6 │ source $s
   ·        ─┬
   ·         ╰── expected string, found nothing
   ╰────
```

The version guard has no effect, because the error occurs before anything is run. I tried a few hacks to get the version guard to run at parse time, but without success (in 0.99, not even the `get` command can run at parse time -- that appears to have changed sometime between 0.99 and 0.111).

So maybe the best course of action is to hold off on this until 0.102 is old enough that we can assume everyone uses it. In the meantime, direnv and `eval $(opam env)` give a pretty reasonable nushell-compatible setup.

kit-ty-kate:
what about simply parsing `nu --version` in `OpamClient.init_checks`? (you just need to pass `shell` to it).

The check would be done only once at init time but i don't expect people to start downgrading their version of nushell, and if so i think they would probably expect something to break.

kit-ty-kate:
Moving to the 2.7 milestone as we won't have time to properly review it in the remaining time before the 2.6.0~alpha1 release

PR #6874, opened 2026-03-29
Follow symlinks transparently in get_diff

### LINKED ISSUE #6410: Opam update fails when repository already exists
It seems like when using the setup-ocaml action, sometimes we get this error:
```
  [default] synchronised from git+https://github.com/ocaml/opam-repository.git
  [ERROR] Could not update repository "default": Cannot remove /root/.opam/repo/default (/__t/opam/2.3.0/x86_64/opam: "unlink" failed on /root/.opam/repo/default: Is a directory).
```
It's not immediately obvious to me what the issue is, but we are running on GHA, and this is after a cache hit has been restored.

smorimoto:
## Problem

When a repository is configured with a `file://` URL pointing to a local directory containing symlinks, `opam update` fails with:

```
[ERROR] Could not update repository "default": Symlinks are unsupported
```

This occurs because `get_diff` in `opamRepositoryBackend.ml` uses `Unix.lstat` to determine file types. `lstat` does not follow symlinks, so any symlink is reported as `S_LNK` and immediately rejected with `failwith "Symlinks are unsupported"`.

The issue was encountered in the opam-repository CI on Windows/Cygwin (ocaml/opam-repository#29613), where the workflow uses `file://${{ github.workspace }}` as the repository URL. The opam-repository checkout contains symlinks (used for package version aliases), and the `file://` backend's local copy handler (`OpamLocal.fetch_repo_update`) preserves them in the quarantine directory. When `get_diff` then compares the quarantine against the existing repository state, it encounters these symlinks and fails.

This is related to #6410, which reports a similar `opam update` failure after a GitHub Actions cache restore with `file://` pinned packages.

### Why `git+file://` works but `file://` does not

Switching the repository URL from `file://` to `git+file://` avoids the problem entirely. The `git+file://` URL uses the git backend (`OpamVCS`), which computes diffs via `git diff` — this never calls `get_diff`. The `file://` URL uses the rsync/local backend (`OpamLocal`), which copies the directory to a quarantine and then calls `get_diff` to compare old and new states. Only this latter code path encounters the `S_LNK` rejection. Users should not need to work around this by changing URL schemes; the `file://` backend ought to handle symlinks properly.

### Why this affects more than just Windows

Although the original report was on Windows, the `"Symlinks are unsupported"` error comes from `get_diff` in `opamRepositoryBackend.ml`, which is platform-independent code. Any `file://` repository containing symlinks triggers the same failure on Linux and macOS. The `git+file://` workaround happens to bypass `get_diff` entirely, which is why it appears to be a Windows-only issue.

## Solution

Replace `Unix.lstat` with `Unix.stat` in the `get_diff` function. `Unix.stat` follows symlinks transparently: a symlink to a regular file is seen as `S_REG`, and a symlink to a directory is seen as `S_DIR`. The diff then operates on the logical content of the repository rather than its filesystem metadata.

### Edge cases handled

**Dangling symlinks** (`ENOENT`): When `Unix.stat` follows a symlink whose target does not exist, it raises `Unix_error(ENOENT, ...)`. This is caught in `stat_opt` and mapped to `None`, treating the entry as absent. The corresponding file path is also set to `None` so that `readfile` is not called on the dangling symlink (which would raise `File_not_found` from `OpamSystem.read`).

**Circular symlinks** (`ELOOP`): When `Unix.stat` encounters a symlink loop, it raises `Unix_error(ELOOP, ...)`. This is caught alongside `ENOENT` for the same treatment.

**Other errors** (e.g. `EACCES`): These are intentionally *not* caught — they represent genuine failures that should propagate as exceptions. This is consistent with the existing behaviour of the old `lstat_opt`, which also did not catch any `Unix_error`.

### Match arm changes

- The `S_LNK` arm is retained for pattern exhaustiveness but changed from `failwith` to `assert false`. `Unix.stat` never returns `S_LNK`, so this arm is unreachable. Using `assert false` (rather than silently skipping) ensures a clear, debuggable failure if this invariant is ever violated.

- The `None, None` arm is changed from `assert false` to `diffs` (skip). Previously, `get_files_for_diff` always produced pairs with at least one `Some` side, and `lstat` never returned `None` for an existing directory entry, so `None, None` was genuinely unreachable. With `stat`, both sides can become `None` when both entries are dangling symlinks. Skipping is correct: there is nothing to diff when neither side has accessible content.

### Test changes

- **Symlink tests** (tests 4–5): Now produce proper diffs instead of `ERROR: Symlinks are unsupported`.
- **Dangling symlink tests** (tests 6–8): Three new test cases exercise the `ENOENT` code path — dangling on first side (treated as file creation), dangling on second side (treated as file deletion), and dangling on both sides (no diff).
- **Test isolation fix**: The shared `linked_file` target is now unconditionally rewritten in `write_setup`. Previously, the existence check (`if not (OpamFilename.exists f)`) meant the file was only written once; when test 4's patch application modified it through the symlink, subsequent tests inherited the corrupted content. This was masked before because symlink tests always errored out before reaching the patch step.
- **Path normalisation**: Added `rm_tmpdir` to replace the platform-specific temporary directory path with `$TMPDIR` in error messages, ensuring the expected output is stable across operating systems.

Fixes #6410

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -63,6 +63,7 @@ users)
 ## Lint
 
 ## Repository
+  * Fix `opam update` failure with `file://` repositories containing symlinks by following symlinks transparently in the internal diff computation [#6874 @smorimoto - fix #6410]
 
 ## Lock
 
@@ -133,6 +134,8 @@ users)
 ## Internal: Windows
 
 ## Test
+  * Add dangling symlink test cases for the internal diff computation [#6874 @smorimoto]
+  * Fix test isolation issue where symlink target file was shared across patchDiff tests [#6874 @smorimoto]
 
 ## Benchmarks
   * Add an even larger real-world diff to benchmark `opam update` [#6567 @kit-ty-kate]
--- a/src/repository/opamRepositoryBackend.ml
+++ b/src/repository/opamRepositoryBackend.ml
@@ -138,11 +138,12 @@ let get_diff parent_dir dir1 dir2 =
     in
     (file, OpamSystem.read real_file)
   in
-  let lstat_opt parent_dir = function
+  let stat_opt parent_dir = function
     | None -> None
     | Some file ->
       let file = Filename.concat (OpamFilename.Dir.to_string parent_dir) file in
-      Some (Unix.lstat file)
+      try Some (Unix.stat file)
+      with Unix.Unix_error ((Unix.ENOENT | Unix.ELOOP), _, _) -> None
   in
   let rec aux diffs dir1 dir2 =
     let files = get_files_for_diff parent_dir dir1 dir2 in
@@ -153,7 +154,13 @@ let get_diff parent_dir dir1 dir2 =
             | None -> diffs
             | Some diff -> diff :: diffs
           in
-          match lstat_opt parent_dir file1, lstat_opt parent_dir file2 with
+          let stat1 = stat_opt parent_dir file1 in
+          let stat2 = stat_opt parent_dir file2 in
+          (* When stat fails (e.g. dangling symlink), treat the file
+             as absent for reading and recursion too. *)
+          let file1 = if Option.is_none stat1 then None else file1 in
+          let file2 = if Option.is_none stat2 then None else file2 in
+          match stat1, stat2 with
           | Some {st_kind = S_REG; _}, None
           | None, Some {st_kind = S_REG; _}
           | Some {st_kind = S_REG; _}, Some {st_kind = S_REG; _} ->
@@ -168,7 +175,8 @@ let get_diff parent_dir dir1 dir2 =
           | Some {st_kind = S_REG; _}, Some {st_kind = S_DIR; _} ->
             failwith "Change from a regular file to a directory is unsupported"
           | Some {st_kind = S_LNK; _}, _ | _, Some {st_kind = S_LNK; _} ->
-            failwith "Symlinks are unsupported"
+            (* Unreachable: Unix.stat resolves symlinks *)
+            assert false
           | Some {st_kind = S_CHR; _}, _ | _, Some {st_kind = S_CHR; _} ->
             failwith "Character devices are unsupported"
           | Some {st_kind = S_BLK; _}, _ | _, Some {st_kind = S_BLK; _} ->
@@ -177,7 +185,7 @@ let get_diff parent_dir dir1 dir2 =
             failwith "Named pipes are unsupported"
           | Some {st_kind = S_SOCK; _}, _ | _, Some {st_kind = S_SOCK; _} ->
             failwith "Sockets are unsupported"
-          | None, None -> assert false)
+          | None, None -> diffs)
         diffs files
     in
     diffs
--- a/src/repository/opamRepositoryBackend.mli
+++ b/src/repository/opamRepositoryBackend.mli
@@ -112,9 +112,13 @@ val job_text:
     subdirs of [parent_dir], returns None if they are equal, and the
     corresponding patch and the list of file-changes otherwise.
 
+    Symlinks are resolved transparently: they are followed to their targets.
+    Dangling symlinks (whose targets do not exist) and circular symlinks
+    are silently skipped.
+
     @raise Stdlib.Failure if an unsupported file type or comparison is
     detected in any of [subdir1] or [subdir2].
-    Unsupported file types: symlinks, character devices, block devices,
-    named pipes, sockets.
+    Unsupported file types: character devices, block devices, named pipes,
+    sockets.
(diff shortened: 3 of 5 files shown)


kit-ty-kate:
Symlinks are not supported on purpose (too many special cases), what is the reason setup-ocaml use them instead of simply copying if the use is package aliases?

As a side-question, i'm sorry to ask this, but given how verbose the PR description etc is, have you used an llm to write this?

smorimoto:
@kit-ty-kate I see, that makes sense. So basically, does that mean using `git+file://` is recommended? When I use `file://`, `opam update` fails, and this is quite a confusing issue for people who aren’t familiar with the "opam codebase" (not opam itself), myself included.

As for the LLM point, that’s half correct. Recently, I’ve been frequently doing translations between Japanese and English using an LLM that I fine-tuned myself. This issue seemed far too complex to get merged without sufficient context.

I’m aware that the OCaml community isn’t very keen on using tools like Claude Code or Codex, so I’ve deliberately avoided using it here (and in any case, the change wasn’t large enough to need it).

rjbou:
Thanks for the PR. We discussed it in dev meeting and there is several points to highlight. First, it was intended to not support symlinks in repository since 2.4 (see [release blog post](https://opam.ocaml.org/blog/opam-2-4-0/#Other-noteworthy-changes)), it originates with the use of `ocaml-patch` instead of the `patch`. If it is reintroduced, we need to check that the library permit to handle it. `diff` handles symlinks as it plain files (it show the linked file change), with option to not follow symlinks, it only tells that the link changed while `git diff` show the content of the link (the linked path).
<details> <summary>diff link handling</summary>

```shell
# SETUP
> ls -l *
a:
total 8
-rw-rw-r-- 1 u u 4 Mar 31 13:27 file1
-rw-rw-r-- 1 u u 5 Mar 31 13:25 file2
lrwxrwxrwx 1 u u 5 Mar 31 13:26 lnk1 -> file1
lrwxrwxrwx 1 u u 5 Mar 31 13:26 lnk2 -> file1

b:
total 8
-rw-rw-r-- 1 u u 4 Mar 31 13:19 file1
-rw-rw-r-- 1 u u 5 Mar 31 13:25 file2
lrwxrwxrwx 1 u u 5 Mar 31 13:26 lnk1 -> file1
lrwxrwxrwx 1 u u 5 Mar 31 13:26 lnk2 -> file2
> batcat */file*
───────┬────────────────────────────────────
       │ File: a/file1
───────┼────────────────────────────────────
   1   │ foo
───────┴────────────────────────────────────
───────┬────────────────────────────────────
       │ File: a/file2
───────┼────────────────────────────────────
   1   │ foo2
───────┴────────────────────────────────────
───────┬────────────────────────────────────
       │ File: b/file1
───────┼────────────────────────────────────
   1   │ bar
───────┴────────────────────────────────────
───────┬────────────────────────────────────
       │ File: b/file2
───────┼────────────────────────────────────
   1   │ foo2
───────┴────────────────────────────────────
# DIFF
> diff a b
diff a/file1 b/file1
1c1
< foo
---
> bar
diff a/lnk1 b/lnk1
1c1
< foo
---
> bar
diff a/lnk2 b/lnk2
1c1
< foo
---
> foo2
# DIFF without following symlinks
> diff --no-dereference a b
diff --no-dereference a/file1 b/file1
1c1
< foo
---
> bar
Symbolic links ‘a/lnk2’ -> ‘file1’ and ‘b/lnk2’ -> ‘file2’ differ
# GIT DIFF
> git --no-pager diff
diff --git a/a/file1 b/b/file1
index 257cc56..5716ca5 100644
--- a/a/file1
+++ b/b/file1
@@ -1 +1 @@
-foo
+bar
diff --git a/a/lnk2 b/b/lnk2
index 08219db..30d67d4 120000
--- a/a/lnk2
+++ b/b/lnk2
@@ -1 +1 @@
-file1
\ No newline at end of file
+file2
\ No newline at end of file
```
</details>
Except how it is handled strictly via the diff, there is repository consistency concerns. On the links themselves, we need to ensure that it is contained in the repository (no absolute path, no escape). On the repository side, it is not recommended to link opam files (it leads to silent unwanted changes), but it is quite safe to have link on extra files (there is the checksum safeguard in opam file).

It worth also adding a cram test for linked files in repository. `PatchDiff` test tests only the opam lib functions of diff/patch, while cram test checks opam workflows: repository consistency and its use in opam (for example extra files retrieving when a package is installed), opam admin, etc. We'll need to add linked files in repositories tests.

Handling symlinks in repositories, with the good safeguards, could be a good to have. However, keep in mind that this piece of code is being rewritten (without symlinks support): we are currently working on #6625 that will change significantly the same pieces of code (diff, repo, extra files, etc.).

PR #6878, opened 2026-04-08
Dependency lower bounds spring cleanup

kit-ty-kate:
A minor spring cleaning adding missing lower-bounds and a very handy script for people who like me have no native way of installing autoconf 2.71.
Reading the commit descriptions should hopefully be sufficient to understand the intent of each one.


shell/docker-configure-autogen.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+
```
rjbou: Add some "doc", something like ? 
```suggestion
# Helpers to regenerate 'configure' with the good version of autoconf
# make sure to have committed your changes in 'configure.ac' before launching it
# Does nor work on worktrees
```
maybe check that there is no remaining `configure.ac` uncommitted changes ? maybe too much for this simple script ?

rjbou (approved):
On standardisation of opam files depends (commit #4), what was used ? it is not alphabetic order, how is it sorted ?

kit-ty-kate:
> On standardisation of opam files depends (commit #4), what was used ? it is not alphabetic order, how is it sorted ?

as mentioned in the extended commit message, the order is:
- ocaml
- dune
- opam libraries
- anything else (unordered, same order as before)

rjbou:
indeed, i didn't understand that the list was the order...

kit-ty-kate:
CI failures (alpine and gentoo) will be fixed by #6888 and #6886, ignoring.

PR #6880, opened 2026-04-12
uname: register roots for intermediate strings to avoid GC corruption

avsm:
uname: register roots for intermediate strings to avoid GC corruption

Reported by @andrew
Backported to 2.5 in https://github.com/ocaml/opam/pull/6893

### DIFF
--- a/src/core/opamUnix.c
+++ b/src/core/opamUnix.c
@@ -22,8 +22,9 @@ CAMLprim value opam_stdout_ws_col(value _unit) {
 #include <sys/utsname.h>
 
 CAMLprim value opam_uname(value _unit) {
+  CAMLparam0();
+  CAMLlocal1(ret);
   struct utsname buf;
-  value ret;
 
   if (-1 == uname(&buf)) {
     caml_uerror("uname", Nothing);
@@ -33,5 +34,5 @@ CAMLprim value opam_uname(value _unit) {
   Store_field(ret, 1, caml_copy_string(buf.release));
   Store_field(ret, 2, caml_copy_string(buf.machine));
 
-  return ret;
+  CAMLreturn(ret);
 }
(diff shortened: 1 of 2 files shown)


src/core/opamUnix.c
```diff
@@ -22,16 +22,20 @@ CAMLprim value opam_stdout_ws_col(value _unit) {
 #include <sys/utsname.h>
 
 CAMLprim value opam_uname(value _unit) {
+  CAMLparam1(_unit);
```
kit-ty-kate: wouldn't `CAMLparam0();` be sufficient here since it's unused?

src/core/opamUnix.c
```diff
@@ -22,16 +22,20 @@ CAMLprim value opam_stdout_ws_col(value _unit) {
...
+  CAMLlocal4(v_ret, v_sysname, v_release, v_machine);
```
kit-ty-kate: very minor, but i feel like the `v_` prefix change here isn't necessary

kit-ty-kate (approved):
LGTM, thanks! I have a couple of very minor comments but the core is of course good

kit-ty-kate:
actually i'm curious, why add new intermediate `v_sysname` etc. variables? I understand why `ret` needs to be registered but i'm unsure why the intermediate ones need to be as no other allocation is present on the same line. Is this to prevent some kind of reordering that would somehow be allowed by the C standard or something?

cc @gasche

kit-ty-kate:
I've chosen to simplify the change and only register the actually required GC root (`ret`). The resulting patch was verified by @octachron.

Unless someone raises a veto in the meantime, i'll merge this in a couple hours.

dra27 (approved):
Both versions absolutely fine - indeed, the problem with the original code was just that `ret` could move if any of the `caml_copy_string` calls resulted in a GC, the actual copied strings went straight into a block, so no issue.

kit-ty-kate:
Rebased above master. Merging once CI is done.

kit-ty-kate:
Thanks a lot!

(ignoring the ocaml-benchmarks failure, opam update failed with error code 40 seems unrelated)

avsm:
Just catching up after some travels; mine was just conservative as each caml_copy_string could go into the GC, but only if someone else modified the code later.  Thanks for the simpler revision.

PR #6882, opened 2026-04-14
Fix a GC corruption on Windows in enumRegistry

kit-ty-kate:
sub_key was used after the allocation, thus sub_key can in rare cases be moved elsewhere and its value becomes invalid. Simply moving the allocation after the last use of any non-immediate values fixes this problem.

Backported to 2.5 in https://github.com/ocaml/opam/pull/6892


dra27 (approved):
Thanks for preparing this one! I'm not sure whether I started `OPAMW_RegEnumValue` by copying and then mis-adapting `OPAMW_ReadRegistry` or was just being distinctly daft, but `sub_key` is now clearly _not_ read after the first allocation in the function.

PR #6883, opened 2026-04-14
Standardise allocating c stubs

kit-ty-kate:
I'm unsure if this would actually help prevent future GC corruptions but i'm proposing it just in case. What do you think @dra27 

The rule of the standardisation is:
- any functions that allocate anything have to register every local values and parameters (with the exception of unused parameters)

Queued on top of #6882

### DIFF
--- a/src/core/opamWindows.c
+++ b/src/core/opamWindows.c
@@ -431,11 +431,6 @@ CAMLprim value OPAMW_RegEnumValue(value hKey, value sub_key, value value_type)
 
   LPWSTR lpEnvironment;
 
-  result = caml_alloc_small(2, 0);
-  Field(result, 0) = Val_int(0);    /* Unused */
-  Field(result, 1) = Val_emptylist; /* The actual result */
-  tail = result;
-
   HKEY key;
   DWORD type;
   LSTATUS ret;
@@ -452,6 +447,11 @@ CAMLprim value OPAMW_RegEnumValue(value hKey, value sub_key, value value_type)
   if (!(lpSubKey = caml_stat_strdup_to_utf16(String_val(sub_key)))) {
     caml_raise_out_of_memory();
   }
+
+  result = caml_alloc_small(2, 0);
+  Field(result, 0) = Val_int(0);    /* Unused */
+  Field(result, 1) = Val_emptylist; /* The actual result */
+  tail = result;
 
   ret = RegOpenKey(roots[Int_val(hKey)], lpSubKey, &key);
   if (ret == ERROR_SUCCESS)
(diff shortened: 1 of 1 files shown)


src/core/opamWindows.c
```diff
@@ -95,14 +95,13 @@ CAMLprim value OPAMW_GetStdHandle(value nStdHandle)
...
+  CAMLreturn(result);
 }
 
 CAMLprim value OPAMW_GetConsoleScreenBufferInfo(value hConsoleOutput)
 {
-  CAMLparam0();
-  CAMLlocal1(result);
-  value coord;
+  CAMLparam1(hConsoleOutput);
+  CAMLlocal1(result, coord);
```
dra27, line 104:
```suggestion
  CAMLlocal2(result, coord);
```

src/core/opamWindows.c
```diff
@@ -191,16 +190,17 @@ CAMLprim value OPAMW_SetConsoleMode(value hConsoleMode, value dwMode)
   return Val_bool(result);
 }
 
-CAMLprim value OPAMW_GetWindowsVersion(value unit)
+CAMLprim value OPAMW_GetWindowsVersion(value _unit)
 {
-  value result;
+  CAMLlocal1(result);
```
dra27, line 195:
```suggestion
  CAMLparam0();
  CAMLlocal1(result);
```

src/core/opamWindows.c
```diff
@@ -747,13 +746,14 @@ CAMLprim value OPAMW_process_putenv(value pid, value key, value val)
...
+  CAMLparam0();
+  CAMLlocal1(result);
   PWSTR path = NULL;
-  value result;
 
   if (SUCCEEDED(SHGetKnownFolderPath(rfid, 0, NULL, &path))) {
     result = caml_copy_string_of_utf16(path);
     CoTaskMemFree(path);
-    return result;
+    CAMLreturn(result);
```
dra27, line 756: `CAMLreturn` has to go at the end of the function

src/core/opamWindows.c
```diff
@@ -777,7 +777,9 @@ CAMLprim value OPAMW_SendMessageTimeout(value vhWnd,
                                         value vwParam,
                                         value vlParam)
 {
-  value result;
+  CAMLparam5(vhWnd, uTimeout, fuFlags, vmsg, vwParam);
+  CAMLparam1(vlParam);
```
dra27, line 781:
```suggestion
  CAMLxparam1(vlParam);
```

dra27 (commented):
There's part of me which prefers it to be that the roots are registered only when it's really necessary... but in the absence of static checking that this is correct, and given that I don't think any of these functions are even close to being performance-sensitive to the extra roots, I think it's safer this way, indeed!

(I don't expect you were planning to, but I wouldn't push this patch out to older releases - the stubs in question do all look fine)

PR #6889, opened 2026-04-14
GHA: speedup macOS builds by stopping testing alternative solvers on macOS

kit-ty-kate:


rjbou (approved):
yup, there is no need to test specifically on macos. thanks!

kit-ty-kate:
Ignoring unfinished ocaml-benchmark (waiting for workers)

PR #6894, opened 2026-04-15
nix depexts: load from json instead of inlining package names

kit-ty-kate:
Fixed and tested by @RyanGibb 

The following pattern was allowed:
```
depexts: [
  ["gmp ]; <arbitrary nix expression>"] {os-distribution = "nixos"}
```

The json values can include strings trying to escape the json, but the `map (name: buildPackages.${name})` ensures any value will either be correct or make nix-build fail which is a reasonable behaviour. Nix syntax does not work like bash and thus ensure that nothing can escape the `${name}`.

Backported to 2.5 in https://github.com/ocaml/opam/pull/6895

### DIFF
--- a/src/state/opamSysInteract.ml
+++ b/src/state/opamSysInteract.ml
@@ -1474,11 +1474,22 @@ let install_packages_commands_t ?(env=OpamVariable.Map.empty) ~to_show st
          OpamFilename.create dir
            (OpamFilename.Base.of_string "env.nix")
        in
-       let packages =
-         String.concat " "
-           (OpamSysPkg.Set.fold (fun p l -> OpamSysPkg.to_string p :: l)
-              OpamSysPkg.Set.Op.(sys_packages.ti_new ++ sys_packages.ti_required) [])
+       let packageFile =
+         OpamFilename.create dir
+           (OpamFilename.Base.of_string "nix-depexts.json")
        in
+       (* We create a JSON object to avoid passing this kind of depexts
+          depexts: [
+            ["gmp ]; <arbitrary nix expression>"] {os-distribution = "nixos"}
+       *)
+       let packages =
+         "[" ^
+         String.concat ", "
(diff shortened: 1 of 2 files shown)


src/state/opamSysInteract.ml
```diff
@@ -1474,11 +1474,18 @@ let install_packages_commands_t ?(env=OpamVariable.Map.empty) ~to_show st
...
+       in
        let packages =
-         String.concat " "
-           (OpamSysPkg.Set.fold (fun p l -> OpamSysPkg.to_string p :: l)
-              OpamSysPkg.Set.Op.(sys_packages.ti_new ++ sys_packages.ti_required) [])
+         "[" ^
+         String.concat ", "
+           (OpamSysPkg.Set.fold (fun p l -> ("\"" ^ OpamSysPkg.to_string p ^ "\"") :: l)
+           OpamSysPkg.Set.Op.(sys_packages.ti_new ++ sys_packages.ti_required) [])
+           ^ "]"
```
```suggestion
              OpamSysPkg.Set.Op.(sys_packages.ti_new ++ sys_packages.ti_required) [])
         ^ "]"
```

src/state/opamSysInteract.ml
```diff
@@ -1474,11 +1474,18 @@ let install_packages_commands_t ?(env=OpamVariable.Map.empty) ~to_show st
          OpamFilename.create dir
            (OpamFilename.Base.of_string "env.nix")
        in
+       let packageFile =
+         OpamFilename.create dir
+           (OpamFilename.Base.of_string "nix-depexts.json")
+       in
        let packages =
```
rjbou, line 1485: (* 

```suggestion
       (* We create a JSON object to avoid passing this kind of depexts
          depexts: [
            ["gmp ]; <arbitrary nix expression>"] {os-distribution = "nixos"}
       *)
       let packages =
```

rjbou (approved):
LGTM!
Tested locally, and it changes well the derivation parsing error to a less harmful json parsing error.

PR #6896, opened 2026-04-15
OpamFilename.might_escape ~sep:Unspecified: Fix escape detection on Windows

kit-ty-kate:
This function with this particular parameter is unused in the code currently so this is only an issue for library users at the moment.


rjbou (commented):
Tha change itself is lgtm, it worth having a test, but i don't have an idea on which one

rjbou:
Integrated in #6897, closing

PR #6905, opened 2026-04-16
GHA depexts improvements

kit-ty-kate:
This speeds up the depexts CI by quite a bit (it doesn't need to compile ocaml everytime), allows us to stay up-to-date with whatever the distributions do and puts on the same level rolling-releases and version-based distributions.

The only distribution that does not have a `latest` docker tag is oraclelinux because they have purposefully removed it (i guess their main user-base want stability)


rjbou (approved):
lgtm! thanks

rjbou:
ocaml-benchmarks external failure, merging.

PR #6906, opened 2026-04-16
Remove the build directory as soon as possible when installing a package

### LINKED ISSUE #5884: opam install should cleanup the build directory right after a successful install instead of waiting until the end of the command
When installing certain number of packages from opam-repository, the size of the `$OPAMROOT/<switch>/.opam-switch/build/` directory can become very large by the end of the command as each package's tarballs get extracted before they are built.

For example, when building `async.v0.16.0`, the directory reaches 2.8GB.
This is an issue for tools such as [opam-health-check-ng](https://github.com/kit-ty-kate/opam-health-check-ng), for which a common technique to improve its performance is to mount `/var/lib/docker/` into `tmpfs`, which moves `$OPAMROOT` into RAM. RAM can be limited on some system and keeping all the build directories until the end makes the RAM usage much bigger than it should.

Code-wise, I believe that this code should be moved further up:
https://github.com/ocaml/opam/blob/eafdc8187676f4669d5e702317a2c5bb79225019/src/client/opamSolution.ml#L849

As a side-note, on Windows removing all the directories all at once at the end can take some time. If the removal were to be parallelized while other packages are being installed, it would be a good boost to performance.

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/5884
~Queued on https://github.com/ocaml/opam/pull/6912~

### DIFF
--- a/src/client/opamSolution.ml
+++ b/src/client/opamSolution.ml
@@ -683,6 +683,14 @@ let parallel_apply t
        | Left conf ->
          add_to_install nv conf;
          store_time ();
+         if not OpamClientConfig.(!r.keep_build_dir) then begin
+           (* NOTE: build_dir as defined above is the
+              "build dir in case of --inplace" so != OpamPath.Switch.build *)
+           let build_dir =
+             OpamPath.Switch.build t.switch_global.root t.switch nv
+           in
+           OpamFilename.rmdir build_dir;
+         end;
          Done (`Successful (OpamPackage.Set.add nv installed, removed))
        | Right exn ->
          store_time ();
@@ -864,17 +872,13 @@ let parallel_apply t
   end;
 
   let cleanup_artefacts graph =
+    (* NOTE: build dir removal is done directly after install
+       instead of here, for performance reasons. *)
     PackageActionGraph.iter_vertex (function
         | `Remove nv ->
           OpamAction.cleanup_package_artefacts t nv
           (* if reinstalled, only removes build dir *)
-        | `Install nv ->
-          if not OpamClientConfig.(!r.keep_build_dir) then
-            let build_dir =
-              OpamPath.Switch.build t.switch_global.root t.switch nv
-            in
-            OpamFilename.rmdir build_dir
-        | `Build _ | `Fetch _ -> ()
+        | `Build _ | `Fetch _ | `Install _ -> ()
         | `Change _ | `Reinstall _  -> assert false)
       graph
   in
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
Tested over 40_000 times with great success together with https://github.com/ocaml/opam/pull/6440. Max disk + RAM usage goes from 4-10GB down to < 2GB per install.

src/client/opamSolution.ml
```diff
@@ -868,13 +876,7 @@ let parallel_apply t
         | `Remove nv ->
           OpamAction.cleanup_package_artefacts t nv
           (* if reinstalled, only removes build dir *)
```
rjbou, line 880: This comment is tied to which line ?

src/client/opamSolution.ml
rjbou, line 877:
```suggestion
    (* Install cleanup, i.e. build dir removal, is done directly after install for performance. *)
    PackageActionGraph.iter_vertex (function
```

rjbou (commented):
_Génial!_ lgtm on the idea.
tiny comment + it worth adding a bench if possible.

src/client/opamSolution.ml
kit-ty-kate: the one above. See b6ed206a1a76871df856850a375e83e827ff3051

src/client/opamSolution.ml
kit-ty-kate: added with a couple of changes

kit-ty-kate:
> tiny comment + it worth adding a bench if possible.

done. I've added the bench in https://github.com/ocaml/opam/pull/6912 and rebased this PR on top of it to check the result.

kit-ty-kate:
Compared to baseline (https://github.com/ocaml/opam/pull/6912), the benchmark shows a 0.3% performance degradation of regular installs and a 5% performance improvement on IO-constrained filesystems (simulated as much as i could)

rjbou (approved):
We need a final check on the bench to rerun once rebased, and if it confirms it should be ok (there is no reason that the gain changes)

rjbou:
> Compared to baseline (#6912), the benchmark shows a 0.3% performance degradation of regular installs

What is the reason of this degradation ?

kit-ty-kate:
> What is the reason of this degradation ?

rm of the build dir are sequential and block the next install when installing fast packages to build so it's very slightly slower when the builds are too fast

kit-ty-kate:
Ignoring the build failure on cygwin (https://github.com/ocaml/opam/pull/7030)

> Compared to baseline (#6912), the benchmark shows a 0.3% performance degradation of regular installs and a 5% performance improvement on IO-constrained filesystems (simulated as much as i could)

Latest diff is -0.1% and -1.3%, well within the margin of fluctuation. The main improvement is RAM usage over time anyway, not the performance improvement.

Issue #3171, opened 2018-01-12
too many system calls!? (cwd evaluated during update)


hannesm:
the scenario I had was as follows (using opam 1.2.2):
- `opam update` in the working directory `/tmp/foo`
- `rm -rf /tmp/foo` in a different terminal

I was surprised that during the update (while opam was downloading), opam complained about "no such file or directory" -- which I then discovered was related to the current working directory which got lost.

--> why does opam need to investigate its cwd (and check existance) while it is downloading things to `~/.opam`?

AltGr:
Probably not the best approach, but the reason boils down to this:
- many operations rely on `chdir` (here it's probably command execution, but even directory normalisation does)
- opam always attempts to replace itself in its original cwd, in case it may need to lookup stuff from there. The issue is that it fails if the directory does not exist. I've already caught this in a few places, but not everywhere it seems.

PR #6910, opened 2026-04-24
Remove unnecessary use of chdir

### LINKED ISSUE #6600: Some commands aren't resistent to missing source directory of pinned packages
As presented in https://github.com/ocaml/opam/pull/6597, `opam pin list` currently fails in the following manner when the source directory isn't present:
```
$ opam install .
[...]
Would you like to install? [Y/n] n
$ opam pin
/usr/local/bin/opam: "chdir" failed on .../.opam-switch/sources/<pkg>: No such file or directory
```
I suspect other opam commands also fail in a similar manner while trying to get the git revision or similar info.

This specific `chdir` issue might be better fixed in the future by https://github.com/ocaml/opam/pull/5966/commits/69b447859a5c64802f5bc84eb5ecfb8ffc58b389 from https://github.com/ocaml/opam/pull/5966 (related to https://github.com/ocaml/opam/issues/3171, https://github.com/ocaml/opam/issues/2442, https://github.com/ocaml/opam/issues/969, …)

NathanReb:
This ports the `chdir` removal from https://github.com/ocaml/opam/pull/5966 as asked for in https://github.com/ocaml/opam/issues/6600.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -15,6 +15,7 @@ users)
 
 ## Global CLI
   * Update Kate's email address [#6808 @kit-ty-kate]
+  * Remove unnecessary uses of `chdir` [#6910 @NathanReb]
 
 ## Plugins
 
@@ -247,6 +248,7 @@ users)
   * `OpamPathName` was added [#6917 @rjbou]
   * `OpamRepositoryPathName` was added [#6917 @rjbou]
   * `OpamRepositoryPath` was moved from `opam-repository` [#6917 @rjbou]
+  * `OpamFilter.expand_interpolations_in_file`: changed argument type from `basename` to `filename` [#6910 @NathanReb]
 
 ## opam-core
   * `OpamCmdliner` was added. It is accessible through a new `opam-core.cmdliner` sub-library [#6755 @kit-ty-kate]
@@ -263,3 +265,8 @@ users)
   * `OpamFilename`: add `is_dir_read_only` [#6489 @rjbou]
   * `OpamFilename.might_escape`: ensure / is detected as a file separator when called with `~sep:Unspecified` on Windows [#6897 @kit-ty-kate]
   * `OpamFilename.Unix` was added abstracting over `/` separated paths regardless of the current system [#6914 @rjbou @kit-ty-kate]
+  * `OpamFilename.in_dir`: removed [#6910 @NathanReb]
+  * `OpamSystem.in_tmp_dir`: removed [#6910 @NathanReb]
+  * `OpamSystem.in_dir`: removed [#6910 @NathanReb]
+  * `OpamSystem.chdir`: removed [#6910 @NathanReb]
+  * `OpamSystem.{command,commands,read_command_output}`: add a `?dir: dirname` optional arg to launch the command in a specific directory [#6910 @NathanReb]
--- a/src/client/opamConfigCommand.ml
+++ b/src/client/opamConfigCommand.ml
@@ -377,8 +377,11 @@ let env gt switch ?(set_opamroot=false) ?(set_opamswitch=false)
 let subst gt fs =
   log "config-substitute";
   OpamSwitchState.with_ `Lock_none gt @@ fun st ->
+  let env = OpamPackageVar.resolve st in
   List.iter
-    (OpamFilter.expand_interpolations_in_file (OpamPackageVar.resolve st))
+    (fun b ->
+       let file = OpamFilename.of_basename b in
+       OpamFilter.expand_interpolations_in_file env file)
     fs
 
 let expand gt str =
--- a/src/format/opamFilter.ml
+++ b/src/format/opamFilter.ml
@@ -461,10 +461,9 @@ let expand_interpolations_in_file_full env ~src ~dst =
   close_out oc
 
 (* Substitute the file contents *)
-let expand_interpolations_in_file env file =
-  let file = OpamFilename.of_basename file in
-  let src = OpamFilename.add_extension file OpamPathName.subst_ext in
-  expand_interpolations_in_file_full env ~src ~dst:file
+let expand_interpolations_in_file env dst =
+  let src = OpamFilename.add_extension dst OpamPathName.subst_ext in
+  expand_interpolations_in_file_full env ~src ~dst
 
 (* Apply filters and interpolations to package commands *)
 
--- a/src/format/opamFilter.mli
+++ b/src/format/opamFilter.mli
@@ -143,12 +143,12 @@ val expand_interpolations_in_file_full: env -> src:filename -> dst:filename -> u
 (** Same as {!expand_interpolations_in_file} but allows to set the source [src] and
     destination [dst] files independently instead of implying [src] = [dst].in *)
 
-(** Rewrites [basename.in] to [basename], expanding interpolations.
+(** Rewrites [filename.in] to [filename], expanding interpolations.
     If the first line begins ["opam-version:"], assumes that expansion of
     variables within strings should be properly escaped. In particular, this
     means that Windows paths should expand correctly when generating .config
     files. *)
-val expand_interpolations_in_file: env -> basename -> unit
+val expand_interpolations_in_file: env -> filename -> unit
 
 
 (** Processes filters evaluation in a command list: parameter expansion and
--- a/src/repository/opamHTTP.ml
+++ b/src/repository/opamHTTP.ml
@@ -94,13 +94,24 @@ end
 (* Helper functions used by opam-admin *)
 
 let make_index_tar_gz repo_root =
-  OpamFilename.in_dir repo_root (fun () ->
-    let to_include = [
-      "version";
-      OpamRepositoryPathName.packages_d;
-      OpamRepositoryPathName.repo_f;
+  let open OpamFilename.Op in
+  let files_to_include = [
+    "version";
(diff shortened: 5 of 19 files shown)


src/format/opamFilter.ml
```diff
@@ -461,10 +461,14 @@ let expand_interpolations_in_file_full env ~src ~dst =
   close_out oc
 
 (* Substitute the file contents *)
-let expand_interpolations_in_file env file =
-  let file = OpamFilename.of_basename file in
-  let src = OpamFilename.add_extension file "in" in
-  expand_interpolations_in_file_full env ~src ~dst:file
+let expand_interpolations_in_file ?dir env file =
```
kit-ty-kate: Are there places where `?dir:None`? If not, i think it would be better to simply replace by
```suggestion
let expand_interpolations_in_file ~dir env file =
```

kit-ty-kate (commented):
The tests need to be adapted to the new API (tests/lib/patcher uses chdir)

kit-ty-kate:
On top of #6600, this should hopefully also fix https://github.com/ocaml/opam/issues/3171

src/format/opamFilter.ml
NathanReb: There's one in `OpamConfigCommand` [here](https://github.com/ocaml/opam/blob/master/src/client/opamConfigCommand.ml#L381).

NathanReb: note to self: make it a non optional labeled arg: `dir: dirname option ->`

NathanReb:
The test failures seem to all come from `opam-rt` builds. I guess it needs to be patched as well.

This could make us reconsider the removal of the `chdir` dependant API entries as this constitutes a breaking API change.

I could simply reintroduce them now that we know they're not used internally anymore but I assume the goal is to discourage their use entirely in opam so removing them feels right.

Let me know what you'd prefer here. I wonder if we could use some neat trick to deprecate them internally without the warning hitting external users, this would make for a good in between.

Edit: I forgot to hit `Comment`. I saw your PR on but still wanted to open the discussion on the breaking API change.

src/format/opamFilter.mli
```diff
@@ -143,12 +143,13 @@ val expand_interpolations_in_file_full: env -> src:filename -> dst:filename -> u
...
-(** Rewrites [basename.in] to [basename], expanding interpolations.
+(** Rewrites [dir/basename.in] to [dir/basename], expanding interpolations.
     If the first line begins ["opam-version:"], assumes that expansion of
     variables within strings should be properly escaped. In particular, this
     means that Windows paths should expand correctly when generating .config
-    files. *)
-val expand_interpolations_in_file: env -> basename -> unit
+    files.
+    [~dir] defaults to cwd. *)
+val expand_interpolations_in_file: dir: dirname option -> env -> basename -> unit
```
rjbou: I think it is better to have either
```
val expand_interpolations_in_file:  env -> filename -> unit
```
or
```
val expand_interpolations_in_file: dirname -> env -> basename -> unit
```
The second one have the benefit to avoid concatenation before calling the function each time, if we have a basename.

kit-ty-kate:
> The test failures seem to all come from `opam-rt` builds. I guess it needs to be patched as well.

done in https://github.com/ocaml-opam/opam-rt/pull/89

src/format/opamFilter.mli
NathanReb: Yeah I like the first one better! I'll do that instead!

NathanReb:
Any idea what's wrong with this build in particular: https://github.com/ocaml/opam/actions/runs/25102824494/job/73556728915?pr=6910

The error:
```
diff --git a/_build/default/tests/reftests/pin.test b/_build/default/tests/reftests/pin.out
File "tests/reftests/pin.test", line 1, characters 0-0:
index 581969b..baf3f8d 100644
/usr/bin/git --no-pager diff --no-index --color=always -u _build/default/tests/reftests/pin.test _build/default/tests/reftests/pin.out
--- a/_build/default/tests/reftests/pin.test
Command exited with code 1.
+++ b/_build/default/tests/reftests/pin.out
@@ -241,8 +241,8 @@ The following actions will be performed:
 -> installed nip-git2.ved
 Done.
 ### opam pin add ./nip-path2 --kind git | sed-cmd git | "128.*" -> "128"
-[ERROR] Command "git -C ${BASEDIR}/nip-path2 ls-files" failed:
-"git -C ${BASEDIR}/nip-path2 ls-files" exited with code 128
+[ERROR] Command "git -C /tmp/build_e05667_dune/opam-reftest-c7128
+"git -C /tmp/build_e05667_dune/opam-reftest-c7128
 # Return code 99 #
 ### : auto
 ### opam pin add ./nip-git3 --kind auto
make: *** [Makefile:217: tests] Error 1
Error: Process completed with exit code 2.
```

NathanReb:
Ah just got it, the regexp matched on the generated directory name! I'll see if I can make the regexp more robust!

NathanReb:
Maybe that should in fact be fixed at the reftest-runner level though, it seems the regexp was applied before the path normalization, I doubt there is any concrete case where we want that. Let me know what you'd prefer!

NathanReb:
I went for the former. A bit of testing showed that the changes in the reftest runner would be a bit more involved than I anticipated as in some tests at least, `sed-cmd` as to be run before the `common_filters` so we'd need a finer orderding than simply `common_filters @ filters`.

NathanReb:
rebased on top of #6922 to show the impact on the `opam pin list` bug.

NathanReb:
Is anything still blocking this?

src/client/opamConfigCommand.ml
```diff
@@ -377,8 +377,11 @@ let env gt switch ?(set_opamroot=false) ?(set_opamswitch=false)
 let subst gt fs =
   log "config-substitute";
   OpamSwitchState.with_ `Lock_none gt @@ fun st ->
+  let env = OpamPackageVar.resolve st in
   List.iter
-    (OpamFilter.expand_interpolations_in_file (OpamPackageVar.resolve st))
+    (fun b ->
+       let file = OpamFilename.of_basename b in
```
kit-ty-kate, line 383: of_basename changes the behaviour quite a lot here. It will always start the string from cwd. From what i understand `b` can be an absolute as well

kit-ty-kate (changes requested):

src/client/opamConfigCommand.ml
NathanReb: This used to happen within `expand_interpolations_in_file` so unless I missed something, this is exactly the same behaviour.

kit-ty-kate: fair enough. This is still a bug to be fixed but we can do that separately later.

kit-ty-kate (approved):

src/repository/opamHTTP.ml
```diff
@@ -97,9 +97,13 @@ end
...
-  OpamFilename.in_dir repo_root (fun () ->
-    let to_include = [ "version"; "packages"; "repo" ] in
-    match List.filter Sys.file_exists to_include with
-    | [] -> ()
-    | d  -> OpamSystem.command ("tar" :: "czhf" :: "index.tar.gz" :: "--exclude=.git*" :: d)
-  )
+  let repo_root = OpamFilename.Dir.to_string repo_root in
+  let (/) = Filename.concat in
+  let to_include = [ "version"; "packages"; "repo" ] in
+  match List.filter (fun i -> Sys.file_exists (repo_root / i)) to_include with
```
rjbou:
```suggestion
  let open OpamFilename.Op in
  let to_include = [ "version"; "packages"; "repo" ] in
  match List.filter (fun i -> OpamFilename.exists (repo_root // i)) to_include with
  | [] -> ()
  | d  ->
    OpamSystem.command
      ("tar" :: "czhf" :: "index.tar.gz" ::
       "-C" :: OpamFilename.Dir.to_string repo_root :: "--exclude=.git*" ::
       d)
```

master_changes.md
```diff
@@ -249,3 +251,10 @@ users)
   * `OpamFilename`: add `is_dir_read_only` [#6489 @rjbou]
   * `OpamFilename.might_escape`: ensure / is detected as a file separator when called with `~sep:Unspecified` on Windows [#6897 @kit-ty-kate]
   * `OpamFilename.Unix` was added abstracting over `/` separated paths regardless of the current system [#6914 @rjbou @kit-ty-kate]
+  * `OpamFilename.in_dir`: removed [#6910 @NathanReb]
+  * `OpamSystem.in_tmp_dir`: removed [#6910 @NathanReb]
+  * `OpamSystem.in_dir`: removed [#6910 @NathanReb]
+  * `OpamSystem.chdir`: removed [#6910 @NathanReb]
+  * `OpamSystem.command`: added a `?dir: dirname` optional arg [#6910 @NathanReb]
+  * `OpamSystem.commands`: added a `?dir: dirname` optional arg [#6910 @NathanReb]
+  * `OpamSystem.read_command_output`: added a `?dir: dirname` optional arg [#6910 @NathanReb]
```
```suggestion
  * `OpamSystem.{command,commands,read_command_output}`: add a `?dir: dirname` optional arg to launch the command in a specific directory [#6910 @NathanReb]
```

rjbou (approved):
Tiny comments, otherwise, lgtm!

kit-ty-kate:
Thanks!

PR #6912, opened 2026-04-28
Add a benchmark for opam install

kit-ty-kate:
As requested in https://github.com/ocaml/opam/pull/6906

This adds two benchmarks:
- one tests a long install with many small and not-so small dependencies that are both able to build in parallel and also sequential 
- the other one tests specifically https://github.com/ocaml/opam/pull/6906 by exaggerating IO pressure to mimic Windows/IO-constrained filesystem behaviour, by allowing the removal of the build directory from one package to be done while another one is being built

### DIFF
--- a/tests/bench/bench.ml
+++ b/tests/bench/bench.ml
@@ -136,6 +136,37 @@ let () =
     in
     List.fold_left (+.) 0.0 l /. float_of_int n
   in
+  let time_install_core =
+    Gc.compact ();
+    launch (fmt "%s switch create -y eight ocaml-base-compiler.4.14.0" bin);
+    launch (fmt "%s install --download-only -y core.v0.15.1" bin);
+    launch "test -d ~/.opam/eight/.opam-switch/sources && rm -rf ~/.opam/eight/.opam-switch/sources";
+    launch "test -d ~/.opam/eight/.opam-switch/build && rm -rf ~/.opam/eight/.opam-switch/build";
+    time_cmd ~exit:0 (fmt "%s install -y core.v0.15.1" bin)
+  in
+  let time_install_big =
+    Gc.compact ();
+    launch (fmt "%s switch create nine --empty" bin);
+    OpamSystem.write "/tmp/opam-bench-bigpkg/repo" {|opam-version: "2.0"
+|};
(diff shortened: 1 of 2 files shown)


tests/bench/bench.ml
```diff
@@ -136,6 +136,37 @@ let () =
     in
     List.fold_left (+.) 0.0 l /. float_of_int n
   in
+  let time_install_core =
+    Gc.compact ();
+    launch (fmt "%s switch create -y eight ocaml-base-compiler.4.14.0" bin);
+    launch (fmt "%s install --download-only -y core.v0.15.1" bin);
+    launch "test -d ~/.opam/eight/.opam-switch/sources && rm -rf ~/.opam/eight/.opam-switch/sources";
+    launch "test -d ~/.opam/eight/.opam-switch/build && rm -rf ~/.opam/eight/.opam-switch/build";
+    time_cmd ~exit:0 (fmt "%s install -y core.v0.15.1" bin)
```
rjbou, line 145: Is the idea here is to have the biggest number of dependencies with core.v0.15.1 ?

tests/bench/bench.ml
```diff
@@ -279,6 +310,16 @@ let () =
           "name": "OpamStd.String.split amortised over 10 runs",
           "value": %f,
           "units": "secs"
+        },
+        {
+          "name": "opam install core.v0.15.1",
```
rjbou: It's not clear what is highlighted in this bench

tests/bench/bench.ml
```diff
@@ -279,6 +310,16 @@ let () =
...
+          "value": %f,
+          "units": "secs"
+        },
+        {
+          "name": "opam install large-pkg large-pkg2",
```
rjbou:
```suggestion
          "name": "opam install lot-io-source lot-io-source2",
```
large-pkg is not explicit on what we want to highlight

tests/bench/bench.ml
kit-ty-kate:
```suggestion
          "name": "opam install large-number-of-deps",
```

rjbou (approved):

PR #6913, opened 2026-04-28
OpamFile: fix comparison between two dummy files

rjbou:
Otherwise, it tries to compare `path/to/<none>` and `./<none>` and fail

TODO:
  * [x] add changelog
  * [x] check documentation

In the story of #6625

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -227,6 +227,7 @@ users)
   * `OpamFile.URL` was moved to `OpamFile.URL_legacy` and a simpler `OpamFile.URL` module was created only containing non-IO functions removing the outdated `url` file support [#6827 @kit-ty-kate]
   * `OpamFile.Descr.of_legacy`: was added [#6827 @kit-ty-kate]
   * `OpamFile.URL.of_legacy`: was added [#6827 @kit-ty-kate]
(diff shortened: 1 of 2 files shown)


kit-ty-kate:
how is `filename` able to become `path/to/<none>` ?

rjbou:
There is an `OpamFilename.of_string/to_string` in `OpamFormat`. Usually, it is not an issue as all dummy files are created in the "same place", but when working on #6625 that had different kind of paths (relative), i had the issue. I don't know if i could reproduce it.

kit-ty-kate:
are there no chance to remove that hideous dummy_file value and use None instead (only using `<none>` when printing the value to string) ?

rjbou:
it is possible, however it requires big changes in `format/opamFile.ml` and maybe in `opam-file-format`

kit-ty-kate:
Fair enough, i've opened a new TODO ticket https://github.com/ocaml/opam/issues/6918 to remind us to do that later

kit-ty-kate (approved):

kit-ty-kate:
ignoring ocaml-benchmarks failure which is unrelated

Issue #6890, opened 2026-04-14
GHA: occasional failures in legacy-local on Windows


kit-ty-kate:
```diff
diff --git a/_build/default/tests/reftests/legacy-local.test b/_build/default/tests/reftests/legacy-local.out
index 3a0fb7b..b233a67 100644
--- a/_build/default/tests/reftests/legacy-local.test
+++ b/_build/default/tests/reftests/legacy-local.out
@@ -1292,20 +1292,40 @@ The following actions will be performed:
 -> retrieved P3.1~weird-version.test  (cached)
 -> retrieved P4.3  (cached)
 -> installed P1.1
--> installed P2.1
+[ERROR] The compilation of P2.1 failed at "./build.sh".
+##% output ###
Error: [m# [ERROR] Variable P1:bytecomp not found in switch 10+a+b
+# Bytecode Compilation
+# File "p2.ml", line 2, characters 2-4:
+# 2 |   P1.x ()
+#       ^^
+# Error: Unbound module P1
+
+
+
+<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
++- The following actions failed
+| - build P2 1
++- 
++- The following changes have been performed (the rest was aborted)
+| - install ocaml 10+a+b
+| - install P1    1
+| - install P3    1~weird-version.test
++- 
+
+
+The former state can be restored with:
+    ${OPAM} switch import "${BASEDIR}/OPAM/10+a+b/.opam-switch/backup/state-today.export"
 -> installed P3.1~weird-version.test
--> installed P4.3
-Done.
```

kit-ty-kate:
after reading the test file i can't find anything \~aggregiously\~ wrong, so the `[ERROR] Variable P1:bytecomp not found in switch 10+a+b` leads me to think more towards some kind of race condition between when P1 is installed and P2 reads the package variables from P1 using `opam var P1:bytecomp`, although i can't think of what could be doing that

kit-ty-kate:
as similar but different failure i encountered in https://github.com/ocaml/opam/pull/6915 is
```diff
diff --git a/_build/default/tests/reftests/legacy-local.test b/_build/default/tests/reftests/legacy-local.out
index 3a0fb7b..f8930f5 100644
--- a/_build/default/tests/reftests/legacy-local.test
+++ b/_build/default/tests/reftests/legacy-local.out
@@ -1147,51 +1147,56 @@ The following actions will be performed:
[...]

PR #6915, opened 2026-04-28
Allow opam update to update repositories that changed a file to a directory of the same name and vice versa

### LINKED ISSUE #3830: opam update patch fails on a file->dir rename in opam repository
the seemingly innocuous ocaml/opam-repository@3044264138073fb14576d59b8278c4f8f9b43e51 change broke all `opam update` calls in CI immediately with:

```
[default] synchronised from git+file:///home/opam/opam-repository
[ERROR] Could not update repository "default": "/usr/bin/patch -p1 -i /home/opam/.opam/log/processed-patch-6-58c514" exited with code 1
The command '/bin/sh -c opam update' returned a non-zero code: 40
The command "bash -ex ./.travis-docker.sh" exited with 40.
```

Looks like its due to a file -> directory rename which is tripping up patch; see https://github.com/ocaml/opam-repository/pull/13992/files for the PR that triggered it.

rjbou:
With the addition of tar.gz archives as an internal representation of a repository (#6625), we want to be able to diff between directories (as now), but also between archives and directory and archive.
`OpamRepositoryBackend.get_diff` was rewritten to support those several types. As tar.gz implementation only list existing files (no directories), it has a side effect to no longer easily detect directory-file renames automatically (see #3830) to error on them. The new implementation allow them and generates a diff accordingly.
~Our patch function didn't support application of this kind of operation, it is then updated to handle it properly.~

~In this PR, there is also the creation of `OpamPatch` module that contains patch related functions. As we have other things that will be added for patch function in #6625, it is moved to its own module in advance.~ extracted as no longer needed here.

The `patchDiff` test is also updated with new tests for this new supported diff operation and better rewrites for consistent output.

TODO:
  * [x] add changelog
  * [x] check documentation

fix #3830
In the story of #6625

### DIFF
--- a/src/repository/opamRepositoryBackend.ml
+++ b/src/repository/opamRepositoryBackend.ml
@@ -71,122 +71,112 @@ let job_text name label =
        (OpamConsole.colorise `green (OpamRepositoryName.to_string name))
        label)
 
-let get_files_for_diff parent_dir dir1 dir2 =
-  let getfiles parent_dir dir =
-    let dir = Filename.concat (OpamFilename.Dir.to_string parent_dir) dir in
-    OpamSystem.get_files_except_vcs dir
+(**  DIFF *)
+
+(* We put back the prefix for patch -p1 harmonisation *)
+let add_prefix repo1 repo2 =
+  let prefix repo =
+    OpamFilename.basename_dir repo
+    |> OpamFilename.Base.to_string
   in
-  match dir1, dir2 with
-  | None, None -> assert false
-  | Some dir, None ->
-    List.map (fun file -> (Some (dir^"/"^file), None))
-      (getfiles parent_dir dir)
-  | None, Some dir ->
-    List.map (fun file -> (None, Some (dir^"/"^file)))
-      (getfiles parent_dir dir)
-  | Some dir1, Some dir2 ->
-    let files1 = List.fast_sort String.compare (getfiles parent_dir dir1) in
-    let files2 = List.fast_sort String.compare (getfiles parent_dir dir2) in
-    let rec aux acc files1 files2 = match files1, files2 with
-      | (file1::files1 as orig1), (file2::files2 as orig2) ->
-        let cmp = String.compare file1 file2 in
-        if cmp = 0 then
-          aux ((Some (dir1^"/"^file1), Some (dir2^"/"^file2)) :: acc)
-            files1 files2
-        else if cmp < 0 then
-          aux ((Some (dir1^"/"^file1), None) :: acc) files1 orig2
-        else
-          aux ((None, Some (dir2^"/"^file2)) :: acc) orig1 files2
-      | file1::files1, [] ->
-        aux ((Some (dir1^"/"^file1), None) :: acc) files1 []
-      | [], file2::files2 ->
-        aux ((None, Some (dir2^"/"^file2)) :: acc) [] files2
-      | [], [] ->
-        acc
+  let p1 x = prefix repo1 ^ "/" ^ x in
+  let p2 x = prefix repo2 ^ "/" ^ x in
+  fun patch ->
+    let operation =
+      match patch.Patch.operation with
+      | Patch.Create f -> Patch.Create (p2 f)
+      | Patch.Delete f -> Patch.Delete (p1 f)
+      | Patch.Edit (f1, f2) -> Patch.Edit (p1 f1, p2 f2)
+      | Patch.Git_ext (f1, f2, ext) -> Patch.Git_ext (p1 f1, p2 f2, ext)
     in
-    aux [] files1 files2
+    {patch with operation}
 
-(* Serves to remove the repository suffix since the quarantine mechanism in
-   local and http patches causes incoherencies with vcs patches *)
-let strip_repo_suffix patch =
-  let rm_prefix f =
-    match OpamStd.String.cut_at f '/' with
-    | None ->
-      log "Internal diff: failed to remove prefix of %s" f;
-      f
-    | Some (_, r) -> r
-  in
-  let operation =
-    match patch.Patch.operation with
-    | Patch.Create f -> Patch.Create (rm_prefix f)
-    | Patch.Delete f -> Patch.Delete (rm_prefix f)
-    | Patch.Edit (f1, f2) -> Patch.Edit (rm_prefix f1, rm_prefix f2)
-    | Patch.Git_ext (f1, f2, ext) ->
-      Patch.Git_ext (rm_prefix f1, rm_prefix f2, ext)
-  in
-  {patch with operation}
-
-let get_diff parent_dir dir1 dir2 =
+let get_diff repo1 repo2 =
   let chrono = OpamConsole.timer () in
-  log "diff: %a/{%a,%a}"
-    (slog OpamFilename.Dir.to_string) parent_dir
-    (slog OpamFilename.Base.to_string) dir1
-    (slog OpamFilename.Base.to_string) dir2;
-  let readfile parent_dir file =
-    let real_file =
-      Filename.concat (OpamFilename.Dir.to_string parent_dir) file
+  log "diff: %a"
+    (fun fmt () ->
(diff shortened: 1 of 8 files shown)


rjbou:
To review with _hide withespace_ enabled

rjbou:
yeah, the doc job is working ^^

tests/reftests/update.test
```diff
@@ -905,3 +905,110 @@ The following actions would be performed:
...
+### :V: Changes that move a file into a directory and vice versa
+### <DREPO/repo>
+opam-version: "2.0"
+### <DREPO/packages/uno/uno.1>
+opam-version: "2.0"
+### <DREPO/packages/due/due.1/opam>
+opam-version: "2.0"
+### <DREPO/packages/due/due.1/files/archivio>
+Am I a file ?
+### <add-xf.sh>
```
kit-ty-kate, line 917:
```suggestion
### <add-xf.sh>
set -eu
```

src/core/opamPatch.ml
```diff
@@ -0,0 +1,408 @@
...
+      (* That seems to be the GNU patch behaviour *)
+      let file = if file1_exists then file1 else file2 in
+      let content = OpamSystem.read file in
+      let content = patch ~file:file (Some content) diff in
+      OpamSystem.write file content;
+      if file1_exists && file1 <> (file2 : string) then
+        OpamSystem.rmdir_cleanup (Filename.dirname file1)
+    | Patch.Delete file | Patch.Git_ext (file, _, Patch.Delete_only) ->
+      let file = get_path file in
+      OpamSystem.remove_file file;
```
kit-ty-kate: the only difference in this commit is the change from `OpamSystem.remove_file_t ~log:false` to `OpamSystem.remove_file`. As far as i can see they're not strictly compatible and do not use each others

src/repository/opamRepositoryBackend.ml
```diff
@@ -71,122 +71,118 @@ let job_text name label =
...
-        diffs files
-    in
-    diffs
+  let diffs, seen =
+    OpamStd.String.Map.fold
+      (fun filename content2 (diffs, seen) ->
+         get_content_diffs filename contents1 content2 diffs seen)
+      contents2 ([], OpamStd.String.Set.empty)
+  in
+  let diffs =
```
kit-ty-kate, line 167: a comment here would probably be nice, to make sure we remember that adding deletions at the beginning in necessary for dir --> file support

src/repository/opamRepositoryBackend.ml
```diff
@@ -71,122 +71,118 @@ let job_text name label =
...
-    | Patch.Delete f -> Patch.Delete (rm_prefix f)
-    | Patch.Edit (f1, f2) -> Patch.Edit (rm_prefix f1, rm_prefix f2)
-    | Patch.Git_ext (f1, f2, ext) ->
-      Patch.Git_ext (rm_prefix f1, rm_prefix f2, ext)
+(* We put back the prefix for patch -p1 harmonisation *)
+let add_prefix repo1 repo2 =
+  let prefix repo =
+    OpamFilename.basename_dir repo
+    |> OpamFilename.Base.to_string
+    |> OpamSystem.back_to_forward
```
kit-ty-kate: back_to_forward is dangerous here. The prefix should never have a /, otherwise the patch isn't p1 anymore.
```suggestion
```

src/repository/opamRepositoryBackend.ml
```diff
@@ -71,122 +71,118 @@ let job_text name label =
...
+      | Patch.Delete f -> Patch.Delete (p1 f)
+      | Patch.Edit (f1, f2) -> Patch.Edit (p1 f1, p2 f2)
+      | Patch.Git_ext (f1, f2, ext) ->
+        let f1, f2 =
+          match ext with
+          | Patch.Rename_only (_, _) -> f1, f2
+          | Patch.Delete_only -> p1 f1, p1 f2
+          | Patch.Create_only -> p2 f1, p2 f2
+        in
+        Patch.Git_ext (f1, f2, ext)
```
kit-ty-kate: Rename_only is the only one that doesn't have a prefix but only for the bits inside Rename_only, not for f1 and f2 from Git_ext.
```suggestion
        Patch.Git_ext (p1 f1, p2 f2, ext)
```

src/repository/opamRepositoryBackend.ml
```diff
@@ -71,122 +71,118 @@ let job_text name label =
...
+            | Unix.S_BLK -> failwith "Block devices are unsupported"
+            | Unix.S_FIFO -> failwith "Named pipes are unsupported"
+            | Unix.S_SOCK -> failwith "Sockets are unsupported"
+          ) acc entries
+      in
+      aux OpamStd.String.Map.empty None dir
     in
-    (file, OpamSystem.read real_file)
+    fun dir ->
+    read_dir_contents (OpamFilename.Dir.to_string dir)
```
kit-ty-kate:
```suggestion
      read_dir_contents (OpamFilename.Dir.to_string dir)
```

kit-ty-kate (commented):
- 885263d160bfc77fd9ebb7ce97006d23868c30e8: most of the change is actually just a debug pretty-printer for the tests. I would rather see this change moved to a different commit, otherwise it's really hard to understand the different changes. I would also prefer this to be in its own function or wrapper or something rather than in the main patch function.
For the actual change, only the `Rename_only` case was changed. In theory, `Edit` might also be affected, but that's more for handwritten patches, for which we can just say "no thanks" for now.

- b3fd273135bf905c7032d29f045ca8ab5b65cdbd: aside from the couple of minor issues that i've highlighted this seems pretty good. Obviously this needs to have a better commit message and maybe a short commit description

src/repository/opamRepositoryBackend.ml
rjbou: removed, it was something kept from a previous version i guess, there is no such meaning to have the back to forward on a basename

src/core/opamPatch.ml
rjbou: Agree, but remove_file_t is not exported. This is an issue in `OpamSystem` module, we shouldn't have several removal functions (there is a same issue with `ls` kind functions). 
The two functions does almost the same thing, the main difference is the logging, and `remove_file_t` checks beforehand that it is a `Sys.win32` before calling the windows unlinking, while `remove_file` wait to fail, and checks it has an `EACCESS` error and that it is `Sys.win32`. Both perform the same unlinking on non windows and on windows.

src/repository/opamRepositoryBackend.ml
rjbou: I extracted this one on its own commit, to be able to check the effect

rjbou:
> https://github.com/ocaml/opam/commit/885263d160bfc77fd9ebb7ce97006d23868c30e8: most of the change is actually just a debug pretty-printer for the tests. I would rather see this change moved to a different commit, otherwise it's really hard to understand the different changes. I would also prefer this to be in its own function or wrapper or something rather than in the main patch function.

Most of the change is the support for dir->file substitution. The test shows only the debugging because it's the only thing that can be shown. I changed the commits order to be able to see better the effect of each commit (diff/patch), and extracted the error pretty-printing in its own commit.

> For the actual change, only the Rename_only case was changed. In theory, Edit might also be affected, but that's more for handwritten patches, for which we can just say "no thanks" for now.

From our discussions, `Edit` is not affected by these changes.

src/repository/opamRepositoryBackend.ml
rjbou: i'm not sure about this one

kit-ty-kate (approved):

rjbou:
LGTM! good to merge once tests are green

src/core/opamPatch.ml
rjbou: no longer in this PR

kit-ty-kate:
Ignoring the Windows failure (https://github.com/ocaml/opam/issues/6890). Thanks!

PR #6917, opened 2026-04-29
Add modules to specify some path elements names

rjbou:
For opam root hierarchy and repository layout.
It avoids some hardcoded directory of filenames
Part of is may be needed for repo tar


src/format/opamRepositoryPathName.mli
```diff
@@ -0,0 +1,18 @@
...
+(*                                                                        *)
+(**************************************************************************)
+
+(** {1 Names of a repository elements layout} *)
+
+val repo_f : string
+val repo_d : string
+val packages_d : string
+val download_cache_d : string
+val files_d : string
```
kit-ty-kate: i don't think having a separate file for just these is worth it. Most of these except for repo_f and download_cache_d are already in OpamPathName. I'd much rather put everything into OpamPathName

rjbou: It's not the same level: `OpamPathName` contains opam root layout names, while `OpamRepositoryPathName` contains repository layout names. In fact, `OpamPathName` and especially `OpamPath` can be moved to `opam-state` (its there that we act on opam root).

We can avoid the new files and have a submodule `Name` in each `Opam*Name` module.

On `OpamRepositoryPathName`, having its own module permit to break the cycle between `OpamRepositoryRoot` and `OpamRepositoryPath` for `delayed_read_repo` (to not hardcode `"repo"`). But there is another solution, the function can take as argument the filename.

src/format/opamPathName.ml
```diff
@@ -0,0 +1,39 @@
...
+let opam_f = "opam"
+let lock_f = "lock"
+let environment_f = "environment"
+
+let changes_suffix = ".changes"
+let install_suffix = ".install"
+let config_suffix = ".config"
+let opam_suffix = ".opam"
+let subst_suffix = ".in"
+let subst_suffix_wodot = "in"
```
rjbou:
```suggestion
let subst_ext = "in"
```

src/format/opamPathName.mli
```diff
@@ -0,0 +1,45 @@
...
+val lock_f: string
+val environment_f: string
+
+(** {2 Suffixes} *)
+val changes_suffix: string
+val install_suffix: string
+val config_suffix: string
+val opam_suffix: string
+val subst_suffix: string
+val subst_suffix_wodot: string
```
rjbou:
```suggestion
val subst_ext: string
```

src/format/opamFilter.ml
```diff
@@ -463,7 +463,7 @@ let expand_interpolations_in_file_full env ~src ~dst =
 (* Substitute the file contents *)
 let expand_interpolations_in_file env file =
   let file = OpamFilename.of_basename file in
-  let src = OpamFilename.add_extension file "in" in
+  let src = OpamFilename.add_extension file OpamPathName.subst_suffix_wodot in
```
rjbou:
```suggestion
  let src = OpamFilename.add_extension file OpamPathName.subst_ext in
```

src/format/opamRepositoryPathName.ml
```diff
@@ -0,0 +1,15 @@
...
+(*  GNU Lesser General Public License version 2.1, with the special       *)
+(*  exception on linking described in the file LICENSE.                   *)
+(*                                                                        *)
+(**************************************************************************)
+
+let repo_d = "repo"
+let repo_f = "repo"
+let packages_d = "packages"
+let download_cache_d = "download-cache"
+let files_d = OpamPathName.files_d
```
rjbou:
```suggestion
let files_d = "files"
```

src/format/opamPathName.ml
```diff
@@ -0,0 +1,39 @@
...
+(*                                                                        *)
+(*    Copyright 2025 OCamlPro                                             *)
+(*                                                                        *)
+(*  All rights reserved. This file is distributed under the terms of the  *)
+(*  GNU Lesser General Public License version 2.1, with the special       *)
+(*  exception on linking described in the file LICENSE.                   *)
+(*                                                                        *)
+(**************************************************************************)
+
+let files_d = "files"
```
rjbou:
```suggestion
let files_d = OpamRepositoryname.files_d
```

src/format/opamPathName.ml
```diff
@@ -0,0 +1,39 @@
...
+let repo_d = "repo"
+let build_d = "build"
+let config_d = "config"
+let hooks_d = "hooks"
+let install_d = "install"
+let log_d = "log"
+let overlay_d = "overlay"
+let packages_d = "packages"
```
rjbou:
```suggestion
let packages_d = OpamRepositoryname.packages_d
```

src/state/opamPinned.ml
```diff
@@ -26,14 +26,17 @@ let version_opt st name = try Some (version st name) with Not_found -> None
...
 
-let possible_definition_filenames dir name = [
-  dir / (OpamPackage.Name.to_string name ^ ".opam") // "opam";
-  dir // (OpamPackage.Name.to_string name ^ ".opam");
-  dir / "opam" / (OpamPackage.Name.to_string name ^ ".opam") // "opam";
-  dir / "opam" // (OpamPackage.Name.to_string name ^ ".opam");
-  dir / "opam" // "opam";
-  dir // "opam"
-]
+let possible_definition_filenames dir name =
```
rjbou, line 29: for the future : should we integrate that in OpamPath ?

src/format/opamPathName.ml
kit-ty-kate: this is not related to repositories, this is `.opam-switch/packages`

src/format/opamPathName.ml
kit-ty-kate: In theory both are distinct and could be renamed to some other names separately

kit-ty-kate (approved):

rjbou:
LGTM!

PR #6920, opened 2026-05-01
Improve visibility of depopts filter note in documentation

### LINKED ISSUE #5367: TODO: Document a little better the behaviour of filters in depopts
Currently the documentation is sort of lacking in this section wrt the tricky difference of behaviour between depopts and conflicts

ccoulombel:
This change makes the existing note more visible by using bold formatting.

Fixes #5367


kit-ty-kate:
Thanks! Maybe the whole line should be bold, but let's start with that and we'll see from there.

kit-ty-kate (approved):

PR #6922, opened 2026-05-04
Add regression test for aborted local install leaving half-pinned pkgs

NathanReb:
This adds a test to show what happens when an install containing local packages is aborted or fails mid run and what packages are still pinned afterwards.

At the moment it only shows that it leaves the switch in a state that `opam pin list` cannot handle and it crashes with a terrible `chdir` error.

After #6910, `opam pin list` should work and it will show that packages pinned during the aborted install are left as such, which is something that we'd like to change and that I'll fix in a separate PR.

### DIFF
--- a/tests/reftests/autopin.test
+++ b/tests/reftests/autopin.test
@@ -428,3 +428,37 @@ Switch invariant: ["ocaml" {>= "4.05.0"}]
 
 Switch initialisation failed: clean up? ('n' will leave the switch partially installed) [Y/n] y
 # Return code 20 #
+### :::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
+### : Aborting install . should not leave half-pinned packages around
+### :::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::::
+### opam switch create abort-install-local --empty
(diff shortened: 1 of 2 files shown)


kit-ty-kate (approved):

tests/reftests/autopin.test
```diff
@@ -428,3 +428,49 @@ Switch invariant: ["ocaml" {>= "4.05.0"}]
...
+On branch master
+
+Initial commit
+
+Untracked files:
+  (use "git add <file>..." to include in what will be committed)
+	abort-install.opam
+
+nothing added to commit but untracked files present (use "git add" to track)
+# Return code 1 #
```
kit-ty-kate:
```suggestion
### git -C abort-install add abort-install.opam
### git -C abort-install commit -qm 'init'
```

kit-ty-kate:
Ignoring the failure on Windows (see https://github.com/ocaml/opam/issues/6890). Thanks!

tests/reftests/autopin.test
NathanReb: ouh thanks, my bad, forgot to clean that one up!

PR #6923, opened 2026-05-05
Offer to remove autopin pkgs on install failure

### LINKED ISSUE #6601: opam install <local> pins packages even if the user doesn't proceed on the installation
https://github.com/ocaml/opam/issues/6600 shows that `opam install .` can pin a package, but if the user doesn't accept the installation, the package stays pinned (but not fetched in the sources directory).

In my opinion `opam install <local>` shouldn't write the new list of pinned packages on disk until the installation actually starts.

NathanReb:
This PR adds a prompt when an `opam install` command fails to install packages or is aborted, offering them to remove packages that were automatically pinned as part of the process.

This mostly affects installation of local packages, i.e. when `opam install` is passed a file or directory as this triggers an automatic pin of the given opam file or any opam file found within the directory, alongside their pin-depends.

It allows users to quickly clean up if they realize that was not their intent or if they are unable to proceed with the install.

Queued on
* [x] #6910

### DIFF
--- a/src/client/opamCommands.ml
+++ b/src/client/opamCommands.ml
@@ -1894,10 +1894,28 @@ let install cli =
       else atoms_or_locals
     in
     if formula = OpamFormula.Empty && atoms_or_locals = [] then `Ok () else
+    let already_pinned = st.pinned in
     let st, atoms =
       OpamAuxCommands.autopin
         st ~recurse ?subpath ~quiet:check ~simulate:(deps_only||check||depext_only)
         ?locked:OpamStateConfig.(!r.locked) atoms_or_locals
+    in
+    let newly_pinned = OpamPackage.Set.diff st.pinned already_pinned in
+    let clean_up_pins_on_failure st () =
+      let newly_pinned_l =
+        OpamPackage.Set.elements newly_pinned
+        |> List.map OpamPackage.name
+      in
+      match newly_pinned_l with
+      | [] -> ()
+      | _ ->
+        OpamConsole.msg
+          "Unpinning packages pinned during install:\n%s"
+          (OpamStd.Format.itemize
+             (fun name ->
+                OpamConsole.colorise `bold (OpamPackage.Name.to_string name))
+             newly_pinned_l);
+        ignore (OpamPinCommand.unpin st newly_pinned_l)
     in
     if formula = OpamFormula.Empty && atoms = [] then
       (OpamConsole.msg "Nothing to do\n";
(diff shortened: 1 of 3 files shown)


NathanReb:
I'm opening this as draft because I have some questions regarding what should be done in those cases:
1. when `--check` is passed and everything is installed already. I'd assume if this is the case, then nothing has been pinned but I might be wrong about that.
2. when `--check` is passed and some packages must be installed.
3. when `dry-run` or `show`
4. when `simulate` (i.e. `--check`, `--deps-only` or `--depext-only`

1 and 2 are kind of a specific case of 4. If my understanding of 4 is correct, no pin is actually written to disk in that case and therefore there's no clean up left to do.

I also need to correctly update the tests so that the installation aborted test makes sure everything is correctly cleaned up, e.g. by adding `pin-depends` into the mix. I'll of course have to test the scenarios described above as well.

src/client/opamCommands.ml
```diff
@@ -1894,11 +1894,33 @@ let install cli =
...
+      | [] -> ()
+      | _ ->
+        OpamConsole.msg
+          "The following packages were pinned during install:\n%s"
+          (OpamStd.Format.itemize
+             (fun name ->
+                OpamConsole.colorise `bold (OpamPackage.Name.to_string name))
+             newly_pinned_l);
+        let unpin =
+          OpamConsole.confirm ~name:"unpin-on-failed-install" "Unpin them?"
```
kit-ty-kate: From a user PoV, it feels redundant with the installation confirmation, and causes weird interactions with `--yes` etc.
I think it should simply be the default to unpin when the user doesn't want to install the proposed packages during opam install.

NathanReb: That sounds good to me. Do you believe it should be the same for install failures?

kit-ty-kate: install failures seem different. The user intended to install (and pin) them. Also it's harder to detect which packages actually failed, they can partially fail (some dependencies might have been installed), etc.
Also it feels like it would be a bigger breaking change than only unpinning them upon non-confirmation.

NathanReb:
I updated the code so it unpins only when the user aborts the install and does so without asking for confirmation. The test was updated accordingly and I added a pin-depend in the mix to make sure it was also unpinned correctly.

tests/reftests/autopin.test
```diff
@@ -441,24 +443,82 @@ homepage: "foo.com"
 bug-reports: "foo.com/bugs"
 license: "MIT"
 synopsis: "Test pkg please ignore"
-depends: [ "dep-lib" ]
-### <pkg:dep-lib.1>
-opam-version: "2.0"
+depends: [ "dep-lib" "pin-dep-lib" ]
+### <add-pin-depends.sh>
```
kit-ty-kate, line 507:
```suggestion
### <add-pin-depends.sh>
set -eu
```

kit-ty-kate (commented):
Could you split the addition of the new test from the fix? Doing it this way makes it very easy to understand what the fix changed by simply reading the diff

kit-ty-kate:
While the PR looks reasonable overall i'd like a bit more time than we have to review it and think it over, so i'm bumping this to the 2.7 milestone

PR #6871, opened 2026-03-24
Repository loading: update load from diff mechanism

rjbou:
It simplifies the readability of the code and no longer loads unrelated paths.

TODO:
  * [x] add changelog
  * [ ] check documentation
  * [x] update with new code
  * [x] sanitise functions

In the story of #6625
Fixes https://github.com/ocaml/opam/issues/6872

PR #6924, opened 2026-05-05
Homogenize the behaviour of initial repository loads and subsequent loads from diff

### LINKED ISSUE #6872: Incremental repository loading loads every subdirectories including those not into packages/
https://github.com/ocaml/opam/pull/6614 (part of opam 2.5.0) added incremental loading of repositories. However, unlike `load_opams_from_dir`, `load_opams_from_diff` does not look into the `packages/` directory but starts from the repository root itself instead.

This means that adding a repository with `/test.1/opam` will not initially add a `test.1` package, but subsequent updates will.

This bug is present in opam 2.5.0 only.

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/6872
Alternative to https://github.com/ocaml/opam/pull/6871

This enforces:
- a consistent behaviour between the initial repo load and loads from diff
- load from diff to only look for packages in `packages/`
- load from diff to support extrafiles in a `files` subdirectory (as in `packages/a.1/files/files/file`)
- load_opams_from_dir to support `opam` files in intermediate directories (e.g. `packages/a/opam` will be ignored but `packages/a/a.1/opam` will now be detected). This is useless in practice but makes implementation easier and simplifies things for both load from diff and #6625 
- ... (TO FILL UP)

Should be reviewed using `Hide whitespace`


src/state/opamRepositoryState.ml
```diff
@@ -123,71 +113,72 @@ let load_opams_from_dir repo_name repo_root =
...
+      OpamFilename.Op.(repo_root / OpamFilename.Unix.Dir.to_string dir)
+    in
+    if OpamFilename.exists_dir full_path then
+      match OpamRepositoryPath.get_pkg_dir OpamFilename.Unix.Op.(dir // "") with
+      | Some (nv, _pkg_dir) ->
+        (* NOTE: we ignore [pkg_dir] because load_opams_from_dir
+           will always scan parent directories before going deeper *)
+        begin match read_package_opam ~repo_name ~repo_root full_path with
+        | Some opam -> OpamPackage.Map.add nv opam r
         | None -> r
```
rjbou, line 126: We no longer have the error `log "ERR: directory name not a valid package: ignored %s"` in this version.

src/state/opamRepositoryState.ml
```diff
@@ -123,71 +113,72 @@ let load_opams_from_dir repo_name repo_root =
...
-      in
-      if OpamFilename.Base.to_string basename = "files" then
-        OpamFilename.Dir.of_string (Filename.dirname full_path)
-      else
-        OpamFilename.Dir.of_string full_path
-    in
-    if OpamFilename.Dir.Set.mem pkg_dir processed_dirs then
+    match OpamRepositoryPath.get_pkg_dir file with
+    | None ->
+      log "ERR: file name not a valid package: ignored %s"
```
rjbou, line 152:
```suggestion
      log "ERR: directory name not a valid package: ignored %s"
```

src/state/opamRepositoryState.ml
```diff
@@ -123,71 +113,72 @@ let load_opams_from_dir repo_name repo_root =
...
-        | Some (nv, opam) -> OpamPackage.Map.add nv opam r
+    let full_path =
+      OpamFilename.Op.(repo_root / OpamFilename.Unix.Dir.to_string dir)
+    in
+    if OpamFilename.exists_dir full_path then
+      match OpamRepositoryPath.get_pkg_dir OpamFilename.Unix.Op.(dir // "") with
+      | Some (nv, _pkg_dir) ->
+        (* NOTE: we ignore [pkg_dir] because load_opams_from_dir
+           will always scan parent directories before going deeper *)
+        begin match read_package_opam ~repo_name ~repo_root full_path with
```
rjbou, line 124: we need to check here if the file is `"opam"`, otherwise we take ito account each file that has a valid opam file content (the extrafile did not change)

tests/reftests/repository-formats.test
```diff
@@ -0,0 +1,628 @@
+N0REP0
```
rjbou, line 1: we need to add the same test with load_from_dir

rjbou: and a test for `files/files` to see the change

src/state/opamRepositoryState.ml
kit-ty-kate: The error cannot happen in the new version since we're checking using get_pkg_dir. Also i'm unable to find this debug log anywhere in our testsuite

src/state/opamRepositoryState.ml
kit-ty-kate: Well %s is not a directory but a file, so it should rather be
```suggestion
      log "ERR: file name is not part of a valid package directory: ignored %s"
```

src/state/opamRepositoryState.ml
kit-ty-kate: mmh i don't think so. Do you have an example where it does that? `read_package_opam` fails if we give it anything but a directory containing a file named `opam`.

src/core/opamFilename.ml
```diff
@@ -813,4 +813,19 @@ module Unix = struct
...
+      match String.split_on_char '/' x with
+      | [] | ""::_::_ -> assert false
+      | [""] -> [current_dir_name]
+      | x ->
+        List.filter (fun seg ->
+            if seg = ".." then
+              invalid_arg "'..' is forbidden"
+            else
+              seg <> "" && not (String.equal current_dir_name seg))
+          x
```
kit-ty-kate, line 828:
```suggestion
        match
          List.filter (fun seg ->
              if seg = ".." then
                invalid_arg "'..' is forbidden"
              else
                seg <> "" && not (String.equal current_dir_name seg))
            x
         with
         | [] -> [current_dir_name] | x -> x
```

src/state/opamRepositoryState.ml
```diff
@@ -123,71 +113,72 @@ let load_opams_from_dir repo_name repo_root =
...
+        let processed_dirs = OpamFilename.Unix.Dir.Set.add pkg_dir processed_dirs in
+        let full_path =
+          OpamFilename.Op.(repo_root / OpamFilename.Unix.Dir.to_string pkg_dir)
+        in
+        match read_package_opam ~repo_name:repo.repo_name ~repo_root full_path with
+        | Some opam ->
+          let added_pkgs = OpamPackage.Set.add nv added_pkgs in
+          OpamPackage.Map.add nv opam opams, processed_dirs, added_pkgs
+        | None ->
+          if is_removal then
```
rjbou, line 168: note: here we read the file even though it a a removal, ie the file does not exists

PR #6930, opened 2026-05-09
added return 0 for zsh to prevent leaking error codes

ta2005:
i added a return 0 in zsh to prevent leaking error codes
i have noticed this because it have status code in my prompt

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -138,6 +138,7 @@ users)
   * Improved depexts handling by caching system package availability during update, avoiding redundant system checks at install time. [#6489 @arozovyk - fix #6461]
 
 ## Shell
+  * added return 0 for zsh/bash/sh/fish to prevent leaking error codes [#6930 @ta2005]
 
 ## Internal
   * Improve cache-loading performance when using OCaml >= 5.4 by using `Gc.ramp_up` [#6515 @dra27]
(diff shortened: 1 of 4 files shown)


kit-ty-kate:
Thanks for the contribution, this change looks good for sh/bash/zsh. I've looked at the other shells too:
- `exit` in csh defaults to 0, so the current code is all good
- powershell and cmd work in different ways
- `return` in fish however, has the same behaviour as with POSIX shell

Would you be able to also change the fish case? Also your commit message only mentions zsh but changes the code for POSIX shell and bash too. If you don't have the time to do this yourself, i can amend your commit no problem.

ta2005:
> you

will be happy to change and test fish behaviour

ta2005:
The opam-init scripts for Zsh  was leaking a non-zero exit code (1) when the opam environment was already initialized. This occurred because the test command would fail and the subsequent return would propagate that failure code, causing issues for users with custom prompts that display the last command's status.

I have updated the templates for both Zsh and Fish to explicitly return 0 and for zsh added the true command, ensuring a clean shell startup. Verified locally on both shells

kit-ty-kate (approved):

rjbou (approved):

kit-ty-kate:
Thanks!

PR #6938, opened 2026-05-18
Add `root` and `rootexec` sections to .install files

### LINKED ISSUE #6919: Add a prefix field to .install files
See https://github.com/ocaml/dune/issues/14393
It looks to me that the only way to support this properly without rolling back part of 2.5.1 would be to allow a new `prefix` field to `.install` files.

Backward compatibility is a bit of a problem as unknown fields are simply ignored in older versions, but hopefully packages that use that feature can simply restrict availability to `opam-version >= "2.6"`. `opam-version: "2.6"` would similarly ignore the whole file.

WardBrian:
Would close #6919

* [x] Queued on #6956
* [x] Queued on #7026

### DIFF
--- a/src/core/opamFilename.mli
+++ b/src/core/opamFilename.mli
@@ -15,6 +15,9 @@
 (* Returns [true] if string contains '..' between directory separators *)
 val might_escape: sep:[`Unix | `Windows | `Unspecified ] -> string -> bool
 
+(* Returns a list of elements between directory separators *)
+val split: sep:[`Unix | `Windows | `Unspecified ] -> string -> string list
+
 (** Basenames *)
 module Base: sig
   include OpamStd.ABSTRACT
--- a/src/format/opamFile.ml
+++ b/src/format/opamFile.ml
@@ -3751,6 +3751,8 @@ module Dot_installSyntax = struct
   let format_version = OpamVersion.of_string "2.0"
 
   type t =  {
+    root    : (basename optional * basename option) list;
+    rootexec  : (basename optional * basename option) list;
     bin     : (basename optional * basename option) list;
     sbin    : (basename optional * basename option) list;
     lib     : (basename optional * basename option) list;
@@ -3768,6 +3770,8 @@ module Dot_installSyntax = struct
   }
 
   let empty = {
+    root     = [];
+    rootexec = [];
     lib      = [];
     bin      = [];
     sbin     = [];
@@ -3784,6 +3788,8 @@ module Dot_installSyntax = struct
     doc      = [];
   }
 
+  let root t = t.root
+  let rootexec t = t.rootexec
   let bin t = t.bin
   let sbin t = t.sbin
   let lib t = t.lib
@@ -3799,6 +3805,8 @@ module Dot_installSyntax = struct
   let lib_root t = t.lib_root
   let libexec_root t = t.libexec_root
 
+  let with_root root t = { t with root }
+  let with_rootexec rootexec t = { t with rootexec }
   let with_bin bin t = { t with bin }
   let with_sbin sbin t = { t with sbin }
   let with_lib lib t = { t with lib }
@@ -3858,31 +3866,59 @@ module Dot_installSyntax = struct
          else OpamFilename.Base.to_string op.c)
 
   let fields =
+    let raise_with_file file ?pos message =
+      Pp.bad_format ?pos ("%s " ^^ message) file
+    in
+    let pp_check_parent_dir =
+      Pp.check ~name:"rel-filename"
+        ~raise:raise_with_file
+        ~errmsg:"references its parent directory."
+        (Fun.negate @@ OpamFilename.might_escape ~sep:`Unspecified)
+    in
+    let pp_check_not_absolute =
+      Pp.check ~name:"rel-filename"
+        ~raise:raise_with_file
+        ~errmsg:"is an absolute filename."
+        Filename.is_relative
+    in
     let pp_field =
       Pp.V.map_list ~depth:1 @@ Pp.V.map_option
         (Pp.V.string -| pp_optional)
         (Pp.opt @@
-         Pp.singleton -| Pp.V.string -| Pp.pp ~name:"rel-filename"
-           (fun ~pos s ->
-              if OpamFilename.might_escape ~sep:`Unspecified s then
-                Pp.bad_format ~pos "%s references its parent directory." s
-              else if Filename.is_relative s then
-                OpamFilename.Base.of_string s
-              else
-                Pp.bad_format ~pos "%s is an absolute filename." s)
-           OpamFilename.Base.to_string)
+         Pp.singleton -| Pp.V.string
+         -| pp_check_parent_dir
+         -| pp_check_not_absolute
+         -| Pp.of_module "file" (module OpamFilename.Base))
+    in
+    let pp_root =
+      Pp.V.map_list ~depth:1 @@ Pp.V.map_option
+        (Pp.V.string -| pp_optional)
+        (Pp.opt @@
+         Pp.singleton -| Pp.V.string
+         -| pp_check_parent_dir
+         -| pp_check_not_absolute
+         -| Pp.check ~name:"rel-filename"
+           ~raise:raise_with_file
+           ~errmsg:"tries to overwrite opam internal files."
+           (fun s ->
+              match OpamFilename.split ~sep:`Unspecified s with
+              | root::_::_ -> (root : string) <> OpamPathName.opamswitch_d
+              | _ -> true)
+         -| Pp.of_module "file" (module OpamFilename.Base))
     in
     let pp_misc =
       Pp.V.map_list ~depth:1 @@ Pp.V.map_option
         (Pp.V.string -| pp_optional)
-        (Pp.singleton -| Pp.V.string -| Pp.pp ~name:"abs-filename"
-           (fun ~pos s ->
-              if not (Filename.is_relative s) then OpamFilename.of_string s
-              else Pp.bad_format ~pos
-                  "%s is not an absolute filename." s)
-           OpamFilename.to_string)
+        (Pp.singleton -| Pp.V.string
+         -| Pp.check ~name:"abs-filename"
+           ~raise:raise_with_file
+           ~errmsg:"is not an absolute filename."
+           (Fun.negate Filename.is_relative)
+         -| Pp.of_module "abs-filename" (module OpamFilename))
     in
     [
+      "root", Pp.ppacc with_root root pp_root;
+      "rootexec", Pp.ppacc with_rootexec rootexec pp_root;
       "lib", Pp.ppacc with_lib lib pp_field;
       "bin", Pp.ppacc with_bin bin pp_field;
       "sbin", Pp.ppacc with_sbin sbin pp_field;
--- a/tests/reftests/dot-install-non-exe-fields.test
+++ b/tests/reftests/dot-install-non-exe-fields.test
@@ -1835,3 +1835,282 @@ Some files in ${BASEDIR}/OPAM/sw-lib_root/.opam-switch/install/pkg-lib_root.inst
 # OPAM/sw-lib_root/.opam-switch/install/pkg-lib_root.changes not found
 ### ocaml cat.ml sw-lib_root pkg-lib_root
 ==> No 'installed-file' found
+### :::::::::::::::::::::::::
+### :V: Field 'root'
+### :::::::::::::::::::::::::
+### mkdir -p REPO/packages/pkg-root/pkg-root.1/files
+### cp installed-file REPO/packages/pkg-root/pkg-root.1/files/installed-file
+### opam switch create sw-root --empty
+### :V:1: No defined destination path
+### <pkg:pkg-root.1>
+opam-version: "2.0"
+### <pkg:pkg-root.1:pkg-root.install>
+root: [ "installed-file" ]
+### OPAMDEBUG=-5 opam install pkg-root
+The following actions will be performed:
+=== install 1 package
+  - install pkg-root 1
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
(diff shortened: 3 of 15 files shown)


dbuenzli:
> I'm currently unable to run the tests locally due to an issue that seems related to #5968

I recently got that aswell on a machine. It disappeared once I used the provided <https://opam.ocaml.org/install.sh> install script rather than download a binary and copy it to `~/.local/bin/opam`. (I was in a hurry so did not try to understand but I suspect the install script does some bubblewrap magic).

WardBrian:
Using install.sh (and installing in /usr/local instead of /usr) is what eventually resolved my bwrap issue as a user, but not when running development versions it seems. Oddly enough, I seem to recall the tests running fine before when I was regularly running into #5968 as a user

src/core/opamFilename.mli
```diff
@@ -15,6 +15,9 @@
 (* Returns [true] if string contains '..' between directory separators *)
 val might_escape: sep:[`Unix | `Windows | `Unspecified ] -> string -> bool
 
+(* Returns a lit of elements between directory separators *)
```
```suggestion
(* Returns a list of elements between directory separators *)
```

src/format/opamFile.ml
```diff
@@ -3829,32 +3837,54 @@ module Dot_installSyntax = struct
...
-              if OpamFilename.might_escape ~sep:`Unspecified s then
-                Pp.bad_format ~pos "%s references its parent directory." s
-              else if Filename.is_relative s then
-                OpamFilename.Base.of_string s
-              else
-                Pp.bad_format ~pos "%s is an absolute filename." s)
-           OpamFilename.Base.to_string)
+         Pp.singleton -| Pp.V.string -| pp_check_relative
+           -| pp_check_not_absolute
+           -| Pp.of_module "filename" (module OpamFilename.Base))
```
rjbou:
```suggestion
           -| Pp.of_module "file" (module OpamFilename.Base))
```

src/format/opamFile.ml
```diff
@@ -3829,32 +3837,54 @@ module Dot_installSyntax = struct
...
+        (Pp.V.string -| pp_optional)
+        (Pp.opt @@
+         Pp.singleton -| Pp.V.string -| pp_check_relative
+           -| pp_check_not_absolute
+           -| pp_check ~name:"rel-filename"
+               ~f:(fun s -> List.exists
+                   (String.equal ".opam-switch")
+                   (OpamFilename.split ~sep:`Unspecified s))
+               "%s tries to overwrite opam internal files."
+          -| Pp.of_module "filename" (module OpamFilename.Base))
```
rjbou:
```suggestion
          -| Pp.of_module "file" (module OpamFilename.Base))
```

src/format/opamFile.ml
```diff
@@ -3829,32 +3837,54 @@ module Dot_installSyntax = struct
...
-           (fun ~pos s ->
-              if OpamFilename.might_escape ~sep:`Unspecified s then
-                Pp.bad_format ~pos "%s references its parent directory." s
-              else if Filename.is_relative s then
-                OpamFilename.Base.of_string s
-              else
-                Pp.bad_format ~pos "%s is an absolute filename." s)
-           OpamFilename.Base.to_string)
+         Pp.singleton -| Pp.V.string -| pp_check_relative
+           -| pp_check_not_absolute
```
rjbou:
```suggestion
         Pp.singleton -| Pp.V.string
           -| pp_check_relative
           -| pp_check_not_absolute
```

src/format/opamFile.ml
```diff
@@ -3829,32 +3837,54 @@ module Dot_installSyntax = struct
          if op.optional then "?" ^ OpamFilename.Base.to_string op.c
          else OpamFilename.Base.to_string op.c)
 
+  let pp_check ~name ~f msg =
+      Pp.pp ~name
+        (fun ~pos s ->
+          if f s then Pp.bad_format ~pos msg s else s)
+        Fun.id
```
rjbou: This function is already defined, you can use `Pp.check` (src/format/opamPp.mli).

src/format/opamFile.ml
```diff
@@ -3829,32 +3837,54 @@ module Dot_installSyntax = struct
...
+    in
+    let pp_prefix =
+      Pp.V.map_list ~depth:1 @@ Pp.V.map_option
+        (Pp.V.string -| pp_optional)
+        (Pp.opt @@
+         Pp.singleton -| Pp.V.string -| pp_check_relative
+           -| pp_check_not_absolute
+           -| pp_check ~name:"rel-filename"
+               ~f:(fun s -> List.exists
+                   (String.equal ".opam-switch")
```
rjbou:
```suggestion
                   (String.equal OpamPathName.opamswitch_d)
```
`OpamPathName.opam-switch_d` doesn't exist yet, it can be added and replace in other files the hardcoded string.

src/format/opamFile.ml
```diff
@@ -3829,32 +3837,54 @@ module Dot_installSyntax = struct
...
          else OpamFilename.Base.to_string op.c)
 
+  let pp_check ~name ~f msg =
+      Pp.pp ~name
+        (fun ~pos s ->
+          if f s then Pp.bad_format ~pos msg s else s)
+        Fun.id
+
   let fields =
+    let pp_check_relative =
```
rjbou:
```suggestion
    let pp_check_parent_dir =
```
this function only checks if the `..` is present in the path

src/format/opamFile.ml
```diff
@@ -3829,32 +3837,54 @@ module Dot_installSyntax = struct
...
+           -| Pp.of_module "filename" (module OpamFilename.Base))
+    in
+    let pp_prefix =
+      Pp.V.map_list ~depth:1 @@ Pp.V.map_option
+        (Pp.V.string -| pp_optional)
+        (Pp.opt @@
+         Pp.singleton -| Pp.V.string -| pp_check_relative
+           -| pp_check_not_absolute
+           -| pp_check ~name:"rel-filename"
+               ~f:(fun s -> List.exists
```
rjbou: The opam switch internal directory is only at the root of the switch prefix, there is nothing preventing users to have internal `.opam-switch` directories inside their packages file. It is possible to check the first element, but there is a need to handle the case where the filename begins with one or several `./././path` (it is allowed)

rjbou (commented):
The PR itself looks good, I've put some comments on the code. It is missing a test where we see the new field (i realise that there is no test that checks all fields, it could be the moment to add, i can do it).

src/format/opamFile.ml
WardBrian: Yes, this is a conservative over-approximation. I believe the same is currently true for the `might_escape` check, since in theory a path like `./foo/bar/../bar` should be fine, but I believe it will be prevented at the moment

The odds of someone having a legitimate use case for a `.opam-switch` path seem low enough that being conservative seems like it would be fine in this case

WardBrian:
Thanks for the review @rjbou. I believe I addressed all your comments other than the desire for more testing

WardBrian:
@rjbou at this point, would you consider this feature safe to assume for the next opam release (i.e., it would be reasonable for `dune` to assume its existence and change what it generates in .install files)?

Do let me know if there is anything else you’d like me to do on the PR!

rjbou:
Updated with the new tests. The main todo is to add prefix check for opam internals (other fields), and an update of the windows test, and it'll be ready.

rjbou:
Added the new checks

src/client/opamInitDefaults.ml
```diff
@@ -148,7 +148,7 @@ module I = OpamFile.InitConfig
 
 let (@|) g f = OpamStd.Op.(g @* f) ()
 
-let init_config ?(sandboxing=true) () =
+let init_config ?(sandboxing=false) () =
```
kit-ty-kate: why is that changed?

src/format/opamFile.ml
```diff
@@ -3857,32 +3865,79 @@ module Dot_installSyntax = struct
          if op.optional then "?" ^ OpamFilename.Base.to_string op.c
          else OpamFilename.Base.to_string op.c)
 
+
```
kit-ty-kate: minor
```suggestion
```

src/format/opamFile.ml
```diff
@@ -3857,32 +3865,79 @@ module Dot_installSyntax = struct
...
+           let paths = OpamTypesBase.all_std_paths in
+           let paths = List.filter (fun x -> x <> Prefix) paths in
+           let paths = List.map OpamTypesBase.string_of_std_path paths in
+           (match OpamFilename.split ~sep:`Unspecified str with
+            | root::_::_ ->
+              if OpamStd.List.mem String.equal root paths then
+                Pp.warn ~pos
+                  "Path '%s' begins with %s directory in 'root' field. \
+                   Use '%s' field instead."
+                  str root root
```
kit-ty-kate: I don't think it's worth adding a warning at this point and there are legitimate cases where using `rootexec: ["file" {"share/blah"}]` will trigger the warning. Could we split it to a separate PR and think about it later?

tests/reftests/dot-install.test
```diff
@@ -396,6 +396,13 @@ misc: [
   "dosiero" { "/tmp/../middle/dosiero" }
   "good"    { "good" }
   ]
+prefix: [
```
kit-ty-kate:
```suggestion
root: [
```

src/format/opamFile.ml
```diff
@@ -3857,32 +3865,79 @@ module Dot_installSyntax = struct
...
+    in
+    let pp_warn_std_paths =
+      Pp.pp ~name:"std-paths"
+        (fun ~pos str ->
+           let paths = OpamTypesBase.all_std_paths in
+           let paths = List.filter (fun x -> x <> Prefix) paths in
+           let paths = List.map OpamTypesBase.string_of_std_path paths in
+           (match OpamFilename.split ~sep:`Unspecified str with
+            | root::_::_ ->
+              if OpamStd.List.mem String.equal root paths then
```
kit-ty-kate: this check is not the right one. The name of the field and the name of the directory it points to can be different (e.g. the `*exec` fields, `stublibs`, `toplevel`, …)

kit-ty-kate (commented):
the branch is a bit messy with all the fixup commits. I would suggest rebasing it with 3 commits (in order):
* OpamFile.Install.fields refactoring
* The addition of the root and rootexec fields
* Additions to the reftests

src/client/opamInitDefaults.ml
WardBrian: This was an accidental check in, I still haven’t figured out bwrap in a dev environment

src/format/opamFile.ml
rjbou: We are not checking regarding the field name, but the root dir of the installation path (if given) versus all defined directories that will/would be created on the switch prefix given the `.install` file.

src/format/opamFile.ml
rjbou: Do you have an example of a legitimate case ?

src/client/opamInitDefaults.ml
rjbou:
```suggestion
let init_config ?(sandboxing=true) () =
```

rjbou:
> the branch is a bit messy with all the fixup commits. 

It is intended, to highlight my changes to @WardBrian and reciprocally

WardBrian:
> > the branch is a bit messy with all the fixup commits.
> 
> It is intended, to highlight my changes to @WardBrian and reciprocally

This has definitely been helpful so far but at this point I’d be okay with you rebasing it and erasing this distinction

src/format/opamFile.ml
```diff
@@ -3858,31 +3866,77 @@ module Dot_installSyntax = struct
...
+                Pp.warn ~pos
```
rjbou: todo: update Pp.warn with an option to always print warning (not guarded by verbose)

src/format/opamFile.ml
rjbou: cf #7039

src/format/opamFile.ml
rjbou: > Could we split it to a separate PR and think about it later?

opened #7039

tests/reftests/dot-install-non-exe-fields.test
```diff
@@ -1835,3 +1835,413 @@ Some files in ${BASEDIR}/OPAM/sw-lib_root/.opam-switch/install/pkg-lib_root.inst
...
+root: [
+  "installed-file" {"lib/installed-file.lib.root"}
+  "installed-file" {"bin/installed-file.bin.root"}
+  "installed-file" {"sbin/installed-file.sbin.root"}
+  "installed-file" {"share/installed-file.share.root"}
+  "installed-file" {"doc/installed-file.doc.root"}
+  "installed-file" {"etc/installed-file.etc.root"}
+  "installed-file" {"man/installed-file.man.root"}
+  "installed-file" {"toplevel/installed-file.lib.root"}
+  "installed-file" {"stublibs/installed-file.stublibs.root"}
```
kit-ty-kate: The same problem that i raised earlier (https://github.com/ocaml/opam/pull/6938#discussion_r3570754991) is still present here in some form. You're trying to check for paths that would otherwise be covered by dedicated fields but stublibs and toplevel aren't at the root of the switch directory, they're in `lib`.

tests/reftests/run.ml
```diff
@@ -698,6 +698,9 @@ let common_filters ?opam dir =
     extra_dotinstall_dirs "man";
     extra_dotinstall_dirs "sbin";
     extra_dotinstall_dirs "share";
+    extra_dotinstall_dirs "toplevel";
+    extra_dotinstall_dirs "stublibs";
```
kit-ty-kate: this shouldn't be needed (see previous comment)

kit-ty-kate (approved):
lgtm modulo the minor issue i raised above

rjbou:
Updated accordingly, and extracted #7039 specific tests in that PR

kit-ty-kate:
Thanks!

PR #6939, opened 2026-05-19
Add HTTP repository reftest

rjbou:
A reftest to test the http fetching/update mechanism of repository handling.
Part of #6625 story as it modifies the handling of http repositories

~~one that highlights all repository layout handled~~ moved to another PR as it need some code modification

### DIFF
--- a/tests/reftests/repository-http.test
+++ b/tests/reftests/repository-http.test
@@ -1,0 +1,172 @@
+N0REP0
+### ::: We test here repository HTTP backend
+### ::: Sections I to IV test the nominal operations
+### :: Setup
+### : Launch an minimal http server
+### http-server
+HTTP SERVER launched
+### : Creation and populate the repository
+### : we need to have repository layout at root as we use `opam admin` to create the archive
+### rm -rf REPO
+### <repo>
+opam-version:"2.0"
+### <packages/satu/satu.1/opam>
+opam-version:"2.0"
+### <packages/dua/dua.2/opam>
+opam-version:"2.0"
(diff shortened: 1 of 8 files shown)


tests/reftests/repository-http.test
```diff
@@ -0,0 +1,164 @@
+N0REP0
```
NathanReb, line 1: It doesn't seem to be the standard in the reftests but I think it would greatly improve their readability to describe the different setup and test steps with comments to provide a bit of context of what's being tested, what we expect the output to be and to clearly delimit each step and test.

rjbou: Updated with some comments

tests/reftests/repository-http.test
```diff
@@ -0,0 +1,177 @@
...
+dua    --
+empat  --
+enam   --
+lima   --
+satu   --
+tiga   --
+### :IV: Update of a package and check that the change is well present after an update
+### <packages/enam/enam.6/opam>
+opam-version: "2.0"
+build: ["true"]
```
NathanReb: How about adding a synopsis instead so that only `opam list --all` is needed to highlight that the change was correctly processed by `opam update`?

NathanReb (commented):
Looks good to me!

I have a minor comment on the last test to simplify it a bit but otherwise looks good!

tests/reftests/repository-http.test
rjbou: Indeed. Updated

NathanReb (approved):

PR #6941, opened 2026-05-20
Reftests add repository layout test

rjbou:
Add a test that shows the handled paths for opam packages, how and if they are read or processed by opam client (update & admin commands).

This mechanism was reworked by #6614 that introduced incremental loading (from diff instead of reading again the whole repository).

It highlights and permit to follow some errors that were unseen during #6614 and that are resolved by #6871 & #6924 (previous handled paths are seen in first commit[1]).

This PR adds the test, that is good to have before #6625 rewriting, to ensure non change on behaviour (even if erroneous).
However, in this PR we change a little bit the behaviour: first by adding a new logging to differentiate load from dir & load from diff calls ; and we sort the files read from disk during loading from dir to ensure an cross system reftest output (especially with `---strict` that error at the first error encountered).

~It also fixes a bug, a different behaviour between Unix & Windows. Stdlib `open_in` functions opens directories on Unix but error on Windows. Now, `OpamSystem.open_in*` error on all platforms if the filename is a directory.~

[1] ~The first commit is here for my personal use. It will be removed from the final version., it higlights the output of the test before before incremental loading merge (at commit https://github.com/ocaml/opam/commit/f3703519e0cdfb2bdd8232c1c897501df13c188a). It will be removed on merge.~ The commit that contains the result of the reftest before incremental loading is in on top of the branch [rjbou/bef-load-from-diff-reftests-repository-layout](https://github.com/rjbou/opam/tree/bef-load-from-diff-reftests-repository-layout)

In the story of #6625
Queued on https://github.com/ocaml/opam/pull/6995

### DIFF
--- a/src/state/opamRepositoryState.ml
+++ b/src/state/opamRepositoryState.ml
@@ -132,6 +132,7 @@ let load_opams_from_dir repo_name repo_root =
       let dir_str = OpamFilename.Dir.to_string dir in
       let fnames = Sys.readdir dir_str in
       let ( / ) = Filename.concat in
+      Array.sort String.compare fnames;
       if Array.exists (fun f -> String.equal f OpamPathName.opam_f &&
                                 not (Sys.is_directory (dir_str / f))) fnames then
         match read_package_opam ~repo_name ~repo_root dir with
--- a/src/state/opamRepositoryState.ml
+++ b/src/state/opamRepositoryState.ml
@@ -12,7 +12,7 @@
 open OpamTypes
 open OpamStateTypes
 
-let log fmt = OpamConsole.log "RSTATE" fmt
+let log ?level fmt = OpamConsole.log ?level "RSTATE" fmt
 let slog = OpamConsole.slog
 
 module Cache = struct
@@ -127,6 +127,8 @@ let load_opams_from_dir repo_name repo_root =
     OpamConsole.status_line "Processing: [%s: loading data]"
       (OpamConsole.colorise `blue (OpamRepositoryName.to_string repo_name));
   (* FIXME: why is this different from OpamPackage.list ? *)
+  log ~level:3 "load repo %a from dir"
+    (slog OpamRepositoryName.to_string) repo_name;
   let rec aux r dir =
     if OpamFilename.exists_dir dir then
       let dir_str = OpamFilename.Dir.to_string dir in
@@ -153,6 +155,8 @@ let load_opams_from_diff repo diffs rt =
   if OpamConsole.disp_status_line () || OpamConsole.verbose () then
     OpamConsole.status_line "Processing: [%s: loading data]"
       (OpamConsole.colorise `blue (OpamRepositoryName.to_string repo.repo_name));
+  log ~level:3 "load repo %a from diff"
+    (slog OpamRepositoryName.to_string) repo.repo_name;
   let existing_opams =
     OpamRepositoryName.Map.find repo.repo_name rt.repo_opams
   in
--- a/tests/reftests/repository-layout.test
+++ b/tests/reftests/repository-layout.test
@@ -146,6 +146,7 @@ RSTATE                          LOAD-REPOSITORY-STATE @ ${BASEDIR}/OPAM
 RSTATE                          Cache found
 [new-default-strict] Initialised
 UPDATE                          Repository has new changes
+RSTATE                          load repo new-default-strict from dir
 FILE(opam)                      Read ${BASEDIR}/OPAM/repo/new-default-strict/packages/good/good.1/opam in 0.000s
 ### OPAMDEBUGSECTIONS="FILE(opam) opam-file RSTATE UPDATE" OPAMDEBUG=-3 opam repository add new-default ./REPO --this-switch
 FILE(config)                    Read ${BASEDIR}/OPAM/config in 0.000s
@@ -153,6 +154,7 @@ RSTATE                          LOAD-REPOSITORY-STATE @ ${BASEDIR}/OPAM
 RSTATE                          Cache found
 [new-default] Initialised
 UPDATE                          Repository has new changes
+RSTATE                          load repo new-default from dir
 FILE(opam)                      Read ${BASEDIR}/OPAM/repo/new-default/packages/good/good.1/opam in 0.000s
 ### :I:1:b: List available package
 ### opam switch create good-sw-from-scratch --empty --repo=new-default,new-default-strict
@@ -186,6 +188,7 @@ RSTATE                          Cache found
 <><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
 [default-strict] synchronised from file://${BASEDIR}/REPO
 UPDATE                          Repository has new changes
+RSTATE                          load repo default-strict from diff
 FILE(opam)                      Read ${BASEDIR}/OPAM/repo/default-strict/packages/good/good.1/opam in 0.000s
 RSTATE                          ERR: Could not load ${BASEDIR}/OPAM/repo/default-strict/packages/good/good.1/files/un/der/opam, ignored
 Now run 'opam upgrade' to apply any package updates.
@@ -197,6 +200,7 @@ RSTATE                          Cache found
 <><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
 [default] synchronised from file://${BASEDIR}/REPO
 UPDATE                          Repository has new changes
+RSTATE                          load repo default from diff
 FILE(opam)                      Read ${BASEDIR}/OPAM/repo/default/packages/good/good.1/opam in 0.000s
 RSTATE                          ERR: Could not load ${BASEDIR}/OPAM/repo/default/packages/good/good.1/files/un/der/opam, ignored
 Now run 'opam upgrade' to apply any package updates.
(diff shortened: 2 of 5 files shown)


tests/reftests/repository-layout.test
```diff
@@ -0,0 +1,974 @@
+N0REP0
+### ::: We test here the several layout of opam packages paths in an opam repository
+### :   the one that are handled andthe one that aren't
+### # We uses 2 repositories, the defualt one and a second `default-strict` that will be called with the `--strict` option
```
NathanReb:
```suggestion
### # We uses 2 repositories, the default one and a second `default-strict` that will be called with the `--strict` option
```

src/state/opamRepositoryState.ml
```diff
@@ -128,13 +128,17 @@ let load_opams_from_dir repo_name repo_root =
     (slog OpamRepositoryName.to_string) repo_name;
   let rec aux r dir =
     if OpamFilename.exists_dir dir then
-      let fnames = Sys.readdir (OpamFilename.Dir.to_string dir) in
-      if Array.exists (fun f -> String.equal f OpamPathName.opam_f) fnames then
+      let fnames =
+        Sys.readdir (OpamFilename.Dir.to_string dir)
+        |> Array.to_list
```
NathanReb: Can't we simply use `Array.sort` here?

tests/reftests/repository-layout.test
```diff
@@ -0,0 +1,974 @@
+N0REP0
+### ::: We test here the several layout of opam packages paths in an opam repository
+### :   the one that are handled andthe one that aren't
```
NathanReb:
```suggestion
### :   the one that are handled and the one that aren't
```

tests/reftests/repository-layout.test
```diff
@@ -0,0 +1,974 @@
+N0REP0
+### ::: We test here the several layout of opam packages paths in an opam repository
```
NathanReb:
```suggestion
### ::: We test here the several layouts of opam packages paths in an opam repository
```

tests/reftests/repository-layout.test
```diff
@@ -0,0 +1,974 @@
...
+### :     We first add in the opam file an extra files declarations
+### :     We want to check then what is updated (layout) and if extra-file mismatching is well detected.
+### :     Then we add the extra files and upate repositories.
+### :     It is updates with and without strict mode, then check available packages
+### :     and if they are installable (if new extra-files were well retrieved)
+### <thrd>
+yet another file, no longer ignored!
+### openssl md5 thrd | '.*= ' -> '' >$ THRD_MD5
+### <xfile-opam.sh>
+# this script rewrites pam file by adding 'thrd' file in dextra files field'
```
NathanReb:
```suggestion
# this script rewrites an opam file, adding 'thrd' to the extra-files field
```

tests/reftests/repository-layout.test
```diff
@@ -0,0 +1,974 @@
...
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> installed z-direct-root.1
+Done.
+### opam install z-no-nv
+[ERROR] No package named z-no-nv found.
+# Return code 5 #
+### :IV: Unnecessary loading when a non opam file is updated.
+### :    In incremental loading, if next to an opam file a file not named opam
+### :    and containing a valid opam file content is updated (and only that
+### :    file in the packae directory), the correspondant opam package
```
NathanReb:
```suggestion
### :    file in the package directory), the correspondant opam package
```

tests/reftests/repository-layout.test
```diff
@@ -0,0 +1,974 @@
...
+### :    shouldn't be loaded. We can see that it that test by the mismatch
```
NathanReb:
```suggestion
### :    shouldn't be loaded. This is shown by the mismatch
```

tests/reftests/repository-layout.test
```diff
@@ -0,0 +1,974 @@
...
+### # then under 'pkg' directory, those should be ignored
+### sh populate.sh pkg/z-good/z-good.1
+### sh populate.sh pkg/z-not/z-no-n-nv.1
+### sh populate.sh pkg/z-ba/z-baz/z-n-nv-not-root.1
+### sh populate.sh pkg/z-v/z-e/z-r/z-y/z-very-inner.1
+### sh populate.sh pkg/z-i/z-n/z-e/z-r/z-very-inned-n-nv/z-very-inner-n-nv.1
+### sh populate.sh pkg/z-direct-root.1
+### sh populate.sh pkg/z-no-nv
+### sh populate.sh pkg/not-opam-filename/not-opam-filename.1
+### cp REPO/pkg/not-opam-filename/not-opam-filename.1/opam REPO/pkg/not-opam-filename/not-opam-filename.1/opam.notopamfile
```
NathanReb: Shouldn't that be:
```suggestion
### mv REPO/pkg/not-opam-filename/not-opam-filename.1/opam REPO/pkg/not-opam-filename/not-opam-filename.1/opam.notopamfile
```

tests/reftests/repository-layout.test
```diff
@@ -0,0 +1,974 @@
...
+### opam install not-same-nv-in.2
+The following actions will be performed:
+=== install 1 package
+  - install not-same-nv-in 2
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> installed not-same-nv-in.2
+Done.
+### opam install not-same-nv-in.2
+[NOTE] Package not-same-nv-in is already installed (current version is 2).
```
NathanReb: This looks like a duplicate. Is it intentional?

NathanReb (commented):
The test is quite large and I wonder if a slightly different structure could improve its readability and eventually even eliminate some duplication, potentially reducing the overall size.

Currently, if we leave out section IV, the test is roughly organised by setting up a repo with every layout under test (e.g. `packages/good/good.1/opam` or `packages/direct-root.1/opam`) and then run commands that will show the behaviour on each individual layout at once.

I'd like to suggest trying the opposite, i.e. for each layout run all the commands we want, describing how it should behave for said layout and then showing how it actually behaves in the commands output.
That would allow to reduce the size of the commands output (especially with the debug logs, its hard to keep track of everything) and keep everything closer to the documentation/comments that are relevant to this particular test.

That would also allow to describe what the layout actually are which could clear out some cases that may look like unintentional typos such as `packages/i/n/e/r/very-inned-n-nv/very-inner-n-nv.1` where `very-inned-n-nv` VS `very-inner-n-nv.1` could either be a typo or part of the test.

I suspect that trying out such a structure might show that some layouts actually exhibit the same behaviour because they go through the same code path in certain situations.

What do you think?

rjbou:
I like the idea!

src/state/opamRepositoryState.ml
```diff
@@ -124,15 +124,21 @@ let load_opams_from_dir repo_name repo_root =
...
+        |> List.sort String.compare
+      in
+      if List.exists (fun f -> String.equal f OpamPathName.opam_f) fnames then
```
```suggestion
        Sys.readdir (OpamFilename.Dir.to_string dir)
        |> Array.sort String.compare
      in
      if Array.exists (fun f -> String.equal f OpamPathName.opam_f) fnames then
```

tests/reftests/repository-layout.test
rjbou: indeed

rjbou:
Updated with the new structure.
The first commit is still to remove before merge

tests/reftests/repository-layout.test
```diff
@@ -0,0 +1,2462 @@
...
+### :VIII: packages/not-same-nv-in/not-same-nv-in.2/opam : package with the
+### :        good layout,but with specified version in the opam fille different
+### :IX:   packages/not-opam-filename/not-opam-filename.1/not-opam-filename :
+### :        package with the good layout, but the opam file is not named 'opam'
+### :X:    packages/opam-dir-nv/opam-dir-nv.1/opam/opam : opam file is under an
+### :        opam directory, with pattern 'packages/name/name.version/opam'
+### :XI:   packages/opam-dir-n/opam/opam : opam file is under an opam directory,
+### :        with pattern 'packages/name/opam'
+### :XII: pkg/z-good/z-good.1 : root directory is not 'packages'
+### :
```
kit-ty-kate, line 54: I think this one is missing
```suggestion
###:XIII: packages/opam : a lone opam file at the root of the packages directory
### :
```

tests/reftests/repository-layout.test
```diff
@@ -0,0 +1,2462 @@
...
+### :I:1:c: install available package
+### opam install good.1
+The following actions will be performed:
+=== install 1 package
+  - install good 1
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+-> installed good.1
+Done.
+### :I:2:a: Update repositories
```
kit-ty-kate: This 2:a step doesn't seem to do anything. Instead, i would've expected to:
- remove the repo
- create an empty repo
- readd the repo
- redo the populate step
- call opam update

This way both ways of parsing the repo are tested (load_from_dir and load_from_diff)

rjbou:
Updated, and updated the main comment

tests/reftests/repository-layout.test
```diff
@@ -0,0 +1,2574 @@
...
+<><> Updating package repositories ><><><><><><><><><><><><><><><><><><><><><><>
+[default] synchronised from file://${BASEDIR}/REPO
+Now run 'opam upgrade' to apply any package updates.
+### # unrelated new package, it is just here to highlight the unrelated update of wrong-opamf-updated
+### <REPO/packages/another-good/another-good.1/opam>
+opam-version: "2.0"
+### # we add change in opam.xxx
+### sh -c "echo >> REPO/packages/wrong-opamf-updated/wrong-opamf-updated.1/opam.xxx"
+### :C:1: Update of repositories
+### OPAMDEBUGSECTIONS="FILE(opam) opam-file RSTATE UPDATE" OPAMDEBUG=-3 opam update default-strict --strict | unordered
```
kit-ty-kate: are the ` | unordered` still useful after the addition of the `Array.sort`?

rjbou: nop

kit-ty-kate: i've removed them. Let's see if it's indeed the case everywhere

kit-ty-kate:
Ignoring ocaml-benchmarks

kit-ty-kate (approved):

PR #6945, opened 2026-05-22
Use `tar` library instead of `tar` tool for some operations

rjbou:
This PR adds in the API functions to manipulate `tar.gz` archives and use it to create archives in the code.
This new `OpamTar` module will be used in #6625 to open and read archives (internal representation of some repositories) without io.
Original work done by @kit-ty-kate

* [x] Queued on #6939
 
In the story of #6625

### DIFF
--- a/src/core/opamFilename.ml
+++ b/src/core/opamFilename.ml
@@ -392,9 +392,6 @@ let extract_in filename dirname =
 let extract_in_job filename dirname =
   OpamSystem.extract_in_job (to_string filename) ~dir:(Dir.to_string dirname)
 
-let make_tar_gz_job filename dirname =
-  OpamSystem.make_tar_gz_job (to_string filename) ~dir:(Dir.to_string dirname)
-
 type generic_file =
   | D of Dir.t
   | F of t
--- a/src/core/opamFilename.ml
+++ b/src/core/opamFilename.ml
@@ -810,4 +810,11 @@ module Unix = struct
 
   let of_filename x = OpamSystem.back_to_forward (concat x.dirname x.basename)
   let to_filename x = {dirname = dirname x; basename = basename x}
+
+  let starts_with prefix filename =
+    if prefix = (filename : string) then true
+    else
+      let prefix = concat prefix "" in
+      OpamCompat.String.starts_with ~prefix filename
+
 end
--- a/src/repository/opamTar.ml
+++ b/src/repository/opamTar.ml
@@ -18,45 +18,50 @@ type archived_file_content = string
 
 let log ?level fmt = OpamConsole.log "TAR" ?level fmt
 
+exception Tar of archive * string
+let raise_error tar fmt =
+  Printf.ksprintf (fun str -> raise (Tar (tar, str))) fmt
+
 let rec safe_read fd buf off len =
   try Unix.read fd buf off len
   with Unix.Unix_error (Unix.EINTR, _, _) -> safe_read fd buf off len
 
-let rec run : type a. Unix.file_descr -> (a, _, _) Tar.t -> a = fun fd -> function
-  | Tar.Read len ->
-    let b = Bytes.create len in
-    let read = safe_read fd b 0 len in
-    if read = 0 then
-      failwith "unexpected end of file"
-    else if len = (read : int) then
-      Bytes.unsafe_to_string b
-    else
-      Bytes.sub_string b 0 read
-  | Tar.Really_read len ->
-    let rec loop fd buf offset len =
-      if offset < (len : int) then
-        let n = safe_read fd buf offset (len - offset) in
-        if n = 0 then
-          failwith "unexpected end of file"
-        else
-          loop fd buf (offset + n) len
-    in
-    let buf = Bytes.create len in
-    loop fd buf 0 len;
-    Bytes.unsafe_to_string buf
-  | Tar.Return (Ok x) -> x
-  | Tar.Return (Error e) ->
-    let error =
-      match e with
-      | `Fatal e -> Format.asprintf "Fatal: %a" Tar.pp_error e
-      | `Eof -> "EOF"
-      | `Gz s -> "gz: "^s
-    in
-    failwith error
-  | Tar.High _ | Tar.Write _ | Tar.Seek _ -> assert false
-  | Tar.Bind (x, f) -> run fd (f (run fd x))
+let run archive =
+  let raise_error fmt = raise_error archive fmt in
+  let rec run : type a. Unix.file_descr -> (a, _, _) Tar.t -> a = fun fd -> function
+    | Tar.Read len ->
+      let b = Bytes.create len in
+      let read = safe_read fd b 0 len in
+      if read = 0 then
+        raise_error "unexpected end of file"
+      else if len = (read : int) then
+        Bytes.unsafe_to_string b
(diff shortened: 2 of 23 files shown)


tests/reftests/repository.test
```diff
@@ -36,7 +36,6 @@ build: ["test" "-f" "baz"]
 some content
 ### sh hash.sh REPO foo.2
 ### opam update default -vv | grep '^\+' | sed-cmd tar
-+ tar "cfz" "${BASEDIR}/OPAM/repo/default.tar.gz.tmp" "-C" "${BASEDIR}/OPAM/repo" "default"
```
rjbou, line 39: todo : add debug section TAR

src/core/opamFilename.ml
```diff
@@ -810,4 +810,8 @@ module Unix = struct
 
   let of_filename x = OpamSystem.back_to_forward (concat x.dirname x.basename)
   let to_filename x = {dirname = dirname x; basename = basename x}
+
+  let starts_with prefix filename =
+    OpamCompat.String.starts_with ~prefix filename
```
NathanReb: Isn't that at risk of suffering the same bug as the other `starts_with` functions?

I.e. `OpamFilename.starts_with ~prefiix:"/tmp/foo" "/tmp/foo-bar/baz"` will unexpectedly return true?

src/repository/opamTar.ml
```diff
@@ -0,0 +1,172 @@
...
+  try Unix.read fd buf off len
+  with Unix.Unix_error (Unix.EINTR, _, _) -> safe_read fd buf off len
+
+let rec run : type a. Unix.file_descr -> (a, _, _) Tar.t -> a = fun fd -> function
+  | Tar.Read len ->
+    let b = Bytes.create len in
+    let read = safe_read fd b 0 len in
+    if read = 0 then
+      failwith "unexpected end of file"
+    else if len = (read : int) then
```
NathanReb: We should use `Int.equal` here, shouldn't we?

src/repository/opamTar.mli
```diff
@@ -0,0 +1,56 @@
...
+(*  exception on linking described in the file LICENSE.                   *)
+(*                                                                        *)
+(**************************************************************************)
+
+(** Tar gz archives manipulation *)
+
+open OpamTypes
+
+type tar = filename
+type tar_file = unix_filename
```
NathanReb: How about:
```suggestion
type archive = filename
type archived_file = unix_filename
```

Which I think would clarify that one refer to a path leading to an archive while the other refers to a path within the archive.

NathanReb: I guess then we would need to make `tar_content` -> `archive_content` to be consistent.

I think it's also good to use a slightly different terminology as `tar` might be mistaken for an alias to `Tar.t` to a reader skimming through the documentation.

NathanReb: Well, circling back to this after reading further, it seems `tar_content` would in fact be `archived_file_content`.

src/repository/opamTar.ml
```diff
@@ -0,0 +1,172 @@
...
+      fun () ->
+        match dispenser () with
+        | None -> Tar.return (Ok None)
+        | Some x -> Tar.return (Ok (Some x))
+    in
+    let tar = Tar.out ~level:Ustar entries in
+    let tar = Tar_gz.out_gzipped ~level:4 ~mtime:0l Gz.Unix tar in
+    let buf = Buffer.create 10_485_760 in
+    to_buffer buf tar;
+    let str = Buffer.contents buf in
```
NathanReb, line 178: Do we have to go through this buffer to then allocate the whole string to then write to a Unix file descriptor?

Couldn't we write a `to_fd` function directly instead of the `to_buffer` one?

Is it so that we don't overwrite the original archive in case an error occurs? If so, then it's probably worth a comment or a bit of documentation.

src/repository/opamTar.ml
```diff
@@ -0,0 +1,172 @@
...
+        | Tar.Return (Error _) ->
+          failwith "something went wrong"
+        | Tar.Bind (x, f) ->
+          run buf (f (run buf x))
+      in
+      run buf t
+    in
+    let entries =
+      let dispenser =
+        Map.fold (fun path content acc ->
```
NathanReb: How about using `to_seq` followed by `Seq.map` here?

Was the manual construction of the `Seq` intentional? Is it a performance issue?

src/repository/opamTar.ml
```diff
@@ -0,0 +1,172 @@
...
+
+  let remove_dir dname t =
+    let content =
+      Map.filter (fun fname _ ->
+          not (OpamFilename.Unix.starts_with dname fname))
+        t.content
+    in
+    { t with content }
+
+  let write (t:t) =
```
NathanReb, line 136: Tiny note on this function: it's based on the map only so it entirely skips directory entries in the archive. This means directories metadata (permissions, mtime, and so on) along with empty directories are lost when modifying an existing archive.

I assume this is intentional because we don't care about those for our own use case (manipulating archived repos) but it might be worth documenting in this module which might seem like a general purpose archive manipulation module and could be reused elsewhere in the future.

src/repository/opamTar.ml
```diff
@@ -170,3 +170,22 @@ module Inplace = struct
     ()
 
 end
+
+let create tar dir =
```
NathanReb, line 187: Same comment as for the `write` function: dirs metadata and empty dirs are not preserved in the archive, probably worth a mention in the doc!

tests/reftests/repository.test
NathanReb, line 39: Shouldn't we have some new corresponding logs here so that the archiving operation still shows in the log, even if it's not done through a separate process anymore?

src/core/opamFilename.ml
rjbou: it is, it would be updated in the same PR that fixes it.

src/repository/opamTar.ml
rjbou: we use this annotation for a better readability

tests/reftests/repository.test
rjbou: we should, the test need to be updated for that. My comment was cryptic, it was about to remember to enable debug and add the debug section TAR which will print the tar creation log, and replace the verbose command display.

NathanReb: Ah yes my bad, my review span overnight and I forgot about your comment entirely. It didn't show up in the per-commit view I used for some reason.

rjbou:
I've updated the PR taking into account some remarks. I'll let @kit-ty-kate answer `OpamTar` ones.

src/core/opamFilename.ml
rjbou: cf #6948

src/repository/opamTar.ml
rjbou: Opening issue to follow up #6950, if there is performance benefit

src/repository/opamTar.ml
rjbou: update pushed

NathanReb (approved):

rjbou:
openSUSE depexts failure is unrelated, skipping

kit-ty-kate (approved):

PR #6947, opened 2026-05-26
Fix `undefined variable` error when a lock file filter contains an undefined variable

### LINKED ISSUE #6946: Undefined variables are fatal with `--locked`
My opam locking issues spree continues...

Supposedly the presence of `with-dev-setup` in package dependencies shouldn't bother opam 2.1.
However, when building [this Dockerfile](https://github.com/goblint/analyzer/blob/bba5c1f9b07d52370a9f2f0d5b5437ce7c6cce10/Dockerfile) (at that commit), the command (not directly visible there)
```sh
opam switch -y create . --deps-only --packages=ocaml-variants.4.14.2+options,ocaml-option-flambda --locked
```
fails with
```
Fatal error: Undefined boolean filter value: with-dev-setup
```

Apparently the undefined variable becomes fatal with `--locked`, rather than defaulting to false. This is still the case on opam 2.5.
Reduced test case from @kit-ty-kate:
```sh
mkdir /tmp/test && cd /tmp/test && (echo 'opam-version: "2.0"' && echo 'depends: "nope" {with-nope}') > test.opam.locked && opam switch create . --deps-only --locked
```

rjbou:
On strict mode, it fails gracefully, and on normal mode, it is "ignored", which is default to false.
fix #6946

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -70,6 +70,7 @@ users)
   * No longer call tar tool to create archives, use tar library instead [#6945 @kit-ty-kate]
 
 ## Lock
+  * [BUG] Fix `undefined variable` error when a lock file filter contains an undefined variables: fail gracefully with strict mode, continue and default the variable to false on normal mode [#6947 @rjbou - fix #6946]
 
 ## Clean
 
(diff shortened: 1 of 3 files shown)


NathanReb (approved):
Looks good to me, thanks!

kit-ty-kate (approved):

PR #6953, opened 2026-06-03
Fix OpamFilename startswith functions

### LINKED ISSUE #6948: `OpamFilename.starts_with` is incorrect
The current implementation of `OpamFilename.starts_with dir file` is a substring check. It can return true even when the dirname of `dir` is a substring of the dirname of `file`.
```
# OpamFilename.starts_with (OpamFilename.Dir.of_string "/a/path") (OpamFilename.of_string "/a/path/to");;
- : bool = true
# OpamFilename.starts_with (OpamFilename.Dir.of_string "/a/path") (OpamFilename.of_string "/a/path-xxx/to");;
- : bool = true
# OpamFilename.starts_with (OpamFilename.Dir.of_string "/a/path/") (OpamFilename.of_string "/a/path-xxx/to");;
- : bool = true

### PR DESCRIPTION
**Pull request by NathanReb**
Fixes #6948 

This PR adds unit tests for `OpamFilename.starts_with` and `OpamFilename.dir_starts_with`, highlighting the existing bugs in its implementation and fixing them:
1. `foo/bar` was considered a prefix of `foo/bar-baz`
2. `/foo/bar` as a dir was considered a prefix of `/foo/bar` as a file.

The PR adds a minimal unit test library/runner. I'm happy to use an existing framework if you feel like the dependency is acceptable!

Backported to 2.5 in https://github.com/ocaml/opam/pull/7003

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -320,3 +320,6 @@ users)
   * `OpamCompat.Seq`: add `to_dispenser` [#6945 @kit-ty-kate]
   * `OpamSystem.directories_with_links`, `OpamSystem.rec_files`, `OpamFilename.rec_files`: add optional `?except_vcs` that default to false to exclude VCS directories [#6945 @kit-ty-kate @rjbou]
   * `OpamFilename.make_tar_gz{_job}`: rename `make_tar_gz_job` into `make_tar_gz` as it no longer need an external process call [#6945 @kit-ty-kate]
+  * `OpamFilename.{,dir_}starts_with`: Fix a bug where `foo/bar` would be considered a prefix of `foo/bar-baz` [#6953 @NathanReb - fix #6948]
+  * `OpamFilename.{,dir_}starts_with`: `/` and `\` are now equivalent on Windows [#6953 @NathanReb]
+  * `OpamFilename.starts_with`: `starts_with "a/b" "a/b"` no longer returns `true` [#6953 @NathanReb]
--- a/master_changes.md
+++ b/master_changes.md
@@ -164,6 +164,7 @@ users)
   * lib/patchDiff: Ensure a more consistent output accross Unix and Windows platforms [#6915 @kit-ty-kate]
   * lib/patchdiff: add dir-file transformations tests [#6915 @rjbou]
   * Add the `lib-tests` target to the main Makefile [#6928 @kit-ty-kate]
+  * Add `opamUnit` as a basic unit test framework [#6953 @NathanReb]
 
 ## Benchmarks
   * Add an even larger real-world diff to benchmark `opam update` [#6567 @kit-ty-kate]
--- a/master_changes.md
+++ b/master_changes.md
@@ -165,6 +165,7 @@ users)
   * lib/patchdiff: add dir-file transformations tests [#6915 @rjbou]
   * Add the `lib-tests` target to the main Makefile [#6928 @kit-ty-kate]
   * Add `opamUnit` as a basic unit test framework [#6953 @NathanReb]
+  * Add unit tests for `OpamFilename.starts_with` and `dir_starts_with` in `tests/lib/core` [#6953 @NathanReb]
 
 ## Benchmarks
   * Add an even larger real-world diff to benchmark `opam update` [#6567 @kit-ty-kate]
--- a/src/core/opamFilename.ml
+++ b/src/core/opamFilename.ml
@@ -344,11 +344,23 @@ let is_exec file =
   with Unix.Unix_error _ ->
     OpamSystem.internal_error "%s does not exist." (to_string file)
 
-let starts_with dirname filename =
-  OpamCompat.String.starts_with ~prefix:(Dir.to_string dirname) (to_string filename)
+let cmpable_dir_string dir =
+  match OpamSystem.forward_to_back (Dir.to_string dir) with
+  | "" -> ""
+  | d -> Filename.concat d ""
 
-let dir_starts_with pfx dir =
-  OpamCompat.String.starts_with ~prefix:(Dir.to_string pfx) (Dir.to_string dir)
+let cmpable_file_string file =
+  OpamSystem.forward_to_back (to_string file)
+
+let starts_with prefix filename =
+  let prefix = cmpable_dir_string prefix in
+  let dir = cmpable_file_string filename in
+  OpamCompat.String.starts_with ~prefix dir
+
+let dir_starts_with prefix dir =
+  let prefix = cmpable_dir_string prefix in
+  let dir = cmpable_dir_string dir in
+  OpamCompat.String.starts_with ~prefix dir
 
 let remove_prefix prefix filename =
   let prefix =
--- a/tests/lib/core/testOpamFilename.ml
+++ b/tests/lib/core/testOpamFilename.ml
@@ -1,0 +1,151 @@
+(**************************************************************************)
+(*                                                                        *)
+(*    Copyright 2026 OCamlPro                                             *)
+(*                                                                        *)
+(*  All rights reserved. This file is distributed under the terms of the  *)
+(*  GNU Lesser General Public License version 2.1, with the special       *)
+(*  exception on linking described in the file LICENSE.                   *)
+(*                                                                        *)
+(**************************************************************************)
+
+let test_starts_with ~ctxt () =
+  let fun_name = "OpamFilename.starts_with" in
+  let test ~on ~prefix_dir ~file ~expected =
+    let test_name = Printf.sprintf "(%s, %s)" prefix_dir file in
+    if List.exists (String.equal Sys.os_type) on then
+      (let prefix_dir' = OpamFilename.raw_dir prefix_dir in
+       let file' = OpamFilename.raw file in
+       let got = OpamFilename.starts_with prefix_dir' file' in
+       if Bool.equal expected got then
+         OpamUnit.success ~ctxt ~fun_name ~test_name ()
+       else
+         let expected = Bool.to_string expected in
+         let got = Bool.to_string got in
+         OpamUnit.failure ~ctxt ~fun_name ~test_name ~expected ~got ())
+    else
+      let reason = Printf.sprintf "Test does not run on %s" Sys.os_type in
+      OpamUnit.skip ~ctxt ~fun_name ~test_name ~reason ()
+  in
+  List.iter
+    (fun (on, prefix_dir, file, expected) ->
+       test ~on ~prefix_dir ~file ~expected)
+    [
+      (* absolute path without parent dirname *)
+      ["Unix"; "Cygwin"], "/no/match", "/foo/bar/file", false;
+      ["Unix"; "Cygwin"], "/foo/bar", "/foo/bar/file", true;
+      ["Unix"; "Cygwin"], "/foo/bar", "/foo/bar/baz/file", true;
+      ["Unix"; "Cygwin"], "/foo/bar", "/foo/bar-baz/file", false;
+      ["Unix"; "Cygwin"], "/foo/bar/", "/foo/bar-baz/file", false;
+      ["Unix"; "Cygwin"], "/foo/bar/file", "/foo/bar/file", false;
+      ["Cygwin"], "/cygdrive/c/foo/bar", "/cygdrive/c/foo/bar/file", true;
+      ["Cygwin"], "/cygdrive/c/foo/bar", "/cygdrive/d/foo/bar/file", false;
+      ["Win32"], {|C:\no\match|}, {|C:\foo\bar\file|}, false;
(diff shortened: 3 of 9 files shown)

### CODE REVIEW

File: src/core/opamFilename.ml [hunk 1]
```diff
@@ -344,16 +344,23 @@ let is_exec file =
   with Unix.Unix_error _ ->
     OpamSystem.internal_error "%s does not exist." (to_string file)
 
-let starts_with dirname filename =
-  OpamCompat.String.starts_with ~prefix:(Dir.to_string dirname) (to_string filename)
+let canonical_dir_string dir =
```
rjbou: maybe just a naming thing, but it doesn't resolve `..` and `./`. `OpamFilename.Dir.of_string` does it, but it is not an invariant. A `dirname` can be constructed via `OpamFilename.raw` which is just the string.

src/core/opamFilename.ml
```diff
@@ -344,16 +344,23 @@ let is_exec file =
...
+  match Dir.to_string dir with
+  | "" -> ""
+  | d -> Filename.concat d ""
 
-let dir_starts_with pfx dir =
-  OpamCompat.String.starts_with ~prefix:(Dir.to_string pfx) (Dir.to_string dir)
+let starts_with prefix filename =
```
rjbou, line 355: `starts_with` doesn't handle `..` pathes, I don't know if it should be resolved in this function, or if we should have a function `is_prefix` that does that resolution.

tests/lib/core/testOpamFilename.ml
```diff
@@ -0,0 +1,81 @@
+let test_starts_with ~ctxt () =
```
rjbou, line 11: I don't know if it worth it, but the 2 test functions are the same except some elements (fun_name, function to call, etc.). Maybe it worth factorising to avoid update in 2 places when needed?

rjbou (commented):
I've added some fixup commits (mostly style + headers).
I've also added some tests about relative paths, cf review comments.

The newly added test framework is working well and a huge good to have (I want to add a lot of tests now). It worth adding some comments to understand the test & the framework.
On the naming of test files, i don't think we need the `test` prefix, we are already in a test directory, and we are testing that module/function. It can't be named simply `opamFilename.ml` as it will clash, but i'm wondering if there is less redundant naming.

src/core/opamFilename.ml
kit-ty-kate: i'd say it is more to the caller to handle that. OpamFilename has no invariant on the presence or not of `..` so usage of `starts_with` can absolutely be `starts_with ".." "../test"`

src/core/opamFilename.ml
kit-ty-kate: I agree. Canonical isn't the right term here. Maybe simply `dir_string` is descriptive enough

src/core/opamFilename.ml
```diff
@@ -344,16 +344,23 @@ let is_exec file =
   with Unix.Unix_error _ ->
     OpamSystem.internal_error "%s does not exist." (to_string file)
 
-let starts_with dirname filename =
-  OpamCompat.String.starts_with ~prefix:(Dir.to_string dirname) (to_string filename)
+let canonical_dir_string dir =
+  match Dir.to_string dir with
+  | "" -> ""
+  | d -> Filename.concat d ""
```
kit-ty-kate, line 350: I'm not sure this is quite right on Windows, e.g. `starts_with "C:\a\b" "C:\a\b/c"` is possible and would've worked with the previous version. Maybe using `OpamSystem.forward_to_back` or something might be the solution although it really depends on the definition/invariants we want to give to these functions and the OpamFilename module as a whole.

Otherwise something like this should also work just fine:
```ocaml
let starts_with prefix filename =
  <old definition> && (filename = "" ||
    let len = String.length prefix in
    len < String.length filename &&
    (filename.[len] = '/' || (Sys.win32 && filename.[len] = '\\'))
```

src/core/opamFilename.ml
NathanReb: I assumed `raw` was discouraged as it could break the `of_string` invariant.

If `raw` is very commonly used I'd be curious to understand why. It might be that a better representation would be to keep the original path string alongside its true canonical form (`..`, `.` and symlinks resolved) as at the moment having to handle mixed `of_string` and `raw` representation feels like jumping through hoops.

That being said, it's definitely a matter for another PR so I'll fix this.

src/core/opamFilename.ml
NathanReb: Again, I assumed `of_string` was the prefered way of instanciating `dirname` and it already handles this, hence why I didn't bother with such cases.

tests/lib/core/testOpamFilename.ml
NathanReb: I'm happy with test code being dumb and straight forward here but I'm happy to factor the code if you'd like as well!

tests/lib/core/testOpamFilename.ml
```diff
@@ -0,0 +1,145 @@
...
+      ["Cygwin"], "/cygdrive/c/foo/bar", "/cygdrive/c/foo/bar/file", true;
+      ["Cygwin"], "/cygdrive/c/foo/bar", "/cygdrive/d/foo/bar/file", false;
+      ["Win32"], {|C:\no\match|}, {|C:\foo\bar\file|}, false;
+      ["Win32"], {|C:\foo\bar|}, {|C:\foo\bar\file|}, true;
+      ["Win32"], {|C:\foo\bar|}, {|D:\foo\bar\file|}, false;
+      ["Win32"], {|C:\foo\bar|}, {|C:\foo\bar\baz\file|}, true;
+      ["Win32"], {|C:\foo\bar|}, {|C:\foo\bar-baz\file|}, false;
+      ["Win32"], {|C:\foo\bar\|}, {|C:\foo\bar-baz\file|}, false;
+      ["Win32"], {|C:\foo\bar\file|}, {|C:\foo\bar\file|}, false;
+      ["Win32"], {|C:\foo/bar|}, {|C:/foo\bar\file|}, true;
```
rjbou:
```suggestion
      ["Win32"], {|C:\foo\bar|}, {|C:\foo\bar/file|}, true;
```

src/core/opamFilename.ml
NathanReb: I think using forward_to_back shoud do the trick shouldn't it? I agree we need to rethink `OpamFilename` and sooner rather than later.

I suggest going with forward_to_back for now!

src/core/opamFilename.ml
NathanReb: renamed to `cmpable_dir_string`

NathanReb:
The tests were using `of_string` instead of `raw` and I think that's what was causing the errors on Windows

NathanReb:
The test failures seem unrelated but I might be wrong!

rjbou:
they are unrelated, see #6978

master_changes.md
```diff
@@ -155,6 +155,9 @@ users)
   * lib/patchDiff: Ensure a more consistent output accross Unix and Windows platforms [#6915 @kit-ty-kate]
   * lib/patchdiff: add dir-file transformations tests [#6915 @rjbou]
   * Add the `lib-tests` target to the main Makefile [#6928 @kit-ty-kate]
+  * Add `opamUnit` as a basic unit test framework [#6953 @NathanReb]
+  * Add unit tests for `OpamFilename.starts_with` and `dir_starts_with` in
+    `tests/lib/core` [#6953 @NathanReb]
```
kit-ty-kate: The 80 columns line limit doesn't apply on master_changes and actually makes my work harder when releasing
```suggestion
  * Add unit tests for `OpamFilename.starts_with` and `dir_starts_with` in `tests/lib/core` [#6953 @NathanReb]
```

master_changes.md
```diff
@@ -296,3 +299,4 @@ users)
   * `OpamCompat.Seq`: add `to_dispenser` [#6945 @kit-ty-kate]
   * `OpamSystem.directories_with_links`, `OpamSystem.rec_files`, `OpamFilename.rec_files`: add optional `?except_vcs` that default to false to exclude VCS directories [#6945 @kit-ty-kate @rjbou]
   * `OpamFilename.make_tar_gz{_job}`: rename `make_tar_gz_job` into `make_tar_gz` as it no longer need an external process call [#6945 @kit-ty-kate]
+  * `OpamFilename.{starts_with/dir_starts_with}`: Fix a bug where `foo/bar` would be considered a prefix of `foo/bar-baz` [#6953 @NathanReb]
```
kit-ty-kate: we tend to use the shell-like syntax for this
```suggestion
  * `OpamFilename.{,dir_}starts_with`: Fix a bug where `foo/bar` would be considered a prefix of `foo/bar-baz` [#6953 @NathanReb]
```

tests/opam-unit/dune
```diff
@@ -0,0 +1,2 @@
+(library
+ (name opamUnit))
```
kit-ty-kate, line 2: personally i'm not a huge fan of adding more complexity to the test system and would've preferred simply using a bunch of `assert`. Is there a particular reason why this was added?

src/core/opamFilename.ml
```diff
@@ -344,16 +344,23 @@ let is_exec file =
...
+let cmpable_dir_string dir =
+  match OpamSystem.forward_to_back (Dir.to_string dir) with
+  | "" -> ""
+  | d -> Filename.concat d ""
 
-let dir_starts_with pfx dir =
-  OpamCompat.String.starts_with ~prefix:(Dir.to_string pfx) (Dir.to_string dir)
+let starts_with prefix filename =
+  let prefix = cmpable_dir_string prefix in
+  let dir = cmpable_dir_string (dirname filename) in
```
kit-ty-kate: This is an unwanted breaking change. `dirname` does not do what we want as it is defined as `t.dirname` which can or cannot be the full directory path depending on what's in `t.filename`.
This behaviour also doesn't correspond to the usage you're gonna do it in your future PR
```suggestion
  let dir = cmpable_dir_string filename in
```

tests/opam-unit/dune
NathanReb: The problem with assert is that the tests stop at the first failure instead of reporting all of them which can make it a bit annoying as you have to iterate through errors until it all works again.

src/core/opamFilename.ml
NathanReb: So is it again a `raw` vs `of_string` problem? I have the impression that `of_string` will always produce a well formed `t.dirname` with only the last component in `filename`. In which other scenario can we break this "invariant which isn't one"?

I'm happy to revert this but then `starts_with (Dir.of_string "/a/b/c") (of_string "a/b/c")` is `true`. I guess it's not super important as it's not very likely that we'll run into such things but this is another bizarerie for `OpamFilename`.

kit-ty-kate: the implicit invariant (`Fpath.is_seg t.basename`) isn't currently upheld everywhere. For example:
https://github.com/ocaml/opam/blob/79d857ccfcb9f96c3976b62eba143eddb7a19dea/src/core/opamFilename.ml#L173-L175

But maybe we should fix these cases instead. I haven't read all usages of starts_with yet to see if adding this new behaviour breaks or fixes things, i'll have a look.

NathanReb: I agree we need to make the function work with what we have right now but I do believe we need to start fixing OpamFilename.

Let's implement your suggestion of reverting the `dirname` thing for now, once we start refactoring OpamFilename we can use newly enforced invariants to rewrite them.

Does that sound good?

src/core/opamFilename.ml
rjbou: > I agree we need to rethink OpamFilename and sooner rather than later.

opened #6986

src/core/opamFilename.ml
```diff
@@ -344,16 +344,26 @@ let is_exec file =
...
+let dir_starts_with prefix dir =
+  let prefix = cmpable_dir_string prefix in
+  let dir = cmpable_dir_string dir in
+  OpamCompat.String.starts_with ~prefix dir
 
 let remove_prefix prefix filename =
-  let prefix =
-    let str = Dir.to_string prefix in
-    if str = "" then "" else Filename.concat str "" in
+  let prefix = cmpable_dir_string prefix in
```
kit-ty-kate: filename isn't guaranteed to contain `\` instead of `/` so it would be best to keep it as-is for now
```suggestion
  let prefix =
    let str = Dir.to_string prefix in
    if str = "" then "" else Filename.concat str "" in
```

NathanReb: I fixed this by using `cmpable_file_string` on the input `filename` as well. I'd prefer this solution as it makes `remove_prefix` consistent with `starts_with`.

kit-ty-kate (approved):

PR #5888, opened 2024-03-19
`opam source --dev` fetching git repositories without `--depth 1`

moyodiallo:
This PR is about to fix #5061. It turns out we doesn't need a flag for fixing the issue, It is agreed during the `opam-dev` meeting to avoid fetching with `--depth 1` for `opam source` because on the client side it could be confusing, an idea from @kit-ty-kate.

- [x] Add an option in the backend
- [x] ~~Add an flag for `opam source` client~~
- [x] opam source is going to be `--deph 1` free when fetching source from dev_repo.
- [x] Add `test` for `opam source` command
- [x] Add test for `opam repository` command, to be sure the behavior remains unchanged.

PR #6955, opened 2026-06-04
Apply opam-specific git config if not full_fetch

### LINKED ISSUE #6451: `opam source --dev repo` sets color.ui to `false`
which I personally find undesirable, just let the user config take over.

```sh
> opam --version
2.3.0
> opam source --dev cmdliner
> git config --local color.ui
false
```

### LINKED ISSUE #7035: `opam source` should do a `git clone` when sourcing git repositories
We use the same code to internally clone packages sources and `opam source`. The resulting `opam source` directory is meant to be used  by users as if they did a git clone, we shouldn't apply opam internals specific configuration : as branch naming, git cache, depth (already removed), etc.

related to #6955 #6451 #6992

sporkl:
`opam source --dev` should not apply opam-specific `git config`s. This commit adds a `will_full_fetch` argument to VCS.init, and only applies opam-specific git configurations when this is false. Also, default to full_fetch=false to mimic prior behavior. This should resolve #6451 .

I do have some failing tests locally, but it looks like the same tests fail on master for me.

Relatively new to contributing to open source, so any feedback is appreciated! I have a fair amount of free time these days, so looking to contribute more to the OCaml ecosystem.

### DIFF
--- a/src/repository/opamGit.ml
+++ b/src/repository/opamGit.ml
@@ -41,28 +41,39 @@ module VCS : OpamVCS.VCS = struct
       let env = env () in
       OpamSystem.make_command ~env ?verbose ?stdout "git" ("-C"::dir::args)
 
-  let init repo_root repo_url =
+  let init ?(for_source = false) repo_root repo_url =
     OpamFilename.mkdir repo_root;
-    OpamProcess.Job.of_list [
-      git repo_root [ "init"; "--initial-branch=main" ];
-      (* Enforce this option, it can break our use of git if set *)
-      git repo_root [ "config" ; "--local" ; "fetch.prune"; "false"];
-      (* We reset diff.noprefix to ensure we get a `-p1` patch and avoid <https://github.com/ocaml/opam/issues/3627>. *)
-      git repo_root [ "config" ; "--local" ; "diff.noprefix"; "false"];
-      (* Disable automatic line-ending conversion and switch core.eol to Unix.
-         THIS DOES NOT MEAN ALL FILES GET LF-ONLY LINE-ENDINGS!
-         This combination of settings means that files will be checked out
-         exactly as they appear in the repository, so if files are checked in
-         with CRLF line-endings (either by not having .gitattributes with
-         core.autocrlf = false, or having an explicit eol=crlf in
-         .gitattributes), then they will still be checked out with CRLF endings.
-       *)
-      git repo_root [ "config" ; "--local" ; "core.autocrlf"; "false"];
-      git repo_root [ "config" ; "--local" ; "core.eol"; "lf"];
-      git repo_root [ "config" ; "--local" ; "color.ui"; "false" ];
-      (* Document the remote for user-friendliness (we don't use it) *)
-      git repo_root [ "remote"; "add"; "origin"; OpamUrl.base_url repo_url ];
-    ] @@+ function
+    let if_not_for_source =
+      if not for_source then Option.some else Fun.const Option.none
+    in
+    (OpamProcess.Job.of_list @@ List.filter_map Fun.id [
+        Some (git repo_root [ "init"; "--initial-branch=main" ]);
+        (* Enforce this option, it can break our use of git if set *)
+        if_not_for_source
+          (git repo_root [ "config" ; "--local" ; "fetch.prune"; "false"]);
+        (* We reset diff.noprefix to ensure we get a `-p1` patch and avoid
+           https://github.com/ocaml/opam/issues/3627 *)
+        if_not_for_source
+          (git repo_root [ "config" ; "--local" ; "diff.noprefix"; "false"]);
+        (* Disable automatic line-ending conversion and switch core.eol to Unix.
+           THIS DOES NOT MEAN ALL FILES GET LF-ONLY LINE-ENDINGS!
+           This combination of settings means that files will be checked out
+           exactly as they appear in the repository, so if files are checked in
+           with CRLF line-endings (either by not having .gitattributes with
+           core.autocrlf = false, or having an explicit eol=crlf in
+           .gitattributes), then they will still be checked out with CRLF
+           endings.
+        *)
+        if_not_for_source
+          (git repo_root [ "config" ; "--local" ; "core.autocrlf"; "false" ]);
+        if_not_for_source
+          (git repo_root [ "config" ; "--local" ; "core.eol"; "lf" ]);
+        if_not_for_source
+          (git repo_root [ "config" ; "--local" ; "color.ui"; "false" ]);
+        (* Document the remote for user-friendliness (we don't use it) *)
+        Some (git repo_root
+                [ "remote"; "add"; "origin"; OpamUrl.base_url repo_url ]);
+      ]) @@+ function
     | None -> Done ()
     | Some (_,err) -> OpamSystem.process_error err
 
--- a/src/repository/opamGit.ml
+++ b/src/repository/opamGit.ml
@@ -77,6 +77,27 @@ module VCS : OpamVCS.VCS = struct
     | None -> Done ()
     | Some (_,err) -> OpamSystem.process_error err
 
+  let clone ?(full_fetch = true) repo_root url =
+    OpamFilename.mkdir repo_root;
+    if full_fetch then
+      git repo_root [ "clone"; OpamUrl.base_url url; "." ]
+      @@> fun r ->
+      OpamSystem.raise_on_process_error r;
+      git repo_root
+        [ "submodule"; "update"; "--init"; "--recursive" ]
+      @@> fun r ->
+      OpamSystem.raise_on_process_error r;
+      Done ()
+    else
+      git repo_root [ "clone"; OpamUrl.base_url url; "."; "--depth"; "1" ]
+      @@> fun r ->
+      OpamSystem.raise_on_process_error r;
+      git repo_root
+        [ "submodule"; "update"; "--init"; "--recursive"; "--depth"; "1" ]
+      @@> fun r ->
+      OpamSystem.raise_on_process_error r;
+      Done ()
+
   let remote_ref url =
     match url.OpamUrl.hash with
     | Some h -> "refs/remotes/opam-ref-"^h
--- a/src/repository/opamVCS.ml
+++ b/src/repository/opamVCS.ml
@@ -15,7 +15,7 @@ open OpamProcess.Job.Op
 module type VCS = sig
   val name: OpamUrl.backend
   val exists: dirname -> bool
-  val init: dirname -> url -> unit OpamProcess.job
+  val init: ?for_source:bool -> dirname -> url -> unit OpamProcess.job
   val fetch:
     ?full_fetch:bool -> ?cache_dir:dirname -> ?subpath:subpath -> dirname -> url ->
     unit OpamProcess.job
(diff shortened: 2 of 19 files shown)


rjbou:
Thanks for this first contribution! We'll try to take a look soon. For the moment we are trying to finalise 2.6 alpha release.

src/repository/opamGit.ml
```diff
@@ -32,37 +32,46 @@ module VCS : OpamVCS.VCS = struct
...
+      | None -> Done ()
+      | Some (_,err) -> OpamSystem.process_error err
 
   let remote_ref url =
     match url.OpamUrl.hash with
     | Some h -> "refs/remotes/opam-ref-"^h
     | None -> "refs/remotes/opam-ref"
 
-  let fetch ?(full_fetch = true) ?cache_dir ?subpath repo_root repo_url =
+  let fetch ?(full_fetch = false) ?cache_dir ?subpath repo_root repo_url =
```
kit-ty-kate: best to keep the default as-is to avoid unnecessary breaking changes
```suggestion
  let fetch ?(full_fetch = true) ?cache_dir ?subpath repo_root repo_url =
```

src/repository/opamGit.ml
```diff
@@ -32,37 +32,46 @@ module VCS : OpamVCS.VCS = struct
     fun ?verbose ?stdout args ->
       OpamSystem.make_command ?verbose ?stdout "git" ("-C"::dir::args)
 
-  let init repo_root repo_url =
+  let init ?(will_full_fetch = false) repo_root repo_url =
```
kit-ty-kate: i think it makes more sense to use the same name for the parameter as for `fetch`
```suggestion
  let init ?(full_fetch = true) repo_root repo_url =
```

tests/reftests/action-disk.test
```diff
@@ -1639,7 +1639,7 @@ FILE(switch-state)              Wrote ${BASEDIR}/OPAM/install-from-git-pin-all/.
 + git "-C" "${BASEDIR}/OPAM/install-from-git-pin-all/.opam-switch/sources/main-gpin" "config" "--local" "color.ui" "false"
 + git "-C" "${BASEDIR}/OPAM/install-from-git-pin-all/.opam-switch/sources/main-gpin" "remote" "add" "origin" "file://${BASEDIR}/main-gpin"
 + git "-C" "${BASEDIR}/OPAM/install-from-git-pin-all/.opam-switch/sources/main-gpin" "remote" "set-url" "origin" "file://${BASEDIR}/main-gpin"
-+ git "-C" "${BASEDIR}/OPAM/install-from-git-pin-all/.opam-switch/sources/main-gpin" "fetch" "-q" "file://${BASEDIR}/main-gpin" "--update-shallow" "+master:refs/remotes/opam-ref-master"
++ git "-C" "${BASEDIR}/OPAM/install-from-git-pin-all/.opam-switch/sources/main-gpin" "fetch" "-q" "file://${BASEDIR}/main-gpin" "--update-shallow" "--depth=1" "+master:refs/remotes/opam-ref-master"
```
kit-ty-kate: an opam install should not end up doing `depth=1`, this would be a regression (i suspect reverting the default value change should fix that)

src/repository/opamGit.ml
```diff
@@ -32,37 +32,46 @@ module VCS : OpamVCS.VCS = struct
...
-      git repo_root [ "config" ; "--local" ; "core.autocrlf"; "false"];
-      git repo_root [ "config" ; "--local" ; "core.eol"; "lf"];
-      git repo_root [ "config" ; "--local" ; "color.ui"; "false" ];
-      (* Document the remote for user-friendliness (we don't use it) *)
-      git repo_root [ "remote"; "add"; "origin"; OpamUrl.base_url repo_url ];
-    ] @@+ function
-    | None -> Done ()
-    | Some (_,err) -> OpamSystem.process_error err
+    if will_full_fetch then
+      OpamProcess.Job.of_list [
```
kit-ty-kate: To avoid duplication and make the diff easier to read, we can simply use `filter_map`
```suggestion
    OpamProcess.Job.of_list @@ List.filter_map Fun.id [
```

tests/reftests/source.test
```diff
@@ -58,6 +58,7 @@ pandore/.git
 pandore/root.ml
 pandore/src
 pandore/src/src.ml
+### cat pandore/.git/config | grep "ui = false"
```
kit-ty-kate: I'm not sure it's the best regression test for this. opam could not set this value in the future, git could also change the number of spaces or even the syntax of the file or something and it would be completely invisible.

sporkl:
Thank you for the feedback! I've switched back to defaulting full_fetch to true, as before. This does break compatibility with reftest-download; I'm not confident what the correct behavior is here.

I think I understood the suggestion about using List.filter_map, and the other stuff should be addressed as well, but please let me know if I should change anything.

Apologies for the commit spam; I did a rebase before pushing but I guess I shouldn't do that?

kit-ty-kate:
> Apologies for the commit spam; I did a rebase before pushing but I guess I shouldn't do that?

no, rebases are most welcome (we would do one anyway if needed before merging), but here it doesn't seem like you did a rebase, rather some type of fast-forward merge. Could you try again with:
```
git pull --rebase origin master
git push --force-with-lease <your-branch> conditional-git-config
```

tests/reftests/download.test
```diff
@@ -296,11 +291,6 @@ SYSTEM                          mkdir ${OPAMTMP}
 SYSTEM                          mkdir ${BASEDIR}/OPAM/download/.opam-switch/sources/qux.1
 Processing  1/1: [qux.1: git]
 + git "-C" "${BASEDIR}/OPAM/download/.opam-switch/sources/qux.1" "init"
-+ git "-C" "${BASEDIR}/OPAM/download/.opam-switch/sources/qux.1" "config" "--local" "fetch.prune" "false"
-+ git "-C" "${BASEDIR}/OPAM/download/.opam-switch/sources/qux.1" "config" "--local" "diff.noprefix" "false"
-+ git "-C" "${BASEDIR}/OPAM/download/.opam-switch/sources/qux.1" "config" "--local" "core.autocrlf" "false"
-+ git "-C" "${BASEDIR}/OPAM/download/.opam-switch/sources/qux.1" "config" "--local" "core.eol" "lf"
-+ git "-C" "${BASEDIR}/OPAM/download/.opam-switch/sources/qux.1" "config" "--local" "color.ui" "false"
```
kit-ty-kate, line 303: this shouldn't indeed change. The issue is that you're using an existing `full_fetch` variable (introduced in https://github.com/ocaml/opam/pull/5888) which doesn't distinguish between `opam source` vs. other uses as you're trying to make happen, but rather between "uses that needs the full git history" and "uses that just needs the latest commit, aka. --depth=1".

What you want to have is instead to introduce a new parameter or something like that that reflect what you are trying to do.

tests/reftests/source.test
```diff
@@ -58,6 +58,8 @@ pandore/.git
 pandore/root.ml
 pandore/src
 pandore/src/src.ml
+### git -C pandore config --local fetch.prune
```
kit-ty-kate: instead of this brittle pattern, while not just have something like
```
### opam source -vv <some-git-pkg> | sed-cmd git | grep -v '^- '
```
This way every git command called would be displayed and it would be easy to detect any regression

tests/reftests/download.test
sporkl: Makes sense. I think I was thrown off by `full_fetch` being set to true in calls to `opam source --dev` in #5888 .

sporkl:
Working on updating the test as suggested, and I realized that while `opam source --dev ...` avoids the opam-specific git configuration as expected, `opam source ...` still includes the the git configuration stuff. I'm inclined to think that `opam source` should also avoid applying any specific git configuration in general, but want to verify that this behavior change makes sense?

(In this case it's going through a call to `OpamUpdate.download_package_source` then `OpamRepository.pull_shared_tree`; as far as I can tell, these functions aren't called anywhere else)

kit-ty-kate:
> I'm inclined to think that `opam source` should also avoid applying any specific git configuration in general, but want to verify that this behavior change makes sense?

I agree

kit-ty-kate (approved):
I've squashed your commits and split it between test addition and fix+test update (helps to show what changed), renamed `from_source` to `for_source` (makes a bit more sense when using the code), and reindented the code to avoid long lines etc (we try to avoid going beyond 80 columns)

Does that look ok for you? If so i'm happy to merge this

sporkl:
Looks good to me! I was wondering about changes along the same lines. Thanks for the help, I'm learning embarrassingly much

src/repository/opamGit.ml
```diff
@@ -32,28 +32,39 @@ module VCS : OpamVCS.VCS = struct
     fun ?verbose ?stdout args ->
       OpamSystem.make_command ?verbose ?stdout "git" ("-C"::dir::args)
 
-  let init repo_root repo_url =
+  let init ?(for_source = false) repo_root repo_url =
```
rjbou, line 44: The naming of the new argument `for_source` (and the internal `if_not_for_source`) is not enough generic. In this PR, it is added because we don't want config when it originates from `opam source`. But when using the API, it is not evident that it will have an effect of dropping the config step (it is written in the doc), and if one day we use it for another process, it will lose its meaning. More generally, at `opam-repository` level, the functions should not be aware of the command that called them (even if their code does things accordingly).
It can be a more generic `no_config` for example. I can do the change.

sporkl: I think this sounds good, `no_config` makes sense to me.

src/repository/opamVCS.ml
```diff
@@ -109,7 +110,7 @@ module Make (VCS: VCS) = struct
         Done (Result None)
     else
```
kit-ty-kate, line 120: https://github.com/ocaml/opam/pull/6992 shows that there are other problems with the current `opam source`, e.g. branches are not fetched, the current branch is set to whatever the default name is on your machine instead of the default branch of the remote git repository, a spurious `opam-ref` remote branch is created, …

I think we should instead do something entirely different in the case of `for_source` (the name in this case makes more sense than `no_config` i would argue), and simply use `git clone` etc:
```suggestion
    else if for_source then
      VCS.clone dirname url @@+ fun () ->
      Done (Result None)
    else
```

sporkl: I think we can get the default latest commit of a remote repo with `git ls-remote origin HEAD`, and use this in the implementation of fetch, to something branch-agnostic without needing to add a new `clone` function. Working on giving this a try, though I'm not convinced that it's necessarily better than using `clone`.

sporkl: We can specifically get the branch name with `git ls-remote --symref origin HEAD`, I  guess the place to use this would be `reset_tree` rather than `fetch`, and to rename the current local branch. I'm suspicious of needing to do branch renaming, so I am leaning towards adding `clone` as suggested. Also avoids the need to parse shell output to get the branch names.

sporkl:
Looking into implementing clone for darcs and mercurial, which I'm not as familiar with. It looks like there aren't any reftests for them yet. I'm thinking of adding some, testing both basic installation and also that `source` works correctly. Does this make sense? It would introduce mercurial and darcs as dev dependencies to get the tests to pass (though maybe can update the build to work around this).

kit-ty-kate:
rebased

sporkl:
Couple changes here. I consolidated the `opam source` git command tests a bit, added hg and darcs reftests for package installation and source, and added `clone` for each VCS.

kit-ty-kate (changes requested):
Since the code significantly changed i'll have to re-review it and we won't have time to review it before the 2.6.0~alpha1 freeze, so bumping to the 2.7 milestone

PR #6957, opened 2026-06-06
bugfix: avoid compiler setup for opam init --bare

### LINKED ISSUE #5815: opam init --bare wastes unnecessary time selecting a compiler
This is a regression compared to opam 2.0.
opam 2.1 also exhibits this issue as well as opam master (2.2.0~beta1):
```
$ opam init -n --bare --debug
[...]
00:08.106  GSTATE                          Failed to evaluate global variable sys-ocaml-version: "ocamlc": command not found.
00:13.974  CLIENT                          Selected default compiler ocaml-base-compiler in 6.143s
```
`--bare` should prevent opam from looking for a default compiler, since none will be installed.

tuesdayjz:
Fixes #5815 - `opam init --bare` was spending a few seconds selecting a default compiler, which is inappropriate for `--bare` that don't need a compiler environment.

Changes:
- Added a new ?no_compiler:bool parameter to OpamClient
- Added logic to skip compiler setup when no_compiler=true
- Updated init function to pass the no_compiler flag from the CLI

verification:
- before
```log
...
00:05.622  CACHE(repository)      Writing the repository cache to ~/.opam/repo/state-3DDC2EF7.cache ...
00:05.796  CACHE(repository)      ~/.opam/repo/state-3DDC2EF7.cache written in 0.174s
00:08.265  CLIENT                 Selected default compiler ocaml-base-compiler in 2.469s
```

- after
```log
...
00:06.044  CACHE(repository)      Writing the repository cache to ~/.opam/repo/state-3DDC2EF7.cache ...
00:06.213  CACHE(repository)      ~/.opam/repo/state-3DDC2EF7.cache written in 0.169s
```

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -218,6 +218,7 @@ users)
   * Add a test making sure the global or system `git` config doesn't change the behaviour of opam [#6992 @kit-ty-kate]
   * Add a test showing the remote and branch names of a git repository extracted by `opam source` [#6992 @kit-ty-kate]
   * Add a test showing the order of install actions for each relevant commands [#6864 @kit-ty-kate]
+  * Add a test showing some of the internal steps of `opam init` [#6957 @kit-ty-kate]
 
 ### Engine
   * Add `http-server` to launch a minimal http server [#6939 @rjbou]
--- a/master_changes.md
+++ b/master_changes.md
@@ -21,6 +21,7 @@ users)
 
 ## Init
   * ✘ Display an appropriate error message when the file given to `opam init --config` does not exist or is in a VCS. This changes the behaviour for VCS local urls that was previously retrieved. [#5979 @kit-ty-kate - fix #5971]
+  * Improve the performance of `opam init --bare` by no longer computing the default compiler in that case [#6957 @tuesdayjz @kit-ty-kate - fix #5815]
 
 ## Config report
 
@@ -276,6 +277,7 @@ users)
 ## opam-client
   * `OpamArg`: add `build_options_no_depexts` getter to retrieve the value of the given flag  [#6489 @rjbou]
   * `OpamArg{,Tools}.cli2_6` was added [#6978 @kit-ty-kate]
+  * `OpamClient.init`: a new `?no_compiler` argument was added to not compute the default compiler and the list of atom returned is now an option to separate empty invariant from no switch [#6957 @kit-ty-kate]
   * `OpamClientConfig.opam_init`: replace `no_depexts` argument by `depexts` [#6489 @rjbou]
   * `OpamSolution` remove the heuristic of recomputing depexts of additional (pinned) packages. [#6489 @arozovyk]
   * `OpamClient` update the system package status check for dependencies during `opam install --deps-only`, including support for pinned packages; also update this in `OpamAuxCommands.autopin` [#6489 @arozovyk]
--- a/src/client/opamClient.ml
+++ b/src/client/opamClient.ml
@@ -1802,6 +1802,7 @@ let init
     ?dot_profile ?update_config ?env_hook ?(completion=true)
     ?(check_sandbox=true)
     ?cygwin_setup ?git_location
+    ?(no_compiler=false)
     shell =
   log "INIT %a"
     (slog @@ OpamStd.Option.to_string OpamRepositoryBackend.to_string) repo;
@@ -1840,7 +1841,7 @@ let init
         OpamConsole.msg
           "... but you have no switches installed, use `opam switch \
            create <compiler-or-version>' to get started.";
-      gt, OpamRepositoryState.load `Lock_none gt, []
+      gt, OpamRepositoryState.load `Lock_none gt, None
     ) else (
       if not root_empty then (
         OpamConsole.warning "%s exists and is not empty"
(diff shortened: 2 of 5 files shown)


src/client/opamClient.ml
```diff
@@ -1917,7 +1918,7 @@ let init
            OpamConsole.error_and_exit `Sync_error
              "Initial download of repository failed.");
         let default_compiler =
-          if dontswitch then [] else
+          if dontswitch || no_compiler then [] else
```
kit-ty-kate:
```suggestion
          if dontswitch then Some []
          else if no_compiler then None
          else
```

src/client/opamClient.mli
```diff
@@ -30,6 +30,7 @@ val init:
   ?check_sandbox:bool ->
   ?cygwin_setup: [ `internal of OpamSysPkg.t list | `default_location | `location of dirname | `no ] ->
   ?git_location:(dirname, unit) either ->
+  ?no_compiler:bool ->
   shell ->
   rw global_state * unlocked repos_state * atom list
```
kit-ty-kate:
```suggestion
  rw global_state * unlocked repos_state * atom list option
```

src/client/opamCommands.ml
```suggestion
    match default_compiler with
    | None -> ()
    | Some default_compiler ->
```

kit-ty-kate (commented):
Thanks a lot! While the PR as proposed is good, to help making the code a bit more maintainable over a longer period i feel like it would be better so take the time to change slightly the API to add the new case.

This PR changes the `OpamClient.init` function in such a way that one of the result is linked to the new `?no_compiler` parameter. Changing the type of this result to avoid the implicit invariant when using the function should make the code simpler to understand: the function can return a list of compiler packages, an empty list (empty switch) or None (no switch at all)

kit-ty-kate (approved):
Note for reviewers: enable the "Hide whitespace" option

src/client/opamClient.ml
```diff
@@ -1917,31 +1918,33 @@ let init
...
-            List.find_opt
-              (OpamSolver.atom_coinstallability_check univ)
-            alternatives
-            |> OpamStd.Option.default []
-          in
-          log "Selected default compiler %s in %0.3fs"
-            (OpamFormula.string_of_atoms default_compiler)
-            (chrono ());
-          default_compiler
+          if dontswitch then Some []
```
rjbou, line 1921: minor: it is possible to avoid the reindent only by having the first 1/3 lines changed.

src/client/opamClient.mli
```diff
@@ -30,8 +30,9 @@ val init:
   ?check_sandbox:bool ->
   ?cygwin_setup: [ `internal of OpamSysPkg.t list | `default_location | `location of dirname | `no ] ->
   ?git_location:(dirname, unit) either ->
+  ?no_compiler:bool ->
```
rjbou, line 33: It worth updating the documentation with the new ~~field~~ optional argument

master_changes.md
```diff
@@ -21,6 +21,7 @@ users)
 
 ## Init
   * ✘ Display an appropriate error message when the file given to `opam init --config` does not exist or is in a VCS. This changes the behaviour for VCS local urls that was previously retrieved. [#5979 @kit-ty-kate - fix #5971]
+  * Improve the performance of `opam init --bare` [#6957 @tuesdayjz @kit-ty-kate - fix #5815]
```
```suggestion
  * Improve the performance of `opam init --bare` by no longer computing the default compiler in that case[#6957 @tuesdayjz @kit-ty-kate - fix #5815]
```

rjbou: There is missing API changes

rjbou (commented):
TODO : add a test to highlight the change

src/client/opamClient.mli
kit-ty-kate: the documentation doesn't document the other arguments either currently so i'd rather not do that and leave it to do later

kit-ty-kate:
done and rebased

src/client/opamClient.mli
rjbou: Which usually means never. We are not enough for the amount of work and this kind of minor update usually are dropped and forgotten. That's why i'm in favor of updating, even if it is partial (reinstall have only 1 argument documented), it's a tiny step, and would encourage next changes to update the doc instead of "there is no doc, no need to update it".
Anyway, I've opened #7036

tests/reftests/init.test
```diff
@@ -402,3 +402,17 @@ this is not the content of a config file
...
+### <opamrc>
+opam-version: "2.0"
+### OPAMDEBUGSECTIONS=CLIENT OPAMDEBUG=-5 opam init --bypass-checks --bare --no-setup --config opamrc REPO/
+Configuring from ${BASEDIR}/opamrc and then from built-in defaults.
+CLIENT                          INIT default from file://${BASEDIR}/REPO
+CLIENT                          updating repository state
+
+<><> Fetching repository information ><><><><><><><><><><><><><><><><><><><><><>
+[default] Initialised
+CLIENT                          Selected default compiler ocaml-base-compiler in 0.000s
```
rjbou:
```suggestion
```
For the last commit

master_changes.md
```diff
@@ -273,6 +275,7 @@ users)
 ## opam-client
   * `OpamArg`: add `build_options_no_depexts` getter to retrieve the value of the given flag  [#6489 @rjbou]
   * `OpamArg{,Tools}.cli2_6` was added [#6978 @kit-ty-kate]
+  * `OpamClient.init`: a new `?no_compiler` argument was added and the list of atom returned is now an option to separate empty invariant from no switch [#6957 @kit-ty-kate]
```
```suggestion
  * `OpamClient.init`: a new `?no_compiler` argument was added to not compute the default compiler and the list of atom returned is now an option to separate empty invariant from no switch [#6957 @kit-ty-kate]
```

rjbou (approved):
Modulo green ci, and reftest, lgtm. thanks!

kit-ty-kate:
Thanks!

tuesdayjz:
Thank you guys for fixing! Sorry to keep ignoring the comments for a month...

kit-ty-kate:
no problem at all

PR #6960, opened 2026-06-08
GHA: fix depexts jobs

rjbou:
* ubuntu job: remove ocaml-compiler-libs requirement, package not found
* opensuse job: force resolution in case of conflicts, there is a conflicts between ocaml & ncurses-devel packages


NathanReb (approved):

rjbou:
Oups, merged even though the openSUSE was failing. Github was showing all the test green, discovered this while following after merge jobs.

PR #6961, opened 2026-06-08
reftest: fix repository http test

rjbou:
Timestamps are added in repo file by 'opam admin index', which made the test non reproducible.

It was unnoticed in #6939. It can be seen only on some windows jobs (with a cache), because of some latency, and only when the minute changes between the beginning & the end of the test.

For the moment, to unlock CI, the test is updated. In the future, maybe we want some option to 'opam admin index' to not embed timestamp, or even an environment variable for tests that disables it.

### DIFF
--- a/tests/reftests/repository-http.test
+++ b/tests/reftests/repository-http.test
@@ -18,11 +18,22 @@ opam-version:"2.0"
 opam-version:"2.0"
 ### <packages/empat/empat.4/opam>
 opam-version:"2.0"
+### # We need this script to remove automatically embed timestamp in repo file
+### # by 'opam admin index'
+### <remove-timestamp.sh>
+set -eu
+cat > repo << EOF
+opam-version:"2.0"
+EOF
+gzip -d index.tar.gz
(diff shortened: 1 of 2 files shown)


tests/reftests/repository-http.test
```diff
@@ -18,11 +18,22 @@ opam-version:"2.0"
 opam-version:"2.0"
 ### <packages/empat/empat.4/opam>
 opam-version:"2.0"
+### # We need this script to remove automatically embed timestamp in repo file
+### # by 'opam admin index'
+### <remove-timestamp.sh>
```
NathanReb, line 23: How about including the original `opam admin index` in this script so that we'd only have to call the script to generate the right files?

rjbou: `opam` called at toplevel is replaced by the path of the currently built one, `opamMain.exe` in `_build`. If it is inside a script, it would be the one in the path that would be called.

NathanReb (approved):

PR #6973, opened 2026-06-25
Packaging.md: add a paragraph explaining that static archives are preferrable when manually releasing packages

mseri:
See: https://github.com/ocaml/opam-repository/issues/30089#issuecomment-4801321415

Not sure if it is worth having this in the master_changes file

### DIFF
--- a/doc/pages/Packaging.md
+++ b/doc/pages/Packaging.md
@@ -114,6 +114,11 @@ url {
 }
 ```
 
+It is recommended to use a static archive for the `src:`. Any change in the
+archive (such as those that can happen when using archives generated
+by git forges such as GitHub) can lead to changes in the checksums and
+consequently break the package on opam-repository.
+
 Then get to https://github.com/ocaml/opam-repository and select `Fork` on the
 top-right. Clone the resulting repository, add your package definition, and
 push back, as such:
--- a/master_changes.md
+++ b/master_changes.md
@@ -214,6 +214,7 @@ users)
   * Bump `actions/cache` to version 6 [#6983 @kit-ty-kate]
 
 ## Doc
+  * Update the packaging page to recommend using static archives for releases [#6973 @mseri]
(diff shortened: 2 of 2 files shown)


doc/pages/Packaging.md
```diff
@@ -114,6 +114,11 @@ url {
 }
 ```
 
+It is recommended to use a static archive for the `src:`, since unexpected
+changes in automatic archive generation (e.g. when a repository is migrated
+to a different organization) can lead to changes in the checksums and
```
kit-ty-kate: this part feels a bit weird to me without a description of what automatic archive generation is and explaining their link to forges/git-frontends. For example, you come from a background where you host your archives yourself this paragraph doesn't really make sense.

kit-ty-kate:
> Not sure if it is worth having this in the master_changes file

also yes, pretty much everything is added to the master_changes. Its goal is to have a temporary file to put everything that was done, that is then processed later (removing things that doesn't make sense in the changelog, merging things together to explain them better, …) once we do a release.

doc/pages/Packaging.md
mseri: Would you just cut it short or would you rather have a more elaborate description (like "if you host your code on github, gitlab or similar services, they can automatically generate archives for tags and commits ...)?

doc/pages/Packaging.md
```diff
@@ -114,6 +114,11 @@ url {
...
+consequently break the package on the opam-repository.
```
kit-ty-kate: maybe something like that?
```suggestion
It is recommended to use a static archive for the `src:`. Any change in the
archive (such as those that can happen when using archives generated
by git forges such as GitHub) can lead to changes in the checksums and
consequently break the package on opam-repository.
```

kit-ty-kate (approved):

kit-ty-kate:
Thanks!

PR #6977, opened 2026-06-26
Workaround symlinks in local opam repositories

dra27-js:
We got caught out by a caching problem in the CI for [oxcaml/opam-repository](https://github.com/oxcaml/opam-repository) because it creates an opam local switch in the checkout for testing. First time round it all worked, but if the cache was restored then the repo being added in setup-ocaml failed because of the symlinks in `_opam`.

This PR presents a few ideas: the simplest one is first which is to extend `rsync`'s ignoring of `_opam*`, `.#*` and `_build`. I also added two possible ideas, given that it's for repos only, either dereferencing or just ignoring the symlinks... I'm not totally convinced by either, though.

### DIFF
--- a/src/core/opamSystem.ml
+++ b/src/core/opamSystem.ml
@@ -123,8 +123,13 @@ let get_files_t ~except_vcs dirname =
   let rec aux files =
     match Unix.readdir dir with
     | "." | ".." -> aux files
-    | ".git" | ".hg" | "_darcs" when except_vcs -> aux files
-    | file -> aux (file :: files)
+    | ".git" | ".hg" | "_darcs" | "_build" when except_vcs -> aux files
+    | file ->
+      let has_prefix prefix = OpamCompat.String.starts_with ~prefix file in
+      if except_vcs && (has_prefix "_opam" || has_prefix ".#") then
+        aux files
+      else
+        aux (file :: files)
     | exception End_of_file -> files
   in
   let files = aux [] in
--- a/src/core/opamSystem.mli
+++ b/src/core/opamSystem.mli
@@ -77,7 +77,8 @@ val copy_file: string -> string -> unit
 val copy_dir: string -> string -> unit
 
 (** Same as [copy_dir] except it avoids copying VCS directories
-    ([.git], [.hg], [_darcs]) *)
+    ([.git], [.hg], [_darcs]), opam switches ([_opam*]), [_build]
+    and editor lock files ([.#]) *)
 val copy_dir_except_vcs : string -> string -> unit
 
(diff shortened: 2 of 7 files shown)


kit-ty-kate:
Ignoring `_opam`, `.#` etc. seems good, but i don't really like the idea of ignoring symlinks. What was your issue with symlinks? If the `_opam` is ignored in the first place it shouldn't appear in `get_diff` or am i missing something?

As for rsync, ideally i'd like to get rid of it in the future so i suppose this goes in this direction, nice.

kit-ty-kate:
answering to my own question: oh sorry i guess i didn't correct read `I also added two possible ideas, [...]` as "here's a list of commits to choose from".

So yes, personally i would only pick the two first commits and would avoid the last three

src/core/opamSystem.ml
```diff
@@ -123,8 +123,13 @@ let get_files_t ~except_vcs dirname =
   let rec aux files =
     match Unix.readdir dir with
     | "." | ".." -> aux files
-    | ".git" | ".hg" | "_darcs" when except_vcs -> aux files
-    | file -> aux (file :: files)
+    | ".git" | ".hg" | "_darcs" | "_build" when except_vcs -> aux files
+    | file ->
+      let has_prefix prefix = OpamCompat.String.starts_with ~prefix file in
+      if except_vcs && (has_prefix "_opam" || has_prefix ".#") then
```
kit-ty-kate, line 129: just to avoid the unnecessary closure when `~except_vcs:true`:
```suggestion
      let has_prefix prefix s = OpamCompat.String.starts_with ~prefix s in
      if except_vcs && (has_prefix "_opam" file || has_prefix ".#" file) then
```

src/core/opamSystem.mli
```diff
@@ -77,7 +77,8 @@ val copy_file: string -> string -> unit
 val copy_dir: string -> string -> unit
 
 (** Same as [copy_dir] except it avoids copying VCS directories
-    ([.git], [.hg], [_darcs]) *)
+    ([.git], [.hg], [_darcs]), opam switches ([_opam*]), [_build]
+    and editor lock files ([.#]) *)
```
kit-ty-kate, line 81: Technically this diverges from the `OpamLocal.rsync` behaviour where `_opam*`, `_build` and `.#*` are always ignored regardless of `~except_vcs`.

Idk if we'd want to add a new parameter or rename the function or something like that to show the new behaviour

kit-ty-kate (commented):
looks good on principle, but i think it's worth discussing a little more first

PR #6978, opened 2026-06-26
Stop the opam 1.2 to 2.0 repository upgrade process from downloading packages without checksums to add a non-trusted md5

kit-ty-kate:
This made the result of `opamrt-big-upgrade.test` depend on whether or not `caml.inria.fr` is down or not.

### DIFF
--- a/src/client/opamAdminCommand.ml
+++ b/src/client/opamAdminCommand.ml
@@ -781,11 +781,8 @@ let upgrade_command cli =
   ]
   in
   let clear_cache_arg =
-    OpamArg.mk_flag ~cli OpamArg.cli_original ["clear-cache"]
-      (Printf.sprintf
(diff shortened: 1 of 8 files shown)


kit-ty-kate:
~Side note for reviewer: the cygwin failure is expected (see https://github.com/ocaml/dune/issues/15340). I'm not exactly sure how we'll fix that yet but we can ignore it for now~ Fixed in #6979

src/client/opamAdminCommand.ml
```diff
@@ -780,13 +780,6 @@ let upgrade_command cli =
           OpamArg.dir_sep)
   ]
   in
-  let clear_cache_arg =
```
rjbou, line 783: Is there a reason to not use the removal framework with `OpamArg.cli_between` ?

rjbou (approved):

PR #6979, opened 2026-06-27
GHA: Fix the basic tests when using the Cygwin backend

kit-ty-kate:
See https://github.com/ocaml/dune/issues/15340


kit-ty-kate:
Every CI run successfully completed the basic test section so ignoring the rest of the runs. Merging to avoid cyclic PR dependency with https://github.com/ocaml/opam/pull/6978

PR #6980, opened 2026-06-27
Add support for using git repositories owned by another local user

### LINKED ISSUE #6963: opam fails on `git ls-files`
I have a CI script failing in an environment where `root` owns the git clone of the project being tested, but the CI script itself runs with a different user (which also has rights to access and create files in the git clone). The problem comes from the fact that `opam` calls `git ls-files` internally, but recent-ish versions of `git` fail when the owner of the repository is not the same as the current user.

Expected behavior: I wish that `opam` would not fail with an error in this case, at least when `--yes` is being passed. Instead it should use the same logic than when it is _not_ in a git directory and thus cannot call `git ls-files`.

Reproduction script:

```shell
# create a reproduction subdirectory
mkdir repro

# populate it with a foo.opam file
echo 'opam-version: "2.0"' >> repro/foo.opam
echo 'depends: [ "ocaml" ]' >> repro/foo.opam
echo 'build: []' >> repro/foo.opam

# initialize it as a git repo
(cd repro; git init)

# give wide access rights, but change ownership to root
chmod ugo+rwx -R repro
sudo chown root -R repro

# now run an opam command from that repository
cd repro
opam install --deps-only . --yes
```

Observed behavior: opam fails with

```
[ERROR] Command "/usr/bin/git ls-files" failed:
"/usr/bin/git ls-files" exited with code 128
```

and returns error code 99.

Expected behavior: the command succeeds.

kit-ty-kate:
Fixes #6963

### DIFF
--- a/.github/scripts/main/main.sh
+++ b/.github/scripts/main/main.sh
@@ -173,6 +173,22 @@ if [ "$OPAM_TEST" = "1" ]; then
   opam install opam-rt --deps-only opam-devel
   opam exec -- make || { opam reinstall opam-client -y; opam exec -- make; }
   (set +x ; echo -en "::endgroup::opam-rt\r") 2>/dev/null
+
+  # Test https://github.com/ocaml/opam/issues/6963 (not possible in reftests because sudo/root access is required)
+  testdir=/tmp/opam-test-6963
+  git init "$testdir"
+  cat > "$testdir/opam" << EOF
+opam-version: "2.0"
+name: "opam-test-6963"
+build: "true"
+EOF
+  chmod ugo+rwx -R "$testdir"
+  sudo chown root -R "$testdir"
+  if git -C "$testdir" ls-files; then
+    echo "Your git version is too old for this test"
+    exit 1
+  fi
+  opam install -y "$testdir"
 fi
 
 test_project () {
--- a/master_changes.md
+++ b/master_changes.md
@@ -245,6 +245,7 @@ users)
   * Upgrade the CI to OCaml 5.5 and 4.14.4 [#6988 @kit-ty-kate]
   * Fix the Cygwin 5.5 build by requiring cygwin gcc 13.4 as 14.4 breaks C++ support [#7007 @kit-ty-kate]
   * Upgrade to use opam 2.5.2 [#7018 @kit-ty-kate]
+  * Add a test showing opam using a git repository owned by root [#6980 @kit-ty-kate]
 
 ## Doc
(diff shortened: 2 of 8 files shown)


kit-ty-kate:
Moved to the 2.7 milestone because GHA's git version is too old for the test to work

src/repository/opamGit.ml
```diff
@@ -30,7 +30,8 @@ module VCS : OpamVCS.VCS = struct
        of git will need to change, as altering PATH could select a different
        Git *)
     fun ?verbose ?stdout args ->
-      OpamSystem.make_command ?verbose ?stdout "git" ("-C"::dir::args)
+      OpamSystem.make_command ?verbose ?stdout
```
rjbou, line 33: Is it possible to use the git config to add that argument ? cf #6992

.github/scripts/main/main.sh
```diff
@@ -173,6 +173,22 @@ if [ "$OPAM_TEST" = "1" ]; then
   opam install opam-rt --deps-only opam-devel
   opam exec -- make || { opam reinstall opam-client -y; opam exec -- make; }
   (set +x ; echo -en "::endgroup::opam-rt\r") 2>/dev/null
+
+  # Test https://github.com/ocaml/opam/issues/6963 (not possible in reftests because sudo/root access is required)
```
rjbou, line 177: It worth having a highlight what is tested instead of just a link.

master_changes.md
```diff
@@ -102,6 +102,7 @@ users)
 
 ## VCS
   * Darcs no longer fall back to “Num Patches” as “Weak Hash” has been supported since 2016 [#6866 @toastal]
+  * Add support for using git repositories owned by another local user [#6980 @kit-ty-kate]
```
rjbou, line 105: \+ an entry in API changelog (or another new Git section?) for the information of the new config option added in git calls

rjbou (commented):
On the idea, lgtm.  It worth to be more explicit on the main comment of the PR and the message of last commit (how is this done)
The Linux & macos tests are failing
```
+ testdir=/tmp/opam-test-6963
+ git init /tmp/opam-test-6963

+ cat
Initialized empty Git repository in /private/tmp/opam-test-6963/.git/
+ chmod ugo+rwx -R /tmp/opam-test-6963
+ sudo chown root -R /tmp/opam-test-6963
+ git -C /tmp/opam-test-6963 ls-files
+ echo 'Your git version is too old for this test'
Your git version is too old for this test
+ exit 1

PR #6981, opened 2026-06-27
Fix the filename check used when parsing arguments to allow '/' to be recognised as a directory separator on Windows

### LINKED ISSUE #6940: Failure to properly handle directory separators on windows platforms
The following behavior can be observed on Windows :

```shellsession
$ opam install dir/*.opam
opam: PACKAGES… arguments: Invalid character '/' in package name "dir/package"
Usage: opam install [OPTION]… [PACKAGES]…
```

This is surprising since this platform supposedly has handled both `/` and `\` as directory separators for a while.

Versions used:
- opam.2.5.1 / mingw64
- ocaml.4.14.2
- (please ask for more if needed, I am not versed on what Windows-specifics may influence this behavior)

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/6940

The only reason this half worked before is because of the `String.starts_with ~prefix:"." s` which should instead have been `s = "." || s = ".."`, else `.pkg` is wrongly detected as a valid filename instead of an invalid package name.

Also for some reason the testsuite always use `./` as a prefix when using directory names etc, so it was never detected.

### DIFF
--- a/src/client/opamArg.ml
+++ b/src/client/opamArg.ml
@@ -872,8 +872,8 @@ let atom =
 
 let atom_or_local =
   let parse str =
-    if OpamStd.String.contains ~sub:Filename.dir_sep str ||
-       OpamCompat.String.starts_with ~prefix:"." str
+    if OpamFilename.is_rel_seg str ||
+       OpamCompat.String.exists OpamFilename.is_dir_sep str
     then
       if OpamFilename.(exists (of_string str)) then
         Ok (`Filename (OpamFilename.of_string str))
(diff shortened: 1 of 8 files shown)


src/client/opamArg.ml
```diff
@@ -872,8 +872,8 @@ let atom =
 
 let atom_or_local =
   let parse str =
-    if OpamStd.String.contains ~sub:Filename.dir_sep str ||
-       OpamCompat.String.starts_with ~prefix:"." str
+    if OpamFilename.is_rel_seg str ||
```
rjbou, line 875: How does this affect (or not) the other commands that rely on atoms_or_local ?

kit-ty-kate: only tree/why, show, install and lock uses it and it affects them the same exact way

rjbou (approved):

rjbou:
thanks!

PR #6992, opened 2026-06-30
Make git calls more deterministic regardless of the global or system config

### LINKED ISSUE #6937: opam update fails between 4ae3c774e4 and latest master: OpamSystem.Internal_patch_error("Patch \"in directory         /home/edwint/.opam/repo/test\" does not apply cleanly.")
Steps to reproduce with opam 2.5.1 (narrowed down to a smaller commit range):

This works:
```
git clone https://github.com/ocaml/opam-repository && cd opam-repository
git checkout 142525538e5694742be38a5edaf7d3d6051acaf8
opam repo remove testbug --all
opam repo add testbug .
git checkout d9cabb4c55
opam update testbug
```

This fails:
```
git clone https://github.com/ocaml/opam-repository && cd opam-repository
git checkout '142525538e5694742be38a5edaf7d3d6051acaf8^'
opam repo remove testbug --all
opam repo add testbug .
git checkout d9cabb4c55
opam update testbug
```

Interestingly running `opam update testbug` again works then (but this doesn't help with the original issue I had).

Although updating directly from 142525538e5694742be38a5edaf7d3d6051acaf8^ to 142525538e5694742be38a5edaf7d3d6051acaf8 works too.

FWIW updating the default https remote works, it is just using a git remote that fails.

```
$ opam --version
2.5.1
$ git --version
git version 2.54.0
$ patch --version
GNU patch 2.8
```

Although it looks like opam processes the patch using an OCaml implementation of patch, so the bug might be there.

kit-ty-kate:
Fixes https://github.com/ocaml/opam/issues/6937

* Set the protocol.file.allow=always git config while running the reftests regardless of GIT_CONFIG_GLOBAL
* When fetching a git repository, the resulting git branch is now deterministically named main instead of taking the system's init.defaultBranch
* Make git calls more deterministic regardless of the global or system config
git calls made in opam had a non-deterministic behaviour based on the
user's local git configuration. Setting GIT_CONFIG_GLOBAL and GIT_CONFIG_SYSTEM
to /dev/null fixes this problem.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -210,6 +210,7 @@ users)
   * Add 2.6 root test cases in opamroot-versions [#6625 @rjbou]
   * Add tests for `.install` fields handling [#6956 #67026 @rjbou]
   * Add a test showing opam pin list not working when the source git directory is missing [#6597 @kit-ty-kate]
+  * Add a test making sure the global or system `git` config doesn't change the behaviour of opam [#6992 @kit-ty-kate]
 
 ### Engine
   * Add `http-server` to launch a minimal http server [#6939 @rjbou]
--- a/master_changes.md
+++ b/master_changes.md
@@ -30,6 +30,7 @@ users)
   * Remove the build directory as soon as possible when installing a package [#6906 @kit-ty-kate - fix #5884]
 
 ## Build (package)
+  * When fetching a git repository, the resulting git branch is now deterministically named `main` instead of taking the system's `init.defaultBranch` [#6992 @kit-ty-kate]
 
 ## Remove
 
@@ -67,6 +68,7 @@ users)
 ## Exec
 
 ## Source
+  * When fetching a git repository (i.e. `--dev-repo` or a package with git url), the resulting git branch is now deterministically named `main` instead of taking the system's `init.defaultBranch` [#6992 @kit-ty-kate]
 
 ## Lint
 
@@ -74,6 +76,7 @@ users)
   * No longer call tar tool to create archives, use tar library instead [#6945 @kit-ty-kate]
   * [BUG] Do not fail on directories named `opam` when scanning the `packages` directory of a repository during `opam repo add` or `opam init` (worked on subsequent `opam update`) [#6941 @kit-ty-kate @rjbou]
   * Speedup repository operations on certain file-systems (e.g. NTFS on Windows or IO constrained machines) by changing its storage in the opam root from plain directory to archive, for HTTP repositories, or non-VCS one if `OPAMREPOSITORYTARRING` is enabled [#6625 @rjbou @kit-ty-kate @arozovyk - fix #5346 #5741 #5648 #5484 #5559 #3050 #6974]
+  * When fetching a git repository, the resulting git branch is now deterministically named `main` instead of taking the system's `init.defaultBranch` [#6992 @kit-ty-kate]
 
 ## Lock
   * [BUG] Fix `undefined variable` error when a lock file filter contains an undefined variables: fail gracefully with strict mode, continue and default the variable to false on normal mode [#6947 @rjbou - fix #6946]
--- a/master_changes.md
+++ b/master_changes.md
@@ -105,6 +105,7 @@ users)
 
 ## VCS
   * Darcs no longer fall back to “Num Patches” as “Weak Hash” has been supported since 2016 [#6866 @toastal]
+  * Make `git` calls more deterministic regardless of the global or system config [#6992 @kit-ty-kate - fix #6937]
 
 ## Build
   * Fix Windows build on MSYS2 [#6862 @Firobe]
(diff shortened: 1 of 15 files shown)


edwintorok:
Looks like this creates some problems for the tests which now behave differently as defaults between git versions change.

> --- a/_build/default/tests/reftests/action-disk.test
>+++ b/_build/default/tests/reftests/action-disk.out
>@@ -1278,6 +1278,19 @@ SYSTEM                          mkdir ${BASEDIR}/OPAM/install-from-git-pin-all/.
> SYSTEM                          mkdir ${BASEDIR}/OPAM/install-from-git-pin-all/.opam-switch/sources/main-gpin
> Processing  1/1: [main-gpin: git]
> + git "-C" "${BASEDIR}/OPAM/install-from-git-pin-all/.opam-switch/sources/main-gpin" "init"
>+- hint: Using 'master' as the name for the initial branch. This default branch name
>+- hint: will change to "main" in Git 3.0. To configure the initial branch name
>+- hint: to use in all of your new repositories, which will suppress this warning,
>+- hint: call:
>+- hint:
>+- hint: 	git config --global init.defaultBranch <name>

Probably need to explicitly set `init.defaultBranch` too to avoid test failures.

tests/reftests/run.ml
```diff
@@ -86,6 +86,10 @@ let base_env =
     "OPAMNOENVNOTICE", "1";
     "OPAMNODEPEXTS", "1";
     "OPAMDOWNLOADJOBS", "1";
+
+    "GIT_CONFIG_COUNT", "1";
+    "GIT_CONFIG_KEY_0", "protocol.file.allow";
+    "GIT_CONFIG_VALUE_0", "always";
```
rjbou, line 92: There is a comment in git.test that need to be removed as we allow protocol file, no ?

master_changes.md
```diff
@@ -67,13 +68,15 @@ users)
 ## Exec
 
 ## Source
+  * When fetching a git repository (e.g. `--dev-repo`), the resulting git branch is now deterministically named `main` instead of taking the system's `init.defaultBranch` [#6992 @kit-ty-kate]
```
rjbou: opened #7035

rjbou: It is possible to highlight this change by adding some verbose to  source.test git commands, and something like 
```
### git -C pandore6 branch -avv
* main             42e6a9c second empty commit
  remotes/opam-ref 42e6a9c second empty commit
### git -C pandore6 remote -v
origin	file://${BASEDIR}/pandev (fetch)
origin	file://${BASEDIR}/pandev (push)
```

tests/reftests/git.test
```diff
@@ -142,3 +142,115 @@ Done.
...
+www
+xxx
+yyy
+zzz
+"""
+### git -C GIT_REPO_DUP init -q --initial-branch=master
+### git -C GIT_REPO_DUP config core.autocrlf false
+### git -C GIT_REPO_DUP add *
+### git -C GIT_REPO_DUP commit -qm init
+### opam repo add dup ./GIT_REPO_DUP
```
rjbou:
```suggestion
### opam repo add dup ./GIT_REPO_DUP --this-switch
```

tests/reftests/git.test
```diff
@@ -142,3 +142,115 @@ Done.
...
+rrr
+sss
+ttt
+uuu
+vvv
+www
+xxx
+yyy
+zzz
+"""
```
rjbou, line 180: What is the reason to have this long post message ?

master_changes.md
rjbou: It doesn't affect `opam source` of  package from the repository with a git url ?

rjbou (commented):
On the idea, lgtm. It worth to be more explicit on the main comment of the PR and the message of last commit.

master_changes.md
```diff
@@ -102,6 +105,7 @@ users)
 
 ## VCS
   * Darcs no longer fall back to “Num Patches” as “Weak Hash” has been supported since 2016 [#6866 @toastal]
+  * Make `git` calls more deterministic regardless of the global or system config [#6992 @kit-ty-kate - fix #6937]
```
rjbou, line 108: \+ an entry in API changelog (or another new Git section?) for the information of the new config option added for git calls

master_changes.md
kit-ty-kate: it does. Why wouldn't it? That's why i'm saying `e.g.` not `i.e.` and `fetching a git repository` is generic enough

tests/reftests/git.test
kit-ty-kate: For safety, to make sure git doesn't suddenly decide that having `copy from` in the diff isn't worth it.

rjbou: For future git changes then. It worth a comment, to be sure it is kept as is, understand the intent (when looking at it later, or if this test inspires another one).

master_changes.md
rjbou: It worth adding it then
```suggestion
  * When fetching a git repository (e.g. `--dev-repo` or package with git url), the resulting git branch is now deterministically named `main` instead of taking the system's `init.defaultBranch` [#6992 @kit-ty-kate]
```

kit-ty-kate:
done and rebased

rjbou (approved):
thanks!

PR #7001, opened 2026-07-03
`OpamRepositoryRoot` tgz folding with memoisation

rjbou:
Extracted from #6625


kit-ty-kate (changes requested):
memoisation would blow up ram usage. I don't think it would be wise to have that

PR #7002, opened 2026-07-03
GHA: fix depends jobs when a opam root format upgrade is needed

rjbou:
Needed for #6625


kit-ty-kate (approved):

kit-ty-kate:
Ignoring the random windows failure (https://github.com/ocaml/opam/issues/6890) and ocaml-benchmarks (build machine unavailable)

Issue #6986, opened 2026-06-29
Rewrite `OpamFilename`


rjbou:
In last months, several PRs had paths handling and `OpamFilename` was painful to use. There is no strong invariant, some functions manipulate raw strings, other resolve paths, others check on disk, the functions are not homogeneous (`OpamFilename.Dir.fun` and `OpamFilename.dir_fun'` coexist), its `basename` module is used on places where it shouldn't (#6967, because there was the simpliest way to do it at the time).
We should rethink that abstraction and rewrite it completely.

PR #7004, opened 2026-07-06
Make OpamFilename.remove_prefix consistent with starts_with

kit-ty-kate:
Extracted from #6953 by @NathanReb
In my opinion the change is too impacting (e.g. #6625) and not as easy to check the correctness of. Also tests were missing and the original issue isn't present on `remove_prefix`.


kit-ty-kate:
In my opinion, i think we should rather focus on https://github.com/ocaml/opam/issues/6986 and find good invariants in the first place than finish this

PR #7005, opened 2026-07-07
Fix bypass switch prefix check on install using symlinks

kit-ty-kate:
CVE-2026-57825
Fix by @NathanReb 
Backported to 2.5 in https://github.com/ocaml/opam/pull/7006


rjbou (approved):
It would worth to add some context in the commits message (not only the number)

rjbou (approved):

kit-ty-kate:
Ignoring the Cygwin failure (see https://github.com/ocaml/opam/pull/7007) and the msvc windows failure (https://github.com/ocaml/opam/issues/6890)

PR #6879, opened 2026-04-08
Remove support for building opam with OCaml 4.08, 4.09 and 4.10

kit-ty-kate:
To review only if/once [my proposal about bumping the oldest supported compiler in opam-repository to 4.11](https://discuss.ocaml.org/t/dropping-some-intermediate-ocaml-versions-from-ci/17947/9) ends up happening

PR #7008, opened 2026-07-07
Re-allow .install files containing '..'

kit-ty-kate:
Partially reverts https://github.com/ocaml/opam/commit/ae877c0f74f696e05c2d524913c682c8aeaad918 (from #6879)

This check is no longer necessary since https://github.com/ocaml/opam/commit/d7f7d8e39afcf8ade6b560224e5b084a924f2b15 (#7005) and broke opam repositories providing cross-compiled packages as such format is used by `dune` (see https://github.com/ocaml/dune/issues/14393)

See https://ocaml.zulipchat.com/#narrow/channel/527832-opam/topic/Cross.20compilation.20with.20opam.202.2E5.2E1/with/592359165

Backported to 2.5 in #7009

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -97,7 +97,7 @@ users)
 
 ## Opamfile
   * The `url` file now only supports the legacy opam 1.2 fields [#6827 @kit-ty-kate]
-  * Filter fields in .install files containing destinations with `..` or absolute filepaths as parse errors [#6897 @kit-ty-kate]
+  * Warn on fields in .install files containing destinations with `..` and filter absolute filepaths as parse errors [#6897 #7008 @kit-ty-kate]
 
 ## External dependencies
   * Restore the distribution detection on Gentoo [#6886 @kit-ty-kate - fix #6887]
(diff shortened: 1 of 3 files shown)


rjbou (commented):
Also, I think it is better to avoid having links, especially non permanent ones on commit messages. If the message is deleted, or the zulip is no longer online, it is not possible to find out the origin that you wanted to highlight.

rjbou (commented):
Even if it wasn't enforced in the code, from the manual, `..` are forbidden in .install target.
My opinion: I'm in favor or removing it (keep as is), especially as there is now a field that permit to no longer need them (#6938). For backward compatibility, we can reintroduce it, but discourage its use with a warning (that errors on strict mode). We can even think about removing the support in a few releases (we need then to give the information from now on release blog posts).

kit-ty-kate:
Per https://github.com/ocaml/opam/pull/7009#issuecomment-4935708302, that backward compatibility is required for 2.6 at least (then we can bring back the stricter rule in 2.7)

rjbou (approved):
I've update the code : 
* rebased
* rebase on top of 7015
* added a warning (error on strict mode) for `..` in target
* added a test with strict mode to highlight the error

kit-ty-kate:
Ignoring the missing macOS runs. Seems like GitHub have nuked their macOS machines out of existence

PR #7007, opened 2026-07-07
GHA: Fix Cygwin build

kit-ty-kate:
~Could this be enough to f.i.x https://github.com/ocaml/ocaml/issues/14911?~

The original gcc package constraints were from https://github.com/ocaml/opam/pull/6624

PR #6956, opened 2026-06-05
reftest: add '.install' exhaustive test for all fields

rjbou:
#6938 add new field to `.install` files, there was no exhaustive test.
The tests are 3 files:
- a cross platform test for non executable fields
- 2 per platform tests for executable fields (it differs from win to nux)
Note: the windows test highlights some bugs, issues will be opened

PR #7009, opened 2026-07-07
[2.5] Re-allow .install files containing '..'

kit-ty-kate:
Backport of https://github.com/ocaml/opam/pull/7008 on 2.5
Queued on #7013

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -117,6 +117,7 @@ users)
 ## Doc
 
 ## Security fixes
+  * Invalidate .install fields containing absolute destination filepath except when in the `misc` field [#6897 #7008 @kit-ty-kate]
   * Fix a bug that allowed a package to install files anywhere on the system using a symlink to an external directory without warning the user and asking for their permission: CVE-2026-57825. [#7005 @NathanReb]
 
 # API updates
--- a/master_changes.md
+++ b/master_changes.md
@@ -107,6 +107,7 @@ users)
 
 ## Reftests
 ### Tests
+  * Add a test showing the behaviour of .install files containing destination filepath trying to escape their scope [#6897 #7008 @rjbou @kit-ty-kate]
   * Add a test ensuring installing files through a .install file can't escape the opam switch (CVE-2026-57825) [#7005 @NathanReb]
 
 ### Engine
--- a/master_changes.md
+++ b/master_changes.md
@@ -21,6 +21,7 @@ users)
 ## Config report
 
 ## Actions
+  * Stop the installation process via .install file from creating unnecessary intermediate directories for paths that contain `..` [#7015 @kit-ty-kate]
 
 ## Install
 
--- a/tests/reftests/dot-install.test
+++ b/tests/reftests/dot-install.test
@@ -370,30 +370,35 @@ lib: [
   "a-file"  { "../../a-file" }
   "fichier" { "/ab/so/lute/path/fichier" }
   "dosiero" { "/tmp/../middle/dosiero" }
+  "dosiero" { "tmp/../../../../middle/dosiero" }
   "good"    { "good" }
   ]
 bin: [
   "a-file"  { "../../a-file" }
   "fichier" { "/ab/so/lute/path/fichier" }
   "dosiero" { "/tmp/../middle/dosiero" }
+  "dosiero" { "tmp/../../../../middle/dosiero" }
   "good"    { "good" }
   ]
 etc: [
   "a-file"  { "../../a-file" }
   "fichier" { "/ab/so/lute/path/fichier" }
   "dosiero" { "/tmp/../middle/dosiero" }
+  "dosiero" { "tmp/../../../../middle/dosiero" }
   "good"    { "good" }
   ]
 share: [
   "a-file"  { "../../a-file" }
   "fichier" { "/ab/so/lute/path/fichier" }
   "dosiero" { "/tmp/../middle/dosiero" }
+  "dosiero" { "tmp/../../../../middle/dosiero" }
   "good"    { "good" }
   ]
 misc: [
   "a-file"  { "../../a-file" }
   "fichier" { "/ab/so/lute/path/fichier" }
   "dosiero" { "/tmp/../middle/dosiero" }
+  "dosiero" { "tmp/../../../../middle/dosiero" }
   "good"    { "good" }
   ]
 ### <pkg:eskape.1:a-file>
@@ -410,7 +415,7 @@ hallo
 [default] no changes from file://${BASEDIR}/REPO
 ### opam switch create escapability --empty
 ### OPAMDEBUGSECTIONS="SYSTEM" OPAMDEBUG=-5
-### opam install eskape
+### opam install eskape | unordered
 FILE(config)                    Read ${BASEDIR}/OPAM/config in 0.000s
 SYSTEM                          LOCK ${BASEDIR}/OPAM/lock (none => read)
 SYSTEM                          LOCK ${BASEDIR}/OPAM/repo/state-magicv.cache (none => read)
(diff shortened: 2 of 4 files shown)


tests/reftests/dot-install.test
```diff
@@ -437,17 +437,218 @@ SYSTEM                          write ${BASEDIR}/OPAM/escapability/.opam-switch/
...
+SYSTEM                          mkdir ${BASEDIR}/OPAM/escapability/share/eskape/tmp
+SYSTEM                          mkdir ${BASEDIR}/OPAM/escapability/share/eskape/tmp/..
+SYSTEM                          mkdir ${BASEDIR}/OPAM/escapability/share/eskape/tmp/../middle
+SYSTEM                          install ${BASEDIR}/OPAM/escapability/.opam-switch/build/eskape.1/dosiero -> ${BASEDIR}/OPAM/escapability/share/eskape/tmp/../middle/dosiero (644)
+SYSTEM                          mkdir ${BASEDIR}/OPAM/escapability/etc
+SYSTEM                          mkdir ${BASEDIR}/OPAM/escapability/etc/eskape
+SYSTEM                          mkdir ${BASEDIR}/OPAM/escapability/etc/eskape/tmp
+SYSTEM                          mkdir ${BASEDIR}/OPAM/escapability/etc/eskape/tmp/..
+SYSTEM                          mkdir ${BASEDIR}/OPAM/escapability/etc/eskape/tmp/../middle
+SYSTEM                          install ${BASEDIR}/OPAM/escapability/.opam-switch/build/eskape.1/dosiero -> ${BASEDIR}/OPAM/escapability/etc/eskape/tmp/../middle/dosiero (644)
```
rjbou: pathes are note resolved (see #7010), so it can be scaped

tests/reftests/dot-install.test
```diff
@@ -437,17 +437,218 @@ SYSTEM                          write ${BASEDIR}/OPAM/escapability/.opam-switch/
...
+    ${OPAM} install --restore
+SYSTEM                          LOCK ${BASEDIR}/OPAM/repo/lock (none => none)
+SYSTEM                          LOCK ${BASEDIR}/OPAM/config.lock (none => none)
+# Return code 31 #
+### ocaml cat.ml escapability eskape
+==> eskape installed file
+Not found: ${BASEDIR}/OPAM/escapability/share/eskape/a-file
+==> eskape changes
+Not found: ${BASEDIR}/OPAM/escapability/.opam-switch/install/eskape.changes
+### <pkg:eskape.1:eskape.install>
```
rjbou: Is it possible to add a comment to highlight what are you testing

tests/reftests/dot-install.test
```diff
@@ -437,17 +437,218 @@ SYSTEM                          write ${BASEDIR}/OPAM/escapability/.opam-switch/
...
+etc: [
+  "a-file"  { "../../a-file" }
+  ]
+share: [
+  "a-file"  { "../../a-file" }
+  ]
+misc: [
+  "a-file"  { "../../a-file" }
+  ]
+### opam reinstall eskape | unordered
```
rjbou: It is better to use a new packages or new switch to avoid having all the recompile log

rjbou (changes requested):
There is a missing test that checks escapability (for ex `{ "tmp/../../../../middle/dosiero" }`. That's how a found #7010.
Besides this real_path behaviour, I understand the reason to reintroduce `..` handling in 2.5.2, but i don't think we should backport it on master, at least for the moment. #6938 is open to propose an alternative solution, but we can discuss it later (but definitively before 2.6 release)

tests/reftests/dot-install.test
```diff
@@ -474,3 +479,286 @@ SYSTEM                          LOCK ${BASEDIR}/OPAM/config.lock (none => none)
...
+lib: [
+  "a-file"  { "../../a-file" }
+]
+etc: [
+  "a-file"  { "../../a-file" }
+]
+share: [
+  "a-file"  { "../../a-file" }
+]
+misc: [
```
rjbou, line 497: I don't know if it worth testing misc for this test, it needs to be an absolute path and we are testing relative paths

tests/reftests/dot-install.test
```diff
@@ -474,3 +479,286 @@ SYSTEM                          LOCK ${BASEDIR}/OPAM/config.lock (none => none)
...
+bin: [
+  "a-file" { "tmp/../middle/a-file" }
+]
+etc: [
+  "a-file" { "tmp/../middle/a-file" }
+]
+share: [
+  "a-file" { "tmp/../middle/a-file" }
+]
+misc: [
```
rjbou, line 578: same

tests/reftests/dot-install.test
```diff
@@ -474,3 +479,286 @@ SYSTEM                          LOCK ${BASEDIR}/OPAM/config.lock (none => none)
...
+-> installed eskape-rel-dotdot-exists.1
+SYSTEM                          mkdir ${BASEDIR}/OPAM/escapability/.opam-switch/packages/eskape-rel-dotdot-exists.1
+SYSTEM                          mkdir ${BASEDIR}/OPAM/escapability/.opam-switch/packages/eskape-rel-dotdot-exists.1/files
+SYSTEM                          read ${BASEDIR}/OPAM/repo/default/packages/eskape-rel-dotdot-exists/eskape-rel-dotdot-exists.1/files/eskape-rel-dotdot-exists.install
+SYSTEM                          write ${BASEDIR}/OPAM/escapability/.opam-switch/packages/eskape-rel-dotdot-exists.1/files/eskape-rel-dotdot-exists.install
+SYSTEM                          read ${BASEDIR}/OPAM/repo/default/packages/eskape-rel-dotdot-exists/eskape-rel-dotdot-exists.1/files/a-file
+SYSTEM                          write ${BASEDIR}/OPAM/escapability/.opam-switch/packages/eskape-rel-dotdot-exists.1/files/a-file
+
+<><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
+              ../../a-file references its parent directory.
```
rjbou: For readability, it is better to promote without unordered, then add unordered for later testing

tests/reftests/dot-install.test
```diff
@@ -474,3 +479,286 @@ SYSTEM                          LOCK ${BASEDIR}/OPAM/config.lock (none => none)
...
+bin: [
+  "a-file" { "tmp/../../../middle/a-file" }
+]
+etc: [
+  "a-file" { "tmp/../../../../middle/a-file" }
+]
+share: [
+  "a-file" { "tmp/../../../../middle/a-file" }
+]
+misc: [
```
rjbou, line 670: same

master_changes.md
```diff
@@ -115,6 +116,7 @@ users)
 ## Doc
 
 ## Security fixes
+  * Invalidate .install fields containing absolute destination filepath except when in the `misc` field [#6897 #7008 @kit-ty-kate]
```
rjbou, line 121:
```suggestion
  * Re-allow `..` in `.install` files, partially reverts 2.5.1 #6879 [#7009 @kit-ty-kate]
```

kit-ty-kate: it doesn't really matter, it's master_changes which is just for us anyway and we're doing a release right after this.

tests/reftests/dot-install.test
kit-ty-kate: since one wrong thing in a field invalidates the whole field i think it's worth checking since this pattern isn't checked

rjbou: Fair enough. It will be in #6956

rjbou (approved):
lgtm with green CI
On the 2nd commit message, I think it is better to avoid having links, especially non permanent ones on commit messages. If the message is deleted, or the zulip is no longer online, it is not possible to find out the origin that you wanted to highlight.

kit-ty-kate:
Ignoring the unfinished Solvers-macOS jobs whose result would be unrelated to this PR anyway, and the failing cygwin build (#7007)

WardBrian:
> I understand the reason to reintroduce `..` handling in 2.5.2, but i don't think we should backport it on master, at least for the moment. #6938 is open to propose an alternative solution, but we can discuss it later (but definitively before 2.6 release)

Not sure if this is the place to discuss it, but the easiest path forward is to include this behavior _and_ the new root/rootexec in a release, since[ dune wants the new fields released before targetting them](https://github.com/ocaml/dune/pull/14600#issuecomment-4766608828). The release following that could restrict `..` again if desired.

rjbou:
>  this behavior and the new root/rootexec in a release

That's what i'm prone to, with some logging in case of '..'. I think we'll discuss it on monday dev meeting to settle it.

PR #7015, opened 2026-07-08
Stop the installation process via .install file from creating unnecessary intermediate directories for paths that contain '..'

kit-ty-kate:
Detected in https://github.com/ocaml/opam/pull/7008

Target paths are now resolve before their creation & install.

Backported to 2.5 in https://github.com/ocaml/opam/pull/7009
It is queued on #7008 (reintroduce '..') to be able to see the changes in tests, but #7008 can't be merged without this PR. So #7008 contains the update of the test added here

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -32,6 +32,7 @@ users)
 ## Actions
   * Reorder the list of actions by increased priority [#6864 @kit-ty-kate - fix #6863]
   * The internal `sources` directory now only serves to cache dev packages instead of every packages, saving disk space. For other packages it is now removed during the build phase. [#6440 @kit-ty-kate - fix #4056 #5448]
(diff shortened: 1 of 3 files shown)


rjbou (commented):
The added test (from #7008) doesn't highlight the changes. I've updated the main comment to add the queued information

rjbou (commented):
Same, the main comment and main commit (last one) doesn't contains the explanation of what was done to no longer create intermediate dirs during the installation process from .install.

rjbou (approved):
Updated commit msg & main comment

kit-ty-kate:
Ignoring the missing macOS runs. Seems like GitHub have nuked their macOS machines out of existence

kit-ty-kate:
The effect of the change can be seen in #7008

PR #7016, opened 2026-07-08
Release 2.5.2

kit-ty-kate:
Modulo https://github.com/ocaml/opam/pull/7009 and subsequent empting of master_changes.md


CHANGES
```diff
@@ -3,6 +3,27 @@ repositories (changes that are automatically handled by the format upgrade tools
...
 
+2.5.2:
+* Fix a bug that allowed a package to install files anywhere on the system
+  using a symlink to an external directory without warning the user and
+  asking for their permission: CVE-2026-57825 / OSEC-2026-10.
+  [#7005 @NathanReb]
+* Re-allow `..` in `.install` files, partially reverts 2.5.1's #6879
+  [#7008 @kit-ty-kate - fix ocaml/dune#14393]
+* Stop the installation process via .install file from creating unnecessary
+  intermediate directories for paths that contain `..` [#7015 @kit-ty-kate]
```
rjbou, line 14:
```suggestion
* When installing files from `.install` resolve them before creating paths.
  It no longer create unnecessary intermediate directories if it contains
  indirections (`..`) [#7015 @kit-ty-kate]
```

CHANGES
```diff
@@ -3,6 +3,27 @@ repositories (changes that are automatically handled by the format upgrade tools
...
+  [#6953 #7011 #7008 #7005 @NathanReb @rjbou @kit-ty-kate]
+* API changes in `opam-core`:
+  * `OpamFilename.{,dir_}starts_with`: Fix a bug where `foo/bar` would be
+    considered a prefix of `foo/bar-baz` [#6953 @NathanReb - fix #6948]
+  * `OpamFilename.{,dir_}starts_with`: `/` and `\` are now equivalent on
+    Windows [#6953 @NathanReb]
+  * `OpamFilename.starts_with`: `starts_with "a/b" "a/b"` no longer returns
+    `true` [#6953 @NathanReb]
+  * `OpamSystem.real_path`: fix a bug where paths after a non existent
+    directory where not resolve [#7011 @kit-ty-kate - fix #7010]
```
rjbou, line 25:
```suggestion
    directory where not resolve [#7011 @kit-ty-kate - fix #7010]
  * `OpamAction.install_package`: when installing files from `.install`, resolve paths before installing them [#7015 @kit-ty-kate]
```

rjbou (approved):

CHANGES
kit-ty-kate: i don't think it's an API change worth mentioning

CHANGES
kit-ty-kate: I'm not sure it's clearer

rjbou: maybe a rewording is needed to make it clearer. The idea was to highlight that the paths are now resolved, which avoid creating unnecessary dirs. It is not obvious with just "no more create unnecessary dirs"

kit-ty-kate: From the users' point of view i don't think it makes sense to mention that paths are resolved. They don't really care. The only effect they could see is the unnecessary directories not being created anymore

rjbou: If we go that way, from user perspective, there is no impact on the usage, its all under opam root, expect for misc.
I've thought about it because if I read an external changelog that contains that line, it would have implied a lot of questions (why it was created in the first place, what changed, is it a bug, etc.), which can be answered directly/not raised by having the "path resolved" information.

kit-ty-kate: > If we go that way, from user perspective, there is no impact on the usage, its all under opam root, expect for misc.

I'm also happy to remove it from the changelog entirely.

> I've thought about it because if I read an external changelog that contains that line, it would have implied a lot of questions (why it was created in the first place, what changed, is it a bug, etc.), which can be answered directly/not raised by having the "path resolved" information.

But if someone were to have question of that type, then they'd already be interested in the opam codebase enough to simply have a look at the PR directly. Then the question would answer itself.

rjbou: > But if someone were to have question of that type, then they'd already be interested in the opam codebase enough to simply have a look at the PR directly. Then the question would answer itself.

Not necessarily, when i read release notes, i'm just interested on the changes, not the codebase itself. Plus, there is a cost to go and dig on an unknown codebase to have these questions answered, that i won't do on most cases.


> I'm also happy to remove it from the changelog entirely.

My opinion : we should keep the line (otherwise half of opam changelog would be removed) and i think it is better to have the "path resolved" information. Now, it is ok for me to keep it as is, the PR is already approved.

kit-ty-kate:
Ignoring cygwin failure (#7007) and alpine failure (https://github.com/ocaml/opam/pull/6888)

PR #7017, opened 2026-07-09
Add opam 2.5.2 to the install scripts

kit-ty-kate:


kit-ty-kate:
Hygiene, Changelog and script install checks are the only relevant CI checks here and are all green, so merging without waiting.

PR #7026, opened 2026-07-13
reftests: run dot install non exe fields test without strict mode

rjbou:
The test was run with the default `OPAMSTRICT=1`, which errors on some warnings.


kit-ty-kate (approved):

kit-ty-kate:
Ignoring cygwin failure (#6890)

PR #7033, opened 2026-07-14
GHA: Ensure a better availability of the micro_httpd dependency

kit-ty-kate:
Going through opam-repository instead of the git repository is faster and more consistent

### DIFF
--- a/.github/scripts/main/reftests.sh
+++ b/.github/scripts/main/reftests.sh
@@ -36,7 +36,14 @@ esac
 
 export PATH="$PREFIX/bin:$PATH"
 
-git clone https://codeberg.org/kit-ty-kate/micro_httpd
+# Fetch micro_httpd
+export OPAMROOT=$(pwd)/tmp-opamroot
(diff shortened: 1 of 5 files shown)


.github/scripts/main/reftests.sh
```diff
@@ -36,7 +36,14 @@ esac
 
 export PATH="$PREFIX/bin:$PATH"
 
-git clone https://codeberg.org/kit-ty-kate/micro_httpd
+# Fetch micro_httpd
```
rjbou, line 39:
```suggestion
# Fetch micro_httpd
# We need to fetch it because it is not present in vendored deps and no opam switch is set at this point
```

rjbou (approved):
tiny comment but lgtm

kit-ty-kate:
GHA is having a moment... so merging without waiting for it based on the fact that the previous force-pushed commit (8b6d68762992ce2e549287dda1a7901c44e6d65a) was green, that https://github.com/ocaml/opam/pull/7094 was also green and that they should not conflict in any way.

PR #7034, opened 2026-07-14
Factorise OpamSystem.remove_file

kit-ty-kate:
Extracted from https://github.com/ocaml/opam/pull/6481 from @rjbou


rjbou:
Discussed in dev meeting: these functions are not strictly equivalent, we need to double check Windows Unix Error.
Related to #7093

PR #7039, opened 2026-07-15
Add some checks on `.install` `root` and `rootexec` fields

rjbou:
This PR adds some checks on the target optional field : it shouldn't begin with another field prefix. There a warning when it is the case (and failure on strict mode)
From https://github.com/ocaml/opam/pull/6938#discussion_r3570687952

Todo :
* [ ] update `Pp.warn`with an optional argument to not display warning only on verbose mode
* [ ] fix toplevel & stublibs paths
* [ ] queued on #6938


kit-ty-kate:
I would argue that having this check would be at best unnecessary and at worst annoying for users generating their .install file themselves. I can honestly see value in giving users the possibility to simply do:
```
root: [
  ... # everything here
]
```
it would make protyping a lot easier for small tools and scripts (e.g. generating opam packages from other package managers)

Issue #7031, opened 2026-07-13
opam repo add triggers a git-maintenance race condition


kit-ty-kate:
I noticed a type of error locally (as well as in CI from time to time, i think):
```diff
diff --git a/tests/reftests/legacy-git.test b/tests/reftests/legacy-git.test
index 3dbd569da..c7942b61b 100644
--- a/tests/reftests/legacy-git.test
+++ b/tests/reftests/legacy-git.test
@@ -1370,24 +1370,11 @@ P4.3
 PASS 1
 ### : REPO :
 ### opam repo add REPO2 REPO -k git
-[REPO2] Initialised
-[WARNING] The repository 'REPO2' at git+file://${BASEDIR}/REPO doesn't have a 'repo' file, and might not be compatible with this version of opam.
-[NOTE] Repository at git+file://${BASEDIR}/REPO doesn't define its version, assuming it's 1.2.
-
-<><> Upgrading repositories from older opam format ><><><><><><><><><><><><><><>
-Upgrading repository "REPO2"...
-Updated ${BASEDIR}/OPAM/repo/REPO2/packages/P1.0/opam
-Updated ${BASEDIR}/OPAM/repo/REPO2/packages/P1.1/opam
-Updated ${BASEDIR}/OPAM/repo/REPO2/packages/P2.1/opam
-Updated ${BASEDIR}/OPAM/repo/REPO2/packages/P3.1~weird-version.test/opam
-Updated ${BASEDIR}/OPAM/repo/REPO2/packages/P4.1/opam
-Updated ${BASEDIR}/OPAM/repo/REPO2/packages/P4.2/opam
-Updated ${BASEDIR}/OPAM/repo/REPO2/packages/P4.3/opam
-Updated ${BASEDIR}/OPAM/repo/REPO2/packages/P5.1/opam
-[NOTE] Repository REPO2 has been added to the selections of switch 10+a+b only.
-       Run `opam repository add REPO2 --all-switches|--set-default' to use it in all existing switches, or in newly created switches, respectively.
-
+[ERROR] Could not update repository "REPO2": ${OPAM}: "lstat" failed on ${BASEDIR}/OPAM/repo/REPO2/.git/objects/maintenance.lock: No such file or directory
+[ERROR] Initial repository fetch failed
+# Return code 40 #
 ### opam repo remove REPO2 --all
+[WARNING] No configured repositories by these names found: REPO2
 ### opam repo remove test --all
 ### : LIST :
 ### opam list -A
```
[...]

PR #7044, opened 2026-07-16
reftests: Stop transforming / into \ in command lines on Windows

kit-ty-kate:
Turns out the error shown in #7029 wasn't due to the port number at all but https://github.com/curl/curl/issues/22337 because patterns such as `${VAR}/x/y` would be rewritten `<content of var>\x\y` even in the command line itself (as opposed to the result).

### DIFF
--- a/tests/reftests/run.ml
+++ b/tests/reftests/run.ml
@@ -142,7 +142,7 @@ let escape_regexps s =
     ) s;
   Buffer.contents buf
 
-let str_replace_path ?escape whichway filters s =
+let str_replace_path ?escape ?(wrap_whole_path=true) whichway filters s =
   let s =
     match escape with
     | Some `Unescape -> Re.(replace_string (compile @@ str "\\\\") ~by:"\\" s)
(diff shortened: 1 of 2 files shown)


tests/reftests/run.ml
```diff
@@ -142,7 +142,7 @@ let escape_regexps s =
     ) s;
   Buffer.contents buf
 
-let str_replace_path ?escape whichway filters s =
+let str_replace_path ?escape ?(in_cmd=false) whichway filters s =
```
rjbou:
```suggestion
let str_replace_path ?escape ?(wrap_whole_path=true) whichway filters s =
```

tests/reftests/run.ml
```diff
@@ -160,11 +160,13 @@ let str_replace_path ?escape whichway filters s =
...
-        ) in
       match by with
+      | Sed by when in_cmd ->
+        Re.replace_string (Re.compile re) s ~by:(escape_regexps by)
       | Sed by ->
+        let re_path = Re.(
+            seq [re; group (rep (diff any (alt [set ":;$\"'"; space])))]
+          ) in
         Re.replace (Re.compile re_path) s ~f:(fun g ->
             escape_regexps by ^ escape_backslashes (whichway (Re.Group.(get g (nb_groups g - 1)))))
```
rjbou:
```suggestion
      | Sed by ->
        if wrap_whole_path then
          let re_path = Re.(
              seq [re; group (rep (diff any (alt [set ":;$\"'"; space])))]
           ) in
           Re.replace (Re.compile re_path) s ~f:(fun g ->
             escape_regexps by ^ escape_backslashes (whichway (Re.Group.(get g (nb_groups g - 1)))))
          else
            Re.replace_string (Re.compile re) s ~by:(escape_regexps by)
      
```

tests/reftests/run.ml
```diff
@@ -766,7 +768,7 @@ let run_cmd ~opam ~dir ?(vars=[]) ?(filter=[]) ?(silent=false) ?(sort=false) cmd
           if a <> "" && a.[0] = '\'' then a
           else
             str_replace_path ~escape:`Backslashes OpamSystem.forward_to_back
-              var_filters a
+              ~in_cmd:true var_filters a
```
rjbou:
```suggestion
              ~wrap_whole_path:false var_filters a
```

rjbou (approved):
Tiny proposal on argument naming.
Thanks!

kit-ty-kate:
Ignoring the macOS failure (https://github.com/ocaml/opam/issues/7031)

PR #7047, opened 2026-07-20
Add opam 2.6.0~alpha1 to the install scripts

kit-ty-kate:


kit-ty-kate:
Ignoring unused CI runs as only install-check is doing anything

PR #7052, opened 2026-07-22
Upgrade the autoconf generated files to 2.72

kit-ty-kate:
Similar to https://github.com/ocaml/ocaml/pull/14953

Only Ubuntu 26.04 has this version of autoconf so we have to upgrade the Hygiene workflow to that version of Ubuntu. 
However the name of the system package that installs dune changed to ocaml-dune so this had to change too.


rjbou (approved):
To rebase & good to go

kit-ty-kate:
Ignoring the ocaml-benchmarks failure (machine out of disk space)

PR #7056, opened 2026-07-27
Improve potentially used backtraces and discard unused backtraces

kit-ty-kate:
Minor cleanup inspired by https://github.com/ocaml/ocaml/issues/6728

This only do two things:
- use `Stdlib.raise_notrace` where the exception isn't used as an exception but as a control flow utility (the backtrace is never used)
- use `OpamStd.Exn.finalise`/`Printexc.raise_with_backtrace` where re-raises are done


gasche:
Nice!

PR #7058, opened 2026-07-27
Do not make opam init --reinit ask to retry the command when upgrading from a 2.1 root

### LINKED ISSUE #7057: opam init --reinit should not give a exit code 10 and ask to redo the command when upgrading from opam 2.1
From a 2.1 root, upgrading to 2.6 leads to:
```
$ opam init --reinit --config .opamrc-sandbox -ni
Configuring from /home/opam/.opamrc-sandbox, then /home/opam/.opamrc, and finally from built-in defaults.

This version of opam requires an update to the layout of /home/opam/.opam from version 2.1 to version 2.6~alpha, which can't be reverted.

Perform the update and continue? [Y/n] y
[NOTE] The 'jobs' option was reset, its value was 1 and its new value will vary according to the current number of cores on your machine. You can restore the fixed value using:
           opam option jobs=1 --global
Upgrading the internal repository format for 'default'...
Format upgrade done.

<><> Rerunning init and update ><><><><><><><><><><><><><><><><><><><><><><><><>

<><> Updating repositories ><><><><><><><><><><><><><><><><><><><><><><><><><><>
[default] no changes from git+file:///home/opam/opam-repository
Update done, please now retry your command.
"/usr/bin/env" "bash" "-c" "opam init --reinit --config .opamrc-sandbox -ni" failed with exit status 10
```

Given `opam init --reinit` is the main command used to upgrade from one version to the other, i don't think the command should fail and ask to redo the command again.

Noticed in opam-repo-ci, hotfixed by https://github.com/ocurrent/opam-repo-ci/commit/c33484821b7a0549aa0363b48caf5762aca9900e from @mtelvers

kit-ty-kate:
Fixes #7057

### DIFF
--- a/tests/reftests/opamroot-versions.test
+++ b/tests/reftests/opamroot-versions.test
@@ -3568,6 +3568,36 @@ STATE                           Switch state loaded in 0.000s
 i-am-package      2           One-line description
 i-am-sys-compiler 2           One-line description
 ### rm -rf _opam
+### :V:8:f: reinit from 2.2 with OPAMREPOTARRING=1
+### ocaml generate.ml 2.2
+### tar czf OPAM/repo/default.tar.gz default
+### rm -r OPAM/repo/default
+### opam init --reinit --bypass-checks --no-setup | "${OPAMROOTVERSION}($|,)" -> "current" | grep -v Cygwin
+No configuration file found, using built-in defaults.
+GSTATE                          LOAD-GLOBAL-STATE @ ${BASEDIR}/OPAM
+[WARNING] Removing global switch 'this-internal-error' as it no longer exists
+FMT_UPG                         Hard config upgrade, from 2.2 to current
+This version of opam requires an update to the layout of ${BASEDIR}/OPAM from version 2.2 to version current which can't be reverted.
+You may want to back it up before going further.
+
+Perform the update and continue? [Y/n] y
+Upgrading the internal repository format for 'default'...
+Format upgrade done.
+
+<><> Rerunning init and update ><><><><><><><><><><><><><><><><><><><><><><><><>
+GSTATE                          LOAD-GLOBAL-STATE @ ${BASEDIR}/OPAM
+RSTATE                          LOAD-REPOSITORY-STATE @ ${BASEDIR}/OPAM
+RSTATE                          No cache found
+RSTATE                          loaded opam files from repo default in 0.000s
+
+<><> Updating repositories ><><><><><><><><><><><><><><><><><><><><><><><><><><>
+[default] no changes from file://${BASEDIR}/default
+Update done, please now retry your command.
+# Return code 10 #
+### test -f OPAM/repo/default.tar.gz
+# Return code 1 #
+### test -d OPAM/repo/default
+### rm -rf _opam
 ### :V:9:a: From 2.6~alpha root, global
 ### ocaml generate.ml 2.6~alpha
 ### # ro global state
--- a/tests/reftests/opamroot-versions.test
+++ b/tests/reftests/opamroot-versions.test
@@ -1382,8 +1382,7 @@ RSTATE                          loaded opam files from repo default in 0.000s
 
 <><> Updating repositories ><><><><><><><><><><><><><><><><><><><><><><><><><><>
 [default] no changes from file://${BASEDIR}/default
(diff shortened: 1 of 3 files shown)


rjbou (approved):
Thanks! Missing changelog & good to go

tests/reftests/opamroot-versions.test
```diff
@@ -1382,8 +1382,7 @@ RSTATE                          loaded opam files from repo default in 0.000s
 
 <><> Updating repositories ><><><><><><><><><><><><><><><><><><><><><><><><><><>
 [default] no changes from file://${BASEDIR}/default
-Update done, please now retry your command.
-# Return code 10 #
```
kit-ty-kate, line 1386: for the record, 2.1\~alpha2 doesn't suffer from this because `?reinit` is passed directly in `hard_upgrade_from_2_1_intermediates` as it is the only version that actually requires it.
2.1\~rc either but this time because hard upgrade isn't necessary.

tests/reftests/opamroot-versions.test
```diff
@@ -3567,6 +3566,36 @@ STATE                           Switch state loaded in 0.000s
 # Name            # Installed # Synopsis
 i-am-package      2           One-line description
 i-am-sys-compiler 2           One-line description
+### :V:8:f: reinit from 2.2 with OPAMREPOTARRING=1
```
rjbou, line 3570:
```suggestion
### rm -rf _opam
### :V:8:f: reinit from 2.2 with OPAMREPOTARRING=1
```

tests/reftests/opamroot-versions.test
```diff
@@ -3568,6 +3567,37 @@ STATE                           Switch state loaded in 0.000s
...
+<><> Rerunning init and update ><><><><><><><><><><><><><><><><><><><><><><><><>
+GSTATE                          LOAD-GLOBAL-STATE @ ${BASEDIR}/OPAM
+RSTATE                          LOAD-REPOSITORY-STATE @ ${BASEDIR}/OPAM
+RSTATE                          No cache found
+RSTATE                          loaded opam files from repo default in 0.000s
+
+<><> Updating repositories ><><><><><><><><><><><><><><><><><><><><><><><><><><>
+[default] synchronised from file://${BASEDIR}/default
+RSTATE                          ERR: Could not load ${BASEDIR}/OPAM/repo/default/packages/foo/foo.1/opam, ignored
+Update done.
```
rjbou:
```suggestion
### tar czf OPAM/repo/default.tar.gz default
### rm -r OPAM/repo/default
### opam init --reinit --bypass-checks --no-setup | "${OPAMROOTVERSION}($|,)" -> "current" | grep -v Cygwin
No configuration file found, using built-in defaults.
GSTATE                          LOAD-GLOBAL-STATE @ ${BASEDIR}/OPAM
[WARNING] Removing global switch 'this-internal-error' as it no longer exists
FMT_UPG                         Hard config upgrade, from 2.2 to current
This version of opam requires an update to the layout of ${BASEDIR}/OPAM from version 2.2 to version current which can't be reverted.
You may want to back it up before going further.

Perform the update and continue? [Y/n] y
Upgrading the internal repository format for 'default'...
Format upgrade done.

<><> Rerunning init and update ><><><><><><><><><><><><><><><><><><><><><><><><>
GSTATE                          LOAD-GLOBAL-STATE @ ${BASEDIR}/OPAM
RSTATE                          LOAD-REPOSITORY-STATE @ ${BASEDIR}/OPAM
RSTATE                          No cache found
RSTATE                          loaded opam files from repo default in 0.000s

<><> Updating repositories ><><><><><><><><><><><><><><><><><><><><><><><><><><>
[default] no changes from file://${BASEDIR}/default
Update done.
```
It avoids creating a almost empty repo, and avoids the extra log from #6872. But it doesn't trigger an update of the repo as there is no change, I don't know if it is wanted.

kit-ty-kate: The idea was to ensure no change happened but i guess we have already the one in repository.test added in #6625 so it should be sufficient. We didn't have a testcase testing updates from tar with no change but this way of doing is a plus. I've added a couple of extra `test -f` etc to that have more insurance as to the behaviour. Good idea, thanks.

rjbou (approved):
Thanks!

PR #7063, opened 2026-07-29
Fix package name inference from project directory names

### LINKED ISSUE #7043: `opam install .foo` add package `oo`
Found during #6981 review, the behaviour is present on master.
The same behaviour can be found with `pin`, `switch create`, etc.
```
### <pin:.foo/opam>
opam-version: "2.0"
### opam install .foo
[NOTE] Package oo does not exist in opam repositories registered in the current switch.
oo is now pinned to file://${BASEDIR}/.foo (version dev)
=== install 1 package
  - install oo dev (pinned)

-> retrieved oo.dev  (file://${BASEDIR}/.foo)
-> installed oo.dev
Done.

MavenRain:
Fixes #7043.

`name_from_project_dirname` (src/client/opamAuxCommands.ml) inferred a package name from a project directory name by deleting every `.` *plus the character following it* (`Re.replace_string` over the pattern `'.' + any`).  For a hidden directory `.foo` it inferred the name `oo`, which `opam install ./.foo` then happily pinned and installed.

This replaces the deletion with a cut at the **first** dot, the same name/version split `OpamPackage.of_string` uses, and infers no name at all when the basename starts with a dot (empty name part), so those directories fall into the existing "could not infer package name" warning path instead of guessing a mangled name.

| argument | before | after |
|---|---|---|
| `opam install ./.foo` | pins and installs `oo` | warning, no inference |
| `opam install .foo/opam` | pins and installs `oo` | "could not infer package name" error |
| `opam install ./2024.report` | pins and installs `2024eport` | "could not infer package name" warning |
| `opam install ./foo.1@2` | warning, no inference | pins `foo` (dev) |
| `opam install ./proj.2.0` | `proj` 2.0 | `proj` 2.0 (unchanged) |
| `opam install ./proj2.12.3` | `proj2` 12.3 | `proj2` 12.3 (unchanged) |

Notes:

* The old pattern also mangled `name.version`-style basenames at the function level (`proj2.12.3` gave `proj22`), but that never surfaced at the CLI: both call sites consult `OpamFile.OPAM.name_opt` first, and reading a bare `opam` file stamps name and version from the enclosing directory via `OpamPackage.of_filename` (the `pp`/`with_nv` hook in opamFile.ml), which already cuts at the first dot.  `name_from_project_dirname` is only consulted when that parse yields no name, i.e. an undotted basename (unchanged by this PR) or one whose name or version part is invalid, of which the leading dot is the practical case.  Cutting at the first dot keeps the fallback consistent with the reader's split instead of leaving a second, disagreeing parser.  The full set of behaviour flips on that fallback path: a leading-dot basename now declines inference (the #7043 fix);  a dotted basename with an unparseable version part, e.g. `foo.1@2`, previously declined and now infers `foo`;  a dotted basename whose stem is not a valid package name, e.g. `2024.report`, previously inferred the mangled `2024eport` and now declines;  a trailing-dot basename such as `foo.` previously declined and now infers `foo` (not in the reftest because Windows strips trailing dots from file names).
* The issue transcript uses a bare `opam install .foo`;  since #6981, `.foo` without a separator is parsed as a package atom and rejected by the CLI, so the reftest uses `./.foo` (directory form) and `.foo/opam` (file form, which exercises the `name_and_dir_of_opam_file` call site).
* The thread converged on the diagnosis but left open whether `.foo` should infer `foo` or decline inference;  this PR declines, on the grounds that a name opam cannot round-trip from the directory name is better asked of the user explicitly.  Happy to switch to stripping the leading dots if you prefer that UX.

Testing: new reftest `tests/reftests/name-inference-7043.test` covering the dot-directory through both call sites (`./.foo` and `.foo/opam`), the other two fallback flips (`2024.report` declines, `foo.1@2` infers `foo`), and `name.version` directories in both directory and file form, whose behaviour is unchanged.  Reverting the src change flips all three fallback cases (`oo` and `2024eport` pinned and installed, `foo.1@2` declining inference), so the test pins the fix.

<sub>Drafted with AI assistance;  root-caused, reviewed and tested by me.</sub>

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -23,6 +23,7 @@ users)
 ## Actions
 
 ## Install
+  * [BUG] Fix package name inference from a local project directory name: the name is everything up to the first dot, and directory names starting with a dot no longer infer a mangled name (`opam install ./.foo` used to infer, pin and install a package named `oo`) [#XXXX @MavenRain - fix #7043]
 
 ## Build (package)
 
@@ -105,6 +106,7 @@ users)
 
 ## Reftests
 ### Tests
+  * Add a test covering package name inference from local project directory names [#XXXX @MavenRain]
 
 ### Engine
 
--- a/src/client/opamAuxCommands.ml
+++ b/src/client/opamAuxCommands.ml
@@ -78,11 +78,16 @@ let remove_files_from_destdir st pfx packages =
       end
   | _ -> ()
 
+(* Guesses a package name from a project directory name: everything up
+   to the first dot, mirroring the name/version split of package
+   strings (e.g. "foo.2.0" gives "foo"). Directory names starting with
+   a dot yield no name (see #7043). *)
 let name_from_project_dirname d =
-  try
-    Some (OpamFilename.(Base.to_string (basename_dir d)) |>
-          Re.(replace_string (compile (seq [char '.'; any]))) ~by:"" |>
-          OpamPackage.Name.of_string)
+  let base = OpamFilename.(Base.to_string (basename_dir d)) in
+  let name =
+    OpamStd.Option.map_default fst base (OpamStd.String.cut_at base '.')
+  in
+  try Some (OpamPackage.Name.of_string name)
   with Failure _ -> None
 
 let url_with_local_branch = function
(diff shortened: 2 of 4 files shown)


MavenRain:
@kit-ty-kate This PR was closed without a comment, and #7043 is still open.  I would like to understand the reason, so I can adjust how I contribute.

If the project does not accept AI-assisted contributions, I will respect that and will not resubmit.

If the problem is the scope, or the choice to decline inference instead of stripping the leading dot, I can rework the patch to the minimal fix you prefer.

If the team plans to fix this internally, that is also fine.  Thank you for your time.

Issue #7000, opened 2026-07-03
opam list --debug --safe does not show any debug log


kit-ty-kate:
This behaviour has been in opam since the introduction of `--safe` in https://github.com/ocaml/opam/commit/8c9d6ed363f825bb11379041ac337cfaf8647f65, however i would argue that it makes no sense and should be changed. It's not like the combination of both would be used by anything but us or anyone actually trying to debug their problem (which this behaviour prevents)

rjbou:
I guess that the origin of that behaviour is that generally people use it in script and redirect stdout to a variable (and no stderr).
If we change this behaviour, we need to advertise it well. Like that people know that they need to update their scripts to discard stderr in order to not have extra printing if it happen `OPAMDEBUG` is set.

kit-ty-kate:
If it was the logic then it wasn't applied to other arguments or environment variables such as `OPAMVERBOSE`, which is arguably worse because `--verbose` will output on stdout and not stderr (for some reason)

Issue #7066, opened 2026-08-03
Be able to use opam libraries (like `OpamCoreConfig` into a sandboxed context


dinosaure:
At robur, we currently develop a little tool which would like to introspect packages from the current OPAM switch. It seems that when we do `OpamCoreConfig` ([here](https://github.com/robur-coop/unic/blob/f938c74a9514c61793865a9b410dfa70197b5643/lib/uniq_opam.ml#L8-L13)), our tool would like to remove the cache of the current switch (saw [here](https://github.com/ocaml/opam-repository/pull/30383)):
```
+ /home/opam/.opam/opam-init/hooks/sandbox.sh "build" "dune" "build" "-p" "unic" "-j" "255" (CWD=/home/opam/.opam/5.2/.opam-switch/build/unic.0.1.0)
Processing  2/4: [unic: dune runtest]
+ /home/opam/.opam/opam-init/hooks/sandbox.sh "build" "dune" "runtest" "-p" "unic" "-j" "255" (CWD=/home/opam/.opam/5.2/.opam-switch/build/unic.0.1.0)
- File "test/infer-simple.t", line 1, characters 0-0:
- /usr/bin/git --no-pager diff --no-index --color=always -u _build/default/test/infer-simple.t _build/default/test/infer-simple.t.corrected
- diff --git a/_build/default/test/infer-simple.t b/_build/default/test/infer-simple.t.corrected
- index be4c8f0..134249c 100644
- --- a/_build/default/test/infer-simple.t
- +++ b/_build/default/test/infer-simple.t.corrected
- @@ -5,6 +5,8 @@ Basic usage of `unic`
-    > let () = Fmt.pr "%d\n%!" 42
-    > EOF
-    $ unic infer pure/
- +  unic: OpamSystem.Internal_error("Cannot remove /home/opam/.opam/5.2/.opam-switch/packages/cache (Unix.Unix_error(Unix.EROFS, \"unlink\", \"/home/opam/.opam/5.2/.opam-switch/packages/cache\")).")
- +  [1]
-  
-  `bstr` is taken as a library with a C stub
-  
- @@ -13,7 +15,8 @@ Basic usage of `unic`
-    > let () = ignore (Bstr.create 42)
-    > EOF
-    $ unic infer stubs/
- -  bstr
[...]

PR #7067, opened 2026-08-03
Respect safe_mode when encountering an outdated cache file

kit-ty-kate:
Partially f.i.x.e.s https://github.com/ocaml/opam/issues/7066

Queued:
* [x] #7091 
* [x] #7092


rjbou (commented):
There is no full safe mode test, meanwhile, It worth having its own tiny test

kit-ty-kate:
Done. I also had to fix https://github.com/ocaml/opam/issues/7000 to be able to properly show the change.

rjbou (approved):
I've updated the test to make it more independent and untied the first one from the second. I've also added some cache content display.

It's now queued it on 2 PRs: #7091 where I extracted the debug on safe mode (i think it worth having its own PR and especially its own discussion space), and #7092 that fix reftest cache printing and display better information.

PR #7070, opened 2026-08-10
Fix generated PowerShell init script

flandia:
## Problem

Since #5864, opam has generated the following PowerShell:

```powershell
if Test-Path "...\variables.ps1" { ... }
```

PowerShell requires parentheses around an `if` condition and rejects this
script with:

```text
Missing '(' after 'if' in if statement.
```

## Fix

Generate valid PowerShell syntax:

```powershell
if (Test-Path "...\variables.ps1") { ... }
```

The Windows reftest now executes the generated script using Windows
PowerShell, ensuring future syntax errors fail the test.

## Testing

```powershell
dune build --profile=release --root . '@reftest-init-scripts.win32' --force
```

The generated script was also executed successfully with Windows PowerShell
5.1 and PowerShell 7.

### DIFF
--- a/tests/reftests/init-scripts.win32.test
+++ b/tests/reftests/init-scripts.win32.test
@@ -57,7 +57,8 @@ if ( -f '${BASEDIR}/root/opam-init/variables.csh' ) source '${BASEDIR}/root/opam
 ### cat root/opam-init/init.cmd
 if exist "${BASEDIR}/root/opam-init/variables.cmd" call "${BASEDIR}/root/opam-init/variables.cmd" >NUL 2>NUL
 ### cat root/opam-init/init.ps1
-if Test-Path "${BASEDIR}/root/opam-init/variables.ps1" { . "${BASEDIR}/root/opam-init/variables.ps1" *> $null }
+if (Test-Path "${BASEDIR}/root/opam-init/variables.ps1") { . "${BASEDIR}/root/opam-init/variables.ps1" *> $null }
+### powershell.exe -NoLogo -NoProfile -NonInteractive -ExecutionPolicy Bypass -File root/opam-init/init.ps1
 ### : Variables scripts :
 ### cat root/opam-init/variables.sh | grep -v man | grep -v MANPATH
 test -z "${OPAM_SWITCH_PREFIX:+x}" || return 0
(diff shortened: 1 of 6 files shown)


tests/reftests/init-scripts.win32.test
```diff
@@ -57,7 +57,8 @@ if ( -f '${BASEDIR}/root/opam-init/variables.csh' ) source '${BASEDIR}/root/opam
 ### cat root/opam-init/init.cmd
 if exist "${BASEDIR}/root/opam-init/variables.cmd" call "${BASEDIR}/root/opam-init/variables.cmd" >NUL 2>NUL
 ### cat root/opam-init/init.ps1
-if Test-Path "${BASEDIR}/root/opam-init/variables.ps1" { . "${BASEDIR}/root/opam-init/variables.ps1" *> $null }
+if (Test-Path "${BASEDIR}/root/opam-init/variables.ps1") { . "${BASEDIR}/root/opam-init/variables.ps1" *> $null }
+### powershell.exe -NoLogo -NoProfile -NonInteractive -ExecutionPolicy Bypass -File root/opam-init/init.ps1
```
kit-ty-kate, line 62: neat idea to at least test the syntax, it should probably be done for every shell we support (you're welcome to do it but it's not strictly necessary for this PR so no pressure)

flandia: I tried to add them but I'm not sure if it works. Would you approve the workflow and test it?

kit-ty-kate: done

flandia: The workflow failed and I guess that's due to an upstream problem which has been fixed. I updated the branch. Would you approve the workflow again? Thanks

kit-ty-kate: For the record, you can open a PR on your own fork and enable GitHub Action on it, if you want to debug something. This way i don't have to approve it every time

flandia: Thanks for the tip. The checks succeeded. If maintainers do not have further comments, I believe this PR is now ready to merge.

PR #7071, opened 2026-08-11
Stop the testsuite from generating files containing CRLF

kit-ty-kate:
Detected in https://github.com/ocaml/opam/pull/7058

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -109,6 +109,7 @@ users)
 ### Tests
 
 ### Engine
+  * Stop the testsuite from generating files containing CRLF [#7071 @kit-ty-kate]
 
 ## Github Actions
 
--- a/tests/reftests/init-ocaml-eval-variables.unix.test
+++ b/tests/reftests/init-ocaml-eval-variables.unix.test
@@ -31,7 +31,7 @@ let _ =
   in
   let name = Filename.concat opamroot "config" in
   let content = get_files Sys.argv.(1) in
-  let fd = open_out name in
+  let fd = open_out_bin name in
   List.iter (fun l -> output_string fd (l^"\n")) content;
   close_out fd
 ### rm -rf "$OPAMROOT"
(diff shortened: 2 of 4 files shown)


tests/reftests/admin.test
```diff
@@ -121,7 +121,7 @@ let () =
        ocaml_version;
      exit 1)
   else
-  let oc = open_out "%{_:name}%.config" in
+  let oc = open_out_bin "%{_:name}%.config" in
```
rjbou:
```suggestion
  let oc = open_out "%{_:name}%.config" in
```
It's gen_ocaml_config.ml

tests/reftests/admin.test
```diff
@@ -219,7 +219,7 @@ let () =
     else
       String.make 32 '0'
   in
-  let oc = open_out "%{_:name}%.config" in
+  let oc = open_out_bin "%{_:name}%.config" in
```
rjbou:
```suggestion
  let oc = open_out "%{_:name}%.config" in
```
It's gen_ocaml_config.ml

rjbou (approved):

tests/reftests/admin.test
kit-ty-kate: I chose to fix opam admin instead

src/client/opamAdminRepoUpgrade.ml
```diff
@@ -44,7 +44,7 @@ let wrapper_conf_script =
   \       ocaml_version;\n\
   \     exit 1)\n\
   \  else\n\
-  \  let oc = open_out \"%{_:name}%.config\" in\n\
+  \  let oc = open_out_bin \"%{_:name}%.config\" in\n\
```
rjbou: The proper fix would be to update upgrade code to generate the current opam repo compiler packages, cf #7074

kit-ty-kate: That seems unrelated to what this PR is trying to do

rjbou: It's true, that issue should be resolved on its own PR.
The purpose of this PR is to avoid the test suite programs to generate files containing CRLF. Generated OCaml files from `OpamAdminRepoUgrade` shouldn't be updated as it is not test code that generate file, and it is not affecting the testsuite as these files are just printed, packages are not installed.

kit-ty-kate: done

rjbou (approved):
thanks!

PR #7072, opened 2026-08-11
Add level 4 debug output of the patch operations during a load from diff

kit-ty-kate:
Useful when debugging issues discovered in https://github.com/ocaml/opam/pull/7058


src/state/opamRepositoryState.ml
```diff
@@ -287,6 +287,7 @@ let load_opams_from_diff repo diffs rt =
       (OpamConsole.colorise `blue (OpamRepositoryName.to_string repo.repo_name));
   log ~level:3 "load repo %a from diff"
     (slog OpamRepositoryName.to_string) repo.repo_name;
+  log ~level:4 "%a" (Format.pp_print_list Patch.pp_operation) diffs;
```
```suggestion
  log ~level:4 "Applying diff\n%a" (Format.pp_print_list Patch.pp_operation) diffs;
```
or something like that. We have no insurance that these 2 logs will be one after the other.

rjbou (approved):

PR #7073, opened 2026-08-11
Disable git gc/maintenance on repositories opam maintains

kit-ty-kate:
Partially f.i.x.e.s https://github.com/ocaml/opam/issues/7031

This should fix most sources of errors when using recent git versions, however this won't help users avoid this type of error during things like `opam pin add -k path .` of a git repository (maybe `--workingdir` although this one i'm not sure) or any copy or directory non-git aware scans of things that contain git repositories.

This will also not fix already created repositories. However an alternative could be to use `GIT_CONFIG_KEY_*` and `GIT_CONFIG_VALUE_*` environment variable instead. Doing it this way would fix it retroactively as well. It might also help for things like https://github.com/ocaml/opam/pull/6955 if we generalise the use of these variables.
**EDIT:** see fixup commit

A downside of disabling these detachable jobs in theory is that we could loose efficiency by not repacking once in a while but i think this should be fine.

### DIFF
--- a/src/core/opamStd.mli
+++ b/src/core/opamStd.mli
@@ -373,6 +373,28 @@ end
 
 module Env : sig
 
+  (** {3 Raw array functions} *)
+
+  type raw = string array
+
+  (** Same as {!Unix.environment} *)
+  val raw_env: unit -> raw
+
+  (** [get_from_array array key] returns the value of the first element
+      of [array] associated with [key]. Returns [None] if nothing was found. *)
+  val get_from_array: raw -> string -> string option
+
+  (** [add_and_replace array (key, value)] returns a fresh array containing
+      at least [key=value] and [array] where all the elements matching [key]
+      will be replaced by [key=value] as well.
+
+      This makes sure any applications will always associate [key] with [value]
+      as per POSIX.1-2024:
+      > If more than one string in an environment of a process has the same name,
+      > the consequences are undefined.
+      https://pubs.opengroup.org/onlinepubs/9799919799/basedefs/V1_chap08.html *)
+  val add_and_replace: raw -> (string * string) -> raw
+
   (** {3 Generic functions} *)
 
   (** Remove from a c-separated list of string the ones with the given prefix *)
@@ -436,7 +458,6 @@ module Env : sig
   val getopt_full: Name.t -> Name.t * string option
 
   val list: unit -> (Name.t * string) list
-  val raw_env: unit -> string array
 
   (** [cyg_env ~env ~cygbin ~git_location] returns [env] environment with its
       PATH variable updated with [git_location] and [cygbin] at the beginning,
--- a/src/repository/opamGit.ml
+++ b/src/repository/opamGit.ml
@@ -14,13 +14,49 @@ open OpamProcess.Job.Op
 
 (* let log fmt = OpamConsole.log "GIT" fmt *)
 
-let git_env = [|
-  "GIT_CONFIG_GLOBAL="^Filename.null;
-  "GIT_CONFIG_SYSTEM="^Filename.null;
-|]
+let git_env = [
+  "GIT_CONFIG_GLOBAL", Filename.null;
+  "GIT_CONFIG_SYSTEM", Filename.null;
+]
 
-let env () =
-  Array.append git_env (OpamProcess.default_env ())
+let git_config = [
+  "gc.autoDetach", "false";
+  "maintenance.autoDetach", "false";
+]
+
+let env =
+  let env = lazy begin
+    let env = OpamProcess.default_env () in
+    let env = List.fold_left OpamStd.Env.add_and_replace env git_env in
+    let config_count_name = "GIT_CONFIG_COUNT" in
+    let old_config_count =
+      let open OpamStd.Option.Op in
+      (OpamStd.Env.get_from_array env config_count_name >>= int_of_string_opt)
+      +! 0
+    in
+    let config_key_prefix = "GIT_CONFIG_KEY_" in
+    let config_value_prefix = "GIT_CONFIG_VALUE_" in
+    let new_config_count = old_config_count + List.length git_config in
+    let env =
+      OpamStd.Env.add_and_replace env
+        (config_count_name, Int.to_string new_config_count)
+    in
+    let rec aux env n = function
+      | [] -> env
+      | (key, value)::xs ->
+        let env =
+          OpamStd.Env.add_and_replace env
+            (config_key_prefix ^ Int.to_string n, key)
+        in
+        let env =
+          OpamStd.Env.add_and_replace env
+            (config_value_prefix ^ Int.to_string n, value)
+        in
+        aux env (n + 1) xs
+    in
+    aux env old_config_count git_config
+  end in
+  fun () -> Lazy.force env
 
 module VCS : OpamVCS.VCS = struct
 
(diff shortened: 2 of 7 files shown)


src/repository/opamGit.ml
```diff
@@ -14,13 +14,66 @@ open OpamProcess.Job.Op
 
 (* let log fmt = OpamConsole.log "GIT" fmt *)
 
-let git_env = [|
-  "GIT_CONFIG_GLOBAL="^Filename.null;
-  "GIT_CONFIG_SYSTEM="^Filename.null;
-|]
+let git_env = [
+  "GIT_CONFIG_GLOBAL", "/dev/null";
+  "GIT_CONFIG_SYSTEM", "/dev/null";
```
rjbou: What is the reason to change `Filename.null` to `/dev/null` ?

src/core/opamStd.mli
```diff
@@ -375,6 +375,10 @@ module Env : sig
 
   (** {3 Generic functions} *)
 
+  val get_from_array: string array -> string -> string option
+
+  val add_or_replace: string array -> (string * string) -> string array
+
```
rjbou: The doc is missing

src/repository/opamGit.ml
```diff
@@ -14,13 +14,66 @@ open OpamProcess.Job.Op
...
-  Array.append git_env (OpamProcess.default_env ())
+  let env =
+    List.fold_left OpamStd.Env.add_or_replace
+      (OpamProcess.default_env ()) git_env
+  in
+  let config_count_name = "GIT_CONFIG_COUNT" in
+  let old_config_count =
+    Option.value ~default:0
+      (Option.bind (OpamStd.Env.get_from_array env config_count_name)
+         int_of_string_opt)
```
rjbou:
```suggestion
    let open OpamStd.Option in
    (OpamStd.Env.get_from_array env config_count_name
    >>= int_of_string_opt)
    +! 0
```
Proposal

src/repository/opamGit.ml
```diff
@@ -45,6 +98,9 @@ module VCS : OpamVCS.VCS = struct
     OpamFilename.mkdir repo_root;
     OpamProcess.Job.of_list [
       git repo_root [ "init"; "--initial-branch=main" ];
+      (* Disable detachable jobs to avoid race conditions *)
+      git repo_root [ "config" ; "--local" ; "gc.autoDetach"; "false"];
+      git repo_root [ "config" ; "--local" ; "maintenance.autoDetach"; "false"];
```
rjbou, line 89: As we have the config done via the environment variables, do we need to keep this config commands?

src/repository/opamGit.ml
rjbou, line 104: Shouldn't we move all these config to environment variable ?

src/repository/opamGit.ml
```diff
@@ -14,13 +14,66 @@ open OpamProcess.Job.Op
...
+]
+
+let git_config = [
+  "gc.autoDetach", "false";
+  "maintenance.autoDetach", "false";
+]
 
 let env () =
-  Array.append git_env (OpamProcess.default_env ())
+  let env =
```
rjbou: Shouldn't this computation be lazy ?

src/repository/opamGit.ml
```diff
@@ -14,13 +14,66 @@ open OpamProcess.Job.Op
...
+        | None -> aux git_config env (n - 1)
+        | Some value ->
+          let value_name = config_value_prefix ^ Int.to_string n in
+          let git_config =
+            OpamStd.List.remove_assoc String.equal key git_config
+          in
+          let env = OpamStd.Env.add_or_replace env (key_name, key) in
+          let env = OpamStd.Env.add_or_replace env (value_name, value) in
+          aux git_config env (n - 1)
+  in
```
rjbou: strictly speaking, there is no need to replace already present `GIT_CONFIG_*` en vars, git will always take the last value and load from 0 to `GIT_CONFIG_COUNT`*. So even if `GIT_CONFIG_KEY_x=gc.autoDetach` is already defined, we can define `GIT_CONFIG_KEY_x+1=gc.autoDetach` to false and it will be good.

*It is not written in the documentation, but the source code [confirm](https://github.com/git/git/blob/dea0ea3582e6980ddbc1173cc8e3e9f9db91cde0/config.c#L759) local testing

rjbou (commented):
On the idea, lgtm! Somme comments to address and good to go

src/repository/opamGit.ml
kit-ty-kate: we do. For example while building a package, if the user build steps call git they might end up spawning a maintenance job which would then interfere with the removal of the build directory later.

src/repository/opamGit.ml
kit-ty-kate: we can think about it later but i think they're fine like that for now and this would be a separate task

src/repository/opamGit.ml
kit-ty-kate: ah good point, `default_env` itself is lazy so we probably should yes

src/repository/opamGit.ml
kit-ty-kate: Extracted to https://github.com/ocaml/opam/pull/7085 for discussion

src/client/opamCommands.ml
```diff
@@ -4421,7 +4421,7 @@ let clean cli =
        List.iter (fun dir ->
            match OpamFilename.(Base.to_string (basename_dir dir)) with
            | "git" ->
-             let env = OpamGit.env () in
+             let env = Lazy.force OpamGit.env in
```
rjbou:
```suggestion
             let env = OpamGit.env () in
```

src/repository/opamGit.mli
```diff
@@ -15,7 +15,7 @@
     used to make git calls more reproducible.
     Note however that it cannot be used if you need values from the global git
     configs (e.g. /etc/gitconfig or ~/.gitconfig) *)
-val env : unit -> string array
+val env : string array Lazy.t
```
rjbou:
```suggestion
val env : unit -> string array
```

src/repository/opamGit.ml
```diff
@@ -14,13 +14,47 @@ open OpamProcess.Job.Op
...
+          (config_key_prefix ^ Int.to_string n, key)
+      in
+      let env =
+        OpamStd.Env.add_and_replace env
+          (config_value_prefix ^ Int.to_string n, value)
+      in
+      aux env (n + 1) xs
+  in
+  aux env old_config_count git_config
+end
```
rjbou: We can keep the same signature, and its easier to use.
```suggestion
let env =
  let env =
    Lazy.from_fun @@ fun () ->
    let env = OpamProcess.default_env () in
    let env = List.fold_left OpamStd.Env.add_and_replace env git_env in
    let config_count_name = "GIT_CONFIG_COUNT" in
    let old_config_count =
      let open OpamStd.Option.Op in
      (OpamStd.Env.get_from_array env config_count_name >>= int_of_string_opt)
      +! 0
    in
    let config_key_prefix = "GIT_CONFIG_KEY_" in
    let config_value_prefix = "GIT_CONFIG_VALUE_" in
    let new_config_count = old_config_count + List.length git_config in
    let env =
      OpamStd.Env.add_and_replace env
        (config_count_name, Int.to_string new_config_count)
    in
    let rec aux env n = function
      | [] -> env
      | (key, value)::xs ->
        let env =
          OpamStd.Env.add_and_replace env
            (config_key_prefix ^ Int.to_string n, key)
        in
        let env =
          OpamStd.Env.add_and_replace env
            (config_value_prefix ^ Int.to_string n, value)
        in
        aux env (n + 1) xs
    in
    aux env old_config_count git_config
  in
  fun () -> Lazy.force env
```

src/repository/opamGit.ml
rjbou: ok! A comment should be added then to be sure that we remember in the future the need to have both.

rjbou (approved):
lgtm! some tiny comments remaining, but the PR is good to merge

src/repository/opamGit.ml
kit-ty-kate: done

src/repository/opamGit.mli
kit-ty-kate: If you want but i'd argue that it is dangerous to not make it explicit that the return value is memoised when returning a mutable value. Although this is a different issue and should be applied to `OpamProcess.default_env` too in a separate PR later. I've opened https://github.com/ocaml/opam/pull/7088 for that.

PR #7088, opened 2026-08-20
Make memoised mutable structure explicitly lazy

kit-ty-kate:
Per https://github.com/ocaml/opam/pull/7073#discussion_r3821587404


kit-ty-kate:
We could change the type to instead not be mutable but an `OpamStd.Env.Map`

PR #7090, opened 2026-08-20
Disable git's fsmonitor on repositories managed by opam

### LINKED ISSUE #7054: repository set-url fails when Git fsmonitor leaves a socket under .git
```
# opam config report
# opam-version         2.5.2
# self-upgrade         no
# system               arch=arm64 os=macos os-distribution=homebrew os-version=26.5.2
# jobs                 9
# repositories         1 (http), 2 (version-controlled)
# current-switch       none set
```

## Description

With `git config --global core.fsmonitor true`, Git creates a Unix socket at `.git/fsmonitor--daemon.ipc`. Commands that copy a repository root (notably `opam repository set-url`) then fail:

```
Fatal error: Copying sockets (.../.opam/repo/default/.git/fsmonitor--daemon.ipc) is unsupported
```

or, during the subsequent update:

```
[ERROR] Could not update repository "default": Copying sockets (.../.opam/repo/default/.git/fsmonitor--daemon.ipc) is unsupported
```

## Reproducer

1. `git config --global core.fsmonitor true`
2. Ensure the default repository is a git clone under `~/.opam/repo/default` (or run any `git` command there so the fsmonitor daemon starts).
3. `opam repository set-url default git+https://github.com/ocaml/opam-repository.git`

## Notes

- `OpamSystem.copy_dir` raises on `Unix.S_SOCK` (`src/core/opamSystem.ml`).
- On 2.5.2 this hits the backup `copy_dir` in `repository set-url`, and again the quarantine `copy_dir` in the VCS first-init path (`OpamFilename.copy_dir` of `repo` → `repo.new` after `git fetch`, which itself creates the socket when fsmonitor is enabled).
- Master isolates git via `GIT_CONFIG_GLOBAL`/`GIT_CONFIG_SYSTEM` (`2b1536a9a`), which avoids creating the socket during opam's own git calls, but an already-present socket (or one created by an external `git` in that tree) still breaks any full directory copy.

## Workaround

```
git -C ~/.opam/repo/default fsmonitor--daemon stop
rm -f ~/.opam/repo/default/.git/fsmonitor--daemon.ipc
# temporarily:
git config --global core.fsmonitor false
```

A robust fix would likely skip uncopyable runtime artefacts such as sockets in `copy_dir` (or stop/remove the fsmonitor daemon before copying `.git`).

kit-ty-kate:
Fixes #7054 
As mentioned in this ticket this is barely worth doing as one would need to set an undocumented environment variable to trigger the issue but i guess it's best to make sure it never happens and we already have the code for it anyway.

### DIFF
--- a/src/repository/opamGit.ml
+++ b/src/repository/opamGit.ml
@@ -22,6 +22,7 @@ let git_env = [
 let git_config = [
   "gc.autoDetach", "false";
   "maintenance.autoDetach", "false";
+  "core.fsmonitor", "false";
 ]
 
 let env =


src/repository/opamGit.ml
```diff
@@ -22,6 +22,7 @@ let git_env = [
 let git_config = [
   "gc.autoDetach", "false";
   "maintenance.autoDetach", "false";
+  "core.fsmonitor", "false";
```
kit-ty-kate, line 25: TODO:
- add a comment here telling people to think about adding `git config --local` variable
- add the `git config --local` variable
- idea from @rjbou: we could also simply map the list to create these `git config --local` instead of duplicating them.

PR #7092, opened 2026-08-20
Fix and enhance some cache handling in restest engine

rjbou:
While reviewing #7067, i found some missing features and a bug:
* when we print a cache, we load it via opam code, that can do operations on cache. It is now a no-op
* more specific cache status:
  * dissociate no cache file and invalid cache
  * print a message when the package selection is empty instead of no printing
  * handle the opam magic number, as it is used in `state-xxx.cache` file (now we can name it in a test), and like that it will automatically replace all magic number automatically (they can appear on debug log, like invalid cache).

### DIFF
--- a/tests/reftests/repository.test
+++ b/tests/reftests/repository.test
@@ -19,10 +19,11 @@ some content
 ### : Internal repository storage as archive or plain directory :
 ### opam switch create tarring --empty
 ### opam update -vv | grep '^\+'
-### ls $OPAMROOT/repo | grep -v "cache"
+### ls $OPAMROOT/repo
 default
 lock
 repos-config
+state-${MAGICV}.cache
 ### opam install foo.1 -vv | grep '^\+' | sed-cmd test
 + test "-f" "bar" (CWD=${BASEDIR}/OPAM/tarring/.opam-switch/build/foo.1)
 ### opam repository add plain ./REPO --this-switch
@@ -38,11 +39,12 @@ some content
 ### OPAMDEBUG=-1 OPAMDEBUGSECTIONS=TAR opam update default | grep TAR
 TAR                             creating archive ${BASEDIR}/OPAM/repo/default.tar.gz from ${BASEDIR}/OPAM/repo/default
 TAR                             creating archive ${BASEDIR}/OPAM/repo/default.tar.gz.new from ${BASEDIR}/REPO
-### ls $OPAMROOT/repo | grep -v "cache"
+### ls $OPAMROOT/repo
 default.tar.gz
 lock
 plain
 repos-config
+state-${MAGICV}.cache
 ### opam install foo.2 -vv | grep '^\+' | sed-cmd test tar
 + test "-f" "baz" (CWD=${BASEDIR}/OPAM/tarring/.opam-switch/build/foo.2)
 ### opam repository remove default --all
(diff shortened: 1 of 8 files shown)


tests/reftests/run.ml
```diff
@@ -990,6 +988,7 @@ let run_test ?(vars=[]) ~opam t =
     "OPAMROOT", opamroot;
     "BASEDIR", dir;
     "PATH", Sys.getenv "PATH";
+    "MAGICV", opam_magicv;
```
rjbou, line 991: I'm not sure about the name, `MAGICV`, `OPAMMAGICV`, `OPAMMAGICVERSION` ?

kit-ty-kate: MAGICV is fine imo

tests/reftests/run.ml
```diff
@@ -1219,73 +1218,100 @@ let run_test ?(vars=[]) ~opam t =
                     nvs)
               OpamPackage.Name.Map.empty nvs
           in
+          let save_and_restore file k =
```
rjbou, line 1221: This part is to review with hiding whitespace

tests/reftests/repository.test
```diff
@@ -133,10 +138,10 @@ The following actions will be performed:
 -> removed   foo.5
 -> installed foo.4
 Done.
-### opam update --debug-level=-3 | "state-[0-9A-Z]{8}" -> "state-magicv"
+### opam update --debug-level=-3 | "state-[0-9A-Z]{8}" -> "state-${MAGICV}"
```
kit-ty-kate: Why not simply remove the Sed?
```suggestion
### opam update --debug-level=-3
```

tests/reftests/repository.test
```diff
@@ -169,10 +174,10 @@ some content
 ### sh hash.sh REPO2 bar.1
 ### opam repository add repo2 ./REPO2 --this-switch
 [repo2] Initialised
-### opam update --debug-level=-3 | "state-[0-9A-Z]{8}" -> "state-magicv"
+### opam update --debug-level=-3 | "state-[0-9A-Z]{8}" -> "state-${MAGICV}"
```
kit-ty-kate:
```suggestion
### opam update --debug-level=-3
```

tests/reftests/run.ml
```diff
@@ -1219,73 +1218,100 @@ let run_test ?(vars=[]) ~opam t =
...
+            let tmpdir = OpamFilename.mk_tmp_dir () in
+            let backup = OpamFilename.Op.(tmpdir // "cache-backup") in
+            OpamFilename.copy ~src:file ~dst:backup;
+            k ();
+            OpamFilename.copy ~src:backup ~dst:file;
+            OpamFilename.rmdir tmpdir
+          in
```
kit-ty-kate, line 1228: in theory, we don't need this if we merge https://github.com/ocaml/opam/pull/7067 first and set safe_mode=true right?

rjbou: In theory yes. But I prefer to keep like that if there is any change, it doesn't affect the tests : printing shouldn't change opam internals state...

tests/reftests/repository.test
rjbou: because i used a sed to update the test, to avoid lines reordering

kit-ty-kate (approved):

kit-ty-kate:
Ignoring the benchmark failure (machine failure) and the windows failure (fixed by https://github.com/ocaml/opam/pull/7094)

PR #7097, opened 2026-08-21
Reproducible windows release

### LINKED ISSUE #6752: Reproducible Windows binary
Currently the Windows build isn't reproducible as the time and date is stored in the binary header.
I'm not exactly sure how the flexlink uses the native linker, but if it uses `link.exe` even with mingw we might be able to pass `/Brepro` to it, which apparently might fix that issue according to [this blog post](https://nikhilism.com/post/2020/windows-deterministic-builds/)

kit-ty-kate:
Fixes #6752 
Queued on:
- [x] https://github.com/ocaml/opam/pull/7096

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -76,6 +76,7 @@ users)
 
 ## Release scripts
   * Harden the release script against known `sshpass` instability [#7096 @kit-ty-kate]
+  * Make the Windows binary reproducible [#7097 @kit-ty-kate - fix #6752]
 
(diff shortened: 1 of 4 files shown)


rjbou (approved):
Discussed in dev meeting: to test for next release, if it succeed, good to go!

kit-ty-kate:
TODO: add a step to check reproducibility (relaunch the release script) in release/readme.txt

PR #7099, opened 2026-08-22
Add 2.6.0~beta1 to the install scripts

kit-ty-kate:


kit-ty-kate:
install-checks CI is green, the rest doesn't matter. Merging.

PR #7101, opened 2026-08-22
Stop scanning opam projects recursively for nautht

### LINKED ISSUE #7098: opam install --deps-only . takes forever on a project containing a large number of files
Reproduction case:
```
git clone https://github.com/ocaml/opam
cd opam
./configure
make test
opam install --deps .
```

kit-ty-kate:
2.6 regression from https://github.com/ocaml/opam/commit/c9dda299a1788a87b8dfb71c16425502099d1f64
Fixes #7098 

When pinning packages, "extra-files" (`descr`, `url`, `files/**`) are scanned in the project's directory, however `rec_files` was used indiscriminately on the root directory thus scanning unused files and causing heavy use of memory and time on larger projects (or projects that have uncommited files such as in `_build`, `.git`, …) for nothing.

### DIFF
--- a/src/state/opamFileTools.ml
+++ b/src/state/opamFileTools.ml
@@ -1518,10 +1518,21 @@ let get_extrafiles ~repo_root dir =
         OpamRepositoryRoot.remove_prefix repo_root file
     | None -> OpamFilename.Unix.of_filename
   in
-  List.fold_left (fun map file ->
-      OpamFilename.Unix.Map.add (to_key file)
-        (lazy (OpamFilename.read file)) map)
-    OpamFilename.Unix.Map.empty (OpamFilename.rec_files dir)
+  let aux map file =
+    OpamFilename.Unix.Map.add (to_key file)
+      (lazy (OpamFilename.read file))
+      map
+  in
+  let add_file_if_exists map file =
+    if OpamFilename.exists file
+    then aux map file
+    else map
+  in
+  let map = OpamFilename.Unix.Map.empty in
+  let map = add_file_if_exists map OpamFilename.Op.(dir // "url") in
+  let map = add_file_if_exists map OpamFilename.Op.(dir // "descr") in
+  let files_dir = OpamFilename.Op.(dir / OpamPathName.files_d) in
+  List.fold_left aux map (OpamFilename.rec_files files_dir)
 
 let get_contents ~repo_root dir =
   let file = OpamFilename.Op.(dir // OpamPathName.opam_f) in
--- a/tests/reftests/action-disk.test
+++ b/tests/reftests/action-disk.test
@@ -110,11 +110,11 @@ SYSTEM                          mkdir ${BASEDIR}/OPAM/repo/default/packages/main
 SYSTEM                          mkdir ${BASEDIR}/OPAM/repo/default/packages/main-repo/main-repo.2/files
 FILE(repo)                      Read ${BASEDIR}/OPAM/repo/default/repo in 0.000s
 Processing: [default: loading data]
-SYSTEM                          scan files recursively (except_vcs: false) in ${BASEDIR}/OPAM/repo/default/packages/main-repo/main-repo.2
+SYSTEM                          scan files recursively (except_vcs: false) in ${BASEDIR}/OPAM/repo/default/packages/main-repo/main-repo.2/files
 SYSTEM                          read ${BASEDIR}/OPAM/repo/default/packages/main-repo/main-repo.2/opam
 FILE(opam)                      Read packages/main-repo/main-repo.2/opam (out of ${BASEDIR}/OPAM/repo/default) in 0.000s
 SYSTEM                          read ${BASEDIR}/OPAM/repo/default/packages/main-repo/main-repo.2/files/x-file.main-repo.2
-SYSTEM                          scan files recursively (except_vcs: false) in ${BASEDIR}/OPAM/repo/default/packages/main-repo/main-repo.1
+SYSTEM                          scan files recursively (except_vcs: false) in ${BASEDIR}/OPAM/repo/default/packages/main-repo/main-repo.1/files
 SYSTEM                          read ${BASEDIR}/OPAM/repo/default/packages/main-repo/main-repo.1/opam
 FILE(opam)                      Read packages/main-repo/main-repo.1/opam (out of ${BASEDIR}/OPAM/repo/default) in 0.000s
 SYSTEM                          read ${BASEDIR}/OPAM/repo/default/packages/main-repo/main-repo.1/files/x-file.main-repo.1
@@ -682,8 +682,8 @@ SYSTEM                          LOCK ${BASEDIR}/OPAM/repo/state-${MAGICV}.cache 
 SYSTEM                          LOCK ${BASEDIR}/OPAM/install-from-path-pin-all/.opam-switch/lock (none => write)
 FILE(switch-config)             Read ${BASEDIR}/OPAM/install-from-path-pin-all/.opam-switch/switch-config in 0.000s
 FILE(switch-state)              Read ${BASEDIR}/OPAM/install-from-path-pin-all/.opam-switch/switch-state in 0.000s
-SYSTEM                          scan files recursively (except_vcs: false) in ${BASEDIR}/OPAM/install-from-path-pin-all/.opam-switch/overlay/main-ppin
-SYSTEM                          scan files recursively (except_vcs: false) in ${BASEDIR}/OPAM/install-from-path-pin-all/.opam-switch/overlay/main-ppin
+SYSTEM                          scan files recursively (except_vcs: false) in ${BASEDIR}/OPAM/install-from-path-pin-all/.opam-switch/overlay/main-ppin/files
+SYSTEM                          scan files recursively (except_vcs: false) in ${BASEDIR}/OPAM/install-from-path-pin-all/.opam-switch/overlay/main-ppin/files
 SYSTEM                          scan files recursively (except_vcs: false) in ${BASEDIR}/OPAM/install-from-path-pin-all/.opam-switch/sources/main-ppin/files
 SYSTEM                          read ${BASEDIR}/OPAM/install-from-path-pin-all/.opam-switch/overlay/main-ppin/opam
 FILE(opam)                      Read ${BASEDIR}/OPAM/install-from-path-pin-all/.opam-switch/overlay/main-ppin/opam in 0.000s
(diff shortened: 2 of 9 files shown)


tests/reftests/package-format-upgrade.test
```diff
@@ -0,0 +1,178 @@
...
+name: "local-pkg"
+version: "dev"
+depends: ["ocaml"]
+build: [
+  ["cat" "some-extrafile"]
+  ["cat" "dir/some-other-extrafile"]
+]
+url {
+  src: "file://${BASEDIR}/local-pkg"
+}
```
kit-ty-kate: This test seems to show that neither `url`, `descr` or the `files` directory are read during a package upgrade on pin. Was that always the case?

kit-ty-kate: I was able to reproduce this with 2.5 (see https://github.com/ocaml/opam/commit/263688ef6b40ce11cf3d52c80721b8388663382e) so this isn't a regression

kit-ty-kate: Actually there is a change of behaviour (the descr and extrafiles are used for the upgrades) compared to 2.1 but only when using autopin. Full fat pins and installs from a repository have the same behaviour.

tests/reftests/action-disk.test
```diff
@@ -766,7 +770,10 @@ SYSTEM                          LOCK ${BASEDIR}/OPAM/repo/state-${MAGICV}.cache
 SYSTEM                          LOCK ${BASEDIR}/OPAM/install-from-path-pin-all/.opam-switch/lock (none => write)
 FILE(switch-config)             Read ${BASEDIR}/OPAM/install-from-path-pin-all/.opam-switch/switch-config in 0.000s
 FILE(switch-state)              Read ${BASEDIR}/OPAM/install-from-path-pin-all/.opam-switch/switch-state in 0.000s
-SYSTEM                          scan files recursively in ${BASEDIR}/OPAM/install-from-path-pin-all/.opam-switch/overlay/main-ppin
+SYSTEM                          scan files recursively in ${BASEDIR}/OPAM/install-from-path-pin-all/.opam-switch/overlay/main-ppin/files
+SYSTEM                          scan files in ${BASEDIR}/OPAM/install-from-path-pin-all/.opam-switch/overlay/main-ppin
+SYSTEM                          scan files recursively in ${BASEDIR}/OPAM/install-from-path-pin-all/.opam-switch/overlay/main-ppin/files
+SYSTEM                          scan files in ${BASEDIR}/OPAM/install-from-path-pin-all/.opam-switch/overlay/main-ppin
```
kit-ty-kate: why are we scanning that directory twice?

kit-ty-kate: oh my bad, it's unordered

tests/reftests/package-format-upgrade.test
rjbou: it worth opening an issue for that

tests/reftests/package-format-upgrade.test
```diff
@@ -0,0 +1,241 @@
...
+Some random unused file
+### <local-pkg/files/some-extrafile>
+here is an extra file
+### <local-pkg/files/dir/some-other-extrafile>
+here is another extra file
+### <local-pkg/descr>
+Some descr
+More descr
+### sh url.sh local-pkg/url
+### opam show --normalise --raw local-pkg/opam | '$ARCHIVE_MD5' -> '+md5+' | '$HTTPSERVERPORT' -> '+port+'
```
rjbou: you can use ``sed-hash``

tests/reftests/package-format-upgrade.test
```diff
@@ -0,0 +1,241 @@
...
+### mkdir dir
+### tar czf file.tar.gz dir
+### openssl md5 file.tar.gz | '.*= ' -> '' >$ ARCHIVE_MD5
+### <url.sh>
+#!/bin/sh
+
+set -eu
+
+cat > "$1" << EOF
+src: "http://localhost:$HTTPSERVERPORT/file.tar.gz"
```
rjbou: It is possible to use a local archive for the test. There is no need to launch the http server

tests/reftests/package-format-upgrade.test
```diff
@@ -0,0 +1,241 @@
...
+  ["cat" "some-extrafile"]
+  ["cat" "dir/some-other-extrafile"]
+]
+### :II: format upgrade a package in a local dir
+### <local-pkg/opam>
+opam-version: "1.2"
+build: [
+  ["cat" "some-extrafile"]
+  ["cat" "dir/some-other-extrafile"]
+]
```
rjbou:
```suggestion
]
synopsis: "A word"
description: "Two words."
authors: "the testing team"
homepage: "egapemoh"
maintainer: "maint@tain.er"
license: "MIT"
dev-repo: "hg+https://pkg@op.am"
bug-reports: "https://nobug"
```
like that there is no linting nagging

tests/reftests/package-format-upgrade.test
rjbou:
```suggestion
### :II:1: via show
### opam show --normalise --raw local-pkg/opam | '$ARCHIVE_MD5' -> '+md5+' | '$HTTPSERVERPORT' -> '+port+'
```

tests/reftests/package-format-upgrade.test
```diff
@@ -0,0 +1,241 @@
...
+depends: ["ocaml"]
+build: [
+  ["cat" "some-extrafile"]
+  ["cat" "dir/some-other-extrafile"]
+]
+url {
+  src: "http://localhost:+port+/file.tar.gz"
+  checksum: "md5=+md5+"
+}
+### :II:1: via pin
```
rjbou:
```suggestion
### :II:2: via pin
```

tests/reftests/package-format-upgrade.test
```diff
@@ -0,0 +1,241 @@
...
+
+<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
++- The following actions failed
+| - build local-pkg dev
++- 
+- No changes have been performed
+# Return code 31 #
+### opam unpin ./local-pkg
+Ok, local-pkg is no longer pinned to file://${BASEDIR}/local-pkg (version dev)
+### :II:2: via autopin
```
rjbou:
```suggestion
### :II:3: via autopin
```

tests/reftests/package-format-upgrade.test
```diff
@@ -0,0 +1,241 @@
...
+
+
+<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
++- The following actions failed
+| - build local-pkg dev
++- 
+- No changes have been performed
+# Return code 31 #
+### opam unpin ./local-pkg
+Ok, local-pkg is no longer pinned to file://${BASEDIR}/local-pkg (version dev)
```
rjbou:
```suggestion
Ok, local-pkg is no longer pinned to file://${BASEDIR}/local-pkg (version dev)
### cat local-pkg/opam
opam-version: "1.2"
build: [
  ["cat" "some-extrafile"]
  ["cat" "dir/some-other-extrafile"]
]
```
As for the latest test, we can check that nothing changed on the source dir

src/state/opamFileTools.ml
```diff
@@ -1518,10 +1518,17 @@ let get_extrafiles ~repo_root dir =
...
+    OpamFilename.Unix.Map.add (to_key file)
+      (lazy (OpamFilename.read file))
+      map
+  in
+  let map = OpamFilename.Unix.Map.empty in
+  let map =
+    let dir = OpamFilename.Op.(dir / OpamPathName.files_d) in
+    List.fold_left aux map (OpamFilename.rec_files dir)
+  in
+  List.fold_left aux map (OpamFilename.files dir)
```
rjbou: Why not just read descr & url ? Like that if there is a lot of files in the directory, we only give in argument the files that we know we would want to read.
Agree that the order of magnitude is isn't the same than a `_build` dir.

kit-ty-kate: we could but this duplicates the logic of read_opam

rjbou: we already retrieve a subset of present files (only the one at root and the ones in the `files/` directory).

tests/reftests/package-format-upgrade.test
kit-ty-kate: i'd rather have the nagging and avoid the cruft personally

src/state/opamFileTools.ml
kit-ty-kate: fair. done

tests/reftests/extrafile.test
```diff
@@ -700,3 +700,331 @@ The following actions will be performed:
...
+           warning 36: Missing field 'bug-reports'
+             error 57: Synopsis must not be empty
+           warning 68: Missing field 'license'
+opam-version: "2.0"
+name: "local-pkg"
+version: "dev"
+build: [
+  ["test" "-f" "standalone-file"]
+  ["test" "-f" "some-extrafile"]
+  ["test" "!" "-f" "dir/some-other-extrafile"]
```
kit-ty-kate: This is a bug

tests/reftests/package-format-upgrade.test
kit-ty-kate: in an ideal world i'd rather have an `opam {install,pin} --no-lint` instead

tests/reftests/extrafile.test
kit-ty-kate: ah nevermind, this is an issue in my test

tests/reftests/package-format-upgrade.test
kit-ty-kate: Done in https://github.com/ocaml/opam/issues/7109

tests/reftests/package-format-upgrade.test
kit-ty-kate: See https://github.com/ocaml/opam/issues/7110

src/state/opamFileTools.ml
```diff
@@ -1518,10 +1518,21 @@ let get_extrafiles ~repo_root dir =
...
+      (lazy (OpamFilename.read file))
+      map
+  in
+  let add_file_if_exists map file =
+    if OpamFilename.exists file
+    then aux map file
+    else map
+  in
+  let map = OpamFilename.Unix.Map.empty in
+  let map = add_file_if_exists map OpamFilename.Op.(dir // "url") in
```
rjbou, line 1532: Not in this PR : in worth adding `url` and `descr` files in `*PathName` module

tests/reftests/extrafile.test
```diff
@@ -700,3 +700,297 @@ The following actions will be performed:
...
+here is another extra file
+### <local-pkg/opam/file-in-opam-dir>
+This shouldn't be an extrafile
+### <local-pkg/opam/files/file-in-opam-dir-slash-files>
+This shouldn't be an extrafile either
+### openssl md5 local-pkg/standalone-file | '.*= ' -> '' >$ STANDALONE_MD5
+### openssl md5 local-pkg/files/some-extrafile | '.*= ' -> '' >$ EXTRAFILE_MD5
+### openssl md5 local-pkg/files/dir/some-other-extrafile | '.*= ' -> '' >$ OTHEREXTRAFILE_MD5
+### openssl md5 local-pkg/opam/file-in-opam-dir | '.*= ' -> '' >$ FILEINOPAM_MD5
+### openssl md5 local-pkg/opam/files/file-in-opam-dir-slash-files | '.*= ' -> '' >$ FILEINOPAMSLASHFILES_MD5
```
rjbou, line 724: if you copy the first file into the others, you don't have the same hash ?

tests/reftests/extrafile.test
```diff
@@ -700,3 +700,297 @@ The following actions will be performed:
...
+### <extrafiles.sh>
+#!/bin/sh
```
rjbou, line 726: no needed as we launche via `sh`

rjbou (approved):
Tiny comment, but lgtm! Good to go once CI is green

tests/reftests/extrafile.test
kit-ty-kate: i prefer having it, as it makes it explicit at creation time, what time of shell is being used.

tests/reftests/extrafile.test
kit-ty-kate: this is on purpose, i wanted multiple files with different content

kit-ty-kate:
As a bonus, this seems to have improved performance of `opam update` by 8% (although ocaml-benchmarks seems to be a bit confused as to which commit did it, given that the merge commit appears before the latest commit of this branch in its timeline)

PR #7106, opened 2026-08-27
Stop scanning certain directories twice and attempt removal of a non-existing directory during removal of packages

kit-ty-kate:
Queued on https://github.com/ocaml/opam/pull/7101 for the relevant logs


rjbou (commented):
On the idea lgtm! If possible add a test, changelog, and good to go.

PR #7107, opened 2026-08-27
reftests: Make unordered test results real

kit-ty-kate:
"Real" as opposed to trying to reduce the diff at the cost of jumbled order that cannot happen in real life.


tests/reftests/dot-install.test
```diff
@@ -830,18 +833,15 @@ SYSTEM                          rmdir ${BASEDIR}/OPAM/escapability/doc/realeskap
...
+SYSTEM                          LOCK ${BASEDIR}/OPAM/escapability/.opam-switch/packages/cache (write => none)
 
 
 <><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
 +- The following actions failed
 | - install realeskape-rel-middotdot-notexists 1
 +- 
 - No changes have been performed
-            - At ${BASEDIR}/OPAM/escapability/.opam-switch/build/realeskape-rel-middotdot-notexists.1/realeskape-rel-middotdot-notexists.install:14:13-14:44::
-              tmp/../../../../middle/a-file is not an absolute filename.
```
kit-ty-kate, line 841: here's an example of "unreal" test output that happened to have been created by the unordered sorting system. There are other examples in this diff

PR #7116, opened 2026-08-31
Upgrade the "lockfile" with all the most up-to-date dependencies available

kit-ty-kate:
Changes:
- ~dune ¯\\\_(ツ)\_/¯ however it cannot be updated to its current latest version due to https://github.com/ocaml/dune/issues/15340~ EDIT: not updated. There are complications with OCaml 5.6
- base64 removed support for OCaml < 4.07
- spdx_licenses upgraded to Version 3.28.0 of the SPDX License List
- menhir removed support for OCaml < 4.08 and fixed a couple of bugs that do not impact us
- patch fixed an issue arising with handwritten patches https://github.com/hannesm/patch/pull/42
- checkseum fixed some C warnings on recent clang versions
- ~decompress sped up compression~ EDIT: not updated. 32bit support is missing

I noticed the C warnings fixed in the `checkseum` release while building the macOS binaries for 2.6.0\~beta2

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -76,6 +76,7 @@ users)
 ## Infrastructure
 
 ## Release scripts
+  * Update the OCaml compiler used for release to 4.14.4 [#7116 @kit-ty-kate]
 
 ## Install script
 
--- a/master_changes.md
+++ b/master_changes.md
@@ -71,6 +71,7 @@ users)
 ## VCS
 
 ## Build
+  * Upgrade the download-if-missing dependencies to their latest version available for us to use (ocaml.4.14.4, base64.3.5.2, spdx\_licenses.1.5.0, menhir.20260209, patch.3.1.2, checkseum.0.5.3) [#7116 @kit-ty-kate]
 
 ## Infrastructure
 
--- a/release/Makefile
+++ b/release/Makefile
@@ -8,7 +8,7 @@ VERSION = $(shell git describe $(TAG))
 OPAM_VERSION = $(subst -,~,$(VERSION))
 GIT_URL = ..
 
-OCAMLV = 4.14.2
+OCAMLV = 4.14.4
 FLEXDLLV = 0.44
 # currently hardcoded in Dockerfile.in
 OCAML_URL = https://github.com/ocaml/ocaml/archive/refs/tags/$(OCAMLV).tar.gz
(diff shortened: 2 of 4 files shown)


hannesm:
excellent work!

kit-ty-kate:
@hannesm thanks for commenting, i had almost forgotten to update the OCaml compiler versions to 4.14.4 :D

```
  File "src_ext/decompress/lib/de.ml", line 4055, characters 20-30:
  4055 |   let _hash_magic = 0x9e3779b1 (* xxHash *)
                             ^^^^^^^^^^
  Error: Integer literal exceeds the range of representable integers of type int
```
@dinosaure does decompress not support 32bit platforms anymore?

dinosaure:
> @dinosaure does decompress not support 32bit platforms anymore?

It's indeed the case due to my last upgrade (`decompress.1.6.0`), I can figure out (with `optint`) if it's really required to keep the support on 32-bits (I was probably too optimistic).

kit-ty-kate:
> let _hash_magic = 0x9e3779b1 (* xxHash *)

doing a quick search yesterday on the why of this magic number i found out that it seems to be used exactly because it fits into 32bit ints. However since OCaml has 31bit ints it doesn't fit anymore. Why not use `Int32.t` or simply reduce the precision to something like `0x4f1bbcdc` (aka. `int_of_float ((sqrt 5. -. 1.) /. 2.) *. (2. ** 31.)`) ?

dinosaure:
> Why not use Int32.t or simply reduce the precision to something like 0x4f1bbcdc (aka. int_of_float ((sqrt 5. -. 1.) /. 2.) *. (2. ** 31.)) ?

This part of decompress is about the deflation, use a boxed `int32` value will probably slow down the deflation process (and how we fill/search on our internal hash table with patterns). We can probably give a try to an `int32` and see if we can have a regression about the `decompress`'s throughput.

For the global perspective, mirage/decompress#171 is basically the difference between the old `zlib` (which would like to keep a certain compatibility with really old computers) and `zlib-ng`. Tell me if you are interested by such experiment (we can also take `0x4f1bbcdc` also).

rjbou (approved):
Thanks!
A tiny comment on first commit message. Proposal
```diff
- dune is not updated due to complications with OCaml 5.6 and
- ocaml/dune#15340
+ dune is not updated due to complications with OCaml 5.6 and
+ build on Cygwin issue (ocaml/dune#15340)

kit-ty-kate:
Ignoring the CI failure on Windows (https://github.com/ocaml/opam/issues/6890)

PR #7119, opened 2026-08-31
Add 2.6.0~beta2 to the install scripts

kit-ty-kate:


kit-ty-kate:
Merging as the install-check run is green.

Issue #5958, opened 2024-05-16
TODO: merge make cold and the release script


kit-ty-kate:
All three scripts (Makefile, release/Makefile and release/Dockerfile.in) compile the OCaml compiler, the version number isn't even kept in sync and it's all very annoying to deal with. On top of that they also all include a way to cold build and all do it in a slightly different way.

The proposal would be to:
* get rid of the duplicated scripts by creating a new `make release` target that would create a stripped (and static where possible) binary
* backport these change into the 2.1 and upcoming 2.2 branches so that the branch itself is used to build the releases instead of taking the release script from the latest commit from master. This would maybe imply that we want to move the part of the Makefile that deals with `make cold`-type rules on their own Makefile for easier sync between master and the maintenance branches

PR #7121, opened 2026-08-31
release: use 'make compiler' instead of duplicating compiler build steps

kit-ty-kate:
Queued on https://github.com/ocaml/opam/pull/7120
Partially f.i.x.e.s https://github.com/ocaml/opam/issues/5958

Two downsides:
- it would no longer possible (or at least harder) to rebuild an opam release with a different ocaml version
- bootstrap-ocaml.sh is harder to read than the current release script


rjbou:
Discussed in dev meeting: The good point of this PR is to avoid changing the compiler information in 2 places, and keep it synchronised, we only need to do it in `Makefile.sources` once. It comes with some cons, as we loose the ability to simply create binaries with a different compiler. Plus, it depends strongly on `bootstrap-ocaml.sh` script which seems not easy to update. Kate wants to clean it up before.

PR #7124, opened 2026-09-01
Speed-up OpamFile.OPAM.*equal when the two parameters are strictly equal

kit-ty-kate:
The impact should be minimal but this should avoid unnecessary allocations in some cases


rjbou:
Discussed in dev meeting: looking at benchmark, it is not obvious that there is a speedup.

PR #7082, opened 2026-08-16
fix docs on conjunctions in conflicts version formula

RyanGibb:
See 

https://github.com/ocaml/opam/commit/d85e29ba13c81da0eb0f789c8e872776737a444b
https://github.com/ocaml/opam/blob/6c5857a2231604aac6da1fbc055499050391d203/src/format/opamFormula.ml#L574
https://github.com/ocaml/opam/pull/2831

PR #7127, opened 2026-09-02
Add an exhaustive test showing the behaviour of the conflicts field

kit-ty-kate:
While reviewing https://github.com/ocaml/opam/pull/7082 i realised that the `conflicts` field wasn't properly tested in our testsuite.

No problem was discovered but this guards against future regression or change of behaviour.

### DIFF
--- a/master_changes.md
+++ b/master_changes.md
@@ -107,6 +107,7 @@ users)
 
 ## Reftests
 ### Tests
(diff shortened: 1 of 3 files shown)


rjbou (approved):
I've added numbers to headings and 1 switch per section.
lgtm!

PR #7130, opened 2026-09-07
Use stdlib version of Map.update

### LINKED ISSUE #4915: OpamStd.Map.update has a slightly confusing signature and is less powerful than the stdlib one
`OpamStd.Map` defines its own `update` function. This was totally justified back when there was no such function in the standard library `Map` functor but probably is not anymore.

The signature and the documentation made me assume it would behave differently than it does, here is the doc:
```
  (** [update k f zero map] updates the binding of [k] in [map] using function
      [f], applied to the current value bound to [k] or [zero] if none *)
  val update: key -> ('a -> 'a) -> 'a -> 'a t -> 'a t
```

I thought `zero` would be used as binding directly rather than passed to `f`. Reading the documentation again with that in mind, it is correct but I naturally assumed from the signature it would use `zero` directly.

This seems a bit unfriendly as not all types can define a meaningful `zero`, e.g. list has `[]`, option has `None` but some user defined type might not and working with this is a bit painful.

The stdlib update is much more powerful, it can delete existing bindings in addition to modifying them or adding new ones and has a much better and user friendly API.

I totally understand why `OpamStd.Map.update` was defined the way it is but I think it's time to let it go.

I also understand you might not want to remove it altogether for compat reasons so maybe re-exposing the stdlib `update` under a different name so it can be used instead would also be a satisfactory solution here.

I'm happy to submit a PR if you agree that would be useful and we work out the details of how you'd like to do it!

NathanReb:
Fixes #4915


src/state/opamSwitchState.ml
```diff
@@ -111,12 +111,19 @@ let infer_switch_invariant_raw
     OpamPackage.Set.fold (fun nv dmap ->
         let deps = resolve_deps nv in
         let dmap =
-          OpamPackage.Map.update nv ((++) deps) OpamPackage.Set.empty dmap
+          OpamPackage.Map.update nv
+            (function
+              | None -> Some deps
+              | Some d -> Some (d ++ deps))
```
kit-ty-kate: it shouldn't change anything but in theory it's not exactly equivalent otherwise
```suggestion
              | Some d -> Some (deps ++ d))
```

src/state/opamSwitchState.ml
```diff
@@ -1525,8 +1535,10 @@ let reverse_dependencies st ~build ~post =
              in
              List.fold_left (fun rev_deps rev_nv ->
                  OpamStd.IntMap.update rev_nv
-                   (fun l -> Hashtbl.hash nv :: l)
-                   [] rev_deps)
+                   (function
+                     | None -> Some (Hashtbl.hash nv :: [])
```
kit-ty-kate:
```suggestion
                     | None -> Some [Hashtbl.hash nv]
```

src/state/opamGlobalState.ml
```diff
@@ -132,23 +132,25 @@ let load lock_kind =
...
+                     in
+                     Some (S (OpamStd.String.strip (String.concat "\n" ret)))
+                   with e ->
+                     OpamStd.Exn.fatal e;
+                     log "Failed to evaluate global variable %a: %a"
+                       (slog OpamVariable.to_string) v
+                       (slog Printexc.to_string) e;
+                     match previous_value with
+                     | None -> None
+                     | Some pv -> Lazy.force (fst pv))),
```
kit-ty-kate:
```suggestion
                     | Some (v, _doc) -> Lazy.force v)),
```

src/client/opamRepositoryCommand.ml
```diff
@@ -183,8 +183,8 @@ let list_all rt ~short =
         let n_repos = List.length repos in
         let acc,_ =
           List.fold_left (fun (acc,i) repo ->
-              OpamRepositoryName.Map.update repo
-                (fun s -> (Some sw, (i, n_repos))::s) [] acc,
+              let elm = Some sw, (i, n_repos) in
```
kit-ty-kate: stylistically i think adding parenthesis here helps highlight the tuple
```suggestion
              let elm = (Some sw, (i, n_repos)) in
```

kit-ty-kate (approved):
lgtm aside from the following minor stylistic comments and the missing change to `src/solver/opamBuiltinZ3.real.ml` as shown in CI.

src/state/opamSwitchState.ml
NathanReb: good catch!

src/state/opamSwitchState.ml
NathanReb: I simplified it into a `add_to_list`!

NathanReb:
Does the windows CI failure seem related to you @kit-ty-kate @rjbou ?

kit-ty-kate:
> Does the windows CI failure seem related to you @kit-ty-kate @rjbou ?

No it's not related. That's https://github.com/ocaml/opam/issues/6890

master_changes.md
```diff
@@ -129,3 +129,5 @@ users)
 ## opam-format
 
 ## opam-core
+
+  * Changed `OpamStd.Map.update` to the stdlib version which has a more flexible API and better performances [#7130 @NathanReb]
```
```suggestion
  * `OpamStd.Map.update` was changed to the stdlib version which has a more flexible API and has better performances [#7130 @NathanReb - fix #4915]
```

rjbou (approved):
thanks!
looking at dependencies jobs, this change broke odep & opam-dune-lint

PR #7131, opened 2026-09-07
Stop opening tar.gz repositories during installation unless necessary

kit-ty-kate:
During installation opam will want to copy extra-files. However since opam 2.3 (https://github.com/ocaml/opam/pull/5564), extra-files is no longer an optional field whose content would be automatically concatenated with whatever is in the files/ directory so we can simply trust that an empty extra-files field means we don't have to look into that directory.

Looking into that directory is only required in 1.x to 2.0 file format upgrades which should update the extra-files field anyway. However while testing with opam 2.5, this isn't the case anyway so we are all good.
This might be a regression from 2.3 and can be delt with later.

Avoiding a complete tar.gz read twice per package installed improves performance of installs by at least 500ms per package.

### DIFF
--- a/tests/reftests/repository-http.test
+++ b/tests/reftests/repository-http.test
@@ -364,12 +364,10 @@ The following actions will be performed:
   - install to-install 1
 
 <><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
-TAR                             Open archive ${BASEDIR}/OPAM/repo/http-repo.tar.gz
 Processing  2/3: [to-install: true]
 + true (CWD=${BASEDIR}/OPAM/install/.opam-switch/build/to-install.1)
 -> compiled  to-install.1
 -> installed to-install.1
-TAR                             Open archive ${BASEDIR}/OPAM/repo/http-repo.tar.gz
 Done.
 ### opam install extra-files-no-field | sed-cmd test
 FILE(config)                    Read ${BASEDIR}/OPAM/config in 0.000s
@@ -380,12 +378,10 @@ The following actions will be performed:
   - install extra-files-no-field 1
 
 <><> Processing actions <><><><><><><><><><><><><><><><><><><><><><><><><><><><>
-TAR                             Open archive ${BASEDIR}/OPAM/repo/http-repo.tar.gz
 Processing  2/3: [extra-files-no-field: test !]
 + test "!" "-f" "some-file" (CWD=${BASEDIR}/OPAM/install/.opam-switch/build/extra-files-no-field.1)
 -> compiled  extra-files-no-field.1
 -> installed extra-files-no-field.1
-TAR                             Open archive ${BASEDIR}/OPAM/repo/http-repo.tar.gz
 Done.
 ### opam install extra-files-legacy-no-field | sed-cmd test
 FILE(config)                    Read ${BASEDIR}/OPAM/config in 0.000s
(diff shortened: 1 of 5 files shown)


tests/reftests/repository-http.test
```diff
@@ -266,4 +293,161 @@ empat   --
...
+### openssl md5 packages/extra-files-no-field.1/files/some-file | '.*= ' -> '' >$ XMD5
+### <extrafile.sh>
+#!/bin/sh
+set -eu
+cat >> "$1" << EOF
+extra-files: ["some-file" "md5=$XMD5"]
+EOF
+### sh extrafile.sh packages/extra-files-with-field.1/opam
+### sh extrafile.sh packages/extra-files-legacy-with-field.1/opam
+### opam admin index
```
rjbou, line 345: Even if other patterns are accepted, it is better to have the `n/n.v` pattern, to have the test suite future proof
```suggestion
### <packages/to-install/to-install.1/opam>
opam-version: "2.0"
build: "true"
### <packages/extra-files-no-field.1/opam>
opam-version: "2.0"
build: [
  ["test" "!" "-f" "some-file"]
]
### <packages/extra-files-no-field/extra-files-no-field.1/files/some-file>
some file
### <packages/extra-files-legacy-no-field.1/opam>
opam-version: "1.2"
build: [
  ["test" "-f" "some-file"]
]
### <packages/extra-files-legacy-no-field/extra-files-legacy-no-field.1/files/some-file>
some file
### <packages/extra-files-with-field/extra-files-with-field.1/opam>
opam-version: "2.0"
build: [
  ["test" "-f" "some-file"]
  ["test" "!" "-f" "some-other-file"]
]
### <packages/extra-files-with-field/extra-files-with-field.1/files/some-file>
some file
### <packages/extra-files-with-field/extra-files-with-field.1/files/some-other-file>
some unlisted file
### <packages/extra-files-legacy-with-field/extra-files-legacy-with-field.1/opam>
opam-version: "1.2"
build: [
  ["test" "-f" "some-file"]
  ["test" "-f" "some-other-file"]
]
### <packages/extra-files-legacy-with-field/extra-files-legacy-with-field.1/files/some-file>
some file
### <packages/extra-files-legacy-with-field/extra-files-legacy-with-field.1/files/some-other-file>
some unlisted file
### openssl md5 packages/extra-files-no-field/extra-files-no-field.1/files/some-file | '.*= ' -> '' >$ XMD5
### <extrafile.sh>
#!/bin/sh
set -eu
cat >> "$1" << EOF
extra-files: ["some-file" "md5=$XMD5"]
EOF
### sh extrafile.sh packages/extra-files-with-field/extra-files-with-field.1/opam
### sh extrafile.sh packages/extra-files-legacy-with-field/extra-files-legacy-with-field.1/opam
### opam admin index
```

tests/reftests/repository-http.test
```diff
@@ -266,4 +293,161 @@ empat   --
 enam    --
 lima    --
 satu    --
+### :VII: Install packages from http repository
```
rjbou:
```suggestion
### :VII: Install packages from http repository and extra files handling
```

rjbou (approved):
lgtm! modulo test comments

tests/reftests/repository-http.test
kit-ty-kate: > Even if other patterns are accepted, it is better to have the `n/n.v` pattern, to have the test suite future proof

why future proof? I don't think we should change that

kit-ty-kate:
For future reference, ocaml-benchmarks does not show any improvement because the main repository is a git repository for reproducibility purpose (see [/bench.Dockerfile#L11](https://github.com/ocaml/opam/blob/9c8eba9cf4c4358977fe410abf7bad4526ac6e8a/bench.Dockerfile#L11)).

PR #7135, opened 2026-09-08
Add opam 2.6.0~rc1 to the install scripts

kit-ty-kate:


rjbou (approved):
thanks!

kit-ty-kate:
Ignoring the rest of CI given that only install-check (green) is relevant

Issue #6433, opened 2025-03-25
Release script: Generate pre-built binaries for all platforms with official support for native-code


kit-ty-kate:
For the sake of testing and completeness it would be nice to have pre-built binaries for all the platforms with official support for OCaml native-code.

On 4.14, the list of platforms with native-code support is [here](https://github.com/ocaml/ocaml/blob/4.14/README.adoc#overview).
Compared to what we have currently, those platforms are missing:
- windows-x86_32
- freebsd-i686
- openbsd-i686
- netbsd-i686
- linux-ppc32
- freebsd-arm64
- freebsd-arm32
- openbsd-arm32
- netbsd-arm32

However on 5.x/trunk, 32bit support has been removed. The list is a bit different and available [here](https://github.com/ocaml/ocaml/blob/trunk/README.adoc#overview).
Compared to what we have currently, those platforms are missing:
- omnios-x86_64 (illumos-x86_64?)
- freebsd-arm64
- openbsd-arm64
- netbsd-arm64
- linux-ppc64be

From having tried before, OmniOS is a hard platform to support (i didn't manage to compile opam).
ppc64be support seems quite shoddy compared to the little-endian version as far as Linux distributions are concerned. Also docker doesn't seem to support it.

So maybe we could at least add `freebsd-arm64` and wait for our release script to upgrade to OCaml 5.x to add `openbsd-arm64` and `netbsd-arm64`.

kit-ty-kate:
Suggestion: according to https://github.com/ocaml/ocaml/commit/077920a58406fad184b19f191d3c756e100b3c51, OCaml 4.14 will stop being LTS after OCaml 5.5. Currently our CI and pre-built binaries are built primarily with 4.14 (in addition to 5.x for the CI). Starting with the planned opam 2.6, we could stop treating 4.14 as a special release in CI and make 5.5 the version of the compiler building the pre-built binaries.

This would have for effect to remove 32bit support, however for people who need it, we could instead build a bytecode target and require people on those platform to have `ocamlrun` installed. At the moment opam is wayyy too slow when compiled to bytecode but i'm hoping to have that much improved by opam 2.6

PR #7136, opened 2026-09-08
Upgrade the lockfile to use OCaml5 by default to build opam on 64bit platforms

kit-ty-kate:
Ideally i think this should wait until the release of OCaml 5.6, but 5.5.1 should be ok in a pinch.

Doing so we're shipping both 4.14 and 5.x in the opam-full-*.tar.gz release archive and choosing which one to use on device. Users that know what they are doing can always override the `IS_64BIT_OS` variable.

Related to https://github.com/ocaml/opam/issues/6433. I've chosen to not go the bytecode path for performance reason for now.


rjbou:
Discussed in dev meeting: this is a good example of #7121, we need to change the scripts in 2 places

PR #7142, opened 2026-09-14
Simplify internal linting function 

rjbou:
Seen while reviewing #7130, queued on it


src/state/opamFileTools.ml
```diff
@@ -407,19 +407,14 @@ let t_lint ?check_extra_files ?(check_upstream=false) ?(all=false) t =
...
-          | Some last ->
-            if compare last elem = 0 then
-              Some elem, OpamStd.String.Map.update (to_str elem) ((+) 1) 1 dbl
-            else
-              Some elem, dbl
-          | None -> Some elem, dbl)
-        (None, OpamStd.String.Map.empty)
-      |> snd
+      List.fold_left (fun dbl elem ->
+              OpamStd.String.Map.update (to_str elem)
```
rjbou, line 413: we could have `OpamStd.MAP` as an argument too have a proper map, but we would need to create the map for `OpamHash.kind` locally, or in the module, just for this use case.

src/state/opamFileTools.ml
```diff
@@ -407,22 +407,14 @@ let t_lint ?check_extra_files ?(check_upstream=false) ?(all=false) t =
...
-          | None -> Some elem, dbl)
-        (None, OpamStd.String.Map.empty)
-      |> snd
+                (function None -> Some 1 | Some i -> Some (i + 1))
+                dbl)
+       OpamStd.String.Map.empty lst
+       |> OpamStd.String.Map.filter (fun _ occ -> occ > 1)
     in
     if OpamStd.String.Map.is_empty double then false, None else
       true,
```
kit-ty-kate, line 420: while we're simplifying this function, i think returning a tuple here is redundant (`fst` = `Option.is_some snd`)

src/state/opamFileTools.ml
kit-ty-kate, line 423: same here, this check is superfluous:
```suggestion
Printf.sprintf "%s: %d occurences"
elem occ)
```

PR #7147, opened 2026-09-16
GHA: auto-cancel unreachable jobs in PRs

kit-ty-kate:
I'm tired of doing it by hand and turns out there is a setting to do it automatically.
Documented in https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency

The only CI job that doesn't auto-cancel is ocaml-benchmarks but that's tracked in https://github.com/ocurrent/current-bench/issues/497


rjbou (approved):
thanks! it is very helpful to no longer have to handle that :)

PR #7148, opened 2026-09-17
Remove st.packages ++ st.installed occurences

### LINKED ISSUE #7141: `OpamSolution.fuzzy_name` seems to do unnecessary computations
The function `OpamSolution.fuzzy_name`, described as follows:

```ocaml
(** Matches package names to their existing counterparts, up to capitalisation.
    If no match exists, returns the name unchanged. *)
val fuzzy_name: 'a switch_state -> name -> name
```

It look for a matching name in the switch, modulo case. What troubles me is the actual implementation:

```ocaml
let fuzzy_name t name =
  let lname = String.lowercase_ascii (OpamPackage.Name.to_string name) in
  let match_name nv =
    lname = String.lowercase_ascii (OpamPackage.name_to_string nv)
  in
  let matches =
    OpamPackage.Set.union
      (OpamPackage.Set.filter match_name t.installed)
      (OpamPackage.Set.filter match_name t.packages)
  in
  let names = OpamPackage.names_of_packages matches in
  match OpamPackage.Name.Set.elements names with
  | [name] -> name
  | _ -> name
```

It filters both `installed` and `packages`, then makes the union of the matching sets and takes the first match. It seems to me that: `installed` should be included in `packages`. There might be some extreme cases where they are out of sync but whether this applies to the call sites of `fuzzy_name` should be easy to determine.

If that seems fine to you I suggest that `fuzzy_name` only looks up `packages` and stops at the first match rather than looking up the whole set.

Let me know what you think, I'll happily send a PR with a fix if I'm correct!

NathanReb:
Fixes #7141 

The set of installed packages is included in the set of known packages making this a costly version of `st.packages`.

These were the occurences I could find via a few `git grep` invocations but I might have missed some.

I looked up sherlocode to find similar usage in revdeps but didn't find anything conclusive, again, might have missed some, I'll try to conduct a more thoroudh investigation.

Opening the PR now also lets me rebase some other refactorings which will benefit from this.


NathanReb:
Not quite sure where to write this one in `master_changes.md`, where does this kind of API transparent refactoring usually go?

kit-ty-kate:
> Not quite sure where to write this one in `master_changes.md`, where does this kind of API transparent refactoring usually go?

The `Internal` section probably

src/client/opamSolution.ml
```diff
@@ -197,22 +197,15 @@ let fuzzy_name t name =
...
-      (OpamPackage.Set.filter match_name t.installed)
-      (OpamPackage.Set.filter match_name t.packages)
-  in
-  let names = OpamPackage.names_of_packages matches in
-  match OpamPackage.Name.Set.elements names with
-  | [name] -> name
-  | _ -> name
+  match OpamPackage.Set.find_opt match_name t.packages with
+  | None -> name
+  | Some match_ -> match_.name
```
kit-ty-kate: `match_` feels like an odd name due to the keyword status of `match`. What about simply:
```suggestion
  | Some pkg -> pkg.name
```

kit-ty-kate (approved):

kit-ty-kate:
looking at the result of ocaml-benchmarks, this seems to be giving ~4% performance gain. Nice

NathanReb:
Added the `master_changes.md` entry!

src/client/opamListCommand.ml
```diff
@@ -906,6 +905,7 @@ let info st ~fields ~raw ~where ?normalise ?(show_empty=false)
          OpamPackage.Set.iter output_package nvs)
       else
         (let choose =
+           let open OpamPackage.Set.Op in
```
rjbou: For next time, it is better to have these removals in another commit, as they are not related to the main purposte

rjbou (approved):

src/client/opamListCommand.ml
kit-ty-kate: I took the opportunity from the required rebase to revert these changes as they are unnecessary

PR #7150, opened 2026-09-17
Add 2.6.0 to the install scripts and documentation

kit-ty-kate:


kit-ty-kate:
GHA's install-check and Doc-Linux run are green. Merging without waiting for the rest

PR #7154, opened 2026-09-21
Remove unnecessary set operations in `OpamSwitchState.universe`

NathanReb:
These are a legacy from a time where `requested` was a set of package names. The move to a package set was done in a conservative manner in #4796 and kept `requested_all_pkgs` as the set of all versions of packages appearing in `requested`.

This has two inconvenients:
1. the universe is prepared for a larger set than what's actually required and while it seems to be relatively harmless for the application of `with-test` and other such filters, it is also used to decide which packages need to be reinstalled and can cause unwanted recompilation of packages.
2. In quite a few places we actually pass `st.packages` as `requested` which makes those set operation a costly identity operation.

I added a test to exhibit the reinstall behaviour, thanks to @AltGr's comment on the original PR: https://github.com/ocaml/opam/pull/4796#discussion_r700998546


tests/reftests/universe-requested-all-pkgs.test
```diff
@@ -0,0 +1,61 @@
...
+build: [["echo" "depb installed!"]]
+### <pkg:foo.1>
+opam-version: "2.0"
+depends: [ "depa" ]
+### <pkg:foo.2>
+opam-version: "2.0"
+depends: [ "depb" ]
+### opam switch create bar --empty
+### : We install foo.1 which logically requires depa.1 to be installed
+### opam install foo.1 -y
```
kit-ty-kate: in the tests we tend to set `OPAMYES=1` at the top instead of using `--yes`, this makes things easier to read and to modify imo

src/state/opamSwitchState.ml
```diff
@@ -1018,19 +1018,15 @@ let universe st
...
   let chrono = OpamConsole.timer () in
-  let names = OpamPackage.names_of_packages requested in
-  let requested_allpkgs =
-    OpamPackage.packages_of_names st.packages names
-  in
   let env =
     package_env_t st
       ~force_dev_deps ~test ~doc ~dev_setup
-      ~requested_allpkgs
+      ~requested_allpkgs:requested
```
kit-ty-kate: since `request_allpkgs` is no longer a different concept compared to `requested` i think it would be worth to rename these labelled arguments

kit-ty-kate (approved):
That looks pretty sound. One small request but otherwise looks great

PR #7159, opened 2026-09-22
Rewrite inefficient set <-> map conversions

NathanReb:
I'm opening this as is because I found this inefficient conversion and didn't want to squeeze it into an unrelated PR.

If it seems acceptable to merge just this one then let's do it, otherwise I'm happy to try and find other such operations to include as part of this PR.


kit-ty-kate (approved):

kit-ty-kate:
Ignoring disk failure on ocaml-benchmarks

